PR #3, opened 2018-11-14
Removed type on variables definition with direct value assign.

abdullahselek:


xwu:
It looks like this project explicitly annotates the type of static and instance members, which is a valid style choice with its own advantages.

benlangmuir:
@xwu is correct. I am cautious about relying on inference in instance and static properties, because I believe it may slow down type-checking non-primary files.  I also like being explicit about the type of anything public, and at that point it's nice to just be consistent for everything internal as well.

I could be convinced to change this in the future, but I would want to see evidence that swift style is standardizing on implicit types _in API_ and that my type-checking performance concern is unfounded or can be mitigated effectively.

aciidb0mb3r:
I think @jrose-apple said it's not really a good idea to rely on type inference for properties.

jrose-apple:
@benlangmuir gave most of the reasons why I believe that. If the repeated type is bothersome, one option is to use `.init` instead:

```
public static let quickFix: CodeActionKind = .init(rawValue: "quickfix")
```

But this would also be a style choice that the maintainers would have to agree is better.

akyrtzi:
I discussed with @benlangmuir and he'll add a style guide so it's clear to contributors.

PR #5, opened 2018-11-15
Add "Document Symbols" to status feature list

benlangmuir:
This is very similar to sourcekitd's "document structure", which can provide a hierarchical view of entities in the current file.


akyrtzi:
Nitpick, could you also change 'Open Quickly' to the LSP name, 'Workspace Symbols' I think ?

benlangmuir:
Sure: https://github.com/apple/sourcekit-lsp/pull/6

PR #13, opened 2018-11-20
[SKCore] If 'Platform.currentPlatform' is nil use 'so' as library extension

akyrtzi:
Platform only detects the 'debian' linux flavor so if it is nil it's most likely an unknown linux flavor

### DIFF
--- a/Sources/SKCore/Toolchain.swift
+++ b/Sources/SKCore/Toolchain.swift
@@ -86,7 +86,8 @@ public final class Toolchain {
       foundAny = true
     }
 
-    let dylibExt = Platform.currentPlatform?.dynamicLibraryExtension ?? "dylib"
+    // If 'currentPlatform' is nil it's most likely an unknown linux flavor.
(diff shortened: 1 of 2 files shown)


pvieito:
`LinuxFlavor` should include an `unknown` case so when `Process.checkNonZeroExit(args: "uname")` returns `linux` we can at least return `.linux(. unknown)` in any case.

akyrtzi:
> `LinuxFlavor` should include an `unknown` case so when `Process.checkNonZeroExit(args: "uname")` returns `linux` we can at least return `.linux(. unknown)` in any case.

That's a good idea but I think the change here is good to have regardless of your suggested improvement to `Platform`.
An unknown platform is more likely some unix-like platform, `.so` is a better fallback extension than `.dylib`.

pvieito:
Ah, yes, of course!

akyrtzi:
I'll tentatively merge this to allow the service to work on non-debian linux.

PR #18, opened 2018-11-24
Cleanup syntax

yhkaplan:
This is just a minor cleanup of the following areas:
- Preferring Void to ()
- Use discardableResult to remove _ =
- Remove unneeded (_)
- Use trailing closure with Kevin's syntax
- Use trailing closures when possible
- Fix CamelCase in one area
- Use for where

Please tell me if you disagree with any differences in style, etc.


Sources/LanguageServerProtocol/CodeActionKind.swift
```diff
@@ -22,25 +22,25 @@ public struct CodeActionKind: RawRepresentable, Codable, Hashable {
   }
 
   /// QuickFix action, such as FixIt.
-  public static let quickFix: CodeActionKind = CodeActionKind(rawValue: "quickfix")
```
LuizZak: Removal of type annotations on public properties has been suggested before, and the consensus seems to be to keep them explicit: https://github.com/apple/sourcekit-lsp/pull/3#issuecomment-438743769

yhkaplan: Fair enough. I'll go ahead and just remove it on internal properties then.

benlangmuir:
> Remove unneeded (_)

What's the motivation here? I don't have a strong preference, but I weakly prefer to be explicit about ignoring the payload.

hartbit:
Got what it’s worth, in SwiftPM, we don’t use trailing closure whenever possible, only when they represent a non-functional function. Some people call it the Rule of Kevin and you can find it summarized at the top of the following page: https://ericasadun.com/2015/11/17/a-handful-of-swift-style-rules-swiftlang/

yhkaplan:
@benlangmuir

>> Remove unneeded (_)

> What's the motivation here? I don't have a strong preference, but I weakly prefer to be explicit about ignoring the payload.

The motivation is to reduce the amount of things the programmer has to keep in mind while reading a particular switch statement, focusing more on what is being accomplished in that section. I figure that if the programmer wants to see about the payload, then they can often have to examine the enum anyways to see what values exist and what properties might be available.

yhkaplan:
Merge conflict: Toolchain.swift (https://github.com/apple/sourcekit-lsp/pull/25/files)

Use theirs, adding discardableResult
```
<<<<<<< HEAD
  public convenience init(
    identifier: String,
    displayName: String,
    xctoolchainPath path: AbsolutePath,
    fileSystem fs: FileSystem = localFileSystem
  ) {
    self.init(identifier: identifier, displayName: displayName, path: path)
    searchForTools(path: path, fileSystem: fs)
  }

  /// Search `path` for tools, returning true if any are found.
  @discardableResult
  func searchForTools(path: AbsolutePath, fileSystem fs: FileSystem = localFileSystem) -> Bool {
=======
    if !searchForTools(path, fileSystem) {
      return nil
    }
  }

  /// Search `path` for tools, returning true if any are found.
  func searchForTools(_ path: AbsolutePath, _ fs: FileSystem = localFileSystem) -> Bool {
>>>>>>> a7999dd3d1c4f3776d00018a8684f7be0520e093
```


Platform was moved to separate file, so use theirs and delete the `(_)` in Platform.swift
```
<<<<<<< HEAD

extension Platform {
  var dynamicLibraryExtension: String {
    switch self {
    case .darwin: return "dylib"
    case .linux: return "so"
    }
  }
}
=======
>>>>>>> a7999dd3d1c4f3776d00018a8684f7be0520e093
```

### Changes made to theirs
```diff
diff --cc Sources/SKCore/Toolchain.swift
index 25a9c9d,391db6f..0000000
--- a/Sources/SKCore/Toolchain.swift
+++ b/Sources/SKCore/Toolchain.swift
@@@ -54,8 -118,7 +118,8 @@@ extension Toolchain
    }

    /// Search `path` for tools, returning true if any are found.
+  @discardableResult
    func searchForTools(_ path: AbsolutePath, _ fs: FileSystem = localFileSystem) -> Bool {
      return
        searchForTools(binPath: path, fs) ||
        searchForTools(binPath: path.appending(components: "bin"), fs) ||
diff --git a/Sources/SKSupport/Platform.swift b/Sources/SKSupport/Platform.swift
index 0fd1f32..c97efb9 100644
--- a/Sources/SKSupport/Platform.swift
+++ b/Sources/SKSupport/Platform.swift
@@ -18,7 +18,7 @@ extension Platform {
   public var dynamicLibraryExtension: String {
     switch self {
     case .darwin: return "dylib"
-    case .linux(_): return "so"
+    case .linux: return "so"
     }
   }
 }

```

yhkaplan:
Thanks for the review! I made the corrections. What do you think?

benlangmuir:
Rather than add and then revert changes, please force push to this PR with the commits removed.  Ideally please incorporate the merge fixes back to the original commits as well, but if that turns out to be hard I'm okay with leaving that.

Otherwise LGTM.

yhkaplan:
@benlangmuir Thanks for review! I fixed the commit history.

benlangmuir:
Looks good, thanks!

PR #24, opened 2018-11-30
Add "textDocument/foldingRanges"

rockbruno:
Hey there, I was reading about the LSP yesterday and wanted to take a shot at implementing the folding requests 🙂 

Although I'm unsure how Xcode implemented it, some spelunking makes me believe you guys are using the`bodyoffsets` from `editor.open` requests, so that's how I tackled it.

The first commit works for perfectly idented bodies but stills fails for weird syntaxes like:

```
func foo() {

return }
```

Since Xcode can treat these, I'll implement the full request to treat the edge cases.

screen shot 2018-11-30 at 10 12 18 am


rockbruno:
I've removed the WIP tag and rebased the commits - It's now supporting all bodies and comments recognized by SourceKit. My only pet peeve is that it acts slightly different than the default folding behaviour from VS Code. While VSCode will fold something with the bracket on the next line:

```
func foo() { ...
}
```
...this implementation will get rid of it:
```
func foo() { ...
```
It does seem to be in line with how Xcode folds it though, and VS Code fails to fold the edge case I first mentioned which makes me think this might not be the worst approach after all.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -547,6 +548,81 @@ extension SwiftLanguageServer {
...
+      var structureStack: [SKResponseArray] = [substructure]
+      while let substructure = structureStack.popLast() {
+        substructure.forEach { _, value in
+          if let offset: Int = value[self.keys.bodyoffset],
+             let length: Int = value[self.keys.bodylength],
+             length > 0,
+             let start: Position = snapshot.positionOf(utf8Offset: offset),
+             let end: Position = snapshot.positionOf(utf8Offset: offset + length) {
+            let range = FoldingRange(startLine: start.line,
+                                     startCharacter: offset,
```
benlangmuir: I'm pretty sure the start/index "character" are columns, not absolute positions.  Otherwise the comment in the spec about defaulting to the length of the line make no sense (also the line+character terminology matches what's in "Position", which is definitely a column index).

So these should be start.utf16index and end.utf16index I think.

Could this be related to what you mentioned in the description about how the line folding didn't match vscode's builtin folding?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -547,6 +548,81 @@ extension SwiftLanguageServer {
...
+        if let kind: sourcekitd_uid_t = value[self.keys.kind],
+           kind == self.values.syntaxtype_comment,
+           let offset: Int = value[self.keys.offset],
+           let start: Position = snapshot.positionOf(utf8Offset: offset),
+           let length: Int = value[self.keys.length],
+           // SourceKit marks the end of a comment as the first non-comment character
+           // after it, so we subtract one to get the real comment range.
+           let end: Position = snapshot.positionOf(utf8Offset: offset + length - 1) {
+          let range = FoldingRange(startLine: start.line,
+                                   startCharacter: offset,
```
benlangmuir: See my comment below about start/end "characters".

akyrtzi: > I need to think about this more, but my current feeling is to accept the current (open a new document ever time) approach and move to swift syntax in the future. /cc @akyrtzi for his thoughts on this question.

Using sourcekitd for syntactic functionality is an evolutionary dead-end, we'll adopt SwiftSyntax at some point. So I wouldn't worry about making it efficient, keep it simple.
The tests you'll add for this will be useful to validate that SwiftSyntax integration works as expected later on, make sure you have tests that check the behavior when edits occur as well.

akyrtzi:
> I need to think about this more, but my current feeling is to accept the current (open a new document ever time) approach and move to swift syntax in the future. /cc @akyrtzi for his thoughts on this question.

Using sourcekitd for syntactic functionality is an evolutionary dead-end, we'll adopt SwiftSyntax at some point. So I wouldn't worry about making it efficient, keep it simple.
The tests you'll add for this will be useful to validate that SwiftSyntax integration works as expected later on, make sure you have tests that check the behavior when edits occur as well.

benlangmuir:
> make sure you have tests that check the behavior when edits occur as well.

Folding ranges doesn't have any real interaction with edits, so it doesn't seem like the right place to test this to me.  When we add swift syntax support I expect we'll want to test the more specific goal of keeping a shared syntax tree in sync with edit requests.

akyrtzi:
> > make sure you have tests that check the behavior when edits occur as well.
> 
> Folding ranges doesn't have any real interaction with edits, so it doesn't seem like the right place to test this to me. When we add swift syntax support I expect we'll want to test the more specific goal of keeping a shared syntax tree in sync with edit requests.

SGTM.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -547,6 +548,81 @@ extension SwiftLanguageServer {
...
+
+    guard let snapshot = documentManager.latestSnapshot(req.params.textDocument.url) else {
+      log("failed to find snapshot for url \(req.params.textDocument.url)")
+      req.reply(nil)
+      return
+    }
+
+    let skreq = SKRequestDictionary(sourcekitd: sourcekitd)
+    skreq[keys.request] = requests.editor_open
+    skreq[keys.name] = "IBSwiftAnnotationProvider:" + snapshot.document.url.path
```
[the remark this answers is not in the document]
rockbruno: Hahaha I left it in because I was trying to cover the day Xcode implements the LSP. I assumed it was important for this request to continue sending it. Is just the URL enough or should we come up with another prefix?

rockbruno:
Thank you for your inputs @benlangmuir and @akyrtzi. I was under the impression that SourceKit had some form of internal cache given that I get a few duplicate editor.open requests after booting a single-file Xcode project, so I blindly went with the always-request approach. I'll implement the necessary changes shortly.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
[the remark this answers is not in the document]
benlangmuir: Use another prefix, otherwise it will conflict with the normal `textDocument/didOpen` code path.  Something like "FoldingRanges:" would be fine.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
rockbruno: That makes a lot of sense! I got confused by the name they used in the protocol.
It didn't changed that behaviour though. I _assume_ they're ignoring these properties completely in favor of folding the entire line, because if you have something like this:
```
func foo() {
    func bar() {

    } //comment
}
```
and subtract one from the end offset (so it ends before the closing brace), `foo` will fold like the builtin version, but `bar` loses the brace and the comment. That would explain why things worked even though the characters were wrong.

Sources/LanguageServerProtocol/FoldingRange.swift
```diff
@@ -0,0 +1,41 @@
...
+
+/// A single folding range result.
+public struct FoldingRange: ResponseType, Hashable {
+
+  /// The zero-based line number from where the folded range starts.
+  public var startLine: Int
+
+  /// The zero-based character offset from where the folded range starts.
+  /// If not defined, defaults to the length of the start line.
+  public var startCharacter: Int?
```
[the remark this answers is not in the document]
rockbruno: The builtin behavior (in .swift files at least) appears to be to fold everything in the lines after and before the braces:

```
func foo() { 1111 //intact
    2222 //folded
3333 } //intact
```
With SourceKit setting startLine to 1 and endLine to 3, the behavior is:
```
func foo() { 1111 //intact
    2222 //folded
3333 } //folded
```
Which is the behavior the comments define. The problem is that they appear to be ignoring the character properties, as I was unable to make `func foo() {` stay intact but fold `1111`. I'm investigating if that's really the case.

Sources/LanguageServerProtocol/FoldingRange.swift
```diff
@@ -18,24 +18,34 @@ public struct FoldingRange: ResponseType, Hashable {
 
   /// The zero-based character offset from where the folded range starts.
   /// If not defined, defaults to the length of the start line.
-  public var startCharacter: Int?
+  public var utf16startIndex: Int?
```
benlangmuir: Should be an upper-case S in `utf16StartIndex`.  Same with E in end.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -547,6 +548,81 @@ extension SwiftLanguageServer {
...
+      syntaxMap.forEach { _, value in
+        if let kind: sourcekitd_uid_t = value[self.keys.kind],
+           kind == self.values.syntaxtype_comment,
+           let offset: Int = value[self.keys.offset],
+           let start: Position = snapshot.positionOf(utf8Offset: offset),
+           let length: Int = value[self.keys.length],
+           // SourceKit marks the end of a comment as the first non-comment character
+           // after it, so we subtract one to get the real comment range.
+           let end: Position = snapshot.positionOf(utf8Offset: offset + length - 1) {
+          let range = FoldingRange(startLine: start.line,
```
benlangmuir: Style: When you have multiple lines of `if let` bindings and conditions, please put the final `{` on its own line. Otherwise it's hard to see where the condition ends and the following code begins (I read this wrong twice now).  There's a similar case below in the structure handling code.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -547,6 +548,81 @@ extension SwiftLanguageServer {
...
+
+      syntaxMap.forEach { _, value in
+        if let kind: sourcekitd_uid_t = value[self.keys.kind],
+           kind == self.values.syntaxtype_comment,
+           let offset: Int = value[self.keys.offset],
+           let start: Position = snapshot.positionOf(utf8Offset: offset),
+           let length: Int = value[self.keys.length],
+           // SourceKit marks the end of a comment as the first non-comment character
+           // after it, so we subtract one to get the real comment range.
+           let end: Position = snapshot.positionOf(utf8Offset: offset + length - 1) {
```
benlangmuir: I don't think this subtraction should be necessary.  Both LSP and sourcekitd use half-open ranges (the range includes the start position and goes up to but *not* including the end position).  They should both expect the end offset to be one-past the last character of the comment AFAICT.

Heres what the spec says about the `Range` type:
> A range in a text document expressed as (zero-based) start and end positions. A range is comparable to a selection in an editor. Therefore the end position is exclusive.

And the comments in the spec about FoldingRange's endCharacter says that it defaults to the line length, which is also one-past the end.

Tests/SourceKitTests/SKLocalSwiftTests.swift
```diff
@@ -217,6 +217,73 @@ final class SKLocalSwiftTests: XCTestCase {
...
+       c3
+      */
+      var abc: Int
+
+      func test(a: Int) {
+        guard a > 0 else { return }
+        self.abc = a
+      }
+    }
+    """)))
```
benlangmuir: I'd like to see some tests that include a doc comment, and maybe some inline /**/ comments within a line.  If you can find cases where a comment right before or after another range end up with the same start or end point, that would be interesting to have tests for as well.

Sources/LanguageServerProtocol/FoldingRange.swift
```diff
@@ -0,0 +1,51 @@
...
+
+/// A single folding range result.
+public struct FoldingRange: ResponseType, Hashable {
+
+  /// The zero-based line number from where the folded range starts.
+  public var startLine: Int
+
+  /// The zero-based character offset from where the folded range starts.
+  /// If not defined, defaults to the length of the start line.
+  public var utf16StartIndex: Int?
```
benlangmuir: Should be `startUTF16Index` and `endUTF16Index` so that the connection with `startLine` and `endLine` is clearer.  I think I told you the wrong thing when I suggested you rename these, sorry about that!

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
rockbruno: The reason I added that is because a simple comment like `//foo` will end up with a different endLine, as the "end" of the comment here is the start of the next line. It was causing VSCode to fold whatever was on the next line even though the comment only really covers a single line. It might have to do with the supposed ignored properties I mentioned earlier in this case, but I wasn't sure how to stop it other than subtracting it. What do you think?

benlangmuir: Hmm, in that case subtracting 1 _before_ finding the Position doesn't seem like a good fix, since it only does the right thing if the end position is on a different line.  Maybe you could compute the correct Position, but then check if the end of the range is at column 0, and in that case change it to the previous line but with column number = line length?  It's the same position in an abstract sense, but if something is only looking at the line number and not the column it will be more correct.

benlangmuir: To clarify, I mean change `Position(line: N, utf16Index: 0)` to `Position(line: N-1, utf16Index: <length of line N-1>)`.  You can get the line length from the snapshot's LineTable.

rockbruno: Definitely sounds better! To be fair, I just noticed the same problem exists if you have something right after a brace, like a promise:
```
foo().then {
   //
}.then {
}
```
Folding the first block would also fold the second `then` if the client only supports folding whole lines. It's the same problem, although the comment one definitely looks weirder. If you're right about the end position being expected to be one character past then perhaps we don't need to treat this at all since the problem will always exist for bodies.

rockbruno:
Alright! I think I finally cracked the issue this time @benlangmuir . It seems that VSCode really only supports line folding:
```
    fillClientCapabilities(capabilites) {
        let capability = ensure(ensure(capabilites, 'textDocument'), 'foldingRange');
        capability.dynamicRegistration = true;
        capability.rangeLimit = 5000;
        capability.lineFoldingOnly = true;
    }
```
[Their models don't even check the offsets.](https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L3696)

In regards to why the last line was getting folded, [their tests seem to show that they're not supposed to be part of the range.](https://github.com/Microsoft/vscode/blob/957518fdffe5b62262e957e2f942e373f825312a/src/vs/editor/contrib/folding/test/syntaxFold.test.ts#L34) I wasn't sure if that's how the LSP is supposed to work but I've coded the feature around it just in case.

I've removed the code that was subtracting the offsets and instead starting passing the client's capabilities to the server. If they only support line folding, I ignore the last line, otherwise I send the real lines and offsets. This makes it work exactly like the built-in behaviour in VSCode, although I'm still not sure if it would behave correctly in an IDE that can fold arbitrary ranges (as in, do we need to ignore the last character?)

I also added support to doc comments, but since SourceKit breaks them into several components I've had to make the response's merge successive comments into one. I took a bit of freedom to make it work so I think you'll want to take a look at it again.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -519,6 +524,117 @@ extension SwiftLanguageServer {
...
+    }
+
+    // FIXME: cancellation
+    _ = handle
+  }
+
+  func addFoldingRange(offset: Int, length: Int, kind: FoldingRangeKind? = nil, in snapshot: DocumentSnapshot, toArray ranges: inout [FoldingRange]) {
+    let capabilities = clientCapabilities.textDocument?.foldingRange
+    if let rangeLimit = capabilities?.rangeLimit, ranges.count == rangeLimit {
+      return
```
rockbruno: Thinking about it now, I should've halted the whole request if the limit was reached.

Tests/SourceKitTests/LocalSwiftTests.swift
```diff
@@ -220,6 +220,115 @@ final class LocalSwiftTests: XCTestCase {
     XCTAssertNotEqual(after, selfDot)
   }
 
+  func testFoldingRanges() {
```
rockbruno: I should write tests to see how the different capabilities can affect the response, but I wasn't sure how to approach it here.

benlangmuir: 👍 I suggest creating a new test class that duplicates the TestServer setup from this class, but moves the InitializeRequest out of `setUp` and into the individual test methods.  It can just go at the bottom of this file for now, and if it grows we can move it later.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
benlangmuir: That seems like a good idea to me; would be interesting to know what other language servers do.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -519,6 +524,117 @@ extension SwiftLanguageServer {
...
+
+      syntaxMap.forEach { _, value in
+        if let kind: sourcekitd_uid_t = value[self.keys.kind],
+           kind.isCommentKind(self.values),
+           let offset: Int = value[self.keys.offset],
+           let length: Int = value[self.keys.length]
+        {
+          if let comment = currentComment {
+            if offset == comment.offset + comment.length {
+              currentComment?.length += length
```
benlangmuir: Cannot be nil, so please use `currentComment!` so it will complain loudly if we ever break this invariant.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -519,6 +524,117 @@ extension SwiftLanguageServer {
...
+      // Merge successive comments into one big comment by adding their lengths.
+      var currentComment: (offset: Int, length: Int)? = nil
+
+      syntaxMap.forEach { _, value in
+        if let kind: sourcekitd_uid_t = value[self.keys.kind],
+           kind.isCommentKind(self.values),
+           let offset: Int = value[self.keys.offset],
+           let length: Int = value[self.keys.length]
+        {
+          if let comment = currentComment {
```
benlangmuir: Could we simplify this `if` a bit:
```swift
if let comment = currentComment, comment.offset + comment.length == offset {
  currentComment!.length += length
  continue
}

if let comment = currentComment != nil {
  self.addFoldingRange(...)
}

currentComment = (offset: offset, length: length)
```

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -519,6 +524,117 @@ extension SwiftLanguageServer {
...
+  }
+
+  func addFoldingRange(offset: Int, length: Int, kind: FoldingRangeKind? = nil, in snapshot: DocumentSnapshot, toArray ranges: inout [FoldingRange]) {
+    let capabilities = clientCapabilities.textDocument?.foldingRange
+    if let rangeLimit = capabilities?.rangeLimit, ranges.count == rangeLimit {
+      return
+    }
+    guard let start: Position = snapshot.positionOf(utf8Offset: offset),
+          let end: Position = snapshot.positionOf(utf8Offset: offset + length) else {
+      return
```
benlangmuir: Worth logging this failure at `.warning` level.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -519,6 +524,117 @@ extension SwiftLanguageServer {
...
+    }
+    let range: FoldingRange
+    // If the client only supports folding full lines, ignore the end character's line.
+    if capabilities?.lineFoldingOnly == true {
+      let lastLineToFold = end.line - 1
+      if lastLineToFold <= start.line {
```
benlangmuir: So you can never fold a single line?  Is that expected?

rockbruno: If the client only supports folding entire lines but the range starts and ends at the same line, then you'll certainly end up folding things outside the range, such as the braces themselves. 

VSCode itself ignores the lines that start/end a range as it folds (startLine+1...endLine), with startLine being used only as a place to display the fold button and endLine being the line before the closing brace, so a range that starts and ends at the same line has no effect.

Tests/SourceKitTests/LocalSwiftTests.swift
rockbruno: That did the trick perfectly. Thanks!

benlangmuir:
LGTM!  Congratulations on landing our first new request since we open-sourced! 🎉

rockbruno:
Thanks for the patience @benlangmuir! Feel free to ping me once SwiftSyntax gets in the picture, I would love to assist in porting this and other features in any way I can.

benlangmuir:
@rockbruno we should update the README [status table](https://github.com/apple/sourcekit-lsp#status).  Would you like to do the honours?

rockbruno:
Absolutely! https://github.com/apple/sourcekit-lsp/pull/45 🎉

PR #27, opened 2018-12-02
[build] Fix warnings about Foundation.Process on Linux

benlangmuir:
launchPath and launch are deprecated. Use exectuableURL and run() when
available.

### DIFF
--- a/Sources/SKSupport/Path.swift
+++ b/Sources/SKSupport/Path.swift
@@ -1,0 +1,22 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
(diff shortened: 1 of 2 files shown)


Sources/SKSupport/Path.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Basic
+import Foundation
+
+extension AbsolutePath {
+
+  /// File URL created from the normalized string representation of the path.
+  public var asURL: URL {
```
krzyzanowskim: I'd say this extension should go up to SwiftPM. I use it everywhere by myself ;-)
https://github.com/apple/swift-package-manager/pull/1894/commits

benlangmuir: Note: I merged this as is, but will remove our copy of this API once we update to a new enough swiftpm.

PR #32, opened 2018-12-05
[swiftpm] Add target triple string to build directory path

benlangmuir:
We were previously depending on the symlink `.build/debug -> .build/<TRIPLE>/debug`, which mostly works, but has subtle inconsistencies that can sometimes break loading clang modules from the build.  Note: the path uses the triple string *without* a version.  That's what swiftpm does, and it's intentional.

rdar://46427413


aciidb0mb3r:
No reason. We can make this implicit in the build params. Can you file a bug or maybe a PR ;)

PR #50, opened 2018-12-14
[Folding nit] Ignore the structure stack after the limit

rockbruno:
Just a little cleanup, with the limit being checked on the loop we don't have to kill the stack anymore. Should make the logic easier to read.


benlangmuir:
Good idea

PR #58, opened 2018-12-27
[WIP] Support workspace folders

krzyzanowskim:
- [x] Adds support of multiple root paths of a workspace
- [ ] Update root paths

It's half-baked. Most of the model is there. Kind of rabbit hole and not even very urgent feature after all.


krzyzanowskim:
something went wrong with my push, I'll create new PR

PR #59, opened 2019-01-05
Add instructions for Vim 8

dduan:
Add integration instructions for Vim 8, using https://github.com/prabirshrestha/vim-lsp.

Also inluded the simplest, not-really-sourcekit-lsp-specific example for
triggering autocomplete.

### DIFF
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -72,6 +72,35 @@ You will need the path to the `sourcekit-lsp` executable and the Swift toolchain
 
 There is an Emacs client for SourceKit-LSP in the [main Emacs LSP repository](https://github.com/emacs-lsp/lsp-sourcekit).
 
+## Vim 8
+
+Install [vim-lsp](https://github.com/prabirshrestha/vim-lsp). In your `.vimrc`, configure vim-lsp to use
+sourcekit-lsp for Swift source files like so:
+
+```
+if executable('sourcekit-lsp')
+    au User lsp_setup call lsp#register_server({
+        \ 'name': 'sourcekit-lsp',
+        \ 'cmd': {server_info->['sourcekit-lsp']},
+        \ 'whitelist': ['swift'],
+        \ })
+endif
+```
+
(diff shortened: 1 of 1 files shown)


Editors/README.md
```diff
@@ -72,6 +72,37 @@ You will need the path to the `sourcekit-lsp` executable and the Swift toolchain
...
+    au User lsp_setup call lsp#register_server({
+        \ 'name': 'sourcekit-lsp',
+        \ 'cmd': {server_info->['sourcekit-lsp']},
+        \ 'whitelist': ['swift'],
+        \ })
+endif
+```
+
+(…assuming `sourckit-lsp` is in your PATH variable, otherwise replace `'sourcekit-lsp'` with path to your
+command location).
```
benlangmuir: Missing a "the".  How about: "with the path to the sourcekit-lsp executable"?

Editors/README.md
```diff
@@ -72,6 +72,37 @@ You will need the path to the `sourcekit-lsp` executable and the Swift toolchain
...
+
+That's it! As a test, open a source file in an Xcode project, put cursor on top of a symbol in normal mode and
```
benlangmuir: Let's say something like "open a swift file" and not mention Xcode.  We don't directly support Xcode project settings anyway at this point.

Editors/README.md
```diff
@@ -72,6 +72,37 @@ You will need the path to the `sourcekit-lsp` executable and the Swift toolchain
...
+
+There are many Vim solutions for code completion. For instance, you may want to use LSP for omnifunc:
+
+```
+autocmd FileType swift setlocal omnifunc=lsp#complete
+```
+
+With this added in `.vimrc`, you can use `<c-x><c-o>` in insert mode to trigger sourcekit-lsp completion.
+
+(Please contribute instructions for other autocomplete methods if you can!)
```
benlangmuir: I'd rather not have this last bit about contributions.  If there are good reasons to add instructions for an alternative autocomplete method, we can talk about it, but as a general rule I would rather keep this document focused on setting up sourcekit-lsp.

dduan:
@benlangmuir addressed feedback

PR #62, opened 2019-01-08
Support for window/showMessage(Request)

DavidGoldman:
This allows the server to display a message in the client and
optionally gives the client the ability to select an action item.

As a follow up, I was thinking of adding `LogOptions` to `Logging`, one of which enables showing some log messages to the client, thoughts?

### DIFF
--- a/Sources/LanguageServerProtocol/ShowMessageRequest.swift
+++ b/Sources/LanguageServerProtocol/ShowMessageRequest.swift
@@ -1,0 +1,75 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Notification from the server containing a message for the client to display.
+///
+/// - Parameters:
+///   - type: The kind of message.
+///   - message: The contents of the message.
+///   - actions: Action items which the user may select (up to 1).
+///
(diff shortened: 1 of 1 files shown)


Sources/LanguageServerProtocol/ShowMessageRequest.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Notification from the server containing a message for the client to display.
```
benlangmuir: This comment is for the notification; should be updated.

benlangmuir:
> As a follow up, I was thinking of adding LogOptions to Logging, one of which enables showing some log messages to the client, thoughts?

Do you mean options from the user/editor, or options passed to a `log()` call in the server?  What would be the criteria for showing the message to the user?  Do you have some concrete example(s) of when you think this should happen?  I'm skeptical about having the language server directly popping up alerts in general.

Sources/LanguageServerProtocol/ShowMessageRequest.swift
DavidGoldman: Done

DavidGoldman:
> > As a follow up, I was thinking of adding LogOptions to Logging, one of which enables showing some log messages to the client, thoughts?
> 
> Do you mean options from the user/editor, or options passed to a `log()` call in the server? What would be the criteria for showing the message to the user? Do you have some concrete example(s) of when you think this should happen? I'm skeptical about having the language server directly popping up alerts in general.

Options passed to a `log()` call on the server which would send it over to the client for display (and  log it locally). I don't expect it to be the case normally but I think in some events (long-running process, unexpected circumstance) it's good to give the client some indication of what is going on.

PR #64, opened 2019-01-09
folderWorkspaces support

krzyzanowskim:
- A server maintains an array of `Workspace`
- Client may add/remove workspaces as defined here: https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders
- Server reports capability
- `Workspace.Configuration` holds the configuration values
- Added more workspace related, but not used here models (request, notification) to the protocol.
- Special purpose `noRootWorkspace` instance is used to collect all the opened documents without an actual workspace


krzyzanowskim:
It's done. Ready to review.

krzyzanowskim:
That's a valid question. I guess I assumed that it's up to the client to close the files first, then remove workspace - it was too naive. Most likely DocumentManager should close all the files and post the notifications. Will do that.

krzyzanowskim:
Second thought. The client can't be notified what documents are closed (no such notification). The only DidCloseDocument is from client -> server. If the server will close documents, and the client won't... it will be out of sync. Either way, the client has to close the documents before removing the workspace folder.

benlangmuir:
There's no requirement that documents be in any workspace folder, is there?

krzyzanowskim:
no, there is no. I'll try to transfer opened documents to the "no workspace" workspace - this should do the trick.

Sources/LanguageServerProtocol/Configuration.swift
```diff
@@ -10,6 +10,30 @@
...
 
+/// Request from the server to the client to fetch configuration settings from the client.
+///
+/// Clients that support workspace folders should set the `workspaceFolders` client capability.
+///
+/// - Note: the format of the settings is implementation-defined.
+///
+/// - Returns: Configuration settings. If the client can't provide a configuration setting for
+///            a given scope then nil need to be present in the returned array.
+public struct ConfigurationRequest: RequestType, Hashable {
```
benlangmuir: How is this related to workspaceFolders?

Sources/LanguageServerProtocol/ClientCapabilities.swift
```diff
@@ -42,6 +42,13 @@ public struct WorkspaceClientCapabilities: Hashable, Codable {
     /// Whether the client supports the `documentChanges` field of `WorkspaceEdit`.
     public var documentChanges: Bool? = nil
 
+    /// The resource operations the client supports. Clients should at least
+    /// support 'create', 'rename' and 'delete' files and folders.
+    public var resourceOperations: [ResourceOperationKind]? = nil
```
benlangmuir: How are these related to workspaceFolders?

Sources/LanguageServerProtocol/WorkspaceEdit.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public struct WorkspaceEdit: Codable {
```
benlangmuir: How is this related to workspaceFolders?

Sources/SourceKit/Workspace.swift
```diff
@@ -27,19 +27,32 @@ import SKSwiftPMWorkspace
...
+
+    init(rootPath: AbsolutePath? = nil, buildSettings: BuildSystem, index: IndexStoreDB? = nil, buildSetup: BuildSetup = .default) {
+      self.rootPath = rootPath
+      self.buildSettings = buildSettings
+      self.index = index
+      self.buildSetup = buildSetup
+    }
+  }
 
-  /// Build setup
```
benlangmuir: This patch changed the server to contain multiple workspaces, but this particular change to factor the fields into a Configuration struct seems to assume the opposite: a single workspace with multiple folders.  Can you clarify the intent?

akyrtzi:
Considering the model as "single workspace with multiple folders" makes sense to me. Not sure about the build system but I think it is valuable to have the index be global for the workspace in such a model.

Even if you have multiple folders, you arranged them in a workspace presumably because they are inter-related and you want to be able to "jump" between them easily. The index being global would allow doing "jump-to-definition" across folders and search for symbols in the whole workspace.
A similar model would be an Xcode workspace with multiple projects, where the index is global for the same benefits.

krzyzanowskim:
> @benlangmuir: In the LSP terminology, I think it would be correct to say there is a single workspace with multiple folders.

see https://github.com/Microsoft/language-server-protocol/issues/298#issuecomment-452650264.  if I understand the protocol designers intent correctly the `workspaceFolders` replaces `rootUri`and creates a container for the workspaces

That's how it looks in VSCode:
screenshot 2019-02-12 at 07 26 08

> @akyrtzi Not sure about the build system but I think it is valuable to have the index be global for the workspace in such a model.

yeah. not sure how to use a single index though.

Sources/LanguageServerProtocol/WorkspaceEdit.swift
krzyzanowskim: it's not. Just a model though. I can separate it to the PR later

Sources/LanguageServerProtocol/FailureHandlingKind.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// The failure handling strategy of a client if applying the workspace edit fails
+public enum FailureHandlingKind: String, Codable, Hashable {
+    /// Applying the workspace change is simply aborted if one of the changes provided
+    /// fails. All operations executed before the failing operation stay executed.
+    case abort
+
+    /// All operations are executed transactional. That means they either all
```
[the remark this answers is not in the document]
krzyzanowskim: ;-) it's specification copy&paste

Sources/SourceKit/Workspace.swift
krzyzanowskim: Looks like post https://github.com/apple/sourcekit-lsp/pull/65 merge issue.
`Configuration` only encapsulate config settings, doesn't change the previous meaning, does it? The `BuildSetup` though is overriding the setup for all workspaces that is not intended here.

Sources/LanguageServerProtocol/WorkspaceEdit.swift
benlangmuir: Sounds good

benlangmuir:
> if I understand the protocol designers intent correctly the `workspaceFolders` replaces `rootUri`and creates a container for the workspaces

I agree about the functionality, but the terminology is that there is a single workspace with multiple (root) folders. For example, the spec says:

> Many tools support more than one root folder per workspace.

It's similar in Xcode where we have a single workspace with one or more projects in it.  If we could preserve that in the code I think it would be clearer how it relates back to the spec.  E.g. one Workspace with multiple WorkspaceFolder/Workspace.Folder/similar.  What do you think?

>> @akyrtzi Not sure about the build system but I think it is valuable to have the index be global for the workspace in such a model.
> yeah. not sure how to use a single index though.

This isn't something we're setup for right now, but it's good to have in mind for future index improvements. If the folders have separate builds I think we'd need to teach the index to handle multple datastores and then figure out where to put the global index files, etc.  I'm sure we'd run into lots of fun problems to solve.

I'm still not sure what to think about the build system side of this.  I like to use `swift package edit` and there are times when I would want it to share a swiftpm workspace and use the index from the parent project, and there are times when I would want to separate them since I'm actively building only the dependency.  Maybe there just isn't a right answer for this.

akyrtzi:
I think the build and index should be interlinked together, meaning a build should be accompanied by an index that is associated with that the build. Trying to using one index with multiple builds is going to lead to a very complex and hard to manage model.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -79,27 +85,27 @@ public final class SourceKitServer: LanguageServer {
   }
 
   func registerWorkspaceRequest<R>(
-    _ requestHandler: @escaping (SourceKitServer) -> (Request<R>, Workspace) -> Void)
+    _ requestHandler: @escaping (SourceKitServer) -> (Request<R>) -> Void)
```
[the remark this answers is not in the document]
krzyzanowskim: it's too generic.
However, it could if `registerWorkspaceRequest` constraint to `TextDocumentRequest` - I'll do that

Sources/SourceKit/Workspace.swift
krzyzanowskim: > The BuildSetup though is overriding the setup for all workspaces that is not intended here.

After revisiting the code, it looks ok to me (I did refactor it to make the initializer intent clearer: d11029c8a823573cb09f682255d4a7c15fb31b7e).
The `BuildSetup` doesn't matter for ".noRoot" Workspace. Each property is setup per Workspace instance.

krzyzanowskim:
> but the terminology is that there is a single workspace with multiple (root) folders. For example, the spec says:

in terms of terminology, I see it as:
1. Rename Workspace -> Project
1. New Workspace is a container for [Project]

krzyzanowskim:
*Don't merge* - I need to re-start it from the scratch applying Workspace/Project https://github.com/apple/sourcekit-lsp/pull/64#issuecomment-508549472 idea. Also current version break "folding" tests (shrug) for some reason.

I'm closing this PR so it won't stay open forever.

PR #65, opened 2019-01-10
Command line arguments

krzyzanowskim:
- Introduce Server configuration model that holds configuration parameters from CLI arguments
- Adds CLI options (I found useful):

```
OVERVIEW: Language Server Protocol implementation for Swift and C-based languages

USAGE: sourcekit-lsp [options]

OPTIONS:
  --build-path          Specify build/cache directory [default: ./.build]
  --configuration, -c   Build with configuration (debug|release) [default: debug]
  --log-level           Set the logging level (debug|info|warning|error) [default: info]
  -Xcc                  Pass flag through to all C compiler invocations
  -Xcxx                 Pass flag through to all C++ compiler invocations
  -Xlinker              Pass flag through to all linker invocations
  -Xswiftc              Pass flag through to all Swift compiler invocations
  --help                Display available options
```


krzyzanowskim:
ok. this one is ready to review. CLI arguments should be easier to use and discover than environments.

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -128,9 +135,9 @@ public final class SwiftPMWorkspace {
     // FIXME: make these configurable
```
krzyzanowskim: should triple be configurable here? if not I guess we can remove the FIXME

Sources/SKCore/Configuration.swift
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Utility
+import PackageModel
```
benlangmuir: We shouldn't be using `PackageModel` outside of the `SwiftPMWorkspace` module - it's the only place that depends on all of libSwiftPM.  Only things in the `Utility` and `Basic` modules of SwiftPM should be used elsewhere.  Specifically, we could duplicate `BuildConfiguration`, which is a very simple enum, and `BuildFlags` is okay because it's already in `Utility`.

Sources/SKCore/Configuration.swift
```diff
@@ -0,0 +1,25 @@
...
+
+public protocol Configuration {
+  var build: SKCore.BuildConfiguration { get }
+}
+
+public protocol BuildConfiguration {
```
benlangmuir: What's the reasoning for using a protocol instead of just using a struct?

Sources/SKCore/Configuration.swift
```diff
@@ -0,0 +1,25 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Utility
+import PackageModel
+
+public protocol Configuration {
```
benlangmuir: What value is `Configuration` providing?  Is the idea to leave room for non-build configuration items?  I suggest dropping this for now. We can easily add it back if we need it.

Sources/SKSupport/BuildConfiguration.swift
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Utility
+import PackageModel
```
benlangmuir: Similarly, we shouldn't be importing `PackageModel` from `SKSupport`.

Sources/SKSupport/Logging.swift
```diff
@@ -12,18 +12,52 @@
...
 import Basic
 import Foundation
+import Utility
 
 #if canImport(os)
 import os // os_log
 #endif
 
-public enum LogLevel: Int, Equatable {
+public enum LogLevel: Int, Equatable, CustomStringConvertible, ArgumentKind {
```
benlangmuir: Style: Please separate conformances to their own extensions. The exception is synthesized conformances like `Equatable` or `Hashable` and conformances that are a core part of what the type is.

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
benlangmuir: Removing the FIXME sounds fine!  We may someday want to customize the triple, but not for now.

Sources/SKCore/Configuration.swift
krzyzanowskim: The initial idea was to have generalized Configuration<BuildConfiguration> and use it here and in tests with specific types, but that's too much hassle after all.

krzyzanowskim:
@benlangmuir thanks! addressed feedback. Build configuration is now held by `BuildSetup`, and introduced our `BuildConfiguration` enum

Sources/SKCore/BuildSetup.swift
```diff
@@ -0,0 +1,39 @@
...
+
+import Foundation
+import Utility
+import SKSupport
+
+/// Build configuration
+public struct BuildSetup {
+
+  /// Default configuration
+  public static let `default` = BuildSetup(configuration: .debug,
```
benlangmuir: I'm curious why you chose a static property for this instead of just making this the default initializer?

krzyzanowskim: verbosity and explicitly I guess.

Sources/sourcekit-lsp/main.swift
```diff
@@ -28,7 +59,7 @@ Logger.shared.addLogHandler { message, _ in
   clientConnection.send(LogMessage(type: .log, message: message))
 }
 
-let server = SourceKitServer(client: clientConnection, onExit: {
+let server = SourceKitServer(client: clientConnection, buildSetup: try parseConfigurationArguments(), onExit: {
```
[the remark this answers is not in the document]
krzyzanowskim: speaking of stderr. `Logger` doesn't output to `stderr`, does it? It looks like it use `os_log` or `NSLog`. I'd say .error should output to `stderr`.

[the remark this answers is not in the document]
benlangmuir: NSLog writes to stderr, no?

Sources/sourcekit-lsp/main.swift
```diff
@@ -14,11 +14,42 @@ import SourceKit
...
 import sourcekitd // Not needed here, but fixes debugging...
 
-Logger.shared.setLogLevel(environmentVariable: "SOURCEKIT_LOGGING")
+func parseConfigurationArguments() throws -> BuildSetup {
+  let arguments = Array(ProcessInfo.processInfo.arguments.dropFirst())
+  let parser = ArgumentParser(usage: "[options]", overview: "Language Server Protocol implementation for Swift and C-based languages")
+  let loggingOption = parser.add(option: "--log-level", kind: LogLevel.self, usage: "Set the logging level (debug|info|warning|error) [default: \(LogLevel.default)]")
+  let buildConfigurationOption = parser.add(option: "--configuration", shortName: "-c", kind: BuildConfiguration.self, usage: "Build with configuration (debug|release) [default: debug]")
+  let buildPathOption = parser.add(option: "--build-path", kind: String.self, usage: "Specify build/cache directory [default: ./.build]")
+  let buildFlagsCc = parser.add(option: "-Xcc", kind: [String].self, usage: "Pass flag through to all C compiler invocations")
```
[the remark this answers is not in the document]
krzyzanowskim: `-Xcc="-fsyntax-only"` works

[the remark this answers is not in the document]
krzyzanowskim: with https://github.com/apple/sourcekit-lsp/pull/65/commits/fd4d495abaec9e3fdd64a6c0f4879206f8d65552 it also works as `-Xcc -fsyntax-only`

Sources/sourcekit-lsp/main.swift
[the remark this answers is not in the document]
krzyzanowskim: `NSLog` does, but `os_log` doesn't seem so:

```
$ ./sourcekit-lsp ----Xcc -fsyntax-only
$
```

the error is logged though (accessible via Console.app)
screenshot 2019-01-30 at 19 52 57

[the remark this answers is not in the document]
benlangmuir: Right, that's expected for os_log.  Sorry, to be clear: logging in general (even at the .error level) should not necessarily use the same output policy as failures during command-line argument parsing.  I think the right thing to do for now is to output explicitly to stderr when command line argument parsing fails.  There's a bigger question about what the ideal logging is for other kinds of issues (there is no one right answer to this), but failure to parse command line arguments can just write to stderr and exit.

krzyzanowskim:
Thanks for the review! I addressed remaining issues. 
- Use `Logger` to log error, (see: https://github.com/apple/sourcekit-lsp/pull/65#discussion_r252383295)
- I'd leave `BuildSetup.default`.

benlangmuir:
FYI I made a few tweaks:

PR #66, opened 2019-01-11
Start using Boyer-Moore-Horspool algorithm.

abdullahselek:
- Changes on `firstIndex` to use Boyer-Moore-Horspool algorithm [swift-algorithm-club](https://github.com/raywenderlich/swift-algorithm-club/tree/master/Boyer-Moore-Horspool) used as a reference.

### DIFF
--- a/Sources/SKSupport/Collection.swift
+++ b/Sources/SKSupport/Collection.swift
@@ -10,7 +10,23 @@
 //
 //===----------------------------------------------------------------------===//
 
-extension RandomAccessCollection where Element: Equatable {
+extension RandomAccessCollection where Element: Equatable & Hashable {
+
+  /// Searches back through both patterns until we find an element that
+  /// doesn’t match or until we’ve reached the beginning of the pattern.
+  @inlinable
+  public func backwards<Pattern>(on pattern: Pattern, i: Index) -> Index? where Pattern: RandomAccessCollection, Pattern.Element == Element {
+    var q = pattern.index(before: pattern.endIndex)
+    var j = i
+    while q > pattern.startIndex {
+      j = index(before: j)
+      q = pattern.index(before: q)
+      if self[j] != pattern[q] {
+        return nil
+      }
+    }
+    return j
+  }
 
   /// Returns the first index where the specified subsequence appears or nil.
   @inlinable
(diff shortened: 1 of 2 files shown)


benlangmuir:
Thanks for taking a look!  Before accepting this it will need performance tests demonstrating the impact, and the data should be chosen based on where we're using this function.  I already made one attempt to replace this back when I originally wrote this code and my implementation turned out to be slower in practice.  It's been a while, but I think part of the issue is that the callers of this code are searching for short strings.

benlangmuir:
I checked: the only place this is currently used (aside from test code) is searching for the two byte sequence "\r\n" in message headers.  Before writing your own perf tests you could also sanity check with the existing `ConnectionPerfTests`.  They don't specifically stress the message parsing, but if there's a big change it might show up there.

abdullahselek:
Below you can find the results for `ConnectionPerfTests` with two different implementations.
**Old**
- `testEcho1` 0,00319s
- `testEcho100Latency` 0,0718s
- `testEcho100Throughput` 0,0253s

**New**
- `testEcho1` 0,00153s
- `testEcho100Latency` 0,066s
- `testEcho100Throughput` 0,0265s

benlangmuir:
Nice!  I'd still like to see a perf test that's specific to this function before we merge.

abdullahselek:
I added performance tests for collection, have a look when you have time.

Tests/SKSupportTests/CollectionPerfTests.swift
```diff
@@ -0,0 +1,65 @@
...
+    entire collection. Conforming to the RandomAccessCollection Protocol: the RandomAccessCollection protocol adds
+    further constraints on the associated Indices and SubSequence types, but otherwise imposes no additional
+    requirements over the BidirectionalCollection protocol. However, in order to meet the complexity guarantees of
+    a random-access collection, either the index for your custom type must conform to the Strideable protocol or
+    you must implement the index(_:offsetBy:) and distance(from:to:) methods with O(1 efficiency.
+    """.utf8)
+
+  func testFirstIndexWithShortestPattern() {
+    let keyword: [UInt8] = [UInt8](".".utf8)
+    self.measureMetrics([.wallClockTime], automaticallyStartMeasuring: true) {
```
benlangmuir: Can be simplified to `self.measure { ... }` (same for all the others.

Sources/SKSupport/Collection.swift
```diff
@@ -10,7 +10,23 @@
 //
 //===----------------------------------------------------------------------===//
 
-extension RandomAccessCollection where Element: Equatable {
+extension RandomAccessCollection where Element: Equatable & Hashable {
+
+  /// Searches back through both patterns until we find an element that
```
benlangmuir: s/both patterns/self and the given pattern/ ?

Sources/SKSupport/Collection.swift
```diff
@@ -10,7 +10,23 @@
...
+  /// doesn’t match or until we’ve reached the beginning of the pattern.
+  @inlinable
+  public func backwards<Pattern>(on pattern: Pattern, i: Index) -> Index? where Pattern: RandomAccessCollection, Pattern.Element == Element {
```
benlangmuir: How does this compare to something like `self[i...].starts(with: pattern)` ?  This whole helper function is pretty weird as an API and there's no real point in returning the index since it's always a known offset from `i`.  If we need to keep a helper like this it should be private at least.

benlangmuir: Err, I messed up that suggestion, but something along the lines of `self[index(i, offsetBy: -pattern.count)..<i] == pattern`.  Basically, we don't care whether this walks backwards of forwards (right?) and the return value is equivalent to a Bool, since we know the result is always either `nil` or `index(i, offsetBy: -pattern.count)`, although I might have an off-by-one error.

Tests/SKSupportTests/CollectionPerfTests.swift
abdullahselek: Fixed.

Sources/SKSupport/Collection.swift
abdullahselek: Could you give me more details?

Sources/SKSupport/Collection.swift
abdullahselek: I think this algorithm aims to compare the pattern to the text from right to left instead of left to right.

Sources/SKSupport/Collection.swift
benlangmuir: There is only one pattern.  It searches in `pattern` and `self`, not in two patterns.

Sources/SKSupport/Collection.swift
benlangmuir: ^^ Just making sure you saw these comments, @abdullahselek.

benlangmuir: @abdullahselek but why does it matter?  It always returns either `nil` or `index(i, offsetBy: -pattern.count)`.   It feels like this code is overly complicated for what it's doing.  If searching backwards has an advantage, it would be nice to better understand what that is, and even still maybe we don't need to return `Index`.

shahmishal:

PR #69, opened 2019-01-16
Try to build SPM workspace at open, if index store is not available.

krzyzanowskim:
Try to build SPM workspace if an index location is not available. This most likely will create an index (or not, but that's ok). Useful for newly opened workspaces - otherwise this command has to be run manually.

Spawn swift-build because SwiftPM doesn't expose `Command` module where build logic is implemented.

PS. also I tried to generate llbuild manifest file and use llbuild directly. In theory this would work, however, there is something missing/wrong and llbuild wasn't able to properly build the project out of that.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -135,6 +135,14 @@ public final class SwiftPMWorkspace {
     self.packageGraph = PackageGraph(rootPackages: [])
 
     try reloadPackage()
+
+    if !fileSystem.exists(self.buildParameters.indexStore) {
+        let swiftBuildPath = swiftPMToolchain.swiftCompiler.parentDirectory.appending(component: "swift-build")
+        try Basic.Process.popen(arguments: [swiftBuildPath.asString,
+                                            "--package-path", packageRoot.asString,
+                                            "--configuration", buildParameters.configuration.rawValue,
+                                            "--build-path", buildPath.asString])
+    }
   }
 
   /// Creates a build system using the Swift Package Manager, if this workspace is a package.


akyrtzi:
I have concerns about starting a, potentially *very expensive*, build in the background, like:

* How will the user be informed what is happening, and why is something suddenly taking over the machine's CPU without them initiating anything
* What is the management of this, can this be cancelled and how ? What if the user initiates a build on their own, are there now 2 separate build invocations trying to build the same project ?

My recommendation would be to avoid any background build operations until we have a clear model of how the user can control this and how we can avoid competing with a real build. This may mean that the sourcekit-lsp service never initiates such a build on its own, but only has the hooks for the editor/IDE to control this.

krzyzanowskim:
@akyrtzi these are valid points and I agree in general, BUT.

> How will the user be informed what is happening, and why is something suddenly taking over the machine's CPU without them initiating anything

> What is the management of this, can this be cancelled and how ? What if the user initiates a build on their own, are there now 2 separate build invocations trying to build the same project ?

This is already happening. Currently, we already start _very expensive_ operation in the background. e.g. for an empty Kitura project, it takes ~40s in my setup (mbp i9) to get any results 

```
sourcekit-lsp[80992:5391524] Completed resolution in 37.25s
```

Because what happens in `SwiftPMWorkspace` triggers dependencies resolution that involves a lot of work. It's like it's already half baked, so maybe it should be fully baked and usable.

> avoid competing with a real build

One think is that sourcekit-lsp shouldn't maybe share default `.build` directory. App build and lsp build is not necessarily for the same purpose.

akyrtzi:
> This is already happening. Currently, we already start _very expensive_ operation in the background. e.g. for an empty Kitura project, it takes ~40s in my setup (mbp i9) to get any results
> 
> ```
> sourcekit-lsp[80992:5391524] Completed resolution in 37.25s
> ```
> Because what happens in `SwiftPMWorkspace` triggers dependencies resolution that involves a lot of work. It's like it's already half baked, so maybe it should be fully baked and usable.

Not sure what would be best to do for the dependency resolution but I don't think it's a good idea to add on top another potentially way more expensive background build operation.

> One think is that sourcekit-lsp shouldn't maybe share default `.build` directory. App build and lsp build is not necessarily for the same purpose.

I don't think this is a good model to have; what you end up with is that you have say a large project that takes 10 mins to build, and after you build you have to also wait for a background build operation that takes even more because it needs to be in lower priority and uses less threads. And that is the case for every incremental builds as well.

The way forward IMO is to have:
* A way to efficiently only generate swift modules, ideally shared with the build.
* Have a background `indexer` (not just do a full build) but in a way where it can be cancelled as soon as the user initiates a build.

In absence of the above I prefer the existing situation over adding a full background build. The most common case is that the user will initiate a build on their own anyway and now you have two builds happening concurrently.

krzyzanowskim:
> there was no way to opt-out of that resolution

isn't the dependency resolution required to build an index anyway? It doesn't have to be done by the server though.

Since build can't (shouldn't) be triggered by the server itself, the sourcekit-lsp can't resolve any symbol until a build is triggered from the outside (IDE/manually). For that, the server needs only index location (am I right?) to work. That means maybe it doesn't need SPM that much (does it?) - whatever tool can create the index, is as good as any other. In particular this can be SPM, but does sourcekit-lsp has to know about it?

@akyrtzi Having and indexer distinct from an actual build is better approach, sure.

> This may mean that the sourcekit-lsp service never initiates such a build on its own, but only has the hooks for the editor/IDE to control this.

Now I wonder how that fit in LSP practices.

PS. To be clear, I'm not forcing this PR change. The "build" here is just another patchwork to replace proper indexer, agree on that. I just want to learn about the project principles in this area and maybe move things forward.

benlangmuir:
> isn't the dependency resolution required to build an index anyway? It doesn't have to be done by the server though.

Correct.

> Since build can't (shouldn't) be triggered by the server itself, the sourcekit-lsp can't resolve any symbol until a build is triggered from the outside (IDE/manually). For that, the server needs only index location (am I right?) to work. That means maybe it doesn't need SPM that much (does it?) - whatever tool can create the index, is as good as any other. In particular this can be SPM, but does sourcekit-lsp has to know about it?

We use `libSwiftPM` to get compiler arguments, not just to provide the location of index data. Many of the editor features such as code-completion are driven by live ASTs in sourcekitd that use these compiler arguments, not from the index.  Getting correct compiler arguments that (when the build has happened) line up with the generated .swiftmodules and other intermediate files is the core thing we need `libSwiftPM` for.  The index data itself is more loosely coupled,  but isn't enough on its own to provide most of our features. We have ideas about making it possible to use index data for more things, but I don't foresee this completely replacing live ASTs.

PR #72, opened 2019-01-19
Add CodeActions/Command structure

rockbruno:
This PR adds the skeleton of the Code Actions request, used to provide actions to the user based on a selection range. Actions are used for refactoring and to provide quick fixes to the user when errors are diagnosed. Part one to make https://github.com/apple/sourcekit-lsp/pull/70 easier to integrate.

Because CodeActions have several kinds, the idea was to create the concept of `CodeActionProviders` and tie them to their specific kinds, allowing the main codeActions function to determine which providers to trigger for a given request. The results are added and returned to the client.

## Command

In the LSP, commands are just an identifier and an `[Any]?` argument list. Because they need to be sent to the client and then parsed back again, that makes them a bit difficult to recognize in the server. To remedy this, I wrapped the `Command` Structure around a `CommandDataType` protocol to make the contents predictable, but this required playing around with `Codable`. It works, but it's not very pretty so I'm ready to listen to better options.

PS: I've left `WorkspaceEdit` out for now - I think the PR that implements ApplyEdit will be better suited for it.


Sources/LanguageServerProtocol/Command.swift
```diff
@@ -0,0 +1,133 @@
...
+
+  /// The internal identifier for this command.
+  public var command: String {
+    return type(of: data).identifier
+  }
+
+  /// The data related to this command.
+  /// This is called `arguments` and is [Any]? in the LSP, but internally we treat it
+  /// differently to make it easier to create and (de)serialize commands.
+  public var data: CommandDataType
```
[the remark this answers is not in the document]
danielmartin: Also, could we keep the name `arguments` for this field, just like in the LSP spec?

Sources/LanguageServerProtocol/Command.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SKSupport
+
+/// A `CommandDataType` represents the underlying data of a custom `Command`.
+public protocol CommandDataType: Codable {
```
danielmartin: I think it's more accurate to describe this as "A `CommandDataType` represents the arguments required to execute a `Command`."

Also, If you agree with my suggestion to keep "arguments", I'll rename it as `CommandArgsType`.

danielmartin: I don't know whether requiring all these three properties may be very restrictive to the model or not. I imagine a hypothetic future "Organize imports" command action that would only require a reference to the `textDocument` to do its job. Also, I'd copy here the property docs you have in `SemanticRefactorCommandDataType`.

Sources/LanguageServerProtocol/CodeAction.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// An array of diagnostics.
+  public var diagnostics: [Diagnostic]
+
+  /// Requested kind of actions to return.
+  /// If provided, actions of these kinds are filtered out by the client before being shown,
+  /// so servers can omit computing them.
+  public var only: [CodeActionKind]?
+
+  public init(diagnostics: [Diagnostic] = [], only: [CodeActionKind]?) {
+    self.diagnostics = diagnostics
```
danielmartin: Nit: Should default arguments come last in an argument list? @benlangmuir

Sources/LanguageServerProtocol/Command.swift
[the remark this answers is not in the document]
rockbruno: Thanks for the comments Daniel! I was concerned that directly naming it `arguments` would be misleading since it works differently from the specs, but I agree that it sounds better.

I made it non-optional because the command identifier comes from the protocol, but I don't remember why I did that. I think making it optional will be better indeed 👍

Sources/LanguageServerProtocol/CodeAction.swift
benlangmuir: All else being equal, yes, but here it's probably best to match the order of the fields.  Maybe `only` should have default of `nil`?

Sources/LanguageServerProtocol/Command.swift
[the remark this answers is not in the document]
rockbruno: I've been applying the suggestions and I now remember why I did that - it's because SourceKit-LSP requires all requests to have a `textDocument` and ExecuteCommand isn't a `TextDocumentRequest`. I forced the argument to always contain it in order to circumvent it because I assumed that making the requests not need it would be a massive change, if possible at all since it uses that to determine which language server to trigger. I'm not sure what's the best approach here now. What do you think?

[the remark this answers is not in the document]
benlangmuir: Can you expand on this? We should not require a textDocument in general.  If the Command is something that will be executed by sourcekitd/clangd then sure, we need to know how to dispatch it, but that doesn't need to be a document name.  We can embed whatever information (e.g. toolchain service) we need into `arguments` and then look it up if we are asked to execute the command.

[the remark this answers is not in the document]
rockbruno: Ah, I was under the impression that we couldn't support argument-less commands in the code as we somehow have to explicitly add the "this is Swift" info in the command's arguments (and a document was the easier way since the requests are built on it), but after some thinking I reworked the structure to make Command's `Codable` inject/retrieve the `textDocument` in/from the first position of the array automatically. The command's JSON itself will never have empty arguments, but the type inside the code will now be able to. 

This is going to do the trick while we don't have a command without a document related to it, but I see now that adding a type of request that doesn't need a document would be straight-forward if we end up getting to that point.

rockbruno:
I changed Command to be an enum instead of protocol/struct combos to make Codable/Hashable conformance easier. Ready for review again!

unnamedd:
@swift-ci Please test

rockbruno:
@benlangmuir Is these a plan for these old PRs?

benlangmuir:
Sorry, I'd like to get back to these soon, but I've been busy and I want to do a detailed review.  I definitely haven't forgotten them!

danielmartin:
I'll rebase/rework https://github.com/apple/sourcekit-lsp/pull/19 on top of this once the commands structure is approved.

benlangmuir:
@rockbruno hey, sorry again for the long delay.  The way this patch represents the Command structure and requests could probably work if we controlled all the pieces, but because we are also supporting clangd I think we need more flexibility.  Specifically, we don't know what clangd will want to put in these Command structures, but we will need to be able to pass them through.

Here's what I suggest; let me know if you think this makes sense:
* We add an AnyJSON type to hold the arbitrary `arguments` data that can round-trip
* The LanguageServerProtocol library doesn't know anything about the specific content
* The swift implementation learns to convert whatever model objects it wants to put in the Command arguments to and from the AnyJSON type
* You don't need to implement clangd support, but we should at least make sure we can round trip some arbitrary data through this type

In a similar vein, the result type of the CodeAction request needs to be something like a CommandOrCodeAction - for backwards compatibility, we're supposed to check the client capabilities before using the newer CodeAction structure.

rockbruno:
No worries! Sounds reasonable. The tough part is how we're getting AnyJSON to conform to Codable/Hashable as the arguments can be anything, but I think I have something in mind.

rockbruno:
Apologies for the delay, I was short on time these last few weeks. 

When I did this for the first time I didn't noticed it had support for different return types - so I took a proper look now and reworked the structure. I wrapped the capabilities and the response into custom types that take the client's capabilities to determine what to encode/decode.

I also got rid of the implemented Commands to make things simpler, so it looks just like the LSP counterpart now. For clang I type erased the arguments, but it's still unclear to me how it will interact with all of this so if that's not enough to support it let me know!

Sources/LanguageServerProtocol/Command.swift
```diff
@@ -0,0 +1,176 @@
...
+  public var arguments: [AnyCommandArgument]?
+
+  public init(title: String, command: String, arguments: [AnyCommandArgument]?) {
+    self.title = title
+    self.command = command
+    self.arguments = arguments
+  }
+}
+
+public protocol CommandArgumentType: Codable, Hashable {}
```
benlangmuir: Could we use an `enum` for this instead?  There are a few places in this patch where we make assumptions about the set of types that can conform to this protocol, so this may be better as an enum where we will get an error if we miss a case.  Splitting Int/Double is a bit weird, but may be unavoidable.

```swift
enum CommandArgumentType {
  case null // JSON is great, isn't it?
  case int(Int)
  case double(Double)
  case bool(Bool)
  case string(String)
  case array([CommandArgumentType])
  case dictionary([String: CommandArgumentType])
}
```

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -735,6 +742,40 @@ extension SwiftLanguageServer {
...
+    guard providers.isEmpty == false else {
+      completion([])
+      return
+    }
+    var providersLeftToProcess = Set<Int>(0..<providers.count)
+    var codeActions = [CodeAction]()
+    for i in 0..<providers.count {
+      providers[i](req.params) { actions in
+        codeActions += actions
+        providersLeftToProcess.remove(i)
```
benlangmuir: Why do we need `providersLeftToProcess`?  When we're at iteration `i`, either we have `providersLeftToProcess == i...providers.count`, or we have somehow missed one and we will exit without calling the completion handler, which is a logic error.

rockbruno: The ones that call sourcekitd can be async, so we keep track of who's left because we don't know the order they're going to finish -- is that what you meant or I misunderstood the comment? If the former, I think I'll add an operation queue to make it less hacky.

benlangmuir: I see. What happens if there's an error?

rockbruno: I'm having them log the error and return an empty list to not affect the other code actions:
https://github.com/apple/sourcekit-lsp/blob/f0b632c6081740d6ead21a398a56b26103f852b4/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift#L698

benlangmuir: Okay, the critical thing is that we always call the request's reply method eventually.

rockbruno: Sure, we can make a test to prove it when a provider is added. I had to make the completion optional for now to avoid a compilation issue, but once we add the refactoring provider I can iron this out.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -735,6 +743,42 @@ extension SwiftLanguageServer {
...
+  func retrieveCodeActions(_ req: Request<CodeActionRequest>, providers: [CodeActionProvider], completion: @escaping CodeActionProviderCompletion) {
+    guard providers.isEmpty == false else {
+      completion([])
+      return
+    }
+    var providersLeftToProcess = Set<Int>(0..<providers.count)
+    var codeActions = [CodeAction]()
+    for i in 0..<providers.count {
+      providers[i](req.params) { actions in
+        DispatchQueue.global(qos: .userInitiated).sync(flags: .barrier) {
```
benlangmuir: I would just do  `self.queue.sync { ... }`.

Tests/SourceKitTests/CodeActionTests.swift
```diff
@@ -0,0 +1,108 @@
...
+import LanguageServerProtocol
+import SKSupport
+import SKTestSupport
+import XCTest
+
+@testable import SourceKit
+
+final class CodeActionTests: XCTestCase {
+  func testCodeActionResponseLegacySupport() {
+    let command = Command(title: "Title", command: "Command", arguments: [1,"f",2.2, nil])
```
benlangmuir: We should have a test that uses nested dictionary/array a bit as well.

rockbruno:
Thanks for the tips! I made the server capabilities into an enum as well as it represents the same idea.

benlangmuir:
@swift-ci please test

rockbruno:
Ah thanks for your patience when looking at my mistakes! I rebased https://github.com/apple/sourcekit-lsp/pull/70, but I think I'll close it and open a new PR as most of the points I mentioned there are no longer relevant.

PR #76, opened 2019-01-24
[DO NOT MERGE] Stress testing SwiftFormat

dabelknap:
The sole purpose of this PR is to test SwiftFormat and allow interested parties to comment on the results. The formatter configuration was intentionally chosen to produce a large diff and more clearly show the effects of the formatter:
- Indents are 6 spaces
- Maximum line length is 80 characters
- User-inputed line breaks are ignored; the formatter controls all line breaks. The only exceptions are blank lines between source code.

### DIFF
--- a/Sources/LanguageServerProtocol/CompletionItem.swift
+++ b/Sources/LanguageServerProtocol/CompletionItem.swift
@@ -13,70 +13,66 @@
 /// A single completion result.
 public struct CompletionItem: Codable, Hashable {
 
-  /// The display name of the completion.
-  public var label: String
+      /// The display name of the completion.
+      public var label: String
 
-  /// An extended human-readable name (longer than `label`, but simpler than `documentation`).
-  public var detail: String?
+      /// An extended human-readable name (longer than `label`, but simpler than `documentation`).
+      public var detail: String?
 
-  /// The name to use for sorting the result. If `nil`, use `label.
-  public var sortText: String?
+      /// The name to use for sorting the result. If `nil`, use `label.
+      public var sortText: String?
 
-  /// The name to use for filtering the result. If `nil`, use `label.
-  public var filterText: String?
+      /// The name to use for filtering the result. If `nil`, use `label.
+      public var filterText: String?
 
-  /// The (primary) edit to apply to the document if this completion is accepted.
-  ///
-  /// This takes precedence over `insertText`.
-  ///
-  /// - Note: The range of the edit must contain the completion position.
-  public var textEdit: TextEdit?
+      /// The (primary) edit to apply to the document if this completion is accepted.
+      ///
+      /// This takes precedence over `insertText`.
+      ///
+      /// - Note: The range of the edit must contain the completion position.
+      public var textEdit: TextEdit?
 
-  /// **Deprecated**: use `textEdit`
-  ///
-  /// The string to insert into the document. If `nil`, use `label.
-  public var insertText: String?
+      /// **Deprecated**: use `textEdit`
+      ///
+      /// The string to insert into the document. If `nil`, use `label.
+      public var insertText: String?
 
-  /// The format of the `textEdit.nextText` or `insertText` value.
-  public var insertTextFormat: InsertTextFormat?
+      /// The format of the `textEdit.nextText` or `insertText` value.
+      public var insertTextFormat: InsertTextFormat?
 
-  /// The kind of completion item - e.g. method, property.
-  public var kind: CompletionItemKind
+      /// The kind of completion item - e.g. method, property.
+      public var kind: CompletionItemKind
 
-  /// Whether the completion is for a deprecated symbol.
-  public var deprecated: Bool?
+      /// Whether the completion is for a deprecated symbol.
+      public var deprecated: Bool?
 
-  // TODO: remaining members
+      // TODO: remaining members
 
-  public init(
-    label: String,
-    detail: String? = nil,
-    sortText: String? = nil,
-    filterText: String? = nil,
-    textEdit: TextEdit? = nil,
-    insertText: String? = nil,
-    insertTextFormat: InsertTextFormat? = nil,
-    kind: CompletionItemKind,
-    deprecated: Bool? = nil)
-  {
-    self.label = label
-    self.detail = detail
-    self.sortText = sortText
-    self.filterText = filterText
-    self.textEdit = textEdit
-    self.insertText = insertText
-    self.insertTextFormat = insertTextFormat
-    self.kind = kind
-    self.deprecated = deprecated
-  }
+      public init(
+            label: String, detail: String? = nil, sortText: String? = nil,
+            filterText: String? = nil, textEdit: TextEdit? = nil,
+            insertText: String? = nil,
+            insertTextFormat: InsertTextFormat? = nil, kind: CompletionItemKind,
+            deprecated: Bool? = nil
+      ) {
+            self.label = label
+            self.detail = detail
+            self.sortText = sortText
+            self.filterText = filterText
+            self.textEdit = textEdit
+            self.insertText = insertText
+            self.insertTextFormat = insertTextFormat
+            self.kind = kind
+            self.deprecated = deprecated
+      }
 }
 
 /// The format of the returned insertion text - either literal plain text or a snippet.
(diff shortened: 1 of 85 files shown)


Sources/LanguageServerProtocol/CompletionItem.swift
```diff
@@ -13,70 +13,66 @@
...
-    self.insertTextFormat = insertTextFormat
-    self.kind = kind
-    self.deprecated = deprecated
-  }
+      public init(
+            label: String, detail: String? = nil, sortText: String? = nil,
+            filterText: String? = nil, textEdit: TextEdit? = nil,
+            insertText: String? = nil,
+            insertTextFormat: InsertTextFormat? = nil, kind: CompletionItemKind,
+            deprecated: Bool? = nil
```
benlangmuir: What happened to always formatting parameters vertically XOR horizontally?  This change is a big regression IMO.  I can see why folks might want it in some specific cases, but it makes declarations harder to read.

Sources/LanguageServerProtocol/Diagnostic.swift
```diff
@@ -12,98 +12,96 @@
...
+            let value = try decoder.singleValueContainer()
+            if let intValue = try? value.decode(Int.self) {
+                  self = .number(intValue)
+            }
+            else if let strValue = try? value.decode(String.self) {
+                  self = .string(strValue)
+            }
+            else {
+                  throw MessageDecodingError.invalidRequest(
+                        "could not decode diagnostic code")
```
benlangmuir: Is it intentional to align the string with Message, or is this a coincidence of using 6 spaces?

Sources/LanguageServerProtocol/CompletionItemKind.swift
```diff
@@ -15,40 +15,63 @@
...
-  public static let value: CompletionItemKind = CompletionItemKind(rawValue: 12)
-  public static let `enum`: CompletionItemKind = CompletionItemKind(rawValue: 13)
-  public static let keyword: CompletionItemKind = CompletionItemKind(rawValue: 14)
-  public static let snippet: CompletionItemKind = CompletionItemKind(rawValue: 15)
-  public static let color: CompletionItemKind = CompletionItemKind(rawValue: 16)
-  public static let file: CompletionItemKind = CompletionItemKind(rawValue: 17)
-  public static let reference: CompletionItemKind = CompletionItemKind(rawValue: 18)
+      // LSP 1 kinds, guaranteed to be supported by all clients.
+      public static let text: CompletionItemKind = CompletionItemKind(
+            rawValue: 1)
```
benlangmuir: This particular wrapping looks pretty weird IMO, but it's not a dealbreaker.

Sources/LanguageServerProtocol/FoldingRange.swift
```diff
@@ -21,73 +21,69 @@
...
-    self.startLine = startLine
-    self.startUTF16Index = startUTF16Index
-    self.endLine = endLine
-    self.endUTF16Index = endUTF16Index
-    self.kind = kind
-  }
+      public init(
+            startLine: Int, startUTF16Index: Int? = nil, endLine: Int,
+            endUTF16Index: Int? = nil, kind: FoldingRangeKind? = nil
+      ) {
```
benlangmuir: `) {` on its own line makes me sad 😢

Sources/LanguageServerProtocol/LanguageServer.swift
```diff
@@ -10,185 +10,207 @@
...
+      ///
+      /// Must be called on `queue`.
+      public func _register<Server, R>(
+            _ requestHandler: @escaping (Server) -> (Request<R>) -> Void
+      ) {
+            // We can use `unowned` here because the handler is run synchronously on `queue`.
+            precondition(self is Server)
+            requestHandlers[ObjectIdentifier(R.self)] = {
+                  [unowned self] request in
+                  requestHandler(self as! Server)(request)
```
benlangmuir: Having the capture list and signature on the same indentation as the body seems like a mistake.

Sources/LanguageServerProtocol/CompletionItem.swift
dabrahams: That rule was discarded after I encouraged it.  It is much too restrictive and can cause unreadably long columnar declarations where you see, e.g.

```swift
     ...parameters...
     red: Float,
     green: Float,
     blue: Float,
     alpha: Float,
     ...parameters...
```

which should really be grouped together on a single line.

```swift
     ...parameters...
     red: Float, green: Float, blue: Float, alpha: Float,
     ...parameters...
```

This goes not just for extremely long parameter lists.  It's very common to find declarations where all the parameters fit on two lines, and there's a natural grouping that makes the declaration read well.
 
Because the new rules allow discretionary line breaks, we can expect people to add breaks where it's best for readability (including making parameter columns when that's best), but the tools aren't going to unwrap my carefully-organized declarations and turn them into long columns.

The reason the declaration you see here looks so bad is that for the purposes of this test, discretionary line breaks are being ignored, but that's not the intended configuration.

Sources/LanguageServerProtocol/CompletionItemKind.swift
dabrahams: The intention is to adjust the rules to encourage the formatter to break before the `=` if it would not increase the overall number of lines.

Sources/LanguageServerProtocol/FoldingRange.swift
dabrahams: That is a standard idiom in the standard library style that this formatter was designed to support.  I'm sorry you don't like it, but it's not just a random side-effect of other rules; it's an intended outcome.

When discretionary breaks are re-enabled, this would also be an allowed result as long as the closing paren fit at the end of the previous line:

```swift
      public init(
            startLine: Int, startUTF16Index: Int? = nil, endLine: Int,
            endUTF16Index: Int? = nil, kind: FoldingRangeKind? = nil)
{
```

Sources/LanguageServerProtocol/LanguageServer.swift
dabrahams: That's intended.  How would you indent it?

aciidb0mb3r:
This is great to see! Is SwiftFormat being built as part of the Swift toolchains?

Sources/LanguageServerProtocol/CompletionItem.swift
allevato: Right—we decided to completely ignore existing line breaks for this test so that we could get an intentionally more "opinionated" diff, to make sure that keeping those line breaks didn't somehow hide other issues.

The final version will preserve them (it's configurable right now, but that's more for development than intended use).

Sources/LanguageServerProtocol/CompletionItemKind.swift
allevato: That grouping is a straightforward change that Austin or I can make tomorrow, and then push a new commit with those changes so we can see them isolated against this original format pass.

Sources/LanguageServerProtocol/LanguageServer.swift
benlangmuir: It's hopefully a rare case, but I prefer keeping the signature separate, so add an additional level of indent:

```
requestHandlers[ObjectIdentifier(R.self)] = {
                                    [unowned self] request in
                  requestHandler(self as! Server)(request)
```

allevato: Closures are one of the two biggest pain points we've had to work with when trying to universally apply the idea of separating the body of a block from the prologue above it (`case` clauses are the other one). We've yet to find an answer that we _loved_ here.

Google's original style guide suggested that if the signature of a closure must be moved down to its own line, it should be surrounded by parentheses to help distinguish it (albeit slightly) from the rest of the body, but inserting or removing non-whitespace characters as part of the line-wrapping algorithm starts getting a little weird.

We could also just force a blank line between them, but I'm not crazy about that either.

dabrahams: My take is that 

- yes, it is rare
- the alternative proposed is quirky—no other rule increases indentation by more than 1 between consecutive lines.
- the programmer can always add a blank line after the "in" clause
- therefore, it's not worth complicating the rules to do something special here.

PR #78, opened 2019-02-01
Change configuration build path to be absolute

benlangmuir:
While specifying the build path relative to something might be a useful idea in principle, we would need a clearer story for how it is specified and resolved. When passing a path to a command-line argument, it should be resolved relative to the CWD, not to the package root by default. Incidentally, it doesn't really make sense to have a global default, since different build systems are likely to want different defaults. For now, simplify this configuration to an optional absolute path (passing a relative path on the command line will resolve relative to the CWD). Each build system can interpret `nil` as its preferred default.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -113,13 +113,7 @@ public final class SwiftPMWorkspace {
     swiftPMToolchain.extraSwiftCFlags = extraSwiftFlags
     swiftPMToolchain.extraCPPFlags = extraClangFlags
 
-
-    let buildPath: AbsolutePath
-    if let absoluteBuildPath = try? AbsolutePath(validating: buildSetup.path) {
-      buildPath = absoluteBuildPath
-    } else {
-      buildPath = packageRoot.appending(component: buildSetup.path)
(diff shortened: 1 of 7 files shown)


Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -113,13 +113,7 @@ public final class SwiftPMWorkspace {
...
     swiftPMToolchain.extraCPPFlags = extraClangFlags
 
-
-    let buildPath: AbsolutePath
-    if let absoluteBuildPath = try? AbsolutePath(validating: buildSetup.path) {
-      buildPath = absoluteBuildPath
-    } else {
-      buildPath = packageRoot.appending(component: buildSetup.path)
-    }
+    let buildPath: AbsolutePath = buildSetup.path ?? packageRoot.appending(component: ".build")
```
krzyzanowskim: SwiftPM can use its default by any chance?

benlangmuir: It's not currently exposed in a way we can use it.  There is `Workspace.create(forRootPackage:manifestLoader:)`, which will use the default path, but it doesn't let you customize it or use an in-memory filesystem (which we use in tests).

PR #80, opened 2019-02-02
[swiftpm] Update to latest master

benlangmuir:
* import Utility -> import SPMUtility
* AbsolutePath.asString -> path.description or "\(path)"

### DIFF
--- a/Sources/SKCore/FallbackBuildSystem.swift
+++ b/Sources/SKCore/FallbackBuildSystem.swift
@@ -12,7 +12,7 @@
 
 import LanguageServerProtocol
 import Basic
-import enum Utility.Platform
+import enum SPMUtility.Platform
 
 /// A simple BuildSystem suitable as a fallback when accurate settings are unknown.
 public final class FallbackBuildSystem: BuildSystem {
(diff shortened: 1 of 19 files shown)


Sources/SKCore/FallbackBuildSystem.swift
```diff
@@ -53,10 +53,10 @@ public final class FallbackBuildSystem: BuildSystem {
     if let sdkpath = sdkpath {
       args += [
         "-sdk",
-        sdkpath.asString,
+        sdkpath.description,
```
aciidb0mb3r: `description` is for producing human readable path, please use `pathString` where you want to guarantee proper string representation.

benlangmuir: Thanks, will do.  I was wondering about this when I did it, and I chose `description` because that's what the SwiftPM change did. I didn't realize that `pathString` didn't exist at the time of that change.  It all makes more sense now.

benlangmuir: https://github.com/apple/sourcekit-lsp/pull/81

PR #86, opened 2019-02-27
Changes for running sourcekit-lsp for Swift on PowerPC64LE

sarveshtamba:


benlangmuir:
It was silly of me to hardcode these triples in the first place.  I will change it to use swiftpm to compute the triple strings, which should handle this already.  Closing in favour of https://github.com/apple/sourcekit-lsp/pull/87

PR #87, opened 2019-03-01
[test] Make swiftpm test robust against host triple differences

benlangmuir:
Hardcoding the triples was silly; the only interesting platform
difference is about which parts have a version number on macOS and which
do not.

### DIFF
--- a/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
+++ b/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
@@ -16,6 +16,7 @@ import PackageModel
 import Basic
 import SPMUtility
 import SKTestSupport
+import Build
 import XCTest
 
 final class SwiftPMWorkspaceTests: XCTestCase {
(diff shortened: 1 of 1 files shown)


Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
```diff
@@ -111,11 +112,11 @@ final class SwiftPMWorkspaceTests: XCTestCase {
 
     check("-target", arguments: arguments) // Only one!
 #if os(macOS)
-    check("-target", "x86_64-apple-macosx10.10", arguments: arguments)
+    check("-target", Triple.hostTriple.tripleString(forPlatformVersion: "10.10"), arguments: arguments)
```
krzyzanowskim: I wonder if due to this issue: https://twitter.com/jckarter/status/1095757933579689985 we shouldn't stick to minimum 10.11 on macOS?

benlangmuir: If this ever breaks we can change it, but this test won't try to generate code, so I think it's probably fine.

PR #90, opened 2019-03-07
Workaround perf test running too many iterations

benlangmuir:


benlangmuir:
@swift-ci please test

Tests/SKSupportTests/SupportPerfTests.swift
```diff
@@ -24,8 +24,7 @@ final class SupportPerfTests: PerfTestCase {
       "\t", "\n"
     ] + (32...126).map { Character(UnicodeScalar($0)) }
 
-    // The debug performance is shockingly bad.
-    #if DEBUG
+    #if DEBUG || !ENABLE_PERF_TESTS
```
aciidb0mb3r: Can you please not run _any_ perf tests on the PR bots?

benlangmuir: It's already not running as a performance test (i.e. I hooked the `measure` family of methods so that they don't check perf and only run their block once).

PR #91, opened 2019-03-07
Remove Package.resolved and move to depending on HEAD

benlangmuir:
Always use the latest indexstore-db and swiftpm from the corresponding branch (currently master). This is important for swiftpm, because it does not *guarantee* stable behaviour, so we need to match the version. It also makes sourcekit-lsp behave more like the rest of the swift toolchain projects, who all live on HEAD for their dependencies.


benlangmuir:
Please test with following pull request:
https://github.com/apple/swift-package-manager/pull/2031

@swift-ci please test

PR #92, opened 2019-03-13
Add how to add code cmd in the terminal

mackoj:


Editors/vscode/README.md
```diff
@@ -22,6 +22,8 @@ $ cd Editors/vscode
 $ npm run createDevPackage
 ```
 
+Enable vscode to running in the terminal by enabling [code](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line) command.
+
```
benlangmuir: We should combine this with the next sentence below to clarify.  How about something like:

You can install the package from the command-line using the `code` command if available (see [Launching from the command line](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line)).

PR #93, opened 2019-03-14
Support WSL by adding additional #os(macOS) check in ClangLanguageServer

fwcd:
This branch fixes the build on Windows Subsystem for Linux that would previously throw the following error:

```swift
.../sourcekit-lsp/Sources/SourceKit/clangd/ClangLanguageServer.swift:166:9: error: value of type 'Process' has no member 'run'
    try process.run()
        ^~~~~~~ ~~~
```

### DIFF
--- a/Sources/SourceKit/clangd/ClangLanguageServer.swift
+++ b/Sources/SourceKit/clangd/ClangLanguageServer.swift
@@ -146,11 +146,15 @@ func makeJSONRPCClangServer(client: MessageHandler, clangd: AbsolutePath, buildS
 
   let process = Foundation.Process()
 
-  if #available(OSX 10.13, *) {
-    process.executableURL = clangd.asURL
-  } else {
+  #if os(macOS)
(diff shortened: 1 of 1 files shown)


benlangmuir:
CC @compnerd

This error doesn't make sense to me: swift-corelibs-foundation's Process.run() should be available unconditionally on Linux and Windows.  In fact, Process.launch() is implemented in terms of run:

```
    @available(*, deprecated, renamed: "run")
    open func launch() {
        do {
            try run()
```

compnerd:
If they are building on WSL, this doesn't even go down the Windows path, that is just a Windows emulation layer running on Windows (think of it as Wine inverted).  The Windows code path is irrelevant there.  I think that they were probably building against an older release of Foundation.

fwcd:
Indeed, that was the issue. I tried building `sourcekit-lsp` using Swift 4.2, `Process.run()` however first got added to `swift-corelibs-foundation` in Swift 5.

Thank you.

PR #96, opened 2019-04-04
On Linux, must include full path to `Block.h`

barrymoo:
`swift build -Xcxx -I/usr/lib/swift` fails on Linux w/

```
.../sourcekit-lsp/.build/checkouts/indexstore-db/lib/CIndexStoreDB/CIndexStoreDB.cpp:20:10: fatal error: 'Block.h' file not found
#include <Block.h>
         ^~~~~~~~~
1 error generated.
```

I needed to add `-I/usr/lib/swift/Block` to make it work. This change make the documentation here consistent with https://github.com/apple/indexstore-db.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ SourceKit-LSP is designed to build against the latest SwiftPM, so if you run int
 The C++ code in the index requires `libdispatch`, but unlike Swift code, it cannot find it automatically on Linux. You can work around this by adding a search path manually.
 
 ```sh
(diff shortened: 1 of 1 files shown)


barrymoo:
@benlangmuir I should mention that with Swift 5, I think this needs to be `-Xcxx -I/usr/lib/swift/Block`.

benlangmuir:
@barrymoo ah, I think that should have been needed all along, but I didn't notice when I reviewed this.  Could you update it?

PR #97, opened 2019-04-12
[test] Fix rare failure in LineTable editing test

benlangmuir:
This test will (rarely) delete enough of the string to trip over empty
line and column ranges. After this fix it should be fine to completely
empty the string.


benlangmuir:
@swift-ci please test

benlangmuir:
FYI @nate-chandler this was the random failure you saw.

PR #98, opened 2019-04-14
Implement document symbol request

Trzyipolkostkicukru:
I implemented document symbol request, which enables "document outline" in visual studio code. 

I have no experience with contributing features to open source, I tried to match the style of the rest of the project. Are there any things I have to change/add?


Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -471,6 +473,117 @@ extension SwiftLanguageServer {
...
+      guard let dict = result.success else {
+        req.reply(.failure(result.failure!))
+        return
+      }
+      guard let results: SKResponseArray = dict[self.keys.substructure] else {
+        return req.reply([])
+      }
+
+      /// Translate sourcekit symbol kinds into LSP kinds.
+      func symbolKind(sourcekitSymbolKind: sourcekitd_uid_t) -> SymbolKind? {
```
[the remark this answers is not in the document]
Trzyipolkostkicukru: I think that's a good idea :)

benlangmuir:
@swift-ci please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -758,4 +840,39 @@ extension sourcekitd_uid_t {
         return nil
     }
   }
+
+  func asSymbolKind(_ vals: sourcekitd_values) -> SymbolKind? {
+    switch self {
+      case vals.decl_class:
+        return .class
+      case vals.decl_function_method_instance,
```
benlangmuir: How about

```
case vals.decl_function_constructor:
  return .constructor
```

Trzyipolkostkicukru: I'm not sure how would I go about this, because sourcekit seems to return "source.lang.swift.decl.function.method.instance" as the type for both inits and normal methods

benlangmuir: Weird, okay we can skip this for now then.

benlangmuir:
@swift-ci please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
krzyzanowskim: I just found this discussion. since `init` and `deinit` is a keyword, maybe we should check method name and apply `.keyword` here, however `.constructor` is semantially more correct)

PR #99, opened 2019-04-18
Add Windows Support to SKSupport

gmittert:
I've ported enough of SwiftPM to Windows that I'm able to start building sourcekit-lsp and thus, start adding Windows support to Sourcekit-lsp.

### DIFF
--- a/Sources/SKSupport/dlopen.swift
+++ b/Sources/SKSupport/dlopen.swift
@@ -13,9 +13,14 @@
 import SPMLibc
 
 public final class DLHandle {
-  var rawValue: UnsafeMutableRawPointer? = nil
+  #if os(Windows)
+    typealias Handle = HMODULE
+  #else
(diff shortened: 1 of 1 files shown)


Sources/SKSupport/dlopen.swift
```diff
@@ -60,32 +73,53 @@ public struct DLOpenFlags: RawRepresentable, OptionSet {
   }
 }
 
+#if os(Windows)
+public enum DLError: Swift.Error {
+  case LoadLibrary(String)
+  case FreeLibrary(String)
+}
+#else
```
benlangmuir: Let's just unify these since they're the same other than the name. How about `DLError.open` and `DLError.close`?

gmittert: Seems reasonable. They're now open and close.

gmittert:
@swift-ci please test

PR #100, opened 2019-04-18
Fixup SourceKit module on Windows

gmittert:


Sources/SourceKit/sourcekitd/SwiftSourceKitFramework.swift
```diff
@@ -54,7 +58,11 @@ final class SwiftSourceKitFramework {
     // Workaround rdar://problem/43656704 by not constructing the value directly.
     // self.api = sourcekitd_functions_t(
     let ptr = UnsafeMutablePointer<sourcekitd_functions_t>.allocate(capacity: 1)
+    #if os(Windows)
+    memset(UnsafeMutableRawPointer(ptr), 0, MemoryLayout<sourcekitd_functions_t>.stride)
+    #else
     bzero(UnsafeMutableRawPointer(ptr), MemoryLayout<sourcekitd_functions_t>.stride)
```
akyrtzi: We should just use `memset` since it is cross-platform.

gmittert: Sure! Changed.

akyrtzi:
@swift-ci Please test

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -169,8 +169,13 @@ public final class SourceKitServer: LanguageServer {
         return nil
       }
 
+      #if os(Windows)
+        let pid: Int = unsafeBitCast(GetCurrentProcess(), to: Int.self)
+      #else
+        let pid: Int = Int(getpid())
+      #endif
```
pvieito: Why we are reimplementing things already in Foundation?

This could simply be:

```swift
let pid = Int(ProcessInfo.processInfo.processIdentifier)
```

Which is already cross-platform.

Sources/SourceKit/sourcekitd/SwiftSourceKitFramework.swift
```diff
@@ -42,7 +42,11 @@ final class SwiftSourceKitFramework {
 
   init(dylib path: AbsolutePath) throws {
     self.path = path
+    #if os(Windows)
+    self.dylib = try dlopen(path.pathString, mode: [])
```
pvieito: `dlopen` is supported on Windows? I think we should use [`LoadLibraryA`](https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-loadlibrarya).

Sources/SourceKit/SourceKitServer.swift
gmittert: Ah fair point, I didn't realize that existed. Plus GetCurrentProcess gives a handle instead of a pid which now that I think about it, is probably the wrong thing to identify a process with. I'll swap it.

PR #103, opened 2019-05-01
[5.1] Update swift-5.1-branch to match the swiftpm branch point

benlangmuir:
Semantically, we want our branch point to match swiftpm, but all of the
changes since then also look good to take into the branch, so just do
one last merge from master.


benlangmuir:
@swift-ci please test

PR #104, opened 2019-05-07
Use SPM UserToolchain. …

krzyzanowskim:
- Use SPM UserToolchain
- Use proper toolchain to compile and manifest loading.

It looks like we don't need to copy UserToolchain any longer.


benlangmuir:
@swift-ci please test

PR #110, opened 2019-06-17
Use name, not signature of the symbol

krzyzanowskim:
Use name, not signature. SourceKit return signature rather than nameeg. for function.

### DIFF
--- a/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
+++ b/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
@@ -497,7 +497,7 @@ extension SwiftLanguageServer {
       }
 
       func documentSymbol(value: SKResponseDictionary) -> DocumentSymbol? {
-        guard let name: String = value[self.keys.name],
+        guard let signature: String = value[self.keys.name],
               let uid: sourcekitd_uid_t = value[self.keys.kind],
               let kind: SymbolKind = uid.asSymbolKind(self.values),
               let offset: Int = value[self.keys.offset],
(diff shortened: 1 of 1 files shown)


Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -507,13 +507,14 @@ extension SwiftLanguageServer {
           return nil
         }
         
+        let name = String(signature.prefix(while: { $0.isLetter }))
```
krzyzanowskim: wild guess. Is `isLetter` compatible with grammar?

benlangmuir:
Can you clarify what behaviour you are seeing vs. what you expect?  In Swift the argument labels are part of the name, so I would expect them to show up.

krzyzanowskim:
for functions, a signature is `init(_:)` or `foo()` while the name is `init` or `foo`. LSP defines just a name for `name`, and signature for `details`.

benlangmuir:
In Swift, the terminology would be base name = `foo`, name = `foo(x:)`, signature = `foo(x: Int) -> Float`.  Is there a concrete issue showing the full name in editors? VSCode has no trouble with it, for example.

krzyzanowskim:
SourceKit return parameters as children (see below), that could be a separate symbol as far as I understand.

```
    {
      key.kind: source.lang.swift.decl.function.method.instance,
      key.accessibility: source.lang.swift.accessibility.internal,
      key.name: "name(_:)",
      key.offset: 41,
      key.length: 43,
      key.nameoffset: 46,
      key.namelength: 21,
      key.bodyoffset: 69,
      key.bodylength: 14,
      key.substructure: [
        {
          key.kind: source.lang.swift.decl.var.parameter,
          key.name: "param",
          key.offset: 51,
          key.length: 15,
          key.typename: "String",
          key.nameoffset: 0,
          key.namelength: 0
        }
      ]
    }
```

krzyzanowskim:
> In Swift, the terminology would be base name = foo, name = foo(x:), signature = foo(x: Int) -> Float. Is there a concrete issue showing the full name in editors? VSCode has no trouble with it, for example.

I didn't look what VSCode is doing with the name. Where this call is used in VSCode?

I expect here *base name* eg. to locate the function name, not the function declaration. Now I have doubts though. I may be wrong in spec interpretation.

benlangmuir:
> I didn't look what VSCode is doing with the name. Where this call is used in VSCode?

The "Outline" view.

> I expect here base name eg. to locate the function name, not the function declaration. Now I have doubts though. I may be wrong in spec interpretation.

The LSP spec doesn't really say anything about this - it just requires that it's a displayable string (below).  I think it's reasonable for a language server to use its own semantics for what is a "name", and in Swift I think the full name is a better choice than the base name.

```typescript
        /**
	 * The name of this symbol. Will be displayed in the user interface and therefore must not be
	 * an empty string or a string only consisting of white spaces.
	 */
	name: string;
```

krzyzanowskim:
>  I think it's reasonable for a language server to use its own semantics for what is a "name", and in Swift I think the full name is a better choice than the base name.

Agree. Thanks for pointing this out

PR #111, opened 2019-06-25
Add markdown escaping for names in hover request

Trzyipolkostkicukru:
Very often when hovering over methods with unnamed arguments, italicized colons start showing up. That's because underscores are being interpreted as markdown. This commit fixes that.

I'm not sure if escaping *everything* is a good idea, or maybe I should only escape `*` and `_`?

### DIFF
--- a/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
+++ b/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
@@ -444,22 +444,13 @@ extension SwiftLanguageServer {
         return
       }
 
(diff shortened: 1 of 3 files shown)


benlangmuir:
@swift-ci please test

Trzyipolkostkicukru:
I noticed that Sublime Text doesn't like escaping colons, so I made it so only `*` and `_` are escaped. I edited the tests, but didn't run them locally, because SPM is complaining about something on linux, and I don't have a mac with me right now.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #112, opened 2019-06-25
Add support for JSON-XPC transport for the ClangdXPC.framework

hyp:
This patch adds support for the XPC transport layer that is used by ClangdXPC.framework.

A couple of notes:
- A lot of code between the JSON and XPC connection is the same. I think it might be possible to share it by using a common base class / protocol. @benlangmuir, WDYT?
- Tested locally, but swift-ci doesn't build the XPC framework yet, so it won't be tested in CI. I'll have to fix that too.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONXPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONXPCConnection.swift
@@ -101,13 +101,16 @@ public final class JSONXPCConnection: Connection {
             }
             let instance = Unmanaged<JSONXPCConnection>.fromOpaque(pointer).takeUnretainedValue()
 
-            guard let result = "LSP".withCString ({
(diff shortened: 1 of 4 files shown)


benlangmuir:
Could you say a bit about the choice not to use XPC's own API for sending a message with a reply?

hyp:
> Could you say a bit about the choice not to use XPC's own API for sending a message with a reply?

It looks like `send_with_reply` only catches the reply if the server sends the response using the `xpc_dictionary_create_reply` function. Unfortunately we currently don't support that in Clangd, and we send responses back without preserving the XPC reply link. Do you think that should be fixed first?

hyp:
Updated to use XPC data instead of string to match : https://reviews.llvm.org/D63961

shahmishal:

PR #113, opened 2019-06-27
Marcin/compile arguments

krzyzanowskim:
- Build the requested file only instead of the whole module. This should help with SourceKit response time, especially when the project doesn't compile successfully.
- `-incremental` has no effect without `output-file-map`

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -266,22 +266,25 @@ extension SwiftPMWorkspace {
     forSwiftFile path: AbsolutePath,
     _ td: SwiftTargetBuildDescription) -> FileBuildSettings?
   {
+    guard td.target.sources.paths.first(where: { $0 == path }) != nil else {
+      log("could not find \(path) in the target: \(td.target.name).", level: .warning)
+      return nil
+    }
+
     // FIXME: this is re-implementing llbuild's constructCommandLineArgs.
     var args: [String] = [
       "-module-name",
       td.target.c99name,
-      "-incremental",
       "-emit-dependencies",
       "-emit-module",
       "-emit-module-path",
       buildPath.appending(component: "\(td.target.c99name).swiftmodule").pathString
-      // -output-file-map <path>
     ]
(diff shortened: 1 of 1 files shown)


benlangmuir:
This will break any code split across multiple files.  At the sourcekitd level we already setting up the `-primary-file` correctly, so we only do full typechecking in the primary file.

benlangmuir:
The change to `shouldEnableManifestCaching` is probably fine though

krzyzanowskim:
Q: Why settings(forSwiftFile..) has a path parameter? it's not, is it?
Q: Where is the "-primary-file" set exactly?

I have a simple (Kitura hello world) case where it takes 19s. to parse 4 swift files. At this is how long it takes to respond SourceKit for a request.

Screenshot 2019-06-27 at 20 18 39

benlangmuir:
> Q: Why settings(forSwiftFile..) has a path parameter? it's not, is it?

It is looking up the compiler arguments for this particular file.  The fact that all the files in the same module will often get the same settings doesn't really matter.

> Q: Where is the "-primary-file" set exactly?

When we query sourcekitd, we send the primary file using `key.name`, or `key.sourcefile` depending on the request, and separately send the compiler arguments which contain all the files in that module.

```
{
  ...
  key.name: "/path/to/foo.swift",
  key.compilerargs : [
    "/path/to/foo.swift",
    "/path/to/bar.swift",
    ...
  ]
}
```

When we setup the compiler invocation, we pass in the primary file:
https://github.com/apple/swift/blob/3e3770fe4228836d08eb9ff88802246aff45aca0/tools/SourceKit/lib/SwiftLang/SwiftASTManager.cpp#L439

Ultimately the function `resolveSymbolicLinksInInputs` is what sets the "isPrimary" bit on the correct input.  This is just an internal way to do the same thing as `-primary-file` does.

---

I'm not sure how to interpret your timings, but in general sourcektid AST building will be roughly the same as building with a single `-primary-file` and only compiling up to the SIL diagnostics phase (no optimization or irgen).

krzyzanowskim:
thanks! makes sense of course. I don't know myself what happened with the times there.

benlangmuir:
Would you mind sending a new PR for the `shouldEnableManifestCaching` change? That seems like it would be good regardless :-)

PR #114, opened 2019-06-28
Add ExecuteCommand/ApplyEdit

rockbruno:
Part 2 to https://github.com/apple/sourcekit-lsp/pull/70: This adds the skeleton for command execution and the resulting edits that are sent back to client.

The thing to note here is that `ExecuteCommand` isn't a `TextDocumentRequest` in the LSP, but we still need to determine which server we send the request to. My solution for this involved creating a special type of `SwiftCommand` that can be detected by the use of a `swift.lsp` prefix in the command identifier, but I would love to know if there are cleaner ways to handle this.

EDIT: It now simply intercepts the CodeAction result and appends a metadata object containing the URL.


Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,110 @@
...
+  static var swiftCommandIdentifier: String { get }
+}
+
+extension SwiftCommand {
+  public static var identifier: String {
+    return Command.swiftCommandIdentifierPrefix + Self.swiftCommandIdentifier
+  }
+
+  /// Converts this `SwiftCommand` to a generic LSP `Command` object.
+  public func asCommand() throws -> Command {
```
rockbruno: You can see how this gets used here:
https://github.com/apple/sourcekit-lsp/pull/70/files#diff-6175f3a32db6757aaeb5b49a98bab346R949

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+extension ExecuteCommandRequest {
+  /// Attempts to convert the underlying `Command` metadata from this request
+  /// to a specific Swift language server `SwiftCommand`.
+  ///
+  /// - Parameters:
+  ///   - type: The `SwiftCommand` metatype to convert to.
+  public func swiftCommand<T: SwiftCommand>(ofType type: T.Type) -> T? {
```
rockbruno: You can see how this gets used here:
https://github.com/apple/sourcekit-lsp/pull/70/files#diff-6175f3a32db6757aaeb5b49a98bab346R969

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -888,6 +891,34 @@ extension SwiftLanguageServer {
       }
     }
   }
+
+  func executeCommand(_ req: Request<ExecuteCommandRequest>) {
```
rockbruno: It's outdated, but the actual request is here:
https://github.com/apple/sourcekit-lsp/pull/70/files#diff-6175f3a32db6757aaeb5b49a98bab346R966

Tests/SourceKitTests/ExecuteCommandTests.swift
```diff
@@ -0,0 +1,94 @@
...
+      workspaceFolders: nil))
+
+    workspace = connection.server!.workspace!
+  }
+
+  func testCommandIsRoutedToTheCorrectServer() {
+    let haveClangd = ToolchainRegistry.shared.toolchains.contains { $0.clangd != nil }
+    if !haveClangd {
+      XCTFail("Cannot find clangd in toolchain")
+    }
```
rockbruno: I actually wasn't able to get this or anything in LocalClangTests to pass, I haven't got the right toolchains yet.

Darrius4g: Thanks

Darrius4g: Thanks

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -367,6 +371,31 @@ extension SourceKitServer {
...
+      req.reply(result)
+    }
+    req.cancellationToken.addCancellationHandler { [weak service] in
+      service?.send(CancelRequest(id: id))
+    }
+  }
+
+  func serviceFor(executeCommandRequest req: ExecuteCommandRequest, workspace: Workspace) -> Connection? {
+    // FIXME: ExecuteCommand requests have no URL associated to it, but we need to determine the server
+    // that it gets sent to. There should be a better way to do this.
```
hanjoes: can we pass url back in the `argument` field of `Commands` returned by `CodeActionResponse`?

rockbruno: We can for Swift, but AFAIK we have no control over what clangd's commands will look like

hanjoes: please correct me if i'm missing something, I assume parameters passed in the `CodeActionResponse` will be sent back to server when user picks the command. If that's the case, even for Clangd, server should still be the best place to generate those arguments, no?

rockbruno:
True, I believe we can intercept the response from clang and inject the url that called the code action itself into the arguments. It should be enough to avoid this logic, so we just need to confirm that doing so won't be a problem for clangd.

Sources/SourceKit/SourceKitServer.swift
rockbruno: True, I believe we can intercept the response from clang and inject the url that called the code action itself into the arguments. It should be enough to avoid this logic, so we just need to confirm that doing so won't be a problem for clangd.

rockbruno: True, I believe we can intercept the response from clang and inject the url that called the code action itself into the arguments. It should be enough to avoid this logic, so we just need to confirm that doing so won't be a problem for clangd.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -269,6 +270,9 @@ extension SourceKitServer {
         clientCapabilities: req.params.capabilities.textDocument?.codeAction,
         codeActionOptions: CodeActionOptions(codeActionKinds: nil),
         supportsCodeActions: false // TODO: Turn it on after a provider is implemented.
+      ),
+      executeCommandProvider: ExecuteCommandOptions(
+        commands: builtinSwiftCommands // FIXME: Clangd commands?
```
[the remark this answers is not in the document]
Darrius4g: You can make any steps to get closer to opening it still learning

rockbruno:
Done, I added a result handler to the CodeActions that intercepts the original result and adds a metadata object to the commands. Worked great!

Sources/LanguageServerProtocol/CodeAction.swift
```diff
@@ -47,6 +48,30 @@ public struct CodeActionRequest: TextDocumentRequest, Hashable {
     self.context = context
     self.textDocument = textDocument
   }
+
+  public func injectMetadata(atResponse response: CodeActionRequestResponse?) -> CodeActionRequestResponse? {
+    let metadata = SourceKitLSPCommandMetadata(textDocument: textDocument)
+    guard let data = try? JSONEncoder().encode(metadata),
+          let metadataArgument = try? JSONDecoder().decode(CommandArgumentType.self, from: data) else
```
benlangmuir: We should just construct the dictionary instead of round-tripping through json.  The json (de) serialization code has a relatively high up front cost even for small payloads.

Sources/LanguageServerProtocol/CodeAction.swift
```diff
@@ -47,6 +48,30 @@ public struct CodeActionRequest: TextDocumentRequest, Hashable {
     self.context = context
     self.textDocument = textDocument
   }
+
+  public func injectMetadata(atResponse response: CodeActionRequestResponse?) -> CodeActionRequestResponse? {
```
benlangmuir: This implementation is not really a general feature of CodeAction, so it should probably move into an extension the SourceKit module.

Sources/LanguageServerProtocol/ExecuteCommand.swift
```diff
@@ -0,0 +1,69 @@
...
+  /// Arguments that the command should be invoked with.
+  public var arguments: [CommandArgumentType]?
+
+  /// The document in which the command was invoked.
+  public var textDocument: TextDocumentIdentifier? {
+    return metadata?.textDocument
+  }
+
+  /// Optional metadata containing SourceKit-LSP infomration about this command.
+  public var metadata: SourceKitLSPCommandMetadata? {
```
benlangmuir: Same here, I'd prefer this moved to extensions in the SourceKit module.

Sources/LanguageServerProtocol/ExecuteCommand.swift
```diff
@@ -0,0 +1,69 @@
...
+    guard case .dictionary(let dictionary)? = arguments?.last else {
+      return nil
+    }
+    guard let data = try? JSONEncoder().encode(dictionary) else {
+      return nil
+    }
+    return try? JSONDecoder().decode(SourceKitLSPCommandMetadata.self, from: data)
```
benlangmuir: Similar to above, we already have a dictionary, so we should map it to our metadata directly without going back through json.

Sources/SourceKit/SourceKitServer.swift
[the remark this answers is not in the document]
benlangmuir: Hmm, that's unfortunate.  We load clangd dynamically through the toolchain, so there's no good way for us to get the response from clang this early.  Maybe someday this would be a good place for the specification to evolve to add dynamic capabilities for the server (there are already dynamic capabilities for the *client*). As a short term hack, we could maintain a list of every command clangd supports, but that doesn't need to be done in this patch.

For now, this is fine.

Sources/LanguageServerProtocol/ExecuteCommand.swift
```diff
@@ -0,0 +1,69 @@
...
+  public init(command: String, arguments: [CommandArgumentType]?) {
+    self.command = command
+    self.arguments = arguments
+  }
+}
+
+/// Represents metadata that SourceKit-LSP injects at every command returned by code actions.
+/// The ExecuteCommand is not a TextDocumentRequest, so metadata is injected to allow SourceKit-LSP
+/// to determine where a command should be executed.
+public struct SourceKitLSPCommandMetadata: Codable, Hashable {
```
benlangmuir: If we're going to inject this into clang responses, I think we need to mangle all the dictionary keys so it can't collide with anything (e.g. sourcekitlsp_textDocument).

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -364,7 +368,23 @@ extension SourceKitServer {
...
+      case .success(let reply):
+        return .success(req.params.injectMetadata(atResponse: reply))
+      default:
+        return result
+      }
+    }, fallback: nil)
+  }
+
+  func executeCommand(_ req: Request<ExecuteCommandRequest>, workspace: Workspace) {
+    guard let url = req.params.textDocument?.url else {
```
benlangmuir: I wonder if we should strip the metadata off before sending it to clang.  What do you think?

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,91 @@
...
+public protocol SwiftCommand: Codable, Hashable {
+  static var identifier: String { get }
+  var title: String { get set }
+}
+
+extension SwiftCommand {
+  /// Converts this `SwiftCommand` to a generic LSP `Command` object.
+  public func asCommand() throws -> Command {
+    let data = try JSONEncoder().encode(self)
+    let argument = try JSONDecoder().decode(CommandArgumentType.self, from: data)
```
benlangmuir: More round-tripping through  JSON that we should avoid.

Sources/SourceKit/SourceKitServer.swift
rockbruno: Yeah, since at this point the correct server can be determined we can strip it for Swift as well. Should make things more predictable 👍

Sources/LanguageServerProtocol/CodeAction.swift
rockbruno: Ah, I was trying to make the compiler synthetize the usage of CommandArguments as much as possible to make it easy to add and edit commands... I changed these guys to encode and decode the dictionaries manually, but let me know if there's something that can be done to reduce the boilerplate.

benlangmuir: I'm not aware of anything good here.  In principle you could write a Decoder implementation that works directly from the dictionary - basically perform the object mapping part of JSONDecoder without doing any of the parsing.  But that seems unnecessarily general to implement in sourcekit-lsp, so unless someone adds such a thing to the stdlib or Foundation I don't see us using it.

Sources/LanguageServerProtocol/WorkspaceEdit.swift
```diff
@@ -0,0 +1,26 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A workspace edit represents changes to many resources managed in the workspace.
+public struct WorkspaceEdit: Codable, Hashable, ResponseType {
+
+  /// The edits to be applied to existing resources.
+  public var changes: [String: [TextEdit]]?
```
benlangmuir: Why not use `URL` directly here? Everywhere else that uses `DocumentUri` we are using `URL` I think.

Sources/SourceKit/SourceKitLSPCommandMetadata.swift
```diff
@@ -0,0 +1,93 @@
...
+import LanguageServerProtocol
+import Foundation
+import SKSupport
+
+/// Represents metadata that SourceKit-LSP injects at every command returned by code actions.
+/// The ExecuteCommand is not a TextDocumentRequest, so metadata is injected to allow SourceKit-LSP
+/// to determine where a command should be executed.
+public struct SourceKitLSPCommandMetadata: Codable, Hashable {
+
+  public static func decode(fromDictionary dictionary: [String: LSPAny]) -> SourceKitLSPCommandMetadata? {
```
benlangmuir: Let's rearrange this so that the instance fields and init are at the top, and the decoding and encoding are together.

Also, how about changing the signature to `init?(fromLSPDictionary:)`?  Seems like it belongs in an `init`.

Sources/SourceKit/SourceKitLSPCommandMetadata.swift
```diff
@@ -0,0 +1,93 @@
...
+  func asCommandArgument() -> LSPAny {
+    let textDocumentArgument = LSPAny.dictionary(
+      [TextDocumentIdentifier.CodingKeys.url.stringValue: .string(sourcekitlsp_textDocument.url.absoluteString)]
+    )
+    return .dictionary([CodingKeys.sourcekitlsp_textDocument.stringValue: textDocumentArgument])
+  }
+}
+
+extension CodeActionRequest {
+  public func injectMetadata(atResponse response: CodeActionRequestResponse?) -> CodeActionRequestResponse? {
```
benlangmuir: Nitpick: `toResponse` seems more grammatical to me

Sources/SourceKit/SourceKitLSPCommandMetadata.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+    guard let metadata = SourceKitLSPCommandMetadata.decode(fromDictionary: dictionary) else {
+      log("failed to decode lsp metadata in executeCommand request", level: .error)
+      return nil
+    }
+    return metadata
+  }
+
+  /// Returns this Command's arguments without SourceKit-LSP's injected metadata, if it exists.
+  public var argumentsWithoutLSPMetadata: [LSPAny]? {
```
benlangmuir: `WithoutSourceKitMetadata` seems more correct, since it's not general to LSP, but specific to SK.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -628,16 +650,28 @@ extension SourceKitServer {
   func toolchainTextDocumentRequest<PositionRequest>(
     _ req: Request<PositionRequest>,
     workspace: Workspace,
+    resultHandler: ((LSPResult<PositionRequest.Response>) -> LSPResult<PositionRequest.Response>)? = nil,
```
benlangmuir: How about `resultTransformer`? To me "handler" implies something else.

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  /// The name of this refactoring action.
+  public var title: String
+
+  /// The sourcekitd identifier of the refactoring action.
+  public var actionString: String
+
+  /// The starting line of the range to refactor.
+  public var line: Int
```
benlangmuir: Why not use a `PositionRange`?

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,122 @@
...
+import LanguageServerProtocol
+import Foundation
+
+/// The set of known Swift commands.
+///
+/// All commands from the Swift LSP should be listed here.
+public let builtinSwiftCommands: [String] = []
+
+/// A `Command` that should be executed by Swift's language server.
+public protocol SwiftCommand: Codable, Hashable {
```
benlangmuir: It looks like none of this is used yet. The basic design looks okay, but I'd like to leave this out until we have code that uses it (basically defer it to the next PR).  WDYT?

Sources/SourceKit/SourceKitLSPCommandMetadata.swift
```diff
@@ -0,0 +1,93 @@
...
+    return SourceKitLSPCommandMetadata(textDocument: textDocument)
+  }
+
+  public var sourcekitlsp_textDocument: TextDocumentIdentifier
+
+  public init(textDocument: TextDocumentIdentifier) {
+    self.sourcekitlsp_textDocument = textDocument
+  }
+
+  func asCommandArgument() -> LSPAny {
```
benlangmuir: How about `encode(toLSPAny:)` to be closer to Codable?

rockbruno:
No worries! I applied the changes here - I just wasn't sure about the `encode(toLSPAny:)` bit, I named it `encodeToLSPAny()` directly because we can't really append/construct the LSPAny type from an external reference in that context, but if you have an idea feel free to share it.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test macOS

benlangmuir:
@swift-ci please test macOS

benlangmuir:
> @testable import SourceKit

I disabled testable support so that we can test in a configuration that matches what we want to distribute. This will need to be changed to a normal import.  If there are specific internals you want to use, you can make them public, add a leading underscore to the name and put `*Public for testing*` in their documentation comment.

rockbruno:
Ah that actually wasn't needed for the tests, I think I had that added out of habit for iOS.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test macOS

PR #115, opened 2019-06-29
[SR-10809] Support clients without completion snippet capabilities

rockbruno:
Changes completion results to strip snippets if they aren't supported by the client.

Note: There are a few FIXMEs about adding the placeholder's name to the snippet, but the names in the examples differ from what the compiler would consider as the display name:
```
      // FIXME: should be "a" in the placeholder.
      XCTAssertEqual(test.insertText, "test(a: ${1:value})")
```
Is this intended? If not, I can add the compiler's editor placeholder parsing logic in this PR as well.


Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -383,20 +390,18 @@ extension SwiftLanguageServer {
     _ = handle
   }
 
-  func rewriteCompletionPlacholders(_ completion: String) -> String {
-    if !completion.contains("<#") {
-      return completion
-    }
-
-    var result = completion
+  func rewriteSourceKitPlaceholders(_ skPlaceholder: String, clientSupportsSnippets: Bool) -> String {
```
rockbruno: The reasoning to make this method generic is because they'll likely add snippet support to WorkspaceEdits in the near future. The refactors can use it as well when it happens.

Tests/SourceKitTests/CompletionTests.swift
```diff
@@ -0,0 +1,193 @@
...
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKSupport
+import SKTestSupport
+import XCTest
+
+@testable import SourceKit
+
+final class CompletionTests: XCTestCase {
```
rockbruno: I moved all completion tests to a separate class to be able to play with the capabilities, but we can make something different if the duplication of these setup methods is bad.

Sources/LanguageServerProtocol/ClientCapabilities.swift
```diff
@@ -132,6 +132,9 @@ public struct TextDocumentClientCapabilities: Hashable, Codable {
 
       /// Whether the client supports the `preselect` property on a CompletionItem.
       public var preselectSupport: Bool? = nil
+
+      public init() {
```
benlangmuir: Let's make these memberwise initializers with default values.

Tests/SourceKitTests/CompletionTests.swift
benlangmuir: Let's call it `SwiftCompletionTests`, but otherwise SGTM.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -383,20 +390,18 @@ extension SwiftLanguageServer {
...
       guard let end = result[start.upperBound...].range(of: "#>") else {
-        log("invalid placholder in \(completion)", level: .debug)
-        return completion
+        log("invalid placeholder in \(string)", level: .debug)
+        return string
       }
       // FIXME: add name to placeholder
-      result.replaceSubrange(start.lowerBound..<end.upperBound, with: "${\(index):value}")
+      let displayName = "value"
+      let placeholder = clientSupportsSnippets ? "${\(index):\(displayName)}" : displayName
```
benlangmuir: I would have expected that we would omit the string entirely.  Otherwise you have to manually delete it before you can type in the new value.

rockbruno: Good point, keeping the names will be useful for the method names in refactoring, but we can add support for that once they actually support snippets.

benlangmuir:
@swift-ci please test

PR #116, opened 2019-06-30
Implement "textDocument/rename" request

Trzyipolkostkicukru:
I'll write some test after I borrow a mac, meanwhile do you know if there is any documentation on what different `SymbolRole`s mean? I just blindly put all of them.

### DIFF
--- a/Sources/LanguageServerProtocol/Messages.swift
+++ b/Sources/LanguageServerProtocol/Messages.swift
@@ -34,6 +34,7 @@ public let builtinRequests: [_RequestType.Type] = [
   DocumentColorRequest.self,
   ColorPresentationRequest.self,
   CodeActionRequest.self,
+  RenameRequest.self,
 
   // MARK: LSP Extension Requests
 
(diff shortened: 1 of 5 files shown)


benlangmuir:
Hi @Trzyipolkostkicukru,

I appreciate you taking the time to look at this.  Unfortunately, this is not going to behave correctly for argument labels, symbols that have different names across languages, or symbols that are involved in overrides or protocol conformances.  There are also ObjC-specific cases like renaming properties that have associated ivars.   Another consideration is that in addition to the existing LSP rename API, we're also interested in bringing the "live rename" functionality that we have in Xcode, and we'd like it to share as much of the implementation as possible with the LSP rename.

This is a case where I expect it would be much easier for me to port the rename code that we have internally, including its test suite, rather than reimplement it piecemeal from scratch.

Trzyipolkostkicukru:
Oh no!

Darrius4g:
Do what you gotta do

Trzyipolkostkicukru:
By 'bringing the "live rename" functionality that we have in Xcode' do you mean adding a new unofficial request to the protocol?

Darrius4g:
The name can be IsrarTv
An yes

benlangmuir:
@Trzyipolkostkicukru yes, it would be an extension.

PR #117, opened 2019-07-06
Add EditorPlaceholder parsing for completion snippets

rockbruno:
Porting the editor placeholder component from the compiler so we can show a completion snippet's true contents.

### DIFF
--- a/Tests/SourceKitTests/SwiftCompletionTests.swift
+++ b/Tests/SourceKitTests/SwiftCompletionTests.swift
@@ -142,8 +142,7 @@ final class SwiftCompletionTests: XCTestCase {
       XCTAssertEqual(test.filterText, "test(a:)")
       // FIXME:
       XCTAssertNil(test.textEdit)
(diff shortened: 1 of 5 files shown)


Tests/SourceKitTests/SwiftCompletionTests.swift
```diff
@@ -142,8 +142,7 @@ final class SwiftCompletionTests: XCTestCase {
       XCTAssertEqual(test.filterText, "test(a:)")
       // FIXME:
       XCTAssertNil(test.textEdit)
-      // FIXME: should be "a" in the placeholder.
-      XCTAssertEqual(test.insertText, "test(a: ${1:value})")
+      XCTAssertEqual(test.insertText, "test(a: ${1:Int})")
```
benlangmuir: We should add a test for completion where the type and display name are not the same:
```
func test(_ b: Int) // <#T##b: Int##Int#>
```

benlangmuir:
@swift-ci please test

PR #118, opened 2019-07-06
Add workspace/symbols support

literalpie:
https://bugs.swift.org/browse/SR-10806

Feedback welcome!

This branch builds on changes made in [this PR](https://github.com/apple/indexstore-db/pull/29) of indexstore-db

This is the feature that is activated in VSCode with CMD+T

### DIFF
--- a/Sources/SourceKit/SourceKitServer.swift
+++ b/Sources/SourceKit/SourceKitServer.swift
@@ -71,6 +71,7 @@ public final class SourceKitServer: LanguageServer {
     registerWorkspaceNotfication(SourceKitServer.didSaveDocument)
     registerWorkspaceRequest(SourceKitServer.completion)
     registerWorkspaceRequest(SourceKitServer.hover)
+    registerWorkspaceRequest(SourceKitServer.workspaceSymbols)
     registerWorkspaceRequest(SourceKitServer.definition)
     registerWorkspaceRequest(SourceKitServer.references)
     registerWorkspaceRequest(SourceKitServer.documentSymbolHighlight)
@@ -269,7 +270,8 @@ extension SourceKitServer {
         clientCapabilities: req.params.capabilities.textDocument?.codeAction,
         codeActionOptions: CodeActionOptions(codeActionKinds: nil),
         supportsCodeActions: false // TODO: Turn it on after a provider is implemented.
-      )
+      ),
+      workspaceSymbolProvider: true
     )))
   }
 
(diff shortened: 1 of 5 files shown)


Sources/SourceKit/SourceKitServer.swift
```diff
@@ -338,6 +340,47 @@ extension SourceKitServer {
...
+    ) {symbol in
+      if !symbol.location.isSystem && !symbol.roles.contains(.accessorOf) {
+        symbolOccurenceResults.append(symbol)
+      }
+      return true
+    }
+    return symbolOccurenceResults
+  }
+
+  func workspaceSymbols(_ req: Request<WorkspaceSymbolsRequest>, workspace: Workspace) {
```
literalpie: Much of the other functionality seems to be done in SwiftLanguageServer, but I think I need to do this here to have access to index. Is that correct?

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -338,6 +340,47 @@ extension SourceKitServer {
...
+    ).map({symbolOccurrence -> SymbolInformation in
+      let symbolPosition = Position(
+        line: symbolOccurrence.location.line - 1, // 1-based -> 0-based
+        // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
+        utf16index: symbolOccurrence.location.utf8Column - 1)
+
+      let symbolLocation = Location(
+        url: URL(fileURLWithPath: symbolOccurrence.location.path),
+        range: Range(symbolPosition))
+      // TODO: symbol kind, detail, and containerName should be added
```
literalpie: I would like some guidance on how to get kind, detail, and containerName.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -338,6 +340,47 @@ extension SourceKitServer {
...
+  func findWorkspaceSymbols(matching: String) -> [SymbolOccurrence] {
+    var symbolOccurenceResults: [SymbolOccurrence] = []
+    workspace?.index?.forEachCanonicalSymbolOccurrence(
+      containing: matching,
+      anchorStart: false,
+      anchorEnd: false,
+      subsequence: true,
+      ignoreCase: true
+    ) {symbol in
+      if !symbol.location.isSystem && !symbol.roles.contains(.accessorOf) {
```
literalpie: I'm not sure what roles should be included. I've excluded `accessorOf` because otherwise getters and setters of variables were returned in addition to the variable itself.

Sources/SourceKit/SourceKitServer.swift
benlangmuir: Yep, this is the right place for it.  Language-agnostic functionality belongs here, which is why we also implement e.g. jump to definition here.

Sources/SourceKit/SourceKitServer.swift
benlangmuir: * kind: We have a symbol kind in the index that we should expose on `class Symbol`. Then we can convert it to an LSP `SymbolKind` here.

* detail: This isn't a field on `SymbolInformation` AFAICT.  Did you mix it up with `DocumentSymbol`?

* containerName:  Ah didn't realize we'd need this.  For this we would need to expose the `SymbolOccurrence`'s relations in the index.  The C++ version of this is `SymbolOccurrence::getRelations`, we just need to thread it through the C and Swift interfaces.  Since you already have the canonical occurrence, you would find the related symbol that has relation role `childOf` and use that symbol's name.

literalpie: You are correct that detail came from `DocumentSymbol`.

for containerName, it looks like I will have to pass an ArrayPointer to Swift, which I can't find a good example of. Maybe I could use `unsafeBufferPointer` as mentioned [here](https://stackoverflow.com/questions/39666832/converting-c-array-ptr-to-native-array-in-swift-3), but it looks like that involves copying which I think we want to avoid. is there a better way to do this?

benlangmuir: Copying should be fine, since the number of relations for a single occurrence will generally be small, and we can do the copy lazily only when you ask for related symbols.  Here's a sketch of the C API I would start with.  We can always optimize this later if needed.

```c++
typedef void *indexstoredb_symbol_relation_t;

...

INDEXSTOREDB_PUBLIC uint64_t
indexstoredb_symbol_relation_get_roles(_Nonnull  indexstoredb_symbol_relation_t);

INDEXSTOREDB_PUBLIC indexstoredb_symbol_t
indexstoredb_symbol_relation_get_symbol(_Nonnull indexstoredb_symbol_relation_t);

/// The relations are owned by the occurrence and shall not be used after the occurrence is freed.
INDEXSTOREDB_PUBLIC bool
indexstoredb_symbol_occurrence_relations(_Nonnull indexstoredb_occurrence_t,
                      bool(^applier)(indexstoredb_symbol_relation_t));
```

On the Swift side, this can lazily create an array from the occurrence.

PR #119, opened 2019-07-08
Bump swift-tools-version to 5.1

benlangmuir:
Changes in swiftpm require swift-5.1, so bump our swift-tools-version to make it explicit. And fix a new deprecation warning.


benlangmuir:
@swift-ci please test

PR #122, opened 2019-07-15
Convert Language enum to struct

krzyzanowskim:
Convert `Language` to a struct.

Reasoning:
- It is defined as a string parameter of `TextDocumentItem` (specification recommends a lot more identifiers, that `sourcekit-lsp` won't support)
- as a struct, it can be extended outside the `LanguageServerProtocol` module if needed.


Sources/LanguageServerProtocol/Language.swift
```diff
@@ -11,23 +11,18 @@
...
-  case c
-  case cpp // C++, not C preprocessor
-  case objective_c = "objective-c"
-  case objective_cpp = "objective-cpp"
-  case swift
+public struct Language: RawRepresentable, Codable, Equatable, Hashable {
+  public typealias LanguageId = String
 
-  case unknown
+  public static let c = Language(rawValue: "c")
```
benlangmuir: Please add type annotations to all struct/class properties.

Sources/LanguageServerProtocol/Language.swift
```diff
@@ -11,23 +11,18 @@
...
 
 /// A source code language identifier, such as "swift", or "objective-c".
-public enum Language: String, Codable, Hashable {
-  case c
-  case cpp // C++, not C preprocessor
-  case objective_c = "objective-c"
-  case objective_cpp = "objective-cpp"
-  case swift
+public struct Language: RawRepresentable, Codable, Equatable, Hashable {
+  public typealias LanguageId = String
```
benlangmuir: Is it useful to have the extra type `Language.LanguageId`, or should we just use `RawValue` which is needed anyway?

Sources/LanguageServerProtocol/Language.swift
```diff
@@ -38,12 +33,22 @@ public enum Language: String, Codable, Hashable {
...
+      default: return nil
     }
   }
 
   /// Clang-compatible language name for a header file. See `xflag`.
   public var xflagHeader: String? {
     return xflag.map { "\($0)-header" }
   }
+
+  public static func ~= (lhs: Language, rhs: Language) -> Bool {
```
benlangmuir: Why is this needed? I would have expected being Equatable was enough.

krzyzanowskim: not needed

Sources/LanguageServerProtocol/Language.swift
krzyzanowskim: I like to cover generic types with more descriptive alias, but I can ditch it if you prefer not

benlangmuir: Okay, doesn't make much difference to me.

benlangmuir:
@swift-ci please test

benlangmuir:
```
CodingTests.swift:206: error: CodingTests.testValueCoding : XCTAssertEqual failed: ("unknown") is not equal to ("asdfAF") - 
```

Sources/SKTestSupport/UnknownLanguage.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+
+public extension Language {
+  static let unknown: Language = Language(rawValue: "unknown")
```
benlangmuir: Is this used anywhere other than the line
```
checkCoding(Language.unknown, json: "\"unknown\"")
```

in the CodingTests.swift above the one you removed?  If so, we don't need to expose this static property at all and can just check coding on `Language(rawValue: "unknown")`.

krzyzanowskim: done

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #123, opened 2019-07-17
Add workspace/symbols support

literalpie:
https://bugs.swift.org/browse/SR-10806

Continuing work from #118 now that symbol relations [has been added](https://github.com/apple/indexstore-db/pull/30) to indexstore-db

### DIFF
--- a/Sources/LanguageServerProtocol/WorkspaceSymbols.swift
+++ b/Sources/LanguageServerProtocol/WorkspaceSymbols.swift
@@ -1,0 +1,55 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request for all symbols that match a certain query string.
+///
+/// This request looks up the canonical occurence of each symbol which has a name that contains the query string.
+/// The list of symbol information is returned
+///
+/// Servers that provide workspace symbol queries should set the `workspaceSymbolProvider` server capability.
+///
+/// - Parameters:
+///   - query: The string that should be looked for in symbols of the workspace.
+///
+/// - Returns: Information about each symbol with a name that contains the query string
+public struct WorkspaceSymbolsRequest: RequestType, Hashable {
+
+  public static let method: String = "workspace/symbol"
+  public typealias Response = [SymbolInformation]?
+
+  /// The document in which to lookup the symbol location.
+  public var query: String
+
+  public init(query: String) {
+    self.query = query
+  }
+}
+
+public struct SymbolInformation: Hashable, ResponseType {
+  public var name: String
+
+  public var kind: SymbolKind
+
+  public var deprecated: Bool?
+
+  public var location: Location
+
+  public var containerName: String?
+
+  public init(name: String, kind: SymbolKind, deprecated: Bool?, location: Location, containerName: String?) {
+    self.name = name
+    self.kind = kind
+    self.deprecated = deprecated
+    self.location = location
+    self.containerName = containerName
+  }
+}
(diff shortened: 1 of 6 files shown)


Sources/LanguageServerProtocol/WorkspaceSymbols.swift
```diff
@@ -0,0 +1,55 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
```
benlangmuir: 2019

Sources/LanguageServerProtocol/WorkspaceSymbols.swift
```diff
@@ -0,0 +1,55 @@
...
+    self.query = query
+  }
+}
+
+public struct SymbolInformation: Hashable, ResponseType {
+  public var name: String
+
+  public var kind: SymbolKind
+
+  public var deprecated: Bool
```
benlangmuir: Should be `Bool?` per the spec.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -523,3 +570,54 @@ public func languageService(
 
 public typealias Notification = LanguageServerProtocol.Notification
 public typealias Diagnostic = LanguageServerProtocol.Diagnostic
+
+extension IndexSymbolKind {
+  func toLspSymbolKind() -> SymbolKind {
```
benlangmuir: Nitpick: We've been using `asFoo` naming convention rather than `toFoo`.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -523,3 +570,54 @@ public func languageService(
...
+    switch self {
+    case Self.class:
```
benlangmuir: Style: prefer `.class` over `Self.class` for enum and OptionSet values.  Also, I would hoist the cases onto a single line since there aren't that many of them.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -523,3 +570,54 @@ public func languageService(
...
+      Self.instanceMethod,
+      Self.staticMethod:
+      return .method
+    case Self.instanceProperty,
+      Self.staticProperty,
+      Self.classProperty:
+      return .property
+    case Self.enum:
+      return .enum
+    case Self.enumConstant: // Unsure about this one
```
benlangmuir: Looks right to me.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -523,3 +570,54 @@ public func languageService(
...
+      Self.conversionFunction: // assuming conversion function is function
+      return .function
+    case Self.variable:
+      return .variable
+    case Self.struct:
+      return .struct
+    case Self.parameter:
+      return .typeParameter
+    case Self.parameter:
+      return .typeParameter
```
benlangmuir: Duplicate case?

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -523,3 +570,54 @@ public func languageService(
...
+
+extension SymbolOccurrence {
+  /// Get the name of the symbol that is a parent of this symbol, if one exists
+  func getContainerName() -> String? {
+    var foundChildRelation: SymbolRelation?
+    self.forEachRelation { relation in
+      if relation.roles.contains(.childOf) {
+        foundChildRelation = relation
+      }
+      return false
```
benlangmuir: Shouldn't this be `return false` if we found one, but `return true` to keep looking?

Sources/LanguageServerProtocol/WorkspaceSymbols.swift
literalpie: oh yes, and I'll return nil instead of returning false for everything

literalpie: Let me know if you want me to include the deprecated property as part of this change, but it seemed like it could be a lot of work for a small gain.

benlangmuir: I don't think we even have enough information for that in the index today, so `nil` is fine.

literalpie:
Feedback should be resolved

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #124, opened 2019-07-22
[build-script] Accept --ninja-bin

benlangmuir:
For now, just thread it through as an environment variable. Future
commits will make use of this during testing.


benlangmuir:
Please test with following pull requests:
https://github.com/apple/swift/pull/26269
https://github.com/apple/indexstore-db/pull/31

@swift-ci please test

PR #126, opened 2019-07-24
Implement "textDocument/implementation" request

Trzyipolkostkicukru:
https://bugs.swift.org/browse/SR-10808

Won't build right now, because it relies on `occurrences(relatedToUSR:roles:)`  that will be introduced in https://github.com/apple/indexstore-db/pull/32 

There's a bug where if you satisfy a protocol requirement in the main body of a type, but conform to the protocol in an extension, indexstore-db will give you the location when looking for the protocol requirement implementation.


benlangmuir:
It looks like this needs an update now that `workspace/symbols` has landed.

I also recently added the missing test support code that will allow you to write tests for this, so let's do that.  I'd love any feedback you have about using the new APIs.  You can find an example test in `testIndexSwiftModules`, which includes a `textDocument/definition` request.  I also wrote a bunch of information about the new infrastructure in https://github.com/apple/sourcekit-lsp/pull/131 and https://github.com/apple/indexstore-db/pull/35.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -468,6 +470,57 @@ extension SourceKitServer {
...
+      guard let usr = symbol.usr, let index = workspace.index else {
+        return req.reply([])
+      }
+    
+      var occurs = index.occurrences(ofUSR: usr, roles: .baseOf)
+      if occurs.isEmpty {
+        occurs = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
+      }
+
+      // FIXME: overrided method logic
```
benlangmuir: What is this FIXME about?

Tests/SourceKitTests/ImplementationTests.swift
```diff
@@ -0,0 +1,67 @@
...
+final class ImplementationTests: XCTestCase {
+  func testImplementation() throws {
+    let ws = try staticSourceKitTibsWorkspace(name: "Implementation")!
+    try ws.buildAndIndex()
+
+    try ws.openDocument(ws.testLoc("a.swift").url, language: .swift)
+    try ws.openDocument(ws.testLoc("b.swift").url, language: .swift)
+
+    func impls(at testLoc: TestLocation) throws -> [Location] {
+      let textDocument = TextDocumentIdentifier(testLoc.url)
```
benlangmuir: I added a shortcut `testLoc.docIdentifier`

Tests/SourceKitTests/ImplementationTests.swift
```diff
@@ -0,0 +1,67 @@
...
+    try XCTAssertEqual(impls(at: testLoc("ProtocolStaticFunction")), [loc("StructStaticFunction")])
+    try XCTAssertEqual(impls(at: testLoc("ProtocolVariable")), [loc("StructVariable")])
+    try XCTAssertEqual(impls(at: testLoc("ProtocolFunction")), [loc("StructFunction")])
+    try XCTAssertEqual(impls(at: testLoc("Class")), [loc("SubclassConformance")])
+    try XCTAssertEqual(impls(at: testLoc("ClassClassVar")), [loc("SubclassClassVar")])
+    try XCTAssertEqual(impls(at: testLoc("ClassClassFunction")), [loc("SubclassClassFunction")])
+    try XCTAssertEqual(impls(at: testLoc("ClassVariable")), [loc("SubclassVariable")])
+    try XCTAssertEqual(impls(at: testLoc("ClassFunction")), [loc("SubclassFunction")])
+
+    try XCTAssertEqual(impls(at: testLoc("Sepulcidae")), [loc("ParapamphiliinaeConformance"), loc("XyelulinaeConformance"), loc("TrematothoracinaeConformance")])
```
benlangmuir: The order of these is not guaranteed in the index.  Maybe we should change `impls` to return a `Set` for testing purposes?

Tests/SourceKitTests/ImplementationTests.swift
```diff
@@ -0,0 +1,67 @@
...
+    try XCTAssertEqual(impls(at: testLoc("Parapamphiliinae")), [loc("MicramphiliusConformance"), loc("PamparaphiliusConformance")])
+    try XCTAssertEqual(impls(at: testLoc("Xyelulinae")), [loc("XyelulaConformance")])
+    try XCTAssertEqual(impls(at: testLoc("Trematothoracinae")), [])
+
+    try XCTAssertEqual(impls(at: testLoc("Prozaiczne")), [loc("MurkwiaConformance2"), loc("SepulkaConformance1")])
+    // FIXME: non-ascii characters break Tibs `TestLocation`s
```
benlangmuir: What is the issue here?

Tests/SourceKitTests/ImplementationTests.swift
```diff
@@ -0,0 +1,67 @@
...
+    // try XCTAssertEqual(impls(at: testLoc("Sepulkowate")), [loc("MurkwiaConformance1"), loc("SepulkaConformance2"), loc("PćmaŁagodnaConformance"), loc("PćmaZwyczajnaConformance")])
+    // FIXME: sourcekit returns wrong locations for the function (subclasses that don't override it, and extensions that don't implement it)
```
benlangmuir: Could you file a bug for this one? It seems worth understanding and fixing either in the index or in sourcekit-lsp.

Sources/LanguageServerProtocol/ServerCapabilities.swift
```diff
@@ -54,6 +54,9 @@ public struct ServerCapabilities: Codable, Hashable {
   /// The server provides workspace symbol support.
   public var workspaceSymbolProvider: Bool?
 
+  /// Whether the server provides "textDocument/implementation".
+  public var implementationProvider: Bool?
```
benlangmuir: Nitpick: could you move this up to right after `definitionProvider` and  do the same in the initializer?

Tests/SourceKitTests/ImplementationTests.swift
Trzyipolkostkicukru: I got confused by late night debugging. There is a tibs bug, but I accidentally dodged it here, because "ć" and "Ł" fit into a single utf-16 code unit, even though they aren't ascii. The bug that the test caught is real though and needs investigating.

About the tibs bug - if I put this in a test file
```
class 😃 {}/*a*/
class B {}/*b*/
class Ć {}/*c*/
class 👩‍👩‍👧‍👧 {}/*d*/
```
and then `print(loc("a").range.lowerBound.utf16index)`, I get 15 for all of them, which suggests that tibs counts Characters instead of utf-16 code units. I think I should get 16, 15, 15, 25 for the column numbers

benlangmuir: Ah, thanks! I'll fix it.

Sources/SourceKit/SourceKitServer.swift
Trzyipolkostkicukru: Just got copy-pasted with the rest of `definition(_:workspace:)`

benlangmuir: I don't think it's relevant here.

Trzyipolkostkicukru: deleted

Tests/SourceKitTests/ImplementationTests.swift
Trzyipolkostkicukru: changed

Tests/SourceKitTests/ImplementationTests.swift
Trzyipolkostkicukru: changed

Sources/LanguageServerProtocol/ServerCapabilities.swift
Trzyipolkostkicukru: done

Tests/SourceKitTests/ImplementationTests.swift
benlangmuir: Okay, I think I fixed this in 
https://github.com/apple/indexstore-db/pull/38
https://github.com/apple/sourcekit-lsp/pull/136

Unfortunately, this will probably expose the latent problem that we are using a utf8 index from the `SymbolOccurrence`, so your comparisons may start failing.  If that's the case, I suggest adding new initializers to `Position(badUTF16: TestLocation)` and similarly for `Location` that do the same incorrect behaviour as we do in the implementation.  It also makes it easier to find the places that should change when we someday fix the underlying utf8 vs. utf16 issue from our index data.

benlangmuir:
@swift-ci please test

Tests/SourceKitTests/ImplementationTests.swift
benlangmuir: Actually, assuming the tests pass we could just land your changes first, and I can fix up any discrepancies in my PR.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test macOS

Tests/SourceKitTests/ImplementationTests.swift
Trzyipolkostkicukru: https://bugs.swift.org/browse/SR-11276

PR #127, opened 2019-07-31
FoundationXML: Adoption

millenomi:
CommentXML.swift needs to import `FoundationXML` if available. Required to land https://github.com/apple/swift-corelibs-foundation/pull/2432, but can land independently of it.


millenomi:
@swift-ci please test

PR #130, opened 2019-08-04
[build-script-helper] Add a search path to be able to find `Block.h` from the just built toolchain

akyrtzi:
This is similar to https://github.com/apple/indexstore-db/pull/33


akyrtzi:
@swift-ci Please test

PR #132, opened 2019-08-08
[swiftpm] Handle symlinks in the workspace path

benlangmuir:
Previously we would fail to get settings if the path didn't match
exactly. For now, we try first with the given path, then fallback to the
realpath (and we always realpath the workspace).


benlangmuir:
@swift-ci please test

PR #133, opened 2019-08-08
[clangd] Set correct argv[0] for clangd to fix c++ header search

benlangmuir:
Clang looks for the C++ standard headers relative to the clang binary
that we provide in the compilation database. So pass through a real path
so that we get the correct headers.


benlangmuir:
@swift-ci please test

PR #137, opened 2019-08-09
Remove `@testable import` and `-enable-testing`

benlangmuir:
Using `-enable-testing` means we unnecessarily exposed internal symbols in our release builds and also has been hiding some cases where we were missing public API, particularly memberwise initializers.  It also meant you couldn't test a release build without passing the extra option.

Most of the changes are good general improvements, including exposing API that should be public anyway. In a few cases we marked an API public only for testing and underscored it.


benlangmuir:
@swift-ci please test

PR #138, opened 2019-08-09
make project.json spec-compliant by removing a trailing comma

Trzyipolkostkicukru:
JSON spec doesn't allow trailing commas, so my editor was complaining


benlangmuir:
@swift-ci please test

PR #139, opened 2019-08-09
 [sourcekit-lsp] Prefer containing toolchain if applicable

benlangmuir:
If the sourcekit-lsp binary is embedded inside a toolchain, pass that install path to the toolchain registry and use it by default.


benlangmuir:
@swift-ci please test

PR #142, opened 2019-08-12
 [build-script] Add installation support for sourcekit-lsp

benlangmuir:
When invoked with the `install` action, build-script-helper.py will rsync the produced sourcekit-lsp to the toolchain's bin directory. On Linux, we add an extra relative rpath to find the swift corelibs (unfotunately we currently have no way to remove the absolute rpath; the same is true for all other swift-built binaries in the toolchain). On macOS, we replace the rpath.


benlangmuir:
@swift-ci please test

benlangmuir:
https://github.com/apple/swift/pull/26619

@swift-ci please build toolchain

PR #143, opened 2019-08-15
Separate toolchain discovery from build settings

rmaz:
Conflating the toolchain discovery into file specific build settings causes multiple settings calls per file. Add a separate interface to fetch the toolchain for a workspace, and default to asking only the first build server. This fixes the duplicate settings calls and allows for faster toolchain lookup for build systems that can determine this quickly.


benlangmuir:
@swift-ci please test

PR #144, opened 2019-08-15
Pass in MessageRegistry instead of using a singleton

rmaz:
In preparation for adding other JSON RPC protocols, the MessageRegistry shared singleton has to be refactored to be passed in state to the JSON RPC connection instead. This allows for each connection to specify the messages it expects.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -51,7 +51,7 @@ public final class JSONRPCConection {
 
   var closeHandler: () -> Void
 
-  public init(messageRegistry: MessageRegistry, inFD: Int32, outFD: Int32, closeHandler: @escaping () -> Void = {}) {
+  public init(protocol messageRegistry: MessageRegistry, inFD: Int32, outFD: Int32, closeHandler: @escaping () -> Void = {}) {
     state = .created
     self.closeHandler = closeHandler
     self.messageRegistry = messageRegistry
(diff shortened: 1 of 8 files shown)


Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -50,9 +51,10 @@ public final class JSONRPCConection {
 
   var closeHandler: () -> Void
 
-  public init(inFD: Int32, outFD: Int32, closeHandler: @escaping () -> Void = {}) {
+  public init(messageRegistry: MessageRegistry, inFD: Int32, outFD: Int32, closeHandler: @escaping () -> Void = {}) {
```
benlangmuir: Maybe we could call this field "protocol" since it's basically providing the application layer protocol?

benlangmuir:
@swift-ci please test

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
rmaz: yes, that is better, will update

rmaz: well, except its a reserved keyword so would have to use funky backticks. `rpcProtocol`?

benlangmuir: `public init(protocol messageRegistry: MessageRegistry ...)` should work without backticks at the declaration or call site.  It's only treated as a keyword when used on its own.

rmaz: TIL

benlangmuir:
@swift-ci please test

rmaz:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #146, opened 2019-08-19
[readme] Update Getting Started for using a Swift.org toolchain

benlangmuir:
Now that sourcekit-lsp is available in the swift.org downloadable
toolchains, update the readme to suggest that as the best way to try out
sourcekit-lsp, and move the information about building from source into
the developer documentation.


benlangmuir:
@swift-ci please test

PR #154, opened 2019-08-29
Add BSP lifecycle notifications

rmaz:
Sends the BSP lifecycle notifications according to the spec. This PR also adds support for reading the index store path from the build server initialize response metadata.

### DIFF
--- a/Sources/SKCore/BuildServerBuildSystem.swift
+++ b/Sources/SKCore/BuildServerBuildSystem.swift
@@ -58,12 +58,14 @@ public final class BuildServerBuildSystem {
   }
 
   deinit {
-    do {
-      _ = try self.buildServer?.sendSync(ShutdownBuild())
-    } catch {
-      log("error shutting down build server: \(error)")
+    if let buildServer = self.buildServer {
+      _ = buildServer.send(ShutdownBuild(), queue: DispatchQueue.global(), reply: { result in
+        if let error = result.failure {
+          log("error shutting down build server: \(error)")
+        }
+        buildServer.send(ExitBuildNotification())
+      })
     }
-    self.buildServer?.send(ExitBuildNotification())
   }
 
   private func initializeBuildServer() throws {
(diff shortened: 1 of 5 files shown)


Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -57,6 +57,15 @@ public final class BuildServerBuildSystem {
     }
   }
 
+  deinit {
+    do {
+      _ = try self.buildServer?.sendSync(ShutdownBuild())
```
benlangmuir: Does this need to be synchronous? Seems like you could do the error logging and exit notification from inside the completion block of an async request.

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -78,12 +87,36 @@ public final class BuildServerBuildSystem {
...
+      self.indexStorePath = AbsolutePath(indexStorePath, relativeTo: self.projectRoot)
+    }
     self.buildServer = buildServer
     self.handler = handler
   }
 }
 
+private func readReponseDataKey(data: LSPAny?, key: String) -> String? {
+  switch data {
+  case .dictionary(let dataDict):
```
benlangmuir: You can simplify this code a bit with `if case` pattern matching.

```swift
if case .dictionary(let dataDict)? = data,
   case .string(let stringVal)? = dataDict[key]
{
  return stringVal
}

return nil
```

benlangmuir:
@swift-ci please test

Sources/SKCore/BuildServerBuildSystem.swift
rmaz: Was struggling with this one, knew there had to be a swiftier way to do this.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #155, opened 2019-09-10
Add SourceKitOptions build server request

rmaz:
Add a request and response for SourceKitOptions, and fetch synchronously as part of the `settings` API.

### DIFF
--- a/Sources/BuildServerProtocol/FileOptions.swift
+++ b/Sources/BuildServerProtocol/FileOptions.swift
@@ -1,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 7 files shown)


Sources/BuildServerProtocol/FileOptions.swift
```diff
@@ -0,0 +1,34 @@
...
+  public var uri: URL
+
+  public init(uri: URL) {
+    self.uri = uri
+  }
+}
+
+public struct SourceKitOptionsResult: ResponseType, Hashable {
+  /// The compiler options required for the requested file.
+  public var options: [String]
```
benlangmuir: "compilerArguments"?

benlangmuir:
@swift-ci please test

Sources/BuildServerProtocol/FileOptions.swift
rmaz: This was named after the `buildTarget/scalacOptions` request in the [BSP](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md) draft, but don't feel strongly abou tit.

benlangmuir: Fair enough, we can leave it.

PR #156, opened 2019-09-11
[index] Fix default value for listening to unit changes

benlangmuir:
Was preventing index from being populated by default.


benlangmuir:
@swift-ci please test

PR #157, opened 2019-09-13
Add support for file watching in the build server protocol

rmaz:
This PR adds requests for registering and unregistering file watching via the build server protocol. This will trigger a notification on changes that includes the updated build settings for a registered file. The same mechanism will be used for dependency changes via a separate notification.

### DIFF
--- a/Sources/BuildServerProtocol/RegisterForChangeNotifications.swift
+++ b/Sources/BuildServerProtocol/RegisterForChangeNotifications.swift
@@ -47,8 +47,5 @@ public struct FileOptionsChangedNotification: NotificationType {
   public var uri: URL
 
   /// The updated options for the registered file.
-  public var options: [String]
-
-  /// The working directory for the compile command.
-  public var workingDirectory: String?
+  public var updatedOptions: SourceKitOptionsResult
 }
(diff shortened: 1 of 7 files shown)


Sources/BuildServerProtocol/RegisterForChangeNotifications.swift
```diff
@@ -0,0 +1,54 @@
...
+/// build server to the language server when it detects
+/// changes to a registered files build settings.
+public struct FileOptionsChangedNotification: NotificationType {
+  public static let method: String = "build/sourceKitOptionsChanged"
+
+  /// The URL of the document that has changed settings.
+  public var uri: URL
+
+  /// The updated options for the registered file.
+  public var options: [String]
```
benlangmuir: Any reason not to use `SourceKitOptionsResult` instead of inlining these?

benlangmuir:
@swift-ci please test

Sources/BuildServerProtocol/RegisterForChangeNotifications.swift
rmaz: Nope, lets go with that instead

benlangmuir:
@swift-ci please test

PR #158, opened 2019-09-13
[test] Loosen test to account for clangd change in master-next

benlangmuir:
rdar://55353263


benlangmuir:
@swift-ci please test

PR #160, opened 2019-09-16
Remove the need to use `Connection` to talk to internal LSPs

DavidGoldman:
`SourceKitServer` now talks to the `SwiftLanguageServer` and
`ClangLanguageServerShim` directly through the `LanguageService`
protocol.

I have some TODOs left in the code to double check the threading
as I'm not quite sure how it should behave. Should each internal
LSP be responsible for handling the threading?

### DIFF
--- a/Sources/SourceKit/SourceKitServer.swift
+++ b/Sources/SourceKit/SourceKitServer.swift
@@ -38,7 +38,7 @@ public final class SourceKitServer: LanguageServer {
 
   let toolchainRegistry: ToolchainRegistry
 
-  var languageService: [LanguageServiceKey: Connection] = [:]
+  var languageService: [LanguageServiceKey: ToolchainLanguageServer] = [:]
 
   public var workspace: Workspace?
 
@@ -69,21 +69,39 @@ public final class SourceKitServer: LanguageServer {
     registerWorkspaceNotfication(SourceKitServer.changeDocument)
     registerWorkspaceNotfication(SourceKitServer.willSaveDocument)
     registerWorkspaceNotfication(SourceKitServer.didSaveDocument)
-    registerWorkspaceRequest(SourceKitServer.completion)
-    registerWorkspaceRequest(SourceKitServer.hover)
+
     registerWorkspaceRequest(SourceKitServer.workspaceSymbols)
-    registerWorkspaceRequest(SourceKitServer.definition)
-    registerWorkspaceRequest(SourceKitServer.implementation)
     registerWorkspaceRequest(SourceKitServer.references)
-    registerWorkspaceRequest(SourceKitServer.documentSymbolHighlight)
-    registerWorkspaceRequest(SourceKitServer.foldingRange)
-    registerWorkspaceRequest(SourceKitServer.symbolInfo)
-    registerWorkspaceRequest(SourceKitServer.documentSymbol)
-    registerWorkspaceRequest(SourceKitServer.documentColor)
-    registerWorkspaceRequest(SourceKitServer.colorPresentation)
-    registerWorkspaceRequest(SourceKitServer.codeAction)
     registerWorkspaceRequest(SourceKitServer.pollIndex)
     registerWorkspaceRequest(SourceKitServer.executeCommand)
+
+    registerToolchainTextDocumentRequest(SourceKitServer.completion,
+                                         CompletionList(isIncomplete: false, items: []))
+    registerToolchainTextDocumentRequest(SourceKitServer.hover, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.definition, [])
+    registerToolchainTextDocumentRequest(SourceKitServer.implementation, [])
+    registerToolchainTextDocumentRequest(SourceKitServer.symbolInfo, [])
+    registerToolchainTextDocumentRequest(SourceKitServer.documentSymbolHighlight, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.foldingRange, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.documentSymbol, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.documentColor, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.colorPresentation, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.codeAction, nil)
+  }
+
+  func registerToolchainTextDocumentRequest<PositionRequest: TextDocumentRequest>(
+    _ requestHandler: @escaping (SourceKitServer) -> (Request<PositionRequest>, ToolchainLanguageServer) -> Void,
+    _ fallback: PositionRequest.Response
+  ) {
+    _register { [unowned self] (req: Request<PositionRequest>) in
+      guard let workspace = self.workspace else {
+        return req.reply(.failure(.serverNotInitialized))
+      }
+      guard let languageService = workspace.documentService[req.params.textDocument.url] else {
+        return req.reply(fallback)
+      }
+      requestHandler(self)(req, languageService)
+    }
   }
 
   func registerWorkspaceRequest<R>(
@@ -165,7 +183,7 @@ public final class SourceKitServer: LanguageServer {
     return nil
   }
 
-  func languageService(for toolchain: Toolchain, _ language: Language) -> Connection? {
+  func languageService(for toolchain: Toolchain, _ language: Language) -> ToolchainLanguageServer? {
     let key = LanguageServiceKey(toolchain: toolchain.identifier, language: language)
     if let service = languageService[key] {
       return service
@@ -178,7 +196,7 @@ public final class SourceKitServer: LanguageServer {
       }
 
       let pid = Int(ProcessInfo.processInfo.processIdentifier)
-      let resp = try service.sendSync(InitializeRequest(
+      let resp = try service.initializeSync(InitializeRequest(
         processId: pid,
         rootPath: nil,
         rootURL: (workspace?.rootPath).map { URL(fileURLWithPath: $0.pathString) },
@@ -192,14 +210,14 @@ public final class SourceKitServer: LanguageServer {
         fatalError("non-incremental update not implemented")
       }
 
-      service.send(InitializedNotification())
+      service.clientInitialized(InitializedNotification())
 
       languageService[key] = service
       return service
     }
   }
 
-  func languageService(for url: URL, _ language: Language, in workspace: Workspace) -> Connection? {
+  func languageService(for url: URL, _ language: Language, in workspace: Workspace) -> ToolchainLanguageServer? {
     if let service = workspace.documentService[url] {
       return service
     }
@@ -234,10 +252,7 @@ extension SourceKitServer: BuildSystemDelegate {
       log("Build settings changed for opened file \(url)")
       if let snapshot = documentManager.latestSnapshot(url),
          let service = languageService(for: url, snapshot.document.language, in: workspace) {
-        service.send(
-          DidChangeConfiguration(settings:
-            .documentUpdated(
-              DocumentUpdatedBuildSettings(url: url, language: snapshot.document.language))))
+        service.documentUpdatedBuildSettings(url, language: snapshot.document.language)
       }
     }
   }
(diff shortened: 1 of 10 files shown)


benlangmuir:
> Should each internal LSP be responsible for handling the threading?

I think this is the only way to handle it. While we currently always initiate the calls from the SourceKitServer's `queue`, the async callbacks from e.g. sourcekitd can still come in anytime, and some of those require access to shared state.  So I  think we still want them all to have their own queue, but now you would need to use that queue in each of these individual methods if you're not funneling through the code that used to do that.

It's unfortunate that we don't have better language support for this and will need boiler plate around the queue handling, but I don't really see an alternative if we don't have a single entry point to dispatch through anymore.

DavidGoldman:
Ok, believe I've fixed the threading issues by having all requests/responses in `SwiftLanguageServer` move to its own queue as well as handling callbacks from the secondary LSPs in `SourceKitServer`'s own queue. Let me know if there's any place where I should swap from strong -> weak references for the queue async blocks.

Sources/SourceKit/LanguageService.swift
```diff
@@ -0,0 +1,45 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+
+/// A `LanguageServer` that exists within the context of the current process.
+public protocol LanguageService: AnyObject {
```
benlangmuir: I've been thinking about exposing an API along these lines to make it more convenient to write code agains the LSP server directly, but I think for this particular change we're better off with a special-purpose API used for talking to the swift and clang language implementations. For example, I would argue `documentUpdatedBuildSettings` and probably `symbolInfo` are more relevant to the implementation.  So I suggest naming this something like `ToolchainLanguageServer`, and moving it into the `SourceKit` module.  We might in the future add back something like this, but I think it serves a somewhat different purpose than the API we want for talking to the toolchain server.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -338,54 +335,63 @@ extension SourceKitServer {
...
+    guard let service = workspace.documentService[req.params.textDocument.url] else {
+      req.reply(CompletionList(isIncomplete: false, items: []))
+      return
+    }
+    service.completion(req)
   }
 
   func hover(_ req: Request<HoverRequest>, workspace: Workspace) {
-    toolchainTextDocumentRequest(req, workspace: workspace, fallback: nil)
+    guard let service = workspace.documentService[req.params.textDocument.url] else {
```
benlangmuir: While `toolchainTextDocumentRequest` no longer makes sense, maybe we should pull the lookup into a method `languageService<PositionRequest: TextDocumentRequest>(for request: PositionRequest)`, since I could see us changing this behaviour and wanting to keep all of them in sync.

Sources/SourceKit/clangd/ClangLanguageServer.swift
```diff
@@ -92,31 +89,84 @@ final class ClangLanguageServerShim: LanguageServer {
...
+  public func willSaveDocument(_ note: WillSaveTextDocument) {
+
+  }
+
+  public func didSaveDocument(_ note: DidSaveTextDocument) {
+
+  }
+
+  public func documentUpdatedBuildSettings(_ url: URL, language: Language) {
+    self.updateDocumentSettings(url: url, language: language)
```
benlangmuir: Do we still need both methods?

Sources/SourceKit/LanguageService.swift
DavidGoldman: Renamed, but it's already in the SourceKit module

Sources/SourceKit/SourceKitServer.swift
DavidGoldman: Done, but what do you think instead of having a `registerToolchainTextDocumentRequest(handler, fallbackValue)` to register these?

Sources/SourceKit/clangd/ClangLanguageServer.swift
DavidGoldman: Nope, removed

Sources/SourceKit/SourceKitServer.swift
benlangmuir: Could you expand a bit on what you mean? I'm not sure what exactly you have in mind.

DavidGoldman: Something like

```
  func registerToolchainTextDocumentRequest<PositionRequest: TextDocumentRequest>(
    _ fallback: PositionRequest.Response,
    _ requestHandler: @escaping (SourceKitServer) -> (Request<PositionRequest>, ToolchainLanguageServer) -> Void
  ) {
    _register { [unowned self] (req: Request<PositionRequest>) in
      guard let workspace = self.workspace else {
        return req.reply(.failure(.serverNotInitialized))
      }
      guard let languageService = workspace.documentService[request.params.textDocument.url] else {
        return req.reply(fallback)
      }

      requestHandler(self)(req, languageService )
    }
  }
```

DavidGoldman: Implemented this

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -42,40 +47,16 @@ public final class SwiftLanguageServer: LanguageServer {
...
-    _register(SwiftLanguageServer.symbolInfo)
-    _register(SwiftLanguageServer.documentSymbol)
-    _register(SwiftLanguageServer.documentColor)
-    _register(SwiftLanguageServer.colorPresentation)
-    _register(SwiftLanguageServer.codeAction)
   }
 
-  func publishDiagnostics(
+  /// Should be called on self.queue.
+  func queue_publishDiagnostics(
```
benlangmuir: I'm not a big fan of this naming convention, because it is likely to get lost if the particular queue changes or gets renamed. I'd prefer to just use a leading underscore and leave the queue-bit for the documentation.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -215,62 +184,71 @@ extension SwiftLanguageServer {
...
     guard let snapshot = documentManager.open(note) else {
       // Already logged failure.
       return
     }
 
+    let url = snapshot.document.url
+
+    // Cache the `BuildSystem`'s settings interally.
+    let settings = buildSystem.settings(for: url, snapshot.document.language)
+    buildSettingsByFile[url] = settings
```
benlangmuir: Don't we need to be on `queue` to access this?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Done

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Currently I have this modified outside of the queue but I can move it onto the queue if you'd like

benlangmuir: I meant, doesn't it need that to be thread-safe?

DavidGoldman: We could rely on always being called on the `SourceKitServer` queue, but I guess for safety we should move it over to its own queue?

DavidGoldman: Swapped over all calls to its own queue

benlangmuir:
@swift-ci please test

Sources/SourceKit/clangd/ClangLanguageServer.swift
```diff
@@ -18,7 +18,7 @@ import LanguageServerProtocolJSONRPC
 import Foundation
 
 /// A thin wrapper over a connection to a clangd server providing build setting handling.
-final class ClangLanguageServerShim: LanguageServer {
+final class ClangLanguageServerShim: LanguageServer, ToolchainLanguageServer {
```
benlangmuir: What are we still using from `LanguageServer` here?  Does it make sense to inline it?  If we're still using the `MessageHandler` implementation I think it will just forward everything to the `SourceKitServer` anyway, right? In which case we can just set that as the message handler, maybe?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -196,109 +171,122 @@ extension SwiftLanguageServer {
...
-   case .clangd:
-     break
-   case .documentUpdated(let settings):
-    documentBuildSettingsUpdated(settings.url, language: settings.language)
-   case .unknown:
-     break
-   }
- }
+  public func documentUpdatedBuildSettings(_ url: URL, language: Language) {
+    documentBuildSettingsUpdated(url, language: language)
```
benlangmuir: Should we combine these now?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -196,109 +171,122 @@ extension SwiftLanguageServer {
...
+      self.buildSettingsByFile[url] = nil
+      self.currentDiagnostics[url] = nil
 
-    var lastResponse: SKResponseDictionary? = nil
+      _ = self.sourcekitd.sendSync(req)
+    }
+  }
+
+  public func changeDocument(_ note: DidChangeTextDocument) {
+    let req = SKRequestDictionary(sourcekitd: self.sourcekitd)
```
benlangmuir: The `edit()` call on the document manager will be called once for every edit (it's an array in LSP), but your modified code only works if it is called 0 or 1 times.  I think the original code was okay except that you need to `async` for the `publishDiagnostics` call now.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -352,72 +347,78 @@ extension SwiftLanguageServer {
...
 
-    // FIXME: SourceKit should probably cache this for us.
-    if let settings = self.buildSettingsByFile[snapshot.document.url] {
-      skreq[keys.compilerargs] = settings.compilerArguments
-    }
-
-    logAsync { _ in skreq.description }
-
-    let handle = sourcekitd.send(skreq) { [weak self] result in
+    queue.async { [weak self] in
```
benlangmuir: General comment: I see you're using `queue.async { [weak self]` everywhere - I don't think we need weak for these.  There is no retain cycle, since the closure itself will be released once it has been executed, which will release the reference to `self`.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -539,74 +540,77 @@ extension SwiftLanguageServer {
     skreq[keys.sourcetext] = snapshot.text
     skreq[keys.syntactic_only] = 1
 
-    let handle = sourcekitd.send(skreq) { [weak self] result in
+    queue.async { [weak self] in
```
benlangmuir: I think any requests that aren't touching state protected by the `queue` can just call `self.sourcekitd.send(..., self.queue)` without doing an extra `queue.async` first.  The requests like open/edit/close, completion and cursor info currently need to touch the build settings and other state, so they need the `queue.async`.  But this request (`documentSymbol`) and I think the others below it don't, with the exception of `documentSymbolHighlight`.

benlangmuir:
@swift-ci please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Hmm I think we'll need to `queue.async` for the `self.sourcekitd.sendSync(req).success`, what do you think?

Sources/SourceKit/clangd/ClangLanguageServer.swift
DavidGoldman: Yep it was just for the `MessageHandler` implementation, but that sounds right - I'll try that

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
benlangmuir: I think you can safely make this one `sync` instead of `async`, since we're also using the synchronous sourcekit API here.  That should simplify it.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Do note that the `async`s effectively made the code run the same way that it was before (with the LSP automatically `async`ing to the `queue`), but if you think they're not needed I can remove them.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Sure although that will now block the `SourceKitServer`'s queue (where it didn't before)

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Yep, removed

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Done

Sources/SourceKit/clangd/ClangLanguageServer.swift
DavidGoldman: Seems to work, swapped over

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
benlangmuir: Fair point. In that case it seems like we need to wrap this whole thing in `queue.async`, but move the sourcekitd request back into callback so it gets called the right number of times.

DavidGoldman: Swapped it back

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
Looks good, thanks!  For additional confidence, I ran the  tests a few dozen times with thread sanitizer enabled and didn't see any issues.

PR #161, opened 2019-09-18
[vscode] Run `npm audit fix`

ikesyo:


Editors/vscode/package-lock.json
```diff
@@ -610,13 +610,13 @@
         },
         "minimist": {
             "version": "0.0.8",
-            "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz",
```
benlangmuir: These changes from https -> http don't look desirable to me, and I don't get these if I run `npm audit fix` locally.

ikesyo: I'm not sure why that was changed to http 🤔 I didn't changed that by hand. Any I'll re-update.

ikesyo: @benlangmuir Just updated.

PR #162, opened 2019-09-25
Add workspace/buildTargets request

rmaz:
This PR adds the `workspace/buildTargets` request for the build server protocol, as well as a BuildSystem interface method to request them.


benlangmuir:
@swift-ci please test

Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -0,0 +1,119 @@
+import LanguageServerProtocol
```
benlangmuir: Copyright header missing.

Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -0,0 +1,119 @@
...
+
+public struct BuildTargetIdentifier: Codable, Hashable {
+  public var uri: URL
+
+  public init(uri: URL) {
+    self.uri = uri
+  }
+}
+
+public enum BuildTargetTag: String, Codable {
```
benlangmuir: This should use a `struct` with `RawRepresentable`, since it's an open set.  I.e.
```swift
public struct BuildTargetTag: Codable, RawRepresentable {
  public var rawValue: String
  public init(rawValue: String) {
    self.rawValue = rawValue
  }

  // Not sure if 'Self' actually works here in the current compiler, but you get the idea.
  public static let library: Self = Self(rawValue: "library")
  ...
}
```

Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -0,0 +1,119 @@
...
+  /// Target contains re-usable functionality for downstream targets. May have any
+   /// combination of capabilities. 
+   case Library = "library"
```
benlangmuir: Naming: these should be lower-case.

Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -0,0 +1,119 @@
...
+  /// Pre-defined tags are listed in `BuildTargetTag` but clients and servers
+  /// are free to define new tags for custom purposes.
+  public var tags: [BuildTargetTag]
+
+  /// The capabilities of this build target. 
+  public var capabilities: BuildTargetCapabilities
+
+  /// The set of languages that this target contains.
+  /// The ID string for each language is defined in the LSP. 
+  public var languageIds: [String]
```
benlangmuir: `Language` from LSP?

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -173,6 +173,15 @@ extension BuildServerBuildSystem: BuildSystem {
     return nil
   }
 
+  public func buildTargets(reply: @escaping ([BuildTarget]?) -> Void) {
```
benlangmuir: What's the reasoning for using optional instead of exposing the `Result` with an explicit error?

rmaz: The idea was that it is legitimate for a build system to not support the concept of targets, and so to return nil, but i'm fine with this being an error also.

Sources/BuildServerProtocol/BuildTargets.swift
rmaz: yes

benlangmuir:
@swift-ci please test

Tests/SKCoreTests/BuildServerBuildSystemTests.swift
```diff
@@ -65,6 +66,41 @@ final class BuildServerBuildSystemTests: XCTestCase {
...
+                                   capabilities: BuildTargetCapabilities(canCompile: true, canTest: false, canRun: false),
+                                   languageIds: [Language.objective_c, Language.swift],
+                                   dependencies: [BuildTargetIdentifier(uri: URL(string: "first_target")!)]),
+                       ])
+        expectation.fulfill()
+      case .failure(let error):
+        XCTFail(error.message)
+      }
+    })
+    XCTAssertEqual(XCTWaiter.wait(for: [expectation], timeout: 1), .completed)
```
benlangmuir: Since this is a positive test (ie. we expect it to not timeout), maybe make this a bit longer - when running in CI things can sometimes be heavily loaded.  For tests that expect to timeout like `testSwiftDocumentBuildSettingsChangedFalseAlarm`, a shorter value is fine.

rmaz: Went with 15s like the other tests

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #163, opened 2019-09-26
Add Local Refactoring

rockbruno:
Adds a refactoring provider to `textDocument/codeActions`. CodeActions can now return SemanticRefactorCommands that are processed and sent back to the client as a `textDocument/applyEdit` request.

Screen Shot 2019-09-26 at 10 25 55 AM

Things to note:
- The LSP doesn't support snippets for CodeActions, so unfortunately refactors with placeholders don't work as well as they could... But there are threads that indicate that this might be added to the protocol soon.
- The indentation of the refactors are funky because the formatting request isn't implemented yet.

### DIFF
--- a/Sources/LanguageServerProtocol/PositionRange.swift
+++ b/Sources/LanguageServerProtocol/PositionRange.swift
@@ -20,6 +20,30 @@ extension Range where Bound == Position {
   }
 }
 
+extension Range: LSPAnyCodable where Bound == Position {
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    guard case .dictionary(let start)? = dictionary[PositionRange.CodingKeys.lowerBound.stringValue],
+          case .int(let startLine) = start[Position.CodingKeys.line.stringValue],
+          case .int(let startutf16index) = start[Position.CodingKeys.utf16index.stringValue],
+          case .dictionary(let end)? = dictionary[PositionRange.CodingKeys.upperBound.stringValue],
+          case .int(let endLine) = end[Position.CodingKeys.line.stringValue],
+          case .int(let endutf16index) = end[Position.CodingKeys.utf16index.stringValue] else
+    {
+      return nil
+    }
+    let startPosition = Position(line: startLine, utf16index: startutf16index)
+    let endPosition = Position(line: endLine, utf16index: endutf16index)
+    self = startPosition..<endPosition
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    return .dictionary([
+      PositionRange.CodingKeys.lowerBound.stringValue: lowerBound.encodeToLSPAny(),
+      PositionRange.CodingKeys.upperBound.stringValue: upperBound.encodeToLSPAny()
+    ])
+  }
+}
+
 /// An LSP-compatible encoding for `Range<Position>`, for use with `CustomCodable`.
 public struct PositionRange: CustomCodableWrapper {
   public var wrappedValue: Range<Position>
--- a/Sources/LanguageServerProtocol/PositionRange.swift
+++ b/Sources/LanguageServerProtocol/PositionRange.swift
@@ -28,7 +28,7 @@ public struct PositionRange: CustomCodableWrapper {
     self.wrappedValue = wrappedValue
   }
 
-  private enum CodingKeys: String, CodingKey {
+  public enum CodingKeys: String, CodingKey {
     case lowerBound = "start"
     case upperBound = "end"
   }
--- a/Sources/LanguageServerProtocol/PositionRange.swift
+++ b/Sources/LanguageServerProtocol/PositionRange.swift
@@ -20,30 +20,6 @@ extension Range where Bound == Position {
   }
 }
 
-extension Range: LSPAnyCodable where Bound == Position {
-  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
-    guard case .dictionary(let start)? = dictionary[PositionRange.CodingKeys.lowerBound.stringValue],
-          case .int(let startLine) = start[Position.CodingKeys.line.stringValue],
-          case .int(let startutf16index) = start[Position.CodingKeys.utf16index.stringValue],
-          case .dictionary(let end)? = dictionary[PositionRange.CodingKeys.upperBound.stringValue],
-          case .int(let endLine) = end[Position.CodingKeys.line.stringValue],
-          case .int(let endutf16index) = end[Position.CodingKeys.utf16index.stringValue] else
-    {
-      return nil
-    }
-    let startPosition = Position(line: startLine, utf16index: startutf16index)
-    let endPosition = Position(line: endLine, utf16index: endutf16index)
-    self = startPosition..<endPosition
-  }
-
-  public func encodeToLSPAny() -> LSPAny {
-    return .dictionary([
-      PositionRange.CodingKeys.lowerBound.stringValue: lowerBound.encodeToLSPAny(),
-      PositionRange.CodingKeys.upperBound.stringValue: upperBound.encodeToLSPAny()
-    ])
-  }
-}
-
 /// An LSP-compatible encoding for `Range<Position>`, for use with `CustomCodable`.
 public struct PositionRange: CustomCodableWrapper {
   public var wrappedValue: Range<Position>
@@ -52,7 +28,7 @@ public struct PositionRange: CustomCodableWrapper {
     self.wrappedValue = wrappedValue
   }
 
-  public enum CodingKeys: String, CodingKey {
+  fileprivate enum CodingKeys: String, CodingKey {
     case lowerBound = "start"
     case upperBound = "end"
   }
@@ -68,5 +44,25 @@ public struct PositionRange: CustomCodableWrapper {
     var container = encoder.container(keyedBy: CodingKeys.self)
     try container.encode(wrappedValue.lowerBound, forKey: .lowerBound)
     try container.encode(wrappedValue.upperBound, forKey: .upperBound)
+  }
+}
+
+extension Range: LSPAnyCodable where Bound == Position {
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    guard case .dictionary(let start)? = dictionary[PositionRange.CodingKeys.lowerBound.stringValue],
+          let startPosition = Position(fromLSPDictionary: start),
+          case .dictionary(let end)? = dictionary[PositionRange.CodingKeys.upperBound.stringValue],
+          let endPosition = Position(fromLSPDictionary: end) else
+    {
+      return nil
+    }
+    self = startPosition..<endPosition
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    return .dictionary([
+      PositionRange.CodingKeys.lowerBound.stringValue: lowerBound.encodeToLSPAny(),
+      PositionRange.CodingKeys.upperBound.stringValue: upperBound.encodeToLSPAny()
+    ])
   }
 }
--- a/Sources/LanguageServerProtocol/WorkspaceEdit.swift
+++ b/Sources/LanguageServerProtocol/WorkspaceEdit.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 /// A workspace edit represents changes to many resources managed in the workspace.
-public struct WorkspaceEdit: Hashable, ResponseType, LSPAnyCodable {
+public struct WorkspaceEdit: Hashable, ResponseType {
 
   /// The edits to be applied to existing resources.
   public var changes: [URL: [TextEdit]]?
@@ -19,54 +19,26 @@ public struct WorkspaceEdit: Hashable, ResponseType, LSPAnyCodable {
   public init(changes: [URL: [TextEdit]]?) {
     self.changes = changes
   }
-
-  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
-    guard case .dictionary(let lspDict) = dictionary[CodingKeys.changes.stringValue] else {
-      return nil
-    }
-    var dictionary = [URL: [TextEdit]]()
-    for (key, value) in lspDict {
-      guard let url = URL(string: key) else {
-        return nil
-      }
-      guard case .array(let array) = value else {
-        return nil
-      }
-      var edits = [TextEdit]()
-      for case .dictionary(let editDict) in array {
-        guard let edit = TextEdit(fromLSPDictionary: editDict) else {
-          return nil
-        }
-        edits.append(edit)
-      }
-      dictionary[url] = edits
-    }
-    self.changes = dictionary
-  }
-
-  public func encodeToLSPAny() -> LSPAny {
-    guard let changes = changes else {
-      return nil
-    }
-    let values = changes.map {
-      ($0.key.absoluteString, LSPAny.array($0.value.map { $0.encodeToLSPAny() }))
-    }
-    let dictionary = Dictionary(uniqueKeysWithValues: values)
-    return .dictionary([
-      CodingKeys.changes.stringValue: .dictionary(dictionary)
-    ])
-  }
 }
 
 // Workaround for Codable not correctly encoding dictionaries whose keys aren't strings.
 extension WorkspaceEdit: Codable {
-  enum CodingKeys: String, CodingKey {
+  private enum CodingKeys: String, CodingKey {
     case changes
   }
 
   public init(from decoder: Decoder) throws {
     let container = try decoder.container(keyedBy: CodingKeys.self)
-    self.changes = try container.decode([URL: [TextEdit]].self, forKey: .changes)
+    let changesDict = try container.decode([String: [TextEdit]].self, forKey: .changes)
+    var changes = [URL: [TextEdit]]()
+    for change in changesDict {
+      guard let url = URL(string: change.key) else {
+        let error = "Changes key is not an URL."
+        throw DecodingError.dataCorruptedError(forKey: .changes, in: container, debugDescription: error)
+      }
+      changes[url] = change.value
+    }
+    self.changes = changes
   }
 
   public func encode(to encoder: Encoder) throws {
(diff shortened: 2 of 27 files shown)


Sources/LanguageServerProtocol/WorkspaceEdit.swift
```diff
@@ -14,9 +14,13 @@
 public struct WorkspaceEdit: Codable, Hashable, ResponseType {
 
   /// The edits to be applied to existing resources.
-  public var changes: [URL: [TextEdit]]?
+  public var changes: [String: [TextEdit]]?
 
   public init(changes: [URL: [TextEdit]]?) {
-    self.changes = changes
+    guard let changes = changes else {
```
rockbruno: I haven't fully investigated the reason, but using `URL` directly here is tripping the VSCode extension. It causes `changes` to be parsed as an array instead of a dictionary:

```
[Trace - 10:16:58 PM] Received request 'workspace/applyEdit - (1)'.
Params: {
    "label": "Extract Method",
    "edit": {
        "changes": [
            "file:///file.swift",
            [
                {
                    "range": {
                        "start": {
                            "line": 20,
                            "character": 0
                        },
```

vs (String):

```
[Trace - 10:19:36 PM] Received request 'workspace/applyEdit - (1)'.
Params: {
    "label": "Extract Method",
    "edit": {
        "changes": {
            "file:///file.swift": [
                {
                    "range": {
                        "start": {
                            "line": 20,
                            "character": 0
                        },
```

Gonna look further into it, but I made the key a string meanwhile.

Trzyipolkostkicukru: Relevant thread on swift forums: https://forums.swift.org/t/using-rawrepresentable-string-and-int-keys-for-codable-dictionaries/26899

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
-      case .success(let response):
-        if response?.applied == false {
-          let reason: String
-          if let failureReason = response?.failureReason {
-            reason = " reason: \(failureReason)"
-          } else {
-            reason = ""
-          }
-          log("client refused to apply edit for \(label)!\(reason)", level: .warning)
+      case .success(let response) where response?.applied == false:
```
benlangmuir: I didn't notice this before, but `ApplyEdit.Respone = ApplyEditResponse?` seems incorrect. The spec says it's non-optional.  If it fails, it should give us an error, not nil.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+        }
+      case .failure(let error):
+        log("semantic refactoring failed \(url): \(error)", level: .warning)
+        return req.reply(nil)
+      }
+    }
   }
 
-  func applyEdit(label: String, edit: WorkspaceEdit) {
+  func applyEdit(label: String, edit: WorkspaceEdit, completion: @escaping () -> Void) {
```
benlangmuir: Should the completion handler pass through any errors so that failing to apply the edit can fail the executeCommand request?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+      completion([])
+      return
+    }
+    guard let startOffset = snapshot.utf8Offset(of: params.range.lowerBound),
+          let endOffset = snapshot.utf8Offset(of: params.range.upperBound) else
+    {
+        completion([])
+        return
+    }
+    let skreq = SKRequestDictionary(sourcekitd: sourcekitd)
```
benlangmuir: This code should use `cursorInfo` (e.g. see how `hover` is implemented) and add any extra flags you need rather than duplicate the logic.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
     }
   }
 
+  func retrieveRefactorCodeActions(_ params: CodeActionRequest, completion: @escaping CodeActionProviderCompletion) {
+    guard let snapshot = documentManager.latestSnapshot(params.textDocument.url) else {
+      log("failed to find snapshot for url \(params.textDocument.url)")
+      completion([])
```
benlangmuir: Should return an error.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+    let length = endOffset - startOffset
+    skreq[keys.length] = length
+    skreq[keys.retrieve_refactor_actions] = 1
+    if let settings = buildSystem.settings(for: snapshot.document.url, snapshot.document.language) {
+      skreq[keys.compilerargs] = settings.compilerArguments
+    }
+
+    let handle = sourcekitd.send(skreq) { [weak self] result in
+      guard let self = self else {
+        completion([])
```
benlangmuir: This doesn't need `weak self` since sourcekitd will not store it after it calls it once (and it always calls it, even if the server crashes or the request is cancelled).

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+        return
+      }
+      guard let dict = result.success else {
+        log("failed to find refactor actions: \(result.failure!)")
+        completion([])
```
benlangmuir: Return the error.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+      }
+      var codeActions = [CodeAction]()
+      results.forEach { _, value in
+        if let name: String = value[self.keys.actionname],
+           let actionuid: sourcekitd_uid_t = value[self.keys.actionuid],
+           let ptr = self.sourcekitd.api.uid_get_string_ptr(actionuid)
+        {
+          let actionName = String(cString: ptr)
+          guard actionName != "source.refactoring.kind.rename.global" else {
+            //TODO: Global refactoring
```
benlangmuir: Formatting: space after `//`.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+            return true
+          }
+          let refactorCommand = SemanticRefactorCommand(title: name,
```
benlangmuir: Formatting:
```swift
let refactorCommand = SemanticRefactorCommand(
  title: name,
  actionString: actionName,
  positionRange: params.range.asRange,
  textDocument: params.textDocument)
```

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -925,28 +925,116 @@ extension SwiftLanguageServer {
...
+                                                        actionString: actionName,
+                                                        positionRange: params.range.asRange,
+                                                        textDocument: params.textDocument)
+          do {
+            let lspCommand = try refactorCommand.asCommand()
+            let codeAction = CodeAction(title: name, kind: .refactor, command: lspCommand)
+            codeActions.append(codeAction)
+          } catch {
+            log("Failed to convert SwiftCommand to Command type: \(error)")
```
benlangmuir: `, level: .error)`

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,131 @@
...
+    self.positionRange = positionRange
+    self.textDocument = textDocument
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    let textDocumentArgument = LSPAny.dictionary(
+      [TextDocumentIdentifier.CodingKeys.url.stringValue: .string(textDocument.url.absoluteString)]
+    )
+    let startRange = LSPAny.dictionary(
+      [Position.CodingKeys.line.stringValue: .int(positionRange.lowerBound.line),
```
benlangmuir: How about adding these as method on `Position` and `PositionRange`?

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,131 @@
...
+      return nil
+    }
+    return type.init(fromLSPDictionary: dictionary)
+  }
+}
+
+public struct SemanticRefactorCommand: SwiftCommand {
+
+  public static var identifier: String {
+    return "semantic.refactor.command"
```
benlangmuir: Any reason not to use a stored `let` instead of a computed `var`? it's going to be static either way.

Sources/SourceKit/sourcekitd/SwiftCommand.swift
```diff
@@ -0,0 +1,131 @@
...
+/// All commands from the Swift LSP should be listed here.
+public let builtinSwiftCommands: [String] = [
+  SemanticRefactorCommand.self
+].map { $0.identifier }
+
+/// A `Command` that should be executed by Swift's language server.
+public protocol SwiftCommand: Codable, Hashable {
+  init?(fromLSPDictionary: [String: LSPAny])
+
+  static var identifier: String { get }
```
benlangmuir: Style nitpick: Put the initializer after the last instance property (`title` here).

Sources/SourceKit/sourcekitd/SemanticRefactoring.swift
```diff
@@ -0,0 +1,157 @@
...
+  ///   - completion: Completion block to asynchronously receive the SemanticRefactoring data, or error.
+  func semanticRefactoring(
+    _ refactorCommand: SemanticRefactorCommand,
+    _ completion: @escaping (Result<SemanticRefactoring, SemanticRefactoringError>) -> Void)
+  {
+    let url = refactorCommand.textDocument.url
+    guard let snapshot = documentManager.latestSnapshot(url) else {
+      return completion(.failure(.unknownDocument(url)))
+    }
+    guard let startOffset = snapshot.utf8Offset(of: refactorCommand.positionRange.lowerBound),
```
benlangmuir: Seems like a good convenience API to add: `utf8OffsetRange(of: Range<Position>) -> Range<Int>?`.  You could put it in the extension on DocumentManager where we implement `utf8Offset`.

Sources/SourceKit/sourcekitd/SemanticRefactoring.swift
```diff
@@ -0,0 +1,157 @@
...
+          let endOffset = snapshot.utf8Offset(of: refactorCommand.positionRange.upperBound) else
+    {
+      return completion(.failure(.invalidRange(refactorCommand.positionRange)))
+    }
+    let skreq = SKRequestDictionary(sourcekitd: sourcekitd)
+    skreq[keys.request] = requests.semantic_refactoring
+    skreq[keys.name] = ""
```
benlangmuir: Worth a comment about what this is: 
// Preferred name for e.g. an extracted variable. Empty string means sourcekitd chooses a name automatically.

Sources/SourceKit/sourcekitd/SemanticRefactoring.swift
```diff
@@ -0,0 +1,157 @@
...
+    skreq[keys.sourcefile] = url.path
+    skreq[keys.line] = refactorCommand.positionRange.lowerBound.line + 1
+    skreq[keys.column] = refactorCommand.positionRange.lowerBound.utf16index + 1 // LSP is zero based, but this request is 1 based.
```
benlangmuir: This column number also needs to go through UTF-16 -> UTF-8 conversion.  The LineTable has a `utf16ColumnAt(line:utf8Column)` method, but here you need the opposite: `utf8ColumnAt(line:utf16Column)`.  The implementation should be similar, but with the conversion flipped.

Sources/SourceKit/sourcekitd/SemanticRefactoring.swift
```diff
@@ -0,0 +1,157 @@
...
+      }
+      edits.forEach { _, value in
+        if let startLine: Int = value[keys.line],
+           let startColumn: Int = value[keys.column],
+           let endLine: Int = value[keys.endline],
+           let endColumn: Int = value[keys.endcolumn],
+           let text: String = value[keys.text]
+        {
+          // The LSP is zero based, but semantic_refactoring is one based.
+          let startPosition = Position(line: startLine - 1, utf16index: startColumn - 1)
```
benlangmuir: These columns need to be converted from UTF-8 (used by sourcekitd) to UTF-16 (used by LSP).  You can use the snapshot's `positionOf(zeroBasedLine:utf8Column:)` to get a unicode-correct position.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
rockbruno: Returning an error would work now since we only have one provider for code actions, but for the future the reasoning was that I didn't wanted to fail the entire request if a single provider failed. Should we go for it anyway?

rockbruno: I added support for the providers to return an error but kept the code action request itself always succeeding. We can make it fail easily now if necessary.

Sources/SKTestSupport/TestJSONRPCConnection.swift
```diff
@@ -92,6 +92,7 @@ public final class TestClient: LanguageServerEndpoint {
   var oneShotNotificationHandlers: [((Any) -> Void)] = []
 
   public var allowUnexpectedNotification: Bool = false
+  public var allowUnexpectedRequest: Bool = false
```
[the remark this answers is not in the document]
rockbruno: I had this to prevent ApplyEdit from crashing, but I ended up replicating the notification behavior to requests as well to be able to retrieve the client's ApplyEdit response - let me know if you think something different works better!

rockbruno:
Here's what's new besides the fixes:
- ExecuteCommand now returns an LSPAny as expected, with WorkspaceEdit and its children being now encodable to it. I made this into a `LSPAnyCodable` protocol so we can create a standard to it.
- I extended the tibs' init so that we can get the codeActions's custom capabilities into the workspace for our tests.

Sources/SKSupport/LineTable.swift
```diff
@@ -172,15 +172,37 @@ extension LineTable {
...
+  /// Returns UTF8 column offset at UTF16 version of logical position.
+  ///
+  /// - parameter line: Line number (zero-based).
+  /// - parameter utf16Column: UTF-16 column offset (zero-based).
+  @inlinable
+  public func utf8ColumnAt(line: Int, utf16Column: Int) -> Int? {
+    return convertColumn(
+      line: line,
+      column: utf16Column,
+      indexFunction: content.utf16.index,
```
benlangmuir: Please use the more-specific syntax `content.utf16.index(_:offsetBy:)`.  Same for `distance(from:to:)`.

Sources/SKTestSupport/TestJSONRPCConnection.swift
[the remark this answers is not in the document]
benlangmuir: Where is it `allowUnexpectedRequest  == true`?

Sources/SourceKit/sourcekitd/CursorInfo.swift
```diff
@@ -33,10 +33,14 @@ struct CursorInfo {
   /// https://github.com/apple/swift/blob/master/bindings/xml/comment-xml-schema.rng
   var documentationXML: String?
 
-  init(_ symbolInfo: SymbolDetails, annotatedDeclaration: String?, documentationXML: String?) {
+  /// The refactor actions available at this position.
+  var refactorActions: SKResponseArray? = nil
```
benlangmuir: All the other data here we are pre-converting - e.g. the SymbolDetails.  What if we converted to an array of `[SemanticRefactorCommand]` before putting it in CursorInfo?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -892,63 +892,167 @@ extension SwiftLanguageServer {
...
+
+    let additionalCursorInfoParameters: ((SKRequestDictionary) -> Void) = { skreq in
+      skreq[self.keys.length] = offsetRange.count
+      skreq[self.keys.retrieve_refactor_actions] = 1
+    }
+
+    cursorInfo(
+      params.textDocument.url,
+      params.range.lowerBound,
+      customCursorOffset: offsetRange.lowerBound,
```
benlangmuir: Isn't this the same offset we will compute for the position we're already passing in?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -892,63 +892,167 @@ extension SwiftLanguageServer {
...
+      additionalParameters: additionalCursorInfoParameters)
+    { result in
+      guard let dict: CursorInfo = result.success ?? nil else {
```
benlangmuir: What is `?? nil` doing?

Tests/SourceKitTests/CodeActionTests.swift
```diff
@@ -149,4 +169,63 @@ final class CodeActionTests: XCTestCase {
...
+    XCTAssertEqual(result, .codeActions([]))
+  }
+
+  func testSemanticRefactorLocationCodeActionResult() throws {
+    guard let ws = try refactorTibsWorkspace() else { return }
+    let loc = ws.testLoc("sr:foo")
+    try ws.openDocument(loc.url, language: .swift)
+
+    let textDocument = TextDocumentIdentifier(loc.url)
+    let start = Position(line: 1, utf16index: 11)
```
benlangmuir: You can put a testLoc here as well.

Tests/SourceKitTests/CodeActionTests.swift
```diff
@@ -149,4 +169,63 @@ final class CodeActionTests: XCTestCase {
...
+    XCTAssertEqual(result, .codeActions([expectedCodeAction]))
+  }
+
+  func testSemanticRefactorRangeCodeActionResult() throws {
+    guard let ws = try refactorTibsWorkspace() else { return }
+    let loc = ws.testLoc("sr:foo")
+    try ws.openDocument(loc.url, language: .swift)
+
+    let textDocument = TextDocumentIdentifier(loc.url)
+    let start = Position(line: 1, utf16index: 2)
```
benlangmuir: testLoc?

Sources/LanguageServerProtocol/WorkspaceEdit.swift
```diff
@@ -11,12 +11,73 @@
...
+
+// Workaround for Codable not correctly encoding dictionaries whose keys aren't strings.
+extension WorkspaceEdit: Codable {
+  enum CodingKeys: String, CodingKey {
+    case changes
+  }
+
+  public init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    self.changes = try container.decode([URL: [TextEdit]].self, forKey: .changes)
```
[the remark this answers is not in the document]
rockbruno: It worked since URL encodes to a string, but it actually failed the `checkCoding` method for a similar reason as before so for safety I've now made it decode to the actual [String: [TextEdit]] type and proceed from there.

Sources/LanguageServerProtocol/PositionRange.swift
```diff
@@ -20,6 +20,30 @@ extension Range where Bound == Position {
   }
 }
 
+extension Range: LSPAnyCodable where Bound == Position {
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    guard case .dictionary(let start)? = dictionary[PositionRange.CodingKeys.lowerBound.stringValue],
```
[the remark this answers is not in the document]
rockbruno: Ah, I missed this one after rebasing.

Sources/SKTestSupport/TestJSONRPCConnection.swift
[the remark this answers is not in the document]
rockbruno: Ended up being a leftover from before, will remove it so it can be added when it's really needed.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
rockbruno: The `result.success` here is a double optional -- I followed the same code style from hover/symbolinfo but I agree it's confusing.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
rockbruno: I was covering the type mismatches, but it turns out that was really unnecessary. Whoops!

I've instead made CursorInfo take a range instead of a specific position -- should make the overall logic cleaner here.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #164, opened 2019-09-30
Add -Xclangd to allow forwarding options to clangd command-line

benlangmuir:
And factor the configuration that we thread through to the sourcekit server.


benlangmuir:
@swift-ci please test

PR #165, opened 2019-10-02
[test] Flip the default value of allowUnexpectedNotification

benlangmuir:
This is/was mostly useful for testing specific notification
interactions. Most new tests shouldn't need it, so remove the
boilerplate.


benlangmuir:
@swift-ci please test

PR #166, opened 2019-10-02
[test] Speed up negative expectation test by 4 seconds

benlangmuir:
Saves 4 seconds from test time, which is ~33% of the total warm test
time on my machine when testing in serial.

I verified a timeout of 1 second is still long enough to reliably check
the desired behaviour. Since this test is designed to timeout on
success, we want the shortest timeout possible that will check the
behaviour.


benlangmuir:
@swift-ci please test

PR #167, opened 2019-10-02
Move custom PositionRange encoding into a property wrapper

benlangmuir:
This lets us use the `Range<Position>` without manually converting back
and forth in the majority of cases. See the note on HoverResponse for
the behaviour with Optional that makes us have to provide a custom
Codable implementation in two places that used to be synthesized.

### DIFF
--- a/Sources/SKSupport/CustomCodable.swift
+++ b/Sources/SKSupport/CustomCodable.swift
@@ -1,0 +1,82 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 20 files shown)


benlangmuir:
@swift-ci please test

Sources/SKSupport/CustomCodable.swift
```diff
@@ -0,0 +1,82 @@
...
+/// }
+/// ```
+///
+/// * Note: Unfortunately this wrapper does not work perfectly with `Optional`.
+///   While you can add a conformance for it, it will `encodeNil` rather than
+///   using `encodeIfPresent` on the containing type, since the synthesized
+///   implementation only uses `encodeIfPresent` if the property itself is
+///   `Optional`.
+@propertyWrapper
+public struct CustomCodable<CustomCoder: CustomCodableWrapper> {
```
aciidb0mb3r: Nice!

PR #169, opened 2019-10-02
[diagnostics] Fix caching of diagnostics across files and reopens

benlangmuir:
We were sending back (and clearing) all cached diagnostics, not just the
ones for the current file.  Also, clear on document close.


benlangmuir:
@swift-ci please test

PR #170, opened 2019-10-04
Add buildTarget/sources and buildTarget/outputPaths

rmaz:
This PR adds support for the `buildTarget/sources` and `buildTarget/outputPaths` methods of the Build Server Protocol.


Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -135,3 +135,75 @@ public struct BuildTargetCapabilities: Codable, Hashable {
...
+  public init(targets: [BuildTargetIdentifier]) {
+    self.targets = targets
+  }
+}
+
+public struct BuildTargetSourcesResult: ResponseType, Hashable {
+  public var items: [SourcesItem]
+}
+
+public struct SourcesItem: Codable, Hashable {
```
benlangmuir: Normally I like matching the protocol type names, but having this almost match `SourceItem` really threw me off.  This name isn't serialized anywhere, so how about changing it to something like `BuildTargetSources` so our API is clearer?  I've made similar exceptions for type names in the LSP support when they are unclear or ambiguous with things already in Swift.

rmaz: Fine by me. I'll change `OutputsItem` to match.

benlangmuir: We talked about this: `BuildTargetSources` conflicts with the request.  I agreed to keep the current names.

benlangmuir:
This test failed on a bot: https://ci.swift.org/job/swift-PR-Linux/16488/
```
14:22:23 Test Case 'BuildServerBuildSystemTests.testBuildTargetSources' started at 2019-10-16 21:22:23.352
14:22:23 Exited with signal code 4
```

benlangmuir:
Hmm, the issue might be more general to these tests.  There was a similar failure in testServerInitialize:
https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-18_04/6877/console
```
13:46:23 Test Case 'BuildServerBuildSystemTests.testServerInitialize' started at 2019-10-16 20:46:23.780
13:46:24 Exited with signal code 4
```

benlangmuir:
I've been trying to reproduce locally. So far I haven't seen this, but I did see (once out of a couple dozen runs):
``
Traceback (most recent call last):
  File "/home/docker_user/src/lsp/sourcekit-lsp/Tests/INPUTS/BuildServerBuildSystemTests.testSettings/server.py", line 79, in <module>
    sys.stdout.flush()
IOError: [Errno 32] Broken pipe
```

rmaz:
Yes, that seems clear enough, can just catch write exceptions and exit 0 from the python code. I don't get why the logs show `Exited with signal code 4` though, any insights on this?

benlangmuir:
Signal code 4 is SIGILL.  This is what you get when a Swift process traps for `fatalError`, force unwraps `nil`, etc.

If you're interested, Swift uses the x86_64 `ud2` instruction which causes an "invalid opcode" exception.  This is a common way to abort a process on error.

rmaz:
I see, and by throwing instead we should get some clearer error output. I'll put up a PR to catch write errors in the test code and exit in that case instead.

I'm still not sure I follow why this pipe is getting closed though, shouldn't the closure in `BuildServerBuildSystem.deinit` hold on to the connection until receiving the response?

benlangmuir:
I think the process itself has exited.  The closure keeps a reference to the connection, but there is nothing waiting on the closure and preventing the process from exiting.  That may be fine - we don't really want to wait around.  As long as the python process also exits in a reasonable time, this should be fine.

PR #171, opened 2019-10-08
[tsan] Fix a few issues caught by TSan

benlangmuir:
Fixes several test issues when running with `--sanitize=thread`. There is at least one remaining issue inside the JSONRPCConnection that I need to fix.


benlangmuir:
@swift-ci please test

PR #173, opened 2019-10-09
Adds buildTarget/didChange notification to the build server protocol

rmaz:
This PR adds the `buildTarget/didChange` notification to the build server protocol, as well as a `BuildSystemDelegate` method that is called when it is received.

### DIFF
--- a/Sources/BuildServerProtocol/BuildTargets.swift
+++ b/Sources/BuildServerProtocol/BuildTargets.swift
@@ -226,12 +226,12 @@ public struct BuildTargetEvent: Codable, Hashable {
   public var target: BuildTargetIdentifier
 
   /// The kind of change for this build target.
-  public var kind: BuildTargetEventKind
+  public var kind: BuildTargetEventKind?
 
   /// Any additional metadata about what information changed.
   public var data: LSPAny?
 
-  public init(target: BuildTargetIdentifier, kind: BuildTargetEventKind, data: LSPAny?) {
+  public init(target: BuildTargetIdentifier, kind: BuildTargetEventKind?, data: LSPAny?) {
     self.target = target
     self.kind = kind
     self.data = data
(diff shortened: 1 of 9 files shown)


Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -207,3 +207,44 @@ public struct OutputsItem: Codable, Hashable {
...
+    self.changes = changes
+  }
+}
+
+public struct BuildTargetEvent: Codable, Hashable {
+  /// The identifier for the changed build target.
+  public var target: BuildTargetIdentifier
+
+  /// The kind of change for this build target.
+  public var kind: BuildTargetEventKind
```
benlangmuir: The spec says this should be optional (https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/specification.md#build-target-changed-notification), although I'm not sure what the semantics of `nil` are here. Any thoughts?

Tests/SKCoreTests/BuildServerBuildSystemTests.swift
```diff
@@ -155,19 +155,52 @@ final class BuildServerBuildSystemTests: XCTestCase {
     })
     XCTAssertEqual(XCTWaiter.wait(for: [expectation], timeout: 15), .completed)
   }
+
+  func testBuildTargetsChanged() {
+    let root = AbsolutePath(
+      inputsDirectory().appendingPathComponent(testDirectoryName, isDirectory: true).path)
+    let buildFolder = AbsolutePath(NSTemporaryDirectory())
+    let buildSystem = try? BuildServerBuildSystem(projectRoot: root, buildFolder: buildFolder)
```
benlangmuir: Prefer `try` so that we will see the error if this fails.

benlangmuir:
@swift-ci please test

Sources/BuildServerProtocol/BuildTargets.swift
rmaz: Yes, that is rather confusing. This value is nice as it allows for not invalidating the build target graph when a new target is created or deleted, you can just request the updated sources and outputs for the changed module. I suppose if not set the language server has to just refresh everything. I'll update this to be optional.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #176, opened 2019-10-11
Swap `BuildSystem` to remove build settings method

DavidGoldman:
Instead the `BuildSystem` should inform its delegate of changes to build settings via `fileBuildSettingsChanged`


DavidGoldman:
Going to re-open as I'm not sure what's going on with GitHub including so many commits

DavidGoldman:
Re-opened as #183

PR #177, opened 2019-10-14
don't add log handler until connection has started

rmaz:
This fixes a launch crash if the log level is set above the default and there are any log messages during toolchain scanning. This would cause an assert as the `clientConnection` would try to send a message before it had started.


benlangmuir:
@swift-ci please test

PR #178, opened 2019-10-15
Add `ToolchainLanguageServerCrashHandler` to handle crashes

DavidGoldman:
The handler will be responsible for handling the crash:
- Inform the user of the crash
- Log the crash
- Choosing to restart the `ToolchainLanguageServer` or
  disabling it

### DIFF
--- a/Sources/SourceKit/ToolchainLanguageServerCrashHandler.swift
+++ b/Sources/SourceKit/ToolchainLanguageServerCrashHandler.swift
@@ -1,0 +1,22 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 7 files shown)


benlangmuir:
@swift-ci please test

Sources/SourceKit/ToolchainLanguageServerCrashHandler.swift
```diff
@@ -0,0 +1,22 @@
...
+
+import Foundation
+
+public protocol ToolchainLanguageServerCrashHandler: AnyObject {
+  /// Called when the given `ToolchainLanguageServer` has crashed and needs to be reinitialized with information
+  /// such as the list of open documents.
+  ///
+  /// The handler may or may not chose to re-open previously open documents and/or
+  /// stop sending any further requests to the given `ToolchainLanguageServer`.
+  func handleCrash(_ languageService: ToolchainLanguageServer, _ debugInfo: String)
```
benlangmuir: What's the idea for the debug info parameter? Right now I see it's always a fixed string - where would we get more information to pass here?

DavidGoldman:
Closing since this is no longer a priority (may re-open later)

PR #179, opened 2019-10-15
Set index db path from build server response

rmaz:
Currently only the index store path is returned, but both are required.


benlangmuir:
Looks like this needs to be rebased before we can merge it.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #181, opened 2019-10-17
Improve BuildServerBuildSystemTests error handling

rmaz:
This changes the python test server to catch write errors and exit 0 in that case. It also adds a missing call to `connection.close()` in `BuildServerBuildSystem.deinit()` to close the open pipes on shutdown.


benlangmuir:
@swift-ci please test

PR #182, opened 2019-10-18
Fix hang/crash when a write failure occurs

benlangmuir:
This code was doing a sync dispatch to a queue it was already running
on. Instead, go through the same path as other close calls.  This would
hang, or if you're lucky crash when dispatch detects the issue.


benlangmuir:
@swift-ci please test

PR #183, opened 2019-10-25
Swap `BuildSystem` to remove build settings method

DavidGoldman:
Instead the `BuildSystem` should inform its delegate of changes to build settings via `fileBuildSettingsChanged`


benlangmuir:
@swift-ci please test

DavidGoldman:
Friendly ping

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -163,6 +169,11 @@ extension BuildServerBuildSystem: BuildSystem {
     return buildFolder?.appending(components: "index", "db")
   }
 
+  public func cachedSettings(for: URL, _ language: Language) -> FileBuildSettings? {
+    // TODO: Implement caching once the change notifications are supported.
```
DavidGoldman: Whoops forgot to remove this

Sources/SKCore/BuildSystem.swift
```diff
@@ -43,7 +40,10 @@ public protocol BuildSystem: AnyObject {
 
   /// Register the given file for build-system level change notifications, such
   /// as command line flag changes, dependency changes, etc.
-  func registerForChangeNotifications(for: URL)
+  ///
+  /// If the `BuildSystem` has cached build settings information, it is recommended to
+  /// immediately notify its delegate, as an optimization.
+  func registerForChangeNotifications(for: URL, language: Language)
```
benlangmuir: We should document that it's expected that there will be a notification with the initial settings if available.  The way it's currently worded makes it sound like an optimization, but this needs to always happen.  Also, something I hadn't considered: what do we do if there are no settings for this file? Maybe we could make the callback, but with `.removed`?

Sources/SKCore/BuildSystemList.swift
```diff
@@ -36,20 +36,11 @@ extension BuildSystemList: BuildSystem {
 
   public var indexDatabasePath: AbsolutePath? { return providers.first?.indexDatabasePath }
 
-  public func settings(for url: URL, _ language: Language) -> FileBuildSettings? {
```
benlangmuir: This is changing behaviour with `BuildSystemList` since the notifications are only implemented for the first one in the list.  We need a new way to inject the fallback settings.

Sources/SKCore/FileBuildSettingsChange.swift
```diff
@@ -0,0 +1,32 @@
...
+public enum FileBuildSettingsChange {
+
+  /// The `BuildSystem` no longer has `FileBuildSettings` for the file.
+  case removed
+
+  /// The `FileBuildSettings` have been modified.
+  case modified(FileBuildSettings)
+}
+
+public extension FileBuildSettingsChange {
```
benlangmuir: Style nitpick: put `public` on the individual members, not the extension.

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -135,16 +135,22 @@ final class BuildServerHandler: LanguageServerEndpoint {
...
+    self.delegate?.fileBuildSettingsChanged([params.uri: .modified(fileBuildSettings)])
   }
 }
 
 extension BuildServerBuildSystem: BuildSystem {
 
   /// Register the given file for build-system level change notifications, such as command
   /// line flag changes, dependency changes, etc.
-  public func registerForChangeNotifications(for url: LanguageServerProtocol.URL) {
+  public func registerForChangeNotifications(
```
benlangmuir: The build server still has a `settings` request; I think we need to make that request here to get the initial settings.

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -365,11 +365,16 @@ extension SourceKitServer {
...
+    let url = textDocument.url
+    let language = textDocument.language
 
-    if let service = languageService(for: textDocument.url, textDocument.language, in: workspace) {
-      service.openDocument(note.params)
-    }
+    // Create the `service` before registering for change notifications in case the `BuildSystem`
+    // immediately callbacks to us (its delegate) with `FileBuildSettings` so we notify this
+    // `service` (which might have been newly created).
+    let service = languageService(for: url, language, in: workspace)
```
benlangmuir: We still have a synchronous `BuildSystem.toolchain(for: URL, ...)` method, which is used behind the covers in `languageService(for: URL, ...)`, but ultimately the build system will need to be able to change the toolchain as well.  Your change here reminded me of this.  The current code cannot handle the toolchain changing.

I'm starting to think we may need to go back to providing the toolchain along with the build settings, since once you get called back by the build system you should know both the compiler arguments and the preferred toolchain at the same time.  You can always open the file with the default toolchain in the meantime to get syntactic support if necessary, but that's a tradeoff against potentially wasting time if we would get the callback quickly.  E.g. if you can get the settings in a couple of milliseconds, maybe it's better to just wait, especially since starting a new toolchain language server is (relatively) expensive.

Sources/SKCore/BuildSystem.swift
DavidGoldman: Yep, there's also a question of whether the initial settings must be immediately called, in which case, why not have this method return them?

Sources/SourceKit/SourceKitServer.swift
DavidGoldman: Yeah I think it should be combined with the build settings or perhaps we can mirror similar functionality to toolchain changes? I'm not sure about delaying the opening of the file, this is tough - if the build system may take awhile, the user gets nothing even if the toolchains are the same.

DavidGoldman:
As mentioned in https://forums.swift.org/t/sourcekit-lsp-file-status-ux/3594, planning on reviving this soon.

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SKCore/BuildServerBuildSystem.swift
DavidGoldman: marking resolved, outdated

Sources/SKCore/BuildSystem.swift
DavidGoldman: Documented this in the latest commit

Sources/SKCore/BuildSystemList.swift
DavidGoldman: This is outdated, BuildSystemList has been removed in a previous PR

Sources/SKCore/BuildServerBuildSystem.swift
DavidGoldman: Hmm I believe this is still necessary right?

Sources/SourceKit/SourceKitServer.swift
DavidGoldman: Outdated, this has been removed

Sources/SKCore/FileBuildSettingsChange.swift
DavidGoldman: done

Sources/SKCore/BuildServerBuildSystem.swift
DavidGoldman: @rmaz what do you think? should we change the protocol or manually fetch the settings here?

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -140,16 +154,37 @@ final class BuildServerHandler: LanguageServerEndpoint {
   }
 
   func handleFileOptionsChanged(_ notification: Notification<FileOptionsChangedNotification>) {
```
DavidGoldman, line 156: Need to double check the threading model here still

Sources/SKCore/BuildServerBuildSystem.swift
rmaz: The register request should be enough to get the settings, the initial settings should be returned as soon as they are available. I don't believe we need the `textDocument/sourceKitOptions` request any more.

DavidGoldman: Okay, left this as is, we can remove the `textDocument/sourceKitOptions` in a follow up if this is working well

Sources/SKCore/BuildServerBuildSystem.swift
DavidGoldman: Fixed

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SKCore/BuildServerBuildSystem.swift
benlangmuir: The way `FileOptionsChangedNotification` and registration is setup does not play nicely with files that are outside the build model.  Right now, if a file is not known to the build server, it's not clear if it will fail to register (in which case it will not get a notification if it is later added), or if it will register but not provide initial "nil" settings (since `FileOptionsChangedNotification` does not have a way to represent lack of settings).  For the same reason, a file that successfully registers will never have its settings removed.

In the previous model, we could handle this because the settings update would trigger re-querying the settings synchronously, which can fail.

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,86 +131,124 @@ extension BuildSystemManager: BuildSystem {
...
+    for (mainFile, status) in changes {
+      let watches = self.watchedFiles.filter { $1.mainFile == mainFile }
+      guard !watches.isEmpty else {
+        // Possible notification after the file was unregistered. Ignore.
+        continue
+      }
+      self.mainFileStatuses[mainFile] = status
+      if let change = status.buildSettingsChange {
+        for watch in watches {
+          changedWatchedFiles[watch.key] = change
```
benlangmuir, line 221: If the status changes but the settings are the same, we should avoid an unnecessary notification.  For example, going from waitingFallback to fallback with the same settings.

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -211,42 +309,57 @@ extension BuildSystemManager: BuildSystem {
...
+      guard !changedFiles.isEmpty else {
+        if let delegate = self._delegate {
+          self.notifyQueue.async {
+            delegate.filesDependenciesUpdated(changedFiles)
+          }
+        }
+        return
+      }
+
+      // Need to map the changed main files back into changed watch files.
```
benlangmuir, line 357: Ah, good catch.

Sources/SKCore/CompilationDatabaseBuildSystem.swift
```diff
@@ -59,22 +59,17 @@ extension CompilationDatabaseBuildSystem: BuildSystem {
...
-      language: language)
+  /// Register the given file for build-system level change notifications, such
+  /// as command line flag changes, dependency changes, etc.
+  ///
+  /// IMPORTANT: When first receiving a register request, the `BuildSystem`
+  /// MUST eventually inform its delegate of any initial settings for the given file
+  /// via the `fileBuildSettingsChanged` method.
+  public func registerForChangeNotifications(for uri: DocumentURI, language: Language) {
+    let settings = self._settings(for: uri, language)
+    self.delegate?.fileBuildSettingsChanged([uri: FileBuildSettingsChange(settings)])
```
benlangmuir, line 70: Should be done asynchronously - e.g. on a global dispatch queue.  Clients usually don't expect to be called synchronously like this.

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,86 +131,124 @@ extension BuildSystemManager: BuildSystem {
...
     get { queue.sync { _delegate } }
     set { queue.sync { _delegate = newValue } }
   }
 
+  /// Register the given file for build-system level change notifications, such
+  /// as command line flag changes, dependency changes, etc.
+  ///
+  /// IMPORTANT: When first receiving a register request, the `BuildSystem`
+  /// MUST eventually inform its delegate of any initial settings for the given file
+  /// via the `fileBuildSettingsChanged` method.
```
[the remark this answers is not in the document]
DavidGoldman: Done

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,86 +131,124 @@ extension BuildSystemManager: BuildSystem {
...
         let mainFiles = self.mainFilesProvider?.mainFilesContainingFile(uri)
-        let mainFile = chooseMainFile(for: uri, from: mainFiles ?? [])
+        mainFile = chooseMainFile(for: uri, from: mainFiles ?? [])
         self.watchedFiles[uri] = (mainFile, language)
-        settings = self.cachedOrRegisterForSettings(mainFile: mainFile, language: language)
       }
 
-      if let delegate = self._delegate {
+      let newStatus = self.handleFetch(for: mainFile, language: language)
+      self.mainFileStatuses[mainFile] = newStatus
```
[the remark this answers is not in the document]
DavidGoldman: Done

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,86 +131,124 @@ extension BuildSystemManager: BuildSystem {
...
+      return status
+    }
+    // This is a new `mainFile` that we need to handle. We need to fetch the
+    // build settings.
+    if let buildSystem = self.buildSystem {
+      buildSystem.registerForChangeNotifications(for: mainFile, language: language)
+
+      // Register the timeout if it's applicable.
+      if let fallback = self.fallbackBuildSystem {
+        self.queue.asyncAfter(deadline: DispatchTime.now() + self.fallbackSettingsTimeout) { [weak self] in
```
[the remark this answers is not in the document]
DavidGoldman: Done

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,86 +131,124 @@ extension BuildSystemManager: BuildSystem {
...
+  }
+
+  /// *Must be called on queue*. Handle the fallback timer firing for a given
+  /// `mainFile`. Since this doesn't occur immediately it's possible that the
+  /// `mainFile` is no longer referenced or is referenced by multiple watched
+  /// files.
+  func handleFallbackTimer(
+    for mainFile: DocumentURI,
+    language: Language,
+    _ fallback: FallbackBuildSystem) {
```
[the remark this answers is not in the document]
DavidGoldman: done

Sources/SKCore/BuildSystemManager.swift
DavidGoldman: Good call, added that but we need to be sure to emit one if we were waiting

Sources/SKCore/BuildSystemManager.swift
DavidGoldman: :)

Sources/SKCore/CompilationDatabaseBuildSystem.swift
DavidGoldman: Hmm at least at the moment this is ok for the LSP itself since the BuildSystemManager handles it just fine via its queues, I guess it's still worth doing this for tests?

Sources/SKCore/BuildServerBuildSystem.swift
DavidGoldman: If the server can tell that the file is outside, it should fail registration (and we handle that OK). Otherwise it should provide initial nil settings.

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SKCore/CompilationDatabaseBuildSystem.swift
benlangmuir: Yes, and for future proofing.

DavidGoldman:
@swift-ci please test

Sources/SKCore/BuildServerBuildSystem.swift
benlangmuir: What will happen in these two cases?

1. file that starts unknown to build system is added to the build system (e.g. after you edit a config file).  Do we ever get notified of the settings? I'm guessing not since registration failed.

2. file that is known by the build system is removed from the build system. How can we be notified that this happened when the update notification has no notion of nil/removed?

DavidGoldman: Ah, by `If the server can tell that the file is outside, it should fail registration (and we handle that OK). Otherwise it should provide initial nil settings.` I meant that the server should only fail registration if it knows that file can never be handled by the build system (e.g. for a monorepo this happens if it's in a different directory). Unless it can do that it should register and provide no settings.

re #2 hmm this we won't handle, @rmaz what do you think?

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -256,29 +357,68 @@ extension SourceKitServer: BuildSystemDelegate {
...
+        // Non-ready documents should be considered open even though we haven't
+        // opened it with the language service yet.
+        guard openDocuments.contains(uri) else { continue }
+        guard self.documentsReady.contains(uri) else {
+          // Case 1: initial settings for a given file. Now we can process our backlog.
+          log("Initial build settings received for opened file \(uri)")
+
+          guard let snapshot = documentManager.latestSnapshot(uri),
+                let service = self.languageService(for: uri, snapshot.document.language, in: workspace) else {
+            log("No snapshot or language service for opened file \(uri)", level: .error)
```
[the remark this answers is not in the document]
DavidGoldman: Hmm, interesting. This case shouldn't happen, if it does we have a file that's considered open which the build system knows about yet we don't have a language service for I guess it's possible if the LS crashes? Cancelling makes sense IMO but once we're in this state we won't get out of it, although only documentedits will queue up, maybe we should change the behavior of changeDocument to account for this? WDYT?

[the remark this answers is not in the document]
benlangmuir: We know it had a language service when we opened it (and indeed a reference to that service is captured in the operation), but it's possible that configuration has changed in the meantime - e.g. available toolchains have changed.  In fact, another failure mode here would be that the language service is available, but different than the one we had before.  We could store the language service explicitly with the queue so that we can be sure to get the same one here and never fail.

Sources/SKCore/CompilationDatabaseBuildSystem.swift
DavidGoldman: Done

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -24,6 +24,45 @@ import TSCUtility
 
 public typealias URL = Foundation.URL
 
+/// Simple struct for pending notifications/requests, including a cancellation handler.
+/// For convenience the notifications/request handlers are type erased via wrapping.
+private struct NotificationRequestOperation {
```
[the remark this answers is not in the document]
DavidGoldman: Done

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -91,21 +137,74 @@ public final class SourceKitServer: LanguageServer {
...
+
+      // If the document is ready, we can handle it right now.
+      guard !self.documentsReady.contains(doc) else {
+        requestHandler(self)(req, workspace, languageService)
+        return
+      }
+
+      // Not ready to handle it, we'll queue it and handle it later.
+      self.documentToPendingQueue[doc, default: DocumentNotificationRequestQueue()].add(operation: {
+        requestHandler(self)(req, workspace, languageService)
```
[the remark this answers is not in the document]
DavidGoldman: Done, there was one other case where self was captured

Sources/SourceKit/SourceKitServer.swift
```diff
@@ -428,54 +568,101 @@ extension SourceKitServer {
...
+    guard !documentsReady.contains(uri) else {
+      workspace.documentService[uri]?.closeDocument(note.params)
+      return
     }
+
+    // Clear any queued notifications via their cancellation handlers.
+    // No need to send the notification since it was never considered opened.
+    self.documentToPendingQueue[uri]?.cancelAll()
+    self.documentToPendingQueue[uri] = nil
+    self.documentsReady.insert(uri)
```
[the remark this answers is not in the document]
DavidGoldman: oof nice catch, should instead be a removal above

Sources/SourceKit/SourceKitServer.swift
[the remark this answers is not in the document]
DavidGoldman: From what I can tell we don't support toolchain for a file changing - in fact once we create the toolchain for a file we can't overwrite it, we always use the cached version.

I think only `openDocument` should use the `self.languageService(for: uri, snapshot.document.language, in: workspace)` call, but this method here should always use the cached version and never create a new one (in practice this is what happens at the moment).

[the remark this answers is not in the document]
DavidGoldman: And hmm storing it seems like a good idea, but if we never have a ToolchainLanguageServer for a file, how can we queue up the edit requests properly? We probably shouldn't right? The code here at the moment does...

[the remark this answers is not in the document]
benlangmuir: > From what I can tell we don't support toolchain for a file changing - in fact once we create the toolchain for a file we can't overwrite it, we always use the cached version.

Fair enough - we ought to support that someday, but it's fine for now.

Since we think this situation should never happen and it should always get the cached one, how about logging an error and cancelling all the requests?  That will make it obvious that something is going wrong if this ever changes.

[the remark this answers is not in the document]
DavidGoldman: Done

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SourceKit/SourceKitServer.swift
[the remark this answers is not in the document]
benlangmuir: This got marked as resolved by the code still has `self.documentsReady.insert(uri)`.  Did this change get missed/dropped?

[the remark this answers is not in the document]
DavidGoldman: You're looking at an older outdated version,  or maybe github is confused? https://github.com/DavidGoldman/sourcekit-lsp-1/blob/469cfa49eac542f6161f1096c18824342344d20e/Sources/SourceKit/SourceKitServer.swift#L584

PR #184, opened 2019-10-31
Project can be built using Python 3

tachoknight:
As part of adding sourcekit-lsp to Fedora's version of swift-lang, it was necessary to make some Python 3-specific changes as post Fedora 30, Python 3 is the default version. 

I understand this might not be applicable now when macOS and Ubuntu still have Python 2 available; whenever Python 3 is made the default this pull request will be ready. :)


benlangmuir:
@swift-ci please test

PR #185, opened 2019-11-01
Fix issue preventing DocumentSymbols from working with clangd

DavidGoldman:
- Clangd uses the HierarchicalDocumentSymbolSupport field of the
  documentSymbol capabilities, if we don't pass it through clangd
  will change its behavior potentially resulting in a response
  error (e.g. in Visual Studio Code)


benlangmuir:
@swift-ci please test

PR #188, opened 2019-11-12
Eliminate _IndirectConnection and replace it by closure

ahoppen:
This allows us to move the semaphore logic out of the message to the connection where it belongs.


ahoppen:
@swift-ci Please test

PR #189, opened 2019-11-12
Move LanguageServer to SKCore

ahoppen:
In an effort to clean up `LanguageServerProtocol`, move the `LanguageServer` type to `SKCore`.


ahoppen:
@swift-ci Please test

PR #191, opened 2019-11-13
Remove dependency of LanguageServerProtocol on SKSupport

ahoppen:
Apart from removing unnecessary imports, the following changes were
performed:
- Move `CancellationToken` from `SKSupport` to `LanguageServerProtocol`
- Move `CustomCodable` to its own module
  - `CustomCodable` belongs in LanguageServerProtocol but moving it there causes compilation to fail (rdar://57088846). As a workaround, move it to its own module.

PR #192, opened 2019-11-14
Reorganize the repository into a set of bare LSP modules and SourceKit modules

ahoppen:
## Note

This PR depends on #191 

-------

This PR prepares the repository to allow splitting the LSP modules off. These LSP modules will provide the ability to write custom LSP servers and clients in Swift. The `sourcekit-lsp` repository will build on top of this new package to provide an LSP implementation that creates a language server for Swift and C-based-languages.

In particular this PR performs the following changes:
- Move various types from `SKSupport` to new module `LSPSupport`
- Move `CustomCodable` to `LanguageServerProtocol`
  - This requires a workaround of directly initializing `_range` using the property wrapper because of a compiler bug that causes compilation to fail otherwise
- Audit `import`s for those that are actually needed
- Exhaustively declare all dependencies in `Package.swift`, not relying on transitive dependencies
  - This made it easier to find dependencies that should not exist and allowed easier moving of files if downstream package dependencies are broken. I think it’s good for review as well
- Remove dependencies that are not needed


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

benlangmuir:
Closing in favour of #195. It reuses most of the commits from this PR, but I stripped out some changes to keep the history cleaner.

PR #193, opened 2019-11-19
Add support for non-URL URIs

ahoppen:
## Note

This PR depends on #192 

---------

According to the LSP specification, arbitrary URIs can be used as document identifiers. Instead of internally assuming that all URIs are URLs, use a `DocumentURI` enum to represent URIs. These can either be file URLs or other URIs whose value as treated as an opaque string.

These non-URL URIs can for example be used to refer to unsaved files. VS Code for example uses `untitled:Document name` to refer to unsaved documents.

### Changes

Apart from a massive amount of renaming the interesting changes are being made in:
- `DocumentURI.swift` where the new type is introduced
- `LocalSwiftTests.swift` which contains a test case for editing a file with a non-URL URI.
- `FallbackBuildSystem.swift` which uses the URI as the file name, even if it is not a file URL. This seems to allow semantic functionality even for non-file URIs.
- The build system files `BuildServerBuildSystem.swift`, `CompilationDatabaseBuildSystem.swift`, `SwiftPMWorkspace.swift`, `ClangLanguageServer.swift` which cannot handle computing build settings for non-file URIs and simply return `nil`
- `ClangLangaugeServer.swift` which ignores an updated build settings for non-file URIs since it can't really forward them to clang
  - This is because clang's workspace uses a compilation database with file paths as strings
  - This can probably be reworked to support 
- `WorkspaceFolder.swift` where the logic for computing the workspace name has been slightly altered to account for non-file URIs.


ahoppen:
@swift-ci Please test

benlangmuir:
Closing in favour of https://github.com/apple/sourcekit-lsp/pull/195.  It reuses most of the commits from this PR, but I stripped out some changes to keep the history cleaner.

benlangmuir:
Sorry wrong PR.

ahoppen:
@swift-ci Please test

Package.swift
```diff
@@ -12,79 +12,164 @@ let package = Package(
...
+
+      .target(
+        name: "LSPHelpers",
+        dependencies: []
+      ),
+
+      .testTarget(
+        name: "LSPSupportTests",
+        dependencies: [
+          "LSPHelpers",
```
benlangmuir: This will need to get rebased on #195

Sources/BuildServerProtocol/BuildTargets.swift
```diff
@@ -10,6 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 import LanguageServerProtocol
+import struct Foundation.URL
+
+public typealias URL = Foundation.URL
```
benlangmuir: We don't need to actually implement the handling for `BuildServerBuildSystem` in this PR, but the BSP protocol layer should use URI as well. I have a patch for this already.

Sources/SKCore/BuildSystem.swift
```diff
@@ -13,6 +13,9 @@
 import BuildServerProtocol
 import LanguageServerProtocol
 import TSCBasic
+import struct Foundation.URL
+
+public typealias URL = Foundation.URL
```
benlangmuir: This is no longer necessary.  We can just `import struct Foundation.URL` in `CompilationDatabaseBuildSystem.swift`.  The rest of the module should be clean.

Sources/SKTestSupport/Extensions.swift
```diff
@@ -0,0 +1,26 @@
+//===----------------------------------------------------------------------===//
```
benlangmuir: Naming convention for this file:
`TextDocumentIdentifier+URI.swift`.

benlangmuir:
@swift-ci Please test

benlangmuir:
@swift-ci Please test

ahoppen:
Rebased and addressed the comments.

@swift-ci Please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test Linux

ahoppen:
@swift-ci please test Linux

benlangmuir:
@swift-ci please test Linux

PR #197, opened 2019-11-26
Add textEdit to code completion responses

ahoppen:
Compute the `textEdit` based on the `num_bytes_to_erase` from SourceKit. This allows us to support replacing `.` in Optional completion, e.g. in 

```swift
let a: Foo? = Foo(bar: 1)
a.ba|
```

complete to 

```swift
a?.bar
```


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -384,13 +384,35 @@ extension SwiftLanguageServer {
           }
           let isInsertTextSnippet = clientSupportsSnippets && text != insertText
 
+          let textEdit: TextEdit?
+          if let text = text {
+            let bytesToErase: Int = value[self.keys.num_bytes_to_erase] ?? 0
```
benlangmuir: Theoretically this needs to be converted from UTF-8 byte count to UTF-16 code unit count.  In practice I think we only currently delete ASCII so it might not be noticeable.  At the very least it deserves a FIXME.

Tests/SourceKitTests/SwiftCompletionTests.swift
```diff
@@ -221,4 +227,30 @@ final class SwiftCompletionTests: XCTestCase {
       position: Position(line: 1, utf16index: 0)))
     XCTAssertTrue(outOfRange2.isIncomplete)
   }
+
+  func testCompletionOptional() {
```
benlangmuir: We should check a case where the deletion range is >1.  E.g. this will replace "func " with "override func foo() {}"
```swift
class Base {
  func foo() {}
}
class C: Base {
  func  <<<HERE>>>
}
```

Similarly, you should check what happens when this case crosses lines.  If sourcekitd deletes before the start of the line we should at least not crash here.
```swift
class C: Base {
  func 
  <<<HERE>>>
}
```

ahoppen: Oh, good catch. I didn’t think of that case. I implemented the correct lookup logic using the snapshot now. This also fixes the UTF-8 to UTF-16 conversion issue. 

I expect this to be fine performance-wise since the extra conversions are only done when `utf8BytesToErase` is not 0 which I expect to happen fairly rarely and especially not on a huge amount of results.

ahoppen:
@swift-ci Please test

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -384,13 +384,42 @@ extension SwiftLanguageServer {
...
 
+          let textEdit: TextEdit?
+          if let text = text {
+            let utf8BytesToErase: Int = value[self.keys.num_bytes_to_erase] ?? 0
+
+            var textEditRangeStart = completionPos
+            if utf8BytesToErase != 0 {
+              if let completionPosOffset = snapshot.lineTable.utf8OffsetOf(line: completionPos.line, utf16Column: completionPos.utf16index),
+                let textEditRangeStartLoc = snapshot.lineTable.lineAndUTF16ColumnOf(utf8Offset: completionPosOffset - utf8BytesToErase) {
+                textEditRangeStart = Position(line: textEditRangeStartLoc.line, utf16index: textEditRangeStartLoc.utf16Column)
```
benlangmuir: The spec says this needs to be on the same line:
> *Note:* The range of the edit must be a single line range and it must contain the position at which completion has been requested.

It's probably safest to just ignore `num_bytes_to_erase` when it crosses a line boundary.  You'll end up with something invalid

```
func
override func blah() {}
```

but I think it's understandable to users that we're only fixing it up on one line.  I doubt it's worth adding an extension just for this case.

Tests/SourceKitTests/SwiftCompletionTests.swift
benlangmuir: The most common case is "." to "?.", which will apply to all the results except those that come from `Optional` itself.  If it's an issue though we can special case `num_bytes_to_erase == 1`, since that can only ever be ASCII (so 1 UTF-16 code unit), or an invalid range that would currently trigger your assertion failure I believe.

ahoppen:
Oh, I didn't remember that LSP only supports single-line replacements. I refactored the replacement code to
- ignore the replacement if it would cross a line boundary
- fast-track if only a single character needs to be replaced so we don't need to hit the line table in the common case

ahoppen:
@swift-ci Please tet

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Tests/SourceKitTests/SwiftCompletionTests.swift
```diff
@@ -221,4 +227,82 @@ final class SwiftCompletionTests: XCTestCase {
...
+    XCTAssertFalse(response.isIncomplete)
+    guard let item = response.items.first(where: { $0.label == "foo()" }) else {
+      XCTFail("No completion item with label 'foo()'")
+      return
+    }
+    XCTAssertEqual(item.filterText, "foo()")
+    XCTAssertEqual(item.textEdit, TextEdit(range: Position(line: 4, utf16index: 2)..<Position(line: 4, utf16index: 6), newText: "override func foo() {\n\n}"))
+  }
+
+  func testCompletionOverrideInNewLine() {
```
benlangmuir: Need to run `swift test --generate-linuxmain`

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -384,13 +384,68 @@ extension SwiftLanguageServer {
...
 
+          let textEdit: TextEdit?
+          if let text = text {
+            let utf8CodeUnitsToErase: Int = value[self.keys.num_bytes_to_erase] ?? 0
+
+            // Compute the TextEdit
+            if utf8CodeUnitsToErase == 0 {
+              // Nothing to delete. Fast path and avoid UTF-8/UTF-16 conversions
+              textEdit = TextEdit(range: completionPos..<req.params.position, newText: text)
+            } else if utf8CodeUnitsToErase == 1 {
```
benlangmuir: This code is getting pretty gnarly, could we factor this into its own function to compute the edit?  Also, can we rearrange it to

1. compute the start location for the range
2. fallback for the multiline case

so that the logic can be shared between the count == 1 case and the general case? I think that would make it clearer what the overall flow is.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -384,13 +384,68 @@ extension SwiftLanguageServer {
...
+              let deletionStartUtf16Offset = line.utf16.distance(from: line.startIndex, to: deletionStartStringIndex)
+
+              // Check if we are only deleting on one line. LSP does not support deleting over multiple lines.
+              if deletionStartUtf16Offset >= 0 {
+                // We are only deleting characters on the same line. Construct the corresponding text edit.
+                let textEditRangeStart = Position(line: completionPos.line, utf16index: deletionStartUtf16Offset)
+                textEdit = TextEdit(range: textEditRangeStart..<req.params.position, newText: text)
+              } else {
+                // We are supposed to delete characters over multiple lines. This is not supported by LSP.
+                // Fall back to not replacing anything.
```
benlangmuir: s/not replacing anything/ignoring num_bytes_to_erase/

It still replaces the identifier characters in the range `completionPos..<req.params.position`

ahoppen:
* Pulled out the `TextEdit` computation logic into its own function
* Rearranged the function to first compute the start range for the replacement and then construct the `TextEdit`
* Generated `LinuxMain.swift`

@swift-ci Please test

benlangmuir:
@swift-ci Please test

PR #200, opened 2019-11-29
Fix typo: JSONRPCConnection should be spelled with two 'n'

ahoppen:
`JSONRPCConnection` only had one "n" in the class name.


ahoppen:
@swift-ci Please test

PR #202, opened 2019-12-03
Update the struct definitions to match version 3.14 of the LSP spec

ahoppen:
I went through the spec, comparing the struct definitions to what is defined in https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/ and changed the definitions accordingly to match the spec.

### DIFF
--- a/Sources/LanguageServerProtocol/ClientCapabilities.swift
+++ b/Sources/LanguageServerProtocol/ClientCapabilities.swift
@@ -225,11 +225,23 @@ public struct TextDocumentClientCapabilities: Hashable, Codable {
 
     /// Capabilities specific to `SignatureInformation`.
     public struct SignatureInformation: Hashable, Codable {
-      /// Documentation formats supported by the client from most to least preferred.
-      public var signatureInformation: [MarkupKind]? = nil
+      public struct ParameterInformation: Hashable, Codable {
+        /// The client supports processing label offsets instead of a simple label string.
+        var labelOffsetSupport: Bool? = nil
 
-      public init(signatureInformation: [MarkupKind]? = nil) {
-        self.signatureInformation = signatureInformation
+        public init(labelOffsetSupport: Bool? = nil) {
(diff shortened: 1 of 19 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/Character+Codable.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Character: Codable {
```
benlangmuir: We shouldn't tie Swift's definition of a Character to serialization - for one thing, the definition of a Character changes over time because of changes to unicode.  It's also unlikely to ever match what other lsp implementations use to define a character.  The spec itself uses String.

ahoppen:
I did the following change to address the review feedback:
- I changed the response of `DefinitionRequest` and `ImplementationRequest` to parse `Location` as `[Location]`
- Similarly `CompletionRequest` parses `[CompletionItem]` as `CompletionList`
- I removed the extension on `Optional` since it’s no longer necessary with your changes
- I added tests for any custom `Codable` conformance

I also split a couple commits off this PR to keep it at a more manageable size. They now live in #206.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #204, opened 2019-12-04
Upgrade CustomCodable to handle optional values using encodeIfPresent

benlangmuir:
The LSP spec requires us to omit keys for nil values rather than using
the JSON `null` constant in most places. This change to CustomCodable
allows us to do it automatically using CustomCodable, removing one of
the limitations of the property wrapper.

The idea for the adding overloads of `encode` and `decode` came from
https://forums.swift.org/t/pre-pitch-codable-customization-using-propertywrappers/30244/


benlangmuir:
@swift-ci please test

PR #205, opened 2019-12-05
build: add CMake build system

compnerd:


CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+if(CMAKE_VERSION VERSION_LESS 3.16 AND CMAKE_SYSTEM_NAME STREQUAL Windows)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+else()
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+endif()
+
+find_package(dispatch QUIET)
+find_package(Foundation QUIET)
+find_package(TSC QUIET)
```
benlangmuir: Is this provided by SwiftPM or is it actually going to use the TSC repo?

Sources/SourceKit/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  Csourcekitd
+  BuildServerProtocol
+  IndexStoreDB
+  LanguageServerProtocol
+  LanguageServerProtocolJSONRPC
+  SKCore
+  SKSwiftPMWorkspace
+  TSCUtility)
+
+if(BUILD_SHARED_LIBS)
```
benlangmuir: Do we need all these shared libs? The main product is the executable. We added a library product for the LanguageServerProtocol module, but not for the others.

CMakeLists.txt
compnerd: This is to use TSC from the tools-support-core repository.  All of the CI that I have going is trying to immediately use the tools-support-core version.

Sources/SourceKit/CMakeLists.txt
compnerd: Yes, sadly we do need the shared libraries because on Windows, we do not support static libraries.

CMakeLists.txt
benlangmuir: This will block us testing cmake on Linux on ci.swift.org until TSC itself has PR testing that includes sourcekit-lsp.

Sources/SourceKit/CMakeLists.txt
benlangmuir: Can't we just combine everything  - or at least everything outside of LangaugeServerProtocol module - into the single executable?

compnerd: As long as we can do that in the s-p-m based build as well (basically collapse the directory structure), yes, that would be possible.

CMakeLists.txt
compnerd: This is not a hard requirement - it will fall back to the system provided TSC if one is available.

Sources/SourceKit/CMakeLists.txt
benlangmuir: Oh I get it now. I was thinking of what we did in indexstoredb, but that was C++ code. With Swift code we can’t combine into one target because we need to keep the module structure.

CMakeLists.txt
benlangmuir: The cases I am concerned about are:

1) We make an API change to the copy of TSC in SwiftPM that requires a change in SourceKit-LSP.  If the cmake build uses the actual TSC repo but the swiftpm build uses TSC from SwiftPM we will need to make a lock-step change to TSC and SwiftPM.  This is a new requirement, but maybe it's "fine"?  That's a question for SwiftPM devs.

2) We make a change to TSC itself that breaks SourceKit-LSP.  Without PR testing for TSC that includes SourceKit-LSP, this will not be caught before merging.

compnerd: Okay, so you want to encourage the use of TSC from SwiftPM?  That is fine - if that is the preferred solution, my understanding as per @aciidb0mb3r was that we want to migrate to tools-support-core.

benlangmuir: I want us to migrate to TSC too, but I want to do that at the same time SwiftPM itself does.  I don't actually know what is holding that up.

compnerd: Well, the thing is, there is already a flag which controls that, and it is possible to build s-p-m against t-s-sc.

benlangmuir: > already a flag which controls that

Ah neat, so if we set `-DUSE_VENDORED_TSC=1` it will do that already?  If so, that's good enough for me.  If we add build using cmake that runs on ci.swift.org it should use that option to match how we build with swiftpm.

benlangmuir:
@swift-ci please test

CMakeLists.txt
compnerd: Yeah, that would do it :)

PR #207, opened 2019-12-10
Add a version field to PublishDiagnostics, according to LSP 3.15

ahoppen:
LSP Spec 3.15 will add a document version field for `PublishDiagnostics`.

This PR cherry-picking that field into SourceKit-LSP.


ahoppen:
@swift-ci Please test

PR #216, opened 2020-01-27
Provide Fix-Its through codeActions and PublishDiagnostics

ahoppen:
Return the Fix-Its from the `codeActions` request. 

Also, `clangd` specifies an LSP extension through which Fix-Its can directly be attached to diagnostics. Implement that extension to also attach Swift Fix-Its in the same way.

rdar://56624104


ahoppen:
@swift-ci Please test

benlangmuir:
@swift-ci please test

PR #217, opened 2020-01-27
Add all language identifiers listed in the LSP spec

ahoppen:


ahoppen:
@swift-ci Please test

PR #218, opened 2020-01-28
Return a more easily readable string from the Language's description property

ahoppen:


ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/SupportTypes/Language.swift
```diff
@@ -37,10 +37,71 @@ public struct Language: RawRepresentable, Codable, Equatable, Hashable {
...
-extension Language: CustomStringConvertible {
-  public var description: String {
+extension Language: CustomStringConvertible, CustomDebugStringConvertible {
+  public var debugDescription: String {
     return rawValue
   }
+
+  public var description: String {
+    switch rawValue {
+    case "abap": return "ABAP"
```
benlangmuir: Why not switch on `self` (e.g. `case .abap`)?

ahoppen: Good question. Because I didn’t think about that. Changed it to switch on `self`.

ahoppen:
@swift-ci Please test

PR #221, opened 2020-02-02
[NFC] Update for diagnostic changes

theblixguy:
A warning diagnostic will be emitted (by https://github.com/apple/swift/pull/29576) when discarding a function's result and the result type is `Void`, since that is redundant. This patch updates a test where we expect no diagnostics to be triggered.


benlangmuir:
This change would make us pass the test only on master swift but fail on older compilers.  While we could do that if it's absolutely necessary, I prefer to maintain a larger window of compatibility.  Please take a look at https://github.com/apple/sourcekit-lsp/pull/222 where I've tried to avoid this issue.

theblixguy:
Oh, I originally thought about changing the test code, then I thought maybe I can just change the assertion instead. I didn’t know about the compatibility problem. Thank you for fixing it!

benlangmuir:
Merged #222, which passes before and after the diagnostic change.  Closing this one.

theblixguy:
Thanks!

PR #222, opened 2020-02-03
Avoid new warning about assignment of Void

benlangmuir:
Modify the test code to avoid the new warning so that we can keep passing the test with older compilers.


benlangmuir:
@swift-ci please test

benlangmuir:
https://github.com/apple/swift/pull/29576
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #223, opened 2020-02-03
Infer index store path from compilation database

benlangmuir:
Find the index store path by searching through the command-line
arguments, and if found, also provide a default database path next to
the index store. Also add command-line arguments so that either of these
can be overridden. We could also easily add these as initialization
options if an LSP client wanted to provide them in the future.


benlangmuir:
@swift-ci please test

PR #224, opened 2020-02-03
Speed up compilation database loading when using "command"

benlangmuir:
The JSON compilation database allows specifying arguments either as an array of string (`"arguments"`) or as a single shell-escaped string (`"command"`). When using "command", we were hitting very slow load times for large compilation databases. This commit speeds up splitting the shell-escaped command by not iterating the UTF-8 view instead of the graphemes.

In release builds I see ~10x speedup of the splitting perf test, and ~3-5x in a debug build. In a real-world test using the cmake-generated `compile_commands.json` for llvm+clang this sped up overall compilation database loading from 11 seconds to 1.2 seconds on my machine.


benlangmuir:
@swift-ci please test

PR #225, opened 2020-02-04
Make initial workspace load asynchronous

benlangmuir:
Some editors treat the initialize request as blocking, or otherwise time
out if it takes too long. We have been scraping that edge for swiftpm
projects (time taken to load the manifest and construct a build graph),
and large compilation databases (time taken to parse the json and then
split commands into separate arguments). In practice, a *debug build* of
sourcekit-lsp was failing to open a large compilation database fast
enough for sublime text's lsp plugin (3 second limit).

This change hides the latency for loading the build system and creating
the index by letting it continue beyond the initialize request, blocking
instead the next request. This is not a complete solution. In addition
to hiding latency in between requests, this also is based on editors
being more forgiving about slow requests beyond the first initialize.`

We should also continue to improve the performance of the initial load
of the build system and index. And we should also consider more
prinicipled ways to use asynchronicity, for example by not doing the
initial load on the server's message queue. The main challenge for that
is that we have some assumptions, particularly in tests, that the index
will be created already, and for the index to load we currently block on
the build system load in order to get the index directory.

### DIFF
--- a/Sources/SourceKit/SourceKitServer.swift
+++ b/Sources/SourceKit/SourceKitServer.swift
@@ -300,36 +300,41 @@ extension SourceKitServer {
       }
     }
 
-    if let uri = req.params.rootURI {
-      self.workspace = try? Workspace(
-        rootUri: uri,
-        clientCapabilities: req.params.capabilities,
-        toolchainRegistry: self.toolchainRegistry,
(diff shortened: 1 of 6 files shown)


benlangmuir:
@swift-ci please test

PR #226, opened 2020-02-04
[codeActions] Ignore "supportedKinds", which isn't really an opt-in list

benlangmuir:
The client guarantees that unsupported kinds will be handled, and in practice some clients, such as Sublime's LSP plugin, use `"codeActionKind":{"valueSet":[]}`, since they support all kinds anyway.

This behaviour was also clarified in a [spec issue report](https://github.com/Microsoft/language-server-protocol/issues/620) where it was confirmed the server can send any actions it wants. It seems preferable to ignore it.


benlangmuir:
@swift-ci please test

PR #228, opened 2020-02-05
[test] Fix test broken by collision of recent changes

benlangmuir:
Two interacting changes were accidentally tested independently instead
of together and this got missed.


benlangmuir:
@swift-ci please test

PR #231, opened 2020-02-06
[buildsystem] Fix race when settings are updated

benlangmuir:
Jump onto `queue` when receiving notifications from the build system,
which could be on any queue.


benlangmuir:
@swift-ci please test

PR #232, opened 2020-02-06
[buildsystem] Introduce BuildSystemManager that handles main-file lookup and caching

benlangmuir:
The build system manage integrates with the indexer to provide header to main file mappings, including keeping those mappings up-to-date as `#include`s are added and removed. This allows us to provide accurate compiler arguments for a header based on a real translation unit. When there are no main files found, we continue to fallback to asking the build system.

The build system manager also gives us a place to put a client-side setting cache, since multiple headers may map to a single main file.

rdar://56599582


benlangmuir:
https://github.com/apple/indexstore-db/pull/68

@swift-ci please test

akyrtzi:
Nice! 🎉

PR #233, opened 2020-02-06
Fix crash with virtual files introduced in c477d8103da8ceb8f517f3c313…

DavidGoldman:
…44f581f12e9fb5


DavidGoldman:
Closing in favor of #234

PR #234, opened 2020-02-06
[index]  Avoid querying main files for a non-file URL

benlangmuir:
IndexStoreDB expects a file path, so exit early if we have something else.


benlangmuir:
@swift-ci please test

PR #235, opened 2020-02-06
[test] Regenerate linuxmain after build-system-manager change

benlangmuir:
I had done an update early in the commit series, but forgot to do one at
the end.


benlangmuir:
@swift-ci please test

PR #236, opened 2020-02-06
[fixit] Add a descriptive title to quickfixes

benlangmuir:
While the diagnostic provides the broader context, there can be multiple
fixits and it is helpful to give some way to see what they will do. This
adds a brief "Insert ';'", "Replace 'let a' with '_'", "Remove 'blah'"
description as the title.


benlangmuir:
@swift-ci please test

PR #238, opened 2020-02-07
[fixit] Add support for fixits that come from notes, and multi-edit fixes

benlangmuir:
Extends the fixit support to also walk notes (related information in LSP terminology), and use the description of the note as the title of the quick-fix, which gives more semantic behaviour and matches how other editors/servers behave.

Also fixes handling of multi-edit fixits. We were treating arrays of fixits as if they were independent actions, but in reality we have at most one quick-fix per diagnostic, which is composed of multiple edits. This fixes cases like renaming a deprecated method where there are multiple edits that need to be combined.


benlangmuir:
@swift-ci please test

PR #241, opened 2020-02-10
Fix BuildSystemManager.filesDependenciesUpdated calling wrong delegate method

DavidGoldman:


benlangmuir:
@swift-ci please test

PR #243, opened 2020-02-12
Disable diagnostics for blacklisted document schemes

DavidGoldman:
- Otherwise can lead to confusing duplicated diagnostics in VSCode
  due to its usage of virtual documents for source control diffbases

- sourcekitd does not properly handle virtual files when the
  `-working-directory` flag is passed

### DIFF
--- a/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
+++ b/Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
@@ -37,6 +37,27 @@ fileprivate extension Range {
   }
 }
 
+/// Explicitly blacklisted `DocumentURI` schemes.
+fileprivate let excludedDocumentURISchemes: [String] = [
+  "git",
+  "hg",
+]
+
+/// Returns true if diagnostics should be emitted for the given document.
+///
+/// Some editors  (like Visual Studio Code) use non-file URLs to manage source control diff bases
+/// for the active document, which can lead to duplicate diagnostics in the Problems view.
+/// As a workaround we explicitly blacklist those URIs and don't emit diagnostics for them.
+///
+/// Additionally, as of Xcode 11.4, sourcekitd does not properly handle non-file URLs when
+/// the `-working-directory` argument is passed since it incorrectly applies it to the input
+/// argument but not the internal primary file, leading sourcekitd to believe that the input
+/// file is missing.
+fileprivate func diagnosticsEnabled(for document: DocumentURI) -> Bool {
+  guard let scheme = document.scheme else { return true }
+  return !excludedDocumentURISchemes.contains(scheme)
+}
+
 public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   /// The server's request queue, used to serialize requests and responses to `sourcekitd`.
@@ -79,6 +100,12 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     response: SKResponseDictionary,
     for snapshot: DocumentSnapshot)
   {
+    let documentUri = snapshot.document.uri
+    guard diagnosticsEnabled(for: documentUri) else {
+      log("Ignoring diagnostics for blacklisted file \(documentUri.pseudoPath)", level: .debug)
+      return
+    }
+
     let stageUID: sourcekitd_uid_t? = response[sourcekitd.keys.diagnostic_stage]
     let stage = stageUID.flatMap { DiagnosticStage($0, sourcekitd: sourcekitd) } ?? .sema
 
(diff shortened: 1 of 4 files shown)


Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -79,6 +79,20 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     response: SKResponseDictionary,
     for snapshot: DocumentSnapshot)
   {
+    // Disable diagnostics for virtual files. Some editors (like Visual Studio Code) use virtual
```
benlangmuir: s/virtual files/non-file URLs/ ? The terminology "virtual file" is highly overloaded and we support various kinds of virtual files.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -79,6 +79,20 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
...
+    // files to manage diff bases for the active document, which can lead to duplicate
+    // diagnostics in the Problems view.
+    //
+    // Additionally, as of Xcode 11.4, sourcekitd does not properly handle virtual files when
```
benlangmuir: Have we got a bug report for this, because we should figure out a real solution for this part.

DavidGoldman: Filed https://bugs.swift.org/browse/SR-12196

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Updated, changed to blacklisted

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -37,6 +37,27 @@ fileprivate extension Range {
   }
 }
 
+/// Explicitly blacklisted `DocumentURI.pseudoPath` prefixes. No diagnostics will be emitted
+/// for these documents.
+fileprivate let documentPathPrefixBlacklist = [
```
benlangmuir: Style: Non-local variables should always have a type annotation.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -37,6 +37,27 @@ fileprivate extension Range {
...
+  "git",
+  "hg",
+]
+
+/// Returns true if the document's URL is considered blacklisted. Some editors  (like Visual
+/// Studio Code) use non-file URLs to manage source control diff bases for the active document,
+/// which can lead to duplicate diagnostics in the Problems view. As a workaround we explicitly
+/// blacklist those URIs and don't emit diagnostics for them.
+fileprivate func isDocumentBlacklisted(_ document: DocumentURI) -> Bool {
+  let path = document.pseudoPath
```
benlangmuir: Instead of using string prefixes, how about we expose `scheme: String?` on DocumentURI from the underlying URL? Then we can consistently talk about this as being a list of excluded URI schemes.

Terminology-wise, would it make sense to  say `enableDiagnostics(for:)`, and `excludedDocumentURISchemes`?

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -79,6 +100,20 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     response: SKResponseDictionary,
     for snapshot: DocumentSnapshot)
   {
+    // Disable diagnostics for blacklisted URLs. Some editors (like Visual Studio Code)
```
benlangmuir: Let's fold all the documentation onto the doc-comment string for the function rather than duplicate it here.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Done

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
DavidGoldman: Thought about it, makes sense, done

DavidGoldman:
PTAL, added tests

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #245, opened 2020-02-15
[test] Disable testMuliEditFixitCodeActionNote

benlangmuir:
SourceKit-LSP appears to be behaving correctly, but the toolchain
sourcekitd is sometimes returning 0 warnings instead of 1 in this test
when run on Linux. Disabling while we investigate, since this does not
appear to be a new issue, nor an issue without sourcekit-lsp itself.

rdar://problem/59487935


benlangmuir:
@swift-ci please test

PR #247, opened 2020-02-18
Support for clangd's go-to-definition for #include/#import

DavidGoldman:
- By forwarding the request to clangd when it fails to give
  symbol information, we are able to use its built in
  go-to-definition support for headers (jump to header file)

### DIFF
--- a/Sources/SourceKit/SourceKitServer.swift
+++ b/Sources/SourceKit/SourceKitServer.swift
@@ -599,6 +599,8 @@ extension SourceKitServer {
     let index = self.workspace?.index
     let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
       guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
+        let handled = languageService.definition(req)
+        guard !handled else { return }
         if let error = result.failure {
           req.reply(.failure(error))
         } else {
--- a/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
@@ -12,6 +12,22 @@
 
 import Foundation
 
+/// Standardize the URL.
+///
+/// This normalizes escape sequences in file URLs, like `%40` --> `@`.
+/// But for non-file URLs, has no effect.
+fileprivate func standardize(url: URL) -> URL {
+  guard url.isFileURL else { return url }
+
+  // This has the side-effect of removing trailing slashes from file URLs
+  // on Linux, so we may need to add it back.
+  let standardized = url.standardizedFileURL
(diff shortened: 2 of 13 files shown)


Sources/SourceKit/SourceKitServer.swift
```diff
@@ -599,6 +599,8 @@ extension SourceKitServer {
     let index = self.workspace?.index
     let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
       guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
+        let handled = languageService.definition(req)
```
benlangmuir: If symbol info failed, how is definition going to work?  I would have expected we would do this call farther down if something goes wrong with the index lookup.  For example, we already handle that in Swift with the "fallbackLocation".  Since you're doing a separate request, maybe we should refactor this code to have a fallback handler closure that either uses the fallback location in sourcekitd or calls definition in clangd.  That would also let us avoid the boolean return type on definition, you could just have the sourcekitd implementation fatalError.

DavidGoldman: FWIW for this particular case `clangd` doesn't return any symbols, so in my mind the case is a bit different than sourcekitd's case.

A fallback handler could work, I'll try that

benlangmuir: Oh, what exactly is the case you're thinking of?  The case I was considering was opening a file but we have no index data.  The compiler still knows how to find the declaration in a header, so it's better than nothing even if it can't find the real definition which might be another .cpp.

DavidGoldman: By goto-header I mean CMD+click on header `#include` or `#import` in order to open the file, e.g.

Open header from include

DavidGoldman:
> This could use a test - a static tibs test workspace should work, just don't call `buildAndIndex` and it will have no index data.

Added a test, PTAL

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test Linux

benlangmuir:
@swift-ci please test

DavidGoldman:
Rebased onto master, please try running on CI again, I can't reproduce the failure in a swift docker container....

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

PR #248, opened 2020-02-19
[test] Make test URLs unique to avoid sourcekitd race

benlangmuir:
If a document is re-opened really quickly after a previous open or
modification, we might get a notification from the previous state of the
document when its AST finishes building. In normal operation, this is
benign, because when we make a request to get the new diagnostics and
semantic highlights, they are guaranteed to be for the new document. But
in testing, we don't generally anticipate these spurious notifications.
The ideal fix for this would be to shutdown and restart sourcekitd
server between tests. Currently there are bugs preventing that from
working in sourcekitd, so as a workaround, make sure all URLs in tests
are uniqued so they will not collide.

rdar://59488048


benlangmuir:
@swift-ci please test

PR #251, opened 2020-02-27
BuildSystemManager fileBuildSettingsChanged fixes for dropped notifications

benlangmuir:
Fix two bugs that caused dropped notifications
* If we registered a header, but not its main file then called `fileBuildSettingsChanged([])` (where empty means assume everything changed) it would not propagate the notification.
* If we unregistered the last user of a given main file and then received a `fileBuildSettingsChanged` that explicitly mentioned that main file we would drop all *other* settings changes from the same notification.


benlangmuir:
@swift-ci please test

PR #252, opened 2020-02-27
Make BuildSystemManager passthrough notifications fully `async

benlangmuir:
This avoids having a gotcha in the API where if a build system calls
these methods while it is being called back for settings or to register
for notifications it deadlocks (or crashes in libdispatch to prevent the
deadlock).


benlangmuir:
@swift-ci please test

DavidGoldman:
For more context, this was happening indirectly before when I had the build system call `fileBuildSettingsChanged ` and then `filesDependenciesUpdated` from its own queue.

`fileBuildSettingsChanged` would bounce to the `BuildSystemManager-queue` and then `sync` back to my build system's queue for build settings, but then `filesDependenciesUpdated` would `sync` on the `BuildSystemManager-queue` from my build system's queue.

PR #254, opened 2020-03-10
[JSONRPC] Do not exit until outstanding I/O has finished

benlangmuir:
Specifically, we care that all outstanding **writes** are finished before we call the close handler, because otherwise we may (a) send corrupted output during shutdown, or (b) drop notifications and replies sent during the shutdown process.  The former is an issue for clients that are not robust about parse failures, and the latter is an issue for reproducibility and robustness during testing/debugging - in particular, we have some integration tests that send data without waiting for individual replies and they need to finish outstanding replies before exiting.

rdar://60159448


benlangmuir:
@swift-ci please test

PR #255, opened 2020-03-17
Increase some low test timeouts

benlangmuir:
While these should be fast the majority of the time, we don't want to
fail in a loaded high variance CI environment.

This is a speculative fix for https://bugs.swift.org/browse/SR-12378
rdar://60552818


benlangmuir:
@swift-ci please test

benlangmuir:
Still reproduces with this change.

PR #257, opened 2020-03-23
[build-script] Remove hard-coded install prefix of /usr and pass it in instead

buttaface:
Because of a bug upstream in the main build-script, the install prefix set by the user wasn't being passed in to this repo on non-Darwin platforms and these files assumed it to be `/usr`. This pull removes that assumption and has the main build-script pass an install prefix appended to the toolchain path instead, apple/swift#30565.

Let me know if this will require changing the name of `SOURCEKIT_TOOLCHAIN_PATH`.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -40,10 +40,10 @@ def get_swiftpm_options(args):
     swiftpm_args += [
       # Dispatch headers
       '-Xcxx', '-I', '-Xcxx',
-      os.path.join(args.toolchain, 'usr', 'lib', 'swift'),
+      os.path.join(args.toolchain, 'lib', 'swift'),
       # For <Block.h>
       '-Xcxx', '-I', '-Xcxx',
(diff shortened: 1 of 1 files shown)


benlangmuir:
apple/swift#30565
@swift-ci please test

buttaface:
I think you'll need to test with both the main build-script pull and the indexstore-db pull?

Looks like this is [the one repo that checked both appending `/bin` and `/usr/bin` to the toolchain path passed in](https://github.com/apple/sourcekit-lsp/blob/775e198f77934d792176560655a11c2d42822ebb/Sources/SKCore/Toolchain.swift#L118), rather than simply assuming and appending `/usr` like all the other Swift repos I patched, so maybe no changes for `SOURCEKIT_TOOLCHAIN_PATH` are required?

benlangmuir:
apple/indexstore-db#75
apple/swift#30565
@swift-ci please test

PR #259, opened 2020-04-01
[gardening] Fix warning and missing test on Linux

benlangmuir:


benlangmuir:
@swift-ci please test

PR #260, opened 2020-04-02
Append working directory to swift compiler args when presented in build settings

ghvg1313:
Currently there's no explicit api for `sourcekit` to recognize swift's `working-directory` other than through compiler args
When build server contains such setting, inject the correct args in place.


benlangmuir:
Currently this is only applied to the BuildServerBuildSystem, but it probably also should be done in at least the CompilationDatabaseBuildSystem as well.  Maybe we should handle this in the SwiftLanguageServer code so that it applies to all sources of compiler arguments?  @akyrtzi any concerns about that approach?

rmaz:
I agree, the changes should be done in SwiftLanguageServer. I think longer term it would be ideal to remove the `buildSystem.settings()` function entirely, something like the approach in #183.

akyrtzi:
> Maybe we should handle this in the SwiftLanguageServer code so that it applies to all sources of compiler arguments? @akyrtzi any concerns about that approach?

Seems like the right approach.

ghvg1313:
@benlangmuir Thanks for the suggestion, now mutating args in SwiftLanguageServer itself
as @rmaz point out, this is more like a short term patch, would love to get this sorted out with https://github.com/apple/sourcekit-lsp/pull/183

Sources/SourceKit/sourcekitd/CursorInfo.swift
```diff
@@ -99,7 +99,7 @@ extension SwiftLanguageServer {
 
     // FIXME: SourceKit should probably cache this for us.
     if let settings = self.buildSettingsByFile[uri] {
-      skreq[keys.compilerargs] = settings.compilerArguments
+      skreq[keys.compilerargs] = settings.getSwiftCompilerArguments()
```
benlangmuir, line 102: Rather than do this every time, which is easy to forget, let's mutate the settings from the build system before we put them in `self.buildSettingsByFile`.  I think there are two places (one synchronous call to `settings` and then one in the document update settings method.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -1374,3 +1374,19 @@ extension sourcekitd_uid_t {
     }
   }
 }
+
+// FIXME: sourcekit only accepts `working-directry` through swift compiler args.
+// Injecting the arg here if workingDirectory presents in the settings,
+// but not in the actual compiler args.
+extension FileBuildSettings {
+  
+  func getSwiftCompilerArguments() -> [String] {
```
benlangmuir, line 1383: I suggest something like `swiftCompilerArgumentsWithWorkingDirectory`.  Also, is this really a "FIXME" in the comment above? What would you do differently here?  I would work that comment into the documentation comment for this API.

Tests/SKCoreTests/FileBuildSettingsTests.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@testable import SKCore
```
benlangmuir, line 13: We don't allow `@testable` imports, because it requires building in a way that's not compatible with how we want to distribute our binary.  Feel free to make the API public to test it, but please (1) add an underscore, and (2) add "public for testing" at the front of its documentation comment.

benlangmuir:
Ah I missed that comment - what particularly would #183 help with here?  It would mean we only need one place to accept arguments from the build system (the delegate method), but we'd still need to modify them to add the working directory, no?

Sources/SourceKit/sourcekitd/CursorInfo.swift
ghvg1313: Mutating it would break the equity check against the build server setting in `openDocument` in swift language server
If we think this is the only way to sync the build server working directory setting, maybe we should change `FileBuildSettings`'s init method directly?

ghvg1313:
@benlangmuir ping for review, changed logic to the file setting model now

Tests/SKCoreTests/FileBuildSettingsTests.swift
```diff
@@ -0,0 +1,28 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
```
benlangmuir, line 5: 2020

ghvg1313:
@swift-ci test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test Linux

benlangmuir:
> BuildServerBuildSystemTests.swift:44: error: -[SKCoreTests.BuildServerBuildSystemTests testSettings] : XCTAssertEqual failed: ("Optional(["-a", "-b", "-working-directory", "/path/to/some"])") is not equal to ("Optional(["-a", "-b"])")

ghvg1313:
@swift-ci test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
> CompilationDatabaseTests.swift:203: error: -[SKCoreTests.CompilationDatabaseTests testCompilationDatabaseBuildSystem] : XCTAssertEqual failed: ("Optional(["-swift-version", "4", "/a/a.swift", "-working-directory", "/a"])") is not equal to ("Optional(["-swift-version", "4", "/a/a.swift"])")

benlangmuir:
@swift-ci please test

ghvg1313:
@benlangmuir Safe to merge?

PR #262, opened 2020-04-10
[test] Workaround an issue with createDirectory on Linux

benlangmuir:
Apply the same workaround as we did in indexstore-db. We haven't seen
this particular instance fail, but it theoretically has the same issue
here.

rdar://61545973


benlangmuir:
https://github.com/apple/indexstore-db/pull/81
@swift-ci please test

PR #264, opened 2020-04-21
Prettify hover response

Trzyipolkostkicukru:
Makes symbol name the same size as the rest of text, and puts a dividing line between declaration and the rest of the documentation. It makes the output similar to other language servers (dart, typescript, C)


benlangmuir:
@swift-ci please test

PR #265, opened 2020-04-21
Add serverOptions extension configuration to allow setting flags for sourcekit-lsp

ernieturner:
I had the need to be able to pass the `-Xlinker` flag to `sourcekit-lsp` from my VSCode extension and noticed that this didn't seem to be a configuration option currently. This should support passing any of the flags for `sourcekit-lsp` via the extension.

### DIFF
--- a/Editors/vscode/package.json
+++ b/Editors/vscode/package.json
@@ -46,10 +46,10 @@
                     "default": "sourcekit-lsp",
                     "description": "The path of the sourcekit-lsp executable"
                 },
-                "sourcekit-lsp.serverOptions": {
+                "sourcekit-lsp.serverArguments": {
                     "type": "array",
                     "default": [],
-                    "description": "Options to pass to sourcekit-lsp. Option keys and values should be provided as separate entries in the array e.g. ['--log-level', 'debug']"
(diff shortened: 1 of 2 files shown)


Editors/vscode/package.json
```diff
@@ -46,6 +46,11 @@
                     "default": "sourcekit-lsp",
                     "description": "The path of the sourcekit-lsp executable"
                 },
+                "sourcekit-lsp.serverOptions": {
+                    "type": "array",
```
benlangmuir, line 50: We could add a schema for the array elements as well:
```
"items": {
    "type": "string"
},
```

ernieturner: Done, but just an FYI, that changes the settings UI so that now there's an 'Add Item' button so adding both an option and a value requires entering multiple "Items". Not a big deal, but hopefully that's obvious for others.

Screen Shot 2020-04-21 at 12 12 45 PM

ernieturner:
Changed from `serverOptions` to `serverArguments`

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

ernieturner:
@benlangmuir Thanks!

PR #268, opened 2020-04-22
[test] Disable --parallel testing temporarily

benlangmuir:
We have a couple of non-deterministic CI failures right now, and they
may be triggered by, or made more frequent by doing parallel testing.
Disbaling while I continue to investigate.

rdar://61837752
rdar://61965529


benlangmuir:
@swift-ci please test

PR #276, opened 2020-05-11
add fileEvents in vscode extension

satishbabariya:
`fileEvents` will enable us to get `workspace.didChangeWatchedFiles` notification from vscode. also, we can always add this behind one **Flag** to enable/disable file watching form client 

For now, only Swift files are added for  `fileEvents`. in future we can add more file extensions for c/c++/obj-c

```ts
synchronize: {
  fileEvents: [
    vscode.workspace.createFileSystemWatcher("**/*.swift"),
 // vscode.workspace.createFileSystemWatcher("**/*.c"),
  ],
}
```

We can always implement `sourcekit-lsp` side file watchers in the future. 
but for now, we can implement `workspace.didChangeWatchedFiles` for editors like vscode

### DIFF
--- a/Editors/vscode/src/extension.ts
+++ b/Editors/vscode/src/extension.ts
@@ -26,7 +26,11 @@ export function activate(context: vscode.ExtensionContext) {
             'objective-c',
             'objective-cpp'
         ],
-        synchronize: undefined
+        synchronize: {
+          fileEvents: [
+            vscode.workspace.createFileSystemWatcher("**/*.swift"),
+          ],
+        }
     };
 
     const client = new langclient.LanguageClient('sourcekit-lsp', 'SourceKit Language Server', serverOptions, clientOptions);


benlangmuir:
What use case do you have in mind? It might make more sense to take this change at the same time as something in sourcekit-lsp that actually uses it.

satishbabariya:
@benlangmuir, 

> What use case do you have in mind?

to provide `Background Indexing` we can invoke indexing or building on `'workspace/didChangeWatchedFiles' message params. only on file created and file deleted. here is the enum from docs for file events
```ts
/**
 * The file event type.
 */
enum FileChangeType {
	/**
	 * The file got created.
	 */
	Created = 1,
	/**
	 * The file got changed.
	 */
	Changed = 2,
	/**
	 * The file got deleted.
	 */
	Deleted = 3
}
```
ref: https://github.com/microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#didchangewatchedfiles-notification




> It might make more sense to take this change at the same time as something in sourcekit-lsp

I was thinking of adding counterpart of `fileEvents` in sourcekit-lsp. but for that I still have many gray areas in the understanding of the implementation of current sourcekit-lsp.

As we discussed previously I think about a year ago. that currently SwiftPM has no implementation for file watching.

so I was thinking that we can implement the indexing for the client that provide file events.

Though for Indexing you may have a detailed idea of what needs to be done here.

benlangmuir:
> to provide Background Indexing
> we can invoke indexing or building on `'workspace/didChangeWatchedFiles' message params. only on file created and file deleted. here is the enum from docs for file events

If you're looking into this area, I suggest having the build system drive the actual indexing work as much as  possible, so we're just triggering it not actually invoking the compiler ourselves.  It would probably be good to discuss the overall design on the forums as well for visibility when you feel like you have a plan.

For swiftpm specifically, before we get to the point of indexing, it might make sense to first implement the ability for us to update our compiler arguments when there are changes - added/removed/moved files, or changes to the Package.swift file.  We have APIs for this in the "BuildSystem" protocol, but we haven't implemented it for swiftpm yet.  That would be a really good use case for file watching and will probably overlap with what we need for indexing.

>> It might make more sense to take this change at the same time as something in sourcekit-lsp
>
> I was thinking of adding counterpart of fileEvents in sourcekit-lsp. but for that I still have many gray areas in the understanding of the implementation of current sourcekit-lsp.
>
> As we discussed previously I think about a year ago. that currently SwiftPM has no implementation for file watching.

Sorry I think I explained myself poorly here. What I meant was that it might be better to add the file watching to sourcekit-lsp when we have code that will *use* it.  Otherwise it's unused code that we can't test right now.

By the way, swiftpm/tools-support-core does have a generic file watching API now:

https://github.com/apple/swift-package-manager/blob/1444b46dbc13477f027959e4d59e520747fc8382/swift-tools-support-core/Sources/TSCUtility/FSWatch.swift

I think ideally swiftpm would also provide a specific API that lets us watch for the specific files that affect compiler arguments.  If we could do that in a way that would work with either the LSP file watching, or the builtin file watching, that would be great.

PR #278, opened 2020-05-19
Remove `BuildSystemList`, merge it into `BuildSystemManager`

DavidGoldman:
- As discussed [here](https://forums.swift.org/t/sourcekit-lsp-file-status-ux/35947) the `BuildSystemList` no longer makes sense. Instead we need a `BuildSystem` which wraps 2 build systems: primary
  and fallback.
- The `BuildSystemManager` already wraps the underlying BuildSystem, making it suitable for handling the primary + fallback build systems
- Also cleans up some code a bit by making sure we have a`Workspace`/non-optional `BuildSystem`


DavidGoldman:
@swift-ci please test

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -47,25 +54,25 @@ public final class BuildSystemManager {
 
   /// Create a BuildSystemManager that wraps the given build system. The new manager will modify the
   /// delegate of the underlying build system.
-  public init(buildSystem: BuildSystem, mainFilesProvider: MainFilesProvider?) {
-    precondition(buildSystem.delegate == nil)
+  public init(buildSystem: BuildSystem?, mainFilesProvider: MainFilesProvider?) {
```
benlangmuir, line 57: I hadn't thought about it before, but would it make sense for the fallback build system to also be optional? It seems useful to be able to turn it off, particularly for test code.  Otherwise it can hide bugs in tests.

benlangmuir: I'd also be okay merging this first and treating that as a separate improvement if you prefer; I don't think it really impacts the rest of the API).

DavidGoldman: Done

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

PR #280, opened 2020-06-01
Send `didSave` notifications to clangd if it supports it

DavidGoldman:
- Minor update to the LSP protocol for `didSave` and `colorProvider` to be more in line with the spec
- Only send color requests over to clangd if it supports it

Rationale for this change: clangd recently implemented didSave support, it is used to rebuild preambles when the user edits a header file referenced by an open file.

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
@@ -194,16 +194,21 @@ public struct TextDocumentSyncOptions: Codable, Hashable {
     public init(includeText: Bool = false) {
       self.includeText = includeText
     }
+
+    public init(from decoder: Decoder) throws {
+      let container = try decoder.container(keyedBy: CodingKeys.self)
+      self.includeText = try container.decodeIfPresent(Bool.self, forKey: .includeText) ?? false
+    }
   }
 
   /// Whether save notifications should be sent to the server.
-  public var save: SaveOptions?
+  public var save: ValueOrBool<SaveOptions>?
 
   public init(openClose: Bool? = true,
               change: TextDocumentSyncKind? = .incremental,
               willSave: Bool? = true,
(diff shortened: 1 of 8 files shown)


DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -194,16 +194,21 @@ public struct TextDocumentSyncOptions: Codable, Hashable {
     public init(includeText: Bool = false) {
       self.includeText = includeText
     }
+
+    public init(from decoder: Decoder) throws {
+      let container = try decoder.container(keyedBy: CodingKeys.self)
+      self.includeText = try container.decodeIfPresent(Bool.self, forKey: .includeText) ?? false
```
benlangmuir, line 200: Shouldn't `includeText` just be optional?

DavidGoldman: I could swap to that, but they're functionality equivalent, this just makes it easier to handle since the default is false

benlangmuir: Originally that is how I was implementing these structs too: making them more convenient, but I now lean toward making it possible to round-trip values without changing them.  This is valuable both for our interaction with clangd (we don't want to silently modify the results), and if anyone wants to use our LSP layer as a library for other languages in the future.

DavidGoldman: Ah I see, swapped over to optional

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
Friendly ping on the review

PR #281, opened 2020-06-03
Withhold diagnostics when fallback arguments are being used

DavidGoldman:
- Also rework the DocumentURI standardization since we need
  to make sure we use the same one when checking the Document
  for publish diagnostics for clangd

TODO: reconsider the standardization and clangd threading

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
@@ -12,25 +12,12 @@
 
 import Foundation
 
-/// Standardize the URL.
-///
-/// This normalizes escape sequences in file URLs, like `%40` --> `@`.
-/// But for non-file URLs, has no effect.
-fileprivate func standardize(url: URL) -> URL {
-  guard url.isFileURL else { return url }
-
-  // This has the side-effect of removing trailing slashes from file URLs
-  // on Linux, so we may need to add it back.
-  let standardized = url.standardizedFileURL
-  if url.absoluteString.hasSuffix("/") && !standardized.absoluteString.hasSuffix("/") {
-    return URL(fileURLWithPath: standardized.path, isDirectory: true)
-  }
-  return standardized
-}
-
 public struct DocumentURI: Codable, Hashable {
   /// The URL that store the URIs value
   private let storage: URL
+
+  /// Standardized URL.
+  private let standardizedStorage: URL
 
   public var fileURL: URL? {
     if storage.isFileURL {
@@ -69,16 +56,27 @@ public struct DocumentURI: Codable, Hashable {
     guard let url = URL(string: string) else {
       fatalError("Failed to construct DocumentURI from '\(string)'")
     }
-    self.init(standardize(url: url))
+    self.init(url)
   }
 
   public init(_ url: URL) {
     self.storage = url
+    self.standardizedStorage = url.standardizedFileURL
     assert(self.storage.scheme != nil, "Received invalid URI without a scheme '\(self.storage.absoluteString)'")
   }
 
   public init(from decoder: Decoder) throws {
     self.init(string: try decoder.singleValueContainer().decode(String.self))
+  }
+
+  /// Equality check to handle escape sequences in URLs, which we otherwise preserve.
+  public static func == (lhs: DocumentURI, rhs: DocumentURI) -> Bool {
+    return lhs.standardizedStorage == rhs.standardizedStorage
+  }
+
+  /// Use the standardizedFileURL like our equality comparison.
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(self.standardizedStorage)
   }
 
   public func encode(to encoder: Encoder) throws {
--- a/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
@@ -16,9 +16,6 @@ public struct DocumentURI: Codable, Hashable {
   /// The URL that store the URIs value
   private let storage: URL
 
-  /// Standardized URL.
-  private let standardizedStorage: URL
-
   public var fileURL: URL? {
     if storage.isFileURL {
       return storage
@@ -61,7 +58,6 @@ public struct DocumentURI: Codable, Hashable {
 
   public init(_ url: URL) {
     self.storage = url
-    self.standardizedStorage = url.standardizedFileURL
     assert(self.storage.scheme != nil, "Received invalid URI without a scheme '\(self.storage.absoluteString)'")
   }
 
@@ -69,14 +65,14 @@ public struct DocumentURI: Codable, Hashable {
     self.init(string: try decoder.singleValueContainer().decode(String.self))
   }
 
-  /// Equality check to handle escape sequences in URLs, which we otherwise preserve.
+  /// Equality check to handle escape sequences in file URLs.
   public static func == (lhs: DocumentURI, rhs: DocumentURI) -> Bool {
-    return lhs.standardizedStorage == rhs.standardizedStorage
+    return lhs.storage.scheme == rhs.storage.scheme && lhs.pseudoPath == rhs.pseudoPath
   }
 
-  /// Use the standardizedFileURL like our equality comparison.
   public func hash(into hasher: inout Hasher) {
-    hasher.combine(self.standardizedStorage)
+    hasher.combine(self.storage.scheme)
+    hasher.combine(self.pseudoPath)
   }
 
   public func encode(to encoder: Encoder) throws {
--- a/Sources/SKCore/FileBuildSettingsChange.swift
+++ b/Sources/SKCore/FileBuildSettingsChange.swift
@@ -16,8 +16,16 @@ public enum FileBuildSettingsChange {
   /// The `BuildSystem` has no `FileBuildSettings` for the file.
   case removedOrUnavailable
 
-  /// The `FileBuildSettings` have been modified.
+  /// The `FileBuildSettings` have been modified or are newly available.
   case modified(FileBuildSettings)
+
+  /// The `BuildSystem` is providing fallback arguments which may not be correct,
+  /// or potentially is missing some generated inputs (e.g. prebuilding required).
+  ///
+  /// This indicates that diagnostics should be withheld for the given file. In addition, going from
+  /// fallback --> modified with the exact same `FileBuildSettings` is supported and indicates that
+  /// the LSP should rebuild its AST for the given file and emit new diagnostics.
+  case fallback(FileBuildSettings)
 }
 
 extension FileBuildSettingsChange {
@@ -27,12 +35,26 @@ extension FileBuildSettingsChange {
       return nil
     case .modified(let settings):
       return settings
+    case .fallback(let settings):
+      return settings
     }
   }
 
-  public init(_ settings: FileBuildSettings?) {
+  /// Whether the change represents fallback arguments.
+  public var isFallback: Bool {
+    switch self {
+    case .removedOrUnavailable:
+      return false
+    case .modified(_):
+      return false
+    case .fallback(_):
+      return true
+    }
+  }
+
+  public init(_ settings: FileBuildSettings?, fallback: Bool = false) {
     if let settings = settings {
-      self = .modified(settings)
+      self = fallback ? .fallback(settings) : .modified(settings)
     } else {
       self = .removedOrUnavailable
     }
--- a/Sources/SKCore/FileBuildSettingsChange.swift
+++ b/Sources/SKCore/FileBuildSettingsChange.swift
@@ -20,8 +20,6 @@ public enum FileBuildSettingsChange {
   case modified(FileBuildSettings)
 
   /// The `BuildSystem` is providing fallback arguments which may not be correct.
-  ///
-  /// This indicates that diagnostics should be withheld for the given file.
   case fallback(FileBuildSettings)
 }
 
--- a/Sources/SKCore/FileBuildSettingsChange.swift
+++ b/Sources/SKCore/FileBuildSettingsChange.swift
@@ -19,12 +19,9 @@ public enum FileBuildSettingsChange {
   /// The `FileBuildSettings` have been modified or are newly available.
   case modified(FileBuildSettings)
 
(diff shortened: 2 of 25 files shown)


Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
```diff
@@ -69,18 +56,29 @@ public struct DocumentURI: Codable, Hashable {
...
     self.storage = url
+    self.standardizedStorage = url.standardizedFileURL
     assert(self.storage.scheme != nil, "Received invalid URI without a scheme '\(self.storage.absoluteString)'")
   }
 
   public init(from decoder: Decoder) throws {
     self.init(string: try decoder.singleValueContainer().decode(String.self))
   }
 
+  /// Equality check to handle escape sequences in URLs, which we otherwise preserve.
```
DavidGoldman, line 72: I tried using `self.storage.standardizedFileURL` instead and it didn't work, I think something there must be mutable. Also considering instead using `stringByRemovingPercentEncoding` from NSString

Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
```diff
@@ -12,26 +12,13 @@
...
-  }
-  return standardized
-}
-
 public struct DocumentURI: Codable, Hashable {
   /// The URL that store the URIs value
   private let storage: URL
 
+  /// Standardized URL.
+  private let standardizedStorage: URL
```
benlangmuir, line 20: This could probably be `lazy`, which avoids the potentially expensive standardieFileURL call if you never compare the URL.

Sources/SKCore/FileBuildSettingsChange.swift
```diff
@@ -16,8 +16,16 @@ public enum FileBuildSettingsChange {
   /// The `BuildSystem` has no `FileBuildSettings` for the file.
   case removedOrUnavailable
 
-  /// The `FileBuildSettings` have been modified.
+  /// The `FileBuildSettings` have been modified or are newly available.
   case modified(FileBuildSettings)
+
+  /// The `BuildSystem` is providing fallback arguments which may not be correct,
+  /// or potentially is missing some generated inputs (e.g. prebuilding required).
```
benlangmuir, line 23: > or potentially is missing some generated inputs (e.g. prebuilding required).

This is a state that is independent of the settings themselves, and I would not expect it to be represented as `fallback`.  It feels like this is trying to shoehorn a more general status into the settings.

Sources/SKCore/FileBuildSettingsChange.swift
```diff
@@ -16,8 +16,16 @@ public enum FileBuildSettingsChange {
...
+  ///
+  /// This indicates that diagnostics should be withheld for the given file. In addition, going from
```
benlangmuir, line 25: This is a policy decision that gets made in the code that uses the settings, not some kind of invariant of the build settings themselves.  I would remove this comment.

Sources/SKCore/FileBuildSettingsChange.swift
```diff
@@ -16,8 +16,16 @@ public enum FileBuildSettingsChange {
...
+  /// fallback --> modified with the exact same `FileBuildSettings` is supported and indicates that
+  /// the LSP should rebuild its AST for the given file and emit new diagnostics.
```
benlangmuir, line 27: Why would you need to rebuild anyway, isn't it just a question of no longer filtering?

Sources/SKTestSupport/SKTibsTestWorkspace.swift
```diff
@@ -215,5 +215,8 @@ extension TibsToolchain {
 }
 
 extension TestLocation {
-  public var docIdentifier: TextDocumentIdentifier { TextDocumentIdentifier(url) }
```
benlangmuir, line 218: Dead change?

Tests/SourceKitTests/BuildSystemTests.swift
```diff
@@ -227,6 +227,107 @@ final class BuildSystemTests: XCTestCase {
...
+    let text = """
+    #ifdef FOO
+    static void foo() {}
+    #endif
+
+    int main() {
+      foo();
+      return 0;
+    }
+    """
```
benlangmuir, line 245: Style: please indent the multi-line string one level.

Tests/SourceKitTests/BuildSystemTests.swift
```diff
@@ -227,6 +227,107 @@ final class BuildSystemTests: XCTestCase {
...
+    buildSystem.buildSettingsByFile[doc] = settings
+    let expectation = XCTestExpectation(description: "refresh due to fallback --> primary")
+    expectation.expectedFulfillmentCount = 2
+    sk.handleNextNotification { (note: Notification<PublishDiagnosticsNotification>) in
+      // Syntactic analysis - no expected errors here.
+      XCTAssertEqual(note.params.diagnostics.count, 0)
+      expectation.fulfill()
+    }
+    sk.appendOneShotNotificationHandler  { (note: Notification<PublishDiagnosticsNotification>) in
+      // Syntactic analysis - one error since `-DFOO` was not passed.
```
benlangmuir, line 319: s/Syntactic/Semantic

Tests/SourceKitTests/BuildSystemTests.swift
```diff
@@ -227,6 +227,107 @@ final class BuildSystemTests: XCTestCase {
...
+    let url = URL(fileURLWithPath: "/\(#function)/a.swift")
+    let doc = DocumentURI(url)
+    let settings = FallbackBuildSystem().settings(for: doc, .swift)!
+
+    let text = """
+    #if FOO
+    func foo() {}
+    #endif
+
+    foo()
```
benlangmuir, line 290: We should include a syntactic diagnostic to test the behaviour of filtering out semantic, not just that there are none.

Sources/SourceKit/sourcekitd/SwiftLanguageServer.swift
```diff
@@ -264,13 +281,16 @@ extension SwiftLanguageServer {
...
   }
 
-  public func documentUpdatedBuildSettings(_ uri: DocumentURI, settings: FileBuildSettings?) {
+  public func documentUpdatedBuildSettings(_ uri: DocumentURI, change: FileBuildSettingsChange) {
     self.queue.async {
-      let compileCommand = settings.map { SwiftCompileCommand($0) }
+      let compileCommand = SwiftCompileCommand(change: change)
       // Confirm that the compile commands actually changed, otherwise we don't need to do anything.
+      // This includes when the compiler arguments are the same but the command is no longer
+      // considered to be fallback.
```
benlangmuir, line 293: If we really do have the same settings, why rebuild the AST instead of just updating the filtering?

Sources/SourceKit/clangd/ClangLanguageServer.swift
```diff
@@ -236,10 +307,11 @@ func makeJSONRPCClangServer(
     "-compile_args_from=lsp",   // Provide compiler args programmatically.
     "-background-index=false",  // Disable clangd indexing, we use the build
     "-index=false",             // system index store instead.
+    "--log=verbose",
```
benlangmuir, line 310: Not by default :-)

Sources/SKCore/FileBuildSettingsChange.swift
DavidGoldman: Yeah I thought it made sense to combine them since a file status containing `generated inputs not available` would mean a very similar thing to the fallback state to the LSP instances themselves. If we separate it out into the status, that means that the status the build system reports will also impact the ASTs and force reloads.

With this approach here, the individual LSPs manage the file state, and we could maybe have the build system just report a textual status update, no AST impact. I'm trying to keep the future file status implementation in mind.

Sources/SKCore/FileBuildSettingsChange.swift
DavidGoldman: With the above case in mind it isn't, we explicitly want to support settings not changing but the fallback state changing - it can mean new inputs are available.

Sources/SKCore/FileBuildSettingsChange.swift
DavidGoldman: I think this would only happen in practice if the build system does it on purpose - from above - to note that generated inputs are now available. How would fallback --> primary happen with the same settings otherwise?

Maybe the documentUpdatedDependencies should handle this instead somehow? Unclear to me how that would work, let me know if you have ideas.

Sources/SourceKit/clangd/ClangLanguageServer.swift
DavidGoldman: removed

Tests/SourceKitTests/BuildSystemTests.swift
DavidGoldman: Done

Tests/SourceKitTests/BuildSystemTests.swift
DavidGoldman: Done although other tests in the file don't abide by the format

Tests/SourceKitTests/BuildSystemTests.swift
DavidGoldman: Done

Sources/SKTestSupport/SKTibsTestWorkspace.swift
DavidGoldman: removed

Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
DavidGoldman: I'll try this + `stringByRemovingPercentEncoding` to see if we can get this working without needing to do anything with the file system.

Sources/SKCore/FileBuildSettingsChange.swift
benlangmuir: If you are missing dependencies, you may *want* to see that reflected in your diagnostics as e.g. missing modules. If your editor doesn't support the file status extension, this may be the only indication of what has gone wrong, since otherwise you just silently lose semantic functionality.  If you later build and then start seeing errors that weren't there before, I think that may be confusing.  This is especially painful if you have  dependencies built and then they are removed (i.e. build is cleaned), so the state goes from showing diagnostics to hiding them.  Without the file status extension, that's very confusing.

For fallback settings, you are in a more or less permanent state (so it won't confusing jump between errors and no errors).

Sources/SKCore/FileBuildSettingsChange.swift
```diff
@@ -16,8 +16,13 @@ public enum FileBuildSettingsChange {
   /// The `BuildSystem` has no `FileBuildSettings` for the file.
   case removedOrUnavailable
 
-  /// The `FileBuildSettings` have been modified.
+  /// The `FileBuildSettings` have been modified or are newly available.
   case modified(FileBuildSettings)
+
+  /// The `BuildSystem` is providing fallback arguments which may not be correct.
+  ///
+  /// This indicates that diagnostics should be withheld for the given file.
```
benlangmuir, line 24: Please drop the second line - the policy (hide diagnostics) is not up to the build system, it's up to the language server.

Sources/SourceKitLSP/Swift/Diagnostic.swift
```diff
@@ -223,9 +223,17 @@ extension CachedDiagnostic {
...
+/// emitting syntactic diagnostics.
+func mergeDiagnostics(
+  old: [CachedDiagnostic],
+  new: [CachedDiagnostic],
+  stage: DiagnosticStage,
+  isFallback: Bool
+) -> [CachedDiagnostic] {
   if stage == .sema {
-    return new
+    return isFallback ? old.filter { $0.stage == .parse } : new
```
benlangmuir, line 236: Even if you're falling back, you want the *new* parse diagnostics, not the old ones (otherwise you just get stuck with the old diagnostics permanently).

Also, I think what we want here is to leave mergeDiagnostics alone so that we always cache the semantic diagnostics regardless of fallback, but then filter them before they get sent back in the response.

DavidGoldman: Ah, so there can be parsing stage diagnostics even when `stage == .sema`?  And OK, that SGTM, will do

DavidGoldman: I tried out your suggestion but it doesn't work in the following case:

old (was fallback): stores semantic and parse diagnostics (only emit parse diagnostics though)
new (no longer fallback): parse case. we keep the old semantic diagnostics around, but they were invalid and ignored, but we don't know that since we don't store that information, so we wrongly emit the old semantic diagnostics + new parse diagnostics.

DavidGoldman:
> Looking really close, just a couple of minor things commented inline. Also, please rebase this so we don't have a merge commit from master in here:
> 
> > Merge branch 'master' into withholddiagnostics
> 
> Thanks!

It looks like GitHub can squash and merge, that should work?

Sources/SKCore/FileBuildSettingsChange.swift
DavidGoldman: Done

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SourceKitLSP/Swift/Diagnostic.swift
benlangmuir: Ah good point. Okay, then I think the only thing left is we want to filter to the new parse diagnostics not the old ones.

DavidGoldman: I just checked that as well, I only see semantic diagnostics for the sema stage, so if I do that we lose all diagnostics. Thus we can need the current behavior.

benlangmuir: Ah I see.  So when in fallback, the flow is:
* parse diagnostics arrive, we will return new (parse) + old (sema, empty)
* sema diagnostics arrive, we will return old (parse), which should be identical to the last publishDiagnostics

Okay

DavidGoldman: Yep

PR #285, opened 2020-06-04
[clang] Workaround module format issue with -gmodules

benlangmuir:
Clangd is not built with support for the "obj" format of modules, used by e.g. `-gmodules`. Workaround by forcing the module format to "raw".

Incidentally, also workaround an issue with working directory for `-fmodules-cache-path` that was causing running the new test added here to create its module cache in the project sources instead of in the test's build directory.


benlangmuir:
@swift-ci please test

PR #286, opened 2020-06-04
Expose SourceKitLSP library product

benlangmuir:
It's not at all stable, so underscore it, but it may be useful for experimentation.


benlangmuir:
@swift-ci please test

PR #287, opened 2020-06-08
[test] Fix notification handler test when run in serial

benlangmuir:
The test was passing with --parallel since each test was being run in
its own process, but not when run in serial since other documents could
trigger unexpected notifications. Also make the uniqueness of the path
reliable using a mutable workspace.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -126,6 +126,7 @@ let package = Package(
         dependencies: [
           "SourceKitD",
           "SKCore",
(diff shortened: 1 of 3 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #289, opened 2020-06-16
Update BuildServerBuildSystem for async initialization

rmaz:
Having a synchronous request to initialize the build server will block the LSP thread for the duration of the build servers initialization, which could be potentially a long time. This PR changes the setup code to use an async request instead. This required adding the index store path and database path to the build server config file, as we need to know this directly after initialization so cannot wait for a response from the build server.


rmaz:
Causes more problems than it solves, going to stick with synchronous initialization and punt the server setup code to later instead.

PR #290, opened 2020-06-18
Add file status extension to the LSP module

DavidGoldman:
- For more information see https://forums.swift.org/t/sourcekit-lsp-file-status-ux/35947

### DIFF
--- a/Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
+++ b/Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
@@ -1,0 +1,48 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// **LSP Extension** Notification from the server about a per-file status update for
+/// the client.
+///
+/// Requires a private SourceKit extension.
+public struct SourceKitFileStatusNotification: TextDocumentNotification, Hashable {
+  public static let method: String = "textDocument/sourcekit.filestatus"
+
+  /// The `textDocument` this status notification is for.
+  public var textDocument: TextDocumentIdentifier
+
+  /// File state used to stylize the status (e.g. with an icon).
+  public var state: SourceKitFileState
+
+  /// Severity of the status. Used to stylize the status (e.g. with a color or icon).
+  public var severity: DiagnosticSeverity
+
+  /// Human readable description.
+  public var message: String
+
+  /// Short operation label for the status, if any.
+  public var operation: String?
+
+  public init(
+    textDocument: TextDocumentIdentifier,
+    state: SourceKitFileState,
+    severity: DiagnosticSeverity,
+    message: String,
+    operation: String? = nil
+  ) {
+    self.textDocument = textDocument
+    self.state = state
+    self.severity = severity
+    self.message = message
+    self.operation = operation
+  }
+}
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -56,7 +56,6 @@ public final class SourceKitServer: LanguageServer {
 
   /// Creates a language server for the given client.
   public init(client: Connection, fileSystem: FileSystem = localFileSystem, options: Options, onExit: @escaping () -> Void = {}) {
-
     self.fs = fileSystem
     self.toolchainRegistry = ToolchainRegistry.shared
     self.options = options
@@ -397,6 +396,28 @@ extension SourceKitServer: BuildSystemDelegate {
       }
     }
   }
+
+  /// Handle a file status change from the `BuildSystem`.
+  ///
+  /// We inform the respective language services as long as the given file is open
+  /// (not queued for opening) and file status notifications have been enabled by the client.
+  public func fileStatusesChanged(_ changes: [DocumentURI: FileStatus]) {
+    queue.async {
+      guard let workspace = self.workspace, self.options.languageServerOptions.fileStatus else {
+        return
+      }
+      let documentManager = workspace.documentManager
+      let openDocuments = documentManager.openDocuments
+      for (uri, change) in changes {
+        // Non-ready documents should be considered open even though we haven't
+        // opened it with the language service yet.
+        guard openDocuments.contains(uri) else { continue }
+        if let service = workspace.documentService[uri] {
+          service.documentFileStatusChanged(uri, fileStatus: change)
+        }
+      }
+    }
+  }
 }
 
 // MARK: - Request and notification handling
@@ -426,6 +447,9 @@ extension SourceKitServer {
         case .some(let invalid):
           log("expected null or int for 'maxResults'; got \(invalid)", level: .warning)
         }
+      }
+      if case .bool(let fileStatus) = options["sourcekit/fileStatus"] {
+        self.options.languageServerOptions.fileStatus = fileStatus
       }
     }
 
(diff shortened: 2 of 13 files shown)


DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Notification from the server about a per-file status update for the client.
```
benlangmuir, line 13: I've been putting `**LSP Extension**` into the first line of the description so it shows up prominently in quick help/hover.

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+public struct SourceKitFileStatusNotification: TextDocumentNotification, Hashable {
+  public static let method: String = "textDocument/sourcekit.filestatus"
+
+  /// The `textDocument` this status notification is for.
+  public var textDocument: TextDocumentIdentifier
+
+  /// Status category, used to stylize the status in the editor (e.g. with an icon).
+  public var category: SourceKitFileStatusCategory
+
+  /// User visible description.
```
benlangmuir, line 30: Maybe "human readable" instead of "user visible" since whether it's shown in the UI is up to the client?

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+  public var message: String
+
+  /// URI to open on click, if the editor supports it. Can be used to point
+  /// users towards an expanded description or error link.
+  public var clickTarget: DocumentURI?
```
benlangmuir, line 35: We should avoid terminology like "click".  The client defines the UI, not us, and we should provide information that is useful to a variety of clients.  How about "infoLink" or something similar?

Sources/LanguageServerProtocol/SupportTypes/SourceKitFileStatus.swift
```diff
@@ -0,0 +1,31 @@
...
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// High level categorization of a file status.
+public enum SourceKitFileStatusCategory: Int, Codable, Hashable {
```
benlangmuir, line 14: Unless you have a particular reason to choose Int, I would suggest String and to make this a struct with static fields so that we can extend it with new kinds over time.

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/SupportTypes/SourceKitFileStatus.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
DavidGoldman: Done

DavidGoldman:
> A few suggestions, but overall seems good to me. I think we should merge this at the same time as the functionality itself though. Also, how about adding documentation for this (e.g. an `Extensions.md` file in Documentation).

SGTM, will do both when I get a chance.

DavidGoldman:
PTAL, I've updated the design and added a `sourcekit/fileStatus` initialization option to disable/enable it. If all looks well I'll go ahead and add tests + documentation.

In follow up PRs I'll add the following:
- sourcekitd and clangd integration (for AST status)
- VSCode extension support

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Sources/SourceKitLSP/ToolchainLanguageServer.swift
```diff
@@ -67,3 +71,13 @@ public protocol ToolchainLanguageServer: AnyObject {
 
   func executeCommand(_ req: Request<ExecuteCommandRequest>)
 }
+
+/// Options for a `ToolchainLanguageServer`.
+public struct InternalServerOptions {
```
benlangmuir, line 76: Why not call it `ToolchainLanguageServerOptions`?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -427,6 +448,9 @@ extension SourceKitServer {
           log("expected null or int for 'maxResults'; got \(invalid)", level: .warning)
         }
       }
+      if case .bool(let fileStatus) = options["sourcekit/fileStatus"] {
```
benlangmuir, line 451: Why an initialization option rather than a client capability? This feels more like a capability to me.

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+  public static let method: String = "textDocument/sourcekit.filestatus"
+
+  /// The `textDocument` this status notification is for.
+  public var textDocument: TextDocumentIdentifier
+
+  /// File state used to stylize the status (e.g. with an icon).
+  public var state: SourceKitFileState
+
+  /// Severity of the status. Used to stylize the status (e.g. with a color or icon).
+  public var severity: DiagnosticSeverity
```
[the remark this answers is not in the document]
akyrtzi: Should this be optional and set to `nil` for the "all-OK state"?

I like the general idea of just classifying as "warning" when it is known that there may be issues. For "error" I'd probably only use it for really catastrophic events like the build service crashing or something.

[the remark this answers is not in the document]
DavidGoldman: Yeah, I was thinking this would only be used for warnings and errors, I could make it a separate type. Example cases I was thinking:

* Warning: file using fallback build settings or needs prebuilding
* Error: Internal error or invalid build set up (e.g. error in Package.swift, error in BUILD file)

I separated this from the working vs. ready status to denote loading when in the error / warning state (e.g. if reloading Package.swift files). In VS Code, you can use a separate color to handle the severity (or an icon), which makes the warning/error states more pronounced. Let me know which you'd prefer.

Sources/LanguageServerProtocol/Notifications/SourceKitFileStatusNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+
+  /// Human readable description.
+  public var message: String
```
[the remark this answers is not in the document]
DavidGoldman: I was just thinking the string would be empty then, but I could make it optional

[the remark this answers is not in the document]
DavidGoldman: Also: clangd provides a description about the work it is doing; I was also thinking it could be useful to denote build system state such as the following:

* Target the build settings are being resolved from
* SDK platform/version, toolchain version

shahmishal:

PR #292, opened 2020-06-29
[Discussion] Add scheme change handler to reduce index visibility

ghvg1313:
`SourcekitServer` will handle scheme change, caching outputs and reduce index visibility accordingly. the term "Build scheme" here refers to the build targets user desire to focus on
It's unclear yet what endpoint we are going to use for this. so the remaining work includes:
- connect to the correct endpoint
- Initialize indexstoredb with explicit mode (or according to build server/ide's ability to provide build scheme)


ghvg1313:
Will add scheme notification and build graph pulling

PR #296, opened 2020-07-20
Dup and redirect stdout to defend against accidental print()

benlangmuir:
We currently have a regression where sourcekit-lsp stops working  in
some packages, because swiftpm started printing a warning to stdout.
While we should also fix swiftpm, it behooves us to be more robust in
sourcekit-lsp in case this happens again. Dup stdout for use by the
protocol, and redirect the stdout file descriptor to stderr, which does
not interfere with the protocol.

### DIFF
--- a/Sources/sourcekit-lsp/main.swift
+++ b/Sources/sourcekit-lsp/main.swift
@@ -107,10 +107,20 @@ do {
   exit(1)
 }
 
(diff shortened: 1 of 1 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please build toolchain

PR #299, opened 2020-08-01
[SKSwiftPMWorkspace] Set swiftpm config file when creating Workspace

aciidb0mb3r:
This is quite important to set otherwise SourceKit-LSP sets off a wrong
package resolution operation potentially resulting in a wrong swiftpm
package graph.

Ideally, SwiftPM should have APIs that don't force clients to handle all
this configuration but we don't have that today.


aciidb0mb3r:
@swift-ci test

PR #300, opened 2020-08-04
[logging] Avoid logging cancellation as warnings/errors

benlangmuir:
Cancellation shows up as an error code, but it is not a real error, so
don't log it like one.

https://bugs.swift.org/browse/SR-13344


benlangmuir:
@swift-ci please test

PR #301, opened 2020-08-05
[gardening] Fix warning about strerror return value

benlangmuir:
strerror should never return nil, and in the worst case it's already in
a fatalError. Fix the warning about converting an optional value to a
string.


benlangmuir:
@swift-ci please test

PR #303, opened 2020-08-27
Fix `#file` warning when building with Swift 5.3

MaxDesiatov:
Since `#file` is deprecated, using it with Swift 5.3 causes a warning. There's a workaround that delegates to either `#file` or `#filePath` depending on the current compiler version. The workaround is copied from https://github.com/apple/swift-nio/pull/1526. 

Let me know if you think just using `#filePath` everywhere and bumping the requirement to Swift 5.3 in `Package.swift` is a better option. I'm not sure if Swift 5.3 is used on CI for this repository though for that option to work well right now.

### DIFF
--- a/Sources/LSPLogging/Logging.swift
+++ b/Sources/LSPLogging/Logging.swift
@@ -42,7 +42,7 @@ public func logAsync(level: LogLevel = .default, messageProducer: @escaping (_ c
 /// If `level >= Logger.shared.currentLevel`, it will be emitted. However, the converse is not necessarily true: on platforms that provide `os_log`, the message may be emitted by `os_log` according to its own rules about log level.
 ///
 /// - parameter message: The message to print.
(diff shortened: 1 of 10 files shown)


MaxDesiatov:
@swift-ci please test

benlangmuir:
As hinted at in the title of the NIO PR, this approach unfortunately does not work.  It gets the file from wherever the default argument is declared, not from the caller.

I was under the impression this warning was going away in 5.3 to let us transition cleanly: https://github.com/apple/swift-evolution/blob/master/proposals/0285-ease-pound-file-transition.md

MaxDesiatov:
Thanks for clarifying, do I understand correctly that the corresponding change hasn't landed in Xcode 12 beta 6 yet, with which I'm still getting the warnings?

benlangmuir:
Correct

PR #304, opened 2020-08-27
Adopt swift-argument-parser for argument parsing

MaxDesiatov:
After https://github.com/apple/swift-package-manager/pull/2653 is merged, I think it would make sense for SourceKit-LSP to adopt the new `ArgumentParser` module as well. I've kept the new completion server arguments as single-dash options (such as `-completion-server-side-filtering`), but as these weren't released in 5.3, I think it's a good opportunity for making them double-dash options for consistency, let me know what you think.

CC @natecook1000

PR #306, opened 2020-09-02
Allow double dash for IndexStore-related CLI arguments

MaxDesiatov:
Following up on #304, double dash is now allowed in addition to the single dash option for backward compatibility. This is possible thanks to the fix in Argument Parser 0.3.1.


MaxDesiatov:
@swift-ci please test

PR #308, opened 2020-09-06
SKSupport: update for `BOOL` conversion on Windows

compnerd:
We would fail to build on Windows due to the conversion of `BOOL` to
`Bool`, and then comparing to `0`.  This repairs the build on Windows.


compnerd:
@swift-ci please test

PR #310, opened 2020-09-06
LanguageServerProtocolJSONRPC: add a special case for Windows

compnerd:
This adds a special case for Windows which does not place the error code
in the `POSIXError` enumeration and is defined using a `#define` rather
than an enumeration resulting in no associated `rawValue` for the error.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -14,6 +14,9 @@ import Dispatch
 import Foundation
 import LanguageServerProtocol
 import LSPLogging
+#if os(Windows)
+import let WinSDK.ECANCELED
+#endif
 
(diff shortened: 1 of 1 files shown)


compnerd:
@benlangmuir I'm not particularly happy with this change.  I don't know if there is a better way to handle this.  Can you please help guide me here?

benlangmuir:
@compnerd can we fix the implementation in corelibs-foundation instead?  If ECANCELED is possible on Windows I would have expected it to be provided here: https://github.com/apple/swift-corelibs-foundation/blob/5b68bf460b81e5fb527f870794210b43bec6d8b1/Sources/Foundation/NSError.swift#L1317

compnerd:
Sure, I can try to take a stab at that.  Thanks!

compnerd:
apple/swift-corelibs-foundation#2887 adds `ECANCELED` so this should no longer be necessary.  Thanks @benlangmuir.

PR #311, opened 2020-09-07
build: use `ucrt` on Windows

compnerd:
This makes the Darwin and Windows cases "special" and makes Glibc the
default as that covers additional platforms.  The `ucrt` case that has
been added enables the manifest to be built on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -220,13 +220,9 @@ let package = Package(
 // by the external environment. This allows sourcekit-lsp to take advantage of the automation used
 // for building the swift toolchain, such as `update-checkout`, or cross-repo PR tests.
 
-#if canImport(Glibc)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

jakepetroules (approved):

benlangmuir:
Would it make sense to import Foundation and use `ProcessInfo.processInfo.environment` instead?

compnerd:
@benlangmuir - that definitely would be a much nicer way to handle this.

compnerd:
@swift-ci please test

benlangmuir (approved):

compnerd:
@benlangmuir could you merge this please?

PR #312, opened 2020-09-07
build: pin argument-parser to 0.3.0

compnerd:
This follows the change in apple/swift-package-manager#2909.  Without
this change, it is not possible to build SourceKit-LSP without the local
dependencies as the version conflict prevents a resolution.

Thanks to @owenv for pointing out the change in s-p-m!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -232,7 +232,7 @@ if getenv("SWIFTCI_USE_LOCAL_DEPS") == nil {
     .package(url: "https://github.com/apple/indexstore-db.git", .branch("master")),
     .package(url: "https://github.com/apple/swift-package-manager.git", .branch("master")),
     .package(url: "https://github.com/apple/swift-tools-support-core.git", .branch("master")),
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

MaxDesiatov:
This would probably break https://github.com/apple/sourcekit-lsp/pull/306, which I think only works with 0.3.1. @natecook1000 would you mind confirming this please? As far as I understand, the API is the same in 0.3.0 and 0.3.1 as required by semantic versioning, but a bug in 0.3.0 prevents it from working correctly?

compnerd:
Unfortunately, building s-p-m against s-a-p 0.3.1 fails due to an assertion triggering.  Right now, it is not possible to build SK-LSP as s-p-m pins to =0.3.0 and SK pins to >=0.3.1.

benlangmuir:
As @MaxDesiatov said, this change is not compatible for #306.  I confirmed that it crashes trying to run `sourcekit-lsp`.  I'm reverting that change in #313 to get us back to a working configuration.  We can reapply #306 when everything is ready.

PR #313, opened 2020-09-08
Revert "Allow double dash for IndexStore-related CLI arguments"

benlangmuir:
Reverts apple/sourcekit-lsp#306

Per https://github.com/apple/swift-package-manager/pull/2909 we need to roll back to swift-argument-parser 0.3.0, and the double-dash change is not compatible with that (it crashes at runtime).

I also discovered that the version of swift-argument-parser being used when building swift.org toolchains is not considering the Package.swift requirements: it is actually taking whatever version is specified in https://github.com/apple/swift/blob/db6e3af0023a3e7fdb15204ba95e3a63a53f6deb/utils/update_checkout/update-checkout-config.json#L66.  Right now that happens to be 0.3.0, but it's specified manually.  We will need to update that version before re-applying this change.


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #316, opened 2020-09-23
[vscode] Update dependencies to fix issues found by `npm audit`

benlangmuir:
Updates to the newer `@types/vscode` package instead of the older deprecated `vscode` one. We don't seem to need the part that was split off into `vscode-test`.


benlangmuir:
@swift-ci please test

PR #317, opened 2020-09-25
Improve performance of `workspace/symbol` for large queries + projects

DavidGoldman:
- Avoid short queries (< 3 characters) since they are slow and noisy. Xcode does the same thing.

- Limit the number of results returned:
  - Improves performance for queries which match many symbols
  - Currently take the first N results to improve performance for large projects but once we support cancellation we should take the top N results.


DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

benlangmuir (approved):

PR #318, opened 2020-09-25
Fix ExpressibleByArgument conformance on LogLevel

benlangmuir:
The initializer is failable, not throwing, and there is a default conformance provided via the integer literal that was not correct.


benlangmuir:
@swift-ci please test

PR #319, opened 2020-09-25
LanguageServerProtocolJSONRPC: make Windows path work

compnerd:
This adjusts the use of Dispatch to build on Windows.  Windows does not
provide `stdout_fileno` and `stderr_fileno`.  However, it is possible to
use `fileno` to get the associated fileno from the descriptor.

Dispatch on Windows does not deal with fd's but rather with handles.
Convert the file descriptor to a handle and pass that off to dispatch.
The handle is a non-owning reference, and should not be closed.
Fortunately, dispatch does not close the handle when the DispatchIO is
closed.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -10,10 +10,16 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(CDispatch)
+import CDispatch
+#endif
 import Dispatch
 import Foundation
 import LanguageServerProtocol
 import LSPLogging
+#if os(Windows)
+import ucrt
+#endif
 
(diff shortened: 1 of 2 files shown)


benlangmuir:
@swift-ci please test

compnerd:
@benlangmuir - hmm, is `CDispatch` not available on Darwin and `Dispatch` just imports the C interfaces as well?

benlangmuir:
I believe so - you could wrap with `hasImport(CDispatch)`

compnerd:
@swift-ci please test

benlangmuir:
@compnerd it looks like `dispatch_fd_t` is explicitly marked unavailable in Swift on Darwin.  Using Int32 on non-Windows should be fine.

compnerd:
Thanks @benlangmuir!  I'll update this.

compnerd:
@swift-ci please test

benlangmuir (approved):

compnerd:
@benlangmuir - would you mind merging the change?

PR #323, opened 2020-10-01
[test] Put shared test projects INPUTS into bundle resources

benlangmuir:
This avoids depending on the absolute path to the INPUTS and allows looking it up at run time even if the tests have moved.


benlangmuir:
@swift-ci please test

PR #324, opened 2020-10-05
SKSwiftPMWorkspace: canonicalize the path prior to use

compnerd:
Ensure that we use the file system representation for the path as
otherwise the path conversion may fail.  This is required for Windows
which ends up with mixed path separators.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -209,9 +209,9 @@ extension SwiftPMWorkspace: SKCore.BuildSystem {
       // We can't determine build settings for non-file URIs.
       return nil
     }
-    guard let path = try? AbsolutePath(validating: url.path) else {
-      return nil
-    }
+    guard let path = (url.withUnsafeFileSystemRepresentation {
+      try? AbsolutePath(validating: String(cString: $0!))
+    }) else { return nil }
 
     if let td = targetDescription(for: path) {
       return settings(for: path, language, td)


Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -209,9 +209,9 @@ extension SwiftPMWorkspace: SKCore.BuildSystem {
       // We can't determine build settings for non-file URIs.
       return nil
     }
-    guard let path = try? AbsolutePath(validating: url.path) else {
-      return nil
-    }
+    guard let path = (url.withUnsafeFileSystemRepresentation {
```
benlangmuir, line 212: I'm surprised this is only needed in one place - why don't we have the same issue for incoming URLs from the editor?

Is there a reason you preferred `withUnsafeFileSystemRepresentation` instead of  `standardizedURL.path`?

compnerd: Well, this was specifically for the test, I'm sure that I will need to do this a fair amount more as we discover the sites.

The reason for that is mostly because I did this before we had a chat, and I just failed to think about `standardizedURL.path`,  I'll want to change to that and test it, as I think it would be better to use.

compnerd: Actually, thinking more about this, I don't think that `standardizedURL.path` will do what we want.  That will give the path as a URI string rather than the FSR, which means that it cannot be used for the `AbsolutePath(validating:)` constructor.

compnerd:
I am closing this off because I believe that this is not the right approach.  I think that the problem is in tools-support-core instead.  I believe that apple/swift-tools-support-core#140 takes care of the underlying issue making this work.

PR #307, opened 2020-09-03
convert tests from #file to using #filePath

heckj:
I wandered in an saw a variety of warnings, this iterates through and cleans them up. I wasn't sure of dependencies for supporting older versions of swift for this project, so offering this up as cleanup work.

- converted the use of `#file` to `#filePath` in the tests

PR #326, opened 2020-10-07
convert tests from #file to using #filePath

heckj:
Originally PR #307 - rebased and re-opened to align to merging into `main` branch:

I wandered in an saw a variety of warnings, this iterates through and cleans them up. I wasn't sure of dependencies for supporting older versions of swift for this project, so offering this up as cleanup work.

- converted the use of `#file` to `#filePath` in the tests

### DIFF
--- a/Tests/LSPLoggingTests/SupportTests.swift
+++ b/Tests/LSPLoggingTests/SupportTests.swift
@@ -23,7 +23,7 @@ final class SupportTests: XCTestCase {
       messages.append((message, level))
     }
 
-    func check(expected: [(String, LogLevel)], file: StaticString = #file, line: UInt = #line) {
+    func check(expected: [(String, LogLevel)], file: StaticString = #filePath, line: UInt = #line) {
       testLogger.flush()
       XCTAssert(messages.count == expected.count, "\(messages) does not match expected \(expected)", file: file, line: line)
       XCTAssert(zip(messages, expected).allSatisfy({ $0.0 == $0.1 }), "\(messages) does not match expected \(expected)", file: file, line: line)
--- a/Sources/LSPTestSupport/CheckCoding.swift
+++ b/Sources/LSPTestSupport/CheckCoding.swift
@@ -17,7 +17,7 @@ import XCTest
 ///
 /// - parameter value: The value to encode/decode.
 /// - parameter json: The expected json encoding.
-public func checkCoding<T>(_ value: T, json: String, file: StaticString = #file, line: UInt = #line) where T: Codable & Equatable {
+public func checkCoding<T>(_ value: T, json: String, file: StaticString = #filePath, line: UInt = #line) where T: Codable & Equatable {
   let encoder = JSONEncoder()
   encoder.outputFormatting.insert(.prettyPrinted)
   if #available(macOS 10.13, *) {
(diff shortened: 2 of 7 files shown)


Tests/LSPLoggingTests/SupportTests.swift
```diff
@@ -25,8 +25,8 @@ final class SupportTests: XCTestCase {
 
     func check(expected: [(String, LogLevel)], file: StaticString = #file, line: UInt = #line) {
       testLogger.flush()
-      XCTAssert(messages.count == expected.count, "\(messages) does not match expected \(expected)", file: file, line: line)
-      XCTAssert(zip(messages, expected).allSatisfy({ $0.0 == $0.1 }), "\(messages) does not match expected \(expected)", file: file, line: line)
+        XCTAssert(messages.count == expected.count, "\(messages) does not match expected \(expected)", file: (file), line: line)
```
benlangmuir, line 28: I think the correct change here is that `#file` should be changed to `#filePath` in `func check` above.

There are a number of other places in this patch that are passing `(file)` that should be unnecessary once the declaration has been changed to use `#filePath`.

Also, there seems to be some unexpected changes to indentation that should be removed.

benlangmuir (changes requested):

Tests/LSPLoggingTests/SupportTests.swift
heckj: Thanks @benlangmuir - I'll run back through it and revisit all the spots I tried to convert and update them, and fix the indentation while I'm at it.

Tests/LSPLoggingTests/SupportTests.swift
```diff
@@ -23,10 +23,10 @@ final class SupportTests: XCTestCase {
       messages.append((message, level))
     }
 
-    func check(expected: [(String, LogLevel)], file: StaticString = #file, line: UInt = #line) {
+    func check(expected: [(String, LogLevel)], file: StaticString = #filePath, line: UInt = #line) {
       testLogger.flush()
-      XCTAssert(messages.count == expected.count, "\(messages) does not match expected \(expected)", file: file, line: line)
-      XCTAssert(zip(messages, expected).allSatisfy({ $0.0 == $0.1 }), "\(messages) does not match expected \(expected)", file: file, line: line)
+      XCTAssert(messages.count == expected.count, "\(messages) does not match expected \(expected)", file: (file), line: line)
```
benlangmuir, line 28: I see you updated some of the (file) cases, but it seems there are still a few more (you can search the patch for "(file)" to find the rest).

heckj: yep - I apologize for not being explicit, I hit the easy stuff, but I want to read and understand what's being done with some of the functions, and just haven't had the time to dig into it when I wasn't panicking about some external thing or another. I'm still working on it, but I don't yet know this codebase as tightly as I probably should to make these updates, so I'm learning and digging to I don't regress what's happening.

heckj:
@benlangmuir Apologies for the delay - I finally nabbed some time to circle back and finish this up, also fixing up the unintentional indenting damage I did earlier while changing things.

I tried to verify everything locally before pushing the updates, but I think I'm missing something - the `swift test` I ran is tossing an error at me, although the test that's failing locally for me isn't one that I've touched or changed, so something may be off - or I may just be doing the wrong thing to validate this locally. I've got swift 5.3 with an Xcode based toolchain, and the test command is failing to build, throwing this error:

```
/Users/heckj/src/sourcekit-lsp/Tests/LanguageServerProtocolJSONRPCTests/CodingTests.swift:130:53: error: 'nil' is not compatible with expected argument type 'RequestID'
    checkMessageCoding(ResponseError.cancelled, id: nil, json: """
                                                    ^
```

So I *think* this PR is otherwise good to go, but wanted to call that out just in case.

benlangmuir (approved):

benlangmuir:
@swift-ci please test

PR #327, opened 2020-10-07
Log to stderr instead of sending window/logMessage

DavidGoldman:
- clangd itself already logs to stderr so sourcekit-lsp
  should do the same for consistency (unless we want to
  capture clangd's stderr and forward it in the LSP)

- Editors such as VS Code will show stderr output in the same
  output channel where it shows logMessage + traces

### DIFF
--- a/Sources/LSPLogging/Logging.swift
+++ b/Sources/LSPLogging/Logging.swift
@@ -83,6 +83,9 @@ public final class Logger {
 
   var disableNSLog: Bool = false
 
+  /// Used to log to stderr when `NSLog` logging is disabled.
+  var dateFormatter: DateFormatter
+
   /// The current logging level.
   public var currentLevel: LogLevel {
     get { return logLevelQueue.sync { _currentLevel } }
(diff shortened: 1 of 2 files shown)


DavidGoldman:
Let me know if you want me to change the time format or also include the logging level; was thinking of sending a follow up change to make requests/responses not logged by default but then log everything upon a debug log level just like clangd.

DavidGoldman:
@swift-ci please test

Sources/LSPLogging/Logging.swift
```diff
@@ -181,11 +187,19 @@ public final class Logger {
...
+    fputs(fullMessage, stderr)
+  }
+
   private func logImpl(_ message: String, level: LogLevel, usedOSLog: Bool) {
 
     if !self.disableNSLog && !usedOSLog {
       // Fallback to NSLog if os_log isn't available.
       NSLog(message)
+    } else {
+      self.logToStderr(message, level: level)
```
benlangmuir, line 202: I just checked: `disableNSLog` is never true except in unit tests.  Should we just rely on NSLog?

DavidGoldman: Hmm we could but won't that duplicate os_logs (os_log from us, NSLog)? If we drop our manual os_log. usage we could but it would be a slight change in behavior (filtering before logging)

benlangmuir: Ah sorry, I misread how it would work.

PR #330, opened 2020-10-08
Add --destination option for cross-compiled projects

MaxDesiatov:
Currently SourceKit-LSP can't be used for projects that can't be compiled on a host platform and need to be cross-compiled with an appropriate destination file passed to `swift build`. This PR adds a `--destination` option to SourceKit-LSP that is then passed to SwiftPM workspaces, which in my undestanding is equivalent to passing `--destination` to the underlying `swift build` invocations.

### DIFF
--- a/Sources/SKCore/BuildSetup.swift
+++ b/Sources/SKCore/BuildSetup.swift
@@ -10,17 +10,15 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SKSupport
 import TSCBasic
 import TSCUtility
-import SKSupport
 
 /// Build configuration
 public struct BuildSetup {
-
   /// Default configuration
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci please test

MaxDesiatov:

benlangmuir:
We should add a test to `SwiftPMWorkspaceTests` for this.  I think you could take a test like `testBuildSetup` and instead of checking the customization of BuildFlags, check the customization on the destination.

MaxDesiatov:
It does not seem to work as I expected, Looks like initializing a `Workspace` with these arguments is not fully equivalent to the `swift build --destination` CLI invocation, the inferred SDK path is not the same, which leads to issues. I'm closing this for now until a better solution is found.

PR #333, opened 2020-10-13
Log error before fatalError(...) when message decoding fails

flowtoolz:
* Uses `log(...)` to give the client some feedback before terminating the process in response to an undecodable message


MaxDesiatov:
@swift-ci please test

benlangmuir (approved):

PR #334, opened 2020-10-14
Synchronously send error response on message decoding failure

flowtoolz:
* Replaces two logs since those don't result in response messages anymore
* Adds a func for sending messages synchronously, to get the two messages out before fatalError

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -172,6 +172,12 @@ public final class JSONRPCConnection {
     }
     return ready
   }
+  
+  private func send(_ message: JSONRPCMessage, async: Bool = true) {
+    send(async: async) { encoder in
+      try encoder.encode(message)
+    }
+  }
 
   /// Parse and handle all messages in `bytes`, returning a slice containing any remaining incomplete data.
   func parseAndHandleMessages(from bytes: UnsafeBufferPointer<UInt8>) -> UnsafeBufferPointer<UInt8>.SubSequence {
@@ -209,9 +215,7 @@ public final class JSONRPCConnection {
         switch error.messageKind {
           case .request:
             if let id = error.id {
-              send { encoder in
-                try encoder.encode(JSONRPCMessage.errorResponse(ResponseError(error), id: id))
-              }
+              send(.errorResponse(ResponseError(error), id: id))
               continue MESSAGE_LOOP
             }
           case .response:
@@ -229,17 +233,19 @@ public final class JSONRPCConnection {
               continue MESSAGE_LOOP
             }
           case .unknown:
-            log("error decoding message: \(error.message)", level: .error)
+            send(.errorResponse(ResponseError(error), id: nil),
+                 async: false) // synchronous because the following fatalError
             break
         }
         // FIXME: graceful shutdown?
-        Logger.shared.flush()
         fatalError("fatal error encountered decoding message \(error)")
 
       } catch {
-        log("error decoding message: \(error.localizedDescription)", level: .error)
+        let responseError = ResponseError(code: .parseError,
+                                          message: "Failed to decode message. \(error.localizedDescription)")
+        send(.errorResponse(responseError, id: nil),
+             async: false) // synchronous because the following fatalError
         // FIXME: graceful shutdown?
-        Logger.shared.flush()
         fatalError("fatal error encountered decoding message \(error)")
       }
     }
@@ -265,8 +271,12 @@ public final class JSONRPCConnection {
       }
       outstanding.replyHandler(.success(response))
     case .errorResponse(let error, id: let id):
+      guard let id = id else {
+        log("Received error response for unknown request: \(error.message)", level: .error)
+        return
+      }
       guard let outstanding = outstandingRequests.removeValue(forKey: id) else {
-        log("Unknown request for \(id)", level: .error)
+        log("No outstanding requests for request ID \(id)", level: .error)
         return
       }
       outstanding.replyHandler(.failure(error))
@@ -303,7 +313,17 @@ public final class JSONRPCConnection {
     send(_rawData: dispatchData)
   }
 
-  func send(encoding: (JSONEncoder) throws -> Data) {
+  private func sendMessageSynchronously(_ messageData: Data) {
+    let header = "Content-Length: \(messageData.count)\r\n\r\n"
+    guard let headerData = header.data(using: .utf8) else {
+      log("error encoding header", level: .error)
+      return
+    }
+    let stdOutHandle = FileHandle(fileDescriptor: sendIO.fileDescriptor)
+    stdOutHandle.write(headerData + messageData)
+  }
+  
+  func send(async: Bool = true, encoding: (JSONEncoder) throws -> Data) {
     guard readyToSend() else { return }
 
     let encoder = JSONEncoder()
@@ -317,7 +337,11 @@ public final class JSONRPCConnection {
       fatalError("unexpected error while encoding response: \(error)")
     }
 
-    send(messageData: data)
+    if async {
+      send(messageData: data)
+    } else {
+      sendMessageSynchronously(data)
+    }
   }
 
   /// Close the connection.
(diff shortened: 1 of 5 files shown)


flowtoolz:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -303,7 +313,17 @@ public final class JSONRPCConnection {
...
   }
 
-  func send(encoding: (JSONEncoder) throws -> Data) {
+  private func sendMessageSynchronously(_ messageData: Data) {
+    let header = "Content-Length: \(messageData.count)\r\n\r\n"
+    guard let headerData = header.data(using: .utf8) else {
+      log("error encoding header", level: .error)
+      return
+    }
+    let stdOutHandle = FileHandle(fileDescriptor: sendIO.fileDescriptor)
```
benlangmuir, line 322: This seems unsafe since there may be other asynchronous writers.  How about something like this?

```swift
let semaphore = DispatchSemaphore(value: 0)
sendIO.barrier {
  ... write to the file descriptor, which we now have exclusive access to ...
  semaphore.signal()
}

_ = semaphore.wait(timeout: .now() + .seconds(3))
```

Although perhaps the specific timeout should be passed in.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -303,7 +313,17 @@ public final class JSONRPCConnection {
     send(_rawData: dispatchData)
   }
 
-  func send(encoding: (JSONEncoder) throws -> Data) {
+  private func sendMessageSynchronously(_ messageData: Data) {
+    let header = "Content-Length: \(messageData.count)\r\n\r\n"
+    guard let headerData = header.data(using: .utf8) else {
+      log("error encoding header", level: .error)
```
benlangmuir, line 319: This seems like it ought to be a fatalError.  There is no reason this should ever happen.

Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
```diff
@@ -17,7 +17,7 @@ public enum JSONRPCMessage {
   case notification(NotificationType)
   case request(_RequestType, id: RequestID)
   case response(ResponseType, id: RequestID)
-  case errorResponse(ResponseError, id: RequestID)
```
benlangmuir, line 20: We decode this value below like so:

```swift
      case (let id?, nil, _, let error?):
        msgKind = .response
        self = .errorResponse(error, id: id)
```

... which assumes that the `id` is non-nil.  That probably needs to get updated to `let id` instead of `let id?`.  Please add a test to `Tests/LanguageServerProtocolJSONRPCTests/CodingTests.swift` ensuring that we can round-trip this correctly when `id == nil`.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -172,6 +172,12 @@ public final class JSONRPCConnection {
     }
     return ready
   }
+  
+  private func send(_ message: JSONRPCMessage, async: Bool = true) {
```
benlangmuir, line 176: We should test the new `async == false` behaviour. We can underscore and make this public:

```swift
/// *Public for testing*
public func _send(...)
```

and add a test in `Tests/LanguageServerProtocolJSONRPCTests/ConnectionTests.swift` that uses this.  There are other tests that do similar things like sending raw data.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -229,17 +233,19 @@ public final class JSONRPCConnection {
               continue MESSAGE_LOOP
             }
           case .unknown:
-            log("error decoding message: \(error.message)", level: .error)
+            send(.errorResponse(ResponseError(error), id: nil),
```
benlangmuir, line 236: We can't test this behaviour directly due to the fatalError, but it might be good to add a test for the JSONRPCConnection receiving an error with `nil` id.  For example, if you manually send an `errorResponse(..., id: nil)` make sure it doesn't crash the receiving connection.  This would fit in `Tests/LanguageServerProtocolJSONRPCTests/ConnectionTests.swift`

benlangmuir (changes requested):

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
flowtoolz: With this one I better ask you before going down the wrong rabbit hole ... I don't see an easy way for testing that a `JSONRPCMessage`s or a concrete error response reaches the receiver. I'd want to write somethin like this:

```
func testSendErrorResponseWithNilID() {
    let client = connection.client

    let expectation = self.expectation(description: "received message")
    client.handleNextMessage { (message: JSONRPCMessage) in
      expectation.fulfill()
    }
  
    connection.serverConnection._send(.errorResponse(.cancelled, id: nil))

    waitForExpectations(timeout: 10)
  }
```

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
flowtoolz: Tried to address this in https://github.com/apple/sourcekit-lsp/pull/334/commits/3f4c1c5c29a509809ab18d405b2ea0a7176afea7 without the `barrier` and preferring existing means ... I hope the `completionHandler` is ok?

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
flowtoolz: done

Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
flowtoolz: done

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
flowtoolz: done

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
benlangmuir: Ah I see we don't have a good way to test this right now.  Let's skip this.

benlangmuir:
@swift-ci please test

benlangmuir (commented):
Looks good to me.  Please regenerate the tests for Linux (`swift test --generate-linuxmain`), and I think it will be ready to go :-)

flowtoolz:
Done. Thanks for the learning opportunity 😊

benlangmuir:
@swift-ci please test

benlangmuir:
```
16:01:54 Test Case '-[LanguageServerProtocolJSONRPCTests.CodingTests testMessageDecodingError]' started.
16:01:54 
/Users/buildnode/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/sourcekit-lsp/Tests/LanguageServerProtocolJSONRPCTests/CodingTests.swift:161: error: -[LanguageServerProtocolJSONRPCTests.CodingTests testMessageDecodingError] : failed - expected error not seen
16:01:54 Test Case '-[LanguageServerProtocolJSONRPCTests.CodingTests testMessageDecodingError]' failed (1.289 seconds).
```

flowtoolz:
Oh, I guess this is now a decodable error response message:

```
{"jsonrpc":"2.0","error":{"code":-32000,"message":""}}
```

So decoding that produces no error in CodingTests.swift:161. Should I remove the test or adjust it? I'm not sure what kind of LSP violation the test is testing for ...

benlangmuir:
Sounds good to me

benlangmuir:
@swift-ci please test

flowtoolz:
done. this time, regenerating the Linux tests had no effect, just so you know 🤷🏼‍♂️

PR #335, opened 2020-10-14
Fix code-completion session repeated failure issue

DavidGoldman:
- Fix for https://bugs.swift.org/browse/SR-13561 by making sure the
session close waits for the open to finish

- Add a regression test

### DIFF
--- a/Tests/SourceKitLSPTests/SwiftCompletionTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftCompletionTests.swift
@@ -575,37 +575,35 @@ final class SwiftCompletionTests: XCTestCase {
         textDocument: TextDocumentIdentifier(url),
       position: Position(line: 8, utf16index: 12), // prin^
       context:CompletionContext(triggerKind: .invoked))
-    let coordinationGroup = DispatchGroup()
 
     // Code completion for "self.forSome"
-    coordinationGroup.enter()
+    let forSomeExpectation = XCTestExpectation(description: "self.forSome code completion")
     _ = sk.send(forSomeComplete) { result in
-      defer { coordinationGroup.leave() }
+      defer { forSomeExpectation.fulfill() }
       guard let list = result.success else {
         XCTFail("Request failed: \(String(describing: result.failure))")
         return
       }
-      XCTAssert(countFs(list) == 2)
+      XCTAssertEqual(2, countFs(list))
     }
 
     // Code completion for "self.prin", previously could immediately invalidate
     // the previous request.
-    coordinationGroup.enter()
+    let printExpectation = XCTestExpectation(description: "self.prin code completion")
     _ = sk.send(printComplete) { result in
-      defer { coordinationGroup.leave() }
+      defer { printExpectation.fulfill() }
       guard let list = result.success else {
         XCTFail("Request failed: \(String(describing: result.failure))")
         return
       }
-      XCTAssertEqual(list.items.count, 1)
+      XCTAssertEqual(1, list.items.count)
     }
 
-    // Wait for the previous 2 completions to finish.
-    coordinationGroup.wait()
+    wait(for: [forSomeExpectation, printExpectation], timeout: 10)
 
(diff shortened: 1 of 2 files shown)


DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -547,6 +547,66 @@ final class SwiftCompletionTests: XCTestCase {
...
+
+    let forSomeComplete = CompletionRequest(
+          textDocument: TextDocumentIdentifier(url),
+        position: Position(line: 4, utf16index: 12), // forS^
+        context:CompletionContext(triggerKind: .invoked))
+    let printComplete = CompletionRequest(
+        textDocument: TextDocumentIdentifier(url),
+      position: Position(line: 8, utf16index: 12), // prin^
+      context:CompletionContext(triggerKind: .invoked))
+    let coordinationGroup = DispatchGroup()
```
benlangmuir, line 578: Using XCTest expectations will behave better during testing.  For one thing, they will provide better behaviour if something goes wrong (timeout indicating which request didn't finish).  Another difference is that if there is anything in the service that needs to use the main thread, XCTest will automatically cycle the main run loop when waiting on an expectation, while blocking would otherwise cause a deadlock in that situation.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -547,6 +547,66 @@ final class SwiftCompletionTests: XCTestCase {
...
+
+    // Code completion for "self.forSome"
+    coordinationGroup.enter()
+    _ = sk.send(forSomeComplete) { result in
+      defer { coordinationGroup.leave() }
+      guard let list = result.success else {
+        XCTFail("Request failed: \(String(describing: result.failure))")
+        return
+      }
+      XCTAssert(countFs(list) == 2)
```
benlangmuir, line 588: XCTAssertEqual gives better reports if somethings fails.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -547,6 +547,66 @@ final class SwiftCompletionTests: XCTestCase {
...
+      }
+      XCTAssertEqual(list.items.count, 1)
+    }
+
+    // Wait for the previous 2 completions to finish.
+    coordinationGroup.wait()
+
+    // Try code completion for "self.forSome" again to verify that it still works.
+    let result = try sk.sendSync(forSomeComplete)
+    XCTAssert(countFs(result) == 2)
```
benlangmuir, line 608: Same here for XCTAssertEqual

benlangmuir (changes requested):

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
DavidGoldman: Done

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
DavidGoldman: Done

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
DavidGoldman: Done

DavidGoldman:
@swift-ci please test

benlangmuir (approved):
LGTM, also verified locally in an editor that this behaves as expected.
* No longer hits the session error easily. Before this change I could somewhat reliably reproduce this by typing `x.` (completion for "x" then completion after ".") and then going back and completing "x" again in the same location.
* If I do manage to hit the error by doing repeated completions in the same location it only happens once and doesn't get stuck in a bad state.

PR #288, opened 2020-06-09
Fix a number of resource leaks when using SourceKit-LSP in tests or as a library

benlangmuir:
* Shutdown toolchain connections on exit. This also fixes [SR-11309](https://bugs.swift.org/browse/SR-11309).
* Fix reference cycle for build system manager delegate.

PR #336, opened 2020-10-16
[5.3-20201012] [completion] Filter completions on the server side

benlangmuir:
Cherry-pick the server-side filtering-related changes to `release/5.3-20201012`

* #298 [completion] Filter completions on the server side
* #335 Fix code-completion session repeated failure issue
* #288 Fix a number of resource leaks when using SourceKit-LSP in tests or as a library 
* Changes to the command-line argument spelling (-- vs -) to match the main branch without pulling in the full swift-argument-parser change.

rdar://69374656


benlangmuir:
@swift-ci please test

akyrtzi (approved):

benlangmuir:
* **Explanation**: Code-completion server-side filtering gives a major performance improvement to completion, particularly when completing global values. For developers who import large-ish modules, this is the difference between completion being responsive and being unusably slow (several seconds per global completion). 
* **Scope**: Affects code-completion in sourcekit-lsp.
* **Risk**: We have been using this change successfully on the main branch for a few months now, and the only impactful known issue was recently fixed and is included in this cherry-pick. If there are unanticipated regressions, command-line options can be used to revert to the old behaviour.  The biggest risk is that there were many sourcekitd-related changes on main branch that we do not have on 5.3 and it is possible that in edge cases related to server shutdown we will see issues.  That would most likely affect unit testing, not real users.
* **Testing**: Regression tests included. Also tested manually in an editor.
* **Issue**: rdar://69374656
* **Reviewer**: Argyrios

@swift-ci please nominate

svermeulen:
This is working for me.  Without this, in my environment in my particular project, source kit LSP is unusably slow. With this, things seem good 👍

PR #337, opened 2020-10-18
[build-script-helper] make it run under Python 3 and run tests in parallel on Android

buttaface:
There's no longer a `platform.linux_distribution()` by default with Python 3.9, so just check for Android early and run the tests in parallel. I haven't seen it hang on Android when running these tests.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -18,7 +18,7 @@ def swiftpm_bin_path(swift_exec, swiftpm_args, env=None):
   swiftpm_args = list(filter(lambda arg: arg != '-v' and arg != '--verbose', swiftpm_args))
   cmd = [swift_exec, 'build', '--show-bin-path'] + swiftpm_args
   print(' '.join(cmd))
-  return subprocess.check_output(cmd, env=env).strip()
+  return subprocess.check_output(cmd, env=env, universal_newlines=True).strip()
 
 def get_swiftpm_options(args):
   swiftpm_args = [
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

Utilities/build-script-helper.py
```diff
@@ -87,7 +87,7 @@ def delete_rpath(rpath, binary):
 
 
 def should_test_parallel():
-  if platform.system() == 'Linux':
+  if platform.system() == 'Linux' and 'ANDROID_DATA' not in os.environ:
     distro = platform.linux_distribution()
     if distro[0] != 'Ubuntu':
       # Workaround hang in Process.run() that hasn't been tracked down yet.
```
benlangmuir, line 93: I actually tracked down the bug this is working around after the comment was written and it is a bug in posix_spawn that only affected glibc-2.24.  I don't think we're specifically trying to support any distros that use the affected glibc right now, so I'd be okay with just removing this check.  If we do want to support this, the correct fix would probably be to patch corelibs-foundation's Process.run to not use posix_spawn on that glibc version (there is already a fallback implementation in foundation not using posix_spawn, but it's currently only enabled when posix_spawn is missing entirely).

buttaface: Removed it and added a comment with your info, in case somebody hits that later.

benlangmuir:
@swift-ci please test

benlangmuir (approved):

PR #339, opened 2020-10-22
Don't log the sourcekit code complete request

DavidGoldman:
Seems to be left over from development; SOURCEKIT_LOGGING is
an alternative to trace this.


DavidGoldman:
@swift-ci please test

benlangmuir (approved):

PR #341, opened 2020-10-28
Fix the Pipe handling on Windows to be correct

compnerd:
This actually addresses the real issue that was ignored earlier about
pipes on Windows.  The FileHandle cannot provide a non-owning file
descriptor (the returned file descriptor would need to be explicitly
`_close`'d by the receiver).  Foundation now vends a `_handle` accessor
to the OS primitive handle.  Use this to create the dispatch loop for
messaging.  We now create the JSONRPCConnection from handles on Windows
which actually should help enable running some of the tests on Windows
as well.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -11,15 +11,12 @@
 //===----------------------------------------------------------------------===//
 
 #if canImport(CDispatch)
-import CDispatch
+import struct CDispatch.dispatch_fd_t
 #endif
 import Dispatch
 import Foundation
 import LanguageServerProtocol
 import LSPLogging
-#if os(Windows)
-import ucrt
-#endif
 
 /// A connection between a message handler (e.g. language server) in the same process as the connection object and a remote message handler (e.g. language client) that may run in another process using JSON RPC messages sent over a pair of in/out file descriptors.
 ///
@@ -62,8 +59,8 @@ public final class JSONRPCConnection {
 
   public init(
     protocol messageRegistry: MessageRegistry,
-    inFD: Int32,
-    outFD: Int32,
+    inFD: FileHandle,
+    outFD: FileHandle,
     syncRequests: Bool = false)
   {
     state = .created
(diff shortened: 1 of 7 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -60,10 +63,16 @@ public final class JSONRPCConnection {
...
+  public typealias PipeHandle = HANDLE
+#else
+  public typealias PipeHandle = Int32
+#endif
+
   public init(
     protocol messageRegistry: MessageRegistry,
-    inFD: Int32,
-    outFD: Int32,
+    inFD: PipeHandle,
```
benlangmuir, line 74: Could we use `FileHandle` here? I think with that change, most users of `JSONRPCConnection` could work without `#if os(Windows)`, and we would only need special cases (1) in the implementation here, and (2) possibly in main.swift - I'm not sure if you can get a FileHandle for `realStdout` in a portable way.

compnerd: Sure, I can convert to passing a `FileHandle` instead of a FD, that should make the ergonomics of this API way better IMO.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/sourcekit-lsp/main.swift
```diff
@@ -161,8 +161,8 @@ struct Main: ParsableCommand {
 
     let clientConnection = JSONRPCConnection(
       protocol: MessageRegistry.lspProtocol,
-      inFD: fileno(stdin),
-      outFD: realStdout,
+      inFD: FileHandle(fileDescriptor: fileno(stdin), closeOnDealloc: false),
```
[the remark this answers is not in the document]
compnerd: Ah, yes, good point.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/sourcekit-lsp/main.swift
```diff
@@ -161,8 +161,8 @@ struct Main: ParsableCommand {
...
+      outFD: FileHandle.standardInput,
```
benlangmuir, line 165: You changed the wrong file handle; should be something like this, I think:

```swift
inFD: FileHandle.standardInput,
outFD: FileHandle(fileDescriptor: realStdout, closeOnDealloc: false),
```

In its current state it's not likely to behave very well 🤣

benlangmuir (changes requested):

Sources/sourcekit-lsp/main.swift
compnerd: Nice!  Totally a force merge candidate :-p

compnerd:
@swift-ci please test

benlangmuir (approved):

PR #344, opened 2020-11-03
Rename SwiftPMConfig to Workspace.Configuration

mattt:
This PR corresponds to public API changes in https://github.com/apple/swift-package-manager/pull/3002, where `SwiftPMConfig` is renamed to `Workspace.Configuration` and its designated initializer now `throws`.

As it stands, CI will fail because the package manifest points to the `main` branch of @apple/swift-package-manager, which currently doesn't include this API change.

(This PR also makes an additional change with 84df7d38609e170962419bd114bfc47ca9c4f4d3 to rename a local variable that's misleadingly named. This is entirely severable, and you're welcome to back that out to be considered in a separate patch)

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -101,17 +101,17 @@ public final class SwiftPMWorkspace {
 
     let triple = toolchain.triple
 
-    let swiftPMConfiguration: PackageModel.BuildConfiguration
+    let buildConfiguration: PackageModel.BuildConfiguration
     switch buildSetup.configuration {
     case .debug:
-      swiftPMConfiguration = .debug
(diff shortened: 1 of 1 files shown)


benlangmuir (approved):

benlangmuir:
https://github.com/apple/swift-package-manager/pull/3002
@swift-ci please test

benlangmuir:
> As it stands, CI will fail because the package manifest points to the main branch of @apple/swift-package-manager, which currently doesn't include this API change.

Swift CI uses the package sources checked out on the bot, not the URL dependency: https://github.com/apple/sourcekit-lsp/blob/0068eae68f220f7f1ae1d9dca590f8f35fd35453/Package.swift#L234

mattt:
@benlangmuir Thanks for clarifying. I noticed @neonichu doing the same trick with the smoke test on swift-package-manager.

PR #345, opened 2020-11-07
Expose diagnostic educational notes as diagnostic codes

owenv:
This is an updated version of https://github.com/apple/sourcekit-lsp/pull/275 which works with both 5.3 and trunk toolchains, and uses the latest version of the Proposed LSP 3.16 spec.

A couple of notes:
- LSP 3.16 still isn't final, but hopefully will be soonish afaict. The "diagnostic links" feature has changed quite a bit since the last version of this PR. Now, the link is contained in an optional `CodeDescription` property on diagnostics.
- Once the spec is final, some of the dependencies of the VSCode extension will need to be updated to take advantage of this change. Using the current version of the extension with this change doesn't break anything, but the codes don't get displayed. Getting users to update might be a problem since they will need to build the extension from source to get the latest features.

Screen Recording 2020-11-07 at 11 22 34 AM

### DIFF
--- a/Sources/SourceKitLSP/Swift/Diagnostic.swift
+++ b/Sources/SourceKitLSP/Swift/Diagnostic.swift
@@ -107,7 +107,9 @@ extension TextEdit {
 extension Diagnostic {
 
   /// Creates a diagnostic from a sourcekitd response dictionary.
-  init?(_ diag: SKDResponseDictionary, in snapshot: DocumentSnapshot) {
+  init?(_ diag: SKDResponseDictionary,
+        in snapshot: DocumentSnapshot,
+        useEducationalNoteAsCode: Bool) {
     // FIXME: this assumes that the diagnostics are all in the same file.
 
     let keys = diag.sourcekitd.keys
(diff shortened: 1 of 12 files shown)


owenv:
@swift-ci please test

Sources/SourceKitLSP/Swift/Diagnostic.swift
```diff
@@ -141,6 +143,23 @@ extension Diagnostic {
...
+    var code: DiagnosticCode? = nil
+    var codeDescription: CodeDescription? = nil
+    // If this diagnostic has one or more educational notes, the first one is
+    // treated as primary and used to fill in the diagnostic code and
+    // description. `useEducationalNoteAsCode` ensures a note name is only used
+    // as a code if the cline supports an extended code description.
+    if useEducationalNoteAsCode,
+       let educationalNotePaths: SKDResponseArray = diag[keys.educational_note_paths],
+       educationalNotePaths.count > 0,
+       let primaryPath = educationalNotePaths.sourcekitd.api.variant_array_get_string(educationalNotePaths.array, 0)
```
benlangmuir, line 155: You could add this to SKDResponseArray, e.g.

```swift
subscript(index: Int) -> String? {
  if let cstr = sourcekitd.api.variant_array_get_string(array, index) {
    return String(cString: cstr)
  }
  return nil
}
```

benlangmuir:
> Once the spec is final, some of the dependencies of the VSCode extension will need to be updated 

Is the support available in a released VSCode yet, or are you anticipating this won't happen until 3.16 is finalized?

Sources/SourceKitLSP/Swift/Diagnostic.swift
owenv: good idea, done

owenv:
@benlangmuir VSCode 1.51.0 was released a few days ago and supports the new field, but previous versions have supported other incompatible versions of the feature so we might still want to wait until the spec is finalized.

owenv:
@swift-ci please test

owenv:
@swift-ci please test

owenv:
Hi @benlangmuir, it looks like the LSP 3.16 spec is finalized now (https://microsoft.github.io/language-server-protocol/specifications/specification-current/), so I think this should be good to go. I also updated the VSCode extension's dependencies to support the new additions to the protocol, but I can split that into a separate PR if you prefer

benlangmuir:
Please add `testEducationalNotesAreUsedAsDiagnosticCodes` to `Tests/SourceKitLSPTests/XCTestManifests.swift`.  We haven't adopted test discovery on Linux yet (waiting until it's available in a release).

Otherwise LGTM!

> I also updated the VSCode extension's dependencies to support the new additions to the protocol, but I can split that into a 
separate PR if you prefer

Keeping it in this PR is fine with me.

owenv:
@swift-ci please test

PR #347, opened 2020-11-13
LanguageServerProtocolJSONRPCTests: add Windows stubs

compnerd:
This set of changes is the last set of changes required to build the
entire test suite on Windows.  Although the tests do not yet fully pass,
this is a prerequisite for getting the test suite passing.


benlangmuir (commented):
To simplify this, what about just using `Sleep(DWORD milliseconds)` and bumping the value to 1 ms on Windows?  It's a probabilistic timing anyway, so it's already likely to be hardware/platform dependent how good that value is and probably not worth tuning for multiple platforms.

compnerd:
Sounds good to me

benlangmuir (approved):

benlangmuir:
@swift-ci please test

PR #348, opened 2020-11-15
Improve FileSystemError by adding associated path

MaxDesiatov:
Currently `FileSystemError` doesn't have an associated path, which makes diagnosing errors harder. This is going to be fixed in TSC and SwiftPM, but this is a breaking change. Usage of `FileSystemError` is updated in the SourceKit-LSP codebase accordingly.


MaxDesiatov:
apple/swift-package-manager#3055
@swift-ci please test

benlangmuir (approved):

PR #352, opened 2020-11-30
SKSwiftPMWorkspace: make `testBasicCXXArgs` pass on Windows

compnerd:
We would previously fail to correctly build the GNU dependency outputs
on Windows since we were checking the file system representation of the
path against the internal path representation.  This fixes the
generation of that command as well as the additional checks in the test
case to allow the test case to succeed on Windows.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #357, opened 2020-12-15
Explicitly import Glibc in files that rely on it

hlopko:
These files rely on Glibc declarations in their public interface, even though they don’t import Glibc. Currently, these declarations are discovered through CoreFoundation, but due to implementation details, Swift canonicalizes the declarations and at the end uses declarations from Glibc (which is imported by other files in the module).

This is fragile as Foundation now relies on implementation details of Glibc, which in fact we are trying to change in apple/swift#32404.

This PR fixes the issue by explicitly importing Glibc in affected files.

### DIFF
--- a/Sources/SourceKitD/SKDRequestArray.swift
+++ b/Sources/SourceKitD/SKDRequestArray.swift
@@ -11,6 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import Csourcekitd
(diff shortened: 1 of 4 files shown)


benlangmuir:
@swift-ci please test

benlangmuir (approved):

hlopko:
Thank you Ben! :)

PR #359, opened 2021-01-12
[CanonicalOSSA] sourcekit-lsp tests; fix weak reference lifetimes

atrick:
Some SourceKit properties are weak references. Several test cases
assign weak references assuming they will outlive subsequent checks
without any subsequent use keeping the reference alive.

The SIL optimizer shortens object lifetimes, breaking the test at
-O. This may even happen at -Onone in the future. The optimizer has
always done this sort of optimization, but enabling OSSA makes it much
more likely to happen in practice.

Fixes rdar://73044531 ([CanonicalOSSA] sourcekit-lsp tests; fix weak
reference lifetimes)

These tests are blocking the OSSA work.


atrick:
@swift-ci test

benlangmuir (approved):

PR #360, opened 2021-01-14
Fix deprecation warnings from swiftpm APIs

benlangmuir:


benlangmuir:
@swift-ci please test

PR #361, opened 2021-01-19
implement DocumentFormattingRequest

Trzyipolkostkicukru:
Rebased to main branch and squashed.


benlangmuir:
@Trzyipolkostkicukru sorry for not providing feedback on this earlier.  Thanks to Karl for bringing it up on https://forums.swift.org/t/the-current-state-of-swift-for-server-and-linux/47732/23.

While some of the information on the original PR is still relevant here, I also realize I didn't write down all of the things that were in my head.

Here are what I see as the blockers to adding a dependency on swift-format in sourcekit-lsp.

#### Testing

Integration of swift-format into Swift's CI system, including testing sourcekit-lsp when we change swift-format. For sourcekit-lsp's use case this doesn't require that swift-format the command-line tool is installed in the toolchain, just that it's getting tested properly. @shahmishal do you have any updates you can share?

#### Workflow

There is a workflow challenge for developing tools that are part of the toolchain and built on top of swift-syntax, like swift-format is. Swift-syntax has an unstable dependency on `lib_InternalSwiftSyntaxParser` from the toolchain. Currently, you must use an identical tag of swift-syntax and the parser library.  At the same time, the Swift API of swift-syntax is not API-stable.  This means that if we use main branch of swift-format/swift-syntax, we cannot use it without building the whole toolchain from source.  Alternatively, if we used the latest released tag version we would always be a full release behind any syntax changes to the language, and we would open ourselves up to incompatibility between what happens in a development build, and what happens when building the toolchain for swift.org.

One potential solution to this would be if we could use a binary dependency for the parser library in swift-syntax, which would allow us to pull down a matching combination of swift-syntax and parser library during development that is still very close to the main branch.  Unfortunately, there is not yet swiftpm support for binary dependencies on Linux.

mackoj:
Now that swift-parser support binary dependencies will you consider adding formatting to swift-lsp ?

ahoppen:
While swift-syntax is starting to create releases that ship the parser library as a binary dependency ([link](https://github.com/apple/swift-syntax/releases/tag/0.50600.0-SNAPSHOT-2022-01-24)), SwiftPM still doesn’t support binary dependencies on Linux, so the problem that @benlangmuir mentioned as the “Workflow” problem unfortunately still exists on Linux.

IOOI-SqAR:
> SwiftPM still doesn’t support binary dependencies on Linux, so the problem that @benlangmuir mentioned as the “Workflow” problem unfortunately still exists on Linux.

Is there already an issue for this in the SwiftPM's repo?

benlangmuir:
> Is there already an issue for this in the SwiftPM's repo?

@IOOI-SqAR I wasn't able to find one. The most recent discussion that I am aware of is https://forums.swift.org/t/binary-dependencies-on-linux/27623/1

IOOI-SqAR:
> The most recent discussion that I am aware of is https://forums.swift.org/t/binary-dependencies-on-linux/27623/1

Hi @benlangmuir ! In Order to discuss this topic in the swift forums I created this: https://forums.swift.org/t/support-binary-dependencies-on-linux-and-windows/57021 (I just don't wanted to raise this old thread you mentioned from the dead).

kemchenj:
Now that Swift had [deleted _InternalSwiftSyntaxParser and its build infrastructure](https://github.com/apple/swift/pull/61898), is there any chance that we could push this pr forward?

spenceryr:
Apologies if my understanding is incorrect, but it seems swift-format mainline [no longer depends](https://github.com/apple/swift-format/blob/main/README.md#matching-swift-format-to-your-swift-version-swift-57-and-earlier) on the Swift toolchain. Is this PR able to proceed?

ahoppen:
Yes, I don’t see a reason that would still block this PR now that the SwiftSyntax dependency on the parser library has been removed.

ahoppen:
I picked this PR up in https://github.com/apple/sourcekit-lsp/pull/769 and I’m thus closing this one.

PR #365, opened 2021-01-29
Fix a nondeterministic test failure in crash recovery

ahoppen:
In some cases (especially if the system is under heavy load), sourcekitd's crash is scheduled while the document update is in progress. In this case, the SwiftLanguageServer's queue is blocked until the document update is handled, which only happens once semantic editor functionality has been restored.
Because the queue is blocked waiting for the document update to be handled, we don't receive the crash notification within 5 seconds. Increasing the timeout to 15 seconds to allow for 10 seconds semantic functionality delay fixes the issue.

While I haven't been able to reproduce rdar://73717447 (test failure in https://ci.swift.org/view/Swift%20rebranch/job/oss-swift-rebranch-incremental-RA-macos/45/), I suspect a similar issue occurs for clang crash recovery and increasing the timeout there should also resolve the issue.

This commit also unifies the two async blocks in SwiftLanguageServer's notification handling into one. While it isn't strictly related to the issue described above, there should be less room for errors if we disallow scheduling of other blocks on the queue in between a notification handling and document update handling.

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -1100,20 +1100,18 @@ extension SwiftLanguageServer: SKDNotificationHandler {
         // Reset the document manager to reflect that.
         self.documentManager = DocumentManager()
       }
-    }
-    
-    guard let dict = notification.value else {
-      log(notification.description, level: .error)
-      return
-    }
 
-    logAsync(level: .debug) { _ in notification.description }
+      guard let dict = notification.value else {
+        log(notification.description, level: .error)
+        return
+      }
 
-    if let kind: sourcekitd_uid_t = dict[self.keys.notification],
-       kind == self.values.notification_documentupdate,
-       let name: String = dict[self.keys.name] {
+      logAsync(level: .debug) { _ in notification.description }
 
-      self.queue.async {
+      if let kind: sourcekitd_uid_t = dict[self.keys.notification],
+         kind == self.values.notification_documentupdate,
+         let name: String = dict[self.keys.name] {
+
         let uri: DocumentURI
         if name.starts(with: "/") {
           // If sourcekitd returns us a path, translate it back into a URL
--- a/Tests/SourceKitDTests/CrashRecoveryTests.swift
+++ b/Tests/SourceKitDTests/CrashRecoveryTests.swift
@@ -84,7 +84,7 @@ final class CrashRecoveryTests: XCTestCase {
 
     sourcekitdServer._crash()
 
-    self.wait(for: [sourcekitdCrashed], timeout: 5)
+    self.wait(for: [sourcekitdCrashed], timeout: 15)
     self.wait(for: [sourcekitdRestarted], timeout: 30)
 
     // Check that we have syntactic functionality again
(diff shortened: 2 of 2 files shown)


ahoppen:
@swift-ci Please test

benlangmuir:
> In some cases (especially if the system is under heavy load), sourcekitd's crash is scheduled while the document update is in progress.

If the issue with the test is the crash happening mid-document-update, instead of messing with the timeout, can we change the test to wait on having two `PublishDiagnosticsNotification`s after the document open like we do in many other tests?

> In this case, the SwiftLanguageServer's queue is blocked until the document update is handled, which only happens once semantic editor functionality has been restored.

What exactly is _blocking_ here?  I would expect sourcekitd requests to fail immediately rather than block in this scenario.

---

> While I haven't been able to reproduce rdar://73717447 (test failure in https://ci.swift.org/view/Swift%20rebranch/job/oss-swift-rebranch-incremental-RA-macos/45/), I suspect a similar issue occurs for clang crash recovery and increasing the timeout there should also resolve the issue.

I'm also suspicious that you were not able to reproduce this issue in clangd, but on the bot it fails consistently every time for all of the clangd crash recovery tests and that we have only seen this issue after the llvm rebranch, so with lots of clangd changes landing.  To confirm: were you building the llvm rebranch when testing?

Also, if the issue is that we are sending a crash pragma text edit while clangd is busy, can we wait on whatever clangd is doing explicitly?  E.g. if it's building an AST, we could wait on the diagnostics showing up (not sure if clangd publishes when there are no diagnostics, but we could introduce one).

ahoppen:
> What exactly is blocking here? I would expect sourcekitd requests to fail immediately rather than block in this scenario.

The issue appears to be that we 
- Receive the document update notification
- Enter the notification [receive handler](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L1077) to about [this point](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L1110)
- sourcekitd crashes
- We tell `sourcekitd` to handle the document update [here](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L1124)
- handleDocumentUpdate waits synchronously for `sourcekitd` [here](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L220) which appears to be blocking until `sourcekitd` has restored semantic functionality.

I didn’t investigate why `sourcekitd` is blocking here.

> If the issue with the test is the crash happening mid-document-update, instead of messing with the timeout, can we change the test to wait on having two PublishDiagnosticsNotifications after the document open like we do in many other tests?

I also thought about preventing this situation to fix the test but decided against it because
a) I couldn’t think of a way how to wait for semantic functionality to be there (didn’t think about the wait for two PublishDiagnostics)
b) honestly I don’t care too much if SourceKit-LSP is waiting for a couple seconds after a crash as long as it restores functionality eventually.

> I'm also suspicious that you were not able to reproduce this issue in clangd, but on the bot it fails consistently every time for all of the clangd crash recovery tests and that we have only seen this issue after the llvm rebranch, so with lots of clangd changes landing. To confirm: were you building the llvm rebranch when testing?

Good thing you ask. I didn’t realize the tests were failing consistently, so I assumed it was just a nondeterministic failure (like the `sourcekitd` one I’m fixing in this PR) that only happened to manifest in `rebranch`. I’ll check the `rebranch` for the `clangd` issue.

benlangmuir:
> I also thought about preventing this situation to fix the test but decided against it because
> a) I couldn’t think of a way how to wait for semantic functionality to be there (didn’t think about the wait for two PublishDiagnostics)

Based on what you said above, the issue is that we're still handling a document update notification that occurs *before* the crash.  What I am suggesting is that when you open the document you wait for the two `PublishDiagnosticsNotifications` to be received, so that they are handled before you trigger the crash.  This is a common pattern - see e.g `testEditing` where when it opens the document it synchronously waits for the diagnostic updates.

ahoppen:
Closing in favor of #366 and #367.

PR #366, opened 2021-02-02
Re-enable clang crash recovery tests

ahoppen:
We now explicitly crash clangd using a SIGKILL instead of relying on it crashing when we add a crash pragma to the source code.

rdar://73717447

### DIFF
--- a/Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
+++ b/Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
@@ -75,6 +75,9 @@ final class ClangLanguageServerShim: LanguageServer, ToolchainLanguageServer {
   /// A callback with which `ClangLanguageServer` can request its owner to reopen all documents in case it has crashed.
   private let reopenDocuments: (ToolchainLanguageServer) -> Void
 
+  /// While `clangd` is running, its PID.
+  private var clangdPid: Int32?
+
   /// Creates a language server for the given client referencing the clang binary at the given path.
   public init(
     client: Connection,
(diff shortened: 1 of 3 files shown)


ahoppen:
https://github.com/apple/llvm-project/pull/2416

@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -137,6 +140,7 @@ final class ClangLanguageServerShim: LanguageServer, ToolchainLanguageServer {
     process.standardInput = usToClangd
     process.terminationHandler = { [weak self] process in
       log("clangd exited: \(process.terminationReason) \(process.terminationStatus)")
+      self?.clangPid = nil
```
benlangmuir, line 143: I think this mutation needs to move to `self.queue.async`.  Maybe something like this (not tested)?

```swift
connectionToClangd.close()
self?.queue.async { [self] in
  self.clangPid = nil
  if process.terminationStatus != 0 {
    self.state = .connectionInterrupted
    self.restartClangd()
  }
}
```

And since I'm looking at it, _crash ought to be `queue` as well.  I think async is fine, since the test is going to need to wait on it coming back anyway.  If we ever need to make a crash function that will interrupt an ongoing operation we could lock the pid variable separately from queue, but it doesn't seem like it matters right now.

ahoppen: I can't really imagine a race here, but sounds good to better be safe and make the crash be polite and patiently wait in the queue to kill the process 😉

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

benlangmuir (approved):

compnerd:
This breaks the build on Windows :(

ahoppen:
@compnerd What's the failure?

compnerd:
@ahoppen, `kill` is not a portable function call.  See #368

PR #367, opened 2021-02-02
Fix a nondeterministic sourcektid crash recovery test failure

ahoppen:
In some cases (especially if the system is under heavy load), we crash `sourcekitd` before having handled the `documentupdate` we received from the open request. If we then perform the 0,0 replacetext synchronously in `handleDocumentUpdate` to receive the semantic diagnostics, `sourcekitd` blocks until it has restored semantic functionality and thus the LSP server's queue is blocked for that entire duration.

To avoid the data race, make sure the document is fully opened before crashing sourcekitd.

### DIFF
--- a/Tests/SourceKitDTests/CrashRecoveryTests.swift
+++ b/Tests/SourceKitDTests/CrashRecoveryTests.swift
@@ -12,6 +12,7 @@
 
 import ISDBTestSupport
 import LanguageServerProtocol
+import LSPLogging
 import SourceKitLSP
 import SourceKitD
 import SKTestSupport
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitDTests/CrashRecoveryTests.swift
```diff
@@ -45,7 +46,20 @@ final class CrashRecoveryTests: XCTestCase {
...
+    ws.sk.handleNextNotification({ (note: LanguageServerProtocol.Notification<PublishDiagnosticsNotification>) in
+      log("Received diagnostics for open - syntactic")
+      documentOpened.fulfill()
+    })
+    ws.sk.appendOneShotNotificationHandler({ (note: LanguageServerProtocol.Notification<PublishDiagnosticsNotification>) in
+      log("Received diagnostics for open - semantic")
+      documentOpened.fulfill()
+    })
     try! ws.openDocument(loc.url, language: .swift)
+    self.wait(for: [documentOpened], timeout: 5)
```
benlangmuir, line 62: We should use a larger base timeout for an AST build since CI tends to be slower than we expect sometimes.  Otherwise LGTM.

ahoppen: Increased the timeout to 10 seconds. Or do you think we need more?

ahoppen:
@swift-ci Please test

PR #370, opened 2021-02-05
[swiftpm] Avoid ambiguous reference to delegate type

benlangmuir:
SwiftPM introduced a type with the same name, so qualify ours to avoid
the conflict.


benlangmuir:
@swift-ci please test

PR #372, opened 2021-03-02
[test] Stop matching error description text that is not from our code

benlangmuir:
The underlying JSON parsing error is from Foundation, so we don't want
to try to compare that in our tests. As a quick fix, just compare a
prefix of the string that we control.


benlangmuir:
@swift-ci please test

PR #375, opened 2021-03-13
Add support for fixed compilation database (compile_flags.txt)

hashemi:
This small change adds support for fixed compilation database (compile_flags.txt), ideal for simple codebases that do not use CMake with an auto-generated compile_commands.json file.

See clangd's [compilation database documentation](https://clang.llvm.org/docs/JSONCompilationDatabase.html) under Alternatives.

@benlangmuir

### DIFF
--- a/Sources/SKCore/CompilationDatabase.swift
+++ b/Sources/SKCore/CompilationDatabase.swift
@@ -87,9 +87,10 @@ public struct FixedCompilationDatabase: CompilationDatabase, Equatable {
   public var allCommands: AnySequence<Command> { AnySequence([]) }
   
   private let fixedArgs: [String]
+  private let directory: String
 
   public subscript(path: URL) -> [Command] {
-    [Command(directory: "", filename: path.path, commandLine: fixedArgs + [path.path])]
+    [Command(directory: directory, filename: path.path, commandLine: fixedArgs + [path.path])]
   }
 }
 
@@ -100,6 +101,7 @@ extension FixedCompilationDatabase {
   }
 
   public init(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+    self.directory = file.dirname
     let bytes = try fileSystem.readFileContents(file)
 
     var fixedArgs: [String] = ["clang"]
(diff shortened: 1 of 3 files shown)


Sources/SKCore/CompilationDatabase.swift
```diff
@@ -68,10 +68,55 @@ public func tryLoadCompilationDatabase(
...
+/// ```
+///
+/// See https://clang.llvm.org/docs/JSONCompilationDatabase.html under Alternatives
+public struct FixedCompilationDatabase: CompilationDatabase, Equatable {
+  public var allCommands: AnySequence<Command> { AnySequence([]) }
+  
+  private let fixedArgs: [String]
+
+  public subscript(path: URL) -> [Command] {
+    [Command(directory: "", filename: path.path, commandLine: fixedArgs + [path.path])]
```
benlangmuir, line 92: We should make Command.directory optional and use `nil` here.

Edit: actually, should we instead store the directory from the path we use when constructing the FixedCompilationDatabase (i.e. the directory that contains compile_commands.txt)?

Tests/SKCoreTests/CompilationDatabaseTests.swift
```diff
@@ -187,6 +187,25 @@ final class CompilationDatabaseTests: XCTestCase {
     XCTAssertNotNil(tryLoadCompilationDatabase(directory: AbsolutePath("/a"), fs))
   }
 
+  func testFixedCompilationDatabase() {
```
benlangmuir, line 190: Please regenerate Linux tests

```
swift build --regenerate-linuxmain
```

If your version of swiftpm is too new and no longer supports that you can add the new test manually to `Tests/SKCoreTests/XCTestManifests.swift`

benlangmuir (changes requested):
@swift-ci please test

benlangmuir:
This looks great, just a couple of minor comments.  Thanks!

hashemi:
Thanks for the review Ben! Made the changes.

benlangmuir:
@swift-ci please test

benlangmuir:
Thanks!

PR #376, opened 2021-03-14
Additions to the LSP module for `workspace/didChangeWatchedFiles`

DavidGoldman:
We can then use this functionality to allow a `BuildSystem` to watch
files (e.g. via the client or even in-process file watching),
which would allow us to do things like detect new files and provide
accurate build settings for them.


Sources/LanguageServerProtocol/SupportTypes/FileSystemWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+    self.globPattern = globPattern
+    self.kind = kind
+  }
+}
+
+/// The type of file event a watcher is interested in.
+///
+/// In LSP, this is an integer, so we don't use a closed set.
+
+// fixme: optionset?
```
DavidGoldman, line 32: LMK if you think it makes sense to swap over

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// FIXME: Find a way to abstract this without deserializing.
```
DavidGoldman, line 15: I'm not sure of a better generalization here, we could do something like what we do for Notifications/Requests but to store it we'd need to type erase it/box it (e.g. an AnyRegistrationOptions)

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/FileEvent.swift
```diff
@@ -0,0 +1,39 @@
...
+
+  public init(uri: DocumentURI, type: FileChangeType) {
+    self.uri = uri
+    self.type = type
+  }
+}
+/// The type of file event.
+///
+/// In LSP, this is an integer, so we don't use a closed set.
+public struct FileChangeType: RawRepresentable, Codable, Hashable {
```
benlangmuir, line 26: Worth adding a test to ensure this is encoded correctly since it needs to use the raw value directly.

Sources/LanguageServerProtocol/SupportTypes/FileSystemWatcher.swift
benlangmuir: Yeah, OptionSet makes sense to me 👍

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -382,7 +382,6 @@ public struct CodeLensOptions: Codable, Hashable {
 }
 
 public struct ExecuteCommandOptions: Codable, Hashable {
-
```
benlangmuir, line 385: Unrelated formatting change?

Sources/LanguageServerProtocol/Requests/RegisterCapabilityRequest.swift
```diff
@@ -0,0 +1,59 @@
...
+/// General parameters to register a capability.
+public struct Registration: Codable, Hashable {
+  /// The id used to register the capability which may be used to unregister support.
+  public var id: String
+
+  /// The method/capability to register for.
+  public var method: String
+
+  /// Options necessary for this registration.
+  public var registerOptions: LSPAny
```
benlangmuir, line 46: I believe this should be optional

```
	/**
	 * Options necessary for the registration.
	 */
	registerOptions?: any;
```

Sources/LanguageServerProtocol/Requests/RegisterCapabilityRequest.swift
```diff
@@ -0,0 +1,59 @@
...
+
+  public init(id: String, method: String, registerOptions: LSPAny) {
+    self.id = id
+    self.method = method
+    self.registerOptions = registerOptions
+  }
+
+  public init(method: String, registerOptions: LSPAny) {
+    self.id = UUID().uuidString
```
benlangmuir, line 55: This deserves some explanation at least in the doc comment - doesn't this prevent you from ever changing the registration?  Is that okay?

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
benlangmuir: There aren't that many of these - is there a reason not to just make `encodeToLSPAny` a protocol requirement and implement it concretely for each concrete type that implements the protocol?

Sources/LanguageServerProtocol/Requests/RegisterCapabilityRequest.swift
DavidGoldman: Not sure what you mean, we randomly generate the id, so you'd need to save it if you want to unregister it.

Sources/LanguageServerProtocol/SupportTypes/FileSystemWatcher.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/SupportTypes/FileEvent.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
DavidGoldman: Reverted

Sources/LanguageServerProtocol/Requests/RegisterCapabilityRequest.swift
DavidGoldman: Done

DavidGoldman:
> There aren't that many of these - is there a reason not to just make encodeToLSPAny a protocol requirement and implement it concretely for each concrete type that implements the protocol?

We could but then we'd probably want to implement it for all of the transitive types embedded there, also note that at the moment these really would only ever be encoded, not decoded since only the server sends them over.

benlangmuir:
> We could but then we'd probably want to implement it for all of the transitive types embedded there, also note that at the moment these really would only ever be encoded, not decoded since only the server sends them over.

This is fine for now as long as we don't have a use case, but we generally expect to be able to write both client and server using this code.

> We could but then we'd probably want to implement it for all of the transitive types embedded there

Is there something I'm missing? The ones added so far seem straightforward/shallow in terms of object graph.

Tests/LanguageServerProtocolTests/LanguageServerProtocolTests.swift
```diff
@@ -14,6 +14,11 @@ import LanguageServerProtocol
 
 import XCTest
 
+fileprivate func AssertDataIsString(_ data: Data, expected: String) {
+  let got = String(decoding: data, as: UTF8.self)
+  XCTAssertEqual(got, expected)
```
benlangmuir, line 19: Should add `file: StaticString = #file, line: UInt = #line` and forward these to `XCTAssertEqual` so that failures are not all attributed to the helper method.

Tests/LanguageServerProtocolTests/LanguageServerProtocolTests.swift
```diff
@@ -33,4 +38,12 @@ final class LanguageServerProtocolTests: XCTestCase {
     XCTAssertEqual(encodedURI, fileURI)
     XCTAssertEqual(encodedURI.hashValue, fileURI.hashValue)
   }
+
+  func testFileChangeTypeEncoding() throws {
```
benlangmuir, line 42: Need to update linux tests for the addition

DavidGoldman:
> > We could but then we'd probably want to implement it for all of the transitive types embedded there, also note that at the moment these really would only ever be encoded, not decoded since only the server sends them over.
> 
> This is fine for now as long as we don't have a use case, but we generally expect to be able to write both client and server using this code.
> 
> > We could but then we'd probably want to implement it for all of the transitive types embedded there
> 
> Is there something I'm missing? The ones added so far seem straightforward/shallow in terms of object graph.

Do we intend to keep LSPAny like it currently is? If so I wonder if it would be more useful to have a generalized encoder/decoder for it. Although that will be a bit more work it would save manually implementing all of these encoders/decoders

benlangmuir:
> Do we intend to keep LSPAny like it currently is?

Is there a good alternative?  We're using it for places where the spec allows `any`.  It would be nice if this was builtin to Foundation so that we could transparently encode/decode between string and LSPAny and between LSPAny and Codable types.

DavidGoldman:
> > Do we intend to keep LSPAny like it currently is?
> 
> Is there a good alternative? We're using it for places where the spec allows `any`. It would be nice if this was builtin to Foundation so that we could transparently encode/decode between string and LSPAny and between LSPAny and Codable types.

Right, if we had a proper `JSON` type that had  Encoder/Decoder we could use that instead of `LSPAny` and `LSPAnyCodable`, I guess I was asking if we planned to keep `LSPAnyCodable`. The simplest thing to do here would be to add an `LSPAnyEncodable` with the `encodeToLSPAny` requirement like you said, but a general fix would be proper coders

benlangmuir:
> I guess I was asking if we planned to keep LSPAnyCodable

`LSPAnyCodable` is a workaround for the lack of proper encoders/decoders, so from that perspective I'd be fine with getting rid of it if we had proper support. That being said, I'm not sure how big or complex the proper encoders/decoders would be to implement.  If proper support reduced the number of lines of code by letting us throw out the `LSPAnyCodable` implementations, it would probably be an obvious win, but if it ends up being bigger than what we have it's harder to justify the maintenance.

DavidGoldman:
> LSPAnyCodable is a workaround for the lack of proper encoders/decoders, so from that perspective I'd be fine with getting rid of it if we had proper support. That being said, I'm not sure how big or complex the proper encoders/decoders would be to implement. If proper support reduced the number of lines of code by letting us throw out the LSPAnyCodable implementations, it would probably be an obvious win, but if it ends up being bigger than what we have it's harder to justify the maintenance.

After looking into the stdlib's JSONEncoder/JSONDecoder which is quite large and complicated, I agree that we should stick with `LSPAnyCodable` for now, I'll swap it over.

Tests/LanguageServerProtocolTests/LanguageServerProtocolTests.swift
DavidGoldman: Done

Tests/LanguageServerProtocolTests/LanguageServerProtocolTests.swift
DavidGoldman: Done

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
```diff
@@ -121,3 +121,52 @@ public protocol LSPAnyCodable {
...
+    self = .some(wrapped)
+  }
+
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    return nil
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    guard let wrapped = self else { return .null }
+    return wrapped.encodeToLSPAny()
```
benlangmuir, line 146: Do we actually use the `null` encoding anywhere, or is this just for completeness?  If possible I strongly prefer skipping nil values instead of encoding them as `null` since that seems to play nicer with other tools and saves space.

Tests/SourceKitDTests/XCTestManifests.swift
```diff
@@ -23,6 +35,7 @@ extension SourceKitDTests {
 
 public func __allTests() -> [XCTestCaseEntry] {
     return [
+        testCase(CrashRecoveryTests.__allTests__CrashRecoveryTests),
```
benlangmuir, line 38: Oops!  Hope these are working on Linux...

benlangmuir (approved):

Tests/SourceKitDTests/XCTestManifests.swift
DavidGoldman: Hmm, CI is failing but those errors link to 

```
/home/buildnode/jenkins/workspace/swift-sourcekit-lsp-PR-Linux/tmp/TemporaryDirectory.YD3VQ4/pkg/Package.swift:3:5: error: type annotation missing in pattern
14:18:28 let pack
```

That seems like an unrelated issue that CI is picking up since it was present and enabled before in `SwiftPMWorkspaceTests.testUnparsablePackage` or perhaps something related to the diagnostics changed?

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
DavidGoldman: Swapped over the place that used it

DavidGoldman:
@swift-ci please test

Tests/SourceKitDTests/XCTestManifests.swift
benlangmuir: It's failing in CrashRecoveryTests.testSourcekitdCrashRecovery, which makes some sense since sourcekitd is in-process on Linux, so that test doesn't really make sense.  We should just skip it on non-Darwin platforms.  Feel free to either do that in this patch (some kind of `try XCTSkipIf(...)`) or else remove the changes related to enabling these tests on Linux and let me or @ahoppen deal with it.

DavidGoldman: Gotcha, done

DavidGoldman:
@swift-ci please test

compnerd:
This broke the Windows bootstrap :-(

DavidGoldman:
Is there any CI or pre-merge hooks for windows support? Or a script to run to update the CMake files or do you have to do it manually?

compnerd:
There're nightly builds at https://dev.azure.com/compnerd/swift-build

The CMake files need to be updated manually.

PR #377, opened 2021-03-17
Add LSP types and server stubs for call hierarchy support

DavidGoldman:
There's some other issues we'll have to handle to add support:

- Need a way to get the initial symbol's kind as well as range
  for the initial `textDocument/prepareCallHierarchy` request.
  This will require modifying the `textDocument/symbolInfo`
  extension and/or forwarding the `textDocument/prepareCallHierarchy`
  to clangd (but we'd need the USR).

- Currently the `CallHierarchyItem` spec provides a range for each
  item but we only have the start location available for us in the
  index. This might be OK, if not we likely will need to change the
  index format to record the end of the range or find another way to
  compute it.


DavidGoldman:
@swift-ci please test

DavidGoldman:
Besides those issues, I think in order to implement the incoming calls, we can do:

[foreachSymbolCallOccurrence](https://github.com/apple/indexstore-db/blob/fce2b363f7e20d02adf00493dbc0d2d49c559d3f/include/IndexStoreDB/Index/IndexSystem.h#L122) (need to expose this in IndexStoreDB's Swift API)-> find calls + their locations -> occurrence.relations (calledBy) to find the caller's `Symbol`. For each Symbol, we need to look up the their definition location via `occurrences(ofUSR:`

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/ClientCapabilities.swift
```diff
@@ -438,6 +448,8 @@ public struct TextDocumentClientCapabilities: Hashable, Codable {
 
   public var foldingRange: FoldingRange? = nil
 
+  public var callHierarchy: CallHierarchy? = nil
```
benlangmuir, line 451: This could use `DynamicRegistrationCapability` instead of adding a specific type for it.

Sources/LanguageServerProtocol/Requests/CallHierarchyPrepareRequest.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+
```
benlangmuir, line 13: Nitpick: extra blank line?

Sources/LanguageServerProtocol/Requests/CallHierarchyIncomingCallsRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// The request is sent from the client to the server to resolve incoming calls for a given call hierarchy item.
+/// It is only issued if a server registers for the `textDocument/prepareCallHierarchy` request.
+public struct CallHierarchyIncomingCallsRequest: RequestType {
```
benlangmuir, line 15: Could we add a brief explanation of what "incoming" means here and similarly for the outgoing one?  Maybe it's just me, but I find "callers"  and "callees" much easier to understand.

Sources/LanguageServerProtocol/Requests/CallHierarchyIncomingCallsRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/// Represents a caller (an incoming call) - an item that makes a call of the original `item`.
+public struct CallHierarchyIncomingCall: ResponseType, Hashable {
+  /// The item that makes the call.
+  public var from: CallHierarchyItem
+
+  /// The ranges at which the calls appear. This is relative to the caller
+  /// denoted by `from`.
+  public var fromRanges: [Range<Position>]
```
benlangmuir, line 33: I haven't looked into this much - is this saying that these ranges are relative to `from.range`?  That's kind of weird.  Also, is it just relative to the start position or does it need to be contained within that range like selectionRange does?

Also, this probably needs to have a similar encoding trick played as we have for `@CustomCodable<PositionRange>`.

Sources/LanguageServerProtocol/SupportTypes/CallHierarchyItem.swift
```diff
@@ -0,0 +1,58 @@
...
+
+  /// More detail for this item, e.g. the signature of a function.
+  public var detail: String?
+
+  /// The resource identifier of this item.
+  public var uri: DocumentURI
+
+  /// The range enclosing this symbol, excluding leading/trailing whitespace
+  /// but including everything else, e.g. comments and code.
+  public var range: Range<Position>
```
benlangmuir, line 29: `@CustomCodable<PositionRange>`

Sources/LanguageServerProtocol/SupportTypes/CallHierarchyItem.swift
```diff
@@ -0,0 +1,58 @@
...
+
+  /// The range that should be selected and revealed when this symbol is being
+  /// picked, e.g. the name of a function. Must be contained by the `range`.
+  public var selectionRange: Range<Position>
```
benlangmuir, line 33: `@CustomCodable<PositionRange>`

Sources/LanguageServerProtocol/Requests/CallHierarchyOutgoingCallsRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/// Represents a callee (an outgoing call) - an item that is called by the original `item`.
+public struct CallHierarchyOutgoingCall: ResponseType, Hashable {
+  /// The item that is called.
+  public var to: CallHierarchyItem
+
+  /// The ranges at which this item is called. This is relative to the caller,
+  /// e.g the item passed to `callHierarchy/outgoingCalls` request.
+  public var fromRanges: [Range<Position>]
```
benlangmuir, line 33: Same comments as the incoming version.

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -96,6 +98,7 @@ public final class SourceKitServer: LanguageServer {
     registerToolchainTextDocumentRequest(SourceKitServer.documentColor, [])
     registerToolchainTextDocumentRequest(SourceKitServer.colorPresentation, [])
     registerToolchainTextDocumentRequest(SourceKitServer.codeAction, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.callHierarchyPrepare, nil)
```
benlangmuir, line 101: I would prefer we don't register these dummy functions until we can actually test the implementation.  Once we resolve the questions I made about the LSP API types, I would be okay with committing those independently as long as any non-trivial serialization gets tests added.  But I think the server changes should wait for an implementation.

Sources/LanguageServerProtocol/SupportTypes/ClientCapabilities.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/Requests/CallHierarchyPrepareRequest.swift
DavidGoldman: Removed

Sources/LanguageServerProtocol/Requests/CallHierarchyIncomingCallsRequest.swift
DavidGoldman: Changed

Sources/LanguageServerProtocol/Requests/CallHierarchyIncomingCallsRequest.swift
DavidGoldman: Reworded and swapped to the customcodable

Sources/SourceKitLSP/SourceKitServer.swift
DavidGoldman: Removed

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -438,7 +438,6 @@ extension SourceKitServer {
   // MARK: - General
 
   func initialize(_ req: Request<InitializeRequest>) {
-
```
benlangmuir, line 441: Unrelated formatting change

Sources/LanguageServerProtocol/PositionRange.swift
```diff
@@ -45,6 +45,41 @@ public struct PositionRange: CustomCodableWrapper {
...
+  fileprivate enum CodingKeys: String, CodingKey {
+    case lowerBound = "start"
+    case upperBound = "end"
+  }
+
+  public init(from decoder: Decoder) throws {
+    var values: [Range<Position>] = []
+    var arrayContainer = try decoder.unkeyedContainer()
+    while !arrayContainer.isAtEnd {
+      let container = try arrayContainer.nestedContainer(keyedBy: CodingKeys.self)
```
benlangmuir, line 65: Can this be simplified to something like
```swift
let range = try arrayContainer.decode(PositionRange.self)
values.append(range.wrappedValue)
```

Incidentally, I think we could do something like

```swift
values.reserveCapacity(arrayContainer.count ?? 1)
```

Sources/LanguageServerProtocol/PositionRange.swift
```diff
@@ -45,6 +45,41 @@ public struct PositionRange: CustomCodableWrapper {
...
+      let rhs = try container.decode(Position.self, forKey: .upperBound)
+      values.append(lhs..<rhs)
+    }
+    self.wrappedValue = values
+  }
+
+  public func encode(to encoder: Encoder) throws {
+    var arrayContainer = encoder.unkeyedContainer()
+    for rangeValue in wrappedValue {
+      var container = arrayContainer.nestedContainer(keyedBy: CodingKeys.self)
```
benlangmuir, line 76: Similarly, I think going through PositionRange for encoding simplifies this and you would no longer need explicit CodingKeys for this type.

benlangmuir (commented):
`PositionRangeArray` could use some tests, and it would be good to also sanity check the encoding/decoding of the types that use it as well.  Should fit in with other CodingTests.

DavidGoldman:
@swift-ci please test

benlangmuir (approved):

PR #380, opened 2021-03-18
Dynamically register completion options for supporting clients

DavidGoldman:
Using dynamic registration (when supported by the client) allows
us to provide different completion options for ObjC and Swift
files.

We should be able to expand this to other capabilities in the future
(e.g. semantic highlighting, execute command support).

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
@@ -80,15 +80,6 @@ public struct CompletionRegistrationOptions: RegistrationOptions, TextDocumentRe
       dict["allCommitCharacters"] = encode(strings: allCommitCharacters)
     }
   }
-
-  public static func == (lhs: CompletionRegistrationOptions, rhs: CompletionRegistrationOptions) -> Bool {
-    return lhs.textDocumentRegistrationOptions == rhs.textDocumentRegistrationOptions && lhs.completionOptions == rhs.completionOptions
-  }
-
-  public func hash(into hasher: inout Hasher) {
-    textDocumentRegistrationOptions.hash(into: &hasher)
-    completionOptions.hash(into: &hasher)
-  }
 }
 
 /// Describe options to be used when registering for file system change events.
--- a/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
@@ -41,14 +41,6 @@ public struct TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
     guard let documentSelector = documentSelector else { return }
     dict["documentSelector"] = documentSelector.encodeToLSPAny()
   }
-
-  public static func == (lhs: TextDocumentRegistrationOptions, rhs: TextDocumentRegistrationOptions) -> Bool {
-    return lhs.documentSelector == rhs.documentSelector
-  }
-
-  public func hash(into hasher: inout Hasher) {
-    documentSelector?.hash(into: &hasher)
-  }
 }
 
 /// Protocol for a type which structurally represents`TextDocumentRegistrationOptions`.
--- a/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
@@ -28,7 +28,7 @@ fileprivate func encode(strings: [String]) -> LSPAny {
 }
 
 /// General text document registration options.
-public class TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
+public struct TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
   /// A document selector to identify the scope of the registration. If not set,
   /// the document selector provided on the client side will be used.
   public var documentSelector: DocumentSelector?
@@ -51,17 +51,25 @@ public class TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
   }
 }
 
+/// Protocol for a type which structurally represents`TextDocumentRegistrationOptions`.
+public protocol TextDocumentRegistrationOptionsProtocol {
+  var textDocumentRegistrationOptions: TextDocumentRegistrationOptions {get}
+}
+
 /// Code completiion registration options.
-public class CompletionRegistrationOptions: TextDocumentRegistrationOptions {
+public struct CompletionRegistrationOptions: RegistrationOptions, TextDocumentRegistrationOptionsProtocol, Hashable {
+  public var textDocumentRegistrationOptions: TextDocumentRegistrationOptions
   public var completionOptions: CompletionOptions
 
   public init(documentSelector: DocumentSelector? = nil, completionOptions: CompletionOptions) {
+    self.textDocumentRegistrationOptions =
+        TextDocumentRegistrationOptions(documentSelector: documentSelector)
     self.completionOptions = completionOptions
-    super.init(documentSelector: documentSelector)
   }
 
-  public override func encodeIntoLSPAny(dict: inout [String: LSPAny]) {
-    super.encodeIntoLSPAny(dict: &dict)
+  public func encodeIntoLSPAny(dict: inout [String: LSPAny]) {
+    textDocumentRegistrationOptions.encodeIntoLSPAny(dict: &dict)
+
     if let resolveProvider = completionOptions.resolveProvider {
       dict["resolveProvider"] = .bool(resolveProvider)
     }
@@ -74,11 +82,11 @@ public class CompletionRegistrationOptions: TextDocumentRegistrationOptions {
   }
 
   public static func == (lhs: CompletionRegistrationOptions, rhs: CompletionRegistrationOptions) -> Bool {
-    return lhs.documentSelector == rhs.documentSelector && lhs.completionOptions == rhs.completionOptions
+    return lhs.textDocumentRegistrationOptions == rhs.textDocumentRegistrationOptions && lhs.completionOptions == rhs.completionOptions
   }
 
-  public override func hash(into hasher: inout Hasher) {
-    super.hash(into: &hasher)
+  public func hash(into hasher: inout Hasher) {
+    textDocumentRegistrationOptions.hash(into: &hasher)
     completionOptions.hash(into: &hasher)
   }
 }
--- a/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
@@ -12,12 +12,23 @@
 
 import Foundation
 
-/// An event describing a file change.
-public protocol RegistrationOptions: Codable, LSPAnyCodable, Hashable {
+/// Protocol for capability registration options, which must be encodable to
+/// `LSPAny` so they can be included in a `Registration`.
+public protocol RegistrationOptions: Hashable {
+  func encodeIntoLSPAny(dict: inout [String: LSPAny])
+}
+
+fileprivate func encode(strings: [String]) -> LSPAny {
+  var values = [LSPAny]()
+  values.reserveCapacity(strings.count)
+  for str in strings {
+    values.append(.string(str))
+  }
+  return .array(values)
 }
 
 /// General text document registration options.
-public struct TextDocumentRegistrationOptions: RegistrationOptions {
+public class TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
   /// A document selector to identify the scope of the registration. If not set,
   /// the document selector provided on the client side will be used.
   public var documentSelector: DocumentSelector?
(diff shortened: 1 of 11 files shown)


DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -12,12 +12,23 @@
...
+  values.reserveCapacity(strings.count)
+  for str in strings {
+    values.append(.string(str))
+  }
+  return .array(values)
 }
 
 /// General text document registration options.
-public struct TextDocumentRegistrationOptions: RegistrationOptions {
+public class TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
```
benlangmuir, line 31: I think making this a class is bad idea.  The most interesting functionality this type provides is serialization, but that creates a foot gun for types that inherit from it, since they need to remember to override the serialization and if they try to call `super.encodeIntoLSPAny(...)` they need to know how the superclass serialization is implemented in order to interop with it correctly.

Also, about half the code in this type and in the types that inherit from it is now from Hashable/Equatable implementations that would be free for a struct.

Finally, it's *surprising* this is a class since none of our other simple data types like this have reference semantics.

If what we need is a type that can be used in a dictionary, I suggest making a protocol for what you need.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,90 @@
...
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+
+/// A class which tracks the client's capabilities as well as our dynamic
+/// capability registrations in order to avoid registering conflicting
+/// capabilities.
+public final class CapabilityRegistry {
+  /// Registered completion options.
+  private var completion: [Registration: CompletionRegistrationOptions] = [:]
```
benlangmuir, line 20: Now that I see it in another context, I'm thinking we should consider renaming `Registration` to something like `CapabilityRegistration`, or even `ClientCapabilityRegistration`.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,90 @@
...
+
+  /// Whether dynamic registration is supported for completion.
+  public var completionSupported: Bool = false
```
benlangmuir, line 23: This name is fairly generic - the "dynamic registration" part is not clear at all without the comment.  How about:

```swift
public var clientHasDynamicCompletionRegistration: Bool { clientCapabilities.completion?.dynamicRegistration == true }
```

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,90 @@
...
+
+  private(set) var clientCapabilities: ClientCapabilities
+
+  public init(clientCapabilities: ClientCapabilities) {
+    self.clientCapabilities = clientCapabilities
+  }
+
+  public func registerCompletion(
```
benlangmuir, line 31: We should clarify the contract here with respect to whether dynamic registration is supported.  Naively I would expect either `precondition(completionSupported)` or `guard completionSupported else { return nil }` or something like that.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,90 @@
...
+/// capability registrations in order to avoid registering conflicting
+/// capabilities.
+public final class CapabilityRegistry {
+  /// Registered completion options.
+  private var completion: [Registration: CompletionRegistrationOptions] = [:]
+
+  /// Whether dynamic registration is supported for completion.
+  public var completionSupported: Bool = false
+
+  private(set) var clientCapabilities: ClientCapabilities
```
benlangmuir, line 25: Is clientCapabilities ever modified? If not maybe we could just make it a let?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -527,7 +548,24 @@ extension SourceKitServer {
...
+    )
+  }
+
+  func registerCapabilities(
+    for server: ServerCapabilities,
+    languages: [Language],
+    registry: CapabilityRegistry
+  ) {
+    guard let completionOptions = server.completionProvider else { return }
+    guard registry.completionSupported && !registry.hasCompletionRegistrations(for: languages) else { return }
```
benlangmuir, line 560: Why does it bail out if there are already registrations? I would have expected we'd check if they're equal.

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -527,7 +548,24 @@ extension SourceKitServer {
...
-    )))
+    )
+  }
+
+  func registerCapabilities(
+    for server: ServerCapabilities,
+    languages: [Language],
+    registry: CapabilityRegistry
+  ) {
+    guard let completionOptions = server.completionProvider else { return }
```
benlangmuir, line 559: The way this method is structured makes it a bit tricky to extend when we eventually want to add other kinds of dynamic capabilities in addition to code-completion.  I wonder if we should move most of this code down to the registry itself (which ultimately is the place that knows what is supported) and just have this method be responsible for sending the message back to the client.

Additionally, have you thought about the tradeoffs between the design you currently have where the registry is simple and single-threaded, but you need to remember to always change the registrations via this API on the server (or else they will not be communicated back to the client), vs. having the registry keep a callback function that it can use to send updates to the client directly?

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
DavidGoldman: The main point of this type was to make it easier to introduce types that extend `TextDocumentRegistrationOptions`, inheritance it simplest way to do this in Swift but I can also instead make the sub-types store a `TextDocumentRegistrationOptions` and manually call the methods. I'm not sure how a protocol would help with that. WDYT?

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Sure I'll probably go with CapabilityRegistration... ClientCapabilityRegistration is technically confusing since we're registering server capabilities dynamically

Sources/SourceKitLSP/SourceKitServer.swift
DavidGoldman: If we have a conflicting registration for completion there's not much we can do unless we try to manually merge them - I guess I could just log it if that happens?

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
benlangmuir: SGTM.  My comment about the protocol was due to you using `[Registration: TextDocumentRegistrationOptions]`, but if you don't need dynamic behaviour there great.

DavidGoldman: Ah I see, yeah a protocol would be useful for that but for now we don't need it.

Sources/SourceKitLSP/SourceKitServer.swift
DavidGoldman: Yeah that seems like it would be good a idea, either using a protocol or callback. Depending on the spec we might need to unregister capabilities before registering new ones, I opened this [issue](https://github.com/microsoft/language-server-protocol/issues/1227) about that but it hasn't gotten a response, will likely need to test it out with VS Code to see how VS Code behaves.

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
DavidGoldman: Actually, going with a protocol since it will be useful for semantic highlighting

DavidGoldman: Done

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Renamed

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Done

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Now this is done inside the Registry

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Done

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -26,21 +37,57 @@ public struct TextDocumentRegistrationOptions: RegistrationOptions {
...
-  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
-    guard let selectorValue = dictionary[CodingKeys.documentSelector.stringValue] else {
-      self.documentSelector = nil
-      return
+  public func encodeIntoLSPAny(dict: inout [String: LSPAny]) {
+    guard let documentSelector = documentSelector else { return }
+    dict["documentSelector"] = documentSelector.encodeToLSPAny()
+  }
+
+  public static func == (lhs: TextDocumentRegistrationOptions, rhs: TextDocumentRegistrationOptions) -> Bool {
```
benlangmuir, line 45: You should be able to use the default synthesized implementations for `==` and `hash(into:)` now, right?

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -26,21 +37,57 @@ public struct TextDocumentRegistrationOptions: RegistrationOptions {
...
+    if let allCommitCharacters = completionOptions.allCommitCharacters {
+      dict["allCommitCharacters"] = encode(strings: allCommitCharacters)
     }
-    self.documentSelector = documentSelector
   }
 
-  public func encodeToLSPAny() -> LSPAny {
-    guard let documentSelector = documentSelector else { return .dictionary([:]) }
-    return .dictionary([CodingKeys.documentSelector.stringValue: documentSelector.encodeToLSPAny()])
+  public static func == (lhs: CompletionRegistrationOptions, rhs: CompletionRegistrationOptions) -> Bool {
```
benlangmuir, line 84: Similarly, this should be able to use the synthesized `==` and `hash(into:)` now, right?

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+
+/// Handler responsible for registering a capability with the client.
+public typealias RegistrationHandler = (_ registration: CapabilityRegistration) -> Void
```
benlangmuir, line 16: Maybe `ClientRegistrationHandler`?

Nitpick: I don't think the "_ registration:" adds anything to this typealias.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -0,0 +1,103 @@
...
+    clientCapabilities.textDocument?.completion?.dynamicRegistration == true
+  }
+
+  /// Dynamically register completion capabilities if the client supports it and
+  /// we haven't yet registered any completion capabilities for the given
+  /// languages.
+  public func registerCompletionIfNeeded(
+    options: CompletionOptions,
+    for languages: [Language],
+    handler: RegistrationHandler
```
benlangmuir, line 41: for "handler", how about the more active `registerOnClient`?

Sources/SourceKitLSP/SourceKitServer.swift
benlangmuir: Looks good, thanks!

Sources/SourceKitLSP/SourceKitServer.swift
benlangmuir: Yeah, how about we log a warning here unless the options are identical to what's already registered for that language.

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Done, much better

Sources/SourceKitLSP/CapabilityRegistry.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
DavidGoldman: Done

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
DavidGoldman: Yup done

DavidGoldman:
@swift-ci please test

DavidGoldman:
@swift-ci please test

benlangmuir (approved):

PR #382, opened 2021-03-19
Globally ignore SIGPIPE on Linux

ahoppen:
We receive a `SIGPIPE` if we write to a pipe that points to a crashed process. This in particular happens if the target of a `JSONRPCConnection` has crashed and we try to send it a message.
On Darwin, `DispatchIO` ignores `SIGPIPE` for the pipes handled by it, but that features is not available on Linux.
Instead, globally ignore `SIGPIPE` on Linux to prevent us from crashing if the `JSONRPCConnection`'s target crashes.

Fixes rdar://75580936

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/DisableSigpipe.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/DisableSigpipe.swift
@@ -1,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+
+#if canImport(Glibc)
+import Glibc
+#endif
+
+#if os(Linux) || os(Android)
+// This is a lazily initialised global variable that when read for the first time, will ignore SIGPIPE.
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocolJSONRPC/DisableSigpipe.swift
```diff
@@ -0,0 +1,33 @@
...
+//===----------------------------------------------------------------------===//
+
+
+#if canImport(Glibc)
+import Glibc
+#endif
+
+#if os(Linux) || os(Android)
+// This is a lazily initialised global variable that when read for the first time, will ignore SIGPIPE.
+private let globallyIgnoredSIGPIPE: Bool = {
```
benlangmuir, line 20: This feels overcomplicated.  Just put this as a static variable inside JSONRPCConnection and access it from the initializer.  We don't need a public API for this.

ahoppen: My line of thinking was that since this is a global property of the process, it should also be toggled by a global function. And I don’t think this really belongs in `JSONRPCConnection` either. `JSONRCPConnection` just requests `SIGPIPE` to be ignored, it shouldn’t need to worry single dispatch etc.

(also: I just copy-pased this code as-is from `swift-nio`)

benlangmuir: Let's make it an internal function at least.

ahoppen: Sounds like a good compromise 👍

ahoppen:
Made `globallyDisableSigpipe` an `internal` function.

@swift-ci Please test

benlangmuir (approved):

PR #386, opened 2021-04-05
Fix lifetime issues in tests

benlangmuir:
A number of tests were failing with the -Onone lifetime changes.
Regardless of what happens with that change, I'd like to keep our tests
passing with the stricter rules since we also test in release builds.


benlangmuir:
@swift-ci please test

PR #387, opened 2021-04-06
[sksupport] Remove unused old TSC argument parsing code

benlangmuir:
Fixes deprecation warnings while building sourcekit-lsp. We haven't used
this code since we switched to swift-argument-parser, so just delete it.


benlangmuir:
@swift-ci please test

PR #389, opened 2021-05-01
Add support for diagnostic tags

fwcd:
This PR adds a prototypical implementation of the new diagnostic tags, which are part of LSP since 3.15.0, for Swift:

```typescript
export interface Diagnostic {
  ...
  /**
   * Additional metadata about the diagnostic.
   *
   * @since 3.15.0
   */
  tags?: DiagnosticTag[];
  ...
}

export namespace DiagnosticTag {
  /**
   * Unused or unnecessary code.
   *
   * Clients are allowed to render diagnostics with this tag faded out
   * instead of having an error squiggle.
   */
  export const Unnecessary: 1 = 1;
  /**
   * Deprecated or obsolete code.
   *
   * Clients are allowed to rendered diagnostics with this tag strike through.
   */
  export const Deprecated: 2 = 2;
}
```

These tags are used by editors like VSCode to style unused or deprecated diagnostics separately:

image

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/Diagnostic.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/Diagnostic.swift
@@ -61,6 +61,9 @@ public struct Diagnostic: Codable, Hashable {
   /// Related diagnostic notes.
   public var relatedInformation: [DiagnosticRelatedInformation]?
 
+  /// Additional metadata about the diagnostic.
+  public var tags: [DiagnosticTag]?
+
   /// All the code actions that address this diagnostic.
   /// **LSP Extension from clangd**.
   public var codeActions: [CodeAction]?
@@ -72,6 +75,7 @@ public struct Diagnostic: Codable, Hashable {
     codeDescription: CodeDescription? = nil,
     source: String?,
     message: String,
+    tags: [DiagnosticTag]? = nil,
     relatedInformation: [DiagnosticRelatedInformation]? = nil,
     codeActions: [CodeAction]? = nil)
   {
@@ -81,9 +85,17 @@ public struct Diagnostic: Codable, Hashable {
     self.codeDescription = codeDescription
     self.source = source
     self.message = message
+    self.tags = tags
     self.relatedInformation = relatedInformation
     self.codeActions = codeActions
   }
+}
+
+/// A small piece of metadata about a diagnostic that lets editors e.g. style the diagnostic
+/// in a special way.
+public enum DiagnosticTag: Int, Codable, Hashable {
+  case unnecessary = 1
+  case deprecated = 2
 }
 
 /// A 'note' diagnostic attached to a primary diagonstic that provides additional information.
--- a/Sources/LanguageServerProtocol/SupportTypes/Diagnostic.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/Diagnostic.swift
@@ -93,9 +93,15 @@ public struct Diagnostic: Codable, Hashable {
 
 /// A small piece of metadata about a diagnostic that lets editors e.g. style the diagnostic
 /// in a special way.
-public enum DiagnosticTag: Int, Codable, Hashable {
-  case unnecessary = 1
-  case deprecated = 2
+public struct DiagnosticTag: RawRepresentable, Codable, Hashable {
+  public var rawValue: Int
+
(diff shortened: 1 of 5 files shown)


tami5:
😍

benlangmuir (changes requested):
Hi @fwcd,

Thanks for the PR!

> Since the SourceKit protocol does not expose the corresponding metadata as part of its diagnostics yet as far as I can tell, this implementation currently parses the human-readable message using a simple heuristic.

I would prefer not to use this heuristic, since it can easily be thrown off e.g. by the name of a method including the word "unused".  Here's a way I think we could support this in an accurate way:

* In sourcekitd, add a new field to the diagnostic that exposes the diagnostic ID as a string or UID, using the names from the Diagnostic*.def files in Swift (e.g. `availability_deprecated`).
* In sourcekit-lsp, keep a set of the diagnostic IDs that correspond to deprecation and/or unused declarations.

Sources/LanguageServerProtocol/SupportTypes/Diagnostic.swift
```diff
@@ -81,11 +85,19 @@ public struct Diagnostic: Codable, Hashable {
...
     self.message = message
+    self.tags = tags
     self.relatedInformation = relatedInformation
     self.codeActions = codeActions
   }
 }
 
+/// A small piece of metadata about a diagnostic that lets editors e.g. style the diagnostic
+/// in a special way.
+public enum DiagnosticTag: Int, Codable, Hashable {
```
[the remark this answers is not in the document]
fwcd: Good point, thanks!

fwcd:
Good idea, thanks! The only thing I'd be worried about is that IDs don't have a 'stable' value and may shift once someone adds a new diagnostic: https://github.com/apple/swift/blob/7123d2614b5f222d03b3762cb110d27a9dd98e24/lib/AST/DiagnosticList.cpp#L20-L23

Therefore perhaps it would be better to determine deprecation/unusedness on the SourceKit-side and e.g. expose them als boolean flags or an integer enum similar to the one LSP uses?

benlangmuir:
Technically the string names are not guaranteed to stay the same either I think, but we can just update sourcekit-lsp as necessary if they ever change.  I doubt we would ever have an "incompatible" change where a previous diagnostic name was later reused for something unrelated.

fwcd:
The diagnostic tags are now computed using the new categories exposed by SourceKit.

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -650,6 +650,7 @@ final class LocalSwiftTests: XCTestCase {
           code: nil,
           source: "sourcekitd",
           message: "initialization of immutable value \'a\' was never used; consider replacing with assignment to \'_\' or removing it",
+          tags: [.unnecessary],
```
benlangmuir, line 653: This is going to make the test fail using the currently released swift.  While we can't always guarantee sourcekit-lsp main branch will work with previously released toolchains, we do make an effort to keep it working when we can.

Since this particular test isn't about the diagnostic itself, I suggest changing this to

```swift
    // Diagnostic returned by code actions cannot be recursive.
    var expectedDiagnostic = diagnostic!
    expectedDiagnostic.codeActions = nil

    // Expected Fix-it: Replace `let a` with `_` because it's never used
    let expectedTextEdit = TextEdit(range: Position(line: 1, utf16index: 2)..<Position(line: 1, utf16index: 7), newText: "_")
    XCTAssertEqual(fixit, CodeAction(
      title: "Replace 'let a' with '_'",
      kind: .quickFix,
      diagnostics: [expectedDiagnostic],
...
```

Unfortunately this means we don't have a good way to test this new DiagnosticTag functionality, since it depends on the new sourcekitd.  For a more complex change we would probably need to teach the test to detect whether sourcekitd has the new feature or not and skip the test if it's not available.  In this particular case, detecting the presence of the new sourcekitd functionality in the test is almost as much work as the change itself here.  Since the sourcekit-lsp side of this change is relatively simple I'm okay with us not testing it for now.

fwcd: Okay, I have applied the suggestion :+1:

benlangmuir:
@swift-ci please test

benlangmuir (approved):

PR #390, opened 2021-05-05
Update Package.swift dependencies for release/5.5

benlangmuir:
This points our dependencies on swifptm/tsc/indexstore-db to use the
corresponding release/5.5 branch. This affects building sourcekit-lsp as
a normal package, but has no impact on how Swift.org toolchains are
built, since they use the locally checked out sources, which were
already release/5.5, and  not the remote dependencies.


benlangmuir:
@swift-ci please test

akyrtzi (approved):

PR #393, opened 2021-05-13
Set revealOutputChannelOn to Never in the VSCode extension

fwcd:
The VSCode terminal becomes hard to use if the language server sends a bunch of error messages in quick succession, as the editor keeps switching over to the `Output` tab automatically.

Setting `revealOutputChannelOn` to `Never` disables this behavior. Note that the user can still manually switch to the `Output` to view logs and error messages from the language server.

### DIFF
--- a/Editors/vscode/src/extension.ts
+++ b/Editors/vscode/src/extension.ts
@@ -26,7 +26,8 @@ export function activate(context: vscode.ExtensionContext) {
             'objective-c',
             'objective-cpp'
         ],
-        synchronize: undefined
+        synchronize: undefined,
+        revealOutputChannelOn: langclient.RevealOutputChannelOn.Never
     };
 
     const client = new langclient.LanguageClient('sourcekit-lsp', 'SourceKit Language Server', serverOptions, clientOptions);


benlangmuir:
> if the language server sends a bunch of error messages in quick succession

What were the errors about? Is there something else we could fix here?

benlangmuir:
@swift-ci please test

fwcd:
The issue I ran into was that deleting the `.build` directory in a project (to perform a clean build) caused the language server to crash:

```
sourcekit-lsp: /home/build-user/swift/lib/Sema/TypeCheckStmt.cpp:491: bool typeCheckConditionForStatement(swift::LabeledConditionalStmt *, swift::DeclContext *): Assertion `!E->getType() && "the bool condition is already type checked"' failed.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Stack dump:
0.	While performing new completion
1.	While evaluating request CodeCompletionSecondPassRequest(source_file "...", )
2.	While evaluating request TypeCheckASTNodeAtLocRequest(0x7f2b1c372450 AbstractFunctionDecl name=handle(message:from:) : (no type set), )
3.	While type-checking statement at ...
[Info  - 6:24:37 PM] Connection to server got closed. Server will restart.
```

Not sure if this really fixable, as the compiler might rely on the user not changing any build files while running.

benlangmuir:
> Not sure if this really fixable, as the compiler might rely on the user not changing any build files while running.

There may be some cases with clang modules and header changes that are hard to fix in the current setup, but other than that I would expect us not to crash.  The specific case you mentioned doesn't sound like that though, so worth a bug report if you can reproduce it.

PR #394, opened 2021-05-20
Disable clangd crash recovery tests on Linux

benlangmuir:
Until we can fix whatever is sporadically crashing in CI.

rdar://78035044


benlangmuir:
@swift-ci please test

PR #395, opened 2021-05-24
[logging] Don't pass arbitrary string to NSLog

benlangmuir:
NSLog expects a format string, so don't pass it arbitrary log strings,
which may contain format specifiers.

Thanks @compnerd and @gwynne for reporting.


benlangmuir:
@swift-ci please test

compnerd (approved):

PR #396, opened 2021-05-24
SourceKitLSP: properly handle paths on Windows

compnerd:
This corrects the path identification to include Windows paths properly.
Although this is not entirely ideal due to the platform specific code,
but avoids crashing the SourceKit LSP server if it encounters a space in
the path.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #398, opened 2021-05-24
[logging] Don't pass arbitrary string to NSLog

compnerd:
NSLog expects a format string, so don't pass it arbitrary log strings,
which may contain format specifiers.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #399, opened 2021-05-25
SwiftLanguageServer: correct path detection on Windows

compnerd:
Correct the polarity of the check.  The path being a UNC path or having a drive letter indicates that it *is* a path, as opposed to the inverted representation that was accidentally committed.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #401, opened 2021-05-25
[swiftpm] Add derived sources to module compiler arguments

benlangmuir:
Now that swiftpm has derived sources, for example the resource bundle
accessor generated for `Bundle.module`, we need to ensure they get added
to compiler arguments for the module as seen by sourcekit-lsp. Luckily
there is now a convenient ".sources" property on the target description.

https://bugs.swift.org/browse/SR-14658


benlangmuir:
@swift-ci please test

PR #403, opened 2021-05-26
List dependencies to build SourceKit-LSP in the Linux section 

ahoppen:
They were previously only listed in the Docker section, but they apply to general Linux equally.

Also add a comment that modifying your `PATH` is really necessary and that you shouldn’t try to be clever and skip that step.

Resolves rdar://77853086 [SR-14614]

### DIFF
--- a/Documentation/Development.md
+++ b/Documentation/Development.md
@@ -44,6 +44,8 @@ $ swift package update
 $ swift build -Xcxx -I<path_to_swift_toolchain>/usr/lib/swift -Xcxx -I<path_to_swift_toolchain>/usr/lib/swift/Block
 ```
 
+Setting `PATH` as described above is important even if `<path_to_swift_toolchain>/usr/bin` is already in your `PATH` because `/usr/bin` must be the **first** path to search.
+
 After building, the server will be located at `.build/debug/sourcekit-lsp`, or a similar path, if you passed any custom options to `swift build`. Editors will generally need to be provided with this path in order to run the newly built server - see [Editors](../Editors) for more information about configuration.
 
 SourceKit-LSP is designed to build against the latest SwiftPM, so if you run into any issue make sure you have the most up-to-date dependencies by running `swift package update`.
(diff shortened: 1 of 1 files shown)


Documentation/Development.md
```diff
@@ -38,25 +38,25 @@ $ swift build
...
+
+* libsqlite3-dev, libncurses5-dev, python, ninja-build
+
 ```sh
 $ export PATH="<path_to_swift_toolchain>/usr/bin:${PATH}"
 $ swift package update
 $ swift build -Xcxx -I<path_to_swift_toolchain>/usr/lib/swift -Xcxx -I<path_to_swift_toolchain>/usr/lib/swift/Block
 ```
 
+Setting `PATH` as described above is important even if `<path_to_swift_toolchain>/usr/bin` is already in your `PATH` because `/usr/bin` must be the **first** path to search.
```
benlangmuir, line 51: > because `/usr/bin` must be the **first** path to search.

Why would it need to be first? Obviously if you have more than one toolchain on your path you need the right one to come before the wrong one, but that's just normal PATH behaviour.

ahoppen: The problem is that in the docker container, we have the following `PATH`: 
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

`sourcekit-lsp` finds `libsourcekitdInProc` and `libIndexStore` relative to `/usr/sbin` in `/usr/sbin/../lib/` (aka. `/usr/lib`). Thus `/usr/sbin` is not discarded and it becomes the default toolchain. But it doesn’t contain the swift (or clang) executable, causing all sorts of later failures. If `/usr/bin` is before `/usr/sbin` in `PATH`, we don’t hit this issue.

benlangmuir: Okay, the note is fine for now, but this seems like something we should fix.  We could recognize when one toolchain is a strict superset of another one (sbin and bin have the same libraries but only bin has the executables).

benlangmuir:
@swift-ci please test

PR #405, opened 2021-05-27
[swiftpm] For development, build a local copy of PackageDescription

benlangmuir:
When developing sourcekit-lsp, it is useful to be able to use a released
version of the swift toolchain. However, because we are using the latest
branch version of libSwiftPM, it may be that the PackageDescription
module packaged with the toolchain is incompatible with that libSwiftPM.
To prevent that problem, we now build our own copy of PackageDescription
and using that instead of the toolchain's. That change is made by
default to allow plain `swift build` and `swift test` invocations to
work without modification. When building as part of the swift toolchain,
we continue to use the toolchain's copy of PackageDescription, because
we know that it is built from the same sources as we are using for
libSwiftPM.

One situation that is not explicitly covered by the above is if
sourcekit-lsp is installed somewhere outside of a swift toolchain. In
that case by default it will try to find a local package description
module next to the sourcekit-lsp binary, but assuming that fails it will
fallback to searching the default toolcahin that it discovers at
runtime, as it does today. If this usecase turns out to be important we
can provide additional ways to control how to find the
PackageDescription module.

rdar://78548287

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -83,9 +83,14 @@ let package = Package(
           "BuildServerProtocol",
           "LanguageServerProtocol",
           "SKCore",
-          .product(name: "SwiftPM-auto", package: "SwiftPM")
-        ],
-        exclude: ["CMakeLists.txt"]),
+          .product(name: "SwiftPM-auto", package: "SwiftPM"),
+        ] + (useLocalPackageDescriptionModule() ? [
+          .product(name: "PackageDescription", package: "SwiftPM"),
+        ] : []),
+        exclude: ["CMakeLists.txt"],
+        swiftSettings: useLocalPackageDescriptionModule() ? [
+          .define("USE_LOCAL_PACKAGE_DESCRIPTION_MODULE")
+        ] : []),
 
       .testTarget(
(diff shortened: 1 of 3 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
Marking this a draft until I can test both development builds and and installed toolchain on both Linux and macOS.

I'd appreciate feedback on this approach, in particular
* Are the module paths I'm checking for correct for all platforms (@compnerd will this work on Windows?)
* I'm changing the default behaviour of `swift build` and `swift test` rather than making building PackageDescription opt-in. To me it's important that `swift test` is as simple as possible.  It will still fallback to the existing path if it cannot find PackageDescription relative to the sourcekit-lsp binary.
* Is specifying `binDir` an acceptable use of the libSwiftPM API?  This is what swiftpm itself does during testing, but obviously what a project does for their own use is different from what they want to support.  If there was a better way to customize the PackageDescription, I'd be happy to adopt that instead.

compnerd:
I'll try to get a listing of the build tree.

ahoppen:
This change has become obsolete.

PR #406, opened 2021-05-28
Add server-side support for inlay hints using CollectExpressionType

fwcd:
This is an early-stage draft of the [proposed](https://summerofcode.withgoogle.com/projects/#5174278531579904) inlay type hints for SourceKit-LSP.

Inlay hints let LSP clients/editors dynamically display inline type annotations, making it easier to read and understand Swift code, without requiring the programmer to spell out every type.

For this purpose, a new, non-standard request is introduced (`sourcekit-lsp/inlayHints`) that currently requires custom support on the client side. Note that this does not affect compatibility with 'vanilla' LSP clients as inlay hints follow a pull model, i.e. the client has to initiate the request.

Since inlay hints are a proposed LSP feature, however, this implementation strives to conform to the proposed request structures as faithfully as possible to make the migration easy once the feature becomes official. More details can be found here:

- https://github.com/microsoft/language-server-protocol/issues/956 (associated issue)
- https://github.com/microsoft/language-server-protocol/pull/1249 (current proposal)
- https://github.com/microsoft/vscode-languageserver-node/pull/609 (slightly older proposal, follows `rust-analyzer` implementation)

Current progress:

- [x] Add `InlayHint` and support structures on the SourceKit-LSP side
- [x] Implement request and handler on the SourceKit-LSP side
- [x] Add unit tests for the request on the SourceKit-LSP side

Next steps:

- Implement a new request on the SourceKit-side specifically for inlay hints and use it (instead of the current `DocumentSymbols` + `CollectExpressionType` combination)

Once possible:

- Migrate to the official LSP API
- Add the corresponding server/client capabilities

See also:

- #407 for VSCode support


Sources/LanguageServerProtocol/Requests/InlayHintsRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// Request for inline annotations to be displayed in the editor.
+///
+/// - Parameters:
+///   - textDocument: The document for which to provide the inlay hints.
+///
+/// - Returns: InlayHints for the entire document
+public struct InlayHintsRequest: TextDocumentRequest, Hashable {
```
benlangmuir, line 19: Please add `LSP Extension` to the doc comment -- see `SymbolInfoRequest` for an example.  If there's a stable place to link to for information about the proposed spec, that might also make sense to add.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -1037,6 +1037,34 @@ extension SwiftLanguageServer {
     completion(.success(codeActions))
   }
 
+  public func inlayHints(_ req: Request<InlayHintsRequest>) {
+    let uri = req.params.textDocument.uri
+    expressionTypeInfos(uri) { result in
+      switch result {
+      case .success(let infos):
+        let hints = infos
```
benlangmuir, line 1045: Probably worth doing `.lazy.map` to avoid the intermediate array since this could be a large array

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,106 @@
...
+  var printedType: String
+
+  init?(_ dict: SKDResponseDictionary, in snapshot: DocumentSnapshot) {
+    let keys = dict.sourcekitd.keys
+
+    guard let offset: Int = dict[keys.expression_offset],
+          let length: Int = dict[keys.expression_length],
+          let startIndex = snapshot.positionOf(utf8Offset: offset),
+          let endIndex = snapshot.positionOf(utf8Offset: offset + length),
+          let printedType: String = dict[keys.expression_type] else { return nil }
```
benlangmuir, line 33: Style nitpick: the return should be on a new line.

Sources/LanguageServerProtocol/Requests/InlayHintsRequest.swift
```diff
@@ -0,0 +1,34 @@
+//===----------------------------------------------------------------------===//
```
benlangmuir, line 1: Please add any new files to CMakeLists.txt -- don't worry if you aren't setup to do a cmake build to verify it, but usually just adding the files is enough.

Sources/LanguageServerProtocol/Requests/InlayHintsRequest.swift
fwcd: Sure, I've linked the PR and noted the commit of the current proposal state.

Sources/LanguageServerProtocol/Requests/InlayHintsRequest.swift
```diff
@@ -0,0 +1,37 @@
...
+  public var textDocument: TextDocumentIdentifier
+
+  /// The range the inlay hints are requested for. If unset,
+  /// hints for the entire document are requested.
+  @CustomCodable<PositionRange?>
+  public var range: Range<Position>?
+
+  /// The categories of hints that are interesting to the client
+  /// and should be filtered.
+  public var only: [InlayHintCategory]?
```
benlangmuir, line 36: `InlayHintsRequest` should have a public memberwise initializer.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -490,6 +490,13 @@ extension ClangLanguageServerShim {
     forwardRequestToClangdOnQueue(req)
   }
 
+  func inlayHints(_ req: Request<InlayHintsRequest>) {
+    // Currently a Swift-specific, non-standard request.
```
benlangmuir, line 494: clangd has its own inlineHints extension that I would assume matches the proposal fairly exactly (since it's a clangd developer who wrote the proposal).  Should we forward this request as `clangd/inlayHints`?

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      guard let skExpressionTypeInfos: SKDResponseArray = dict[keys.expression_type_list] else {
+        return completion(.success([]))
+      }
+
+      var expressionTypeInfos: [ExpressionTypeInfo] = []
+
+      skExpressionTypeInfos.forEach { (_, skExpressionTypeInfo) -> Bool in
+        guard let info = ExpressionTypeInfo(skExpressionTypeInfo, in: snapshot) else { return false }
```
ahoppen, line 83: I don’t think we should be silently dropping expression type info that fail to deserialize. That makes it hard to debug. I’d propose to add an `assert(false)` to the failure case. That way we’re failing in debug builds but are not tearing the server down in release builds.

A different option would be to report a failure if the SourceKit response fails to deserialize, but I’d vote for the `assert` since it’s an internal inconsistency that should never happen and therefore shouldn’t be exposed to the user.

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,108 @@
...
+  case responseError(ResponseError)
+}
+
+extension SwiftLanguageServer {
+  /// Must be called on self.queue.
+  func _expressionTypeInfos(
+    _ uri: DocumentURI,
+    _ completion: @escaping (Swift.Result<[ExpressionTypeInfo], ExpressionTypeInfoError>) -> Void
+  ) {
+    guard let snapshot = documentManager.latestSnapshot(uri) else {
```
ahoppen, line 56: IIUC `_expressionTypeInfos` is always expected to be called on `queue`. I think it would be good to make that assumption explicit with a `dispatchPrecondition` like [here](https://github.com/apple/sourcekit-lsp/blob/09da6a4f961720b81687567ce8afea633cbd0f44/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L123-L126). And I know we’re not doing it for the other requests.

ahoppen (commented):
Looking good to me overall so far, I’ve put a few comments inline. 

I think you’re also still missing the client and server capabilities for the request and I didn’t see it on your to-do list. Are you planning to do that in the future?

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
fwcd: Yes that would be nice, `clangd` currently implements a slightly-older version of the proposal though that isn't strictly compatible from what I can tell: https://github.com/llvm/llvm-project/blob/316da54/clang-tools-extra/clangd/Protocol.h#L1488-L1527

Therefore I wasn't sure whether temporarily adding a custom request for `clangd` would be worth the effort currently. WDYT?

fwcd:
Adding the server/client capabilities is a good point, though I think this would first become relevant as soon as the inlay hints get upstreamed: Clients that aren't currently interested in inlay hints would simply not send the request and those that are interested already have some special logic on the client-side for that.

Once they are added to the official spec though, capabilities would become useful to the client for negotiating with the (from its POV arbitrary) language server whether it supports the request.

I have added it to the todo list.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -490,6 +490,13 @@ extension ClangLanguageServerShim {
...
+    // Once inlay hints have been upstreamed to LSP, we
+    // can forward them to clangd.
+    req.reply(.success([]))
```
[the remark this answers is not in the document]
benlangmuir: Errors tend to mean something unexpected happened and may be noisy in various editor.  If we don't forward the request to clangd, returning `[]` seems best to me.

fwcd:
image

I've got an early-stage demo of inlay hints to work, although the results are still very noisy with every expression type being highlighted.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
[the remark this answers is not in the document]
ahoppen: That makes sense. Thanks for clarifying.

Editors/vscode/src/inlayHints.ts
```diff
@@ -0,0 +1,202 @@
...
+        editor.setDecorations(hintStyle.decorationType, decorations);
+    }
+
+    private hintsToDecorations(hints: InlayHint[]): vscode.DecorationOptions[] {
+        const converter = this.client.protocol2CodeConverter;
+        return hints.map(h => hintStyle.makeDecoration(h, converter));
+    }
+
+    private async fetchHints(file: SourceFile): Promise<InlayHint[]> {
+        file.inFlightInlayHints?.cancel();
```
ahoppen, line 177: This means, we can‘t currently fetch inlay hints for two files concurrently, right? Wouldn’t it be better to make `inFlightInlayHints` be a map `URI -> cancellation token` and only cancel the request for the file that we are now requesting new hints for?

It’s not really relevant yet since SourceKit supports neither cancellation nor concurrent requests, but it would be nice to be on the safe side now already.

Editors/vscode/src/inlayHints.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private get visibleSourceKitLSPEditors(): vscode.TextEditor[] {
+        return vscode.window.visibleTextEditors.filter(e => this.isSourceKitLSPDocument(e.document));
+    }
+
+    private isSourceKitLSPDocument(document: vscode.TextDocument): boolean {
+        // TODO: Add other SourceKit-LSP languages if/once we forward inlay
+        // hint requests to clangd.
+        return document.languageId === 'swift' && document.uri.scheme === 'file';
```
ahoppen, line 164: Disclaimer: I am not at all familiar with how VSCode handles open documents and language servers.

Couldn’t the user in theory select some other language server for Swift and use SourceKit-LSP for C-family code only? I’m not sure if the Swift files would then be reported as part of the visible documents, but we might need a more sophisticated way of determining whether we are providing language server functionality for that file in the future.

This is fine for now though.

Editors/vscode/src/inlayHints.ts
```diff
@@ -0,0 +1,202 @@
...
+    private onDidChangeTextDocument(event: vscode.TextDocumentChangeEvent): void {
+        if (event.contentChanges.length !== 0 && this.isSourceKitLSPDocument(event.document)) {
+            this.syncCacheAndRenderHints();
+        }
+    }
+
+    private syncCacheAndRenderHints(): void {
+        this.sourceFiles.forEach(async (file, uri) => {
+            const hints = await this.fetchHints(file);
+            if (!hints) return;
```
ahoppen, line 144: How would we end up in the case that `!hints`. IIUC `fetchHints` is returning `InlayHint[]` and JavaScript (thus probably also TypeScript) considers the empty array as boolean `true`.

ahoppen, line 144: How would we end up in the case that `!hints`. IIUC `fetchHints` is returning `InlayHint[]` and JavaScript (thus probably also TypeScript) considers the empty array as boolean `true`.

Editors/vscode/src/inlayHints.ts
fwcd: Yes that's true, it's more of a heuristic currently. Once LSP includes inlay hints, it won't be needed anymore anyway, since VSCode would internally handle inlay hint requests and make sure they get sent to the correct language server.

Editors/vscode/src/inlayHints.ts
fwcd: Good point!

fwcd:
image

The handler now uses document symbols to figure out the positions of variable bindings and only renders the inlay hints there. While this is already a lot less noisy than previously, there are still some limitations:

- Inlay hints are rendered after the statement, rather than between the variable name and the `=` (i.e. in same spot as a regular type annotation)
- Hints are always rendered, regardless of whether an explicit type annotation is already provided or not

As far as I can tell, the sourcekitd-provided document symbols do not provide any information about whether a type annotation is present, making this a bit challenging. Perhaps it would make sense to introduce a new request on the sourcekitd-side for inlay hints?

Tests/SourceKitLSPTests/InlayHintsTests.swift
```diff
@@ -0,0 +1,159 @@
...
+    XCTAssertEqual(hints, [])
+  }
+
+  func testBindings() {
+    let text = """
+    let x = 4
+    var y = "test" + "123"
+    """
+    let hints = performInlayHintsRequest(text: text)
+    XCTAssertEqual(hints, [
```
ahoppen, line 71: Could you add a `FIXME` to these tests that the type hint should be displayed after the variable name if possible instead of after the expression?

Tests/SourceKitLSPTests/InlayHintsTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        category: .type,
+        label: "Int"
+      ),
+      InlayHint(
+        position: Position(line: 11, utf16index: 31),
+        category: .type,
+        label: "Double"
+      ),
+    ])
+  }
```
ahoppen, line 158: Could you also add a test case with variables that already have type annotations and add a `FIXME` to it that we shouldn’t be displaying the type hint there?

ahoppen (commented):
I’ve just reviewed the PR as a whole and left some comments inline, mostly small improvement suggestions.

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,116 @@
...
+      guard let skExpressionTypeInfos: SKDResponseArray = dict[keys.expression_type_list] else {
+        return completion(.success([]))
+      }
+
+      var expressionTypeInfos: [ExpressionTypeInfo] = []
+
+      skExpressionTypeInfos.forEach { (_, skExpressionTypeInfo) -> Bool in
+        guard let info = ExpressionTypeInfo(skExpressionTypeInfo, in: snapshot) else {
+          assert(false, "ExpressionTypeInfo failed to deserialize")
+          return false
```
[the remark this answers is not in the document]
fwcd: Isn't it the other way around, i.e. `return false` causes iteration to terminate?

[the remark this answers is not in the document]
ahoppen: Yes, but I don’t think we need to terminate iteration here. We can just continue deserializing the next element. Or am I missing something?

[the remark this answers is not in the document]
fwcd: Ah sorry, I misread your comment, sure, that makes sense.

ahoppen (approved):
Looks good to me now. Could you squash the commits? Then I’ll run CI and we can merge this.

fwcd:
Sure, here you go.

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,117 @@
...
+      guard let skExpressionTypeInfos: SKDResponseArray = dict[keys.expression_type_list] else {
+        return completion(.success([]))
+      }
+
+      var expressionTypeInfos: [ExpressionTypeInfo] = []
+      expressionTypeInfos.reserveCapacity(skExpressionTypeInfos.count)
+
+      skExpressionTypeInfos.forEach { (_, skExpressionTypeInfo) -> Bool in
+        guard let info = ExpressionTypeInfo(skExpressionTypeInfo, in: snapshot) else {
+          assert(false, "ExpressionTypeInfo failed to deserialize")
```
benlangmuir, line 90: Nitpick: `assert(false, ...)` is better expressed as `assertionFailure(...)`.

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -0,0 +1,117 @@
...
+  case responseError(ResponseError)
+}
+
+extension SwiftLanguageServer {
+  /// Must be called on self.queue.
+  private func _expressionTypeInfos(
+    _ uri: DocumentURI,
+    _ completion: @escaping (Swift.Result<[ExpressionTypeInfo], ExpressionTypeInfoError>) -> Void
+  ) {
+    if #available(macOS 10.12, *) {
```
benlangmuir, line 57: Our minimum deployment target is 10.15, so this availability check shouldn't be necessary.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -507,86 +507,105 @@ extension SwiftLanguageServer {
...
   }
 
-  public func documentSymbol(_ req: Request<DocumentSymbolRequest>) {
-    let keys = self.keys
+  // Must be called on self.queue
+  private func _documentSymbols(
+    _ uri: DocumentURI,
+    _ completion: @escaping (Result<[DocumentSymbol], ResponseError>) -> Void
+  ) {
+    if #available(macOS 10.12, *) {
```
benlangmuir, line 515: Same comment about availability check

benlangmuir (commented):
A few minor things, but otherwise LGTM

benlangmuir (approved):

PR #408, opened 2021-06-15
Implement server-side support for inlay hints using CollectVariableType

fwcd:
This is an early draft of support for inlay hints that uses the new [`CollectVariableType`](https://github.com/apple/swift/pull/37867) SourceKit request.

### Implementation Progress

- [x] Add UIDs and structures for `CollectVariableType`
- [x] Support ranged `CollectVariableType` requests
- [x] Reimplement inlay hints using `CollectVariableType`
- [x] Update existing tests
- [x] Add new test for inlay hints on closure parameters

> Note that this also requires a very recent build of the Swift toolchain, otherwise the tests will fail

### DIFF
--- a/Editors/vscode/src/inlayHints.ts
+++ b/Editors/vscode/src/inlayHints.ts
@@ -53,9 +53,9 @@ const hintStyle: InlayHintStyle = {
 
     makeDecoration: (hint, converter) => ({
         range: converter.asRange({
-            start: hint.position,
-            end: { ...hint.position, character: hint.position.character + 1 } }
-        ),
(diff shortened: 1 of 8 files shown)


fwcd:
I have tested this branch with the newest toolchain and it's looking pretty nice already:

image

There are still two minor issues that probably have to be addressed on the SourceKit-side:
* The flag that determines whether a variable declaration has an explicit type doesn't seem to work if the declaration is inside a `guard`
* Wildcard patterns (`_`) seem to generate position that is off by one character

ahoppen (approved):
LGTM

ahoppen:
@swift-ci Please test

PR #410, opened 2021-06-26
VSCode: updated package-lock.json for Windows

compnerd:
The changes to the package-lock.json here are due to the execution of
`npm install` on Windows.


Editors/vscode/package-lock.json
```diff
@@ -1,882 +1,8 @@
 {
     "name": "sourcekit-lsp",
     "version": "0.0.1",
-    "lockfileVersion": 2,
+    "lockfileVersion": 1,
```
benlangmuir, line 4: I think if you update npm itself you will be able to use lockfile version 2 and won't see this change.  That's what I did anyway.

compnerd: Hmm, I installed the latest LTS node.js version.  Is that not the version to use?

compnerd: Updating nom outside of node.js seems to have done the trick.  I think that this deserves a note in the development notes, but that can live for another PR.

PR #411, opened 2021-06-27
Packaging Improvements

compnerd:
This updates the packaging for the VSCode extension to use esbuild to merge and minify the JavaScript.  It also fixes the packaging rules to shrink the plugin ~50x (~325 KiB -> 65 KiB).

### DIFF
--- a/Editors/vscode/.vscodeignore
+++ b/Editors/vscode/.vscodeignore
@@ -1,9 +1,5 @@
-.vscode/**
-.vscode-test/**
-out/test/**
-out/**/*.map
-src/**
+**/*.ts
+**/tsconfig.json
 .gitignore
-tsconfig.json
-vsc-extension-quickstart.md
-tslint.json
+.vscode/
+node_modules/
--- a/Editors/vscode/.vscodeignore
+++ b/Editors/vscode/.vscodeignore
@@ -2,4 +2,5 @@
 **/tsconfig.json
 .gitignore
 .vscode/
+.vscode-test/
 node_modules/
--- a/Editors/vscode/README.md
+++ b/Editors/vscode/README.md
@@ -19,7 +19,8 @@ The following commands build the extension and creates a `.vsix` package in the 
 
 ```
 $ cd Editors/vscode
-$ npm run createDevPackage
+$ npm install
+$ npm run dev-package
 ```
 
 You can install the package from the command-line using the `code` command if available (see [Launching from the command line](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line)).
--- a/Editors/vscode/README.md
+++ b/Editors/vscode/README.md
@@ -26,7 +26,7 @@ $ npm run dev-package
 You can install the package from the command-line using the `code` command if available (see [Launching from the command line](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line)).
 
 ```
-code --install-extension out/sourcekit-lsp-vscode-dev.vsix
+code --install-extension sourcekit-lsp-development.vsix
 ```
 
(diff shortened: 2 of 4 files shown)


compnerd:
The changes here also enable building on Windows as the new scripts avoid bash specific syntax for packaging, relying on the developer to run `npm install`.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Editors/vscode/README.md
```diff
@@ -19,7 +19,8 @@ The following commands build the extension and creates a `.vsix` package in the
 
 ```
 $ cd Editors/vscode
-$ npm run createDevPackage
+$ npm install
+$ npm run dev-package
```
benlangmuir, line 23: Please also update the line below that mentions the path involving "out":

> code --install-extension **out**/sourcekit-lsp-vscode-dev.vsix

Editors/vscode/package.json
```diff
@@ -19,19 +19,22 @@
...
-        "vscode:prepublish": "npm run compile",
-        "compile": "tsc -p ./",
-        "watch": "tsc -watch -p ./",
-        "createDevPackage": "npm install && ./node_modules/.bin/vsce package -o ./out/sourcekit-lsp-vscode-dev.vsix"
+        "vscode:prepublish": "npm run -S esbuild-base -- --minify",
+        "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/main.js --external:vscode --format=cjs --platform=node",
+        "esbuild": "npm run -S esbuild-base -- --sourcemap",
+        "esbuild-watch": "npm run -S esbuild-base -- --sourcemap --watch",
+        "package": "vsce package",
+        "dev-package": "vsce package -o sourcekit-lsp-development.vsix",
```
benlangmuir, line 29: Sadly the trailing comma is rejected:

```
npm ERR! JSON.parse Unexpected token "}" (0x7D) in JSON at position 1080 while parsing near "...elopment.vsix\",\n    },\n    \"dependencies..."
npm ERR! JSON.parse Failed to parse JSON data.
npm ERR! JSON.parse Note: package.json must be actual JSON, not just JavaScript.
```

Editors/vscode/.vscodeignore
```diff
@@ -1,9 +1,6 @@
-.vscode/**
-.vscode-test/**
-out/test/**
-out/**/*.map
-src/**
+**/*.ts
```
benlangmuir, line 1: Why this instead of `src/`?

Editors/vscode/package.json
compnerd: Ugh, I keep making that mistake.

Editors/vscode/.vscodeignore
compnerd: Recommendation from the VSCode documentation.  I don't have a strong opinion on this.

compnerd:
@swift-ci please test

Editors/vscode/.vscodeignore
benlangmuir: Huh, I was looking at https://code.visualstudio.com/api/working-with-extensions/bundling-extension which suggests "src/", but it's not a big deal either way I guess.

benlangmuir (approved):

Editors/vscode/.vscodeignore
compnerd: https://code.visualstudio.com/api/working-with-extensions/publishing-extension#.vscodeignore is what I was looking through.  You're right that they accomplish more or less the same thing in our particular case.  The difference is that any non-type-script resources would be bundled with the new filter.  I think that if we decide that this is not the right thing for this project, we can easily change this.

PR #413, opened 2021-07-01
Use dynamic registration for folding range options

DavidGoldman:
Dynamically register because clangd doesn't currently support it, but may in the future. Without this change, we don't provide any code folding in Objective-C files because our empty results override's VS Code's default implementation.


DavidGoldman:
@swift-ci please test

DavidGoldman:
Hmm, actually gonna see if there's an easier way - maybe VS Code will ignore our results if we instead return an error?

DavidGoldman:
Nope, VS Code still hides its own folding ranges when we return an error.

benlangmuir (approved):

benlangmuir:
@swift-ci please test

PR #419, opened 2021-07-21
build: add the ability to pass in SwiftCollections

compnerd:
This allows building against SwiftCollections from a build tree when
building with SwiftPM from a build tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(IndexStoreDB QUIET)
 find_package(SwiftPM QUIET)
 find_package(LLBuild QUIET)
 find_package(ArgumentParser CONFIG REQUIRED)
+find_package(SwiftCollections QUIET)
 
 include(SwiftSupport)
 


compnerd:
CC: @tomerd
This is related to apple/swift-package-manager#3595

compnerd:
@swift-ci please test

tomerd:
seems fine to me

benlangmuir (approved):

tomerd:
we are likely to rollback the SwiftPM change, but lets keep this open when we are ready to roll forward again

cc @neonichu

neonichu (approved):

compnerd:
I think that we should merge this we are holding off on apple/swift-package-manager#3628 (CC: @tomerd)

benlangmuir:
@compnerd to confirm my cmake understanding: this will not cause any errors if collections is not found and is not needed (due to `QUIET` and not having `REQUIRED`)?  If so, I agree there is no downside to merging this PR

compnerd:
@benlangmuir precisely - this allows us to build SK-LSP with and without a build tree.

PR #420, opened 2021-07-27
[test] Update tests to pass with 5.5 toolchain

benlangmuir:
If the toolchain does not support the variable.type request, allow the
inlay type hint tests to be skipped.


benlangmuir:
@swift-ci please test

PR #421, opened 2021-07-29
Adjust SwiftPMWorkspaceTests/testBuildSetup for apple/swift-package-manager#3636,  which changes the behavior of the -Xcc command-line flag

buttaface:
SPM no longer passes `-Xcc` command-line flags to the Swift compiler after that change, so have this test pass the flag in through `-Xswiftc` also.


buttaface:
See buttaface/swift-package-manager@823cfed for the SPM change that requires this test change.

buttaface:
No longer needed as that mooted change was removed.

PR #424, opened 2021-08-20
[build-script-helper] Add support for Android cross-compilation

buttaface:
Add new flags `--cross-compile-host` and `--cross-compile-config` and use them to cross-compile, gated only for Android for now. Add a `--prefix` flag to install to a cross-compiled toolchain instead and set `SWIFT_EXEC` when using SPM, so it uses the right Swift toolchain.

Along with apple/swift#36917 and [an updated version of this `build-script` patch](https://github.com/buttaface/termux-packages/blob/swift55/packages/swift/swift-utils-swift_build_support-swift_build_support-products-indexstoredb.py.patch) to pass the right flags in, I have a working sourcekit-lsp running on Android now, after cross-compiling it from linux for the first time. I will submit that second linked patch once this is in.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 import argparse
 import os
 import platform
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


Utilities/build-script-helper.py
```diff
@@ -112,6 +116,8 @@ def handle_invocation(swift_exec, args):
   print('Cleaning ' + args.build_path)
   shutil.rmtree(args.build_path, ignore_errors=True)
 
+  env['SWIFT_EXEC'] = '%sc' % (swift_exec)
```
buttaface, line 119: I had to add this for [the same reason as `swift-driver`](https://github.com/apple/swift-driver/pull/743#discussion_r663638507), so see that note.

Utilities/build-script-helper.py
```diff
@@ -148,6 +154,8 @@ def add_common_args(parser):
     parser.add_argument('--sanitize', action='append', help='build using the given sanitizer(s) (address|thread|undefined)')
     parser.add_argument('--sanitize-all', action='store_true', help='build using every available sanitizer in sub-directories of build path')
     parser.add_argument('--verbose', '-v', action='store_true', help='enable verbose output')
+    parser.add_argument('--cross-compile-host', help='cross-compile for another host instead')
```
buttaface, line 157: Unlike other `build-script` products that added this build flag, [I made it singular](https://github.com/apple/swift/pull/36917#issue-615550282), as we only ever pass one cross-compilation host at a time.

benlangmuir:
@swift-ci please test

Utilities/build-script-helper.py
```diff
@@ -50,25 +51,28 @@ def get_swiftpm_options(args):
...
-      ]
+  if 'ANDROID_DATA' in os.environ or (args.cross_compile_host and re.match(
+    'android-', args.cross_compile_host)):
+    swiftpm_args += [
+      '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/android',
+      # SwiftPM will otherwise try to compile against GNU strerror_r on
+      # Android and fail.
+      '-Xswiftc', '-Xcc', '-Xswiftc', '-U_GNU_SOURCE',
+    ]
+  elif platform.system() == 'Linux':
```
benlangmuir, line 62: Was this check necessary for some reason now (previously it was just "else")? Mostly curious.

Utilities/build-script-helper.py
```diff
@@ -159,6 +167,7 @@ def add_common_args(parser):
 
   install_parser = subparsers.add_parser('install', help='build the package')
   add_common_args(install_parser)
+  install_parser.add_argument('--prefix', dest='install_prefixes', nargs='*', metavar='PATHS', help="paths to install sourcekit-lsp, default: 'toolchain/bin'")
```
benlangmuir, line 170: When would we want more than one prefix at a time?

Utilities/build-script-helper.py
buttaface: Yes, I moved this entire block out of the non-Darwin block above, so this project could be cross-compiled from macOS too, then I had to make sure the linux rpath isn't added for Darwin.

Utilities/build-script-helper.py
buttaface: No idea, the other `build-script` helper scripts do this, even though `build-script` only passes in one at a time, so I did the same here.

Utilities/build-script-helper.py
benlangmuir: Note to self: don't review python code changes with whitespace changes ignored!

Utilities/build-script-helper.py
```diff
@@ -131,7 +137,10 @@ def handle_invocation(swift_exec, args):
     bin_path = swiftpm_bin_path(swift_exec, swiftpm_args, env)
     swiftpm_args += ['-Xswiftc', '-no-toolchain-stdlib-rpath']
     swiftpm('build', swift_exec, swiftpm_args, env)
-    install(bin_path, args.toolchain)
+
+    if not args.install_prefixes:
+      args.install_prefixes = [args.toolchain]
```
buttaface, line 142: I had tested this default case like this initially, then modified it to just passing `args.install_prefixes if not None else [args.toolchain]` to the line below and not tested the default case again. For whatever reason, that doesn't work, so going back to this.

buttaface:
Ping, just needs another CI run.

MaxDesiatov:
@swift-ci please test

buttaface:
This can be merged, as it doesn't depend on any other pull.

PR #427, opened 2021-09-15
adjust SwiftPMWorkspace to latest SwiftPM APIs

tomerd:
motivation: do not use deprecated APIs

changes:
* explicitly import TSCBasic since SwiftPM stops re-exporting it
* adjust Workspace creation to use up-to-date APIs

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -91,9 +91,17 @@ public final class SwiftPMWorkspace {
         location.workingDirectory = customWorkingDirectory
     }
 
+      // since we are customizing the workspace location, we need to explicitly pass the mirrors configuration
+    let mirrorsConfiguration = try Workspace.Configuration.Mirrors(
+        forRootPackage: packageRoot,
+        sharedMirrorFile: location.sharedMirrorsConfigurationFile,
+        fileSystem: fileSystem
+    )
+
     self.workspace = try Workspace(
         fileSystem: fileSystem,
         location: location,
+        mirrors: mirrorsConfiguration.mirrors,
         customManifestLoader: ManifestLoader(toolchain: toolchain.configuration, cacheDir: location.workingDirectory),
         resolverUpdateEnabled: false
     )
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -86,25 +86,15 @@ public final class SwiftPMWorkspace {
     let destination = try Destination.hostDestination(destinationToolchainBinDir)
     let toolchain = try UserToolchain(destination: destination)
 
-    let buildPath: AbsolutePath = buildSetup.path ?? packageRoot.appending(component: ".build")
-
-    let mirrorsConfiguration = try Workspace.Configuration.Mirrors.init(
-        localMirrorFile: packageRoot.appending(components: ".swiftpm", "config"),
-        sharedMirrorFile: .none,
-        fileSystem: fileSystem
-    )
+    var location = Workspace.Location(forRootPackage: packageRoot, fileSystem: fileSystem)
+    if let customWorkingDirectory = buildSetup.path {
+        location.workingDirectory = customWorkingDirectory
+    }
 
     self.workspace = try Workspace(
         fileSystem: fileSystem,
-        location: .init(
-            workingDirectory: buildPath,
-            editsDirectory: packageRoot.appending(component: "Packages"),
-            resolvedVersionsFile: packageRoot.appending(component: "Package.resolved"),
-            sharedCacheDirectory: .none,
-            sharedConfigurationDirectory: .none
-        ),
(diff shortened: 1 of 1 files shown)


Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -87,16 +88,25 @@ public final class SwiftPMWorkspace {
...
+            workingDirectory: buildPath,
+            editsDirectory: packageRoot.appending(component: "Packages"),
+            resolvedVersionsFile: packageRoot.appending(component: "Package.resolved"),
+            sharedCacheDirectory: .none,
+            sharedConfigurationDirectory: .none
+        ),
+        mirrors: mirrorsConfiguration.mirrors,
+        customManifestLoader: ManifestLoader(toolchain: toolchain.configuration, cacheDir: buildPath),
+        resolverUpdateEnabled: false
+    )
```
tomerd, line 109: @benlangmuir I was not sure if there was a strong reason to define the custom locations for the LSP workspace instead of using SwiftPM default setup. if there is no such reason, I would want to adjust this to use default locations instead

benlangmuir: The reason it's hardcoding the default paths itself is that when this code was written there was no public API in SwiftPM that provided the default paths, and at least at the time we couldn't use `SwiftTool` to create our `Workspace` due to it not exposing control over the toolchain/destination (there may be other issues, but that's the one I hit first).  I see that now we have custom destination support, but it would still require writing a file to disk that `SwiftTool` will trigger parsing of, so it seems like that API is more focused on serving SwiftPM's CLI and not suitable for programatic use.

It would be great to replace any code where we are duplicating that default path computation, such as the editsDirectory, resolvedVersionsFile, etc. with the new `Workspace.DefaultLocations` APIs 😃   The only one we allow customization of right now is the build data path (I guess now that's called the Workspace's "working directory"?), but we could at least fix the default value to something like this:

```swift
let buildPath: AbsolutePath = buildSetup.path ?? Workspace.DefaultLocations.workingDirectory(forRootPackage: packageRoot)
```

Incidentally, do we need to pass in the mirrors since we are specifying `resolverUpdateEnabled: false`?

tomerd: cc @neonichu @abertelrud

tomerd: @benlangmuir sounds good, so I will adjust this PR to use more fo the defaults, with customization of the working directory as you suggested

> Incidentally, do we need to pass in the mirrors since we are specifying resolverUpdateEnabled: false?

I kept it in place since it was there before, but less is more in this case imo

tomerd: @benlangmuir see https://github.com/apple/sourcekit-lsp/pull/427/commits/6618be669e245344fdf81af484e8e81e0a9f187f

btw, why do we set `resolverUpdateEnabled: false` (previously `skipUpdate: true`) in this case?

tomerd:
@swift-ci test

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
benlangmuir: We don't want sourcekit-lsp hitting the network behind your back or modifying state of your Package.resolved by mistake.  If we were using our own build directory instead of sharing it we might revise our policy and give the user a way to let us pull down the dependencies ourselves.

benlangmuir (approved):

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
tomerd: what does the user workflow look like when using this? do they need to use SwiftPM CLI to update the packages first?

benlangmuir: Yes, and more importantly build so that we have all the .swiftmodules available and indexing data.  Long term I would like us to use a separate build directory, generate the .swiftmodules efficiently without typechecking function bodies, and do background indexing in sourcekit-lsp, but for now the model is that you build your project from the CLI.

benlangmuir:
@swift-ci please test macOS

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
tomerd: in that case we probably want the mirrors setup back, so that the resolved file is handles the same way (it maps back and forth from the mirrors). i will amend the PR

benlangmuir: Okay, sorry for leading you astray in that case!

tomerd:
@swift-ci test

tomerd:
@benlangmuir this should be ready once CI passes

benlangmuir:
@swift-ci please test macOS

tomerd:
@swift-ci please test macOS

PR #428, opened 2021-09-20
Fix document symbols regression

DavidGoldman:
This appears to have been accidentally removed in 17f656865ddc7abe6273ec8752b596578232d440, the `children` field is
optional in the LSP spec.


DavidGoldman:
@swift-ci please test

benlangmuir (approved):

ahoppen:
Thanks for fixing this. I didn’t think about the LSP spec when I suggested we make it non-optional.

PR #229, opened 2020-02-05
Update README "Status" table for fixit support

benlangmuir:
As of #216 by @ahoppen

PR #430, opened 2021-09-28
Use new way when building paths based on triple.

3405691582:
In TSC pr #229 and SPM pr #3576, we introduced a method named
`platformBuildPathComponent` to distinguish the valid platform path name
versus the valid target name, where one may have an embedded version
number and the other may not.

SourceKit test suite SwiftPMIntegrationTests currently fails. This is
presumably because this change may be required as it is one of two
uses of `tripleString`. The other uses `tripleString` as the value for
the `-target` field, so it is correct as written. The other -- as in
this commit -- builds a path.

This commit must be committed in concert with the aforementioned prs.


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift-tools-support-core/pull/229
https://github.com/apple/swift-package-manager/pull/3576
@swift-ci Please test

benlangmuir:
apple/swift-tools-support-core#229
apple/swift-package-manager#3576
@swift-ci Please test

benlangmuir (approved):

PR #431, opened 2021-10-02
[LSP] Add `RenameRequest` and `PrepareRenameRequest`

Snowy1803:
Add support for renaming in LanguageServerProtocol. Rename options are restricted to pre-3.16 options, the ones added in 3.16, such as prepareSupportDefaultBehavior, aren't supported.

### DIFF
--- a/Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
@@ -15,7 +15,7 @@
 /// Looks up the symbol at the given position and returns the range of the symbol the user is renaming,
 /// or null if there is no symbol to be renamed at the given position
 ///
-/// Servers that provide rename preparation should set`prepareProvider` to true in the `renameProvider` server capability.
+/// Servers that provide rename preparation should set `prepareProvider` to true in the `renameProvider` server capability.
 ///
 /// - Parameters:
 ///   - textDocument: The document in which to lookup the symbol location.
--- a/Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
@@ -1,0 +1,88 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request to test the validity of a rename operation at the given location.
+///
+/// Looks up the symbol at the given position and returns the range of the symbol the user is renaming,
+/// or null if there is no symbol to be renamed at the given position
+///
+/// Servers that provide rename preparation should set`prepareProvider` to true in the `renameProvider` server capability.
+///
+/// - Parameters:
+///   - textDocument: The document in which to lookup the symbol location.
+///   - position: The document location at which to lookup symbol information.
+///
+/// - Returns: A range for the symbol, and optionally a placeholder text of the string content to be renamed.
+public struct PrepareRenameRequest: TextDocumentRequest, Hashable {
+  public static let method: String = "textDocument/prepareRename"
+  public typealias Response = PrepareRenameResponse?
+
+  /// The document in which to lookup the symbol location.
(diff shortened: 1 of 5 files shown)


Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
```diff
@@ -0,0 +1,88 @@
...
+  /// The document location at which to lookup symbol information.
+  public var position: Position
+
+  public init(textDocument: TextDocumentIdentifier, position: Position) {
+    self.textDocument = textDocument
+    self.position = position
+  }
+}
+
+public struct PrepareRenameResponse: ResponseType, Hashable {
```
DavidGoldman, line 41: Looks like in the spec it's defined as `Range | { range: Range, placeholder: string } | { defaultBehavior: boolean } | null` - do we need to handle the defaultBehavior case too?

DavidGoldman:
@swift-ci please test

Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
Snowy1803: defaultBehavior was added in LSP 3.16, and as I said, I’m not supporting that (would require adding it to the client capability, adding the appropriate rename option for what the default behavior is, etc). IMO it isn’t necessary

Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
```diff
@@ -0,0 +1,88 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request to test the validity of a rename operation at the given location.
+///
+/// Looks up the symbol at the given position and returns the range of the symbol the user is renaming,
+/// or null if there is no symbol to be renamed at the given position
+///
+/// Servers that provide rename preparation should set`prepareProvider` to true in the `renameProvider` server capability.
```
ahoppen, line 18: Missing space 
```suggestion
/// Servers that provide rename preparation should set `prepareProvider` to true in the `renameProvider` server capability.
```

Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
ahoppen: I think it’s interesting that LSP added a new response option with a new version of LSP. AFAICT the server can’t know whether the client supports this response option or am I missing something?

---

In any case, wouldn’t it be cleaner to explicitly model the three different response types by e.g. making `PrepareRenameResponse` as an `enum`? As it stands now, we use `placeholder == nil` as a flag to convey whether the response was a raw `Range` or `{ range: Range, placeholder: string }`. This means that 
```json
{
  "range": …,
  "placeholder": null
}
```
wouldn’t round-trip. I know, the spec requires `placeholder` to be non-optional but still, we might as well prepare for `defaultBehavior` now.

ahoppen:
Thanks for contributing to SourceKit-LSP @Snowy1803. I added one more response to @DavidGoldman’s comment, otherwise it’s looking good to me 👍.

Sources/LanguageServerProtocol/Requests/PrepareRenameRequest.swift
Snowy1803: Client Capabilities have an additional `renameProvider.prepareSupportDefaultBehavior` property to specify support

ahoppen: Ah, that makes sense. I’d be fine to merge this as-is.

ahoppen (approved):

Snowy1803:
@DavidGoldman @benlangmuir May I ask what is needed in order to get this PR merged?

ahoppen:
Thanks for the ping @Snowy1803, I didn’t have the PR on my radar anymore.

I think the comment https://github.com/apple/sourcekit-lsp/pull/431#discussion_r721365241 is still open. Could you comment on that thread?

Snowy1803:
It has already been answered, hasn't it? As long as the client capability `renameProvider.prepareSupportDefaultBehavior` isn't provided, we don't have to support `defaultBehavior`

ahoppen:
I’m sorry @Snowy1803. I was looking at the wrong view in GitHub that didn’t show your reply.

ahoppen:
@swift-ci Please test and merge

Snowy1803:
Do we have to wait some more or is there an issue with the bot?

ahoppen:
@swift-ci Please test

ahoppen:
Forgot that this repo doesn’t support testing + merging in one step.

Snowy1803:
That's odd

PR #432, opened 2021-10-05
Implement `completionItem/resolve`

LebJe:
This pull request adds support for `completionItem/resolve` to SourceKit-LSP.

Unfortunately, the tests are disabled (and this PR is a draft) because SourceKit cannot provide `cursorinfo` via an USR for anything other than things like `struct`s:
```
{
  key.internal_diagnostic: "Unable to resolve type from USR."
}
```

Test Code:

```swift
/// `a` (does not work)
let a = 1

/// `b` (does not work)
func b(c: Int) {}

/// `E` (works)
enum E: CaseIterable {
	/// `a` (does not work)
	case a
	
	/// `b` (does not work)
	case b
	
	/// `c` (does not work)
	case c
}

/// `A` (works)
struct A: Equatable {
	
	/// `a` (does not work)
	let a: String
		
  	/// - Parameters:
  	///    - a: `String`
	/// - Note: does not work
	init(a: String) {
		self.a = a
	}
}
```

To test the above code, type something like `E.a`, or `A.init`, and see if the \`\` is rendered as a code block in the completion popup.

Perhaps this is related to [this issue](https://bugs.swift.org/browse/SR-13958).

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
@@ -10,8 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A single completion result.
-public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Codable, Hashable {
+/// `CompletionItem` is a request for the `completionItem/resolve` method and a response type to the `textDocument/completion` request.
+public struct CompletionItem: TextDocumentRequest, ResponseType, Codable, Hashable {
+  public static var method: String = "completionItem/resolve"
   public typealias Response = CompletionItem
     
   public var textDocument: TextDocumentIdentifier {
@@ -25,9 +26,6 @@ public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Co
       fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
     }
   }
-  
-  public static var method: String = "completionItem/resolve"
-
 
   /// The display name of the completion.
   public var label: String
@@ -56,7 +54,7 @@ public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Co
 
   /// **Deprecated**: use `textEdit`
   ///
-  /// The string to insert into the document. If `nil`, use `label.
+  /// The string to insert into the document. If `nil`, use `label`.
   public var insertText: String?
 
   /// The format of the `textEdit.nextText` or `insertText` value.
@@ -92,20 +90,6 @@ public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Co
     self.kind = kind
     self.deprecated = deprecated
     self.data = data
-  }
-  
-  enum CodingKeys: String, CodingKey {
-    case label,
-    detail,
-    documentation,
-    sortText,
-    filterText,
-    textEdit,
-    insertText,
-    insertTextFormat,
-    kind,
-    deprecated,
-    data
   }
 }
 
--- a/Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
@@ -11,7 +11,23 @@
 //===----------------------------------------------------------------------===//
 
 /// A single completion result.
-public struct CompletionItem: Codable, Hashable {
+public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Codable, Hashable {
+  public typealias Response = CompletionItem
+    
+  public var textDocument: TextDocumentIdentifier {
+    if 
+      let data = self.data,
+      case let .dictionary(dict) = data,
+      case let .string(textDocURI) = dict["textDocURI"]
+    {
+      return TextDocumentIdentifier(DocumentURI(string: textDocURI))
+    } else {
+      fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
+    }
+  }
+  
+  public static var method: String = "completionItem/resolve"
+
 
   /// The display name of the completion.
   public var label: String
@@ -48,6 +64,9 @@ public struct CompletionItem: Codable, Hashable {
 
   /// Whether the completion is for a deprecated symbol.
   public var deprecated: Bool?
+  
+  /// A data entry field that is preserved on a `CompletionItem` between a completion and a completion resolve request.
+  public var data: LSPAny?
 
   public init(
     label: String,
@@ -59,7 +78,8 @@ public struct CompletionItem: Codable, Hashable {
     textEdit: TextEdit? = nil,
     insertText: String? = nil,
     insertTextFormat: InsertTextFormat? = nil,
-    deprecated: Bool? = nil)
+    deprecated: Bool? = nil,
+    data: LSPAny? = nil)
   {
     self.label = label
     self.detail = detail
@@ -71,6 +91,21 @@ public struct CompletionItem: Codable, Hashable {
     self.insertTextFormat = insertTextFormat
     self.kind = kind
     self.deprecated = deprecated
+    self.data = data
+  }
+  
+  enum CodingKeys: String, CodingKey {
+    case label,
+    detail,
+    documentation,
+    sortText,
+    filterText,
+    textEdit,
+    insertText,
+    insertTextFormat,
+    kind,
+    deprecated,
+    data
   }
 }
 
--- a/Sources/SourceKitLSP/Swift/CodeCompletion.swift
+++ b/Sources/SourceKitLSP/Swift/CodeCompletion.swift
@@ -170,19 +170,9 @@ extension SwiftLanguageServer {
       let notRecommended = (value[self.keys.not_recommended] as Int?).map({ $0 != 0 })
 
       let kind = (value[self.keys.kind] as sourcekitd_uid_t?)?.asCompletionItemKind(self.values) ?? .value
-      var usr: String? = nil
-      
-      switch kind {
-        // Don't fill in `usr` for things other than Swift symbols.
-        case .class, .struct, .constant, .variable, .enum, .enumMember, .constructor,
-             .field, .function, .interface, .method, .module, .operator, .property, .typeParameter:
-          usr = value[self.keys.associated_usrs]
-        default: break
-      }
-    
       var data: [String: LSPAny] = ["textDocURI": .string(snapshot.document.uri.stringValue)]
 
-      if usr != nil { data["usr"] = .string(usr!) }
+      if let usr: String = value[self.keys.associated_usrs] { data["usr"] = .string(usr) }
       
       result.items.append(CompletionItem(
         label: name,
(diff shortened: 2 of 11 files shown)


Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -11,7 +11,23 @@
...
+    
+  public var textDocument: TextDocumentIdentifier {
+    if 
+      let data = self.data,
+      case let .dictionary(dict) = data,
+      case let .string(textDocURI) = dict["textDocURI"]
+    {
+      return TextDocumentIdentifier(DocumentURI(string: textDocURI))
+    } else {
+      fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
```
ahoppen, line 25: I don’t think we should make any assumptions about the requests sent to us by the client and, in particular, not crash if it is malformed. Could we return a failure value (e.g.`nil`) here instead of `fatalError`ing?

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -71,6 +91,21 @@ public struct CompletionItem: Codable, Hashable {
     self.insertTextFormat = insertTextFormat
     self.kind = kind
     self.deprecated = deprecated
+    self.data = data
+  }
+  
+  enum CodingKeys: String, CodingKey {
```
ahoppen, line 97: Maybe a stupid question but why do we need to specify the coding keys here? AFAICT we only provide it in other types if we are customizing the decoding initializer or mapping the coding keys to different strings and I think none of this is happening here.

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -11,7 +11,23 @@
 //===----------------------------------------------------------------------===//
 
 /// A single completion result.
-public struct CompletionItem: Codable, Hashable {
+public struct CompletionItem: TextDocumentRequest, RequestType, ResponseType, Codable, Hashable {
```
ahoppen, line 14: Could you add a comment e.g.
```swift
/// CompletionItem is a request for the `completionItem/resolve` method and a response type to the `textDocument/completion` request.
```
It’s dual behavior confused me for quite a while.

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -11,7 +11,23 @@
...
+      case let .dictionary(dict) = data,
+      case let .string(textDocURI) = dict["textDocURI"]
+    {
+      return TextDocumentIdentifier(DocumentURI(string: textDocURI))
+    } else {
+      fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
+    }
+  }
+  
+  public static var method: String = "completionItem/resolve"
```
ahoppen, line 29: I think the `method` property is always the first property in a request type. Could you move it up to match this style?

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -169,18 +169,33 @@ extension SwiftLanguageServer {
...
       let notRecommended = (value[self.keys.not_recommended] as Int?).map({ $0 != 0 })
 
-      let kind: sourcekitd_uid_t? = value[self.keys.kind]
+      let kind = (value[self.keys.kind] as sourcekitd_uid_t?)?.asCompletionItemKind(self.values) ?? .value
+      var usr: String? = nil
+      
+      switch kind {
+        // Don't fill in `usr` for things other than Swift symbols.
+        case .class, .struct, .constant, .variable, .enum, .enumMember, .constructor,
+             .field, .function, .interface, .method, .module, .operator, .property, .typeParameter:
```
ahoppen, line 178: Which cases do we need this check for? Couldn’t we just check if the `associated_usrs` key is available in `value`?

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -283,4 +298,56 @@ extension SwiftLanguageServer {
 
     return TextEdit(range: textEditRangeStart..<requestPosition, newText: newText)
   }
+  
+    /// Must be called on `queue`.
```
ahoppen, line 302: If this must be called on `queue`, could you add a dispatch precondition checking it. E.g. https://github.com/apple/sourcekit-lsp/blob/d29d1254a4bf19e91d04eec441c56b6142b2a4c6/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L170

ahoppen (commented):
Thanks a lot for looking into this @LebJe. I’ll need to look into why `sourcektid` doesn’t return cursor info for the USR. Maybe @benlangmuir has an idea if this is something that’s supported or what might be going wrong here from the top of his head.

For now, I’ve added some comments inline.

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
LebJe: I originally thought `textDocument` would be encoded, but I just tried removing `CodingKeys` and it actually isn't encoded, so I'll remove this.

LebJe:
@benlangmuir, are you saying I should:

- Have a `Dictionary<String, SKDResponseDictionary>` that stores the `codecomplete` results.
- Rewrite `completionsFromSKDResponse` so that it:
	- Clears the old completions form the dictionary, and fills it with `<unique id>: <completion response>`.
	- Returns the minimum amount of information to the LSP client.
- Rewrite `_completionResolve` so that it:
	- Uses the completion dictionary to fill in the rest of the data, and do any necessary processing (like rewriting placeholders).
	
With this approach though, it is still impossible to get full documentation.

Is there something I'm not noticing in the `codecomplete` results that would help me get the full documentation?

---
> Cursor info on USRs only works on a subset of USRs for nominal types IIRC. Please don't use this functionality. We should never have added it.

Was the implementation never finished, or is something causing cursor info via USRs on things other than nominal types to not work?

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -10,8 +10,22 @@
...
+    
+  public var textDocument: TextDocumentIdentifier {
+    if 
+      let data = self.data,
+      case let .dictionary(dict) = data,
+      case let .string(textDocURI) = dict["textDocURI"]
+    {
+      return TextDocumentIdentifier(DocumentURI(string: textDocURI))
+    } else {
+      fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
```
kcmed420, line 26:
```suggestion
/// `CompletionItem` is a request for the `completionItem/resolve` method and a response type to the `textDocument/completion` request.
public struct CompletionItem: TextDocumentRequest, ResponseType, Codable, Hashable {
  public static var method: String = "completionItem/resolve"
  public typealias Response = CompletionItem
    
  public var textDocument: TextDocumentIdentifier {
    if 
      let data = self.data,
      case let .dictionary(dict) = data,
      case let .string(textDocURI) = dict["textDocURI"]
    {
      return TextDocumentIdentifier(DocumentURI(string: textDocURI))
    } else {
      fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
```

kcmed420: ****

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
```diff
@@ -59,7 +76,8 @@ public struct CompletionItem: Codable, Hashable {
     textEdit: TextEdit? = nil,
     insertText: String? = nil,
     insertTextFormat: InsertTextFormat? = nil,
-    deprecated: Bool? = nil)
```
kcmed420, line 62: +

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -169,18 +169,23 @@ extension SwiftLanguageServer {
       // Map SourceKit's not_recommended field to LSP's deprecated
       let notRecommended = (value[self.keys.not_recommended] as Int?).map({ $0 != 0 })
 
-      let kind: sourcekitd_uid_t? = value[self.keys.kind]
```
kcmed420, line 172: ?=value

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -169,18 +169,23 @@ extension SwiftLanguageServer {
...
+      let kind = (value[self.keys.kind] as sourcekitd_uid_t?)?.asCompletionItemKind(self.values) ?? .value
+      var data: [String: LSPAny] = ["textDocURI": .string(snapshot.document.uri.stringValue)]
+
+      if let usr: String = value[self.keys.associated_usrs] { data["usr"] = .string(usr) }
+      
       result.items.append(CompletionItem(
         label: name,
-        kind: kind?.asCompletionItemKind(self.values) ?? .value,
```
kcmed420, line 175: +

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -283,4 +288,58 @@ extension SwiftLanguageServer {
 
     return TextEdit(range: textEditRangeStart..<requestPosition, newText: newText)
   }
+  
+    /// Must be called on `queue`.
```
kcmed420, line 292: "Queue"

Sources/SourceKitLSP/Swift/CodeCompletion.swift
kcmed420: image

Sources/SourceKitLSP/Swift/CodeCompletion.swift
```diff
@@ -283,4 +288,58 @@ extension SwiftLanguageServer {
 
     return TextEdit(range: textEditRangeStart..<requestPosition, newText: newText)
   }
```
kcmed420, line 285: {

kcmed420 (approved):

Sources/LanguageServerProtocol/SupportTypes/CompletionItem.swift
kcmed420: > ```suggestion
> /// `CompletionItem` is a request for the `completionItem/resolve` method and a response type to the `textDocument/completion` request.
> public struct CompletionItem: TextDocumentRequest, ResponseType, Codable, Hashable {
>   public static var method: String = "completionItem/resolve"
>   public typealias Response = CompletionItem
>     
>   public var textDocument: TextDocumentIdentifier {
>     if 
>       let data = self.data,
>       case let .dictionary(dict) = data,
>       case let .string(textDocURI) = dict["textDocURI"]
>     {
>       return TextDocumentIdentifier(DocumentURI(string: textDocURI))
>     } else {
>       fatalError("Missing textDocURI in CompletionItem.data: \(data ?? "(CompletionItem.data is nil)")")
> ```
>

Sources/SourceKitLSP/Swift/CodeCompletion.swift
kcmed420: - [ ]

thanvanthanh (approved):

PR #433, opened 2021-10-21
adjust API to changes in SwiftPM

tomerd:
motivation: stay with the times

changes:
* use observability system instead of diganostics engine
* some APIs in SwiftPM are now throwing, adjust to this change

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -247,9 +247,9 @@ extension SwiftPMWorkspace: SKCore.BuildSystem {
     // TODO: Support for change detection (via file watching)
     var settings: FileBuildSettings? = nil
     do {
-        settings = try self.settings(for: uri, language)
+      settings = try self.settings(for: uri, language)
     } catch {
-        log("error computing settings: \(error)")
+      log("error computing settings: \(error)")
     }
     DispatchQueue.global().async {
       if let settings = settings {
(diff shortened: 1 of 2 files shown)


Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -366,7 +369,7 @@ extension SwiftPMWorkspace {
     args += ["-c"]
     args += td.sources.map { $0.pathString }
     args += ["-I", buildPath.pathString]
-    args += td.compileArguments()
+    args += try! td.compileArguments()
```
benlangmuir, line 372: Why do we know this cannot throw here?  Same question for `basicArguments` below.

tomerd: to be clear I am happy to bring these changes back in, but they end up effecting things like the BuildSystem protocol (registerForChangeNotifications needs to handle errors or marked as throwing) which is a bit beyond my understanding of LSP

tomerd: I think errors from here are rare, but fatalError seems wrong since it will cause a crash. logging is fine I guess. would you like me to do that?

tomerd: see https://github.com/apple/sourcekit-lsp/pull/433/commits/4c36e5a322e51fc9dceef334ba376b0ba1e7fc6a

tomerd:
@swift-ci please test

tomerd:
@swift-ci please test

benlangmuir (approved):

PR #438, opened 2021-12-05
Remove available check below macOS 10.15.4

Kyle-Ye:
Since the Package.swift declare the minimal target for macOS is 10.15.4, we can remove the uncessary available check code


Kyle-Ye:
@swift-ci please test

ahoppen (approved):
Thanks for cleaning this up. LGTM.

PR #439, opened 2021-12-05
Fix deprecated os_log warning

Kyle-Ye:
os_log function is marked as deprecated in os framework

It is suggested that we should "Use log(level:_:) instead."

Screen Shot 2021-12-05 at 17 27 53


Kyle-Ye:
@swift-ci please test

Sources/LSPLogging/Logging.swift
```diff
@@ -240,3 +244,8 @@ public class AnyLogHandler: LogHandler {
     handler(message, level)
   }
 }
+
+@available(macOS 11.0, *)
+extension os.Logger {
+  fileprivate static let shared = os.Logger()
+}
```
ahoppen, line 251: I think this needs to be guarded by `#if canImport(os)`. It’s causing a compilation error on Linux.

Kyle-Ye:
@swift-ci Please test

Sources/LSPLogging/Logging.swift
```diff
@@ -240,3 +244,10 @@ public class AnyLogHandler: LogHandler {
     handler(message, level)
   }
 }
+
+#if canImport(os)
+@available(macOS 11.0, *)
+extension os.Logger {
+  fileprivate static let shared = os.Logger()
+}
+#endif
```
ahoppen, line 253: Sorry for not mentioning this earlier, but is there a reason why we have a shared `os.Logger` and not one for each of our own `Logger` instances? If not, I think it would be cleaner if each `Logger` had its own `os.Logger`.

Kyle-Ye: I think the same at first, but the compiler won't let us do it.
It seems we can't have @available check with stored property.
https://stackoverflow.com/questions/41904724/using-available-with-stored-properties
```
public final class Logger {
#if canImport(os)
  @available(macOS 11.0, *)
  var oslogger = os.Logger() // Error
#endif
}
```

ahoppen (approved):

PR #441, opened 2021-12-24
Use the new VSCode Extension

0xTim:
This PR changes the editor documentation to point to the [new VSCode Swift](https://github.com/swift-server/vscode-swift) extension [recently announced](https://forums.swift.org/t/introducing-swift-for-visual-studio-code/54246) by the SSWG.

It also removes the VSCode extension code as it has been consumed by the extension. Opening the PR for comment to see if we think it's worth removing the code or leaving as is.


benlangmuir:
Thanks for the PR, @0xTim. Congratulations on releasing the new VSCode extension!  It may take a little while to get back to you on this PR as folks are still on holidays.

akyrtzi:
Thank you @0xTim , the new VSCode extension is amazing!

akyrtzi (approved):

PR #442, opened 2022-01-06
Don't report duplicate folding ranges

ahoppen:
I’m not entirely sure when this started occurring but when retrieving folding ranges e.g. for
```swift
func foo() {
    print("someText")
}
```
We report the folding range of `"someText"` twice: Once as the body of the call to `print` and once as the body of the first argument. Each folding range on its own makes sense because you might want to collapse either all arguments or only a single argument. But it doesn’t make sense to report the same range twice.

### DIFF
--- a/Sources/SKTestSupport/INPUTS/FoldingRange/FoldingRangeDuplicateRanges.swift
+++ b/Sources/SKTestSupport/INPUTS/FoldingRange/FoldingRangeDuplicateRanges.swift
@@ -1,0 +1,3 @@
+func foo() {
+    print("/*fr:duplicateRanges*/")
+}
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -953,7 +953,10 @@ extension SwiftLanguageServer {
           return req.reply([])
         }
 
-        var ranges: [FoldingRange] = []
+        /// Some ranges might occur multiple times.
+        /// E.g. for `print("hi")`, `"hi"` is both the range of all call arguments and the range the first argument in the call.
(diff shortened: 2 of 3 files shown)


ahoppen:
@swift-ci Please test

Sources/SKSupport/Array+RemoveDuplicatesFromSorted.swift
```diff
@@ -0,0 +1,18 @@
+public extension Array where Element: Comparable {
+  /// Assuming that the array is sorted, remove all duplicate entries.
+  mutating func removeDuplicatesFromSorted() {
+    var lastElement: Element?
+    for (index, element) in zip(self.indices, self).reversed() {
+      if lastElement == element {
+        self.remove(at: index)
```
benlangmuir, line 7: This is quadratic since it needs to shift the trailing elements every time, and it requires an allocation since it captures self in the zip before modifying it.  You could use `removeAll(where:)` to do it in linear time, or else build a new array from the uniqued values instead of modifying the existing array in-place.

For the current use-case I expect this doesn't matter, since you're probably not removing many elements.  But I figured I would mention it since this is a generic method on Array.

benlangmuir (approved):

Sources/SKSupport/Array+RemoveDuplicatesFromSorted.swift
ahoppen: Very good catch. Another classical case of premature optimization I fell for. I thought this would be faster than `Set(ranges).sorted()` because we can harvest the property that the array is already sorted. But doing some measurements shows I’m totally wrong. `Set(ranges).sorted()` outperforms anything I tried and is way simpler.

ahoppen:
@swift-ci Please test

PR #444, opened 2022-01-18
Delete LinuxMain and XCTestManifests and automatically discover test on Linux

ahoppen:
Automatic test discovery is enabled by default since Swift 5.4. I think we can start relying on it.


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #445, opened 2022-01-21
Try determining diagnostic ranges

ahoppen:
If the diagnostic has a range associated with it that starts at the same location as the diagnostics position, we can use it to retrieve a proper range for the diagnostic, instead of just reporting a zero-length range.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
Linux is failing because it’s running on Ubuntu 16.04 but Swift requires a more recent clang than installed on those machines. We need to wait until the CI job is updated to a more recent Ubuntu.

benlangmuir (approved):

PR #446, opened 2022-01-21
Only referesh semantic tokens if SourceKit returned semantic information

ahoppen:
In certain race condition scenarios, we might receive a document update to which the 0,0 replacetext request returned only syntactic information. In that case we shouldn’t update the sematnic syntax highlighting tokens.

The race condition rarely surfaced as a test failure of `testSemanticMultiEdit`.

CC @fwcd

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -302,6 +302,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   /// Should be called on self.queue.
   func handleDocumentUpdate(uri: DocumentURI) {
+    dispatchPrecondition(condition: .onQueue(queue))
     guard let snapshot = documentManager.latestSnapshot(uri) else {
       return
     }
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Linux is failing because it’s running on Ubuntu 16.04 but Swift requires a more recent clang than installed on those machines. We need to wait until the CI job is updated to a more recent Ubuntu.

ahoppen:
@swift-ci Please test Linux

benlangmuir (approved):

fwcd:
Looks good to me, thanks for investigating this rare case! Do we get semantic information in a subsequent document update if this occurs?

ahoppen:
Yes, we do, because [`testSemanticMultiEdit`](https://github.com/apple/sourcekit-lsp/blob/06df8239a0b6efeb3608f32bb262cd3f2a74a0c3/Tests/SourceKitLSPTests/SemanticTokensTests.swift#L673) always passed with this change and thus the document update with semantic information must have been delivered later on.

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Linux

PR #450, opened 2022-01-27
Set didChangedWatchedFiles glob pattern to `**/*.ext`

adam-fowler:
`**.ext` will only look in the base folder.

@ahoppen


ahoppen:
@swift-ci Please test

ahoppen (approved):

ahoppen:
Thanks a lot for fixing this, Adam. I think the situation was even worse and `**.ext` didn’t even watch files in the base folder because apparently `**` only matches path components but not file name components.

benlangmuir (approved):

PR #451, opened 2022-01-28
[5.6] Don't report duplicate folding ranges

ahoppen:
Cherry-pick https://github.com/apple/sourcekit-lsp/pull/442 to release/5.6.

---

I’m not entirely sure when this started occurring but when retrieving folding ranges for e.g.
```
func foo() {
    print("someText")
}
```
SourceKit reports the folding range of `"someText"` twice: Once as the body of the call to `print` and once as the body of the first argument. Each folding range on its own makes sense because you might want to collapse either all arguments or only a single argument. But it doesn’t make sense to report the same range twice and this is causing SourceKit-LSP tests to fail.

rdar://88172766


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #452, opened 2022-01-28
[5.6] Only referesh semantic tokens if SourceKit returned semantic information

ahoppen:
Cherry-pick https://github.com/apple/sourcekit-lsp/pull/446 to release/5.6

---

In certain race condition scenarios, we might receive a document update to which the 0,0 replacetext request returned only syntactic information. In that case we shouldn’t update the sematnic syntax highlighting tokens.

The race condition rarely surfaced as a test failure of `testSemanticMultiEdit`.

rdar://88173764


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #454, opened 2022-01-31
remove unnecessary dependency on `TSCLibc`

compnerd:
This reduces the dependency on the `TSCLibc` module which was not
actually used for anything but for importing the C library in
`SKSupport`.  Although on Apple platforms and Linux this could largely
be dealt with by statically linking, Windows does not support static
linking of Swift content properly yet.  This avoids the dependency on
the library instead.

### DIFF
--- a/Sources/SKSupport/dlopen.swift
+++ b/Sources/SKSupport/dlopen.swift
@@ -10,7 +10,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-import TSCLibc
+#if os(Windows)
+import CRT
+import WinSDK
+#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+import Darwin
+#else
+import Glibc
+#endif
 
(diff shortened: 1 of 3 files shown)


benlangmuir:
@swift-ci please test

Sources/SKSupport/dlopen.swift
```diff
@@ -10,7 +10,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-import TSCLibc
```
benlangmuir, line 13: There's a _reference_ to TSCLibc on line 92 that will need to be changed
```swift
 guard let handle = TSCLibc.dlopen(path, mode.rawValue) else {
```

compnerd: Ah, thanks; that actually is the re-exported libc function, so we should be able to just drop `TSCLibc.`.

compnerd:
@swift-ci please test

PR #455, opened 2022-02-02
explicitly indicate the TSCUtility interfaces used

compnerd:
This explicitly identifies the TSCUtility interfaces that SourceKit
depends on.  This helps identify a "burn down list" of interfaces that
remain in TSC(Utility) which are in use.  As these interfaces are
replaced, we can easily monitor the remaining interfaces that are in
use.

### DIFF
--- a/Sources/SKCore/BuildSetup.swift
+++ b/Sources/SKCore/BuildSetup.swift
@@ -11,8 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import TSCBasic
-import TSCUtility
 import SKSupport
+
+import struct TSCUtility.BuildFlags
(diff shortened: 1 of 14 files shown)


compnerd:
CC: @tomerd

benlangmuir:
I assume this means the goal is to remove or move these APIs to a new home?

compnerd:
@benlangmuir yes, that is the ultimate goal - find a more appropriate home for them and replace the usage of these interfaces.

tomerd:
this is great. I would love to see us deprecate TSC over time

compnerd:
@swift-ci please test

PR #456, opened 2022-02-08
Add option to not clean build directory prior to performing action

ahoppen:
We are repeatedly building SourceKit-LSP in CI because we clean the build directory prior to to testing & installing, which shouldn’t really be necessary.

rdar://82448307

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -106,8 +106,9 @@ def handle_invocation(swift_exec, args):
   if args.sanitize and 'thread' in args.sanitize:
     env['TSAN_OPTIONS'] = 'halt_on_error=true'
 
-  print('Cleaning ' + args.build_path)
-  shutil.rmtree(args.build_path, ignore_errors=True)
+  if not args.no_clean:
+    print('Cleaning ' + args.build_path)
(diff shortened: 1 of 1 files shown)


benlangmuir (approved):
The install action is adding `-no-toolchain-stdlib-rpath` which is not present in the build/test phases. Are we rebuilding in the install phase still?  We may need to change that as well.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

buttaface:
> Are we rebuilding in the install phase still?

Yes, the SPM commands are still run as normal, this flag simply doesn't wipe the previous build before the test/install steps, which was forcing a from-scratch build for each step. SPM is pretty smart and simply applies the needed changes to update the past build, so that rpath flag works, though it's always possible that something else is missed on these non-clean builds that the tests simply haven't caught.

I had [disabled this unnecessary cleaning on my Android CI that cross-compiles all new snapshot tags of this repo for Android](https://github.com/buttaface/swift-android-sdk/blob/c343118023aab7e390f81aeb1ef7bceabff0936a/swift-android-ci.patch#L19), so I'm happy to remove that now. 😄

PR #458, opened 2022-02-14
Update to new SwiftPM Workspace initializer

benlangmuir:


Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -104,19 +104,19 @@ public final class SwiftPMWorkspace {
         location.workingDirectory = customWorkingDirectory
     }
 
-      // since we are customizing the workspace location, we need to explicitly pass the mirrors configuration
-    let mirrorsConfiguration = try Workspace.Configuration.Mirrors(
-        forRootPackage: packageRoot,
-        sharedMirrorFile: location.sharedMirrorsConfigurationFile,
-        fileSystem: fileSystem
-    )
+    let configuration = WorkspaceConfiguration(
```
tomerd, line 107:
```
var configuration = WorkspaceConfiguration.default
configuration.skipDependenciesUpdates: true // <-- example of customization of the default
```

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -104,19 +104,19 @@ public final class SwiftPMWorkspace {
...
 
     self.workspace = try Workspace(
         fileSystem: fileSystem,
         location: location,
-        mirrors: mirrorsConfiguration.mirrors,
-        customManifestLoader: ManifestLoader(toolchain: toolchain.configuration, cacheDir: location.workingDirectory),
-        resolverUpdateEnabled: false
+        configuration: configuration,
+        customHostToolchain: toolchain,
+        customManifestLoader: ManifestLoader(toolchain: toolchain.configuration, cacheDir: location.workingDirectory)
```
tomerd, line 119: I wonder if we still need a custom manifest loader now that we are providing a custom toolchain.

ManifestLoader has only two argument: the toolchain (which we now provide explicitly) and the cacheDir. I think in the case of LSP we can use the default SwiftPM cache dir so that would mean we can just drop this argument.

tomerd:
looks good. some ideas inline

tomerd:
> I'm also not completely clear how our mirrors setup differs from the default. The last comment on this from you https://github.com/apple/sourcekit-lsp/pull/427#discussion_r709553582 implied that we still need it. Is that still true? The default implementation of mirrors looks reasonable to me, but it's a bit hard to trace through exactly how it would differ from what we're currently doing.

I think the way you have it is correct. the old initializers there was no link between the location and the mirrors, but that ws part of the changes that led up to the new initializers

benlangmuir:
Thanks @tomerd, updated based on your suggestions

benlangmuir:
@swift-ci please test

tomerd (approved):
lgtm

shahmishal:
@swift-ci please test Linux

shahmishal:
@swift-ci please test macOS

ahoppen:
What did the mirrors configuration do? I looked inside the SwiftPM repo but couldn’t find any documentation on it.

benlangmuir:
It was added here: https://github.com/apple/sourcekit-lsp/pull/299

tomerd:
correct, we now have SwiftPM handle this detail internally base on the Location that is initialized to the project root

ahoppen (approved):

PR #459, opened 2022-02-24
Only send response to ShutdownRequest when all toolchain servers have shut down

ahoppen:
Especially in test we would sometimes end up in situations where the SourceKitServer (i.e. the main sourcekit-lsp handler) is being deallocated before clangd has finished shutting down. This can result in clangd sending us notifications while SourceKitServer is halfway deallocated and thus cause crashes due to memory corruption.

To fix the issue, wait until all toolchain servers have shut down before sending a response to the ShutdownRequest.

This should fix SourceKit-LSP test failures on Linux like https://ci.swift.org/job/swift-package-manager-Linux-smoke-test/93/console

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -647,11 +647,21 @@ extension SourceKitServer {
 
   func shutdown(_ request: Request<ShutdownRequest>) {
     _prepareForExit()
+    let shutdownGroup = DispatchGroup()
     for service in languageService.values {
-      service.shutdown()
+      shutdownGroup.enter()
+      service.shutdown() {
+        shutdownGroup.leave()
+      }
     }
     languageService = [:]
-    request.reply(VoidResponse())
+    // Wait for all services to shut down before sending the shutdown response.
+    // Otherwise we might terminate sourcekit-lsp while it still has open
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -647,8 +647,18 @@ extension SourceKitServer {
...
+      service.shutdown() {
+        sema.signal()
+      }
+    }
+    // Wait for all services to shut down before sending the shutdown response.
+    // Otherwise we might terminate sourcekit-lsp while it still has open
+    // connections to the toolchain servers, which could send messages to
+    // sourcekit-lsp while it is being deallocated, causing crashes.
+    for _ in languageService.values {
+      sema.wait()
```
benlangmuir, line 661: Better to use a DispatchGroup instead of multiple waiting on a counting semaphore.  Also, is there a reason you want to block the server queue, or would it be okay to have the rest of this code run async? You can have the DispatchGroup call you back instead of blocking.

ahoppen: DispatchGroup is definitely the better solution. I didn’t know about it.

ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #460, opened 2022-02-27
build: bump CMake requirement, remove workarounds

compnerd:
The swift project uses a minimum of CMake 3.19.6, and the current CMake
release is 3.22.  Update the CMake requirement to a newer version and
remove the workaround for older releases of CMake.


compnerd:
@swift-ci please test

ahoppen:
Would it make sense to bump the required CMake version all the way to 3.19.6 to match the requirement in the main Swift project?

compnerd:
Sure, I've got no objections to that, I just wanted to bump the minimum to remove the workaround.  I suspect that if we are going to raise it further, I can potentially remove more workarounds.

ahoppen:
Sounds good to me. I think in general, it’s a good idea if sourcekit-lsp’s CMake requirement matches that of the main Swift project.

ahoppen (approved):

PR #461, opened 2022-02-27
build: force link against dependencies

compnerd:
This will add an additional link request for dispatch and Foundation
libraries.  These are really required on non-Darwin targets, and should
be satisfied either by the library search path or by explicitly
indicating where the dependencies can be found.

### DIFF
--- a/Sources/LSPLogging/CMakeLists.txt
+++ b/Sources/LSPLogging/CMakeLists.txt
@@ -3,12 +3,8 @@ add_library(LSPLogging
   Logging.swift)
 set_target_properties(LSPLogging PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  if(Foundation_FOUND)
-    target_link_libraries(LSPLogging PRIVATE
-      Foundation)
-  endif()
-endif()
+target_link_libraries(LSPLogging PRIVATE
+  $<$<NOT:$<PLATFORM_ID:Darwin>>:Foundation>)
 
 if(BUILD_SHARED_LIBS)
   get_swift_host_arch(swift_arch)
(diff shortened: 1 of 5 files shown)


compnerd:
@swift-ci please test

ahoppen:
Linux failure looks unrelated. I’m investigating it.

@swift-ci Please test Linux

ahoppen:
Do we have documentation how to build sourcekit-lsp using CMake? If building sourcekit-lsp using CMake is currently supported on macOS and Linux, I’d like to make sure it still works after this change.

compnerd:
Hmm, I thought we did?  Either way, it's really rather simple, the complicated part is actually sitting for the build.

~~~
cmake -B out -D SwiftSystem_DIR=/BinaryCache/5/cmake/modules -D TSC_DIR=/BinaryCache/6/cmake/modules -D LLBuild_DIR=/BinaryCache/7/cmake/modules -D ArgumentParser_DIR=/BinaryCache/9/cmake/modules -D SwiftCollections_DIR=/BinaryCache/12/cmake/modules -D SwiftPM_DIR=/BinaryCache/13/cmake/modules -D IndexStoreDB_DIR=/BinaryCache/14/cmake/modules -G Ninja -S /SourceCache/sourcekit-lsp
~~~

This obviously will need lots of changes.  It just assumes that you have a previous build of the various dependencies, a swift toolchain in your path (including Foundation, dispatch available so they do not need to be rebuilt or passed in, though you _could_ explicit pass in pointers to their builds as well).  It will just put the output into `out` and defaults to debug (you can use `-D CMAKE_BUILD_TYPE=Release` for a release mode build).

ahoppen (approved):
I discussed this with @benlangmuir and the CMake build is really only used to build sourcekit-lsp for Windows. We don’t have any infrastructure that is exercising the CMake build on any other platform, so I’m happy to take this without verifying that it works on Linux/macOS. Should we ever need the CMake build on Linux/macOS (which I currently doubt), we can go through any possible issues at that point.

PR #462, opened 2022-02-28
SKCore: make a couple of members more accessible

compnerd:
This opens up access to `requestHandlers` and `notificationHandlers` as
it gets inlined through `_register` across module boundaries in
`SourceKitLSP`.  Making this more accessible enables building
SourceKit-LSP as a statically linked binary on Windows.

### DIFF
--- a/Sources/SKCore/LanguageServer.swift
+++ b/Sources/SKCore/LanguageServer.swift
@@ -39,9 +39,9 @@ open class LanguageServerEndpoint {
   /// All incoming requests start on this queue, but should reply or move to another queue as soon as possible to avoid blocking.
   public let queue: DispatchQueue = DispatchQueue(label: "language-server-queue", qos: .userInitiated)
 
-  private var requestHandlers: [ObjectIdentifier: Any] = [:]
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

benlangmuir:
I don't understand why this is necessary.  If this is needed to satisfy linker visibility requirements, doesn't it represent a general issue that should be treated as a compiler bug?

compnerd:
It isn't linker visibility, the accessors are simply not generated at all, which are expected to be generated as there are no uses of it in the module itself.

benlangmuir:
How are the accessors being referenced from outside the module without it being a bug to not generate them though?  If it's via inlining it seems like either the accessors should be generated, or the inlining should not have been allowed.

compnerd:
Turns out it is in the vtable entries (the full type metadata contains the pointers).  I'll see if I can figure out why the compiler is not emitting the functions.

compnerd:
I think I understand what is going on, and that is a compiler issue.  Sorry about the noise.

PR #463, opened 2022-03-02
build: build SourceKit-LSP with static linking

compnerd:
This adjusts the sourcekit-lsp build to use static linking for the
internal libraries.  It is not currently possible to build
SourceKitLSP.dll as that requires re-exporting the interfaces from the
consumed modules.  However, this allows us to reduce the overall size of
the distribution of SourceKit-LSP by ~1 MiB and reduces the
distributed file set.  The values here assume partial static linking of
swift-package-manager, which helps reduce the total size.

Before:

~~~
   228,352 BuildServerProtocol.dll
 1,773,056 LanguageServerProtocol.dll
   114,688 LanguageServerProtocolJSONRPC.dll
    49,152 LSPLogging.dll
   262,656 SKCore.dll
    54,784 SKSupport.dll
    80,896 SKSwiftPMWorkspace.dll
   150,528 SourceKitD.dll
   645,632 SourceKitLSP.dll

    70,144 sourcekit-lsp.exe

 3,429,888 bytes
~~~

After:

~~~
 2,416,640 sourcekit-lsp.exe

 2,416,640 bytes
~~~

### DIFF
--- a/Sources/BuildServerProtocol/CMakeLists.txt
+++ b/Sources/BuildServerProtocol/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_library(BuildServerProtocol
+add_library(BuildServerProtocol STATIC
   BuildTargets.swift
   FileOptions.swift
   InitializeBuild.swift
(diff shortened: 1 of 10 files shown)


compnerd:
@swift-ci please test

benlangmuir (approved):
LGTM. This also brings Windows closer to other platforms where we already link sourcekit-lsp statically.

ahoppen (approved):

PR #466, opened 2022-03-13
Import `CRT` to resolve `free` on Windows

stevapple:
```
S:\sourcekit-lsp\Sources\SourceKitD\SKDResponse.swift:55:13: error: cannot find 'free' in scope
    defer { free(ptr) }
            ^~~~
S:\sourcekit-lsp\Sources\SourceKitD\SKDResponseArray.swift:65:13: error: cannot find 'free' in scope
    defer { free(ptr) }
            ^~~~
S:\sourcekit-lsp\Sources\SourceKitD\SKDRequestArray.swift:39:13: error: cannot find 'free' in scope
    defer { free(ptr) }
            ^~~~
S:\sourcekit-lsp\Sources\SourceKitD\SKDRequestDictionary.swift:64:13: error: cannot find 'free' in scope
    defer { free(ptr) }
            ^~~~
S:\sourcekit-lsp\Sources\SourceKitD\SKDResponseDictionary.swift:64:13: error: cannot find 'free' in scope
    defer { free(ptr) }
```

### DIFF
--- a/Sources/SourceKitD/SKDRequestArray.swift
+++ b/Sources/SourceKitD/SKDRequestArray.swift
@@ -13,6 +13,8 @@
 import Csourcekitd
 #if canImport(Glibc)
 import Glibc
+#elseif canImport(CRT)
+import CRT
 #endif
 
 public final class SKDRequestArray {
--- a/Sources/SourceKitD/SKDRequestDictionary.swift
+++ b/Sources/SourceKitD/SKDRequestDictionary.swift
@@ -13,6 +13,8 @@
 import Csourcekitd
 #if canImport(Glibc)
 import Glibc
+#elseif canImport(CRT)
+import CRT
 #endif
 
 public final class SKDRequestDictionary {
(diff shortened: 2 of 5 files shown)


compnerd (approved):
Odd that we haven't hit this so far, but seems like a reasonable thing.

compnerd:
@swift-ci please test

compnerd:
@stevapple how come this is not needed on Darwin btw?  I don't see an `import Darwn` here.

stevapple:
@compnerd I didn’t figure out yet. It may share the same reason why we didn’t run into it before on Windows.

I guess this could be related with re-exported CStdlib symbols from some modules.

compnerd:
Oh, it almost certainly is related to re-exporting.  But, it would be good to understand what's the root cause for the change in that.

stevapple:
> But, it would be good to understand what's the root cause for the change in that.

May possibly be related to my fix on C header visibility… I opened this and apple/swift-corelibs-foundation#3151 as preparation to test that fix.

ahoppen:
Windows noob question: Is `CRT` Window’s C standard library equivalent that defines `free` etc?

ahoppen (approved):
The change sounds reasonable to me and I trust @compnerd’s judgment that it makes sense from a Windows perspective.

PR #469, opened 2022-04-11
Increase timeout for semantic diagnostics in tests

ahoppen:
AFAICT getting semantic diagnostics takes more than 5 seconds if a specific SDK/compiler combination is used for the first time, increase the timeout to make the test pass.


ahoppen:
@swift-ci Please smoke test

ahoppen:
@swift-ci Please test

ahoppen:
I think that makes sense. I’ll take a look at it in a follow-up PR (rdar://91615376)

PR #473, opened 2022-04-20
Support opening multiple workspaces in a single sourcekit-lsp instance

ahoppen:
rdar://89083082 [SR-15877]

### DIFF
--- a/Sources/SKCore/BuildServerBuildSystem.swift
+++ b/Sources/SKCore/BuildServerBuildSystem.swift
@@ -221,6 +221,26 @@ extension BuildServerBuildSystem: BuildSystem {
   }
 
   public func filesDidChange(_ events: [FileEvent]) {}
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
+    guard let fileUrl = uri.fileURL else {
+      return .unhandled
+    }
+
+    // FIXME: We should not make any assumptions about which files the build server can handle.
+    // Instead we should query the build server which files it can handle (#492).
+    let path = AbsolutePath(fileUrl.path)
+    if projectRoot.isAncestorOfOrEqual(to: path) {
+      return .handled
+    }
+
+    let realpath = resolveSymlinks(path)
+    if realpath != path, projectRoot.isAncestorOfOrEqual(to: realpath) {
+      return .handled
+    }
+
+    return .unhandled
+  }
 }
 
 private func loadBuildServerConfig(path: AbsolutePath, fileSystem: FileSystem) throws -> BuildServerConfig {
--- a/Sources/SKCore/BuildSystem.swift
+++ b/Sources/SKCore/BuildSystem.swift
@@ -14,6 +14,18 @@ import BuildServerProtocol
 import LanguageServerProtocol
 import TSCBasic
 
+/// Defines how well a `BuildSystem` can handle a file with a given URI.
+public enum FileHandlingCapability: Comparable {
+  /// The build system can't handle the file at all
+  case unhandled
+
+  /// The build system has fallback build settings for the file
+  case fallback
+
+  /// The build system knows how to handle the file
+  case handled
+}
+
 /// Provider of FileBuildSettings and other build-related information.
 ///
 /// The primary role of the build system is to answer queries for
@@ -62,6 +74,8 @@ public protocol BuildSystem: AnyObject {
 
   /// Called when files in the project change.
   func filesDidChange(_ events: [FileEvent])
+
+  func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability
 }
 
 public let buildTargetsNotSupported =
--- a/Sources/SKCore/CompilationDatabaseBuildSystem.swift
+++ b/Sources/SKCore/CompilationDatabaseBuildSystem.swift
@@ -187,6 +187,19 @@ extension CompilationDatabaseBuildSystem: BuildSystem {
       }
     }
   }
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
+    guard let fileUrl = uri.fileURL else {
+      return .unhandled
+    }
+    return queue.sync {
+      if database(for: fileUrl) != nil {
+        return .handled
+      } else {
+        return .unhandled
+      }
+    }
+  }
 }
 
 extension CompilationDatabaseBuildSystem {
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -361,6 +361,19 @@ extension SwiftPMWorkspace: SKCore.BuildSystem {
       }
     }
   }
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
+    guard let fileUrl = uri.fileURL else {
+      return .unhandled
+    }
+    return self.queue.sync {
+      if targetDescription(for: AbsolutePath(fileUrl.path)) != nil {
+        return .handled
+      } else {
+        return .unhandled
+      }
+    }
+  }
 }
 
 extension SwiftPMWorkspace {
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -222,6 +222,7 @@ extension SwiftPMWorkspace {
       }
     }
     delegate.fileBuildSettingsChanged(changedFiles)
+    delegate.fileHandlingCapabilityChanged()
   }
 }
 
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -421,10 +421,10 @@ extension SourceKitServer: BuildSystemDelegate {
       // Split the changedFiles into the workspaces they belong to.
       // Then invoke affectedOpenDocumentsForChangeSet for each workspace with its affected files.
       let changedFilesAndWorkspace = changedFiles.map({
-        return ($0, self.workspaceForDocument(uri: $0))
+        return (uri: $0, workspace: self.workspaceForDocument(uri: $0))
       })
       for workspace in self.workspaces {
-        let changedFilesForWorkspace = Set(changedFilesAndWorkspace.filter({ $0.1 === workspace }).map(\.0))
+        let changedFilesForWorkspace = Set(changedFilesAndWorkspace.filter({ $0.workspace === workspace }).map(\.uri))
         if changedFilesForWorkspace.isEmpty {
           continue
         }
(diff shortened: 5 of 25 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -65,24 +83,53 @@ public final class SourceKitServer: LanguageServer {
...
+
+  /// Returns the most specific workspace for the given `uri`.
+  /// This is done as follows:
+  ///  - The rootUri of the workspaces and `uri` are canonicalized to remove symlinks, '..' etc.
+  ///  - We select the workspace whose canonicalized rootUri is the longest prefix match to the canonicalized path of `uri`
+  /// This method returns `nil` if `uri` is in no open workspace.
+  public func workspaceForDocument(uri: DocumentURI) -> Workspace? {
+    let canonicalizedFilePath = uri.canonicalizedPseudoPath
+    var bestWorkspaceMatch: (workspace: Workspace, canonicalizedRootPath: String)? = nil
+    for (workspace, canonicalizedWorkspaceRootPath) in _workspaces where canonicalizedFilePath.hasPrefix(canonicalizedWorkspaceRootPath) {
```
DavidGoldman, line 105: If there's only one workspace, should we maintain the old behavior? I think we would regress in workspaces that give a rootUri but can the user could open files outside of there (e.g. derived files, etc)

DavidGoldman: Also not sure how safe this is to generalize - for SwiftPM you might be able to have multiple BuildSystem instances, but for a monorepo with a unified build system you can really only have one

DavidGoldman: > I was thinking the same but decided against it because I thought it was confusing if opening a second workspace changed the way that files were matched to these workspaces. I also assumed that opening files outside of that rootUri isn’t common, but maybe that assumption is wrong. I wouldn’t be terribly opposed to matching all files to a workspace if only one is open.

Yeah, it seems like we need a way to explicitly force the one-workspace behavior. Is it possible that `workspaceFolders` are added/removed after initialization, I think so, right? So in that case maybe a flag to force the old behavior is OK?

> Would issues would you expect here? Having multiple build system instances for unrelated projects seems like it should be possible to me. And for a monorepo I would expect that the rootUri is the base of the monorepo? Or am I making wrong assumptions here?

It's possible that the user has multiple folders of the monorepo open (2 `workspaceFolders`), e.g. `llvm-project/clang` and `llvm-project/clang-tools-extra`, but the root-uri we'd want to use is `llvm-project` and only share a single instance, not have two. IIUC correctly, with this implementation as is, we have no way to support this use case, right?

ahoppen: > Yeah, it seems like we need a way to explicitly force the one-workspace behavior. Is it possible that `workspaceFolders` are added/removed after initialization, I think so, right? So in that case maybe a flag to force the old behavior is OK?

Sleeping over it, I think maintaining the current behavior if only a single workspace is open is reasonable. If you only have one workspace, there’s no reason to assume that a file doesn’t belong to that workspace.

I also thought about adding a similar fallback logic for multiple workspaces but I think falling back to the first or last workspace will be surprising.

> It's possible that the user has multiple folders of the monorepo open (2 `workspaceFolders`), e.g. `llvm-project/clang` and `llvm-project/clang-tools-extra`, but the root-uri we'd want to use is `llvm-project` and only share a single instance, not have two. IIUC correctly, with this implementation as is, we have no way to support this use case, right?

Hmm, this really makes things tricky because clangd doesn’t support multiple workspace folders. I can think of multiple options
1. We refuse to open files from two different workspace folders in clangd. That is, you open a file from workspace folder A, it succeeds, you open a file from workspace folder B and it fails with an error because workspace folder A is already open in clangd and clangd can’t handle multiple workspace folders. This doesn’t support your use case but your use case isn’t supported today either, so that shouldn’t be a regression.
2. Create a separate clangd instance for the different workspace folders. I am not entirely sure how well multiple clangd instance can co-exist on the same project, do you have any experience with that @DavidGoldman? If there’s no issue, I think this would be a very elegant solution.
3. Have a single `clangd` instance but don’t open it with the rootUri of any particular workspace, but with the deprecated `rootUri` property of the `initialize` request. This would support your use case but I really hate this option because it breaks the assumption that the deprecated `rootUri` property in the initialize request is going to be ignored if you pass `workspaceFolders`

benlangmuir: > I think we would regress in workspaces that give a rootUri but the user could open files outside of there (e.g. derived files, etc

This does seem important to me. Files in a build directory are either going to be opened directly if the editor doesn't know anything about them, or else the editor might want to add a root directory for the build directory, and in either case the correct interpretation is to consider them part of the same overall workspace.  If you have multiple build systems across workspace roots, that means we need to figure out which one the file belongs to to the best of our ability.  This seems kind of orthogonal from workspace roots anyway, since you could have multiple build systems in a single root.  I wonder if we should keep a single `Workspace` and have the build system layer deal with the multi-root support.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SKCore/CompilationDatabaseBuildSystem.swift
```diff
@@ -111,6 +111,17 @@ extension CompilationDatabaseBuildSystem: BuildSystem {
   }
 
   public func filesDidChange(_ events: [FileEvent]) {}
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
+    guard let fileUrl = uri.fileURL else {
+      return .handled
```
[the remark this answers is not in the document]
ahoppen: Yes. Thanks for catching this.

Sources/SKCore/BuildSystem.swift
```diff
@@ -62,6 +74,8 @@ public protocol BuildSystem: AnyObject {
 
   /// Called when files in the project change.
   func filesDidChange(_ events: [FileEvent])
+
+  func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability
```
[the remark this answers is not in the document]
ahoppen: If we had async/await, I would say yes. 

Given that we don’t use async/await in SourceKit-LSP yet, I would like to avoid making it async. If `fileHandlingCapability` becomes async, `workspaceForDocument` would also need to be async and thus all request handling functions that call `workspaceForDocument` need to we wrapped in a completion handler – it’s not impossible but does uglify the code. 

Which performance part are you particularly concerned about? AFAICT the most expensive parts of the implementations are and none of them strike me as worrisome enough to justify having to take care of the completion handler in `SourceKitServer`.
- `BuildServerBuildSystem`: Resolve symlinks + a little substring matching
- `BuildSystemManager`: Just calls `max`
- `CompilationDatabaseBuildSystem`: Load compilation database, but should only happen once
- `FallbackBuildSystem`: Always returns `.fallback`
- `SwiftPMWorkspace`: Resolve symlinks + dictionary lookup

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -361,6 +362,18 @@ extension SwiftPMWorkspace: SKCore.BuildSystem {
       }
     }
   }
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
+    guard let fileUrl = uri.fileURL else {
+      return .unhandled
+    }
+    let absoluteFilePath = resolveSymlinks(AbsolutePath(fileUrl.path))
```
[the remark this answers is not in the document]
ahoppen: Good idea 👍

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -24,31 +24,68 @@ import PackageLoading
...
-  public var workspace: Workspace?
+  private let documentManager = DocumentManager()
+
+  /// **Public for testing**
+  public var _documentManager: DocumentManager {
+    return documentManager
+  }
+
+  /// Queue guarding `workspaces` and `uriToWorkspaceCache`.
+  let workspaceQueue = DispatchQueue(label: "", qos: .userInitiated)
```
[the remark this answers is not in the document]
ahoppen: Looking through the code again, I can’t remember any reason why I introduced this queue in the first place instead of just using `self.queue` so I just removed it.

ahoppen:
@swift-ci Please test

Sources/SKCore/BuildSystem.swift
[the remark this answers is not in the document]
benlangmuir: Okay, let's keep it sync unless it becomes an issue in practice.  

> BuildServerBuildSystem: Resolve symlinks + a little substring matching

This one I'm not sure is correct without querying the server.

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -68,7 +68,7 @@ public final class SwiftPMWorkspace {
   /// - `packageGraph`
   /// - `fileToTarget`
   /// - `sourceDirToTarget`
-  let queue: DispatchQueue = .init(label: "SwiftPMWorkspace.queue", qos: .utility)
+  let queue: DispatchQueue = .init(label: "SwiftPMWorkspace.queue", qos: .userInitiated)
```
benlangmuir, line 71: We don't generally want reloading the package to be run at high priority, or it risks hurting performance in the rest of the system.  We also don't need this for the current change, because
the `self.queue.sync` in `fileHandlingCapability` will cause priority donation if the calling code is running at higher priority than utility.

benlangmuir (approved):
LGTM other than the QoS change in swiftpm workspace.

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
ahoppen: Oh, I didn't know that GCD automatically increases the priority if the caller has a priority higher than the called queue. That’s really good to know 👍 It was the only reason why I increased priority here. I’ll revert it.

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -221,6 +221,21 @@ extension BuildServerBuildSystem: BuildSystem {
   }
 
   public func filesDidChange(_ events: [FileEvent]) {}
+
+  public func fileHandlingCapability(for uri: DocumentURI) -> FileHandlingCapability {
```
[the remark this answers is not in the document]
ahoppen: I added a FIXME and opened https://github.com/apple/sourcekit-lsp/issues/492 for now.

ahoppen:
@swift-ci Please test

DavidGoldman:
What's the reasoning behavior having multiple clangd instances - one per workspace?

ahoppen:
Yes, we launch a separate `clangd` instance for each workspace folder.

PR #474, opened 2022-04-22
Fix a crash on Linux if a ShutdownRequest was sent without params

ahoppen:
Workaround for SR-16097 to avoid hitting SR-16095: VSCode will send shutdown requests without a 'params' key. On macOS getting the superDecoder for a non-existent container returns an empty decoder, on Linux it throws `DecodingError.keyNotFound`, causing a crash.

Perform a targeted fix: If we don't have 'params' and the request is a ShutdownRequest, manually create a `ShutdownRequest` without any parameters.

rdar://91288093

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
@@ -77,7 +77,15 @@ extension JSONRPCMessage: Codable {
           throw MessageDecodingError.methodNotFound(method)
         }
 
-        let params = try messageType.init(from: container.superDecoder(forKey: .params))
+        let params: _RequestType
+        do {
+          params = try messageType.init(from: container.superDecoder(forKey: .params))
+        } catch DecodingError.keyNotFound(CodingKeys.params, _) where messageType == ShutdownRequest.self {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
```diff
@@ -77,7 +77,15 @@ extension JSONRPCMessage: Codable {
           throw MessageDecodingError.methodNotFound(method)
         }
 
-        let params = try messageType.init(from: container.superDecoder(forKey: .params))
+        let params: _RequestType
+        do {
+          params = try messageType.init(from: container.superDecoder(forKey: .params))
+        } catch DecodingError.keyNotFound(CodingKeys.params, _) where messageType == ShutdownRequest.self {
```
benlangmuir, line 83: This is kind of a layering violation to have knowledge of specific requests, but I think it's fine for a workaround.

benlangmuir (approved):

Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
ahoppen: Filed rdar://92254952 to track removing the workaround when SR-16097 is resolved.

PR #476, opened 2022-04-22
[5.7] Fix a crash on Linux if a ShutdownRequest was sent without params

ahoppen:
* **Explanation of the issue**: VSCode is shutting down the sourcekit-lsp server when switching workspaces. On Linux sourcekit-lsp fails to process the shutdown request  and causes VSCode to abort the shutdown process at a point it cannot recover from, thus making it impossible to restart the language server. To recover semantic functionality after switching workspaces, VSCode needs to be restarted. 
* **Explanation of the fix**: Perform a targeted fix: If we don't have 'params' and the request is a ShutdownRequest, manually create a `ShutdownRequest` without any parameters. This issue does not occur on macOS because the underlying bug (SR-16097) only affects Linux: VSCode will send shutdown requests without a 'params' key. On macOS getting the superDecoder for a non-existent container returns an empty decoder, on Linux it throws `DecodingError.keyNotFound`, causing a crash. 
* **Scope**: Only affects the shutdown requests that failed to deserialize before
* **Risk**: Very low
* **Testing**: Added a test case
* **Issue**: rdar://91288093
* **Reviewer**: @benlangmuir on https://github.com/apple/sourcekit-lsp/pull/474

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/MessageCoding.swift
@@ -77,7 +77,15 @@ extension JSONRPCMessage: Codable {
           throw MessageDecodingError.methodNotFound(method)
         }
 
-        let params = try messageType.init(from: container.superDecoder(forKey: .params))
+        let params: _RequestType
+        do {
+          params = try messageType.init(from: container.superDecoder(forKey: .params))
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

airspeedswift (approved):

PR #478, opened 2022-04-22
Increase timeout for semantic diagnostics in tests a little bit more

ahoppen:
30 seconds wasn’t sufficient to fix the failure in CI, 60 should be enough.


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #470, opened 2022-04-19
[5.7] Increase timeout for semantic diagnostics in tests

ahoppen:
Cherry-pick https://github.com/apple/sourcekit-lsp/pull/469 to release/5.7.

---

AFAICT getting semantic diagnostics takes more than 5 seconds if a specific SDK/compiler combination is used for the first time, increase the timeout to make the test pass.

PR #479, opened 2022-04-22
[5.7] Increase timeout for semantic diagnostics in tests a little bit more

ahoppen:
* **Explanation**: Increasing the timeout for this test from 5 seconds to 30 seconds in https://github.com/apple/sourcekit-lsp/pull/470 fixed the issue locally, but not in CI. I verified that 60 second is enough.
* **Scope**: SourceKit-LSP test case
* **Risk**: Super low (only affects tests)
* **Testing**: Verified that tests now pass in CI
* **Reviewer**: Waiting for review on https://github.com/apple/sourcekit-lsp/pull/478


ahoppen:
@swift-ci Please test

PR #481, opened 2022-04-26
Translate placeholders to LSP snippets in CodeActions

ahoppen:
Previously, we would insert `<#code#>` placeholders when inserting missing protocol requirements (both via the Fix-It and the refactor command). Translate them to LSP placeholders before inserting them.

rdar://92161144


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #482, opened 2022-04-27
Enable crash recovery tests again

ahoppen:
It turned out that the test failures were not specific to the crash recovery tests and we needed to disable all SourceKit-LSP tests on Linux. There’s no need to disable the crash recovery tests specifically anymore


benlangmuir (approved):

benlangmuir:
> and we needed to disable all SourceKit-LSP tests on Linux.

That seems bad

Issue #486, opened 2022-04-27
Watch compilation_database.json and compile_flags.txt


ahoppen:
Similar to how we reload the package if Package.swift is changed, we could also watch compilation_database.json and compile_flags.txt and notify the build system delegate that build settings changed if these files are modified.

rdar://92388223

ahoppen:
Implemented in https://github.com/apple/sourcekit-lsp/pull/487.

PR #487, opened 2022-04-28
Watch for changes to `compile_commands.json` and `compile_flags.txt`

ahoppen:
Similar to how we reload the package if Package.swift is changed, we also watch `compilation_database.json` and `compile_flags.txt` and notify the build system delegate that build settings changed if these files are modified.

rdar://92388223  [#486]


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #490, opened 2022-04-28
Only run tests in parallel on Darwin

ahoppen:
Running tests in parallel causes nondeterministic failures on Linux (rdar://92260631). Disable parallel testing on all platforms except Darwin until that issue has been fixed.

This might still be caused by the same underlying issue as https://github.com/apple/sourcekit-lsp/pull/337.


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #554, opened 2022-05-13
Only run tests in parallel on Linux again

ahoppen:
The underlying issue in the SwiftPM test runner has been fixed.

rdar://92262856


ahoppen:
@swift-ci Please test

PR #556, opened 2022-05-24
Support creation of fat binaries that run on x86_64 and arm64

ahoppen:
Currently, when building an open source toolchain, SourceKit-LSP is only built for x86_64.  Copy the necessary cross-compilation parts from SwiftPM’s build script to also produce a fat sourcekit-lsp executable for both x86_64 and arm64.

rdar://78039145

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -9,6 +9,29 @@ import re
 import shutil
 import subprocess
 import sys
+import json
+
+def error(message):
+  print("--- %s: error: %s" % (os.path.basename(sys.argv[0]), message), file=sys.stderr)
+  sys.stderr.flush()
+  raise SystemExit(1)
+
+def get_build_target(swift_exec, args):
+  """Returns the target-triple of the current machine or for cross-compilation."""
+  try:
+    command = [swift_exec, '-print-target-info']
+    target_info_json = subprocess.check_output(command,
+                         stderr=subprocess.PIPE, universal_newlines=True).strip()
+    args.target_info = json.loads(target_info_json)
+    if platform.system() == 'Darwin':
+      return args.target_info["target"]["unversionedTriple"]
+    return args.target_info["target"]["triple"]
+  except Exception as e:
+    # Temporary fallback for Darwin.
+    if platform.system() == 'Darwin':
+      return 'x86_64-apple-macosx'
+    else:
+      error(str(e))
 
 def swiftpm(action, swift_exec, swiftpm_args, env=None):
   cmd = [swift_exec, action] + swiftpm_args
@@ -21,7 +44,7 @@ def swiftpm_bin_path(swift_exec, swiftpm_args, env=None):
   print(' '.join(cmd))
   return subprocess.check_output(cmd, env=env, universal_newlines=True).strip()
 
-def get_swiftpm_options(args):
+def get_swiftpm_options(swift_exec, args):
   swiftpm_args = [
     '--package-path', args.package_path,
     '--build-path', args.build_path,
@@ -65,8 +88,15 @@ def get_swiftpm_options(args):
       '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/linux',
     ]
 
+  build_target = get_build_target(swift_exec, args)
   if args.cross_compile_host:
-    swiftpm_args += ['--destination', args.cross_compile_config]
+    if build_target == 'x86_64-apple-macosx' and args.cross_compile_host == "macosx-arm64":
+      swiftpm_args += ["--arch", "x86_64", "--arch", "arm64"]
+    elif re.match('android-', args.cross_compile_host):
+      print('Cross-compiling for %s' % args.cross_compile_host)
+      swiftpm_args += ['--destination', args.cross_compile_config]
+    else:
+      error("cannot cross-compile for %s" % args.cross_compile_host)
 
   return swiftpm_args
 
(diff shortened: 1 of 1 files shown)


Utilities/build-script-helper.py
```diff
@@ -149,7 +181,7 @@ def add_common_args(parser):
     parser.add_argument('--sanitize-all', action='store_true', help='build using every available sanitizer in sub-directories of build path')
     parser.add_argument('--no-clean', action='store_true', help='Don\'t clean the build directory prior to performing the action')
     parser.add_argument('--verbose', '-v', action='store_true', help='enable verbose output')
-    parser.add_argument('--cross-compile-host', help='cross-compile for another host instead')
+    parser.add_argument('--cross-compile-hosts', help='List of cross compile hosts targets', default=[])
```
benlangmuir, line 184: Is this actually a list, or is it just a string? The default value is `[]`, but I would have expected to see `nargs=*` or similar.

Utilities/build-script-helper.py
```diff
@@ -51,8 +77,8 @@ def get_swiftpm_options(args):
       os.path.join(args.toolchain, 'lib', 'swift', 'Block'),
     ]
 
-  if 'ANDROID_DATA' in os.environ or (args.cross_compile_host and re.match(
-    'android-', args.cross_compile_host)):
+  if 'ANDROID_DATA' in os.environ or (args.cross_compile_hosts and re.match(
```
benlangmuir, line 80: Have you confirmed this line works? Matching a list doesn't seem to work for me.  Or maybe this is only working because it's actually a string right now (see my other comment on the argparse code).

```python
>>> re.match('as', 'asdf')
<re.Match object; span=(0, 2), match='as'>

>>> re.match('as', ['asdf'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/AppleInternal/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/re.py", line 191, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object
```

Utilities/build-script-helper.py
```diff
@@ -65,8 +91,12 @@ def get_swiftpm_options(args):
       '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/linux',
     ]
 
+  build_target = get_build_target(swift_exec, args)
   if args.cross_compile_host:
-    swiftpm_args += ['--destination', args.cross_compile_config]
+    if build_target == 'x86_64-apple-macosx' and args.cross_compile_host == "macosx-arm64":
+        swiftpm_args += ["--arch", "x86_64", "--arch", "arm64"]
```
buttaface, line 97: Would you bring the two Android lines deleted below up here and move the config line deleted two lines above after printing the cross-compilation host? Otherwise, this is going to break [cross-compiling sourcekit-lsp for Android](https://github.com/termux/termux-packages/blob/master/packages/swift/build.sh#L162).

Utilities/build-script-helper.py
```diff
@@ -65,8 +91,14 @@ def get_swiftpm_options(args):
       '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/linux',
     ]
 
-  if args.cross_compile_host:
-    swiftpm_args += ['--destination', args.cross_compile_config]
+  build_target = get_build_target(args)
+  cross_compile_hosts = args.cross_compile_hosts
+  if build_target == 'x86_64-apple-macosx' and "macosx-arm64" in cross_compile_hosts:
```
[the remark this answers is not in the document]
ahoppen: We don’t have any build presets that cross compile from arm64 for x86_64, so I’m not entire sure what to look here (I would be guessing `macosx-x86_64`, but that’s really mostly guessing). So I’d prefer to stick with what SwiftPM does and only support cross-compilation from x86_64 to arm64 for now.

ahoppen:
@swift-ci Please test

ahoppen:
I simplified the PR a little bit to only take a single `cross-compile-host` (which seems like it was the only supported configuration anyway) and addressed your feedback. It now successfully produced a universal `sourcekit-lsp` binary and I’m happy with the current state. Could you take another look at it?

Utilities/build-script-helper.py
```diff
@@ -9,6 +9,32 @@
...
+  sys.stdout.flush()
+  raise SystemExit(1)
+
+def get_build_target(swift_exec, args, cross_compile=False):
+  """Returns the target-triple of the current machine or for cross-compilation."""
+  try:
+    command = [swift_exec, '-print-target-info']
+    if cross_compile:
+      cross_compile_json = json.load(open(args.cross_compile_config))
+      command += ['-target', cross_compile_json["target"]]
```
buttaface, line 25: These three lines and the `cross_compile` flag can be removed as they are unused on macOS. With SPM, [this flag was only set for Android to get the target directory](https://github.com/apple/swift-package-manager/blob/main/Utilities/bootstrap#L221), which is unneeded here.

ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -9,6 +9,29 @@
 import shutil
 import subprocess
 import sys
+import json
+
+def error(message):
```
benlangmuir, line 14: 😅 oops, guess we forgot to define this one

Utilities/build-script-helper.py
```diff
@@ -9,6 +9,29 @@
...
+  print("--- %s: error: %s" % (os.path.basename(sys.argv[0]), message))
```
benlangmuir, line 15: Should this go to stderr?

ahoppen: Good idea. SwiftPM (where I copied the function from) prints to stdout but stderr seems like the better choice.

ahoppen:
@swift-ci Please test

benlangmuir (approved):

Issue #557, opened 2022-05-26
Multi-workspace: Sourcekit-LSP can use the wrong workspace if one is inside another.


adam-fowler:
VSCode automatically adds workspace folders to the LSP server (sends `workspace/didChangeWorkspaceFolders` notification) for all workspace folders added to VSCode. If the VSCode workspace folder has subfolders that contain packages when these subfolders are added to the LSP server, they are ignored because SourceKit-LSP will always choose the workspace root folder, over the subfolder (because it was added first).

This causes an issue in the following example. Below is a example workspace structure. We have a `RootFolder` without a `Package.swift` and then we have two subfolders that both contain SwiftPM packages.

```
- RootFolder
  - folder1
    - Package.swift
    - ...
  - folder2
    - Package.swift
    - ...
```

When we load this project, VSCode automatically adds `RootFolder` as a LSP workspace folder. When we go to open either `folder1` or `folder2` the vscode-swift extension will then add that folder as a LSP workspace folder, but as indicated above SourceKit-LSP will ignore the subfolder because `RootFolder` takes precedence even though it doesn't hold a `Package.swift`. Could we get SourceKit-LSP to prioritise workspace folders that hold a Package.swift over folders that don't.

ahoppen:
rdar://94026341

ahoppen:
Fixed by https://github.com/apple/sourcekit-lsp/pull/558

PR #558, opened 2022-05-27
When receiving a didChangeWorkspaceFolders notificaiton, reopen documents that moved to a different workspace

ahoppen:
A didChangeWorkspaceFolders notification might cause documents that are already open to now be considered part of a different workspace. To reflect this, close documents whose workspace has changed in the old workspace and re-open them in the new workspace.

rdar://94026341 [#557]


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -900,6 +904,31 @@ extension SourceKitServer {
...
+    // For each document that has moved to a different workspace, close it in
+    // the old workspace and open it in the new workspace.
+    for docUri in self.documentManager.openDocuments {
+      let oldWorkspace = preChangeWorkspaces[docUri]
+      let newWorkspace = self.workspaceForDocument(uri: docUri)
+      if newWorkspace !== oldWorkspace {
+        guard let snapshot = documentManager.latestSnapshot(docUri) else {
+          continue
+        }
+        if let oldWorkspace {
```
benlangmuir, line 917: This syntax is not supported by the released version of swift (5.6.1), only by 5.7+.  We should keep using the old syntax for now so people can build sourcekit-lsp more easily.

ahoppen: I thought we required the latest Swift version to build SourceKit-LSP like we do for SwiftSyntax but I agree that it’s better to be conservative here because it’s reasonable for people to build SourceKit-LSP on its own.

ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #560, opened 2022-06-01
[5.7] When receiving a didChangeWorkspaceFolders notification, reopen documents that moved to a different workspace

ahoppen:
* **Explanation**: A didChangeWorkspaceFolders notification might cause documents that are already open to now be considered part of a different workspace. To reflect this, close documents whose workspace has changed in the old workspace and re-open them in the new workspace. Without this fix, these documents that switched workspaces weren’t getting any semantic functionality
* **Scope**: SourceKit-LSP after changing workspaces using the `didChangeWorkspaceFolders` notification
* **Risk**: I can’t think of any risk and the current behavior is pretty bad
* **Testing**: Added regression test case
* **Issue**: rdar://94026341 [#557]
* **Reviewer**: @benlangmuir on https://github.com/apple/sourcekit-lsp/pull/558

### DIFF
--- a/Sources/SKTestSupport/INPUTS/ChangeWorkspaceFolders/Package.swift
+++ b/Sources/SKTestSupport/INPUTS/ChangeWorkspaceFolders/Package.swift
@@ -1,0 +1,19 @@
+// swift-tools-version: 5.5
+
+import PackageDescription
+
+let package = Package(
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

PR #561, opened 2022-06-01
[5.7] Support creation of fat binaries that run on x86_64 and arm64

ahoppen:
* **Explanation**: Currently, when building an open source toolchain, SourceKit-LSP is only built for x86_64.  Copy the necessary cross-compilation parts from SwiftPM’s build script to also produce a fat sourcekit-lsp executable for both x86_64 and arm64.
* **Scope**: The way SourceKit-LSP is built in open source toolchains
* **Risk**: These changes to the build process are always a little risky, but I think we should be able to catch any issues once the next open source toolchain is released
* **Testing**: Verified that the binary produced by https://github.com/apple/swift/pull/59046 was fat (x86_64 and arm64) and ran on arm64
* **Issue**: rdar://78039145
* **Reviewer**: @benlangmuir on https://github.com/apple/sourcekit-lsp/pull/556

---

Companion of https://github.com/apple/swift/pull/59196.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -9,6 +9,29 @@ import re
 import shutil
 import subprocess
 import sys
+import json
+
+def error(message):
+  print("--- %s: error: %s" % (os.path.basename(sys.argv[0]), message), file=sys.stderr)
+  sys.stderr.flush()
+  raise SystemExit(1)
+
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

airspeedswift (approved):

PR #562, opened 2022-06-03
Support index store path remappings

DavidGoldman:
This allows sourcekit-lsp to make use of IndexStoreDB path remappings (https://github.com/apple/indexstore-db/pull/148) to remap canonical/remote paths into their local equivalents.

### DIFF
--- a/Sources/SKCore/PathPrefixMapping.swift
+++ b/Sources/SKCore/PathPrefixMapping.swift
@@ -2,10 +2,10 @@ import Foundation
 
 public struct PathPrefixMapping {
   /// Path prefix to be replaced, typically the canonical or hermetic path.
-  let original: String
+  public let original: String
 
   /// Replacement path prefix, typically the path on the local machine.
-  let replacement: String
+  public let replacement: String
 
   public init(original: String, replacement: String) {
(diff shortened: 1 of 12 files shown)


Sources/SKCore/PathPrefixMapping.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public struct PathPrefixMapping {
```
DavidGoldman, line 3: I could just use `PathMapping` from IndexStoreDB instead and it import it/depend upon it from SKCore, LMK which you'd prefer.

DavidGoldman:
@swift-ci please test

Sources/SKCore/PathPrefixMapping.swift
ahoppen: I don’t really have a strong opinion. If this struct stays as simple as it its, avoiding the dependency sounds good to me 👍

ahoppen (approved):
LGTM

DavidGoldman:
@swift-ci Please test

benlangmuir:
@swift-ci please test

PR #172, opened 2019-10-08
[refactoring] Disable "local rename"

benlangmuir:
This needs similar support to global rename before it actually works.
Disable for now.

PR #563, opened 2022-06-05
Clarify renaming capabilities

resolritter:
It's misleading to qualify "**Global** Rename" as not implemented because neither global nor local rename work. It's more straightforward and clear to say that renaming is not supported at all.

From what I can tell the current wording could be related to a time where local renaming used to work as a refactor action item (https://github.com/apple/sourcekit-lsp/pull/172); this is merely a guess since I was not following this repository's development back then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ SourceKit-LSP is still in early development, so you may run into rough edges wit
 | Find References | ✅ | |
 | Background Indexing | ❌ | Build project to update the index using [Indexing While Building](#indexing-while-building) |
 | Workspace Symbols | ✅ | |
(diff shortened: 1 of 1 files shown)


benlangmuir:
@swift-ci please test

benlangmuir (approved):

benlangmuir:
> From what I can tell the current wording could be related to a time where local renaming used to work as a refactor action item (https://github.com/apple/sourcekit-lsp/pull/172); this is merely a guess since I was not following this repository's development back then.

No, it never worked.  It was accidentally possible to trigger the underlying functionality for local rename from sourcekitd, but there was no way to specify the new name.

PR #565, opened 2022-06-06
allow custom per-workspace build flags via initializationOptions

### LINKED ISSUE #564: per-workspace build flags
From what I can tell by the source code, the CLI build flags are used for all workspaces:

https://github.com/apple/sourcekit-lsp/blob/1fdb15676051610d1546eb4ec241d0d27f799024/Sources/sourcekit-lsp/main.swift#L132

https://github.com/apple/sourcekit-lsp/blob/b8d8181a8ceda1b705f2dba0cc6971061868d986/Sources/SourceKitLSP/SourceKitServer.swift#L538

However some workspaces might need different build flags for e.g. different include paths.

Would it be feasible to detect a custom file in the workspace, or custom `Workspace => BuildOptions` dictionary to be provided in e.g. `initializationOptions`, for build flags customization? An example of the custom file approach is `.ccls` (https://github.com/MaskRay/ccls/wiki/Project-Setup#ccls-file).

resolritter:
close #564

Opening the PR earlier, before everything is ready, for feedback before continuing to the remaining tasks

# TODO

- [x] Implementation
- [ ] Documentation
- [ ] Tests

Example usage:

```json
{
  "initializationOptions": {
    "workspaces": {
      "/path/to/workspace": {
        "cxxCompilerFlags": [
          "-I/path/to/include",
          "-I/path/to/include"
        ]
      }
    }
  }
}
```

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -19,6 +19,7 @@ import LSPLogging
 import SKCore
 import SKSupport
 import TSCBasic
+import struct TSCUtility.BuildFlags
 
 import PackageLoading
 
@@ -72,6 +73,9 @@ public final class SourceKitServer: LanguageServer {
   /// Documents that are ready for requests and notifications.
   /// This generally means that the `BuildSystem` has notified of us of build settings.
   var documentsReady: Set<DocumentURI> = []
+
+  /// Custom per-workspace BuildFlags
+  private var workspaceBuildFlags: [String: BuildFlags] = [:]
 
   private var documentToPendingQueue: [DocumentURI: DocumentNotificationRequestQueue] = [:]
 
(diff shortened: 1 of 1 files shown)


resolritter:
> I the JSON format some kind of standard or did you come up with it?

I came up with it as a way of specifying the build flags per-workspace... Though I don't quite understand the intent of the question. The `"workspaces"` is a dictionary of `WorkspacePath => BuildFlags`. I don't know why it would be necessary to have a standard for this since it's something particular to this server.

> Can’t we use different compile_flags.txt or compile_commands.json located in the different workspaces to specify per-workspace build settings.

Oh, so it's already supported? I didn't know. However, does that only works for C++ flags? The solution presented here works for all the kinds of flags supported in BuildFlags (https://github.com/apple/sourcekit-lsp/blob/8791f7ded16022bb86dc004105838732bec93a63/Sources/SKCore/BuildSetup.swift#L33). Basically it instantiates one `BuildFlags` per workspace in the initialization options.

ahoppen:
> I came up with it as a way of specifying the build flags per-workspace... Though I don't quite understand the intent of the question. The `"workspaces"` is a dictionary of `WorkspacePath => BuildFlags`. I don't know why it would be necessary to have a standard for this since it's something particular to this server.

I just wanted to know whether you are implementing a previously defined standard that already applies to other language servers (like `compile_flags.txt` or `compile_commands.json`, which are documented at https://clang.llvm.org/docs/JSONCompilationDatabase.html) or came up with a format that only applies to SourceKit-LSP.

In the former case, I would have agreed that we should support that format to be compatible with other language servers, in the latter case (which it is) I would like to evaluate whether it’s possible to achieve the same thing with tools we already have before adding a new feature.

> Oh, so it's already supported? I didn't know. However, does that only works for C++ flags?

`compile_flags.txt` and `compile_commands.json` should allow you to specify compiler arguments for C, C++ and Swift. For reference, this is where we construct the build system and `CompilationDatabaseBuildSystem` is always the fallback option. https://github.com/apple/sourcekit-lsp/blob/472a06c88af97135b6faa450c5854e8f1a0ab8de/Sources/SourceKitLSP/Workspace.swift#L89-L103

I think it might not be possible to specify linker flags using these files, but linker flags shouldn’t be needed for the semantic functionality provided by SourceKit-LSP, so I don’t think this would be needed.

resolritter:
Thank you for the explanations! Closing this, then

PR #571, opened 2022-06-14
Set a default timeout duration

ahoppen:
Instead of having ad-hoc timeout durations in all the test cases, specify a default timeout duration that can be used by tests.

This allows us increase the timeout duration for all tests if we discover that e.g. sourcekitd is slower in CI setups.

rdar://91615376

### DIFF
--- a/Tests/SourceKitLSPTests/SemanticTokensTests.swift
+++ b/Tests/SourceKitLSPTests/SemanticTokensTests.swift
@@ -18,9 +18,6 @@ import XCTest
 
 private typealias Token = SyntaxHighlightingToken
 
-/// The number of seconds that we wait for a semantic token refresh.
-private let semanticRefreshTimeout: TimeInterval = 60
-
 final class SemanticTokensTests: XCTestCase {
   /// Connection and lifetime management for the service.
   private var connection: TestSourceKitServer! = nil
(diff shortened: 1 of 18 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -18,9 +18,6 @@ import XCTest
 
 private typealias Token = SyntaxHighlightingToken
 
-/// The number of seconds that we wait for a semantic token refresh.
-private let semanticRefreshTimeout: TimeInterval = 60
```
benlangmuir, line 22: I'm concerned that lowering this to 15 seconds will cause regressions.  CI environments, particularly if they are oversubscribed, or running in a VM, can be unexpectedly slow sometimes, and semantic operations seem to exercise those cases.  I would either increase the default timeout, or consider having two global constants: one for general use and one for semantic operations.  

Otherwise, replacing all the various hard-coded timeouts seems like a clear win.

ahoppen: Initially, I had two separate timeouts for syntactic and semantic operations but from a brief scan almost all timeouts we use were semantic, so I dropped the distinction for simplicity.

Regarding the value, I think most semantic operations used 15 seconds, so I stuck with that value to avoid unnecessarily long wait times if a callback doesn’t get call. My idea was that we could start with a semantic timeout of 15 seconds for now and increase it if we see that it’s not sufficient for CI environments. My impression was that the timeout of 60 seconds here was pretty arbitrary as well and might not need to be that large.

benlangmuir: Okay, that's a convincing argument.

benlangmuir (approved):

PR #575, opened 2022-06-28
Clean up index lookup requests

fwcd:
As suggested by the `FIXME` comments, this patch aims to reduce the code duplication between the `definition`, `implementation` and `references` request handlers in `SourceKitServer` by moving the index request into a new method `findIndexedOccurrences`. This will also make it easier to implement new index-related methods in the future (e.g. for call/type hierarchy requests).

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1172,6 +1172,15 @@ extension SourceKitServer {
     languageService.inlayHint(req)
   }
 
+  /// Extracts the locations of an indexed symbol's occurrences,
+  /// e.g. for definition or reference lookups.
+  /// 
+  /// - Parameters:
+  ///   - result: The symbol to look up
+  ///   - index: The index in which the occurrences will be looked up
+  ///   - useLocalFallback: Whether to consider the best known local declaration if no other locations are found
+  ///   - extractOccurrences: A function fetching the occurrences by the desired roles given a usr from the index
+  /// - Returns: The resolved symbol locations
   private func extractIndexedOccurrences(
     result: LSPResult<SymbolInfoRequest.Response>,
     index: IndexStoreDB?,
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1172,6 +1172,46 @@ extension SourceKitServer {
     languageService.inlayHint(req)
   }
 
+  private func extractIndexedOccurrences(
+    result: LSPResult<SymbolInfoRequest.Response>,
+    index: IndexStoreDB?,
+    useLocalFallback: Bool = false,
+    extractOccurrences: (String, IndexStoreDB) -> [SymbolOccurrence]
+  ) -> LSPResult<[Location]> {
+    guard case .success(let symbols) = result else {
+      return .failure(result.failure!)
+    }
+
+    guard let symbol = symbols.first else {
+      return .success([])
+    }
+
+    let fallbackLocation = useLocalFallback
+      ? [symbol.bestLocalDeclaration].compactMap { $0 }
+      : []
+
+    guard let usr = symbol.usr, let index = index else {
+      return .success(fallbackLocation)
+    }
+
+    let occurs = extractOccurrences(usr, index)
+    let locations = occurs.compactMap { occur -> Location? in
+      if occur.location.path.isEmpty {
+        return nil
+      }
+      return Location(
+        uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
+        range: Range(Position(
+          line: occur.location.line - 1, // 1-based -> 0-based
+          // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
+          utf16index: occur.location.utf8Column - 1
+          ))
+      )
+    }
+
+    return .success(locations.isEmpty ? fallbackLocation : locations)
+  }
+
   func definition(
     _ req: Request<DefinitionRequest>,
     workspace: Workspace,
@@ -1179,65 +1219,36 @@ extension SourceKitServer {
   ) {
     let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
     let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
-    // If we're unable to handle the definition request using our index, see if the
-    // language service can handle it (e.g. clangd can provide AST based definitions).
-    let resultHandler: ([Location], ResponseError?) -> Void = { (locs, error) in
-      guard locs.isEmpty else {
-        req.reply(.locations(locs))
-        return
-      }
-      let handled = languageService.definition(req)
-      guard !handled else { return }
-      if let error = error {
-        req.reply(.failure(error))
-      } else {
-        req.reply(.locations([]))
-      }
-    }
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        resultHandler([], result.failure)
-        return
-      }
-
-      let fallbackLocation = [symbol.bestLocalDeclaration].compactMap { $0 }
-
-      guard let usr = symbol.usr, let index = index else {
-        resultHandler(fallbackLocation, nil)
-        return
-      }
-
-      log("performing indexed jump-to-def with usr \(usr)")
-
-      var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
-      if occurs.isEmpty {
-        occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
-      }
-
-      // FIXME: overrided method logic
-
-      let locations = occurs.compactMap { occur -> Location? in
-        if occur.location.path.isEmpty {
-          return nil
+    let callback = callbackOnQueue(self.queue) { (result: LSPResult<SymbolInfoRequest.Response>) in
+      let extractedResult = self.extractIndexedOccurrences(result: result, index: index, useLocalFallback: true) { (usr, index) in
+        log("performing indexed jump-to-def with usr \(usr)")
+        var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
+        if occurs.isEmpty {
+          occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
         }
-        return Location(
-          uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
-          range: Range(Position(
-            line: occur.location.line - 1, // 1-based -> 0-based
-            // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
-            utf16index: occur.location.utf8Column - 1
-            ))
-        )
+        return occurs
       }
 
-      resultHandler(locations.isEmpty ? fallbackLocation : locations, nil)
+      switch extractedResult {
+      case .success(let locs):
+        // If we're unable to handle the definition request using our index, see if the
+        // language service can handle it (e.g. clangd can provide AST based definitions).
+        guard locs.isEmpty else {
+          req.reply(.locations(locs))
+          return
+        }
+        let handled = languageService.definition(req)
+        guard !handled else { return }
+        req.reply(.locations([]))
+      case .failure(let error):
+        req.reply(.failure(error))
+      }
     }
     let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
                           cancellation: req.cancellationToken, reply: callback)
     languageService.symbolInfo(request)
   }
 
-  // FIXME: a lot of duplication with definition request
   func implementation(
     _ req: Request<ImplementationRequest>,
     workspace: Workspace,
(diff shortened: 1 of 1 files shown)


Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1230,111 +1214,105 @@ extension SourceKitServer {
...
-    let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
-    let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        if let error = result.failure {
+    findIndexedOccurrences(
+      textDocument: req.params.textDocument,
+      position: req.params.position,
+      id: req.id,
+      cancellation: req.cancellationToken,
```
fwcd, line 1233: We are passing quite a few request-derived parameters here. An idea I had was to pass the request directly to the method. That would however require introducing a new protocol, e.g. `TextDocumentPositionRequest`, that encompasses both a `textDocument` and `position` parameter, to formulate the corresponding generic constraint, as far as I can tell. Additionally it would make the method 'aware' of the request, which we currently abstract away, therefore I went with the 'manual' parameter-passing approach for now.

fwcd (commented):
A note on the implementation below.

fwcd:
@swift-ci please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1230,111 +1214,105 @@ extension SourceKitServer {
...
   ) {
-    let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
-    let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        if let error = result.failure {
+    findIndexedOccurrences(
+      textDocument: req.params.textDocument,
+      position: req.params.position,
+      id: req.id,
```
ahoppen, line 1232: Looking at the old implementation of `definition`, I think you need to pass `useLocalFallback: true` here.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: What do you think about passing these parameters to `findIndexedOccurrences` as `Request< SymbolInfoRequest>`? I think that would contain exactly the parameters the method needs without the need for a new abstraction level.

Also: It wasn’t obvious to me what `useLocalFallback` and `extractOccurrences` did just from their parameter name and `completion` made me wonder whether we also used this function for code completion (hooray to ambiguous names). Would you mind adding a doc comment for these parameters and maybe renaming `completion` to `completionHandler` to avoid the ambiguity?

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: Good catch, thanks!

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: That's a good idea, though since the request takes a callback, should we move the logic that performs the `SymbolInfo` request into the individual methods (`definition`, etc. )?

ahoppen: New idea: `definition` etc. run a normal `symbolInfo` request to the language service and we just factor out the functionality to extract locations from the symbol info response.

I just played around with this idea and this is what I came up with the following. WDYT?

<details>

```diff
diff --git a/Sources/SourceKitLSP/SourceKitServer.swift b/Sources/SourceKitLSP/SourceKitServer.swift
index 58b2b9e..c370a0d 100644
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1172,6 +1172,42 @@ extension SourceKitServer {
     languageService.inlayHint(req)
   }
 
+  private func extractIndexedOccurrances(
+    result: Result<SymbolInfoRequest.Response, ResponseError>,
+    index: IndexStoreDB?,
+    useLocalFallback: Bool = false,
+    extractOccurrences: (String, IndexStoreDB) -> [SymbolOccurrence]
+  ) -> Result<[Location], ResponseError> {
+    guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
+      return .failure(result.failure!)
+    }
+
+    let fallbackLocation = useLocalFallback
+      ? [symbol.bestLocalDeclaration].compactMap { $0 }
+      : []
+
+    guard let usr = symbol.usr, let index = index else {
+      return .success(fallbackLocation)
+    }
+
+    let occurs = extractOccurrences(usr, index)
+    let locations = occurs.compactMap { occur -> Location? in
+      if occur.location.path.isEmpty {
+        return nil
+      }
+      return Location(
+        uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
+        range: Range(Position(
+          line: occur.location.line - 1, // 1-based -> 0-based
+          // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
+          utf16index: occur.location.utf8Column - 1
+          ))
+      )
+    }
+
+    return .success(locations.isEmpty ? fallbackLocation : locations)
+  }
+
   func definition(
     _ req: Request<DefinitionRequest>,
     workspace: Workspace,
@@ -1179,65 +1215,34 @@ extension SourceKitServer {
   ) {
     let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
     let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
-    // If we're unable to handle the definition request using our index, see if the
-    // language service can handle it (e.g. clangd can provide AST based definitions).
-    let resultHandler: ([Location], ResponseError?) -> Void = { (locs, error) in
-      guard locs.isEmpty else {
-        req.reply(.locations(locs))
-        return
+    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
+                          cancellation: req.cancellationToken) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
+      let extractedResult = self.extractIndexedOccurrances(result: result, index: index) { (usr, index) in
+        log("performing indexed jump-to-def with usr \(usr)")
+        var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
+        if occurs.isEmpty {
+          occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
+        }
+        return occurs
       }
-      let handled = languageService.definition(req)
-      guard !handled else { return }
-      if let error = error {
-        req.reply(.failure(error))
-      } else {
+      switch extractedResult {
+      case .success(let locs):
+        // If we're unable to handle the definition request using our index, see if the
+        // language service can handle it (e.g. clangd can provide AST based definitions).
+        guard locs.isEmpty else {
+          req.reply(.locations(locs))
+          return
+        }
+        let handled = languageService.definition(req)
+        guard !handled else { return }
         req.reply(.locations([]))
+      case .failure(let error):
+        req.reply(.failure(error))
       }
     }
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        resultHandler([], result.failure)
-        return
-      }
-
-      let fallbackLocation = [symbol.bestLocalDeclaration].compactMap { $0 }
-
-      guard let usr = symbol.usr, let index = index else {
-        resultHandler(fallbackLocation, nil)
-        return
-      }
-
-      log("performing indexed jump-to-def with usr \(usr)")
-
-      var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
-      if occurs.isEmpty {
-        occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
-      }
-
-      // FIXME: overrided method logic
-
-      let locations = occurs.compactMap { occur -> Location? in
-        if occur.location.path.isEmpty {
-          return nil
-        }
-        return Location(
-          uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
-          range: Range(Position(
-            line: occur.location.line - 1, // 1-based -> 0-based
-            // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
-            utf16index: occur.location.utf8Column - 1
-            ))
-        )
-      }
-
-      resultHandler(locations.isEmpty ? fallbackLocation : locations, nil)
-    }
-    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
-                          cancellation: req.cancellationToken, reply: callback)
     languageService.symbolInfo(request)
   }
 
-  // FIXME: a lot of duplication with definition request
   func implementation(
     _ req: Request<ImplementationRequest>,
     workspace: Workspace,
@@ -1245,95 +1250,49 @@ extension SourceKitServer {
   ) {
     let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
     let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        if let error = result.failure {
-          req.reply(.failure(error))
-        } else {
-          req.reply(.locations([]))
+    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
+                          cancellation: req.cancellationToken) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
+      let extractedResult = self.extractIndexedOccurrances(result: result, index: index) { (usr, index) in
+        var occurs = index.occurrences(ofUSR: usr, roles: .baseOf)
+        if occurs.isEmpty {
+          occurs = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
         }
-        return
-      }
-
-      guard let usr = symbol.usr, let index = index else {
-        return req.reply(.locations([]))
-      }
-
-      var occurs = index.occurrences(ofUSR: usr, roles: .baseOf)
-      if occurs.isEmpty {
-        occurs = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
+        return occurs
       }
-
-      let locations = occurs.compactMap { occur -> Location? in
-        if occur.location.path.isEmpty {
-          return nil
-        }
-        return Location(
-          uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
-          range: Range(Position(
-            line: occur.location.line - 1, // 1-based -> 0-based
-            // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
-            utf16index: occur.location.utf8Column - 1
-            ))
-        )
+      switch extractedResult {
+      case .success(let locs):
+        req.reply(.locations(locs))
+      case .failure(let error):
+        req.reply(.failure(error))
       }
-
-      req.reply(.locations(locations))
     }
-    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
-                          cancellation: req.cancellationToken, reply: callback)
     languageService.symbolInfo(request)
   }
 
-  // FIXME: a lot of duplication with definition request
   func references(
     _ req: Request<ReferencesRequest>,
     workspace: Workspace,
     languageService: ToolchainLanguageServer
   ) {
     let symbolInfo = SymbolInfoRequest(textDocument: req.params.textDocument, position: req.params.position)
-    let callback = callbackOnQueue(self.queue) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
-      guard let symbols: [SymbolDetails] = result.success ?? nil, let symbol = symbols.first else {
-        if let error = result.failure {
-          req.reply(.failure(error))
-        } else {
-          req.reply([])
+    let index = self.workspaceForDocument(uri: req.params.textDocument.uri)?.index
+    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
+                          cancellation: req.cancellationToken) { (result: Result<SymbolInfoRequest.Response, ResponseError>) in
+      let extractedResult = self.extractIndexedOccurrances(result: result, index: index) { (usr, index) in
+        log("performing indexed jump-to-def with usr \(usr)")
+        var roles: SymbolRole = [.reference]
+        if req.params.context.includeDeclaration {
+          roles.formUnion([.declaration, .definition])
         }
-        return
-      }
-
-      guard let usr = symbol.usr, let index = workspace.index else {
-        req.reply([])
-        return
-      }
-
-      log("performing indexed jump-to-def with usr \(usr)")
-
-      var roles: SymbolRole = [.reference]
-      if req.params.context.includeDeclaration {
-        roles.formUnion([.declaration, .definition])
+        return index.occurrences(ofUSR: usr, roles: roles)
       }
-
-      let occurs = index.occurrences(ofUSR: usr, roles: roles)
-
-      let locations = occurs.compactMap { occur -> Location? in
-        if occur.location.path.isEmpty {
-          return nil
-        }
-        return Location(
-          uri: DocumentURI(URL(fileURLWithPath: occur.location.path)),
-          range: Range(Position(
-            line: occur.location.line - 1, // 1-based -> 0-based
-            // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
-            utf16index: occur.location.utf8Column - 1
-            ))
-        )
+      switch extractedResult {
+      case .success(let locs):
+        req.reply(locs)
+      case .failure(let error):
+        req.reply(.failure(error))
       }
-
-      req.reply(locations)
     }
-    let request = Request(symbolInfo, id: req.id, clientID: ObjectIdentifier(self),
-                          cancellation: req.cancellationToken, reply: callback)
     languageService.symbolInfo(request)
   }
 
```
</details>

fwcd: Looks nice, this way we wouldn't have to deal with asynchrony either and could conveniently use trailing closures

fwcd:
@swift-ci please test

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: Ah yes, sorry I missed this. Should be added

ahoppen (approved):
Thanks. LGTM. 👍

PR #578, opened 2022-06-30
Implement call hierarchy

fwcd:
This branch implements the LSP requests for querying incoming and outgoing function calls. The implementation is similar to the existing definition/reference/implementation lookup in that it queries the index for symbols related by the `calledBy` role.

To do:
- [x] Implement incoming/outgoing call request handlers
- [ ] Investigate if and how we could provide `fromRanges` in `CallHierarchy{Incoming,Outgoing}Call` structures
- [ ] Add unit tests


Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1298,13 +1313,126 @@ extension SourceKitServer {
...
+      kind: symbol.kind.asLspSymbolKind(),
+      tags: nil,
+      uri: location.uri,
+      range: location.range,
+      selectionRange: location.range,
+      // We encode usr and uri for incoming/outgoing call lookups in the implementation-specific data field
+      data: .dictionary([
+        "usr": .string(symbol.usr),
+        "uri": .string(location.uri.stringValue),
+      ])
```
fwcd, line 1335: We encode `usr` and `uri` in the (implementation-specific) LSP data field since incoming/outgoing call requests are only provided with a `CallHierarchyItem` in the request parameters.

fwcd (commented):
A quick note on the implementation

fwcd:
@swift-ci please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1172,6 +1177,24 @@ extension SourceKitServer {
...
+  private func indexToLSPLocation(_ location: SymbolLocation) -> Location? {
+    guard !location.path.isEmpty else { return nil }
+    return Location(
+      uri: DocumentURI(URL(fileURLWithPath: location.path)),
+      range: Range(Position(
+        // 1-based -> 0-based
+        // Note that we still use max(0, ...) as a fallback if the location is zero.
+        line: max(0, location.line - 1),
+        // FIXME: we need to convert the utf8/utf16 column, which may require reading the file!
+        utf16index: max(0, location.utf8Column - 1)
```
fwcd, line 1193: This change btw fixes a bug with definition lookup too, e.g. ctrl+clicking a standard library function like `print` would (at least on my Mac) yield `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/swift/Swift.swiftmodule/arm64e-apple-macos.swiftinterface` along with `(0, 0)` as a position (which would subsequently get turned into `(-1, -1)` by this conversion and cause errors on the LSP client-side.

fwcd: Aside from that it would of course be cool to have proper positions even within a `*.swiftinterface`, but for now having a fallback at least lets us link to the file correctly.

ahoppen: Yes, having proper locations in`.swiftinterface` files would be great, but IIRC that’s a bigger undertaking. For now, I think this is a good fix.

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1195,30 +1218,23 @@ extension SourceKitServer {
       return .success([])
     }
 
-    let fallbackLocation = useLocalFallback
-      ? [symbol.bestLocalDeclaration].compactMap { $0 }
-      : []
+    let fallback: [(occurrence: SymbolOccurrence?, location: Location)] =
+      useLocalFallback
+        ? (symbol.bestLocalDeclaration.map { [(occurrence: nil, location: $0)] } ?? [])
+        : []
```
ahoppen, line 1224: Just wondering whether this would read better as follows. Maybe it’s just my mind that’s weird but I find thinking about `map` on `Optional` harder to reason about than `if` statements. I’ll leave it up to you what you like better.
```suggestion
    let fallback: [(occurrence: SymbolOccurrence?, location: Location)]
    if useLocalFallback, let bestLocalDeclaration = symbol.bestLocalDeclaration {
      fallback = [(occurrence: nil, location: $0)]
    } else {
      fallback = []
    }
```

Tests/SourceKitLSPTests/CallHierarchyTests.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    // Test incoming call hierarchy
+
+    XCTAssertEqual(try incomingCalls(at: testLoc("a")), [
+      inCall(item("b(x:)", .function, usr: bUsr, at: "b"), at: "b->a"),
+      inCall(item("c()", .function, usr: cUsr, at: "c"), at: "c->a"),
+    ])
+    XCTAssertEqual(try incomingCalls(at: testLoc("b")), [
+      inCall(item("b(x:)", .function, usr: bUsr, at: "b"), at: "b->b"),
+    ])
```
ahoppen, line 129: I suppose incoming calls of `c` also contain the recursive `c` call, right? Also: Do you know how VSCode represents recursive calls?

ahoppen (approved):
Two minor comments, otherwise this looks very good to me. 👍

Tests/SourceKitLSPTests/CallHierarchyTests.swift
fwcd: Yes, incoming calls on `c` would contain the recursive call of itself too. Since we only query one level at a time, LSP and VSCode can deal with this (infinite) call tree just fine.

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: I have no strong preference here, let's go with the if-version then :+1:

fwcd:
@swift-ci please test

PR #579, opened 2022-06-30
Tests: replace use of `sleep` with `Thread.sleep(forTimeInterval:)`

compnerd:
Replace the use of `sleep` which is a POSIX construct for the portable
`Thread.sleep(forTimeInterval:)` which has similar semantics but allows
the code to be built for other platforms.  This allows us to finally
build the test suite for Windows.


compnerd:
@swift-ci please test

ahoppen (approved):
Sounds reasonable to me 👍

PR #580, opened 2022-07-04
SKCore: add an additional spelling for `PATH`

compnerd:
Windows uses `Path` for the environment variable spelling, but expects
that environment variables are case-insensitive.  Foundation's
implementation treats the environment variables as being case sensitive
and will not properly look up the value  Add a case corrected spelling
for that support.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #582, opened 2022-07-05
Implement type hierarchy

fwcd:
This branch implements support for querying supertype and subtype hierarchies via the new requests introduced in #377 (note that we include extensions in the list of subtypes):

image

To do:
- [x] Add type hierarchy provider
- [x] Implement the requests
- [x] Add unit test


Tests/SourceKitLSPTests/TypeHierarchyTests.swift
```diff
@@ -0,0 +1,167 @@
...
+    ])
+    assertEqualIgnoringData(try supertypes(at: testLoc("C")), [
+      item("B", .class, at: "B"),
+    ])
+    assertEqualIgnoringData(try supertypes(at: testLoc("D")), [
+      item("A", .class, at: "A"),
+    ])
+    assertEqualIgnoringData(try supertypes(at: testLoc("S")), [
+      item("P", .interface, at: "P"),
+    ])
```
fwcd, line 127: The supertype hierarchy doesn't seem to include the protocol `X`, which `P` retroactively conforms to:

```swift
protocol P {}
protocol X {}
struct S: P {}
extension S: X {}
```

This is due to the index treating the extension `S: X` as a separate entity (it is also listed as such when enumerating the subtypes of `X`).

Tests/SourceKitLSPTests/TypeHierarchyTests.swift
```diff
@@ -0,0 +1,167 @@
...
+      _ expected: [TypeHierarchyItem],
+      file: StaticString = #file,
+      line: UInt = #line
+    ) {
+      XCTAssertEqual(
+        actual.map(withoutData),
+        expected.map(withoutData),
+        file: file,
+        line: line
+      )
```
fwcd, line 81: We ignore the LSP-specific `data` field in comparisons since its USRs are hard to test in the presence of extensions (which have individual USRs) and are not user-facing anyway. Thus we consider the `data` field as an implementation detail here that the test treats as a black box.

fwcd (commented):
Two notes on the implementation

fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1443,6 +1448,160 @@ extension SourceKitServer {
...
+      let definition = index.occurrences(ofUSR: related.symbol.usr, roles: [.definition, .declaration]).first
+      let definitionSymbolLocation = definition.map(\.location)
+      let definitionLocation = definitionSymbolLocation.flatMap(indexToLSPLocation)
+
+      return indexToLSPTypeHierarchyItem(
+        symbol: related.symbol,
+        moduleName: definitionSymbolLocation?.moduleName,
+        location: definitionLocation ?? location, // Use occurrence location as fallback
+        index: index
+      )
```
ahoppen, line 1600: I wonder whether it would make sense to extract that code to a function because it’s shared between super type and subtype. 

And now that I think about it, we could probably also extract the closure in `incomingCalls`/`outgoingCalls`.

Tests/SourceKitLSPTests/TypeHierarchyTests.swift
ahoppen: Could we work around this by doing an index query of extensions for `S` first and also including all the extension’s supertypes?

I think as far as subtypes are concerned, it’s fine to show the extension. I don’t have strong feelings about that.

Tests/SourceKitLSPTests/TypeHierarchyTests.swift
```diff
@@ -0,0 +1,167 @@
...
+    assertEqualIgnoringData(try subtypes(at: testLoc("B")), [
+      item("C", .class, at: "C"),
+    ])
+    assertEqualIgnoringData(try subtypes(at: testLoc("P")), [
+      item("B", .class, at: "B"),
+      item("S", .struct, at: "S"),
+      item("E", .enum, at: "E"),
+    ])
+    assertEqualIgnoringData(try subtypes(at: testLoc("S")), [
+      item("S: X", .null, detail: "Extension at a.swift:13", at: "extS:X"),
```
ahoppen, line 147: I think an extension that adds multiple protocol conformances would also be interesting.

E.g. 
```swift
struct T {}
extension T: X, P {}
```

I would expect the extension to be displayed as `T: X, P`.

Also: I really like how you show the conformed types in the extension names. Makes it much easier to find the extension you are looking for IMO.

Tests/SourceKitLSPTests/TypeHierarchyTests.swift
fwcd: Yeah, I think we could do that, we would probably have do an index query per extension though (which could be a few).

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: Yeah that would be nice indeed. The main difficulty with call hierarchy is that `CallHierarchyIncomingCall` and `CallHierarchyOutgoingCall` are different types

Tests/SourceKitLSPTests/TypeHierarchyTests.swift
ahoppen: I don’t expect these queries to take very long, so I think it would be worth it.

Sources/SourceKitLSP/SourceKitServer.swift
fwcd: Did you mean type hierarchy? The problem is that there are still some subtleties to be considered, e.g. the `subtypes` closure uses the related symbol rather than the occurrence directly, leaving us essentially with only the definition lookup, which we could abstract over (also `subtypes` will get some more special logic with https://github.com/apple/sourcekit-lsp/pull/582/files#r916912420). Since we need both the definition and the symbol location, we'd need these values separately anyway. If you feel strongly about this, I could try and see, but I don't think we would gain much from factoring/unifying these methods.

ahoppen: Yes, I meant type hierarchy.

Oh, I missed those differences. In that case I withdraw my commend and agree that we don’t need to unify these.

ahoppen (approved):
Thanks. Looks good to me. 👍

PR #585, opened 2022-07-12
SourceKitLSP: change an assert to a check

compnerd:
Rather than asserting, raise an error for the user and abort the
operation.  This is motivated by the desire to get the test suite mostly
working on Windows so that the underlying issues can be worked through
more easily.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #586, opened 2022-07-12
Add Package.swift root path to error message

keith:
Previously in the case you have an invalid Package.swift file the LSP
log would just show:

```
[2022-07-12 14:57:23.940] failed to create SwiftPMWorkspace: the package does not contain a buildable target
```

With this change it at least hints you towards where the invalid file
is:

```
[2022-07-12 15:00:45.106] failed to create SwiftPMWorkspace /Users/ksmiley/dev/oss-swift/swiftpm/Sources/PackageModel: the package does not contain a buildable target
```


benlangmuir (approved):
@swift-ci please test

benlangmuir:
@swift-ci please test

keith:
looks like we're all green 🎉 I don't have access to this one, anything else I should do before merging?

benlangmuir:
Nope, that was it, thanks!

PR #587, opened 2022-07-12
Lightly validate discovered Package.swift files

keith:
Previously if you had a normal source file named Package.swift
sourcekit-lsp would pick it up and fail. Now there is some very light
validation that the file looks valid, checking that it has a
PackageDescription import, before accepting it, otherwise it keeps
looking.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -50,6 +50,8 @@ public final class SwiftPMWorkspace {
 
   let workspacePath: AbsolutePath
   let packageRoot: AbsolutePath
+  /// *Public for testing*
+  public var _packageRoot: AbsolutePath { packageRoot }
   var packageGraph: PackageGraph
   let workspace: Workspace
   public let buildParameters: BuildParameters
@@ -545,7 +547,15 @@ private func findPackageDirectory(
   containing path: AbsolutePath,
   _ fileSystem: FileSystem) -> AbsolutePath? {
   var path = path
-  while !fileSystem.isFile(path.appending(component: "Package.swift")) {
+  while true {
+    let packagePath = path.appending(component: "Package.swift")
+    if fileSystem.isFile(packagePath) {
+      let contents = try? fileSystem.readFileContents(packagePath)
+      if contents?.cString.contains("PackageDescription") == true {
+        return path
+      }
+    }
+
     if path.isRoot {
       return nil
     }
--- a/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
+++ b/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
@@ -567,6 +567,30 @@ final class SwiftPMWorkspaceTests: XCTestCase {
         }), "missing resource_bundle_accessor.swift from \(arguments)")
     }
   }
+
+  func testNestedInvalidPackageSwift() throws {
+    let fs = InMemoryFileSystem()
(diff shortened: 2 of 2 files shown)


keith:
This was suggested on https://github.com/apple/swift-package-manager/pull/5651

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -545,7 +545,15 @@ private func findPackageDirectory(
   containing path: AbsolutePath,
   _ fileSystem: FileSystem) -> AbsolutePath? {
   var path = path
-  while !fileSystem.isFile(path.appending(component: "Package.swift")) {
+  while true {
+    let packagePath = path.appending(component: "Package.swift")
+    if fileSystem.isFile(packagePath) {
+      let contents = try? fileSystem.readFileContents(packagePath)
+      if contents?.cString.contains("import PackageDescription") == true {
```
benlangmuir, line 552: Maybe we should just check for "PackageDescription" to avoid false negatives due to whitespace or comments?

benlangmuir (commented):
If I understand correctly, this would happen if you open a subdirectory of a package as the "workspace directory" rather than the package root. I don't love having to scan the file, but I don't have a better solution. @ahoppen ?

This should have a test in `SwiftPMWorkspaceTests.swift`. The other tests in that file should   serve as a reasonable pattern, and the test package itself can live in `SKTestSupport/INPUTS`

ahoppen:
I’m not really happy with the solution as well but I think it solves the problem, which is already pretty rare in a straightforward way, so I’m fine with it.

keith:
> If I understand correctly, this would happen if you open a subdirectory of a package as the "workspace directory" rather than the package root.

Correct, as in SwiftPM before that linked PR above, when they had some models named Package so they naturally named the files Package.swift

> This should have a test in `SwiftPMWorkspaceTests.swift`. The other tests in that file should serve as a reasonable pattern, and the test package itself can live in `SKTestSupport/INPUTS`

I was able to write one without any fixtures with the in memory file system, let me know if you'd prefer the fixtures. I did have to make something public here (regardless of fixtures), which might not be ideal

keith:
@swift-ci test

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -49,7 +49,7 @@ public final class SwiftPMWorkspace {
   public weak var delegate: SKCore.BuildSystemDelegate? = nil
 
   let workspacePath: AbsolutePath
-  let packageRoot: AbsolutePath
+  public let packageRoot: AbsolutePath
```
ahoppen, line 52: Our general paradigm for these would be to have

```suggestion
  let packageRoot: AbsolutePath
  /// *Public for testing*
  public let _packageRoot: AbsolutePath { packageRoot }
```

That way we can check that the underscored method is only used from the tests using call hierarchy.

ahoppen (commented):
The test case looks good. Thanks for adding it.

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
keith: done!

keith:
@swift-ci test

ahoppen (approved):

keith:
I don't have merge access to this one, can someone land this if it's ready?

ahoppen:
Interesting, I thought you could also merge because you could trigger CI. Anyway, I merged it now.

keith:
I think CI triggering is global but contributor access is only added as needed when I asked

PR #589, opened 2022-07-16
SourceKitLSPTests: replace some `fatalError` with `XCTFail`

compnerd:
This replaces many instances of `fatalError` with `XCTFail`.  Although
these code paths should not come to pass, there are currently failures
in the test suite on Windows where these trigger.  By using `XCTFail`
instead, we allow the tests to continue execution which is helpful for
investigating the current state and make progress towards fixing the
issues.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #592, opened 2022-07-19
SKCore,SKTestSupport: adjust test server handling

compnerd:
The test system depended on the shebang to locate the python
interpreter.  However, this is not a portable system.  Instead, prefer
to explicitly search for the interpreter prior to the execution.  This
enables supporting execution of the script support on all platforms. A
secondary change of the printed string is required for Windows. Python
will replace `\n` with `\r\n` resulting in `\r\n` being emitted as
`\r\r\n` on Windows breaking the expectations on the receiver.  Adjust
this by explicitly writing out a binary string to the raw underlying
buffer to avoid the translation.

### DIFF
--- a/Sources/SKCore/BuildServerBuildSystem.swift
+++ b/Sources/SKCore/BuildServerBuildSystem.swift
@@ -18,6 +18,20 @@ import LSPLogging
 import SKSupport
 import TSCBasic
 
+enum BuildServerTestError: Error {
+    case executableNotFound(String)
+}
+
+func executable(_ name: String) -> String {
+#if os(Windows)
+  guard !name.hasSuffix(".exe") else { return name }
+  return "\(name).exe"
+#else
+  return name
+#endif
+}
+
+
 /// A `BuildSystem` based on communicating with a build server
 ///
 /// Provides build settings from a build server launched based on a
@@ -31,6 +45,8 @@ public final class BuildServerBuildSystem {
 
   var handler: BuildServerHandler?
   var buildServer: JSONRPCConnection?
+
+  let searchPaths: [AbsolutePath]
 
   public private(set) var indexDatabasePath: AbsolutePath?
   public private(set) var indexStorePath: AbsolutePath?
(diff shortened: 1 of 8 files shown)


compnerd:

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -18,6 +18,23 @@ import LSPLogging
...
 import TSCBasic
 
+enum BuildServerTestError: Error {
+    case executableNotFound(String)
+}
+
+func executable(_ name: String) -> String {
+#if os(Windows)
+  if name.count > 4,
+     name.suffix(from: name.index(name.endIndex, offsetBy: -4)) == ".exe" {
```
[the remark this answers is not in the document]
ahoppen: I think this is still open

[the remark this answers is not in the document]
compnerd: Yeah, I wanted to try to see if the failure was infrastructure or not.  I'll make this change subsequently.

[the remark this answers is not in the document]
ahoppen: OK, that’s fine. Thought this got unnoticed because you addressed the other comment ;-)

compnerd:
@swift-ci please test

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -85,8 +110,13 @@ public final class BuildServerBuildSystem {
   }
 
   private func initializeBuildServer() throws {
-    let serverPath = AbsolutePath(serverConfig.argv[0], relativeTo: projectRoot)
-    let flags = Array(serverConfig.argv[1...])
+    let interpreter =
+        lookupExecutablePath(filename: executable("python3"), searchPaths: searchPaths)
+            ?? lookupExecutablePath(filename: executable("python"), searchPaths: searchPaths)
```
ahoppen, line 115: Since the Python scripts had a shebang of `#!/usr/bin/env python3` I don’t think we need to look for the `python` executable. That will just result in Python 2 vs Python 3 syntax errors down the line.

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -85,8 +110,13 @@ public final class BuildServerBuildSystem {
...
+    guard let interpreter = interpreter else {
+      throw BuildServerTestError.executableNotFound("python")
```
ahoppen, line 117:
```suggestion
      throw BuildServerTestError.executableNotFound("python3")
```

ahoppen (commented):
Looks like CI is failing to find the python3 executable.

Sources/SKCore/BuildServerBuildSystem.swift
compnerd: We do need to search for `python` as Windows uses `python.exe` always.

compnerd:
How did this work previously?  Was the test run altering `PATH`?

Sources/SKCore/BuildServerBuildSystem.swift
ahoppen: I see. Makes sense 👍

ahoppen:
I don’t think it was. I have no idea why it fails to find `python3` 🤔

ahoppen:
rdar://99561858

compnerd:
@swift-ci please test

ahoppen (approved):
LGTM but I think you mistakingly committed .server.py.swp

compnerd:
@swift-ci please test

PR #593, opened 2022-07-21
Increase default timeout to 60 seconds

ahoppen:
The reduced timeout of 15 seconds was causing issues in CI.


ahoppen:
@swift-ci Please smoke test

ahoppen:
@swift-ci Please test

PR #594, opened 2022-07-22
SKSwiftPMWorkspace: remove unnecessary optionality

compnerd:
These functions will return a value and never nil, so remove the
optional return type.  This helps make it easier to reason about
the code paths.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #597, opened 2022-07-29
Several changes in build-script-helper.py to support building SourceKit-LSP in a unified build

ahoppen:
This PR contains a couple of small changes that are necessary to build SourceKit-LSP in a unified SwiftPM build, removing the need to build indexstore-db twice (once for indexstore-db testing and once for SourceKit-LSP testing).

- Don’t clean build directory prior to building
  - AFAICT cleaning the build directory prior to building was done to work around a bug in SwiftPM at some point. I don’t think that it’s needed anymore.
- Explicitly build all targets
  - When we build SourceKit-LSP in a unified SwiftPM workspace, a simple `swift build` invocation will build all products in the unified workspace (not just SourceKit-LSP products).
- Add a `--multiroot-data-file` option to build sourcekit-lsp in a unified build
- Print environment variables that were specified when executing commands
  - Just some logging improvement
- Add flake8 configuration
  - I formatted the build script according to these rules, maybe we can keep that formatting in the future
- Don’t pass -no-toolchain-stdlib-rpath when installing sourcekit-lsp
  - Since https://github.com/apple/swift-package-manager/pull/4208 SwiftPM no longer adds rpath to the toolchain libraries, so we don’t need to exclude them.
- Ignore stderr when retrieving SwifitPM’s bin path
  - I just noticed that since https://github.com/apple/sourcekit-lsp/pull/598 the bin-path computation was wrong because the returned bin path included warnings about `--build-path` being deprecated in favor of `--scratch-path`
- Rename `--build-path` to `--scratch-path` in SwiftPM invocation
  - `--build-path` has been deprectated in favor of `--scratch-path` in SwiftPM.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -47,10 +47,14 @@ def check_call(cmd: List[str], additional_env: Dict[str, str] = {}, verbose: boo
     subprocess.check_call(cmd, env=env_with_additional_env(additional_env), stderr=subprocess.STDOUT)
 
 
-def check_output(cmd: List[str], additional_env: Dict[str, str] = {}, verbose: bool = False) -> str:
+def check_output(cmd: List[str], additional_env: Dict[str, str] = {}, capture_stderr: bool = True, verbose: bool = False) -> str:
     if verbose:
         print_cmd(cmd=cmd, additional_env=additional_env)
-    return subprocess.check_output(cmd, env=env_with_additional_env(additional_env), stderr=subprocess.STDOUT, encoding='utf-8')
+    if capture_stderr:
+        stderr = subprocess.STDOUT
+    else:
+        stderr = subprocess.DEVNULL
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -185,27 +210,21 @@ def run_tests(swift_exec: str, args: argparse.Namespace) -> None:
         '--disable-testable-imports',
         '--test-product', 'SourceKitLSPPackageTests'
     ] + swiftpm_args
-    check_call(cmd, env=env, verbose=args.verbose)
+    check_call(cmd, additional_env=additional_env, verbose=args.verbose)
```
benlangmuir, line 213: If you change the build but not the tests, won't this rebuild everything?

benlangmuir: I was confused, disregard.

benlangmuir (approved):

ahoppen:
@swift-ci Please test

PR #598, opened 2022-07-29
Improvements to build-script.py

ahoppen:
This should contain no functionality change. I just
- added type annotations
- reordered some code into different functions
- changed indentation to 4 space to be consistent with Python scripts in the swift repo
- Removed some dead code

This is a fallout of trying to get sourcekit-lsp and indexstore-db into the unified SwiftPM build.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -100,18 +100,14 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
 
   return swiftpm_args
 
-def install(swiftpm_bin_path: str, prefixes: List[str], toolchain: str) -> None:
+def install(swiftpm_bin_path: str, prefixes: List[str]) -> None:
   for prefix in prefixes:
-    install_binary('sourcekit-lsp', swiftpm_bin_path, os.path.join(prefix, 'bin'), toolchain)
+    install_binary('sourcekit-lsp', swiftpm_bin_path, os.path.join(prefix, 'bin'))
 
-def install_binary(exe: str, source_dir: str, install_dir: str, toolchain: str) -> None:
+def install_binary(exe: str, source_dir: str, install_dir: str) -> None:
   cmd = ['rsync', '-a', os.path.join(source_dir, exe), install_dir]
   print(' '.join(cmd))
   subprocess.check_call(cmd)
-
-  if platform.system() == 'Darwin':
-    result_path = os.path.join(install_dir, exe)
-    stdlib_rpath = os.path.join(toolchain, 'lib', 'swift', 'macosx')
 
 def handle_invocation(swift_exec: str, args: argparse.Namespace) -> None:
   swiftpm_args = get_swiftpm_options(swift_exec, args)
@@ -161,7 +157,7 @@ def handle_invocation(swift_exec: str, args: argparse.Namespace) -> None:
 
     if not args.install_prefixes:
       args.install_prefixes = [args.toolchain]
-    install(bin_path, args.install_prefixes, args.toolchain)
+    install(bin_path, args.install_prefixes)
   else:
     assert False, 'unknown action \'{}\''.format(args.action)
 
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -11,225 +11,225 @@ import json
 from typing import List, Optional, Dict
 
 def error(message: str) -> None:
-  print("--- %s: error: %s" % (os.path.basename(sys.argv[0]), message), file=sys.stderr)
-  sys.stderr.flush()
-  raise SystemExit(1)
+    print("--- %s: error: %s" % (os.path.basename(sys.argv[0]), message), file=sys.stderr)
+    sys.stderr.flush()
+    raise SystemExit(1)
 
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -8,230 +8,306 @@
...
+        elif re.match('android-', args.cross_compile_host):
+            print('Cross-compiling for %s' % args.cross_compile_host)
+            swiftpm_args += ['--destination', args.cross_compile_config]
+        else:
+            fatal_error("cannot cross-compile for %s" % args.cross_compile_host)
+
+    return swiftpm_args
+
+
+def get_swiftpm_environment_variables(swift_exec: str, args: argparse.Namespace) -> Dict[str, str]:
```
benlangmuir, line 126: What do you think of instead having `get_swiftpm_options` return a tuple of arguments and environment? My thinking is that you always need both to be correct.

Utilities/build-script-helper.py
```diff
@@ -8,230 +8,306 @@
...
+def install_binary(exe: str, source_dir: str, install_dir: str, verbose: bool) -> None:
+    cmd = ['rsync', '-a', os.path.join(source_dir, exe), install_dir]
+    check_call(cmd, verbose=verbose)
+
+def install(swift_exec: str, args: argparse.Namespace) -> None:
+    swiftpm_args = get_swiftpm_options(swift_exec, args)
+    env = get_swiftpm_environment_variables(swift_exec, args)
+
     bin_path = swiftpm_bin_path(swift_exec, swiftpm_args, env)
     swiftpm_args += ['-Xswiftc', '-no-toolchain-stdlib-rpath']
```
benlangmuir, line 205: Incidentally, I guess this causes rebuilds during `install` actions since the arguments have changed?

Utilities/build-script-helper.py
```diff
@@ -8,230 +8,306 @@
...
-      # Dispatch headers
-      '-Xcxx', '-I', '-Xcxx',
-      os.path.join(args.toolchain, 'lib', 'swift'),
-      # For <Block.h>
-      '-Xcxx', '-I', '-Xcxx',
-      os.path.join(args.toolchain, 'lib', 'swift', 'Block'),
+        return arg
+
+
+def check_call(cmd: List[str], cwd: Optional[str] = None, env: Dict[str, str] = dict(os.environ), verbose: bool = False):
```
benlangmuir, line 30: Is there a reason you're defaulting to `os.environ` instead of `None` to match the underlying `subprocess.check_call`?

Utilities/build-script-helper.py
```diff
@@ -8,230 +8,306 @@
...
-    test_args = swiftpm_args
-    test_args += ['--parallel', '--disable-testable-imports']
-    swiftpm('test', swift_exec, test_args, env)
-  elif args.action == 'install':
+    
+    cmd = [
+        swift_exec, 'test', 
+        '--parallel', 
+        '--disable-testable-imports',
+        '--test-product', 'SourceKitLSPTests'
```
benlangmuir, line 186: Does this test only the `SourceKitLSPTests` target?

Utilities/build-script-helper.py
ahoppen: I actually like to keep them separate. That way each method is easier to reason about on its own. Otherwise the unified function becomes larger again and you don’t have a clear separation between the part that sets environments and the part that creates the command line arguments.

Utilities/build-script-helper.py
ahoppen: Yes, I think so. Looking into the actual build time of SourceKit-LSP is the next thing I want to do.

Utilities/build-script-helper.py
ahoppen: No, no particular reason. I changed it now.

Utilities/build-script-helper.py
ahoppen: Thanks for catching this, this should have been `SourceKitLSPPackageTests`, which includes all tests of the package.

ahoppen: I should have probably also put this change in the PR that actually starts doing a unified build because it’s only necessary there to make sure we’re not testing the other projects as well.

benlangmuir: Doesn't matter to me if you want to keep it in this PR or not.

Utilities/build-script-helper.py
benlangmuir: My concern is that you can forget to use the env variables at all, especially when things like check_call have default values for `env`.  But I don't feel strongly enough about it to block this PR if you disagree.

ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
ahoppen: Let’s keep it in this PR.

Utilities/build-script-helper.py
ahoppen: I’ll just remove the default value for `env` from `check_call`.

ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #600, opened 2022-07-31
SKCoreTests: skip a test on Windows

compnerd:
This test seems to hang for some reason.  Skip the test for the time
being to allow us to execute the test suite until we can investigate
this issue.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #602, opened 2022-08-02
Empty TextEdits now return nil to avoid diagnostic crashes

ianhanken:
This adds a check when creating a `TextEdit` to return `nil` if both the fixit and the replacement being used to create a `TextEdit` are empty. A `TextEdit` being created like this and then used in a diagnostic can cause a crash.

### DIFF
--- a/Sources/SourceKitLSP/Swift/Diagnostic.swift
+++ b/Sources/SourceKitLSP/Swift/Diagnostic.swift
@@ -100,6 +100,12 @@ extension TextEdit {
       // Snippets are only suppored in code completion.
       // Remove SourceKit placeholders from Fix-Its because they can't be represented in the editor properly.
       let replacementWithoutPlaceholders = rewriteSourceKitPlaceholders(inString: replacement, clientSupportsSnippets: false)
+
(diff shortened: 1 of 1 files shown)


DavidGoldman:
@swift-ci please test

ahoppen (commented):
Do you have an example that generates a fixit that has `length == 0` and no `replacementWithoutPlaceholders`?

Also: Where does such a `TextEdit` cause a crash? Inside sourcekit-lsp or inside the editor?

ianhanken:
This was found by accident, and I believe it was due to other changes that I was able to create an example where both the fixit had `length == 0` and `replacementWithoutPlaceholders` was empty. Using `apple/main`, I can't find a specific example, but this is meant to avoid the crash if one does pop up.

This `TextEdit` causes a SourceKit-LSP crash [here](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/Swift/Diagnostic.swift#L83). If a diagnostic causes this, it will also cause a crash loop since this will cause a crash every time SourceKit-LSP restarts.

ahoppen (approved):
I see, makes sense 👍 Thanks for the fix.

PR #603, opened 2022-08-03
SourceKitLSPTests: remove many cases of forced unwrapping

compnerd:
When running the test suite on Windows, these forced unwraps would fail.
Adopt `throws` on the cases and be more lenient of the failure as that
is reported as a test failure.  This allows us to run the test suite to
completion even if it fails.


compnerd:
@swift-ci please test

benlangmuir (approved):

ahoppen (approved):

PR #604, opened 2022-08-04
Don’t build SourceKitLSPPackageTests in build-script-helper’s build action

ahoppen:
Building SourceKitLSPPackageTests with cross-compilation enabled fails with
```
error: Could not find target named 'SourceKitLSPPackageTests_586B622B321A3B86_PackageProduct'
error: fatalError
```

For testing, we work around this by disabling cross-compilation. This is not possible for building, so we need to explicitly build all products.

This fixes a current toolchain build failure.


ahoppen:
@swift-ci Please test

ahoppen:
Merging this now to unblock the Swift package bot.

PR #608, opened 2022-08-09
Add rpath of /usr/lib/swift and swift-5.5 compatibility library

ahoppen:
Without it, sourcekit-lsp crashes on launch with the same errors as in https://github.com/apple/swift-package-manager/issues/5467.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -107,9 +107,9 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
 
     if platform.system() == 'Darwin':
         swiftpm_args += [
-            # Relative library rpath for swift; will only be used when /usr/lib/swift
-            # is not available.
+            '-Xlinker', '-rpath', '-Xlinker', '/usr/lib/swift',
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -107,9 +107,9 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
 
     if platform.system() == 'Darwin':
         swiftpm_args += [
-            # Relative library rpath for swift; will only be used when /usr/lib/swift
-            # is not available.
+            '-Xlinker', '-rpath', '-Xlinker', '/usr/lib/swift',
```
benlangmuir, line 110: Why do we need this one? It's already there by default when I build.

ahoppen: For reasons I don’t fully understand it’s not present if we build in CI. I also see the rpath when I build locally. I suspect that it’s because CI is still running macOS 11.5.2, but I haven’t verified that.

benlangmuir (approved):

PR #610, opened 2022-08-12
Provide semantic highlighting for operators in declarations

### LINKED ISSUE #595: operator and operator lexeme declarations should be highlighted as identifiers
> note: this issue was split off from https://github.com/swift-server/vscode-swift/issues/366

**Describe the bug**

operator lexeme declarations should be highlighted as identifiers.

```
infix operator ?= :ComparisonPrecedence
```

Screenshot from 2022-07-22 14-20-47

there are **three** bugs here. two are vscode-swift bugs, and the other one is a sourcekit-lsp bug:

| lexeme | `?=` | `:` | `ComparisonPrecedence` |
| --- |  --- | --- | --- |
| current semantic highlight | none | none | `identifier` |
| expected semantic highlight | `identifier` | none | `identifier` |
| sourcekit-lsp bug? | **yes** | no | no |
| current fallback label | `operator` | `invalid` | `invalid` |
| expected fallback label | `operator` | none | none |
| [vscode](https://github.com/microsoft/vscode/blob/main/extensions/swift/syntaxes/swift.tmLanguage.json) bug? | no | **yes** | **yes**<sup>[1]</sup> |

> [1]: has no effect, because the correct sourcekit-lsp highlight overrides the wrong fallback label)

the wrong highlight for the colon is a vscode problem, and probably not in our power to fix. the only way to correct this behavior would be to make sourcekit-lsp provide more comprehensive highlights (classifying punctuation and identifiers in a way it does not do currently) and migrate the community to swift-specific syntax coloring themes that do not suffer from this defect. however, the operator lexeme highlight *is* within our power to fix.

this also seems to affect operator declarations, but the impact is mitigated because the fallback grammar is able to fill in for the semantic highlighter in this case.

Screenshot from 2022-07-22 14-21-07

**Environment**
 - OS: Ubuntu 20.04
 - vscode-swift version: 0.6.0
 
```
$ swift --version
Swift version 5.8-dev (LLVM 396c6969ee90ef4, Swift 02efc3a2d28e348)
Target: x86_64-unknown-linux-gnu
$ swiftenv version
DEVELOPMENT-SNAPSHOT-2022-07-06-a 
$ code --version
1.69.2
3b889b090b5ad5793f524b5d1d39fda662b96a2a
x64
```

ahoppen:
Adopts https://github.com/apple/swift/pull/60210

Resolves #595
rdar://97961816

### DIFF
--- a/Sources/SourceKitD/sourcekitd_uids.swift
+++ b/Sources/SourceKitD/sourcekitd_uids.swift
@@ -285,6 +285,7 @@ public struct sourcekitd_values {
   public let syntaxtype_doccomment: sourcekitd_uid_t
   public let syntaxtype_doccomment_field: sourcekitd_uid_t
   public let syntaxtype_keyword: sourcekitd_uid_t
+  public let syntaxtype_operator: sourcekitd_uid_t
   public let syntaxtype_number: sourcekitd_uid_t
   public let syntaxtype_string: sourcekitd_uid_t
   public let syntaxtype_string_interpolation_anchor: sourcekitd_uid_t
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #613, opened 2022-08-19
LSP: always use FSR for URI

compnerd:
When converting the URI to a path string, ensure that we convert to the
file system representation.  This is important as this ensures that we
are always passing SourceKit the native path string.  With this change,
the code completion behaviour for the LSP test suite on Windows is
repaired.


compnerd:
@swift-ci please test

ahoppen (approved):
Seems reasonable to me

PR #614, opened 2022-08-23
SourceKitD,SourceKitLSP: replace use of `Lock` with `NSLock`

compnerd:
Replace the use of the deprecated `Lock` from swift-tools-support-core
with `NSLock`.  The benefit of the deprecated interface was the
`withLock` function which is easily replicated and very unlikely to
change.  By doing so we avoid the warnings about `Lock` being deprecated
when building SourceKit-LSP.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #615, opened 2022-08-25
Tests: correct path separator for Windows

compnerd:
Windows uses `;` as a path separator rather than `:`.  Adjust the test
accordingly.  This allows the test to pass on Windows.

### DIFF
--- a/Tests/SKCoreTests/ToolchainRegistryTests.swift
+++ b/Tests/SKCoreTests/ToolchainRegistryTests.swift
@@ -204,7 +204,7 @@ final class ToolchainRegistryTests: XCTestCase {
     let separator: String = ":"
 #endif
 
-    try! ProcessEnv.setVar("SOURCEKIT_PATH", value: "/bogus\(separator)\(binPath)\(separator)/bogus2")
+    try! ProcessEnv.setVar("SOURCEKIT_PATH", value: ["/bogus", binPath.pathString, "/bogus2"].joined(separator: separator))
     defer { try! ProcessEnv.setVar("SOURCEKIT_PATH", value: "") }
 
     tr.scanForToolchains(fs)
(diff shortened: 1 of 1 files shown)


Tests/SKCoreTests/ToolchainRegistryTests.swift
```diff
@@ -198,7 +198,13 @@ final class ToolchainRegistryTests: XCTestCase {
...
     XCTAssert(tr.toolchains.isEmpty)
 
-    try! ProcessEnv.setVar("SOURCEKIT_PATH", value: "/bogus:\(binPath):/bogus2")
+#if os(Windows)
+    let separator: String = ";"
+#else
+    let separator: String = ":"
+#endif
+
+    try! ProcessEnv.setVar("SOURCEKIT_PATH", value: "/bogus\(separator)\(binPath)\(separator)/bogus2")
```
ahoppen, line 207: I think this would read nicer as
```suggestion
    try! ProcessEnv.setVar("SOURCEKIT_PATH", value: ["/bogus", "\(binPath)", "/bogus2"].joined(separator: separator))
```

Same below

ahoppen (approved):
Stylistic comment, otherwise LGTM.

Tests/SKCoreTests/ToolchainRegistryTests.swift
compnerd: I definitely have to agree, thanks for the suggestion!

Tests/SKCoreTests/ToolchainRegistryTests.swift
```diff
@@ -217,7 +223,7 @@ final class ToolchainRegistryTests: XCTestCase {
     XCTAssertNil(tc.libIndexStore)
 
     let binPath2 = AbsolutePath("/other/my_toolchain/bin")
-    try! ProcessEnv.setVar("SOME_TEST_ENV_PATH", value: "/bogus:\(binPath2):/bogus2")
+    try! ProcessEnv.setVar("SOME_TEST_ENV_PATH", value: "/bogus\(separator)\(binPath2)\(separator)/bogus2")
```
compnerd, line 226:
```suggestion
    try! ProcessEnv.setVar("SOME_TEST_ENV_PATH", value: ["/bogus", binPath2, "bogus2"].joined(separator: separator))
```

Tests/SKCoreTests/ToolchainRegistryTests.swift
compnerd: Going to tweak it a bit further, I think that using `binPath.pathString` is nicer rather than the string interpolation for the string conversion.

compnerd:
@swift-ci please test

PR #617, opened 2022-08-27
SKCoreTests: use a tiny PE binary for placeholder executables

compnerd:
Use a tiny PE binary for the placeholder executables for tests as that
serves as the signal for executables on Windows (content is processed
rather than attributes).  This is required for many of the tests to be
handled properly on Windows.


ahoppen (approved):
Just to be clear, we never execute that binary you added, right?

compnerd:
Coreect, the binary is never executed, only checked for being executable.  We use the same binary in SPM for the test suite FWIW.

compnerd:
@swift-ci please test

PR #619, opened 2022-08-28
LanguageServerProtocolJSONRPCTests: silence a warning

compnerd:
`Sleep` is a `void` returning function, which makes the blackhole
redundant.  Remove the unnecessary assignment which also silences
the warning associated with it.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #623, opened 2022-09-02
SourceKitLSPTests: use a UUID in place of `#function`

compnerd:
Rather than using `#function` use a UUID for the generated unique name.
This avoids placing `:`, `(`, and `)` in the file path which are not
portable characters.


compnerd:
@swift-ci please test

ahoppen (approved):

ahoppen:
rdar://99561862

PR #628, opened 2022-09-08
SKTestSupport: form native path for root of test project

compnerd:
The root of the test project would be in POSIX spelling which would cause mismatches between the formed paths and the value returned from the LSP resulting in test failures on Windows.

### DIFF
--- a/Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
+++ b/Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
@@ -53,8 +53,8 @@ public final class SKSwiftPMTestWorkspace {
   public init(projectDir: URL, tmpDir: URL, toolchain: Toolchain, testServer: TestSourceKitServer? = nil) throws {
     self.testServer = testServer ?? TestSourceKitServer(connectionKind: .local)
 
-    self.projectDir = projectDir
-    self.tmpDir = tmpDir
+    self.projectDir = URL(fileURLWithPath: resolveSymlinks(AbsolutePath(projectDir.path)).pathString)
+    self.tmpDir = URL(fileURLWithPath: resolveSymlinks(AbsolutePath(tmpDir.path)).pathString)
     self.toolchain = toolchain
 
     let fm = FileManager.default


compnerd:
@swift-ci please test

Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
```diff
@@ -67,7 +67,7 @@ public final class SKSwiftPMTestWorkspace {
     let databaseDir = tmpDir.appendingPathComponent("db", isDirectory: true)
     try fm.createDirectory(at: databaseDir, withIntermediateDirectories: true, attributes: nil)
 
-    self.sources = try TestSources(rootDirectory: sourceDir)
+    self.sources = try TestSources(rootDirectory: URL(fileURLWithPath: resolveSymlinks(AbsolutePath(sourceDir.path)).pathString))
```
benlangmuir, line 70: It's odd to me to resolve this so late - should we canonicalize projectDir and tmpDir immediately?  Would also be good to have a comment since it's non-obvious that this is about getting a native windows path.  Is there no better way to get a native path than to "resolve symlinks"?

compnerd: I can certainly try to resolve this earlier, I think that this question will keep coming up a few times as I work through the patch series.  Where would you recommend to resolve it?

The conversion to a native path string occurs by means of the `AbsolutePath`.  The `resolveSymlinks` is required in order to peer through symbolic links, which occurs at various points by SourceKit and SourceKit-LSP.  Failure to do so results in some path strings being resolved and some being unresolved, which subsequently force test failures.

benlangmuir: Maybe we should do it on the inputs to this initializer and then use those fixed up base paths?  That way we have a more consistent view of the paths everywhere underneath this class.

```swift
    self.projectDir = URL(fileURLWithPath: resolveSymlinks(AbsolutePath(projectDir.path))
    self.tmpDir = URL(fileURLWithPath: resolveSymlinks(AbsolutePath(tmpDir.path))
```

compnerd:
@swift-ci please test

benlangmuir (approved):

ahoppen (approved):

PR #629, opened 2022-09-09
SKCore: introduce `swift` member to `Toolchain`

compnerd:
This adds a `swift` member to the `Toolchain` type to allow invocations of the swift driver to be constructed for tests.

### DIFF
--- a/Sources/SKCore/Toolchain.swift
+++ b/Sources/SKCore/Toolchain.swift
@@ -43,6 +43,9 @@ public final class Toolchain {
   /// The path to the Clang compiler if available.
   public var clang: AbsolutePath?
 
+  /// The path to the Swift driver if available.
+  public var swift: AbsolutePath?
+
   /// The path to the Swift compiler if available.
   public var swiftc: AbsolutePath?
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

benlangmuir:
`swiftc` should already be the path to the driver. What's going on?

compnerd:
`swiftc` is the path to the frontend, not the driver.  There are a few tests which explicitly wish to use the driver `swift`.  Dropping the `c` by means of `dropLast` is a problem if you have a file extension (i.e. `.exe`).  This adds a proper member to use to reference that path.

benlangmuir (approved):

benlangmuir:
> `swiftc` is the path to the frontend

Running swiftc runs the driver not the frontend. The fact `swiftc` may be a symlink to the `swift-frontend` file is an implementation detail, and `swift-frontend` immediately executes the driver when run as `swiftc`.

> There are a few tests which explicitly wish to use the driver swift.

I see, so this is about commands like `swift build`. Seems reasonable.

ahoppen (approved):

PR #630, opened 2022-09-09
explicitly import TSCBasic types

compnerd:
Explicitly import interfaces from TSCBasic which now allows us to identify all the swift-tools-support-core interfaces which are in use in SourceKit-LSP.

### DIFF
--- a/Sources/SKCore/BuildServerBuildSystem.swift
+++ b/Sources/SKCore/BuildServerBuildSystem.swift
@@ -16,7 +16,14 @@ import LanguageServerProtocol
 import LanguageServerProtocolJSONRPC
 import LSPLogging
 import SKSupport
-import TSCBasic
+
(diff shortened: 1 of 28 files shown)


compnerd:
CC: @CodaFi

compnerd:
@swift-ci please test

ahoppen:
LGTM in general. But why are we doing this? Are we planning to drop the dependency on TSC?

compnerd:
Yes, it would be nice to be able to drop the t-s-c dependency at some point.  Having the types be explicitly stated helps with identifying single use types that can be migrated out of the module.  As an example, this made it easy to identify that the `SQLite` API from TSC could be migrated out of the package to the single package user (SPM).

compnerd:
@swift-ci please test

ahoppen (approved):
Makes sense. Thank you.

PR #631, opened 2022-09-10
SKTestSupport,SourceKitLSPTests: use new `swift` member

compnerd:
Migrate to `swift` member for the swift driver rather than trying to create it by dropping the last character in the path to the swift compiler which may be part of the file extension.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #632, opened 2022-09-13
SKCoreTests: adjust toolchain library stubs for Windows

compnerd:
The toolchain library runtime components are installed into the binaries directory and the import library (link components) are installed into the lib directory.  Darwin splits the runtime (dylib) and link time (tbd) across the SDK and the installation and Linux uses the same file for both.  Account for the difference on Windows to correct some of the test expectations.

### DIFF
--- a/Tests/SKCoreTests/ToolchainRegistryTests.swift
+++ b/Tests/SKCoreTests/ToolchainRegistryTests.swift
@@ -560,16 +560,25 @@ private func makeToolchain(
     makeExec(binPath.appending(component: "swiftc\(execExt)"))
   }
 
-  let dylibExt = Platform.currentPlatform?.dynamicLibraryExtension ?? ".so"
+  let dylibSuffix = Platform.currentPlatform?.dynamicLibraryExtension ?? ".so"
 
   if sourcekitd {
     try! fs.createDirectory(libPath.appending(component: "sourcekitd.framework"))
     try! fs.writeFileContents(libPath.appending(components: "sourcekitd.framework", "sourcekitd") , bytes: "")
   }
   if sourcekitdInProc {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/SKCoreTests/ToolchainRegistryTests.swift
```diff
@@ -560,16 +560,25 @@ private func makeToolchain(
...
+  let dylibSuffix = Platform.currentPlatform?.dynamicLibraryExtension ?? ".so"
 
   if sourcekitd {
     try! fs.createDirectory(libPath.appending(component: "sourcekitd.framework"))
     try! fs.writeFileContents(libPath.appending(components: "sourcekitd.framework", "sourcekitd") , bytes: "")
   }
   if sourcekitdInProc {
-    try! fs.writeFileContents(libPath.appending(component: "libsourcekitdInProc\(dylibExt)") , bytes: "")
+#if os(Windows)
+    try! fs.writeFileContents(binPath.appending(component: "sourcekitdInProc\(dylibSuffix)") , bytes: "")
```
benlangmuir, line 571: This matches how Toolchain.swift works, but  I wonder if we should instead have
* libPath = binPath on Windows
* dylibPrefix = "lib" **except** on Windows

I think then the only weird case is that libIndexStore has "lib" even on Windows.

WDYT?

compnerd: I don't think that we want "libPath = binPath" on Windows.  libpath is lib - that is where the import libraries are.  binpath is bin - that is where the dll is.  Having a dylibPrefix makes sense, but it seemed that we would only have a single use given the oddity of `libIndexStore.dll`.

benlangmuir: Okay fair enough

benlangmuir (approved):

PR #633, opened 2022-09-15
SKCoreTests,SKSwiftPMWorkspaceTests: native path spelling changes

compnerd:
Adjust a couple of sites to use native path spelling rather than POSIX paths.  This is required to ensure that the tests pass on Windows which uses `\` rather than `/` for the path separator.


compnerd:
@swift-ci please test

ahoppen (approved):

PR #637, opened 2022-09-21
SKCore: use `resolvingSymlinks` to resolve symlinks

compnerd:
Rather than using the `resolvingSymlinksInPath` on `URL` use `resolvingSymlink` from tools-support-core as the semantics for the two differ ever so slightly which matters for Windows (the former expands symlinks on the terminal arc iterative where as the latter simply expands the path).


compnerd:
@swift-ci please test

ahoppen (approved):

PR #638, opened 2022-09-22
SourceKitLSPTests: convert to native path spellings

compnerd:
Explicitly convert paths from POSIX spelling to native spelling in a number of cases.  These changes help improve the test coverage on Windows.


compnerd:
@swift-ci please test

Tests/SourceKitLSPTests/CallHierarchyTests.swift
```diff
@@ -80,12 +81,12 @@ final class CallHierarchyTests: XCTestCase {
         kind: kind,
         tags: nil,
         detail: detail,
-        uri: location.uri,
+        uri: DocumentURI(URL(fileURLWithPath: resolveSymlinks(AbsolutePath(location.uri.pseudoPath)).pathString)),
```
ahoppen, line 84: Do you think it would make sense to extract this snippet to a `nativeURI` computed property on `DocumentURI`. I think that would make it clearer why this pattern is needed. It might also be worth migrating your previous PRs to that computed property.

compnerd: I can see that as being a good approach.  Thanks for the idea!

compnerd: Replaced the appropriate cases.

compnerd:
@swift-ci please test

Tests/SourceKitLSPTests/ImplementationTests.swift
```diff
@@ -36,7 +37,8 @@ final class ImplementationTests: XCTestCase {
       ws.testLoc(name)
     }  
     func loc(_ name: String) -> Location {
-      Location(badUTF16: ws.testLoc(name))
```
ahoppen, line 39: Shouldn’t `ws.testLoc` already return a native URI?

Tests/SourceKitLSPTests/SourceKitTests.swift
```diff
@@ -10,10 +10,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+import ISDBTestSupport
```
ahoppen, line 13: Why do we need to import `ISDBTestSupport` here?

compnerd: For `TestLocation`.

Tests/SourceKitLSPTests/ImplementationTests.swift
compnerd: It currently does not.  I can certainly try to see if we can get away with it, but that would change a number of location spellings, and I am worried that it might result in new failures.

ahoppen: Oh, OK. I just thought it might be easier if we change the return value of `testLoc`. I’m fine to keep it as-is for now.

Tests/SourceKitLSPTests/SourceKitTests.swift
ahoppen: 👍 Makes sense

ahoppen (approved):

PR #639, opened 2022-09-25
LanguageServerProtocolJSONRPC: avoid failing cast on Windows

compnerd:
This error code is poorly mapped on Windows and results in a runtime failure.  Avoid the cast to improve stability.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -132,9 +132,11 @@ public final class JSONRPCConnection {
 
     receiveIO.read(offset: 0, length: Int.max, queue: queue) { done, data, errorCode in
       guard errorCode == 0 else {
+#if !os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

ahoppen:
Could it be that you committed the wrong change? The description is talking about a cast but I don’t see a cast being changed in the diff.

compnerd:
The cast is on the Foundation side, where it tries to convert the error code to `ECANCELED`.  I'm happy to change the commit message, but not really sure how to make it clearer :-(.

ahoppen (approved):

compnerd:
It is an explicit conversion that is failing for some reason.  https://github.com/apple/swift-corelibs-foundation/blob/9bc1a18236f643b73eb60f9633f2687e45dab487/Sources/Foundation/NSError.swift#L1319

ahoppen:
Ah, I see. The access of `POSIXError.ECANCELED` causes the crash. I thought the comparison using `!=` caused the crash.

PR #642, opened 2022-09-30
SourceKitLSP: use `AbsolutePath` to compute the filename

compnerd:
Rather than treating the filename as an opaque string and splitting on `/`, use `AbsolutePath` to perform the `basename` operation on the path. This ensures that we split the path properly on platforms which do not use the POSIX path separator `/`.


compnerd:
@swift-ci please test

ahoppen (approved):

Issue #635, opened 2022-09-17
SourceKit Language Server server crashes


LazyOtto:
Repeatable crash when used with VSCode on Linux Mint.
Installed with: swift-5.7-RELEASE-ubuntu22.04.tar.gz
Code to cause it to crash - second line below:
```
var someInts: [Int] = []
someInts.append()
// someInts.append(contentsOf: [5, 7, 9])
print("An array of \(someInts)) with count of: \(someInts.count)")
```
Stack trace of thread 29377:
#0  0x0000555b412ae48e n/a (/home/user/Applications/swift-5.7-RELEASE-ubuntu22.04/usr/bin/sourcekit-lsp + 0x9e648e)

```
Core was generated by `sourcekit-lsp'.
Program terminated with signal SIGILL, Illegal instruction.
#0  Swift runtime failure: precondition failure () at /home/build-user/sourcekit-lsp/Sources/SourceKitLSP/Swift/Diagnostic.swift:83
83	/home/build-user/sourcekit-lsp/Sources/SourceKitLSP/Swift/Diagnostic.swift: No such file or directory.
[Current thread is 1 (Thread 0x7f2b92cf7640 (LWP 29377))]
```

LazyOtto:
I replaced the precondition check in "fixitTitle" as below. Have not run any suite of tests on the patch other than the case which I stumbled upon above.
```
  public static func fixitTitle(replace oldText: String, with newText: String) -> String {
    switch (oldText.isEmpty, newText.isEmpty) {
    case (false, false):
      return "Replace '\(oldText)' with '\(newText)'"
    case (false, true):
      return "Remove '\(oldText)'"
    case (true, false):
      return "Insert '\(newText)'"
    case (true, true):
      // preconditionFailure("FixIt makes no changes")
      log("FixIt makes no changes", level: .warning)
      return ""
    }
  }
```

LazyOtto:
Tests supplied with the package pass:
```
Test Suite 'All tests' passed at 2022-09-17 10:22:47.836
Executed 249 tests, with 4 tests skipped and 0 failures (0 unexpected) in 32.119 (32.119) seconds
```

ahoppen:
rdar://100158700

ahoppen:
I suspect you are using an editor that has snippets disabled, in which case this issue should be fixed by https://github.com/apple/sourcekit-lsp/pull/602.

To verify that this is indeed the case could you:
[...]

PR #646, opened 2022-10-07
[5.7] Empty TextEdits now return nil

ahoppen:
Cherry-picks https://github.com/apple/sourcekit-lsp/pull/602 to release/5.7.

---

This adds a check when creating a `TextEdit` to return `nil` if both the fixit and the replacement being used to create a `TextEdit` are empty. A `TextEdit` being created like this can cause a crash.

Resolves https://github.com/apple/sourcekit-lsp/issues/635


ahoppen:
@swift-ci Please test

benlangmuir (approved):

PR #647, opened 2022-10-10
Implement textDocument/declaration

CodaFi:
A declaration request is similar to a definition request, except that it is expected to return (potentially) many results across the workspace for a given reference. For example, an inline function or macro may have many declarations in the workspace, but only one "good" or canonical definition. For now, this is only implemented by forwarding the request on to clangd since I'm unfamiliar with a SourceKit query for this.

For languages like Swift that lack such a sharp declaration/definition split, we could potentially use this request to provide navigable metadata on linked definitions. For example, the declaration for a type reference would include all extensions of that type in the workspace.

### DIFF
--- a/Tests/SourceKitLSPTests/SourceKitTests.swift
+++ b/Tests/SourceKitLSPTests/SourceKitTests.swift
@@ -314,10 +314,7 @@ final class SKTests: XCTestCase {
       textDocument: mainLoc.docIdentifier, position: includePosition)
     let resp = try withExtendedLifetime(ws) { try ws.sk.sendSync(goToInclude) }
 
-    guard let locationsOrLinks = resp else {
-      XCTFail("No response for go-to-#include")
-      return
-    }
+    let locationsOrLinks = try XCTUnwrap(resp, "No response for go-to-#include")
     switch locationsOrLinks {
     case .locations(let locations):
       XCTAssert(!locations.isEmpty, "Found no locations for go-to-#include")
@@ -346,10 +343,7 @@ final class SKTests: XCTestCase {
       textDocument: refLoc.docIdentifier, position: refPos)
     let resp = try withExtendedLifetime(ws) { try ws.sk.sendSync(goToDefinition) }
 
-    guard let locationsOrLinks = resp else {
-      XCTFail("No response for go-to-definition")
-      return
-    }
+    let locationsOrLinks = try XCTUnwrap(resp, "No response for go-to-definition")
     switch locationsOrLinks {
     case .locations(let locations):
       XCTAssert(!locations.isEmpty, "Found no locations for go-to-definition")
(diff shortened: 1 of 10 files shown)


CodaFi:
@swift-ci test

Tests/SourceKitLSPTests/SourceKitTests.swift
```diff
@@ -363,4 +363,37 @@ final class SKTests: XCTestCase {
...
   }
+
+  func testClangdGoToDeclaration() throws {
+    guard let ws = try staticSourceKitTibsWorkspace(name: "BasicCXX") else { return }
+    guard ToolchainRegistry.shared.default?.clangd != nil else { return }
+
+    let mainLoc = ws.testLoc("Object:ref:newObject")
+    let expectedDoc = ws.testLoc("Object:decl:newObject").docIdentifier.uri
+    let includePosition =
+        Position(line: mainLoc.position.line, utf16index: mainLoc.utf16Column + 2)
```
ahoppen, line 374: Why do you need to do `+ 2` to the column here?

Tests/SourceKitLSPTests/SourceKitTests.swift
```diff
@@ -363,4 +363,37 @@ final class SKTests: XCTestCase {
...
+    try ws.openDocument(mainLoc.url, language: .c)
+
+    let goToInclude = DeclarationRequest(
+      textDocument: mainLoc.docIdentifier, position: includePosition)
+    let resp = try! ws.sk.sendSync(goToInclude)
+
+    guard let locationsOrLinks = resp else {
+      XCTFail("No response for go-to-declaration")
+      return
+    }
```
ahoppen, line 385: I think this can be abbreviated to 
```suggestion
    let locationsOrLinks = try XCTUnwrap(resp)
```

ahoppen (commented):
Thanks. Mostly looks good to me.

CodaFi:
@swift-ci test

Tests/SourceKitLSPTests/SourceKitTests.swift
ahoppen: Ah, could you add a comment with that explanation?

Tests/SourceKitLSPTests/SourceKitTests.swift
```diff
@@ -346,10 +343,7 @@ final class SKTests: XCTestCase {
       textDocument: refLoc.docIdentifier, position: refPos)
     let resp = try withExtendedLifetime(ws) { try ws.sk.sendSync(goToDefinition) }
 
-    guard let locationsOrLinks = resp else {
-      XCTFail("No response for go-to-definition")
-      return
-    }
+    let locationsOrLinks = try XCTUnwrap(resp,  "No response for go-to-definition")
```
ahoppen, line 346:
```suggestion
    let locationsOrLinks = try XCTUnwrap(resp, "No response for go-to-definition")
```

CodaFi:
@swift-ci test

PR #651, opened 2022-10-16
Pull SwiftSyntax into SourceKit-LSP

CodaFi:
The general idea is to try to use the syntax tree to provide an in-process source of truth about syntactic and structural document requests. Having the syntax tree in process also enables syntactic transformations and analyses that can be used to provide code actions, programmatic language features like (keyword) hovering, etc. both natively in Swift and without having to incur a hop out to SourceKit.

To that end, re-core the "lexical" token stream on top of SwiftSyntax and call into the parser to (re-)generate these "lexical tokens" when the document changes. Leave the semantic tokens alone since they take some non-zero amount of name lookup to fully resolve references to fully replicate.

A follow-on will demonstrate what we can start to do with this infrastructure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ find_package(LLBuild QUIET)
 find_package(ArgumentParser CONFIG REQUIRED)
 find_package(SwiftCollections QUIET)
 find_package(SwiftSystem CONFIG REQUIRED)
+find_package(SwiftSyntax CONFIG REQUIRED) 
 
 include(SwiftSupport)
 
--- a/Sources/SourceKitLSP/CMakeLists.txt
+++ b/Sources/SourceKitLSP/CMakeLists.txt
@@ -41,6 +41,11 @@ target_link_libraries(SourceKitLSP PUBLIC
   SKCore
   SKSwiftPMWorkspace
   SourceKitD
+  SwiftSyntax::SwiftBasicFormat
+  SwiftSyntax::SwiftParser
+  SwiftSyntax::SwiftDiagnostics
+  SwiftSyntax::SwiftSyntax
   TSCUtility)
 target_link_libraries(SourceKitLSP PRIVATE
   $<$<NOT:$<PLATFORM_ID:Darwin>>:FoundationXML>)
+
--- a/Sources/SourceKitLSP/DocumentTokens.swift
+++ b/Sources/SourceKitLSP/DocumentTokens.swift
@@ -11,31 +11,100 @@
 //===----------------------------------------------------------------------===//
 
 import LanguageServerProtocol
+import SwiftSyntax
 
 /// Syntax highlighting tokens for a particular document.
 public struct DocumentTokens {
-  /// Lexical tokens, e.g. keywords, raw identifiers, ...
-  public var lexical: [SyntaxHighlightingToken] = []
+  /// The syntax tree representing the entire document.
+  public var syntaxTree: SourceFileSyntax?
   /// Semantic tokens, e.g. variable references, type references, ...
   public var semantic: [SyntaxHighlightingToken] = []
+}
 
-  private var merged: [SyntaxHighlightingToken] {
-    lexical.mergingTokens(with: semantic)
+extension DocumentSnapshot {
+  /// Computes an array of syntax highlighting tokens from the syntax tree that
+  /// have been merged with any semantic tokens from SourceKit. If the provided
+  /// range is non-empty, this function restricts its output to only those
+  /// tokens whose ranges overlap it. If no range is provided, tokens for the
+  /// entire document are returned.
+  ///
+  /// - Parameter range: The range of tokens to restrict this function to, if any.
+  /// - Returns: An array of syntax highlighting tokens.
+  public func mergedAndSortedTokens(in range: Range<Position>? = nil) -> [SyntaxHighlightingToken] {
+    guard let tree = self.tokens.syntaxTree else {
+      return self.tokens.semantic
+    }
+    let range = range.flatMap({ $0.byteSourceRange(in: self) })
+             ?? ByteSourceRange(offset: 0, length: tree.byteSize)
+    return tree
+      .classifications(in: range)
+      .flatMap({ $0.highlightingTokens(in: self) })
+      .mergingTokens(with: self.tokens.semantic)
+      .sorted { $0.start < $1.start }
   }
-  public var mergedAndSorted: [SyntaxHighlightingToken] {
-    merged.sorted { $0.start < $1.start }
-  }
+}
 
-  /// Modifies the syntax highlighting tokens of each kind
-  /// (lexical and semantic) according to `action`.
-  public mutating func withMutableTokensOfEachKind(_ action: (inout [SyntaxHighlightingToken]) -> Void) {
-    action(&lexical)
-    action(&semantic)
+extension Range where Bound == Position {
+  fileprivate func byteSourceRange(in snapshot: DocumentSnapshot) -> ByteSourceRange? {
+    return snapshot.utf8OffsetRange(of: self).map({ ByteSourceRange(offset: $0.startIndex, length: $0.count) })
   }
+}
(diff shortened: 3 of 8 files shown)


compnerd (changes requested):
Please update the CMake builds, the build-windows-toolchain.bat in apple/swift, and this will require coordination as https://GitHub.com/compnerd/swift-build will need updates to keep the official builds from breaking.

ahoppen (commented):
A few from my comments on https://github.com/apple/sourcekit-lsp/pull/652 also apply here. I only saw this PR after adding my review comments to the other one.

CodaFi:
apple/swift-syntax#970

@swift-ci test

CodaFi:
https://github.com/apple/swift-syntax/pull/970

@swift-ci test

Sources/SourceKitLSP/DocumentTokens.swift
```diff
@@ -11,31 +11,92 @@
...
+  /// entire document are returned.
+  ///
+  /// - Parameter range: The range of tokens to restrict this function to, if any.
+  /// - Returns: An array of syntax highlighting tokens.
+  public func mergedAndSortedTokens(in range: Range<Position>? = nil) -> [SyntaxHighlightingToken] {
+    guard let tree = self.tokens.syntaxTree else {
+      return self.tokens.semantic
+    }
+    let range = range.flatMap({ self.utf8OffsetRange(of: $0) }).map({ ByteSourceRange(offset: $0.startIndex, length: $0.count) })
+             ?? ByteSourceRange(offset: 0, length: tree.byteSize)
```
ahoppen, line 62: Do you think it would be worth putting this in an extension on `Range<Position>`. It seems to me like we’ll be needing it more often in the future.

Something like

```swift
  func byteSourceRange(in snapshot: DocumentSnapshot): ByteSourceRange {
    return = self.flatMap({ snapshot.utf8OffsetRange(of: $0) }).map({ ByteSourceRange(offset: $0.startIndex, length: $0.count) })
             ?? ByteSourceRange(offset: 0, length: snapshot.tree.byteSize)
  }
```

Sources/SourceKitLSP/DocumentTokens.swift
```diff
@@ -11,31 +11,92 @@
...
+      .sorted { $0.start < $1.start }
   }
+}
 
-  // Replace all lexical tokens in `range`.
-  public mutating func replaceLexical(in range: Range<Position>, with newTokens: [SyntaxHighlightingToken]) {
-    lexical.removeAll { $0.range.overlaps(range) }
-    lexical += newTokens
+extension SyntaxClassification {
+  fileprivate var parseKindAndModifiers: (SyntaxHighlightingToken.Kind, SyntaxHighlightingToken.Modifiers)? {
```
ahoppen, line 72: Should this be named `highlightingKindAndModifiers`?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -168,13 +170,12 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   /// Updates the lexical tokens for the given `snapshot`.
   /// Must be called on `self.queue`.
   private func updateLexicalTokens(
```
ahoppen, line 172: This should probably be called `updateSyntacticTokens` now because the new highlighting is slightly more complex than lexical.

ahoppen (commented):
Looks good to me. I’ve got a few comments, mostly around naming.

CodaFi:
I think this should do it. @compnerd please review the cmake changes when you get a chance. There's an oddity around some paths in SwiftSyntax::SwiftParser that's forcing me to link with TARGET_OBJECTS, but otherwise I was able to get a good local build.

@swift-ci test

ahoppen (approved):

CodaFi:
@swift-ci test

DougGregor (approved):
This looks fantastic!

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ find_package(LLBuild QUIET)
 find_package(ArgumentParser CONFIG REQUIRED)
 find_package(SwiftCollections QUIET)
 find_package(SwiftSystem CONFIG REQUIRED)
+include(${SwiftSyntax_DIR}/cmake/SwiftSyntaxTargets.cmake)
```
compnerd, line 24: This needs an associated `find_package`.  Currently this is relying on someone happening to specify a variable that is similarly named and uses that to load a file that may or may not be named the same.

```suggestion
find_package(SwiftSyntax CONFIG REQUIRED)
```

Sources/SourceKitLSP/CMakeLists.txt
```diff
@@ -41,6 +41,16 @@ target_link_libraries(SourceKitLSP PUBLIC
   SKCore
   SKSwiftPMWorkspace
   SourceKitD
+  SwiftSyntax::SwiftBasicFormat
+  $<TARGET_OBJECTS:SwiftSyntax::SwiftParser>
```
compnerd, line 45: Please don't do this.  This is going to cause problems for other platforms.  There is no guarantee that `SwiftParser` is built for static linking; furthermore this says that SourceKitLSP **VENDS** the SwiftParser interfaces.  This should be a static or dynamic link based upon how SwiftParser was built.

```suggestion
  SwiftSyntax::SwiftParser
```

Sources/SourceKitLSP/CMakeLists.txt
```diff
@@ -41,6 +41,16 @@ target_link_libraries(SourceKitLSP PUBLIC
...
+  SwiftSyntax::SwiftSyntax
   TSCUtility)
 target_link_libraries(SourceKitLSP PRIVATE
   $<$<NOT:$<PLATFORM_ID:Darwin>>:FoundationXML>)
+
+add_dependencies(SourceKitLSP
+  SwiftSyntax::SwiftBasicFormat
+  SwiftSyntax::SwiftParser
+  SwiftSyntax::SwiftDiagnostics
+  SwiftSyntax::SwiftSyntax)
```
compnerd, line 56: This seems wrong.  The `target_link_libraries` should handle the dependencies.

```suggestion

```

compnerd (changes requested):

Sources/SourceKitLSP/CMakeLists.txt
CodaFi: When I use SwiftSyntax::SwiftParser I'm greeted with a lovely configure-time error:

```
CMake Error in Sources/SourceKitLSP/CMakeLists.txt:
  Imported target "SwiftSyntax::SwiftParser" includes non-existent path

    "/Users/rwidmann/swift/build/Ninja-DebugAssert/swift-syntax/swift /Users/rwidmann/swift/swift-syntax/Sources/SwiftParser"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



CMake Error in Sources/SourceKitLSP/CMakeLists.txt:
  Imported target "SwiftSyntax::SwiftParser" includes non-existent path

    "/Users/rwidmann/swift/build/Ninja-DebugAssert/swift-syntax/swift /Users/rwidmann/swift/swift-syntax/Sources/SwiftParser"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

CodaFi:
@compnerd and I worked out these issues on a separate channel. Patching Swift Syntax at apple/swift-syntax#991

CodaFi:
https://github.com/apple/swift-syntax/pull/991

@swift-ci test

compnerd (commented):
Thanks, the build side of things look good to me.

PR #652, opened 2022-10-16
Redo the Folding Range Request on top of the Syntax Tree 

CodaFi:
Building upon the infrastructure that requests the lexical structure of the document from SwiftSyntax, redo the folding ranges request using the tree directly. This corrects a number of inconsistencies in the tests mostly due to incorrect SourceLocations in the semantic ASTs and the line comment merging logic.

To extract folding ranges, we were previously opening a temporary document to issue a syntactic structure request to SourceKit.

This change is not NFC - there are a number of quirks of the C++ version of the structure walkers that I think were incorrect and are fixed here. For example, the folding range for the `guard` in the FoldingRangeBase test occurs before the actual `{`! However, this PR does try to recreate the adjacent line comment merging behavior in the C++, but does also correct some bogus source ranges as a result.

Built upon #651

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -961,8 +961,6 @@ extension SwiftLanguageServer {
   }
 
   public func foldingRange(_ req: Request<FoldingRangeRequest>) {
-    let keys = self.keys
-
     queue.async {
       guard let snapshot = self.documentManager.latestSnapshot(req.params.textDocument.uri) else {
         log("failed to find snapshot for url \(req.params.textDocument.uri)")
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -978,143 +963,202 @@ extension SwiftLanguageServer {
...
+              _ = self.addFoldingRange(
+                start: start,
+                end: start + piece.sourceLength.utf8Length,
+                kind: .comment)
+            case .lineComment(_), .docLineComment(_):
+              if lineCommentStart == nil {
+                lineCommentStart = start
+              }
+            case .newlines(1), .carriageReturns(1), .spaces(_), .tabs(_):
+              if lineCommentStart != nil {
```
ahoppen, line 1046: IIUC this means that line comments separated by multiple newlines will be considered as one folding range. E.g. the following will be considered as one folding range instead of three, right? (I think these should be three separate folding ranges).

```swift
// One line
// Followed by a line

// Another line
// Followed by another line


// An another line
func foo() {}
```

---

Also, does this mean that we are reporting folding ranges for single line comments like the following? (I think we should not).

```swift
// Single line comment
func foo() {}
```

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -978,143 +963,202 @@ extension SwiftLanguageServer {
...
+            return .visitChildren
+          }
+
+          guard let start: Position = snapshot.positionOf(utf8Offset: start),
+                let end: Position = snapshot.positionOf(utf8Offset: end) else {
+            log("folding range failed to retrieve position of \(snapshot.document.uri): \(start)-\(end)", level: .warning)
+            return .visitChildren
+          }
+          let range: FoldingRange
+          // If the client only supports folding full lines, ignore the end character's line.
```
ahoppen, line 1117: You probably meant the following, right?
```suggestion
          // If the client only supports folding full lines, ignore the end character's correct.
```

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -978,143 +963,202 @@ extension SwiftLanguageServer {
...
+          return self.addFoldingRange(
+            start: node.argumentList.position.utf8Offset,
+            end: node.argumentList.endPosition.utf8Offset)
+        }
+
+        __consuming func finalize() -> Set<FoldingRange> {
+          return self.ranges
+        }
+
+        private func addFoldingRange(start: Int, end: Int, kind: FoldingRangeKind? = nil) -> SyntaxVisitorContinueKind {
```
ahoppen, line 1106: Is this returning a `SyntaxVisitorContinueKind` so you can `return self.addFoldingRange` in the `visit` methods? To me it implied that this method had some smart logic whether it should visit children or not and then I was surprised that it always returned `visitChildren`.

I think it would be good if you could either
- Add a doc comment like: ``This method always returns `.visitChildren` so `visit` method implementations can be `return self.addFoldingRange(…)` `` or
- Have this method return `Void`

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
CodaFi: There are two newlines here

```
// One line
// Followed by a line

// Another line
// Followed by another line
```

and three here

```
// Another line
// Followed by another line


// An another line
```

These comments are therefore not merged.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -876,7 +861,7 @@ extension SwiftLanguageServer {
         return
       }
 
-      let tokens = snapshot.tokens.mergedAndSorted
+      let tokens = snapshot.tokens.semantic.sorted(by: { $0.start < $1.start })
```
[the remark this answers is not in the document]
CodaFi: How difficult would it be to redo the syntactic highlighting with the syntax tree? Specifically, it's not clear to me where SourceKit respects the "syntactic only" flag when generating the syntactic token stream in the first place.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
CodaFi: I've updated the doc comment to be clearer. This was the old comment FWIW.

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -153,18 +153,16 @@ public final class DocumentManager {
...
-                if let adjustedRange = rangeAdjuster.adjust(token.range) {
-                  token.range = adjustedRange
-                  return token
-                } else {
-                  return nil
-                }
-              })
-          }
+          document.latestTokens.semantic = Array(document.latestTokens.semantic.lazy
+            .compactMap {
```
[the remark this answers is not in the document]
CodaFi: It will not have a positive effect since we're immediately draining the results into a fresh array to store it here. I left this code alone when I migrated the lexical token stream, so I'll drop it in #651.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
CodaFi: > Also, does this mean that we are reporting folding ranges for single line comments like the following? 

Outside of line range mode, both SourceKit and this PR report that single-line comment as a folding range.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
CodaFi: This was always meant to return `.skipChildren` if the number of ranges exceeded the requested limit. Thanks.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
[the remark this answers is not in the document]
fwcd: > IIRC merging the syntactic and semantic highlighting provided from sourcekitd was essential to provide good semantic highlighting because sourcekitd’s semantic highlights assume that the syntactic highlights shine through.

This is true, although we also have some level of lexical/syntactic highlighting on the VSCode side with its Swift grammar already (this is IIRC the original motivation for calling the LSP feature 'semantic highlighting', so only the heavy lifting is delegated to the language server). Still, other LSP clients might not have a separate grammar, additionally the quality of sourcekitd's syntactic tokens is (in my experience) considerably better than VSCode's TextMate grammar, so I'd be in favor of keeping lexical/syntactic highlighting if possible.

> How difficult would it be to redo the syntactic highlighting with the syntax tree?

Doing the syntactic highlighting with swift-syntax would probably be quite doable if the plan is to maintain a syntax tree on the language server side anyway. Is the long-term plan to let sourcekitd only handle the semantic tokens?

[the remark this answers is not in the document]
ahoppen: @CodaFi and I discussed this offline yesterday and I think we should be able to use SwiftSyntax’s `SyntaxClassifier` for syntactic highlighting (in my memory that one was even superior to `sourcekitd`’s syntactic highlighting) and only use `sourcekitd` for semantic highlighting.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Do we have a test case that we are not merging these comments into one folding range?

CodaFi: Yes, these two lines assert what you're looking for https://github.com/apple/sourcekit-lsp/blob/main/Tests/SourceKitLSPTests/FoldingRangeTests.swift#L73-L74

The corresponding comments are

```
//
// MARK: - A mark! -
//

//
// FIXME: a fixme
//
```

ahoppen: Nice. Thank you 👍

ahoppen (approved):
Thank you, looks good to me and it’s great to see us get rid of another `sourcekitd` request.

PR #659, opened 2022-10-25
SKCore: use `PackageModel.BuildFlags` instead of TSC

MaxDesiatov:
As TSC would like to deprecate `BuildFlags` (https://github.com/apple/swift-tools-support-core/pull/359), let's use new `struct BuildFlags` from SwiftPM's `PackageModel`.


MaxDesiatov:
https://github.com/apple/swift-package-manager/pull/5837
@swift-ci please test

ahoppen (approved):

PR #660, opened 2022-10-26
Add dependency on IDEUtils

ahoppen:
Reflect for the fact that SyntaxClassifier has moved to a new `IDEUtils` module.


ahoppen:
@swift-ci Please test

PR #661, opened 2022-10-26
SourceKitLSP: add new dependency introduced in SwiftSyntax

compnerd:
Add a dependency on IDEUtils that is missing in the build.


compnerd:
Please test with following PRs:
apple/swift-syntax#1029

@swift-ci please test

PR #668, opened 2022-11-17
SourceKitLSP: generate Swift textual interfaces for module references

bwhiteley:
When doing a jump to definition on a Swift import statement, generate the textual interface for the module so the IDE can load the generated interface file.

### DIFF
--- a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
@@ -1,0 +1,39 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request a textual interface of a module to display in the IDE.
+/// **(LSP Extension)**
+public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
+  public static let method: String = "textDocument/openInterface"
+  public typealias Response = InterfaceDetails?
+
+  /// The document whose compiler arguments should be used to generate the interface.
+  public var textDocument: TextDocumentIdentifier
+
+  /// The module to generate an index for.
+  public var name: String
+
+  public init(textDocument: TextDocumentIdentifier, name: String) {
+    self.textDocument = textDocument
+    self.name = name
+  }
+}
+
+/// The textual output of a module interface.
+public struct InterfaceDetails: ResponseType, Hashable {
+
+  public var uri: DocumentURI
+
+  public init(uri: DocumentURI) {
+    self.uri = uri
+  }
+}
--- a/Sources/SourceKitD/SourceKitD.swift
+++ b/Sources/SourceKitD/SourceKitD.swift
@@ -112,6 +112,10 @@ extension SourceKitD {
 
     return handle
   }
+  
+  public func cancel(_ handle: sourcekitd_request_handle_t) {
+    api.cancel_request(handle)
+  }
 }
 
 private func logRequest(_ request: SKDRequestDictionary) {
--- a/Sources/SourceKitLSP/Swift/OpenInterface.swift
+++ b/Sources/SourceKitLSP/Swift/OpenInterface.swift
@@ -1,0 +1,84 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import SourceKitD
+import LanguageServerProtocol
+import LSPLogging
+
+struct InterfaceInfo {
+  var contents: String
+}
+
+extension SwiftLanguageServer {
+  public func openInterface(_ request: LanguageServerProtocol.Request<LanguageServerProtocol.OpenInterfaceRequest>) {
+    let uri = request.params.textDocument.uri
+    let moduleName = request.params.name
+    self.queue.async {
+      let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent("\(moduleName).swiftinterface")
+      let interfaceDocURI = DocumentURI(interfaceFilePath)
+      self._openInterface(request: request, uri: uri, name: moduleName, interfaceURI: interfaceDocURI) { result in
+        switch result {
+        case .success(let interfaceInfo):
+          do {
+            try interfaceInfo.contents.write(to: interfaceFilePath, atomically: true, encoding: String.Encoding.utf8)
+            request.reply(.success(InterfaceDetails(uri: interfaceDocURI)))
+          } catch {
+            request.reply(.failure(ResponseError.unknown(error.localizedDescription)))
+          }
+        case .failure(let error):
+          log("open interface failed: \(error)", level: .warning)
+          request.reply(.failure(ResponseError(error)))
+        }
+      }
+    }
+  }
+    
+  /// Open the Swift interface for a module.
+  ///
+  /// - Parameters:
+  ///   - request: The OpenInterfaceRequest.
+  ///   - uri: The document whose compiler arguments should be used to generate the interface.
+  ///   - name: The name of the module whose interface should be generated.
+  ///   - interfaceURI: The file where the generated interface should be written.
+  ///   - completion: Completion block to asynchronously receive the InterfaceInfo, or error.
+  private func _openInterface(request: LanguageServerProtocol.Request<LanguageServerProtocol.OpenInterfaceRequest>,
+                              uri: DocumentURI,
+                              name: String,
+                              interfaceURI: DocumentURI,
+                              completion: @escaping (Swift.Result<InterfaceInfo, SKDError>) -> Void) {
+    let keys = self.keys
+    let skreq = SKDRequestDictionary(sourcekitd: sourcekitd)
+    skreq[keys.request] = requests.editor_open_interface
+    skreq[keys.modulename] = name
+    skreq[keys.name] = interfaceURI.pseudoPath
+    skreq[keys.synthesizedextensions] = 1
+    if let compileCommand = self.commandsByFile[uri] {
+      skreq[keys.compilerargs] = compileCommand.compilerArgs
+    }
+    
+    let handle = self.sourcekitd.send(skreq, self.queue) { result in
+      switch result {
+      case .success(let dict):
+        return completion(.success(InterfaceInfo(contents: dict[keys.sourcetext] ?? "")))
+      case .failure(let error):
+        return completion(.failure(error))
+      }
+    }
+    
+    if let handle = handle {
+      request.cancellationToken.addCancellationHandler { [weak self] in
+        self?.sourcekitd.cancel(handle)
+      }
+    }
+  }
+}
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -10,8 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
 import Dispatch
-import struct Foundation.CharacterSet
 import LanguageServerProtocol
 import LSPLogging
 import SKCore
@@ -106,6 +106,9 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   let clientCapabilities: ClientCapabilities
 
   let serverOptions: SourceKitServer.Options
+  
+  /// Directory where generated Swift interfaces will be stored.
+  let generatedInterfacesPath: URL
 
   // FIXME: ideally we wouldn't need separate management from a parent server in the same process.
   var documentManager: DocumentManager
@@ -154,6 +157,8 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     self.documentManager = DocumentManager()
     self.state = .connected
     self.reopenDocuments = reopenDocuments
+    self.generatedInterfacesPath = options.generatedInterfacesPath.asURL
+    try FileManager.default.createDirectory(at: generatedInterfacesPath, withIntermediateDirectories: true)
   }
 
   public func canHandle(workspace: Workspace) -> Bool {
@@ -259,7 +264,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
       }
     })
   }
-
+  
   /// Publish diagnostics for the given `snapshot`. We withhold semantic diagnostics if we are using
   /// fallback arguments.
   ///
@@ -1576,6 +1581,8 @@ extension sourcekitd_uid_t {
       case vals.decl_extension:
         // There are no extensions in LSP, so I return something vaguely similar
         return .namespace
+      case vals.ref_module:
+        return .module
       default:
         return nil
     }
--- a/Sources/sourcekit-lsp/main.swift
+++ b/Sources/sourcekit-lsp/main.swift
@@ -130,6 +130,10 @@ struct Main: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    help: "Specify the directory where generated interfaces will be stored"
+  )
+  var generatedInterfacesPath = defaultDirectoryForGeneratedInterfaces
 
   @Option(
     help: "Whether to enable server-side filtering in code-completion"
@@ -156,6 +160,7 @@ struct Main: ParsableCommand {
     serverOptions.indexOptions.indexPrefixMappings = indexPrefixMappings
     serverOptions.completionOptions.serverSideFiltering = completionServerSideFiltering
     serverOptions.completionOptions.maxResults = completionMaxResults
+    serverOptions.generatedInterfacesPath = generatedInterfacesPath
 
     return serverOptions
   }
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -1,0 +1,132 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+import LSPTestSupport
+import SKTestSupport
+import SourceKitLSP
+import XCTest
+
+final class SwiftInterfaceTests: XCTestCase {
+
+  /// Connection and lifetime management for the service.
+  var connection: TestSourceKitServer! = nil
+
+  /// The primary interface to make requests to the SourceKitServer.
+  var sk: TestClient! = nil
+
+  var documentManager: DocumentManager! {
+    connection.server!._documentManager
+  }
+
+  override func setUp() {
+    connection = TestSourceKitServer()
+    sk = connection.client
+    _ = try! sk.sendSync(InitializeRequest(
+      processId: nil,
+      rootPath: nil,
+      rootURI: nil,
+      initializationOptions: nil,
+      capabilities: ClientCapabilities(workspace: nil,
+                                       textDocument: TextDocumentClientCapabilities(
+                                        codeAction: .init(
+                                          codeActionLiteralSupport: .init(
+                                            codeActionKind: .init(valueSet: [.quickFix])
+                                          )),
+                                        publishDiagnostics: .init(codeDescriptionSupport: true)
+                                       )),
+      trace: .off,
+      workspaceFolders: nil))
+  }
+  
+  override func tearDown() {
+    sk = nil
+    connection = nil
+  }
+  
+  func testSystemModuleInterface() throws {
+    let url = URL(fileURLWithPath: "/\(UUID())/a.swift")
+    let uri = DocumentURI(url)
+
+    sk.send(DidOpenTextDocumentNotification(textDocument: TextDocumentItem(
+      uri: uri,
+      language: .swift,
+      version: 1,
+      text: """
+      import Foundation
+      """)))
+    
+    let _resp = try sk.sendSync(DefinitionRequest(
+      textDocument: TextDocumentIdentifier(url),
+      position: Position(line: 0, utf16index: 10)))
+    let resp = try XCTUnwrap(_resp)
+    guard case .locations(let locations) = resp else {
+      XCTFail("Unexpected response: \(resp)")
+      return
+    }
+    XCTAssertEqual(locations.count, 1)
+    let location = try XCTUnwrap(locations.first)
+    XCTAssertTrue(location.uri.pseudoPath.hasSuffix("/Foundation.swiftinterface"))
+    let fileContents = try XCTUnwrap(location.uri.fileURL.flatMap({ try String(contentsOf: $0, encoding: .utf8) }))
+    // Sanity-check that the generated Swift Interface contains Swift code
+    XCTAssertTrue(fileContents.hasPrefix("import "))
+  }
+  
+  func testOpenInterface() throws {
+    guard let ws = try staticSourceKitSwiftPMWorkspace(name: "SwiftPMPackage") else { return }
+    try ws.buildAndIndex()
+    let importedModule = ws.testLoc("lib:import")
+    try ws.openDocument(importedModule.url, language: .swift)
+    let openInterface = OpenInterfaceRequest(textDocument: importedModule.docIdentifier, name: "lib")
+    let interfaceDetails = try XCTUnwrap(ws.sk.sendSync(openInterface))
+    XCTAssertTrue(interfaceDetails.uri.pseudoPath.hasSuffix("/lib.swiftinterface"))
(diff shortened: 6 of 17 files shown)


adam-fowler:
@ahoppen do you think this will work with my suggested path to displaying symbols in VSCode. https://github.com/swift-server/vscode-swift/issues/408

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request a textual interface of a module to display in the IDE.
```
ahoppen, line 13: Since this is a non-standard LSP request, could you annotate it with
```swift
/// **(LSP Extension)***
```

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+/// The textual output of a module interface.
+public struct InterfaceDetails: ResponseType, Hashable {
+
+  public var uri: DocumentURI
+
+  public init(uri: DocumentURI)
+  {
```
ahoppen, line 36: Nitpick to match the style in the rest of the project

```suggestion
  public init(uri: DocumentURI) {
```

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+private var textualInterfaces: [String:DocumentURI] = [:]
+
+extension SwiftLanguageServer {
+  public func openInterface(_ request: LanguageServerProtocol.Request<LanguageServerProtocol.OpenInterfaceRequest>) {
+    let uri = request.params.textDocument.uri
+    let moduleName = request.params.name
+    self.queue.async {
+
```
ahoppen, line 29: Superfluous newline

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,85 @@
...
+import Foundation
+import SourceKitD
+import LanguageServerProtocol
+import LSPLogging
+
+struct InterfaceInfo {
+  var contents: String
+}
+
+private var textualInterfaces: [String:DocumentURI] = [:]
```
ahoppen, line 22:
```suggestion
private var textualInterfaces: [String: DocumentURI] = [:]
```

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,85 @@
...
+
+extension SwiftLanguageServer {
+  public func openInterface(_ request: LanguageServerProtocol.Request<LanguageServerProtocol.OpenInterfaceRequest>) {
+    let uri = request.params.textDocument.uri
+    let moduleName = request.params.name
+    self.queue.async {
+      
+      if let uri = textualInterfaces[moduleName] {
```
ahoppen, line 30: I don’t think we can cache textual interfaces like this. The contents of the module might have changed since it was last requested and thus the cache would be out of date. 

I think caching would be safe for system modules that are part of the SDK and are thus very, very unlikely to change (unless you re-install Xcode during a session). IIRC there should be a bit in either the symbol or the module response that says whether the module is a system module.

Also the compiler arguments would need to be part of the key because the generated interface request can return different results for different compiler arguments (think about the generated interface from an Objective-C module that uses `#if FOO` and in one invocation you pass `-DFOO` in the compiler arguments and in the other you don’t.

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -112,6 +112,10 @@ extension SourceKitD {
 
     return handle
   }
+  
+  public func cancel(_ handle: sourcekitd_request_handle_t) {
+    api.cancel_request(handle)
```
ahoppen, line 117: 🎉

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -0,0 +1,39 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request a textual interface of a module to display in the IDE.
+public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
+  public static let method: String = "textDocument/openInterface"
+  public typealias Response = InterfaceDetails
+
+  /// The document in which to lookup the module.
```
ahoppen, line 18: I think a better description here would be.

```suggestion
  /// The document whose compiler arguments should be used to generate the interface.
```

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+      }
+    }
+  }
+  
+  private func _openInterface(_ request: LanguageServerProtocol.Request<LanguageServerProtocol.OpenInterfaceRequest>,
+                              _ uri: DocumentURI,
+                              _ name: String,
+                              _ interfaceURI: DocumentURI,
+                              _ completion: @escaping (Swift.Result<InterfaceInfo, SKDError>) -> Void) {
```
ahoppen, line 59: Could you add descriptions for these parameters? Also: Is there any particular reason why you chose to not use argument labels for any of these? I personally always prefer argument label.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -106,6 +106,8 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   let clientCapabilities: ClientCapabilities
 
   let serverOptions: SourceKitServer.Options
+  
+  let generatedInterfacesPath: URL
```
ahoppen, line 110: Could you add a doc comment for the property?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -259,6 +262,20 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
       }
     })
   }
+  
+  /// Obtain the path of the directory to write generated module interfaces.
+  private static func generatedInterfacesPath() throws -> URL {
+    if let path = ProcessInfo.processInfo.environment["SOURCEKIT_GENERATED_INTERFACES_PATH"] {
```
ahoppen, line 268: What do you think about making this an argument to the `sourcekit-lsp` executable instead of an opaque environment variable?

ahoppen (commented):
Thanks a lot for the PR. I think the idea of generating the `.swiftinterfaces` into a temporary directory is a good solution for now. @adam-fowler I have a vague memory that we discussed that LSP/VSCode has/will have a notion of generated/transient/non-disk files but I can’t find it anymore. Do you remember anything about it?

The implementation looks good to me overall, I have some comments inline. Also: Could you add a test case for jump to definition to symbols in other modules and a test case for a standalone open interface request?

adam-fowler:
> @adam-fowler I have a vague memory that we discussed that LSP/VSCode has/will have a notion of generated/transient/non-disk files but I can’t find it anymore. Do you remember anything about it?

That'll be in the issue I linked to https://github.com/swift-server/vscode-swift/issues/408#issuecomment-1277815702

ahoppen:
Is there any precedence from language servers that use these kinds of transient documents or an LSP proposal? I think you are the best person to judge how the contents of the interface should be transferred to the editor. The nice thing about the temporary file implementation is that it will work for all editors.

adam-fowler:
> Is there any precedence from language servers that use these kinds of transient documents or an LSP proposal? I think you are the best person to judge how the contents of the interface should be transferred to the editor. The nice thing about the temporary file implementation is that it will work for all editors.

We can go with the temporary file solution. You'll need to be careful to ensure you generate separate versions of these for different swift versions.

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,85 @@
...
+    let uri = request.params.textDocument.uri
+    let moduleName = request.params.name
+    self.queue.async {
+      
+      if let uri = textualInterfaces[moduleName] {
+        request.reply(.success(InterfaceDetails(uri: uri)))
+        return
+      }
+      
+      let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent("\(moduleName).swift")
```
bwhiteley, line 35: @ahoppen Should these generated interface files have a `.swiftinterface` extension instead of `.swift`?

ahoppen: Oh, yes I think they should.

Sources/SourceKitLSP/Swift/OpenInterface.swift
bwhiteley: I removed caching for now. We could reintroduce it later while addressing the issues you raised.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
bwhiteley: I'll make that change.

bwhiteley:
Thanks for the feedback. I've addressed the issues with the exception wanting to add a couple of more tests. 

>  add a test case for jump to definition to symbols in other modules

Is there an existing test that is similar to this that you could point me to? I've found a couple of candidates, but one of the challenges I am facing is that some of the tests fail for me even in upstream/main. 
For example:
image

Sources/SourceKitLSP/Swift/OpenInterface.swift
ahoppen: OK, I think even without caching this is a very good PR.

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -0,0 +1,84 @@
...
+  }
+    
+  /// Open the Swift interface for a module.
+  ///
+  /// - Parameters:
+  ///   - request: The OpenInterfaceRequest.
+  ///   - uri: The document currently being edited.
+  ///   - name: The module name.
+  ///   - interfaceURI:The file where the generated interface should be written.
+  ///   - completion: Completion block to asynchronously receive the InterfaceInfo, or error.
```
ahoppen, line 53: I’ve got a couple of minor improvement suggestions for this comment.

```suggestion
  /// Open the Swift interface for a module.
  ///
  /// - Parameters:
  ///   - request: The OpenInterfaceRequest.
  ///   - uri: The document whose compiler arguments should be used to generate the interface.
  ///   - name: The name of the module whose interface should be generated.
  ///   - interfaceURI: The file where the generated interface should be written.
  ///   - completion: Completion block to asynchronously receive the InterfaceInfo, or error.
```

Sources/sourcekit-lsp/main.swift
```diff
@@ -130,6 +130,11 @@ struct Main: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    name: .customLong("generated-interfaces-path", withSingleDash: true),
```
ahoppen, line 134: I would prefer this to have two dashes, like all the options below. `index-store-path` and friends only have a single dash to be compatible with `clang` compiler arguments. And if we do that, I think it’s also not necessary to specify a `name` for the `Option`.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -259,6 +263,19 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
...
+    if let overridePath = overrideInterfacesPath {
+      try FileManager.default.createDirectory(at: overridePath, withIntermediateDirectories: true)
+      return overridePath
+    } else {
+      let tempFolderURL = URL(fileURLWithPath: NSTemporaryDirectory())
+      let genFolder = tempFolderURL.appendingPathComponent("GeneratedInterfaces", isDirectory: true)
+      try FileManager.default.createDirectory(at: genFolder, withIntermediateDirectories: true)
+      return genFolder
+    }
+  }
```
ahoppen, line 278: What do you think about defaulting the generated interface path to a temporary directory in `sourcekit-lsp/main.swift`. If you do that, the documentation of the `--generated-interfaces-path` could mention that it defaults to a temporary directory, which makes the behavior cleaner IMO.

ahoppen:
That’s odd that the tests are failing for you. They are passing fine for me locally. Which Xcode version are you using and are you using a Swift Open Source Toolchain? I’m using Xcode 14.0 with the default toolchain.

I think you can use Tests/SourceKitLSPTests/ImplementationTests.swift as a blueprint for your tests. I don’t think you need to check that many locations but the `impls(at:)` function should give you a good idea of how to invoke an LSP request in the tests.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -0,0 +1,154 @@
...
+    let resp = try XCTUnwrap(_resp)
+    guard case .locations(let locations) = resp else {
+      XCTFail("Unexpected response: \(resp)")
+      return
+    }
+    XCTAssertEqual(locations.count, 1)
+    let location = try XCTUnwrap(locations.first)
+    XCTAssertTrue(location.uri.pseudoPath.hasSuffix("/Foundation.swiftinterface"))
+    let fileContents = try XCTUnwrap(location.uri.fileURL.flatMap({ try String(contentsOf: $0, encoding: .utf8) }))
+    XCTAssertTrue(fileContents.hasPrefix("import "))
```
ahoppen, line 81: Is this testing that `Foundation.swiftinterface` is not empty? I think that’s perfectly fine but a short comment like 
```
// Sanity-check that the generated Swift Interface contains Swift code
```
would explain this, I think.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -0,0 +1,154 @@
...
+  func testOpenInterface() throws {
+    guard let ws = try staticSourceKitSwiftPMWorkspace(name: "SwiftPMPackage") else { return }
+    try ws.buildAndIndex()
+    let call = ws.testLoc("Lib.foo:call")
+    try ws.openDocument(call.url, language: .swift)
+    let workspace = try XCTUnwrap(ws.testServer.server?.queue.sync {
+      try XCTUnwrap(ws.testServer.server?.workspaceForDocument(uri: call.docUri))
+    })
+    let swiftLangServer = try XCTUnwrap(ws.testServer.server?._languageService(for: call.docUri,
+                                                                               .swift,
```
ahoppen, line 93: This is being rendered really weirdly on GitHub (the `.` is inside the `s` of `swift`). Is this a normal period or some weird Unicode character?

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -0,0 +1,154 @@
...
+        XCTFail(error.localizedDescription)
+      }
+      expectation.fulfill()
+    })
+    
+    // Send an arbitrary request through the front door first or SourceKitServer won't be properly initialized. 
+    _ = try ws.sk.sendSync(HoverRequest(
+      textDocument: call.docIdentifier,
+      position: Position(line: 0, utf16index: 8)))
+    swiftLangServer.openInterface(request)
```
ahoppen, line 121: Is there any reason why this can’t be `ws.sk.sendSync(openInterface)`. If you send the request that way, I think you also don’t need to send the hover request.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -0,0 +1,154 @@
...
+        position: Position(line: 0, utf16index: 8)))
+    }
+    let resp = try XCTUnwrap(_resp)
+    guard case .locations(let locations) = resp else {
+      XCTFail("Unexpected response: \(resp)")
+      return
+    }
+    XCTAssertEqual(locations.count, 1)
+    let location = try XCTUnwrap(locations.first)
+    XCTAssertTrue(location.uri.pseudoPath.hasSuffix("/lib.swiftinterface"))
```
ahoppen, line 143: Shouldn’t we be jumping to the original source file if it’s in the same package instead of to the Swift interface. AFAICT that’s what we are doing right now. Jumping to the Swift interface here seems like a regression to me.

ahoppen (commented):
Thanks for adding the test cases. I’ve added a few questions inline.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
bwhiteley: It jumps to the Swift interface only if the symbol is a module. Note the `position` in the request is on an import statement. Otherwise it goes to the appropriate line in the source file.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
bwhiteley: If I do it that way it hits this `fatalError`
https://github.com/apple/sourcekit-lsp/blob/main/Sources/LSPTestSupport/TestJSONRPCConnection.swift#L145

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
ahoppen: Hmm, weird. Must have been a rendering issue in Safari then.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
ahoppen: Oh, I didn’t read carefully enough and was assuming that you are jumping to definition on the call. Instead of having the slightly fragile hardcoded position in this test case, the intended design would be to modify `SwiftPMPackage/Sources/exec/main.swift` to

```swift
import /*lib:import*/lib

Lib() . /*Lib.foo:call*/foo()
```

and then you can use define
```diff
- let call = ws.testLoc("Lib.foo:call")
+ let importedModule = ws.testLoc("lib:import")
```
and use `importedModule.position` instead of `Position(line: 0, utf16index: 8)`.

Same for the other test cases in this file.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
ahoppen: Oh, that’s a very important catch that would have prevented the open interface request from working in the live `sourcekit-lsp` server as well: You need to add `OpenInterfaceRequest.self` to `builtinRequests` in `Messages.swift`.

bwhiteley: I added `OpenInterfaceRequest.self` to `builtinRequests` in `Messages.swift` but still get the same crash.

ahoppen: I just did a little bit more debugging myself and I think these are the changes you need. Essentially, what needs to be done is that SourceKit-LSP has to know about the request type and which method implements the request handling (`registerToolchainTextDocumentRequest`). That also needs to have a fallback response, that's why I changed the return type to be optional.

```diff
diff --git a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
index db33f31..6292c1d 100644
--- a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
@@ -14,7 +14,7 @@
 /// **(LSP Extension)**
 public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
   public static let method: String = "textDocument/openInterface"
-  public typealias Response = InterfaceDetails
+  public typealias Response = InterfaceDetails?
 
   /// The document whose compiler arguments should be used to generate the interface.
   public var textDocument: TextDocumentIdentifier
diff --git a/Sources/SourceKitLSP/SourceKitServer.swift b/Sources/SourceKitLSP/SourceKitServer.swift
index 28f93a7..ef07a3e 100644
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -184,6 +184,7 @@ public final class SourceKitServer: LanguageServer {
     registerToolchainTextDocumentRequest(SourceKitServer.completion,
                                          CompletionList(isIncomplete: false, items: []))
     registerToolchainTextDocumentRequest(SourceKitServer.hover, nil)
+    registerToolchainTextDocumentRequest(SourceKitServer.openInterface, nil)
     registerToolchainTextDocumentRequest(SourceKitServer.declaration, .locations([]))
     registerToolchainTextDocumentRequest(SourceKitServer.definition, .locations([]))
     registerToolchainTextDocumentRequest(SourceKitServer.references, [])
@@ -986,6 +987,14 @@ extension SourceKitServer {
     languageService.hover(req)
   }
 
+  func openInterface(
+    _ req: Request<OpenInterfaceRequest>,
+    workspace: Workspace,
+    languageService: ToolchainLanguageServer
+  ) {
+    languageService.openInterface(req)
+  }
+
   /// Find all symbols in the workspace that include a string in their name.
   /// - returns: An array of SymbolOccurrences that match the string.
   func findWorkspaceSymbols(matching: String) -> [SymbolOccurrence] {
@@ -1279,9 +1288,11 @@ extension SourceKitServer {
         let request = Request(openInterface, id: req.id, clientID: ObjectIdentifier(self),
                               cancellation: req.cancellationToken, reply: { (result: Result<OpenInterfaceRequest.Response, ResponseError>) in
           switch result {
-          case .success(let interfaceDetails):
+          case .success(let interfaceDetails?):
             let loc = Location(uri: interfaceDetails.uri, range: Range(Position(line: 0, utf16index: 0)))
             req.reply(.locations([loc]))
+          case .success(nil):
+            req.reply(.failure(.unknown("Could not generate Swift Interface for \(name)")))
           case .failure(let error):
             req.reply(.failure(error))
           }
diff --git a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
index df39e33..44da807 100644
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -91,35 +91,19 @@ final class SwiftInterfaceTests: XCTestCase {
       try XCTUnwrap(ws.testServer.server?.workspaceForDocument(uri: importedModule.docUri))
     })
     let swiftLangServer = try XCTUnwrap(ws.testServer.server?._languageService(for: importedModule.docUri, .swift, in: workspace))
-    let expectation = expectation(description: "open interface request")
     let openInterface = OpenInterfaceRequest(textDocument: importedModule.docIdentifier, name: "lib")
-    let request = Request(openInterface, id: .number(1), clientID: ObjectIdentifier(swiftLangServer),
-                          cancellation: CancellationToken(), reply: { (result: Result<OpenInterfaceRequest.Response, ResponseError>) in
-      do {
-        let interfaceDetails = try result.get()
-        XCTAssertTrue(interfaceDetails.uri.pseudoPath.hasSuffix("/lib.swiftinterface"))
-        let fileContents = try XCTUnwrap(interfaceDetails.uri.fileURL.flatMap({ try String(contentsOf: $0, encoding: .utf8) }))
-        XCTAssertTrue(fileContents.contains("""
-          public struct Lib {
-          
-              public func foo()
-          
-              public init()
-          }
-          """))
-      } catch {
-        XCTFail(error.localizedDescription)
-      }
-      expectation.fulfill()
-    })
-    
-    // Send an arbitrary request through the front door first or SourceKitServer won't be properly initialized.
-    _ = try ws.sk.sendSync(HoverRequest(
-      textDocument: importedModule.docIdentifier,
-      position: importedModule.position))
-    swiftLangServer.openInterface(request)
     
-    waitForExpectations(timeout: 15)
+    let interfaceDetails = try XCTUnwrap(ws.sk.sendSync(openInterface))
+    XCTAssertTrue(interfaceDetails.uri.pseudoPath.hasSuffix("/lib.swiftinterface"))
+    let fileContents = try XCTUnwrap(interfaceDetails.uri.fileURL.flatMap({ try String(contentsOf: $0, encoding: .utf8) }))
+    XCTAssertTrue(fileContents.contains("""
+      public struct Lib {
+
+          public func foo()
+
+          public init()
+      }
+      """))
   }
 
   func testSwiftInterfaceAcrossModules() throws {
```

bwhiteley: Thanks for looking into that! I've made the recommended changes.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -0,0 +1,132 @@
...
+      uri: uri,
+      language: .swift,
+      version: 1,
+      text: """
+      import Foundation
+      """)))
+    
+    let _resp = try sk.sendSync(DefinitionRequest(
+      textDocument: TextDocumentIdentifier(url),
+      position: Position(line: 0, utf16index: 10)))
```
ahoppen, line 71: Do you think you could also define a `testLoc` for this instead of hardcoding the position in the tests?

ahoppen (commented):
Thanks. This looks good to me. Let's get this merged. I've got one final comment inline. Could you also squash your commits after you've addressed the comment to create a nicer git history?

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
bwhiteley: Unlike the other tests in that file, this particular test doesn't have a TestWorkspace, so `testLoc` is not readily available from what I can see. Rather, the code is provided via `DidOpenTextDocumentNotification`. Similarly, `LocalSwiftTests.swift` is full of `Positions` with hard-coded offsets.

ahoppen: Oh, sorry. I should have read the test case more carefully.

ahoppen:
@swift-ci Please test

artemcm:
Not sure how yet, but I think this is related to a failure in swift-driver Windows CI:
https://github.com/apple/swift-driver/pull/1240
https://ci-external.swift.org/job/swift-driver-PR-windows/227/console 
```
C:\Users\swift-ci\jenkins\workspace\swift-driver-PR-windows\sourcekit-lsp\Sources\LanguageServerProtocol\Messages.swift:59:3: error: cannot find 'OpenInterfaceRequest' in scope
```

PR #682, opened 2022-12-20
Allow folding range response from clangd

bnbarham:
clangd supports folding range in stable/20221013. Update the test to allow an actual response.


bnbarham:
@swift-ci please test

PR #686, opened 2023-01-02
SourceKit: sever dependency on `TSCUtility`

compnerd:
The dependency on `TSCUtility` was strictly for identifying the platform of execution. This logic is relatively self-contained and effectively an extension over an enumeration. Replicate this logic with updates for new syntactic improvements. This allows us to partially reduce dependency on swift-tools-support-core. The dependency on TSCBasic is more complicated to remove due to the extensive use of `AbsolutePath`.


compnerd:
@swift-ci please test

Tests/SourceKitDTests/CrashRecoveryTests.swift
```diff
@@ -42,7 +43,7 @@ fileprivate extension HoverResponse {
 
 final class CrashRecoveryTests: XCTestCase {
   func testSourcekitdCrashRecovery() throws {
-    try XCTSkipUnless(isDarwinHost, "Linux and Windows use in-process sourcekitd")
+    try XCTSkipUnless(Platform.current == .some(.darwin), "Linux and Windows use in-process sourcekitd")
```
ahoppen, line 46: Nit: Wouldn’t this be nicer as
```suggestion
    try XCTSkipUnless(Platform.current == Platform.darwin, "Linux and Windows use in-process sourcekitd")
```

Or I think `.darwin` might even work as well, but I’m not sure.

Tests/SourceKitDTests/SourceKitDTests.swift
```diff
@@ -13,25 +13,23 @@
...
 import SKCore
+import SKSupport
 import TSCBasic
 import ISDBTibs
 import ISDBTestSupport
+import Foundation
 import XCTest
 
-import enum TSCUtility.Platform
+import class TSCBasic.Process
```
ahoppen, line 23: How come we need to import `TSCBasic.Process` where we didn’t need to import it before?

Sources/SKSupport/Platform.swift
```diff
@@ -10,10 +10,74 @@
...
+  public static var current: Platform? = {
+    #if os(Windows)
+    return .windows
+    #else
+    switch try? Process.checkNonZeroExit(args: "uname")
+        .trimmingCharacters(in: .whitespacesAndNewlines)
+        .lowercased() {
+    case .some("darwin"):
+      return .darwin
+    case .some("linux"):
```
ahoppen, line 68: Nit: I think this reads better as

```suggestion
    case "darwin"?:
      return .darwin
    case "linux"?:
```

Sources/SKSupport/Platform.swift
```diff
@@ -10,10 +10,74 @@
...
+  }
+
+  return (try? localFileSystem.readFileContents(AbsolutePath(validating: "/etc/system-release")))?.cString.hasPrefix("Amazon Linux release 2") ?? false
+}
+
+public enum Platform: Equatable {
+  public enum LinuxDistribution: Equatable {
+    case debian
+    case fedora
+  }
```
ahoppen, line 48: I don’t think we are differentiating between Debian and Fedora anywhere in SourceKit-LSP (and I don’t see us needing the differentiation anytime soon), so maybe we should just remove the distinction, especially since the Debian vs Fedora check seems to be the most complicated.

ahoppen (commented):
It’s a shame we need to duplicate the logic but if we want to get rid of the TSC it’s something we just have to do 😢

compnerd:
Yes, except, it seems that this functionality is used only in LSP.  The other piece of the interface is in SPM only.  As a result we can split it up this way and deprecate it in TSC.

Tests/SourceKitDTests/CrashRecoveryTests.swift
compnerd: Ah, right, I think that `.darwin` might work, and that is better.

Tests/SourceKitDTests/SourceKitDTests.swift
compnerd: Because `TSCUtility` dragged it along for us.  Something something `@_implementationOnly` all the things and explicit imports.

Sources/SKSupport/Platform.swift
compnerd: I have to agree, thanks!

Sources/SKSupport/Platform.swift
compnerd: I ... uh, yeah, I should've done that.  I don't know why I didn't.  Sorry.

compnerd:
@swift-ci please test

compnerd:
https://github.com/apple/swift-package-manager/pull/6011
https://github.com/apple/swift-tools-support-core/pull/378

Are the other parts of this change that together deprecate `TSCUtility.Platform`.  It really is unfortunate that the symbol has to be public as it is used across multiple modules.

ahoppen (approved):

compnerd:

PR #693, opened 2023-01-13
Remove workaround for SR-16097 to decode Shutdown request

ahoppen:
rdar://92254952


ahoppen:
@swift-ci Please test

PR #698, opened 2023-01-24
Add failure message to assert in testSystemModuleInterface

ahoppen:
This helps us debug if the test fails in CI.


ahoppen:
@swift-ci Please test

PR #699, opened 2023-01-24
NFC: add `.vscode` to `.gitignore`

MaxDesiatov:
When running tests with VS Code, it creates additional `launch.json` which there's little sense to share between repository clones. Let's prevent it from being committed by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Package.resolved
 /*.sublime-workspace
 /.swiftpm
 .*.sw?
+.vscode


MaxDesiatov:
@swift-ci test

compnerd:
I'd rather not see this happen - `.vscode/` does contain stuff that we do want to enable e.g. `extensions.json` for extension recommendations.  I can see the desire to exclude `launch.json`, but having a workspace centric `launch.json` is still useful, users can always add a user specific `launch.json` IIRC.

MaxDesiatov:
This makes sense, I only based this on my past experience that most maintainers prefer these files to be excluded from commits.

@benlangmuir @ahoppen would you be interested in having this VS Code-specific workspace configuration committed to SourceKit-LSP repository instead of ignored like in this PR?

benlangmuir:
Do we have any concrete .vscode/ files we want to share ready to commit?  If not, I'm inclined to take this gitignore change.  We can always change gitignore again later.  But I'll defer to @ahoppen if he feels differently.

compnerd:
I think that we should add a `.vscode/extensions.json` with a pointer to `sswg.swift-lang`.

MaxDesiatov:
While running tests with `sswg.swift-lang` it created this `.vscode/launch.json`:

```json
{
    "configurations": [
        {
            "type": "lldb",
            "request": "launch",
            "name": "Debug sourcekit-lsp",
            "program": "${workspaceFolder:sourcekit-lsp}/.build/debug/sourcekit-lsp",
            "args": [],
            "cwd": "${workspaceFolder:sourcekit-lsp}",
            "preLaunchTask": "swift: Build Debug sourcekit-lsp"
        },
        {
            "type": "lldb",
            "request": "launch",
            "name": "Release sourcekit-lsp",
            "program": "${workspaceFolder:sourcekit-lsp}/.build/release/sourcekit-lsp",
            "args": [],
            "cwd": "${workspaceFolder:sourcekit-lsp}",
            "preLaunchTask": "swift: Build Release sourcekit-lsp"
        }
    ]
}
```

It doesn't contain any absolute paths, but I assume any user would have it recreated immediately after they start using the extension. OTOH, if someone starts adding more launch tasks to their configuration, they could commit those changes by accident I reckon?

compnerd:
The workspace wide configuration should be portable, and allows us to add any recommended flags, which is convenient for getting started.

MaxDesiatov:
Closing in favor of #700.

PR #704, opened 2023-01-31
Move `enum Platform` to SwiftPM

MaxDesiatov:
SwiftPM uses `Platform` type from TSC, which was recently deprecated after moving it to SourceKit-LSP. Let's share it with SwiftPM instead of having copies around.


MaxDesiatov:
https://github.com/apple/swift-package-manager/pull/6087

@swift-ci test

compnerd:
I'm not sure if this is really ideal - this type is simpler to deal with what LSP needs.  Does it make sense for SPM to vend this?  Perhaps we should vend this from swift-driver?

MaxDesiatov:
We'd like SourceKit-LSP to play better with cross-compilation destinations in the near future, which are currently handled by SwiftPM. From that perspective, sharing `Platform` would be beneficial. I don't mind moving it to swift-driver, but then do we want SourceKit-LSP to gain an explicit dependency on swift-driver?

ahoppen (approved):
Thanks. Sharing more code is always good 👍

PR #705, opened 2023-01-31
Add extended logging to SwiftIntefaceTests.testSystemModuleInterface

ahoppen:
This enables lower-level logging for a single test case, which will hopefully help us figure out why that test is sometimes failing in CI.

rdar://104871745 tracks that we should remove this extended logging again once it has fulfilled its purpose.


ahoppen:
@swift-ci Please test

PR #711, opened 2023-02-09
Fix CMake build not handling `@main` attribute

MaxDesiatov:
Looks like `@main` atttribute is not automatically recognized with CMake in executables that have a single source file declared.


MaxDesiatov:
@swift-ci test

ahoppen (approved):
If you say so, I’m fine with this change. CC @compnerd

compnerd:
I forgot about SR-12683.  An alternative is to pass `-parse-as-library`.

MaxDesiatov:
Passing `-parse-as-library` for an executable feels very wrong to me. I personally prefer to rename it back to `main.swift`.

ahoppen:
> Passing `-parse-as-library` for an executable feels very wrong to me. I personally prefer to rename it back to `main.swift`.

I agree

MaxDesiatov:
Closing in favor of https://github.com/apple/sourcekit-lsp/pull/712

PR #717, opened 2023-02-23
Check the target triple to determine what build flags to apply when building sourcekit-lsp

buttaface:
~This is needed for apple/swift#63782, which changes the Unix toolchain to look for libraries in architecture-specific directories.~

I ~also~ updated some checks for the runtime target triple, rather than checking the host system, and removed unnecessary regex checking.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -4,7 +4,6 @@ import argparse
 import json
 import os
 import platform
-import re
 import shutil
 import subprocess
 import sys
@@ -68,13 +67,16 @@ def swiftpm_bin_path(swift_exec: str, swiftpm_args: List[str], additional_env: D
     return check_output(cmd, additional_env=additional_env, capture_stderr=False, verbose=verbose).strip()
 
 
-def get_build_target(swift_exec: str, args: argparse.Namespace) -> str:
+def get_build_target(swift_exec: str, args: argparse.Namespace, cross_compile: bool = False) -> str:
     """Returns the target-triple of the current machine or for cross-compilation."""
     try:
         command = [swift_exec, '-print-target-info']
+        if cross_compile:
+            cross_compile_json = json.load(open(args.cross_compile_config))
+            command += ['-target', cross_compile_json["target"]]
         target_info_json = subprocess.check_output(command, stderr=subprocess.PIPE, universal_newlines=True).strip()
         args.target_info = json.loads(target_info_json)
-        if platform.system() == 'Darwin':
+        if '-apple-macosx' in args.target_info["target"]["unversionedTriple"]:
             return args.target_info["target"]["unversionedTriple"]
         return args.target_info["target"]["triple"]
     except Exception as e:
@@ -105,7 +107,9 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
         for san in args.sanitize:
             swiftpm_args += ['--sanitize=%s' % san]
 
-    if platform.system() == 'Darwin':
+    build_target = get_build_target(swift_exec, args, cross_compile=(True if args.cross_compile_config else False))
+    build_os = build_target.split('-')[2]
+    if build_os.startswith('macosx'):
         swiftpm_args += [
             '-Xlinker', '-rpath', '-Xlinker', '/usr/lib/swift',
             '-Xlinker', '-rpath', '-Xlinker', '@executable_path/../lib/swift/macosx',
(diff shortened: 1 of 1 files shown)


Utilities/build-script-helper.py
```diff
@@ -68,13 +68,16 @@ def swiftpm_bin_path(swift_exec: str, swiftpm_args: List[str], additional_env: D
     return check_output(cmd, additional_env=additional_env, capture_stderr=False, verbose=verbose).strip()
 
 
-def get_build_target(swift_exec: str, args: argparse.Namespace) -> str:
+def get_build_target(swift_exec: str, args: argparse.Namespace, cross_compile: bool = False) -> str:
```
ahoppen, line 71: Since we are currently passing all arguments via the `args` parameter, could we also avoid passing `cross_comile` and instead check whether `args.corss_compile_config` is set inside this function?

Utilities/build-script-helper.py
```diff
@@ -68,13 +68,16 @@ def swiftpm_bin_path(swift_exec: str, swiftpm_args: List[str], additional_env: D
...
     """Returns the target-triple of the current machine or for cross-compilation."""
     try:
         command = [swift_exec, '-print-target-info']
+        if cross_compile:
+            cross_compile_json = json.load(open(args.cross_compile_config))
+            command += ['-target', cross_compile_json["target"]]
         target_info_json = subprocess.check_output(command, stderr=subprocess.PIPE, universal_newlines=True).strip()
         args.target_info = json.loads(target_info_json)
-        if platform.system() == 'Darwin':
+        if re.search('-apple-macosx', args.target_info["target"]["unversionedTriple"]):
```
ahoppen, line 80: Do we need do do a full regex search here? Wouldn’t substring matching be sufficient?

```suggestion
        if '-apple-macosx' in args.target_info["target"]["unversionedTriple"]:
```

Utilities/build-script-helper.py
```diff
@@ -105,7 +108,10 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
         for san in args.sanitize:
             swiftpm_args += ['--sanitize=%s' % san]
 
-    if platform.system() == 'Darwin':
+    build_target = get_build_target(swift_exec, args, cross_compile=(True if args.cross_compile_config else False))
+    build_arch = build_target.split('-')[0]
+    build_os = build_target.split('-')[2]
+    if re.search('-apple-macosx', build_target):
```
ahoppen, line 114: Again, can’t we just use substring matching here?

```suggestion
    if '-apple-macosx' in build_target:
```

Alternatively, wouldn’t it even be better to just check the OS? I.e something like `build_os.startswith('macosx')`

Utilities/build-script-helper.py
buttaface: The reason it is done this way is because I copied it over from how [I did it for SPM a couple years ago](https://github.com/apple/swift-package-manager/commit/38df7c37e7ff705c6f48544a28d3d7a4839e290f#diff-fdc89d912e967e4b700e1bb41e08552bb416b5c5551bbae5b1ccac134db6c3e5R284). This is more general and allows the caller to manually enable checking for cross-compilation triples only when needed, which is used by that SPM bootstrap script.

While you're right that such generality is not needed for this script right now, if you think it might be later, better to keep it this way.

Utilities/build-script-helper.py
buttaface: No, I simply copied this over from [the existing check](https://github.com/apple/sourcekit-lsp/blob/main/Utilities/build-script-helper.py#L140), as I'm unfamiliar with the Python alternatives. I'll change them all to your suggestions.

Utilities/build-script-helper.py
buttaface: Will do.

Utilities/build-script-helper.py
ahoppen: OK, fair enough 👍 I think you also added it to SourceKit-LSP back then and I removed it when I refactored `build-script-helper.py` because we didn’t need it.

Utilities/build-script-helper.py
ahoppen: I wonder why they use regex matching. Does case sensitivity matter here? Python’s `in` is case-senstivive while `re.search` might be case-insensitive.

buttaface:
Made the changes you asked for but don't merge yet, just looking for review until this can be tested on the CI alongside the other pulls.

ahoppen (approved):
LGTM. Thanks

ahoppen:
Let me know when you’d like to run tests for this PR and merge it.

buttaface:
@ahoppen, I've been asking for someone to run all these pulls together on the CI, [see comment on linked compiler pull](https://github.com/apple/swift/pull/63782#issuecomment-1442093981), would be great if you would kick that CI run off.

ahoppen:
https://github.com/apple/swift/pull/63782

@swift-ci Please test

ahoppen:
I triggered PR testing on the main PR. Ping me if you need more testing.

buttaface:
Thanks, looks like I missed installing the local libicu for linux in this new location, will investigate.

buttaface:
@ahoppen, fixed the two issues that cropped up on the CI, ready for another CI run on the linked compiler pull with all 7 of these other pulls.

buttaface:
Removed the runpath changes since [the Swift devs want to take a different approach](https://github.com/apple/swift/pull/63782#pullrequestreview-1353429744), leaving only the build script updates we discussed.

One last CI run and this can go in, @ahoppen.

buttaface:
Ping @ahoppen, something wrong with getting this in?

buttaface:
Ping @benlangmuir, Alex appears to be busy lately, would you run through CI and merge?

benlangmuir:
@swift-ci please test

ahoppen:
Sorry, I just forgot about it and was indeed busy the last two days. Merging it now.

PR #731, opened 2023-04-18
Fix package manifest API `.branch` deprecation

MaxDesiatov:
With Swift 5.7 we should use `branch:` name argument overload instead.


MaxDesiatov:
@swift-ci test

ahoppen (approved):
Thank

Issue #577, opened 2022-06-30
Support Progress (method: '$/progress')


kkharji:
Hello,

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress

It would be great to support progress as it helper users of the server to have some form of feedback on server initialization

Thanks

ahoppen:
I agree that having that functionality would be great but it’s unlikely that it will be something that we’ll be able to implement in the near future because none of the infrastructure in sourcekitd (which is providing the semantic functionality for Swift in SourceKit-LSP) is set up to provide progress notifications. 

Specifically regarding server initialization: SourceKit-LSP doesn’t actually do a lot of work when the server is initialized because there is no background indexing going on (we require you to build the project to update the index [Link](https://github.com/apple/sourcekit-lsp#caveats)). So I don’t think progress for initialization would pretty much just finish immediately.

ahoppen:
rdar://96198579

ahoppen:
Since we don’t have any data regarding progress from sourcekitd and by the way things are defined, there’s no way we will get progress data in the future, this is not actionable, so I’m closing it. 

We can open more specific issues about progress support once SourceKit-LSP has some sort of long-running operations for which it makes sense to provide progress.

PR #732, opened 2023-04-19
Send token with WorkDoneProgress

JCWasmx86:
This fixes the [$/progress](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress) notification, as it requires that the ProgressToken is sent with the value.

`swift test` works, theoretically it should probably be called `ProgressParams`, but it wouldn't reflect that it is a notification, so I just changed the enum to `WorkDoneProgressType`

### DIFF
--- a/Sources/LanguageServerProtocol/Notifications/WorkDoneProgress.swift
+++ b/Sources/LanguageServerProtocol/Notifications/WorkDoneProgress.swift
@@ -10,9 +10,22 @@
 //
 //===----------------------------------------------------------------------===//
 
-public enum WorkDoneProgress: NotificationType, Hashable {
+public struct WorkDoneProgress: NotificationType, Hashable {
   public static var method: String = "$/progress"
 
+  /// The progress token provided by the client or server.
+  public var token: ProgressToken
+
+  /// The progress data.
+  public var value: WorkDoneProgressKind
+
+  public init(token: ProgressToken, value: WorkDoneProgressKind) {
+    self.token = token
+    self.value = value
+  }
+}
+
+public enum WorkDoneProgressKind: Codable, Hashable {
   case begin(WorkDoneProgressBegin)
   case report(WorkDoneProgressReport)
   case end(WorkDoneProgressEnd)
--- a/Tests/LanguageServerProtocolTests/CodingTests.swift
+++ b/Tests/LanguageServerProtocolTests/CodingTests.swift
@@ -1034,21 +1034,33 @@ final class CodingTests: XCTestCase {
   }
 
   func testWorkDoneProgress() {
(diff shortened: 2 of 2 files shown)


Sources/LanguageServerProtocol/Notifications/WorkDoneProgress.swift
```diff
@@ -10,9 +10,22 @@
...
+public struct WorkDoneProgress: NotificationType, Hashable {
   public static var method: String = "$/progress"
 
+  /// The progress token provided by the client or server.
+  public var token: ProgressToken
+
+  /// The progress data.
+  public var value: WorkDoneProgressType
+
+  public init(token: ProgressToken, value: WorkDoneProgressType) {
```
MaxDesiatov, line 22: Is this initializer used anywhere outside of the tests? I don't see any other references to it in this PR.

JCWasmx86: No due to #577. I use your type definitions in another language server and want  to use $/progress => I implement it and send it upstream

JCWasmx86: No due to #577. I use your type definitions in another language server and want  to use $/progress => I implement it and send it upstream

MaxDesiatov:
@swift-ci test

Sources/LanguageServerProtocol/Notifications/WorkDoneProgress.swift
```diff
@@ -10,9 +10,22 @@
...
+    self.token = token
+    self.value = value
+  }
+}
+
+public enum WorkDoneProgressType: Codable, Hashable {
```
ahoppen, line 28: Nitpick: Could you rename this to `WorkDoneProgressKind`? `Type` always makes me think of metatypes or generic types in the swift language, which this isn’t. That’s also more consistent with namings in the rest of the package.

ahoppen (commented):
Thank you. I must have missed that when going through all the struct definitions in https://github.com/apple/sourcekit-lsp/pull/669. I’ve got one naming suggestion, otherwise this looks good to me 👍🏽

MaxDesiatov:
@swift-ci test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

ahoppen:
Thank you. I don’t know how much you care but if you’d like this to be in the Swift 5.9 release, you will need to create a separate PR for the `release/5.9` branch.

JCWasmx86:
Personally it's not that important to have it at any tag, as I just use the main branch from this repo

ahoppen:
OK, in that case there’s no reason to cherry-pick it.

PR #733, opened 2023-04-19
ToolchainRegistryTests: avoid identity checks for `localFileSystem`

MaxDesiatov:
We'd like to remove `AnyObject` requirement from `FileSystem` in https://github.com/apple/swift-tools-support-core/pull/410, which will make identity checks on `any FileSystem` impossible. These checks were an anti-pattern anyway, especially in tests, which should be flexible enough to support an arbitrary file system implementation. If certain behavior is not available on `FileSystem` we should modify that protocol instead.


MaxDesiatov:
@swift-ci test

ahoppen (approved):

PR #737, opened 2023-04-21
Fix warnings for use of `try` in non-throwing context

MaxDesiatov:
Fixes "No calls to throwing functions occur within 'try' expression" warning that occurs in two difference places when building `SKSupport` and `SKSwiftPMWorkspace` targets.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
The warning was caused by outdated dependencies in my check out, closing as invalid.

PR #741, opened 2023-05-02
adjust to latest SwiftPM APIs

tomerd:
motivation: support changes in SwiftPM underlying path APIs

changes:
* adjust call sites in SwiftPMWorkspace to use SwiftPM's path

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -28,9 +28,11 @@ import Workspace
 import Dispatch
 import struct Foundation.URL
 
+import struct Basics.AbsolutePath
+import struct Basics.TSCAbsolutePath
+
 import func TSCBasic.resolveSymlinks
 import protocol TSCBasic.FileSystem
-import struct TSCBasic.AbsolutePath
 import var TSCBasic.localFileSystem
 
 /// Swift Package Manager build system and workspace support.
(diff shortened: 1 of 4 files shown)


tomerd:
tested via https://github.com/apple/swift-package-manager/pull/6509

MaxDesiatov (approved):

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -28,9 +28,11 @@ import Workspace
 import Dispatch
 import struct Foundation.URL
 
+import struct Basics.AbsolutePath
+import struct Basics.TSCAbsolutePath
```
ahoppen, line 32: What’s the rule about using `TSCAbsolutePath` vs. `AbsolutePath`? Or which features will one support but not the other?

tomerd: we are slowly transitioning SwiftPM code base off of TSC. I wanted to keep the change set smaller so lets all the LSP code using the TSC version, but since some of SwiftPM public APIs are now using SwiftPM's AbsolutePath we need do some bridging in SwiftPMWorkspace. eventually all of SwiftPM public APIs will be based on swift-system Path so we will be aligned again, this is the first step to get there

ahoppen (approved):

tomerd:
@swift-ci test

PR #743, opened 2023-05-03
Add boilerplate for pull-model diagnostics

tristanlabelle:
Adds the capabilities, options and registration structs for pull-model diagnostics. Does not report support for them yet so this code currently has no impact. This PR is also to get guidance on how to implement the document diagnostic request in `SwiftLanguageServer`. Any pointers are welcome.

# How tested
- Locally added the snippet to report support for pull request diagnostics in `SwiftLanguageServer.initializeSync`, then built, ran locally in VS Code and confirmed that I could get a dummy diagnostic shown in VS Code.
- Then removed the feature reporting snippet (the state of the code in this PR), rebuilt, reran in VS Code, and confirmed that `publishDiagnostics` still work.

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
@@ -156,6 +156,30 @@ public struct InlayHintRegistrationOptions: RegistrationOptions, TextDocumentReg
   }
 }
 
+/// Describe options to be used when registering for pull diagnostics. Since LSP 3.17.0
+public struct DiagnosticRegistrationOptions: RegistrationOptions, TextDocumentRegistrationOptionsProtocol {
+  public var textDocumentRegistrationOptions: TextDocumentRegistrationOptions
+  public var diagnosticOptions: DiagnosticOptions
+  
+  public init(
+    documentSelector: DocumentSelector? = nil,
+    diagnosticOptions: DiagnosticOptions
+  ) {
+    textDocumentRegistrationOptions = TextDocumentRegistrationOptions(documentSelector: documentSelector)
+    self.diagnosticOptions = diagnosticOptions
+  }
+
+  public func encodeIntoLSPAny(dict: inout [String: LSPAny]) {
+    textDocumentRegistrationOptions.encodeIntoLSPAny(dict: &dict)
+
+    dict["interFileDependencies"] = .bool(diagnosticOptions.interFileDependencies)
+    dict["workspaceDiagnostics"] = .bool(diagnosticOptions.workspaceDiagnostics)
+    if let workDoneProgress = diagnosticOptions.workDoneProgress {
+      dict["workDoneProgress"] = .bool(workDoneProgress)
+    }
+  }
+}
+
 /// Describe options to be used when registering for file system change events.
 public struct DidChangeWatchedFilesRegistrationOptions: RegistrationOptions {
   /// The watchers to register.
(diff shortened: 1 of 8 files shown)


tristanlabelle:
@ahoppen You mentioned that support for pull diagnostic requests already exists in sourcekit. Any pointers on how to get to that data from the dummy `SwiftLanguageServer.documentDiagnostic` function?

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -156,6 +156,30 @@ public struct InlayHintRegistrationOptions: RegistrationOptions, TextDocumentReg
...
+  }
+
+  public func encodeIntoLSPAny(dict: inout [String: LSPAny]) {
+    textDocumentRegistrationOptions.encodeIntoLSPAny(dict: &dict)
+
+    dict["interFileDependencies"] = .bool(diagnosticOptions.interFileDependencies)
+    dict["workspaceDiagnostics"] = .bool(diagnosticOptions.workspaceDiagnostics)
+    if let workDoneProgress = diagnosticOptions.workDoneProgress {
+      dict["workDoneProgress"] = .bool(workDoneProgress)
+    }
```
ahoppen, line 179: I think you also need to encode `documentSelector` (from `TextDocumentRegistrationOptions` in the LSP spec) and `id` (from `StaticRegistrationOptions` in the LSP spec) here.

ahoppen (approved):
LGTM. I really wonder how I missed all of this when updating the definitions to LSP 3.17 🤔

ahoppen:
@swift-ci Please test

ahoppen:
The main entry point where the diagnostic request is being handled in `sourcekitd` is here:

https://github.com/apple/swift/blob/0bdb39f51e54e959c44f3c0fea377d7d64155a51/tools/SourceKit/tools/sourcekitd/lib/Service/Requests.cpp#L1783-L1805

https://github.com/apple/swift/blob/main/test/SourceKit/Diagnostics/diags.swift has some examples of how the diagnostics request can be invoked using `sourcekitd-test`. You can see the raw sourcekitd request by running the test using lit with the `-a` option, which will print the output. In case you don’t want to do this yourself, this is what it produces for me.

```
{
  key.request: source.request.diagnostics,
  key.compilerargs: [
    "-module-cache-path",
    "/Users/alex/src/mbuild/Release+Asserts/swift-macosx-arm64/swift-test-results/arm64-apple-macosx10.13/clang-module-cache",
    "/Users/alex/src/swift/test/SourceKit/Diagnostics/diags.swift"
  ],
  key.sourcefile: "/Users/alex/src/swift/test/SourceKit/Diagnostics/diags.swift",
  key.primary_file: "/Users/alex/src/swift/test/SourceKit/Diagnostics/diags.swift"
}
```

And an example of how to send a request to `sourcekitd` can be found here:

https://github.com/apple/sourcekit-lsp/blob/e84b7b2b83ad4368926ca59a97c5c7484789267e/Sources/SourceKitLSP/Swift/CursorInfo.swift#L91-L106

I hope this gives you enough pointers to get started. Let me know if you need more help.

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
tristanlabelle: `documentSelector` is already added to the dict by `textDocumentRegistrationOptions.encodeIntoLSPAny(dict: &dict)` at least.

ahoppen: Whoops, and I apparently approved and merged the PR, forgotting that I still had a comment. If you could create a follow-up PR, that would be great.

PR #746, opened 2023-05-11
Implement pull-model documentDiagnostics

### LINKED ISSUE #500: [SR-16080] SourceKit-LSP isn't aware of changes in other files
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-16080      |
   |Radar             | None         |
   |Original Reporter | @adam-fowler      |
   |Type              | Bug    |

<details>
  <summary>Environment</summary>

swift-driver version: 1.45.2 Apple Swift version 5.6 (swiftlang-5.6.0.323.62 clang-1316.0.20.8)

Visual Studio Code: 1.65.2

vscode-swift: 0.4.1

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | SourceKit-LSP    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 00ca6d48e63005108f7d259bffdb1d9a

  </details>

**Issue Description:**

Changes in one file are not always registered by another.

Repro:

-   Create a new project

-   Open VSCode

-   Create a file and add a struct called Test

-   Create a second file and add a function using a struct called Test2

-   Go back to first file and rename Test to Test2 such that your code will compile

-   The "cannot find type 'Test2' in scope" error still displays

-   Compile the project (notice it compiles fine)

-   The "cannot find type 'Test2' in scope" error still displays.

-   The only way to fix this is to edit the second file and re-type Test2

Expected:

Changes that fix an error in another file should register as fixing that error

tristanlabelle:
Implements LSP 3.17.0's pull-model documentDiagnostics request, which fix stale squiggly issues after fixing an error due to the relationship between code in two files. When using pull-model documentDiagnostics, we disable the push-model publishDiagnostics to avoid reporting them twice.

PullDiagsDemo
(this gif was using a temporary "pulldiag" source name to verify that it wasn't publish diagnostics in action)

Fixes #500

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -128,7 +128,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     // in addition to the existing push-based publish notifications.
     // If the client supports pull diagnostics, we report the capability
     // and we should disable the publish notifications to avoid double-reporting.
-    return capabilityRegistry.pullDiagnosticsRegistration(.swift) == nil
+    return capabilityRegistry.pullDiagnosticsRegistration(for: .swift) == nil
   }
   
   private var state: LanguageServerState {
@@ -152,7 +152,6 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   public init?(
     client: LocalConnection,
     toolchain: Toolchain,
-    clientCapabilities: ClientCapabilities?,
     options: SourceKitServer.Options,
     workspace: Workspace,
     reopenDocuments: @escaping (ToolchainLanguageServer) -> Void
@@ -977,6 +976,7 @@ extension SwiftLanguageServer {
   }
 
   public func foldingRange(_ req: Request<FoldingRangeRequest>) {
+    let foldingRangeCapabilities = capabilityRegistry.clientCapabilities.textDocument?.foldingRange
     queue.async {
       guard let snapshot = self.documentManager.latestSnapshot(req.params.textDocument.uri) else {
         log("failed to find snapshot for url \(req.params.textDocument.uri)")
@@ -1155,17 +1155,16 @@ extension SwiftLanguageServer {
         }
       }
 
-      let capabilities = capabilityRegistry.clientCapabilities.textDocument?.foldingRange
       // If the limit is less than one, do nothing.
-      if let limit = capabilities?.rangeLimit, limit <= 0 {
+      if let limit = foldingRangeCapabilities?.rangeLimit, limit <= 0 {
         req.reply([])
         return
       }
 
       let rangeFinder = FoldingRangeFinder(
         snapshot: snapshot,
-        rangeLimit: capabilities?.rangeLimit,
-        lineFoldingOnly: capabilities?.lineFoldingOnly ?? false)
+        rangeLimit: foldingRangeCapabilities?.rangeLimit,
+        lineFoldingOnly: foldingRangeCapabilities?.lineFoldingOnly ?? false)
       rangeFinder.walk(sourceFile)
       let ranges = rangeFinder.finalize()
 
@@ -1180,12 +1179,12 @@ extension SwiftLanguageServer {
     ]
     let wantedActionKinds = req.params.context.only
     let providers = providersAndKinds.filter { wantedActionKinds?.contains($0.1) != false }
+    let codeActionCapabilities = capabilityRegistry.clientCapabilities.textDocument?.codeAction
     retrieveCodeActions(req, providers: providers.map { $0.provider }) { result in
       switch result {
       case .success(let codeActions):
-        let capabilities = capabilityRegistry.clientCapabilities.textDocument?.codeAction
         let response = CodeActionRequestResponse(codeActions: codeActions,
-                                                 clientCapabilities: capabilities)
+                                                 clientCapabilities: codeActionCapabilities)
         req.reply(response)
       case .failure(let error):
         req.reply(.failure(error))
@@ -1366,15 +1365,14 @@ extension SwiftLanguageServer {
 
     let supportsCodeDescription = capabilityRegistry.clientHasDiagnosticsCodeDescriptionSupport
 
-    let handle = self.sourcekitd.send(skreq, self.queue) { [weak self] response in
-      guard let self = self else { return }
+    let handle = self.sourcekitd.send(skreq, self.queue) { response in
       guard let dict = response.success else {
         return completion(.failure(ResponseError(response.failure!)))
       }
 
       var diagnostics: [Diagnostic] = []
       dict[keys.diagnostics]?.forEach { _, diag in
-        if var diagnostic = Diagnostic(diag, in: snapshot, useEducationalNoteAsCode: supportsCodeDescription) {
+        if let diagnostic = Diagnostic(diag, in: snapshot, useEducationalNoteAsCode: supportsCodeDescription) {
           diagnostics.append(diagnostic)
         }
         return true
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -1377,7 +1377,6 @@ extension SwiftLanguageServer {
       var diagnostics: [Diagnostic] = []
       dict[keys.diagnostics]?.forEach { _, diag in
         if var diagnostic = Diagnostic(diag, in: snapshot, useEducationalNoteAsCode: supportsCodeDescription) {
-          diagnostic.source = "pulldiag"
           diagnostics.append(diagnostic)
         }
         return true
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -103,7 +103,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   let sourcekitd: SourceKitD
 
-  let clientCapabilities: ClientCapabilities
+  let capabilityRegistry: CapabilityRegistry
 
   let serverOptions: SourceKitServer.Options
   
@@ -128,7 +128,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     // in addition to the existing push-based publish notifications.
     // If the client supports pull diagnostics, we report the capability
     // and we should disable the publish notifications to avoid double-reporting.
-    return clientCapabilities.textDocument?.diagnostic == nil
+    return capabilityRegistry.pullDiagnosticsRegistration(.swift) == nil
   }
   
   private var state: LanguageServerState {
@@ -160,7 +160,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     guard let sourcekitd = toolchain.sourcekitd else { return nil }
     self.client = client
     self.sourcekitd = try SourceKitDImpl.getOrCreate(dylibPath: sourcekitd)
-    self.clientCapabilities = clientCapabilities ?? ClientCapabilities(workspace: nil, textDocument: nil)
+    self.capabilityRegistry = workspace.capabilityRegistry
     self.serverOptions = options
     self.documentManager = DocumentManager()
     self.state = .connected
@@ -250,7 +250,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   /// Inform the client about changes to the syntax highlighting tokens.
   private func requestTokensRefresh() {
-    if clientCapabilities.workspace?.semanticTokens?.refreshSupport ?? false {
+    if capabilityRegistry.clientHasSemanticTokenRefreshSupport {
       _ = client.send(WorkspaceSemanticTokensRefreshRequest(), queue: queue) { result in
         if let error = result.failure {
           log("refreshing tokens failed: \(error)", level: .warning)
@@ -293,8 +293,7 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
     let stageUID: sourcekitd_uid_t? = response[sourcekitd.keys.diagnostic_stage]
     let stage = stageUID.flatMap { DiagnosticStage($0, sourcekitd: sourcekitd) } ?? .sema
 
-    let supportsCodeDescription =
-           (clientCapabilities.textDocument?.publishDiagnostics?.codeDescriptionSupport == true)
+    let supportsCodeDescription = capabilityRegistry.clientHasDiagnosticsCodeDescriptionSupport
 
     // Note: we make the notification even if there are no diagnostics to clear the current state.
     var newDiags: [CachedDiagnostic] = []
@@ -1156,7 +1155,7 @@ extension SwiftLanguageServer {
         }
       }
 
-      let capabilities = self.clientCapabilities.textDocument?.foldingRange
+      let capabilities = capabilityRegistry.clientCapabilities.textDocument?.foldingRange
       // If the limit is less than one, do nothing.
       if let limit = capabilities?.rangeLimit, limit <= 0 {
         req.reply([])
@@ -1184,7 +1183,7 @@ extension SwiftLanguageServer {
     retrieveCodeActions(req, providers: providers.map { $0.provider }) { result in
       switch result {
       case .success(let codeActions):
-        let capabilities = self.clientCapabilities.textDocument?.codeAction
+        let capabilities = capabilityRegistry.clientCapabilities.textDocument?.codeAction
         let response = CodeActionRequestResponse(codeActions: codeActions,
                                                  clientCapabilities: capabilities)
         req.reply(response)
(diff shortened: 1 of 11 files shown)


Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -122,6 +122,14 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   var keys: sourcekitd_keys { return sourcekitd.keys }
   var requests: sourcekitd_requests { return sourcekitd.requests }
   var values: sourcekitd_values { return sourcekitd.values }
+
+  var enablePublishDiagnostics: Bool {
+    // Since LSP 3.17.0, diagnostics can be reported through pull-based requests,
+    // in addition to the existing push-based publish notifications.
+    // If the client supports pull diagnostics, we report the capability
+    // and we should disable the publish notifications to avoid double-reporting.
+    return clientCapabilities.textDocument?.diagnostic == nil
```
tristanlabelle, line 131: Should this rather be testing `clientCapabilities.textDocument?.diagnostic?.dynamicRegistration != true`? I'm not sure if we would receive the pull requests if dynamic registration is not supported by the client.

tristanlabelle:
Two questions before I publish this PR:
-

tristanlabelle:
@ahoppen Good news! I got this to work and it fixes some of the stale squiggly cases we knew of. Can you check the two comments I left on the PR as well? Thanks for the earlier pointers!

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: I think this is the right approach. Correct me if you think I’m wrong but checking for `dynamicRegistration` would only check if the client supports dynamic registration of the capability. If, for some reason, it supported registration of the capability at server initialization but not dynamically after the fact, checking for `dynamicRegistration` wouldn’t catch that.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -122,6 +122,14 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   var keys: sourcekitd_keys { return sourcekitd.keys }
   var requests: sourcekitd_requests { return sourcekitd.requests }
   var values: sourcekitd_values { return sourcekitd.values }
+
+  var enablePublishDiagnostics: Bool {
+    // Since LSP 3.17.0, diagnostics can be reported through pull-based requests,
+    // in addition to the existing push-based publish notifications.
+    // If the client supports pull diagnostics, we report the capability
+    // and we should disable the publish notifications to avoid double-reporting.
```
ahoppen, line 130: Is it documented somewhere that LSP servers shouldn’t publish diagnostics if the client supports `pullDiagnostics`? Or, if not, do you know if other LSP servers are also doing this as well or if VSCode is relying on it?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -370,7 +381,10 @@ extension SwiftLanguageServer {
         range: .bool(true),
         full: .bool(true)),
       inlayHintProvider: .value(InlayHintOptions(
-        resolveProvider: false))
+        resolveProvider: false)),
+      diagnosticProvider: DiagnosticOptions(
+        interFileDependencies: true,
+        workspaceDiagnostics: false)
```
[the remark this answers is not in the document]
ahoppen: No, there is no `workspaceDiagnostic` equivalent in sourcekitd.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -370,7 +381,10 @@ extension SwiftLanguageServer {
         range: .bool(true),
         full: .bool(true)),
       inlayHintProvider: .value(InlayHintOptions(
-        resolveProvider: false))
+        resolveProvider: false)),
+      diagnosticProvider: DiagnosticOptions(
+        interFileDependencies: true,
```
ahoppen, line 386: How do you model the `interFileDependencies`? I don’t think sourcekitd keeps track of that in any way, so I would think that we should return `false` here.

ahoppen (commented):
Thank you for implementing support for the diagnostics request. The implementation is actually a lot shorter than I thought it would be, so great job! I’ve got a couple of questions, but overall, this looks great to me. 

One request though: Could you also write a test case for diagnostic pulling?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle:
```
/**
	 * Whether the language has inter file dependencies meaning that
	 * editing code in one file can result in a different diagnostic
	 * set in another file. Inter file dependencies are common for
	 * most programming languages and typically uncommon for linters.
	 */
```

My read of this is that by setting `interFileDependencies: true`, we're telling the LSP client that it's worth re-querying diagnostics for file B after changes in file A since it might have fixed them, so this is what allows the behavior in the GIF in this PR summary.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: We have the same understanding of `dynamicRegistration`. My read of `SourceKitServer` is that it doesn't report supporting pull diagnostics at initialization. Later, when it is asked about a swift file, it starts `SwiftLanguageServer`, initializes it, sees that it reports supporting pull diagnostics and at that point dynamically registers support for pull diagnostics to the LSP client. So, we depend on `dynamicRegistration` being supported.

Or maybe we should unconditionally report support for pull diagnostics during `SourceKitServer`'s initialization, avoiding dynamic registration? If our underlying language server doesn't support the request, `SourceKitServer` will return no diagnostics so I think there's no harm done.

tristanlabelle:
@ahoppen

> One request though: Could you also write a test case for diagnostic pulling?

I'll do my best. I was under the impression that the sourcekit-lsp tests weren't working on Windows and my macOS-fu is poor at best. :/

compnerd:
> @ahoppen
> 
> > One request though: Could you also write a test case for diagnostic pulling?
> 
> I'll do my best. I was under the impression that the sourcekit-lsp tests weren't working on Windows and my macOS-fu is poor at best. :/

The tests worked (at one point) on Windows.  They cannot be built currently due to the swift-syntax dependency and SPM miscompiling code. You should be able to roll back to prior to the swift-syntax adoption, port your code to that point and run the tests.

Alternatively, you can enumerate all test cases into a manifest, add CMake rules, and build the tests locally with CMake.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: That shouldn't be a problem. `SourceKitServer.initialize()` can unconditionally report support for pull diagnostics, but `SwiftLanguageServer`'s logic is based on the `ClientCapabilities` coming from the LSP Client, so it will only disable publish diagnostics if the LSP client supports pull diagnostics.

tristanlabelle:
Or someone could ask swift-ci to build and test this PR for me? 😅 I have added a simple test mimicking the publish diagnostics test so there's a nonzero chance that it'll pass without me having to do much local build shenanigans... this time around.

compnerd:
@swift-ci please test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
adam-fowler: Ah sorry I'm getting mixed up between the two language servers

ahoppen:
> Or someone could ask swift-ci to build and test this PR for me? 😅 I have added a simple test mimicking the publish diagnostics test so there's a nonzero chance that it'll pass without me having to do much local build shenanigans... this time around.

Sorry, I have meant to reply to this one and then forgot. I ran the test case locally and it had a few syntactic issues but I was able to fix them easily. Unfortunately, I forgot about the changes and discarded them. I can fix your test case again after you addressed https://github.com/apple/sourcekit-lsp/pull/746#discussion_r1194117022.

tristanlabelle:
@ahoppen On it! Sorry I was pulled into other things momentarily but I aim to get this all wrapped up and merged by EOW.

ahoppen:
No worries, @tristanlabelle. Thanks for doing this work in the first place.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: @ahoppen I'm almost done implementing your suggestion but now that it's taking shape, I'm not sure this is the right approach. It adds some complexity for determining if we should or not publish diagnostics, and it requires dynamic capability registration.

I'm thinking a better approach would be to stop publishing diagnostics in the language-specific server as soon as we receive a first pull diagnostics request, so we have proof that the client is using them. We could publish an empty diagnostics array at that point to clear up any previously published diagnostics, then set a flag to not publish any further diagnostics and only respond to pull requests.

I also think it makes sense to unconditionally reporting pull diagnostics as supported in the SourceKitServer (independent of language-specific servers), rather than using dynamic registration when one language-specific server comes online and requires it. We have a well-defined response if we get a pull diagnostics request and the language-specific server doesn't support it (return an empty diagnostics array), and it won't stop the client from responding to publish diagnostics messages by language-specific servers who do not support pull diagnostics. Thoughts?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: I suggest a solution to this here: https://github.com/apple/sourcekit-lsp/pull/746/files#r1206839125

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Honestly, what I really want, would be to see how other language servers handle this. But I haven’t found a single major LSP server that supports both `publishDiagnsotics` and `textDocument/diagnostic`. Also, what surprises me, is that I can’t find a PR or issue on https://github.com/microsoft/language-server-protocol that mentions `textDocument/diagnostics`, which kind of surprises me because I thought there would be some kind of discussion about these features. 

Anyway, given the lack of precedence or clarification, I think we really need to make up our own and out of all the alternatives we have considered, I think not sending `publishDiagnostics` notifications if both the server and the client support “Pull Diagnostics” still seems like the cleanest design to me.

---

> It adds some complexity for determining if we should or not publish diagnostics, and it requires dynamic capability registration.

What’s the added complexity here? I think passing down the `CapabilityRegistry` instead of `ClientCapabilities` as you do it in your PR works rather nicely. I haven’t looked at it in too much detail though.

> I'm thinking a better approach would be to stop publishing diagnostics in the language-specific server as soon as we receive a first pull diagnostics request, so we have proof that the client is using them. 

Hmm, I’m not a huge fan because this means that sending a request silently changes the entire behavior of SourceKit-LSP, which is very confusing IMO. I think whether `publishDiagnostic` notifications should be determined when the client and server agree on capabilities.

> I also think it makes sense to unconditionally reporting pull diagnostics as supported in the SourceKitServer (independent of language-specific servers), rather than using dynamic registration when one language-specific server comes online and requires it. We have a well-defined response if we get a pull diagnostics request and the language-specific server doesn't support it (return an empty diagnostics array), and it won't stop the client from responding to publish diagnostics messages by language-specific servers who do not support pull diagnostics. Thoughts?

But wouldn’t that leave us with the following problem: `clangd` publishes its diagnostics using `publishDiagnostics`. Then the client requests diagnostics using `textDocument/diagnostics` and SourceKit-LSP returns an empty array, clearing any diagnostics displayed on the client.

tristanlabelle: Oh I think that your last point reveals our different assumptions. My assumption is that the namespacing of pull and publish diagnostics is separate and they don't interact because that would be hard to manage, but I'm still looking for some confirmation of this (specs or reference implementations).

ahoppen: Oh, yes. If they are indeed in different namespaces, that would certainly change things.

tristanlabelle: Let's see if this gets us more info: https://github.com/microsoft/language-server-protocol/issues/1743

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: In the meanwhile, I finished and pushed your suggested implementation, so we can make some progress. It's ready to review now.

MaxDesiatov:
@swift-ci test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -1326,10 +1338,75 @@ extension SwiftLanguageServer {
...
+          diagnostics.append(diagnostic)
+        }
+        return true
+      }
+
+      completion(.success(diagnostics))
+    }
+
+    // FIXME: cancellation
+    _ = handle
```
ahoppen, line 1385: I think you should be able to implement cancellation similar to here

https://github.com/apple/sourcekit-lsp/blob/b7d427069c869efa1ae8764879935e6b877dab59/Sources/SourceKitLSP/Swift/OpenInterface.swift#L97-L101

Tests/SourceKitLSPTests/PullDiagnosticsTests.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    let report: DocumentDiagnosticReport
+    do {
+      report = try sk.sendSync(request)
+    } catch let error as ResponseError where error.message.contains("unknown request: source.request.diagnostics") {
+      throw XCTSkip("toolchain does not support source.request.diagnostics request")
+    }
+
+    guard case .full(let fullReport) = report else {
+        XCTFail("Unexpected diagnostics report type: \(report)")
```
ahoppen, line 64: Nitpick

```suggestion
      XCTFail("Unexpected diagnostics report type: \(report)")
```

ahoppen (commented):
Thanks. I think we have landed at a good design now and I’m curious to see if we’ll get a response from issue you filed to the LSP spec repo. 

Also the refactoring of how we check for capabilities looks like a general improvement to me which usually shows that we’re on the right track. Thanks for your continued effort on this PR.

I’ve got two minor comments inline, otherwise, I think this is good to be merged.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
tristanlabelle: Could we leave this to a PR that fixes all cancellation at once? This was copied from [many places](https://github.com/search?q=repo%3Aapple%2Fsourcekit-lsp%20%2F%2F%20FIXME%3A%20cancellation&type=code)

tristanlabelle:
I would appreciate a ci kick. I'm still writing the unit test blindly due to current Windows build limitations. (@ahoppen if you get a chance to build locally that'd be super helpful)

MaxDesiatov:
@swift-ci test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Sounds good to me as well

ahoppen:
I just ran the test locally and if you apply the following diff, they pass

```diff
diff --git a/Tests/SourceKitLSPTests/PullDiagnosticsTests.swift b/Tests/SourceKitLSPTests/PullDiagnosticsTests.swift
index fa2b43f8..a28f23f4 100644
--- a/Tests/SourceKitLSPTests/PullDiagnosticsTests.swift
+++ b/Tests/SourceKitLSPTests/PullDiagnosticsTests.swift
@@ -16,6 +16,10 @@ import SKTestSupport
 import XCTest
 
 final class PullDiagnosticsTests: XCTestCase {
+  enum Error: Swift.Error {
+    case unexpectedDiagnsoticReport
+  }
+
   /// Connection and lifetime management for the service.
   var connection: TestSourceKitServer! = nil
 
@@ -61,8 +65,7 @@ final class PullDiagnosticsTests: XCTestCase {
     }
 
     guard case .full(let fullReport) = report else {
-      XCTFail("Unexpected diagnostics report type: \(report)")
-      return
+      throw Error.unexpectedDiagnsoticReport
     }
 
     return fullReport.items
```

tristanlabelle:
Thank you, @ahoppen ! I applied your diff.

ahoppen:
@swift-ci Please test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
Windows builds need to be started separately unfortunately.

tristanlabelle:
Woo! All green :D

ahoppen:
@tristanlabelle Can you also cherry-pick this PR to the `release/5.9` branch? I think it would be useful to get this into the 5.9 release.

tristanlabelle:
Cherry-picked to #751 (after requashing - this PR was completed by merging so there's a bit of my dirty laundry in the history now)

PR #747, opened 2023-05-16
Swiftinterface symbol lookup

adam-fowler:
On calling `definition` if a symbol returns it is in a file with a `.swiftinterface` extension it will use OpenInterface to create the generated interface file. It will then call sourcekitd `source.request.editor.find_usr` to find the location of the symbol in the generated file.

- `OpenInterfaceRequest` has been extended to include an optional symbol name.
- Broke out openInterface code in `SourceKitServer.definition` into a separate function: `respondWithInterface`, as it is called from multiple sites now
- If first occurrence of symbol definition occurs inside a `swiftInterface` file then use `respondWithInterface` to build generated interface and then find symbol location in generated interface.
- Use DocumentManager to store instances of generated interface files, to avoid generating them on every call to `openInterface` and also cache the line table.
- Added `_findUSR` function that calls `source.request.editor.find_usr`

### DIFF
--- a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
@@ -22,9 +22,13 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
   /// The module to generate an index for.
   public var name: String
 
-  public init(textDocument: TextDocumentIdentifier, name: String) {
+  /// The symbol to search for in the generated module interface.
+  public var symbol: String?
+
+  public init(textDocument: TextDocumentIdentifier, name: String, symbol: String?) {
     self.textDocument = textDocument
     self.name = name
+    self.symbol = symbol
   }
 }
 
@@ -32,8 +36,10 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
 public struct InterfaceDetails: ResponseType, Hashable {
 
   public var uri: DocumentURI
+  public var position: Position?
 
-  public init(uri: DocumentURI) {
+  public init(uri: DocumentURI, position: Position?) {
     self.uri = uri
+    self.position = position
   }
 }
--- a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
@@ -20,15 +20,35 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
   public var textDocument: TextDocumentIdentifier
 
   /// The module to generate an index for.
-  public var name: String
+  public var moduleName: String
+
+  /// The module group name.
+  public var groupNames: [String]
 
   /// The symbol to search for in the generated module interface.
   public var symbol: String?
 
   public init(textDocument: TextDocumentIdentifier, name: String, symbol: String?) {
     self.textDocument = textDocument
-    self.name = name
     self.symbol = symbol
+    // Stdlib Swift modules are all in the "Swift" module, but their symbols return a module name `Swift.***`.
+    let splitName = name.split(separator: ".")
+    if splitName.count == 1 {
+      self.moduleName = name
+      self.groupNames = []
+    } else {
+      self.moduleName = String(splitName[0])
+      self.groupNames = [String.SubSequence](splitName.dropFirst()).map { String($0)}
+    }
+  }
+
+  /// Name of interface module name with group names appended
+  public var name: String {
+    if groupNames.count > 0 {
+      return "\(self.moduleName).\(self.groupNames.joined(separator: "."))"
+    } else {
+      return self.moduleName
+    }
   }
 }
 
--- a/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
@@ -25,21 +25,16 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
   /// The module group name.
   public var groupNames: [String]
 
-  /// The symbol to search for in the generated module interface.
-  public var symbol: String?
+  /// The symbol USR to search for in the generated module interface.
+  public var symbolUSR: String?
 
-  public init(textDocument: TextDocumentIdentifier, name: String, symbol: String?) {
+  public init(textDocument: TextDocumentIdentifier, name: String, symbolUSR: String?) {
     self.textDocument = textDocument
-    self.symbol = symbol
+    self.symbolUSR = symbolUSR
     // Stdlib Swift modules are all in the "Swift" module, but their symbols return a module name `Swift.***`.
     let splitName = name.split(separator: ".")
-    if splitName.count == 1 {
-      self.moduleName = name
-      self.groupNames = []
-    } else {
-      self.moduleName = String(splitName[0])
-      self.groupNames = [String.SubSequence](splitName.dropFirst()).map { String($0)}
-    }
+    self.moduleName = String(splitName[0])
+    self.groupNames = [String.SubSequence](splitName.dropFirst()).map(String.init)
   }
 
   /// Name of interface module name with group names appended
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1307,7 +1307,7 @@ extension SourceKitServer {
 
       // If this symbol is a module then generate a textual interface
       if case .success(let symbols) = result, let symbol = symbols.first, symbol.kind == .module, let name = symbol.name {
-        self.respondWithInterface(req, moduleName: name, symbol: nil, languageService: languageService)
+        self.respondWithInterface(req, moduleName: name, symbolUSR: nil, languageService: languageService)
         return
       }
 
@@ -1330,7 +1330,7 @@ extension SourceKitServer {
           self.respondWithInterface(
             req, 
             moduleName: moduleName, 
-            symbol: firstResolved.occurrence?.symbol.usr,
+            symbolUSR: firstResolved.occurrence?.symbol.usr,
             languageService: languageService
           )
           return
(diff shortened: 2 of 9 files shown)


Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -81,4 +110,46 @@ extension SwiftLanguageServer {
...
+    }
+    let keys = self.keys
+    let skreq = SKDRequestDictionary(sourcekitd: sourcekitd)
+
+    skreq[keys.request] = requests.find_usr
+    skreq[keys.sourcefile] = uri.pseudoPath
+    skreq[keys.usr] = symbol
+
+    if let compileCommand = self.commandsByFile[uri] {
+      skreq[keys.compilerargs] = compileCommand.compilerArgs
```
adam-fowler, line 132: Wasn't sure if this was necessary

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1354,6 +1354,34 @@ extension SourceKitServer {
...
+    _ req: Request<DefinitionRequest>,
+    moduleName: String,
+    symbol: String?,
+    languageService: ToolchainLanguageServer
+  ) {
+      var moduleName = moduleName
+      // Stdlib Swift modules are all in the "Swift" module, but their symbols return a module name `Swift.***`.
+      if moduleName.hasPrefix("Swift.") {
+        moduleName = "Swift"
+      }
```
adam-fowler, line 1367: Is there a better way to do this? Are there other modules that might be split up like this?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1335,6 +1322,19 @@ extension SourceKitServer {
 
       switch extractedResult {
       case .success(let resolved):
+        // if first resolved location is in `.swiftinterface` file. Use moduleName to return 
+        // textual interface 
+        if let firstResolved = resolved.first, 
+           let moduleName = firstResolved.occurrence?.location.moduleName, 
+           firstResolved.location.uri.fileURL?.pathExtension == "swiftinterface" {
```
adam-fowler, line 1329: Is checking the extension of location uri enough here?

bnbarham:
I thought I might give some extra background here. I haven't looked into the current infrastructure too much, but I'll mention what I would expect anyway from a Swift perspective and then we can discuss whether that's feasible to do or not.

The cursor info request returns the group name (which is where the module splitting is coming from) as well as whether it's "dynamic", ie. whether it could be overridden. In the dynamic case we will also provide the "receiver" USRs which are the statically known types of the member that's being called.

If the returned symbol isn't dynamic, then there's no need to do an index lookup - we can jump straight to the location provided by the result (*). In the case that location is a header we can make a choice:
1. Jump to the header
2. Do an index lookup for the definition and jump to that (if we have the definition)
3. Jump to the generated interface for that module
4. Provide the user a choice for one of these

If it is dynamic, we should always do an index lookup - looking up and down the hierarchy from the given receiver types. This is basically to avoid going off into a separate subtree, eg. if we had D: C, C: B, B: A, E: B and a call on `C.foo()` we wouldn't want to provide `E.foo()` as a result, but we would want to give `D.foo()` since it's possible that's the actual call.

(*) This result wouldn't be any more/less up to date than the index since we only index-while-building. Though if we instead had background indexing, it might be better to consider the index result all the time. We could also background build though so... 🤷The dynamic/receiver USR data is also in the index.

-----

With that said...

> Is checking the extension of location uri enough here?

Depends on what we actually want to do regarding the header case above. Another case to consider is what happens when library evolution isn't enabled, though I'm not sure we really want to to support that in general. I *believe* (but haven't checked) that we don't return a path at all in that case, but ideally we'd still jump to the generated interface. The index wouldn't contain any results from such modules anyway.

If this is just about providing the generated interface for swift modules, then this check works well enough.

> Is there a better way to do this? Are there other modules that might be split up like this?

Ideally we'd pass the group name to the generated interface request in this case. We could either get that by splitting (which is fine as long as we check that the symbol is from swift and not C-family) or from the cursor info request. Technically any library could provide this via `-group-info-path`, though it really only matters for when there's no source.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: To extend on @bnbarham’s comment, you can specify the group name by adding the following to the `editor.open.interface` request.

```
key.groupname: "String",
```

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -14,30 +14,59 @@ import Foundation
...
+      // has interface already been generated
+      if let snapshot = self.documentManager.latestSnapshot(interfaceDocURI) {
+        self._findUSR(request: request, uri: interfaceDocURI, snapshot: snapshot, symbol: symbol) { result in
+          switch result {
+            case .success(let info):
+              request.reply(.success(InterfaceDetails(uri: interfaceDocURI, position: info.position)))
+            case .failure:
+              request.reply(.success(InterfaceDetails(uri: interfaceDocURI, position: nil)))
+          }
+        }
```
ahoppen, line 44: I wonder whether we can share this code between the two `if` branches

Sources/SourceKitLSP/Swift/OpenInterface.swift
ahoppen: Compiler args are not used for the `find_usr` request, so there’s no need to specify them

https://github.com/apple/swift/blob/31032b9cae300878312dc21ddbd515c744553db6/tools/SourceKit/tools/sourcekitd/lib/Service/Requests.cpp#L3706-L3723

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -81,4 +110,46 @@ extension SwiftLanguageServer {
...
+    
+    let handle = self.sourcekitd.send(skreq, self.queue) { result in
+      switch result {
+      case .success(let dict):
+          if let offset: Int = dict[keys.offset],
+            let position = snapshot.positionOf(utf8Offset: offset) {
+              return completion(.success(FindUSRInfo(position: position)))
+          } else {
+            return completion(.success(FindUSRInfo(position: nil)))
+          }
```
ahoppen, line 143: Should be indented by 2 spaces, I think.

adam-fowler: Do you use one of the swift formats to format the code?

ahoppen: No, SourceKit-LSP is manual 2 space indentation

Sources/SourceKitLSP/SourceKitServer.swift
adam-fowler: I moved this code into the initialiser for OpenInterfaceRequest, made it slightly more generic and also set `key.groupname` when calling `editor.open.interface`. It appears to take an array of Strings so you can set ["Math", "Integers"] for sub-modules of sub-modules.

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -20,20 +20,46 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
...
-    self.name = name
+    self.symbol = symbol
+    // Stdlib Swift modules are all in the "Swift" module, but their symbols return a module name `Swift.***`.
+    let splitName = name.split(separator: ".")
+    if splitName.count == 1 {
+      self.moduleName = name
+      self.groupNames = []
+    } else {
+      self.moduleName = String(splitName[0])
+      self.groupNames = [String.SubSequence](splitName.dropFirst()).map { String($0)}
```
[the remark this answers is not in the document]
adam-fowler: DropFirst returns an ArraySlice not an Array so I don't your change would work.

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -20,20 +20,46 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
...
+
+  public init(textDocument: TextDocumentIdentifier, name: String, symbol: String?) {
     self.textDocument = textDocument
-    self.name = name
+    self.symbol = symbol
+    // Stdlib Swift modules are all in the "Swift" module, but their symbols return a module name `Swift.***`.
+    let splitName = name.split(separator: ".")
+    if splitName.count == 1 {
+      self.moduleName = name
+      self.groupNames = []
```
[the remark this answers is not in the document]
adam-fowler: Yeah there was a reason for that but then I changed the code and it probably isn't needed anymore

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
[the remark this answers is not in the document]
ahoppen: Ah, I see. In that case: There’s a missing space before the `}`. Sorry for being so picky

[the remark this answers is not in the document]
adam-fowler: You need to setup a formatter then all of these things would be sorted automatically. 😄

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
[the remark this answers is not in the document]
adam-fowler: Removed special case

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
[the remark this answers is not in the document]
adam-fowler: Used String.init instead

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
```diff
@@ -20,20 +20,46 @@ public struct OpenInterfaceRequest: TextDocumentRequest, Hashable {
...
 
   /// The module to generate an index for.
-  public var name: String
+  public var moduleName: String
 
-  public init(textDocument: TextDocumentIdentifier, name: String) {
+  /// The module group name.
+  public var groupNames: [String]
+
+  /// The symbol to search for in the generated module interface.
```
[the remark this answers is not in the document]
adam-fowler: Renamed to symbolUSR

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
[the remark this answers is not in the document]
adam-fowler: Done

Sources/LanguageServerProtocol/Requests/OpenInterfaceRequest.swift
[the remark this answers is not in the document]
ahoppen: Yes, I agree. We do it for swift-syntax now. We should also run swift-format on SourceKit-LSP

ahoppen (approved):
Thank you so much for this

ahoppen:
@swift-ci Please test

adam-fowler:
@ahoppen Hopefully fixed the tests. Had to create a new test SwiftPM project for the swift interface tests as my edits of an existing one broke another test. Also added tests for symbols from multiple libs, with submodules and not.

ahoppen (approved):
Thanks

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@adam-fowler Could you also create a cherry-pick PR for release/5.9? I think it would be worth to include this improvement in the 5.9 release.

Issue #618, opened 2022-08-28
Add support for jump to definition for stdlib/Foundation symbols


adam-fowler:
This work involves generating an interface view for a module.
Working out the location of the symbol within this generated file.
Somehow passing all of this in a response.

How this is all done and how much of it should be up to sourcekit-lsp is undefined at this point in time. I've basically added this as a placeholder for work to be done in the future, once we have a better idea what that work should be.

adam-fowler:
I see SourceKit-LSP already supplies the path to the `.swiftinterface` file in the `textDocument\definition` response. If SourceKit-LSP could provide an actual position in that file and we had a method of combining the `.swiftinterface` and the related `.swiftdoc` we could possibly do a lot of the work for this on the VS Code side.

ahoppen:
rdar://99329577

adam-fowler:
See https://github.com/swift-server/vscode-swift/issues/408 for update

PR #750, opened 2023-05-24
Swiftinterface symbol lookup (release/5.9)

adam-fowler:
* **Explanation**: Jumping to a definition to a type that is defined in the SDK and not the user’s code currently doesn’t work because SourceKit-LSP doesn’t know where to jump to. This add support for jumping to a declaration in the SDK by using the existing infrastructure to generate a generated interface for the SDK module, using sourcekitd to figure out where in that interface the type is declared and jumping to it.
* **Scope**: Jump to definition of SDK symbols
* **Risk**: Low, previously jump to definition didn’t work for these symbols add all 
* **Testing**: Added test case 
* **Issue**: https://github.com/apple/sourcekit-lsp/issues/618 / rdar://99329577
* **Reviewer**:  @ahoppen on #747


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test macOS

PR #761, opened 2023-07-03
Port incremental parse ability to sourcekit-lsp

StevenWong12:
To sync changes in https://github.com/apple/swift-syntax/pull/1685

This feature will be used when we call `changeDocument` in SwiftLanguageServer.

TODO: 
- [x] Add test cases for it

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -118,6 +118,9 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   var currentCompletionSession: CodeCompletionSession? = nil
 
   var commandsByFile: [DocumentURI: SwiftCompileCommand] = [:]
+  
+  /// *For Testing*
+  public var reusedNodeCallback: ReusedNodeCallback?
 
   var keys: sourcekitd_keys { return sourcekitd.keys }
   var requests: sourcekitd_requests { return sourcekitd.requests }
@@ -197,13 +200,27 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
   }
 
   /// Returns the updated lexical tokens for the given `snapshot`.
+  ///
+  /// - Parameters:
+  ///   - edits: If we are in the context of editing the contents of a file, i.e. calling ``SwiftLanguageServer/changeDocument(_:)``, we should pass `edits` to enable incremental parse. Otherwise, `edits` should be `nil`.
   private func updateSyntaxTree(
-    for snapshot: DocumentSnapshot
+    for snapshot: DocumentSnapshot,
+    with edits: ConcurrentEdits? = nil
   ) -> DocumentTokens {
     logExecutionTime(level: .debug) {
       var docTokens = snapshot.tokens
+      
+      var parseTransition: IncrementalParseTransition? = nil
+      if let previousTree = snapshot.tokens.syntaxTree,
+         let lookaheadRanges = snapshot.tokens.lookaheadRanges,
+         let edits {
+        parseTransition = IncrementalParseTransition(previousTree: previousTree, edits: edits, lookaheadRanges: lookaheadRanges, reusedNodeCallback: reusedNodeCallback)
+      }
+      let (tree, nextLookaheadRanges) = Parser.parseIncrementally(
+        source: snapshot.text, parseTransition: parseTransition)
 
-      docTokens.syntaxTree = Parser.parse(source: snapshot.text)
+      docTokens.syntaxTree = tree
+      docTokens.lookaheadRanges = nextLookaheadRanges
 
       return docTokens
     }
@@ -527,27 +544,36 @@ extension SwiftLanguageServer {
 
   public func changeDocument(_ note: DidChangeTextDocumentNotification) {
     let keys = self.keys
+    var edits: [IncrementalEdit] = []
 
     self.queue.async {
       var lastResponse: SKDResponseDictionary? = nil
 
-      let snapshot = self.documentManager.edit(note) { (before: DocumentSnapshot, edit: TextDocumentContentChangeEvent) in
+      let snapshot = self.documentManager.edit(note) {
+        (before: DocumentSnapshot, edit: TextDocumentContentChangeEvent) in
         let req = SKDRequestDictionary(sourcekitd: self.sourcekitd)
         req[keys.request] = self.requests.editor_replacetext
         req[keys.name] = note.textDocument.uri.pseudoPath
 
         if let range = edit.range {
-          guard let offset = before.utf8Offset(of: range.lowerBound), let end = before.utf8Offset(of: range.upperBound) else {
+          guard let offset = before.utf8Offset(of: range.lowerBound),
+            let end = before.utf8Offset(of: range.upperBound)
+          else {
             fatalError("invalid edit \(range)")
           }
 
+          let length = end - offset
           req[keys.offset] = offset
-          req[keys.length] = end - offset
+          req[keys.length] = length
 
+          edits.append(IncrementalEdit(offset: offset, length: length, replacementLength: edit.text.utf8.count))
         } else {
           // Full text
+          let length = before.text.utf8.count
           req[keys.offset] = 0
-          req[keys.length] = before.text.utf8.count
+          req[keys.length] = length
+
+          edits.append(IncrementalEdit(offset: 0, length: length, replacementLength: edit.text.utf8.count))
         }
 
         req[keys.sourcetext] = edit.text
@@ -556,7 +582,7 @@ extension SwiftLanguageServer {
         self.adjustDiagnosticRanges(of: note.textDocument.uri, for: edit)
       } updateDocumentTokens: { (after: DocumentSnapshot) in
         if lastResponse != nil {
-          return self.updateSyntaxTree(for: after)
+          return self.updateSyntaxTree(for: after, with: ConcurrentEdits(fromSequential: edits))
         } else {
           return DocumentTokens()
         }
--- a/Tests/SourceKitLSPTests/LocalSwiftTests.swift
+++ b/Tests/SourceKitLSPTests/LocalSwiftTests.swift
@@ -16,6 +16,7 @@ import LSPTestSupport
 import SKTestSupport
 import SourceKitLSP
 import XCTest
+import SwiftSyntax
 
 // Workaround ambiguity with Foundation.
 typealias Notification = LanguageServerProtocol.Notification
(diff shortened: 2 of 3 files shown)


Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -119,6 +119,10 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   var commandsByFile: [DocumentURI: SwiftCompileCommand] = [:]
 
+  var documentParseTransition: [DocumentURI: (previousTree: SourceFileSyntax, collector: IncrementalParseNodeAffectRangeCollector)] = [:]
```
ahoppen, line 122: We already store the tree in `DocumentSnapshot`. I think the cleaner design would be to also store the `IncrementalParseNodeAffectRangeCollector` there instead of adding this dictionary to `SwiftLanguageServer`.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -119,6 +119,10 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
...
+  
+  public var documentReusedNodeCollector: [DocumentURI: IncrementalParseReusedNodeCollector] = [:]
+
```
ahoppen, line 125: We shouldn’t be recording which nodes got re-used nodes unless it’s needed for testing purposes. I would suggest that we change `IncrementalParseReusedNodeCollector` from a delegate to a closure as I commented in https://github.com/apple/swift-syntax/pull/1685/files#r1254465986. Then this one could be an optional closure as well and you need to explicitly set it in the test cases to make it be called.

I would also add a comment that it’s for testing purposes only.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -198,13 +202,31 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   /// Returns the updated lexical tokens for the given `snapshot`.
   private func updateSyntaxTree(
-    for snapshot: DocumentSnapshot
+    for snapshot: DocumentSnapshot,
+    with edits: [IncrementalEdit]? = nil
```
ahoppen, line 206: Why not make this parameter `ConcurrentEdits?` instead of `[IncrementalEdit]?`. That way the caller can decide whether the edits are sequential or concurrent.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -520,34 +542,46 @@ extension SwiftLanguageServer {
...
             fatalError("invalid edit \(range)")
           }
 
+          let length = end - offset
           req[keys.offset] = offset
-          req[keys.length] = end - offset
+          req[keys.length] = length
 
+          edits.append(IncrementalEdit(offset: offset, length: length, replacementLength: edit.text.utf8.count)
+          )
```
ahoppen, line 577:
```suggestion
          edits.append(IncrementalEdit(offset: offset, length: length, replacementLength: edit.text.utf8.count))
```

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -520,34 +542,46 @@ extension SwiftLanguageServer {
...
       self.commandsByFile[uri] = nil
       self.currentDiagnostics[uri] = nil
+      self.documentParseTransition[uri] = nil
+      self.documentReusedNodeCollector[uri] = nil
 
       _ = try? self.sourcekitd.sendSync(req)
     }
   }
 
   public func changeDocument(_ note: DidChangeTextDocumentNotification) {
```
ahoppen, line 552: Have you checked whether the LSP edits are sequential or concurrent? I’m not sure if the spec defines it so you might need to just check what kind of edits VSCode sends when you do a multi-cursor edit.

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1476,4 +1476,49 @@ final class LocalSwiftTests: XCTestCase {
...
+      log("Received diagnostics for text edit - semantic")
+    })
+    
+    
+    let swiftLanguageServer = connection.server!._languageService(for: uri, .swift, in: connection.server!.workspaceForDocumentOnQueue(uri: uri)!) as! SwiftLanguageServer
+    
+    guard let nodes = swiftLanguageServer.documentReusedNodeCollector[uri]?.nodes else {
+      XCTFail("Can not find reused nodes in \(uri)")
+      return
+    }
```
ahoppen, line 1512: This seems racy to me. IIRC the `SwiftLanguageServer` executes `updateSyntaxTree` asynchronously when receiving the note and thus might not have performed the re-parse when you are getting the re-used nodes. This should be easy to fix if you wait for the first `PublishDiagnosticsNotification`.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
StevenWong12: The [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didChangeTextDocumentParams) indicates that the edits are sequential IIUC.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
StevenWong12: I've added a `*For Testing*` comment there.

I will sync this design change here when I change the delegate to a closure in `SwiftParser`.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -198,13 +203,32 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   /// Returns the updated lexical tokens for the given `snapshot`.
   private func updateSyntaxTree(
-    for snapshot: DocumentSnapshot
+    for snapshot: DocumentSnapshot,
+    with edits: ConcurrentEdits? = nil
   ) -> DocumentTokens {
```
ahoppen, line 208: Could you document what `edits = nil` means?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -119,6 +119,11 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   var commandsByFile: [DocumentURI: SwiftCompileCommand] = [:]
 
+  var documentParseTransition: [DocumentURI: LookaheadRanges] = [:]
```
ahoppen, line 122: The variable is now incorrectly named. It doesn’t contain any parse transitions.

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1476,4 +1476,55 @@ final class LocalSwiftTests: XCTestCase {
...
+      version: 0,
+      text: """
+      func foo() {
+      }
+      class bar {
+      }
+      """
+    )))
+    
+    let didChangeTextDocumentExpectation = XCTestExpectation(description: "didChangeTextDocument")
```
ahoppen, line 1498: You can do `self.expectation` here. That way we will get an error if you don’t wait for the expectation

```suggestion
    let didChangeTextDocumentExpectation = self.expectation(description: "didChangeTextDocument")
```

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1476,4 +1476,55 @@ final class LocalSwiftTests: XCTestCase {
...
+    sk.sendNoteSync(DidChangeTextDocumentNotification(textDocument: .init(uri, version: 1), contentChanges: [
+      .init(range: Range(Position(line: 2, utf16index: 7)), text: "a"),
+    ]), { (note: LanguageServerProtocol.Notification<PublishDiagnosticsNotification>) -> Void in
+      log("Received diagnostics for text edit - syntactic")
+      didChangeTextDocumentExpectation.fulfill()
+    }, { (note: LanguageServerProtocol.Notification<PublishDiagnosticsNotification>) -> Void in
+      log("Received diagnostics for text edit - semantic")
+    })
+    
+    let result = XCTWaiter.wait(for: [didChangeTextDocumentExpectation], timeout: defaultTimeout)
```
ahoppen, line 1508: And if you’re using `self.expectation`, you should also be using `self.wait`. IIRC it will fail the test if the expectation isn’t fulfilled, so you don’t need to check `result` either.

```suggestion
    let result = self.wait(for: [didChangeTextDocumentExpectation], timeout: defaultTimeout)
```

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1476,4 +1476,55 @@ final class LocalSwiftTests: XCTestCase {
...
+    if result != .completed {
+      fatalError("error \(didChangeTextDocumentExpectation) waiting for changing text document notification")
+    }
+    
+    let swiftLanguageServer = connection.server!._languageService(for: uri, .swift, in: connection.server!.workspaceForDocumentOnQueue(uri: uri)!) as! SwiftLanguageServer
+
+    guard let nodes = swiftLanguageServer.documentReusedNodeCollector[uri]?.nodes else {
+      XCTFail("Can not find reused nodes in \(uri)")
+      return
+    }
```
ahoppen, line 1518:
```suggestion
    let nodes = try XCTUnwrap(swiftLanguageServer.documentReusedNodeCollector[uri]?.nodes)
```

ahoppen (commented):
Just some minor comments.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -119,6 +119,11 @@ public final class SwiftLanguageServer: ToolchainLanguageServer {
 
   var commandsByFile: [DocumentURI: SwiftCompileCommand] = [:]
 
+  var documentLookaheadRanges: [DocumentURI: LookaheadRanges] = [:]
```
ahoppen, line 122: I think we should store the `LookaheadRanges` in the `DocumentSnapshot` as I mentioned here: https://github.com/apple/sourcekit-lsp/pull/761#discussion_r1255417400

Or is there a reason why we can’t do that?

I noticed this just now again by trying to write a doc comment for it and I couldn’t come up with a very nice explanation for it.

StevenWong12: Ah, seems I mis-understood that comment.

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -24,6 +25,13 @@ public struct DocumentSnapshot {
   /// the tokens are updated independently and only used internally.
   public var tokens: DocumentTokens
 
+  /// This information is used to determine whether a syntax node can be re-used in incremental parsing.
+  ///
+  /// The property is not nil only after the document is parsed.
+  public var lookaheadRanges: LookaheadRanges? {
+    document.latestLookaheadRanges
+  }
+
```
ahoppen, line 34: We also need to store the `LookaheadRanges` in the `DocumentSnapshot`. At the moment, you could get into the following race condition: 

- We get a document snapshot and which stores its syntax tree (in `DocumentSnapshot.tokens` in `SwiftLanguageServer.changeDocument`)
- Now an edit is processed, which modifies the `Document` that’s underlying the `DocumentSnapshot`
- If we access `snapshot.lookaheadRanges`, we now get the `LookaheadRanges` from the updated document, which is out-of-sync with the syntax tree stored in the `DocumentSnapshot` and the text stored in the `DocumentSnapshot`

I would just store the `LookaheadRanges` in `DocumentTokens`. That way it’s stored right next to the syntax tree they belong to and you also don’t need the `updateLookaheadRanges` method to update them independently of the syntax tree.

---

Edit: On second thought, the race condition won’t actually happen because `updateSyntaxTree` is always executed on `SwiftLangaugeServer.queue` but still, we shouldn’t be relying on that invariant.

StevenWong12: I think storing `lookaheadRanges` in `DocumentTokens` is a good idea. 👍

ahoppen:
@swift-ci Please test

Issue #707, opened 2023-02-08
SourceKit-LSP overwriting results of `swift package update`


adam-fowler:
If I run a `swift package update` on a project that has SourceKit-LSP running on it. The results of the update are overwritten by SourceKit-LSP.

If I have a dependency on a package `from: 1.0.0` and I currently have version `1.0.1` but a `1.0.2` has been released when I run `swift package update` I would expect it to update to using version `1.0.2`. It appears when SourceKit-LSP is running the version updates to `1.0.2` and then is reverted back to `1.0.1`.

This is with swift 5.7 and running inside VSCode.

ahoppen:
rdar://105173375

PR #762, opened 2023-07-07
Don’t update Package.resolved from sourcekit-lsp

ahoppen:
`SwiftPMWorkspace.reloadPackage` called `loadPackageGraph`, which causes package resolution and could thus update `Package.resolved`. This caused race conditions when `swift package update` was run from terminal while sourcekit-lsp is running since sourcekit-lsp was notified about the file changes and thus reloaded the pacakge to get build settings for the modified files.

Set `forceResolvedVersions` to only resolve packages based on the versions in `Package.resolved`, eliminating this problem.

Fixes https://github.com/apple/sourcekit-lsp/issues/707 rdar://105173375


bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #763, opened 2023-07-07
[5.9] Don’t update Package.resolved from sourcekit-lsp

ahoppen:
* **Explanation**: There was a race condition when running `swift package update` while sourcekit-lsp was running that could cause `Package.resolved` to not get updated. Fix the issue by not writing an updated Package.resolved from sourcekit-lsp
* **Risk**: Low
* **Testing**: Verified locally that Package.resolved does not get updated with this change
* **Issue**: rdar://105173375 / https://github.com/apple/sourcekit-lsp/issues/707
* **Reviewer**:  @bnbarham


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #766, opened 2023-07-11
Disable 'testSystemModuleInterface'

ahoppen:
This test is flaky. Disable it until we determined the root cause.


ahoppen:
@swift-ci Please test

PR #770, opened 2023-07-17
Don’t offer text edits from inlay hints if they are syntactically invalid

### LINKED ISSUE #759: InlayHint textEdits are not always correct
The text edits supplied with InlayHints don't always provide code that will compile eg adding the inlay hint to a closure argument would generate invalid code
```swift
func test(_ process: (String) -> ()) {
    process("hello")
}
test { string in 
    print(string)
}
```
The inlay hint for string is `: String`. The text edit that goes along with it is the same and inserting it would generate the following code

```swift
test { string: String in
    print(string)
}
```
which does not compile.

Another example is enums with associated values
```swift
case .test(let value):
```
becomes
```swift
case .test(let value: (String)):
```

Maybe we should disable the text edit option until we can be sure it returns valid code.

See https://github.com/swift-server/vscode-swift/issues/554 where this was originally reported

ahoppen:
Don’t offer any text edits for inlay hints displayed in closure paramters or enum case items.

Fixes #759
rdar://111559715

### DIFF
--- a/Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
+++ b/Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
@@ -13,6 +13,35 @@
 import Dispatch
 import LanguageServerProtocol
 import SourceKitD
+import SwiftSyntax
+
+fileprivate extension TokenSyntax {
+  /// Returns `false` if it is known that this token can’t be followed by a type
+  /// annotation.
+  var canBeFollowedByTypeAnnotation: Bool {
+    var node = Syntax(self)
+    LOOP: while let parent = node.parent {
+      switch parent.kind {
+      case .caseItem, .closureParam:
+        // case items (inside a switch) and closure parameters can’t have type
+        // annotations.
+        return false
+      case .codeBlockItem, .memberDeclListItem:
+        // Performance optimization. If we walked the parents up to code block item,
+        // we can’t enter a case item or closure param anymore. No need walking
+        // the tree any further.
+        break LOOP
+      default:
+        break
+      }
+      node = parent
+    }
+
+    // By default, assume that the token can be followed by a type annotation as
+    // most locations that produce a variable type info can.
+    return true
+  }
+}
 
 /// A typed variable as returned by sourcekitd's CollectVariableType.
 struct VariableTypeInfo {
@@ -22,6 +51,9 @@ struct VariableTypeInfo {
   var printedType: String
   /// Whether the variable has an explicit type annotation in the source file.
   var hasExplicitType: Bool
+  /// Whether we should suggest making an edit to add the type annotation to the
+  /// source file.
+  var canBeFollowedByTypeAnnotation: Bool
 
   init?(_ dict: SKDResponseDictionary, in snapshot: DocumentSnapshot) {
     let keys = dict.sourcekitd.keys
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
```diff
@@ -34,10 +66,12 @@ struct VariableTypeInfo {
           let hasExplicitType: Bool = dict[keys.variable_type_explicit] else {
       return nil
     }
+    let tokenAtOffset = snapshot.tokens.syntaxTree?.token(at: AbsolutePosition(utf8Offset: offset))
 
     self.range = startIndex..<endIndex
     self.printedType = printedType
     self.hasExplicitType = hasExplicitType
+    self.canBeFollowedByTypeAnnotation = tokenAtOffset?.canBeFollowedByTypeAnnotation ?? true
```
bnbarham, line 74: Would a `false` default be better here?

Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
```diff
@@ -13,6 +13,35 @@
...
+  var canBeFollowedByTypeAnnotation: Bool {
+    var node = Syntax(self)
+    LOOP: while let parent = node.parent {
+      switch parent.kind {
+      case .caseItem, .closureParam:
+        // case items (inside a switch) and closure parameters can’t have type
+        // annotations.
+        return false
+      case .codeBlockItem, .memberDeclListItem:
+        // Performanc optimization. If we walked the parents up to code block item,
```
bnbarham, line 30:
```suggestion
        // Performance optimization. If we walked the parents up to code block item,
```

Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
```diff
@@ -13,6 +13,35 @@
...
+import SwiftSyntax
+
+fileprivate extension TokenSyntax {
+  /// Returns `false` if it is known that this token can’t be followed by a type
+  /// annotation.
+  var canBeFollowedByTypeAnnotation: Bool {
+    var node = Syntax(self)
+    LOOP: while let parent = node.parent {
+      switch parent.kind {
+      case .caseItem, .closureParam:
```
bnbarham, line 25: I was going to ask if this handles the case that the param is in `()`. But apparently it does, because `closureParam` is only the non-signature case. `x in` is a `ClosureParamListSyntax` where as `(x) in` is a `ClosureParameterClauseSyntax`. The former has `ClosureParam`s and the latter `ClosureParameter`s 😅.

Not something that needs fixing here, but worth knowing. I don't have a great name for the former case 🤔.

bnbarham (approved):

Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
ahoppen: Not really. `canBeFollowedByTypeAnnotation` is a blacklist-based method that also defaults to `true`. That’s why I think we should also default to `true` here for consistency.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #772, opened 2023-07-18
Add support for Musl libc with `canImport(Musl)` checks

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

Musl is a low footprint libc that's used in Linux distributions such as Alpine Linux, which allows producing fairly small container images. Additionally, unlike Glibc, musl allows full static linking, meaning apps can be easily distributed to an arbitrary Linux distribution that may have a version of Glibc incompatible with the one that Swift is usually built with or no Glibc installed at all.

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/DisableSigpipe.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/DisableSigpipe.swift
@@ -13,13 +13,15 @@
 
 #if canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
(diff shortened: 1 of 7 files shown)


al45tair:
@swift-ci Please test Windows platform

ahoppen (approved):

MaxDesiatov:
@swift-ci test windows

PR #773, opened 2023-07-18
[NFC] Use `SwiftSDK` instead of deprecated `Destination`

MaxDesiatov:
Code utilizing libSwiftPM should use `SwiftSDK` instead, with a few other properties renamed for clarity on related types. In this PR specifically we should use new `targetTriple` instead of the deprecated `triple` on the `Toolchain` type.

Depends on https://github.com/apple/swift-package-manager/pull/6707.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -119,8 +119,8 @@ public final class SwiftPMWorkspace {
         throw Error.cannotDetermineHostToolchain
     }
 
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Are these just renames or does `SwiftSDK` signify something different than `Destionation`, but just happens to be used for the same thing here?

MaxDesiatov:
Plain renames, Swift SDK is the same thing that SwiftPM called a destination previously.

MaxDesiatov:
@swift-ci test windows

ahoppen:
OK, thanks for the clarification.

PR #775, opened 2023-07-19
[5.9] Update tests to allow multiple -F options

MaxDesiatov:
Cherry-pick of https://github.com/apple/sourcekit-lsp/pull/742 required to unblock https://github.com/apple/swift-package-manager/pull/6709.

Updates tests to allow compile arguments with multiple `-F` options such as `-F <foo> -Xcc -F -Xcc <foo>`. This change is required to support SwiftPM changes which forward C compiler flags to the Swift compiler in more cases.

### DIFF
--- a/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
+++ b/Tests/SKSwiftPMWorkspaceTests/SwiftPMWorkspaceTests.swift
@@ -123,7 +123,7 @@ final class SwiftPMWorkspaceTests: XCTestCase {
       let versionString = PackageModel.Platform.macOS.oldestSupportedVersion.versionString
       check("-target", hostTriple.tripleString(forPlatformVersion: versionString), arguments: arguments)
       check("-sdk", arguments: arguments)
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

ahoppen (approved):

bnbarham:
Just to be clear, is there ever any `-F foo -F other` or is it just to accommodate `-Xcc -F -Xcc foo`? I'm okay taking this to unblock things either way, but we could handle that better rather than just arbitrarily allowing any number. The original PR (https://github.com/apple/sourcekit-lsp/pull/28) is more about there being a single `-target` though, so maybe allowing any number of `-F` is still fine 🤷

rauhul:
The change is needed because `-F` flags can arise in a number of places in SPM and it is not valid for sourcekit to assume only a single occurrence. In this case `-F` now appears twice, once for `swift-frontend` as `-F foo` and once for `cc1` as `-Xcc -F -Xcc foo`.

rauhul (approved):

PR #784, opened 2023-07-25
Uppercase first letter of diagnostics

ahoppen:
sourcekitd returns diagnostics with the first letter lowercase. Xcode, for example, shows the messages with the first letter uppercases. I think that looks nicer and we should also uppercase the first letter in sourcekit-lsp.

CC @tristanlabelle I noticed this in https://github.com/apple/swift/issues/67510. While it doesn’t change the messages to be imperative, having them start with an uppercase letter maybe makes them stick out a little less.

### DIFF
--- a/Sources/SourceKitLSP/Swift/Diagnostic.swift
+++ b/Sources/SourceKitLSP/Swift/Diagnostic.swift
@@ -125,6 +125,19 @@ extension TextEdit {
   }
 }
 
+fileprivate extension String {
+  /// Returns this string with the first letter uppercased.
+  ///
+  /// If the string does not start with a letter, no change is made to it.
+  func withFirstLetterUppercased() -> String {
+    if let firstLetter = self.first {
+      return firstLetter.uppercased() + self.dropFirst()
+    } else {
+      return self
+    }
+  }
+}
+
 extension Diagnostic {
 
   /// Creates a diagnostic from a sourcekitd response dictionary.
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

benlangmuir:
My concern here is that the first word in the diagnostic might be a keyword or identifier that shouldn't be modified. If this is preferred style, why not update the compiler diagnostics themselves?

ahoppen:
@swift-ci Please test

ahoppen:
I think on the command line we want to emit diagnostics that start with a lowercase letter to match clang. But in IDEs the errors show up more nicely if they start with an uppercase letter.

If the diagnostic starts with an identifier or keyword, that identifier/keyword should be wrapped in single quotes, which means that the quote would be the first character, so we wouldn’t uppercase the identifier or keyword.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #767, opened 2023-07-12
Show a progress indicator in the editor if SourceKit-LSP is reloading packages

ahoppen:
I noticed that the initial package loading can take ~5s. It’s good behavior to inform the client that sourcekit-lsp is busy reloading the package, showing the user that semantic functionality might not be ready yet.

https://github.com/apple/sourcekit-lsp/issues/620
rdar://111917300

Issue #620, opened 2022-08-29
Jump to definition requires a reload after build


adam-fowler:
If I load a project with dependencies into VS Code which has not been built when I right click on a symbol to go to definition (run a textDocument/definition request) it returns an empty result. This is a known issue.

If I then build the project and repeat I still get an empty result. I would assume once the project is built SourceKit-LSP should be able to return a position for the symbol.

I have to restart VS Code (ie restart the SourceKit-LSP server) for SourceKit-LSP to return a valid position for the symbol.

ahoppen:
rdar://99329579

adam-fowler:
This look like it is resolved, will re-open if I see it again

adam-fowler:
Re-opening saw it again

adam-fowler:
Actually it is just slow to respond. Is there a way we can indicate the LSP server is loading the index after a build, or whatever it is doing post build that means it doesn't have the built symbols available.

ahoppen:
I’m adding a progress indicator for package reloading in https://github.com/apple/sourcekit-lsp/pull/767. That’s all I can think of that might take a non-trivial amount of time and happens on the SourceKit-LSP side after the project has been modified.

ahoppen:
Oh, and now I’ve been able to reproduce the issue as well, independent of the work done progress for package loading.

The issue appears to be that we don’t get compiler arguments if the dependencies have not been checked out and we don’t reload the package after the dependencies are checked out.

Steps to reproduce:
- Set `SOURCEKIT_LOGGING=99` in the VSCode Swift plugin settings
- Clone the soucekit-lsp repository
- `swift package reset`
- Open the project in VSCode
- Open DocumentTokens.swift
- Notice that DocumentTokens.swift gets opened with compiler arguments
- Run a build
- Close and re-open DocumentTokens.swift
- Notice that it still gets opened without compiler arguments

ahoppen:
What’s happening is that after building, you need to make an edit to the file to get jump-to-definition working. Could that match what you’ve been seeing @adam-fowler?

On a technical level
- When we open the file, we build an AST for it. That AST has errors because the dependent modules haven’t been built yet.
- Now we build the dependent modules and the dependent module files exist
[...]

PR #791, opened 2023-08-04
Show a progress indicator in the editor if SourceKit-LSP is reloading packages

ahoppen:
This re-lands https://github.com/apple/sourcekit-lsp/pull/767 after it got reverted. The toolchain build failure should be fixed by adding a check if the client supports work done progress.

```swift
guard capabilityRegistry.clientCapabilities.window?.workDoneProgress ?? false else {
  // Client doesn’t support work done progress
  return
}
```

I noticed that the initial package loading can take ~5s. It’s good behavior to inform the client that sourcekit-lsp is busy reloading the package, showing the user that semantic functionality might not be ready yet.

https://github.com/apple/sourcekit-lsp/issues/620
rdar://112498447


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #800, opened 2023-08-17
Add a message to some test assertions

bnbarham:
The output when these assertions fail is just `XCTAssertTrue failed`, which is very unhelpful. Add in the strings we're comparing.


bnbarham:
@swift-ci please test

ahoppen (approved):

PR #801, opened 2023-08-18
Update assertions to add more detail

bnbarham:
`testSystemSwiftInterface` is called multiple times in a single test, add file/line to the test output so it's clear which case actually failed.

### DIFF
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -114,23 +114,25 @@ final class SwiftInterfaceTests: XCTestCase {
     _ testLoc: TestLocation, 
     ws: SKSwiftPMTestWorkspace, 
     swiftInterfaceFile: String, 
-    linePrefix: String
+    linePrefix: String,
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -114,7 +114,9 @@ final class SwiftInterfaceTests: XCTestCase {
     _ testLoc: TestLocation, 
     ws: SKSwiftPMTestWorkspace, 
     swiftInterfaceFile: String, 
-    linePrefix: String
+    linePrefix: String,
+    file: StaticString = #filePath,
+    line: UInt = #line
```
benlangmuir, line 119: There are more XCT assertions here that should probably take line/file:
* XCTFail in the guard else
* XCTUnwrap called twice below

bnbarham:
@swift-ci please test

bnbarham:

ahoppen:
We haven’t seen this failure ever since we stopped using the system-wide module cache.

PR #804, opened 2023-08-25
Improve folding of line comments

### LINKED ISSUE #803: issue from vscode-swift#587
issue from [vscode-swift#587](https://github.com/swift-server/vscode-swift/issues/587)

It seems that source-kit has provided some error values

ahoppen:
Previously, the folding range of a block of line comments was ended after the newline of the last line comment, which would usually be the line of the function’s start. That caused VS Code to not show a folding range for the function body itself since it started on the same line that the line comment folding range ended.

Rewrite the trivia folding logic so that the folding range of those line comments ends on the line of the last line comment.

Fixes #803
rdar://114428202


benlangmuir (approved):

ahoppen:
@swift-ci Please test

PR #810, opened 2023-09-03
Update nvim instructions in `Editors/README.md`

TyrfingMjolnir:
Added example that acutally works in `nvim` with basic syntax coloring and LSP feedback; yet this example lacks keymaps from above example.

### DIFF
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -188,6 +188,28 @@ vim.api.nvim_create_autocmd('LspAttach', {
 })
 ```
 
+while I was unable to make the example above work, below works for getting basic syntax coloring and feedback from LSP, yet lacks the keymaps mentioned above.
+
+```lua
+local swift_lsp = vim.api.nvim_create_augroup("swift_lsp", { clear = true })
+ vim.api.nvim_create_autocmd("FileType", {
+  pattern = { "swift" },
+  callback = function()
+   local root_dir = vim.fs.dirname(vim.fs.find({
+    "Package.swift",
+    ".git",
+   }, { upward = true })[1])
+   local client = vim.lsp.start({
+    name = "sourcekit-lsp",
+    cmd = { "sourcekit-lsp" },
+    root_dir = root_dir,
+   })
+   vim.lsp.buf_attach_client(0, client)
+  end,
+  group = swift_lsp,
+ })
+```
+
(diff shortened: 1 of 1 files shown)


etcwilde (commented):
So generally this looks like it should work, though I'm not sure that the autogroup/autocommand for the filetype is entirely necessary unless you're configuring the kemaps since the filetypes are set up in the default configuration for the sourcekit-lsp server in lspconfig; https://github.com/neovim/nvim-lspconfig/blob/4266f9bb36b4fb09edd19b67d95043cf7ff88ddf/lua/lspconfig/server_configurations/sourcekit.lua#L6.

I think adding `require('lspconfig').sourcekit.setup {}` to the "suggested" configuration in the `lspconfig` readme should actually get you pretty much this setup, along with the key-mappings, unless you like calling the functions manually.

lspconfig suggested configuration:
https://github.com/neovim/nvim-lspconfig/blob/master/README.md#suggested-configuration

SourceKit LSP server setup:
https://github.com/neovim/nvim-lspconfig/blob/4266f9bb36b4fb09edd19b67d95043cf7ff88ddf/lua/lspconfig/server_configurations/sourcekit.lua#L4-L8

I'm also a little worried about setting a precedent of making this a dumping ground for everyone's configs. We'll end up with an untenable list of various configurations without much rhyme or reason. I would prefer seeing if we can't figure out what is happening in your config and adjust the existing example. Just based on the config you've proposed here, it looks like you already have sourcekit-lsp on your `$PATH`, so you shouldn't need this `cmd = {'$TOOLCHAIN_PATH/usr/bin/sourcekit-lsp'}` configuration.

I've got my own config that's run through the `lazy.nvim` plugin manager here that sets up clangd and sourcekit-lsp;
https://github.com/etcwilde/dotfiles/blob/8f9b8f66f68c75542c2170a2a6ff3625b3788e1a/nvim/lua/plugins/lsp.lua#L17-L23 Technically I shouldn't need the `root_dir` config anymore either since the default config now has `compile_commands.json` and `buildServer.json`. These days I should only need to call `lspconfig.sourcekit.setup {}`.

ahoppen:
I’m going to close this PR for now. As @etcwilde has commented, I think it would be valuable if you could figure out why the exiting configuration didn’t work for you and adjust it as needed. If you do want to do that, feel free to re-open the PR or open a new one.

PR #814, opened 2023-09-11
Introduce a `relatedDependenciesBranch` variable

ahoppen:
This way we only need to update one value when branching for a new release of Swift.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #821, opened 2023-09-15
Fix typo in `reloadPackage` token name

zacharytamas:
I noticed that #791 introduced a misspelling of `SourceKit` in the `WorkDoneProgressState` token for `reloadPackage` so I corrected it.

This value did not seem to be explicitly covered in tests and given the statelessness (between server restarts) of the protocol I believe this doesn't have further ramifications.


benlangmuir (approved):

benlangmuir:
@swift-ci please test

ahoppen (approved):
Thanks

PR #824, opened 2023-09-19
Reduce usages of `try!` in tests

ahoppen:
The usage of `try!` was still a relict of when `XCTest` didn’t allow throwing test method.


ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test

PR #826, opened 2023-09-20
Remove `--sync` option from sourcekit-lsp

ahoppen:
AFAICT this isn’t used and I have never used it for debugging sourcekit-lsp since taking it over. Remove it because it won’t easily be possible to implement the guarding semaphore once we adopt Swift concurrency.


ahoppen:
@swift-ci Please test

ahoppen:
Oh, well. Looks like we need to keep it then… 😢

benlangmuir:
You could alternatively made the python code for that test more complex so that it waits for responses itself; not sure if that's the right tradeoff though.

ahoppen:
I’ll see if we can keep `--sync` around. If that does end up getting in the way of incorporating Swift concurrency cleanly, I think I’ll have to re-write the Python script.

PR #828, opened 2023-09-21
Don’t crash when unregistering for change notifications of a file that isn’t watched

ahoppen:
I believe this should fix the crash reported in https://forums.swift.org/t/vscode-with-5-9-stopping-server-failed/67397.

I am still investigating where the unbalanced open / close calls might be coming from, but in either case, we shouldn’t crash.


ahoppen:
@swift-ci Please test

bnbarham (approved):
Might be worth adding a warning here so we can catch bad things going on in logs.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #833, opened 2023-09-27
Make long test execution opt-out instead of opt-in

ahoppen:
This matches what swift-syntax is doing and ensures that new contributors run all tests by default.

### DIFF
--- a/Sources/SKTestSupport/LongTestsEnabled.swift
+++ b/Sources/SKTestSupport/LongTestsEnabled.swift
@@ -12,8 +12,8 @@
 
 import Foundation
 
-public var longTestsEnabled: Bool {
-  if let value = ProcessInfo.processInfo.environment["SOURCEKIT_LSP_ENABLE_LONG_TESTS"] {
+public var longTestsDisabled: Bool {
+  if let value = ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] {
     return value == "1" || value == "YES"
   }
   return false
--- a/Tests/SourceKitDTests/CrashRecoveryTests.swift
+++ b/Tests/SourceKitDTests/CrashRecoveryTests.swift
@@ -46,7 +46,7 @@ fileprivate extension HoverResponse {
 final class CrashRecoveryTests: XCTestCase {
   func testSourcekitdCrashRecovery() throws {
     try XCTSkipUnless(Platform.current == .darwin, "Linux and Windows use in-process sourcekitd")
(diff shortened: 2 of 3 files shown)


ahoppen:
@swift-ci Please test

bnbarham (commented):
How long are the long tests?

In terms of environment variable, I do actually prefer namespaced (`SOURCEKIT_LSP_*`).

ahoppen:
They increase parallel testing from 18s to 40s, IIRC. So, quite significant.

> In terms of environment variable, I do actually prefer namespaced (`SOURCEKIT_LSP_*`).

Which other long tests should it be skipping? I actually like having the exact same name for swift-syntax and sorucekit-lsp. Plus, it’s shorter to type without the prefix.

bnbarham:
> They increase parallel testing from 18s to 40s, IIRC. So, quite significant.

Oof. Still, probably fine since if you're actually working on it often you can add the environment variable.

> Which other long tests should it be skipping?

Who knows, people don't necessarily just work on our projects 😅. It could be that some other random project uses an environment variable with the same name too. Plus if I had a random environment variable in my `.*rc` and it was called `SKIP_LONG_TESTS` I would later wonder what on earth it's for (yeah yeah I could add a comment for it).

ahoppen:
> Oof. Still, probably fine since if you're actually working on it often you can add the environment variable.

Yes, it’s not that bad that we definitely always want to skip them. Skipping them is really just an optimization for faster local iteration times if you know what you’re doing IMO.

> Who knows, people don't necessarily just work on our projects 😅. It could be that some other random project uses an environment variable with the same name too. Plus if I had a random environment variable in my `.*rc` and it was called `SKIP_LONG_TESTS` I would later wonder what on earth it's for (yeah yeah I could add a comment for it).

Hmm, I never considered setting them as environment variable in my shell profile. Is that something you would do? What we suggest is that you either set them as `SKIP_LONG_TESTS=1 swift test` or in the Xcode scheme and in both of these cases it’s not ambiguous at all.

bnbarham (approved):
> Hmm, I never considered setting them as environment variable in my shell profile. Is that something you would do?

No, but who knows what others have done in other projects. It's not the end of the world for me either way since it doesn't really matter if they're skipped or not. I just generally prefer namespaced environment variables :). If you strongly prefer without, feel free to keep.

ahoppen:
Let’s keep without until it causes real-world ambiguities or issues.

PR #836, opened 2023-09-28
Migrate `BuildSystemManager` and all the build systems to be actors

ahoppen:
This is the last big piece for actorification of the codebase. All follow-up PRs should be about cleaning things up.

### DIFF
--- a/Sources/SKCore/BuildServerBuildSystem.swift
+++ b/Sources/SKCore/BuildServerBuildSystem.swift
@@ -43,10 +43,7 @@ func executable(_ name: String) -> String {
 ///
 /// Provides build settings from a build server launched based on a
 /// `buildServer.json` configuration file provided in the repo root.
-public final class BuildServerBuildSystem: MessageHandler {
-  /// The handler's request queue. `delegate` will always be called on this queue.
-  public let queue: DispatchQueue = DispatchQueue(label: "language-server-queue", qos: .userInitiated)
-
+public actor BuildServerBuildSystem: MessageHandler {
   let projectRoot: AbsolutePath
   let buildFolder: AbsolutePath?
   let serverConfig: BuildServerConfig
@@ -73,7 +70,7 @@ public final class BuildServerBuildSystem: MessageHandler {
   /// The build settings that have been received from the build server.
   private var buildSettings: [DocumentURI: FileBuildSettings] = [:]
 
-  public init(projectRoot: AbsolutePath, buildFolder: AbsolutePath?, fileSystem: FileSystem = localFileSystem) throws {
+  public init(projectRoot: AbsolutePath, buildFolder: AbsolutePath?, fileSystem: FileSystem = localFileSystem) async throws {
     let configPath = projectRoot.appending(component: "buildServer.json")
     let config = try loadBuildServerConfig(path: configPath, fileSystem: fileSystem)
 #if os(Windows)
@@ -95,12 +92,11 @@ public final class BuildServerBuildSystem: MessageHandler {
   /// Creates a build system using the Build Server Protocol config.
   ///
   /// - Returns: nil if `projectRoot` has no config or there is an error parsing it.
-  public convenience init?(projectRoot: AbsolutePath?, buildSetup: BuildSetup)
-  {
+  public init?(projectRoot: AbsolutePath?, buildSetup: BuildSetup) async {
     if projectRoot == nil { return nil }
 
     do {
-      try self.init(projectRoot: projectRoot!, buildFolder: buildSetup.path)
+      try await self.init(projectRoot: projectRoot!, buildFolder: buildSetup.path)
     } catch _ as FileSystemError {
       // config file was missing, no build server for this workspace
       return nil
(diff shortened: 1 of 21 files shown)


ahoppen:
@swift-ci Please test

Sources/SKCore/CompilationDatabaseBuildSystem.swift
```diff
@@ -100,42 +90,32 @@ extension CompilationDatabaseBuildSystem: BuildSystem {
   public func buildSettings(for document: DocumentURI, language: Language) async throws -> FileBuildSettings? {
     // FIXME: (async) Convert this to an async function once `CompilationDatabaseBuildSystem` is an actor.
     return await withCheckedContinuation { continuation in
-      self.queue.async {
-        continuation.resume(returning: self.settings(for: document))
-      }
+      continuation.resume(returning: self.settings(for: document))
     }
```
hamishknight, line 94: Not sure if this gets cleaned up in a future commit, but just in case it isn't, the `withCheckedContinuation` could be removed here :)

hamishknight: Also seems like the comment is now outdated

Sources/SKCore/BuildServerBuildSystem.swift
```diff
@@ -171,13 +167,11 @@ public final class BuildServerBuildSystem: MessageHandler {
...
-        self.handleBuildTargetsChanged(Notification(params, clientID: clientID))
-      } else if let params = params as? FileOptionsChangedNotification {
-        self.handleFileOptionsChanged(Notification(params, clientID: clientID))
-      }
+  public func handle(_ params: some NotificationType, from clientID: ObjectIdentifier) async {
+    if let params = params as? BuildTargetsChangedNotification {
+      await self.handleBuildTargetsChanged(Notification(params, clientID: clientID))
+    } else if let params = params as? FileOptionsChangedNotification {
+      await self.handleFileOptionsChanged(Notification(params, clientID: clientID))
     }
```
hamishknight, line 175: Same question regarding waiting for the delegate method to complete, looks like conformers of BuildSystemDelegate end up dispatching a `Task` asynchronously, so I don't think it's an issue, just wondering if it's something guaranteed by BuildSystemDelegate (or whether it even matters)

hamishknight (approved):

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -181,35 +183,31 @@ extension BuildSystemManager {
...
-         let delegate = self._delegate {
-        let change = self.convert(change: mainChange, ofMainFile: mainFile, to: uri)
-        self.notifyQueue.async {
-          delegate.fileBuildSettingsChanged([uri: change])
-        }
-      }
+    if let mainChange = newStatus.buildSettingsChange,
+       let delegate = self._delegate {
+      let change = self.convert(change: mainChange, ofMainFile: mainFile, to: uri)
+      await delegate.fileBuildSettingsChanged([uri: change])
```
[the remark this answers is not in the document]
ahoppen: Yes, that’s intentional. I checked that none of these methods are doing a huge amount of work. I decided to await their results here because otherwise we cannot guarantee that the callbacks will be called in order (if tasks aren’t executed in the order they were created).

Sources/SourceKitLSP/Sequence+AsyncMap.swift
```diff
@@ -0,0 +1,42 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Sequence {
+  /// Just like `Sequence.map` but allows an `async` transform function.
+  func asyncMap<T>(
+    _ transform: (Element) async throws -> T
+  ) async rethrows -> [T] {
+    var result: [T] = []
```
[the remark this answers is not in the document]
ahoppen: Good idea 👍🏽

Sources/SKCore/CompilationDatabaseBuildSystem.swift
ahoppen: Good catch. Thanks. I would have probably caught it sooner or later by fixing the TODO but we can do it now already.

Sources/SKCore/BuildServerBuildSystem.swift
ahoppen: Once `BuildSystemDelegate` has been asyncified, the implementers of `BuildSystemDelegate` will no longer create tasks. Again, the reason for awaiting here are the ordering concerns. We can revisit whether ordering is relevant once the functions in `BuildSystemManager` have been re-designed to no longer be order-dependent: https://github.com/ahoppen/sourcekit-lsp/commit/e22d1a4965875391b6b9297137761cce41a5081b

Sources/SKCore/BuildSystemManager.swift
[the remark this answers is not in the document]
hamishknight: Makes sense, thanks

ahoppen:
@swift-ci Please test

PR #839, opened 2023-09-29
Move semaphore signaling to block main thread until request has received a reply instead of until it has been handled

ahoppen:
This was causing the sourcekit-lsp integration test to fail.


bnbarham (approved):

bnbarham:
@swift-ci please test

PR #841, opened 2023-09-30
Keep track of build settings in `BuildSystemManager` instead of `SourceKitServer`

ahoppen:
After two minor preliminary commits, there are two core commits here that go hand-in-hand and significantly simplify the build settings handling.

### Change the build system to only notify delegate about changed files, not about new build settings

This defines away an entire class of data races if delegate callbacks are delivered out-of-order. If we aren’t providing the new build settings in the delegate callback, then it doesn’t matter if two `fileBuildSettingsChanged` calls change order since they don’t carry any state. Instead of storing the build settings returned by the delegate call, the language server needs to ask the `BuildSystemManager` for the current build settings of the file.

### Remove tracking of file build settings status in `SourceKitServer` and `BuildSystemManager`

The core idea here is that the toolchain language servers always call into `BuildSystemManager` and `BuildSystemManager` will always reply with build settings. If it hasn’t computed them yet, it will reply with fallback settings.

With that assumption, we can remove the `documentToPendingQueue` from `SourceKitServer` since there are no longer any documents that are pending – everything has a build settings immediately.

Similarly, `BuildSystemManager.mainFileStatuses` also isn’t needed anymore.

And lastly, since we know that `BuildSystemManager.buildSettings` will always return a value `registerForChangeNotifications` is changed not call `fileBuildSettingsChanged` immediately. Instead, it will only cause `fileBuildSettingsChanged` to be called when the file’s build settings change after the `registerForChangeNotifications` call.

### DIFF
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -89,9 +89,6 @@ public actor BuildSystemManager {
   /// The set of watched files, along with their main file and language.
   var watchedFiles: [DocumentURI: (mainFile: DocumentURI, language: Language)] = [:]
 
-  /// Statuses for each main file, containing build settings from the build systems.
-  var mainFileStatuses: [DocumentURI: MainFileStatus] = [:]
-
   /// The underlying primary build system.
   let buildSystem: BuildSystem?
 
@@ -208,11 +205,8 @@ extension BuildSystemManager {
       self.watchedFiles[uri] = (mainFile, language)
     }
 
-    let newStatus = await self.cachedStatusOrRegisterForSettings(for: mainFile, language: language)
-
-    if newStatus.buildSettingsChange != nil, let delegate = self._delegate {
-      await delegate.fileBuildSettingsChanged([uri])
-    }
+    await buildSystem?.registerForChangeNotifications(for: mainFile, language: language)
+    fallbackBuildSystem?.registerForChangeNotifications(for: mainFile, language: language)
   }
 
   /// Return settings for `file` based on  the `change` settings for `mainFile`.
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -208,11 +205,8 @@ extension BuildSystemManager {
       self.watchedFiles[uri] = (mainFile, language)
     }
 
-    let newStatus = await self.cachedStatusOrRegisterForSettings(for: mainFile, language: language)
-
-    if newStatus.buildSettingsChange != nil, let delegate = self._delegate {
-      await delegate.fileBuildSettingsChanged([uri])
-    }
+    await buildSystem?.registerForChangeNotifications(for: mainFile, language: language)
+    fallbackBuildSystem?.registerForChangeNotifications(for: mainFile, language: language)
```
hamishknight, line 209: This is now a no-op, should it be removed? Or is there value in keeping it?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -270,18 +256,7 @@ public actor SourceKitServer {
...
-      await requestHandler(request, workspace, languageService)
-      return
-    }
-
-    // Not ready to handle it, we'll queue it and handle it later.
-    self.documentToPendingQueue[doc, default: DocumentNotificationRequestQueue()].add(operation: {
-      await requestHandler(request, workspace, languageService)
-    }, cancellationHandler: {
-      request.reply(fallback)
-    })
```
hamishknight, line 284: 🎉

hamishknight (approved):
Nice!

Sources/SKCore/BuildSystemManager.swift
ahoppen: Oh, good point. I just made `FallbackBuildSystem` not conform to `BuildSystem` and then we can remove `registerForChangeNotifications` there altogether.

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -183,6 +183,28 @@ extension BuildSystemManager {
...
+  /// If the document doesn't have builds settings by itself, eg. because it is
+  /// a C header file, the build settings will be inferred from the primary main
+  /// file of the document. In practice this means that we will compute the build
+  /// settings of a C file that includes the header and replace any file
+  /// references to that C file in the build settings by the header file.
+  public func buildSettingsInferredFromMainFile(
+    for document: DocumentURI,
+    language: Language
+  ) async -> (buildSettings: FileBuildSettings, isFallback: Bool)? {
+    if let mainFile = mainFilesProvider?.mainFilesContainingFile(document).first {
```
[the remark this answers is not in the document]
ahoppen: It turned out that this was a good area to refactor and it got a little bit larger and more involved. I will open a follow-up PR for it.

PR #845, opened 2023-09-30
Add an overload of `firstNonNil` that allows the default value to be `nil`

ahoppen:
Without this overload, `T` was inferred to be `Something?`, thus the first parameter was inferred to be `Something??` and the first parameter was always wrapped in an optional, effectively making `if let optional` never be hit.


ahoppen:
@swift-ci Please test

ahoppen:
Closing this because I’ve included it in the https://github.com/apple/sourcekit-lsp/tree/ahoppen/asyncification branch, in which I’m collecting all the asyncification changes, which will then be merged into `main` in one go.

PR #846, opened 2023-09-30
Revert asyncificaiton changes

ahoppen:
The asyncification changes caused some non-deterministic test failures. I believe that some of these are due to race conditions that are the result of the partial transition to actors.

Instead of merging the asyncification piece by piece, I will collect the changes asyncification changes in a branch and then qualify that branch througougly (running CI multiple times) before merging it into `main`.

This resets the state of `main` back to 73bbba385a5d7b45d9f7659a45c5b22a9bb29da8, which was the last commit before all the asyncification changes. Since then, only asyncification PRs have been made to `main`.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #847, opened 2023-10-02
Refactor the computation of main files

ahoppen:
The motivating change for this was to deterministically pick a main file for a header file instead of picking the first element in a set, which is not deterministic.

While doing this, I also changed the main file computation to not carry any state about previous main files around. If a header file is associated with a main file b.cpp and a new a.cpp gets added that imports the header as well, we should be using a.cpp for the build settings. That way we will get the same build settings if we close and re-open the project.

And this was a good opportunity to refactor some of the main file handling into smaller, more dedicated functions.

---

Also includes the changes I made to https://github.com/apple/sourcekit-lsp/pull/841#discussion_r1343182026 because I forgot to push those changes before hitting merge.

### DIFF
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -120,7 +120,6 @@ public actor BuildSystemManager {
 
   public func filesDidChange(_ events: [FileEvent]) async {
     await self.buildSystem?.filesDidChange(events)
-    self.fallbackBuildSystem?.filesDidChange(events)
   }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/SKCore/BuildSystemManager.swift
```diff
@@ -329,28 +330,63 @@ extension BuildSystemManager: MainFilesDelegate {
...
+    }
+    guard let mainFilesProvider else {
+      return uri
+    }
+
+    let mainFiles = mainFilesProvider.mainFilesContainingFile(uri)
+    if mainFiles.contains(uri) {
+      // If the main files contain the file itself, prefer to use that one
+      return uri
+    } else if let mainFile = mainFiles.sorted(by: { $0.pseudoPath < $1.pseudoPath }).first {
```
hamishknight, line 384: Nit: This could be better written as `min(by: { $0.pseudoPath < $1.pseudoPath })`

hamishknight (approved):
Much nicer!

Sources/SKCore/BuildSystemManager.swift
ahoppen: Nice. I didn’t know about `min(by:)` 👍🏽

PR #849, opened 2023-10-03
Asynchronously return the request result for folding range requests

ahoppen:
This is the last infrastructural change to asyncify sourcekit-lsp, everything afterwards should be fairly localized fixes for `FIXME: (async)` comments and the translation of all the other requests to return results asynchronously as a return value.

The basic idea is that we change `Connection` to be a class with synchronous functions again. This ensures that all messages within the connection are handles sequentially. `SourceKitServer` is responsible for transitioning into the Swift concurrency world by having an `AsyncQueue`. That queue is concurrent, which means that requests can be served out-of-order. To ensure that we don’t get out-of-order opens/edits/closes (or any other kind of message that modifies global state), we make these notifications dispatch barriers, which means that all messages before the edit need to be handled before the edit can be processed and all messages received after the edit are handled after the edit has been sent to sourcekitd/clangd.

This kind of concurrent execution matches what we are doing in `sourcekitd` where open/edit/close are dispatch barriers.

Technically, we could optimize this further by having an `AsyncQueue` for each file, because edits on one file should not block requests on another file from executing but, at least in Swift, this would get us any real benefits at the moment because sourcekitd only has a single, global queue, instead of a queue per file. Additionally, usually you are editing one file in a source editor, which means that concurrent requests to multiple files tend to be rare. Thus, I will not pursue this optimization for now.

### DIFF
--- a/Sources/LanguageServerProtocol/AsyncQueue.swift
+++ b/Sources/LanguageServerProtocol/AsyncQueue.swift
@@ -33,53 +33,93 @@ extension NSLock {
   }
 }
 
-/// A serial queue that allows the execution of asyncronous blocks of code.
+/// A queue that allows the execution of asyncronous blocks of code.
 public final class AsyncQueue {
-  /// Lock guarding `lastTask`.
-  private let lastTaskLock = NSLock()
+  public enum QueueKind {
+    /// A queue that allows concurrent execution of tasks.
+    case concurrent
 
-  /// The last scheduled task if it hasn't finished yet.
-  ///
-  /// Any newly scheduled tasks need to await this task to ensure that tasks are
-  /// executed syncronously.
-  ///
-  /// `id` is a unique value to identify the task. This allows us to set `lastTask`
-  /// to `nil` if the queue runs empty.
-  private var lastTask: (task: AnyTask, id: UUID)?
+    /// A queue that executes one task after the other.
+    case serial
+  }
 
-  public init() {
-    self.lastTaskLock.name = "AsyncQueue.lastTaskLock"
+  private struct PendingTask {
+    /// The task that is pending.
+    let task: any AnyTask
+
+    /// Whether the task needs to finish executing befoer any other task can
+    /// start in executing in the queue.
+    let isBarrier: Bool
+
+    /// A unique value used to identify the task. This allows tasks to get
+    /// removed from `pendingTasks` again after they finished executing.
+    let id: UUID
+  }
+
+  /// Whether the queue allows concurrent execution of tasks.
+  private let kind: QueueKind
+
+  ///  Lock guarding `pendingTasks`.
+  private let pendingTasksLock = NSLock()
+
+  /// Pending tasks that have not finished execution yet.
+  private var pendingTasks = [PendingTask]()
+
+  public init(_ kind: QueueKind) {
+    self.kind = kind
+    self.pendingTasksLock.name = "AsyncQueue"
   }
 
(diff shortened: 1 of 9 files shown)


Sources/LanguageServerProtocol/AsyncQueue.swift
```diff
@@ -33,53 +33,97 @@ extension NSLock {
...
+  /// added later will wait until the barrier finishes execution.
   @discardableResult
   public func async<Success: Sendable>(
     priority: TaskPriority? = nil,
+    barrier isBarrier: Bool = false,
     @_inheritActorContext operation: @escaping @Sendable () async -> Success
   ) -> Task<Success, Never> {
+    // We treat a serial queue just like a concurrent queue in which every task
+    // is a barrier.
+    let isBarrier = isBarrier || kind == .serial
```
hamishknight, line 89: Is this a holdover from a previous design? Best I can tell, when the kind is `.serial`, we don't care about `isBarrier`

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -45,6 +45,19 @@ actor ClangLanguageServerShim: ToolchainLanguageServer, MessageHandler {
   /// The queue on which clangd calls us back.
   public let clangdCommunicationQueue: DispatchQueue = DispatchQueue(label: "language-server-queue", qos: .userInitiated)
 
+  /// The queue on which all messages that originate from clangd are handled.
+  ///
+  /// This includes requests and notifications sent *from* clangd and does not
+  /// include replies from clangd.
+  ///
+  /// These are requests and notifications sent *from* clangd, not replies from
+  /// clangd.
```
hamishknight, line 54: Nit: duplication

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -553,11 +574,11 @@ extension ClangLanguageServerShim {
...
   }
 
-  func foldingRange(_ req: Request<FoldingRangeRequest>) {
+  func foldingRange(_ req: FoldingRangeRequest) async throws -> [FoldingRange]? {
     if self.capabilities?.foldingRangeProvider?.isSupported == true {
-      forwardRequestToClangd(req)
+      return try await forwardRequestToClangd(req)
     } else {
-      req.reply(.success(nil))
+      return nil
```
hamishknight, line 581: This was the case before, but IMO this would be suitable for a guard

hamishknight (approved):

Sources/LanguageServerProtocol/AsyncQueue.swift
ahoppen: No, this is essential and is where the serial part of the queue comes in. The queue itself doesn’t really have an inherrent notion of being serial or concurrent (we don’t use dispatch queues or anything like that). It’s just that if the queue is serial, every task gets enqueued as a barrier, which means that they will be executed one after the other.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: Thanks 👍🏽

Sources/LanguageServerProtocol/AsyncQueue.swift
ahoppen: Hmm, wait. Actually that’s not true and you are right, we don’t need this. Forget my comment above.

PR #853, opened 2023-10-04
Asyncify the handling of more requests

ahoppen:
We can now harvest the benefits of migrating things to Swift concurrency and change the handling of all requests to asynchronously return the response instead of taking the implicit completion handler via the `Request`.

All changes are very straightforward. Again, this is probably best reviewed with whitespace changes hidden.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1268,11 +1268,11 @@ extension SourceKitServer {
   }
 
   func documentColor(
-    _ req: Request<DocumentColorRequest>,
+    _ req: DocumentColorRequest,
     workspace: Workspace,
     languageService: ToolchainLanguageServer
-  ) async {
-    await languageService.documentColor(req)
+  ) async throws -> [ColorInformation]{
+    return try await languageService.documentColor(req)
   }
 
   func documentSemanticTokens(
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -579,7 +579,7 @@ extension SourceKitServer: MessageHandler {
       case let request as Request<CallHierarchyOutgoingCallsRequest>:
         await self.handleRequest(request, handler: self.outgoingCalls)
       case let request as Request<TypeHierarchySupertypesRequest>:
-        await self.supertypes(request)
+        await self.handleRequest(request, handler: self.supertypes)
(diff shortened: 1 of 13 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/VariableTypeInfo.swift
```diff
@@ -91,11 +91,10 @@ extension SwiftLanguageServer {
   ///   - completion: Completion block to asynchronously receive the VariableTypeInfos, or error.
```
hamishknight, line 91: Same here

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -548,65 +565,65 @@ extension SourceKitServer: MessageHandler {
...
+        await self.handleRequest(request, handler: self.supertypes)
       case let request as Request<TypeHierarchySubtypesRequest>:
-        await self.subtypes(request)
+        await self.handleRequest(request, handler: self.subtypes)
       case let request as Request<CompletionRequest>:
-        await self.withLanguageServiceAndWorkspace(for: request, requestHandler: self.completion, fallback: CompletionList(isIncomplete: false, items: []))
+        await self.handleRequest(for: request, requestHandler: self.completion, fallback: CompletionList(isIncomplete: false, items: []))
       case let request as Request<HoverRequest>:
-        await self.withLanguageServiceAndWorkspace(for: request, requestHandler: self.hover, fallback: nil)
+        await self.handleRequest(for: request, requestHandler: self.hover, fallback: nil)
```
hamishknight, line 588: Presumably the fact that we're now awaiting for results means that e.g if we've sent off a completion request we'll now need to wait for a response before we can process any notifications that require a barrier. Is that expected? Is this something that will be helped by cancellation?

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):
Nice! Couple of minor comments and a question

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: We discussed this offline. We should have a more granular tracking of dependencies in `AsyncQueue`. I’ve got some ideas of how to do this and will create a follow-up PR in the next couple of weeks.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #852, opened 2023-10-04
`closeHandler` callback can be `nil` in an initializer


AndreyKlebanov:
https://github.com/apple/sourcekit-lsp/blob/05b3abd4ed8cbea7fa6539ad9deaf8abe8abafcd/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift#L108

self.closeHanlder callback can be nil in constructor when DispatchIO() fails with _"IO error \(error)"_.

One way to approach this - check if closeHandler as assigned before calling it:
`if (self.closeHandler != nil) { self.closeHandler() }`

Alternatively we could make .init() throwable and throw when IO error happens.

ahoppen:
Tracked in Apple’s issue tracker as rdar://116460760

ahoppen:
Oh, good catch @AndreyKlebanov. Would you like to open a PR to fix this? I think changing `closeHandler` from an implicitly unwrapped optional to an optional would be a good thing to do in general.

ahoppen:
Fixed by https://github.com/apple/sourcekit-lsp/pull/855

PR #855, opened 2023-10-05
Don't call closeHandler if it is nil

AndreyKlebanov:
Issue: #852 closeHandler callback can be nil in an initializer


MaxDesiatov (commented):
Would you be able to change `closeHandler`'s type to an explicit `?` optional type instead of `!` IOU? That way you can also write `await self.closeHandler?()` instead of doing a verbose `if self.closeHandler != nil` check.

MaxDesiatov (approved):
LGTM, thanks!

MaxDesiatov:
@swift-ci test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):
Thanks for fixing this so quickly after filing the issue 😄

AndreyKlebanov:
@ahoppen My pleasure. Great project!

PR #856, opened 2023-10-06
Remove dead code

ahoppen:
A couple of cleanups, mostly deleting completion handler code that was no longer called.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -765,7 +757,16 @@ extension SwiftLanguageServer {
       let closeHelperReq = SKDRequestDictionary(sourcekitd: self.sourcekitd)
       closeHelperReq[keys.request] = self.requests.editor_close
       closeHelperReq[keys.name] = helperDocumentName
-      _ = self.sourcekitd.send(closeHelperReq, .global(qos: .utility), reply: { _ in })
+      // FIXME: (async) We might receive two concurrent document color requests for the 
+      // same document, in which race to open/close a document with the same name in 
+      // sourcekitd. The solution is to either
+      //  - Not open the helper document and instead rely on the document that is already 
+      //    open or
```
bnbarham, line 764: Do you know if there's some reason we wouldn't just do this (same for the other case as well)?

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
```diff
@@ -1,103 +0,0 @@
...
-
-import Dispatch
-import LanguageServerProtocol
-import SourceKitD
-
-/// A typed expression as returned by sourcekitd's CollectExpressionType.
-///
-/// A detailed description of the structure returned by sourcekitd can be found
-/// here: https://github.com/apple/swift/blob/main/tools/SourceKit/docs/Protocol.md#expression-type
-struct ExpressionTypeInfo {
```
bnbarham, line 21: Were inlay hints just never hooked up, or are they done some other way now?

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: I *think* it’s because we are getting the substructure back as a reply to the open and we didn’t capture them during the initial opening/updates.

The real solution here should probably to either create a sourcekitd request that returns the document structure or migrate it to use swift-syntax. I haven’t looked into what exactly `DocumentSymbols` does and whether migrating it to swift-syntax is even possible.

Sources/SourceKitLSP/Swift/ExpressionTypeInfo.swift
ahoppen: This was the part of inlay hints that could show type annotations for expressions, not variables. I’m not sure if that ever made it in or if we removed it pretty early on because annotating `1` and `2` in `let x = 1 + 2` to `let x: Int = x: Int + 2: Int` wasn’t really useful.

`VariableTypeInfo` still exists.

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #857, opened 2023-10-06
Store syntax trees in semantic token in syntax tree/semantic token managers instead of in the `Document`

ahoppen:
Storing the syntax tree and semantic tokens in `Document` was an anti-pattern because those are Swift-related information, but `Document` should be language agnostic.

It also fixes the issue that we were polling for the syntax tree to be created. Instead, we can now await the construction of the syntax tree.


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -382,9 +329,9 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
```
bnbarham, line 329: Do you have a TODO anywhere for changing this to the diagnostic request + a separate semantic token request (possibly part of enabling cancellation)?

Sources/SourceKitLSP/Swift/SyntaxTreeManager.swift
```diff
@@ -0,0 +1,112 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftParser
+import SwiftSyntax
+
+/// Keeps track of SwiftSyntax trees for document snapshots and computes the
+/// SwiftSyntax trees on demand.
```
bnbarham, line 17: Is the thought that the highlight ranges in the semantic case aren't as expensive as the entire syntax tree? ie. why does one have a cache limit but the other doesn't? They're also both quite similar, ie. take an edit and cache the result (which here is the syntax tree and in the semantic is the highlight ranges).

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1510,6 +1504,9 @@ final class LocalSwiftTests: XCTestCase {
       log("Received diagnostics for open - semantic")
     })
 
+    // Send a request that triggers a syntax tree to be built.
+    _ = try sk.sendSync(FoldingRangeRequest(textDocument: .init(uri)))
```
bnbarham, line 1508: What's this needed for?

bnbarham (commented):
So many managers 😅 

> Storing the syntax tree and semantic tokens in Document was an anti-pattern because those are Swift-related information, but Document should be language agnostic.

Is Document itself not really just a Swift thing anyway? I assume open/edit/close are all just forwarded directly to Clang, so is it even used there?

Sources/SourceKitLSP/Swift/SyntaxTreeManager.swift
```diff
@@ -0,0 +1,112 @@
...
+actor SyntaxTreeManager {
+  /// A task that parses a SwiftSyntax tree from a source file, producing both
+  /// the syntax tree and the lookahead ranges that are needed for a subsequent
+  /// incremental parse.
+  private typealias SyntaxTreeComputation = Task<(tree: SourceFileSyntax, lookaheadRanges: LookaheadRanges), Never>
```
hamishknight, line 22: Not really a comment on this PR, but is there a reason `Parser.parseIncrementally` returns a tuple instead of a dedicated struct? Seems like a single type would be easier to manage, and would help ensure you don't pass a mismatching syntax tree + lookahead ranges to `IncrementalParseTransition` (and would make it trivial to add any additional state later down the road if needed).

Sources/SourceKitLSP/Swift/SemanticTokensManager.swift
```diff
@@ -0,0 +1,90 @@
...
+import LanguageServerProtocol
+
+/// Keeps track of the semantic tokens that sourcekitd has sent us for given
+/// document snapshots.
+actor SemanticTokensManager {
+  private var semanticTokens: [DocumentSnapshot.ID: [SyntaxHighlightingToken]] = [:]
+
+  /// The semantic tokens for the given snapshot or `nil` if no semantic tokens
+  /// have been computed yet.
+  func semanticTokens(for snapshotID: DocumentSnapshot.ID) -> [SyntaxHighlightingToken]? {
```
hamishknight, line 22: Personally I would prefer to wrap `[SyntaxHighlightingToken]` in its own struct, e.g then you could put members like `lspEncoded` and `mergingTokens` on it, and maybe move the edit processing logic onto it too. I don't feel that strongly about it though.

hamishknight (approved):

Sources/SourceKitLSP/Swift/SemanticTokensManager.swift
```diff
@@ -0,0 +1,90 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+
+/// Keeps track of the semantic tokens that sourcekitd has sent us for given
+/// document snapshots.
+actor SemanticTokensManager {
+  private var semanticTokens: [DocumentSnapshot.ID: [SyntaxHighlightingToken]] = [:]
```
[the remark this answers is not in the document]
ahoppen: I would prefer to keep it as-is. If we ever decide that we need to store semantic tokens for older version as well, it’s easier to update and conceptually we are storing semantic tokens a document snapshot, the cache eviction logic is separate from that.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: There already exists the separate diagnostic request https://github.com/apple/sourcekit-lsp/blob/1a7939304b3caf4680945272ac70e2dae94fdf24/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L1320-L1351.

We won’t be able to get rid of the diagnostic pushing though, because most editor (AFAIK every editor except VS Code) doesn’t support the pull diagnostics request. I added an item to my To Do list to define the diagnostic push in terms of the diagnostic pull internally though. That way we can hopefully make the document update syntactic only. I will file GitHub issues for all the items on my to do list once I’m done reaving havoc around the entire code base and it’s actually possible for multiple people to work on sourcekit-lsp again.

Sources/SourceKitLSP/Swift/SyntaxTreeManager.swift
ahoppen: The difference is that we can re-compute the syntax tree on demand but we can’t (in the current design) re-compute the semantic tokens. That’s why we can be more aggressive about discarding syntax trees than semantic tokens. If we wanted to change that, getting semantic tokens would need to become a sourcekitd request as well, instead of being sent over implicitly on every update (which we should probably do).

Tests/SourceKitLSPTests/LocalSwiftTests.swift
ahoppen: Previously, we were always eagerly computing the syntax tree whenever we opened a file. Now, we only compute it once the first request requires it (and then implicitly keep it up to date on every edit, the thought being that you’ll probably need it again and doing incremental parses is faster than re-parsing from scratch when it’s needed next). 

To test that we are parsing incrementally on the next edit, we need to trigger that initial build of a syntax tree, which is what I’m doing here.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -198,37 +199,21 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
...
         let tokenParser = SyntaxHighlightingTokenParser(sourcekitd: sourcekitd)
         var tokens: [SyntaxHighlightingToken] = []
         tokenParser.parseTokens(skTokens, in: snapshot, into: &tokens)
 
-        docTokens.semantic = tokens
+        return tokens
       }
 
-      return docTokens
+      return nil
```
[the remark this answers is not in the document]
ahoppen: Good idea

Sources/SourceKitLSP/Swift/SyntaxTreeManager.swift
ahoppen: Yeah, I thought the same when using the API. We should probably change that

Filed https://github.com/apple/swift-syntax/issues/2267

Sources/SourceKitLSP/Swift/SemanticTokensManager.swift
ahoppen: Good idea. It’s a slightly bigger change and I’ll do it in a follow-up PR.

ahoppen:
> Is Document itself not really just a Swift thing anyway? I assume open/edit/close are all just forwarded directly to Clang, so is it even used there? Or should Document/DocumentManager/etc just be Swift-only and then we handle all the things in there rather than the separate syntax/semantic managers?

We still need to keep track of the document contents for clang so that we can re-open the documents in clangd if clangd crashes (for which we need to know the current document contents).

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
bnbarham: > I added an item to my To Do list to define the diagnostic push in terms of the diagnostic pull internally though

👍

Sources/SourceKitLSP/Swift/SemanticTokensManager.swift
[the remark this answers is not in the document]
bnbarham: > conceptually we are storing semantic tokens a document snapshot

Heh, I see this as the other way. We're storing tokens for the latest version of a document. The request for tokens is really "give me the latest tokens of the document which should match version <n>". If it doesn't, then we're out of date and the response isn't likely to matter anyway.

Or to put another away:

> If we ever decide that we need to store semantic tokens for older version as well

Doesn't seem like a use case we care about.

> We still need to keep track of the document contents for clang so that we can re-open the documents in clangd if clangd crashes (for which we need to know the current document contents).

🙇

PR #858, opened 2023-10-06
Handle notifications from `sourcekitd` on a serial queue

ahoppen:
The issue is that we handle the (crash) notifications from sourcekitd on a concurrent queue at the moment.

When sourcekitd crashes, we re-open the documents within it immediately and the re-opening of documents will be sent to sourcekitd (and return) after sourcekitd comes back online.

But since we handle notifications from sourcekitd concurrently at the moment, we might handle the notification that sourcekitd is back online before we had a chance to re-open the documents and now we’re sending a cursor info to sourcekitd for a document that isn’t open.

To fix this, handle notifications from sourcekitd on a serial queue so that we serve them in-order.

rdar://116584307

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -111,6 +111,10 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
 
   let sourcekitd: SourceKitD
 
+  /// Queue on which notifications from sourcekitd are handled to ensure we are
+  /// handling them in-order.
+  let sourcekitdNotificationHandlingQueue = AsyncQueue(.serial)
+
   let capabilityRegistry: CapabilityRegistry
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -111,6 +111,10 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
 
   let sourcekitd: SourceKitD
 
+  /// Queue on which notifications from sourcekitd are handled to ensure we are
+  /// handling them in-order.
+  let sourcekitdNotificationHandlingQueue = AsyncQueue(.serial)
```
xedin, line 116: Should this be an AsyncSequence instead instead of rolling our own types?

PR #860, opened 2023-10-09
Implement request cancellation

ahoppen:
When receiving a `CancellationNotification`, we cancel the task that handles the request with that ID.

This will cause `cancel_notification` to be sent to sourcekitd or a `CancellationNotification` to be sent to `clangd`, which ultimately cancels the request.

rdar://117492860


ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/Connection.swift
```diff
@@ -136,3 +137,29 @@ extension LocalConnection: Connection {
...
+  /// connection if the task it is executing in is being cancelled.
+  ///
+  /// - Warning: Because this message is `async`, it does not provide any ordering
+  ///   guarantees. If you need to gurantee that messages are sent in-order
+  ///   use the version with a completion handler.
+  public func send<R: RequestType>(_ request: R) async throws -> R.Response {
+    let requestIDWrapper = ThreadSafeBox<RequestID?>(initialValue: nil)
+    return try await withTaskCancellationHandler {
+      try await withCheckedThrowingContinuation { continuation in
+        let requestID = self.send(request) { result in
```
hamishknight, line 154: Should this be checking for `Task.checkCancellation()` before sending the request? Asking because I noticed you're doing so in the sourcekitd use of `withTaskCancellationHandler`, and after looking at the docs for `withTaskCancellationHandler` it seems like if the task is already cancelled, it will call `onCancel` _before_ the operation.

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -581,6 +597,14 @@ extension SourceKitServer: MessageHandler {
...
       }
+      // We have handled the request and can't cancel it anymore.
+      // Stop keeping track of it to free the memory.
+      await self.setInProgressRequest(for: id, task: nil)
+    }
+    // Start tracking the request with a high priority to minimize the chance
+    // of cancellation
+    cancellationMessageHandlingQueue.async {
+      await self.setInProgressRequest(for: id, task: task)
     }
```
hamishknight, line 606: Is this meant to be set to high priority?

Sources/SKSupport/ThreadSafeBox.swift
```diff
@@ -0,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
```
hamishknight, line 5:
```suggestion
// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
```

hamishknight (approved):
Nice!

Sources/LanguageServerProtocol/Connection.swift
ahoppen: Oh, very good catch. We should definitely call `Task.checkCancellation()` here.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: No, it’s not. Just adding the ID -> Task mapping to the dictionary isn’t really important. Once we actually cancel a request, the cancellation task (which has high priority), will be blocked on this task, which will raise its priority to `high` as well (https://developer.apple.com/wwdc23/10170?time=515). So, actually, we should run this at a very low priority because it’s just bookkeeping that doesn’t matter until you actually cancel something.

hamishknight: Should the comment be changed then? :)

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Yes

ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/Connection.swift
```diff
@@ -136,3 +137,30 @@ extension LocalConnection: Connection {
...
+  /// This method automatically sends a `CancelRequestNotification` to the
+  /// connection if the task it is executing in is being cancelled.
+  ///
+  /// - Warning: Because this message is `async`, it does not provide any ordering
+  ///   guarantees. If you need to gurantee that messages are sent in-order
+  ///   use the version with a completion handler.
+  public func send<R: RequestType>(_ request: R) async throws -> R.Response {
+    let requestIDWrapper = ThreadSafeBox<RequestID?>(initialValue: nil)
+    try Task.checkCancellation()
+    return try await withTaskCancellationHandler {
```
hamishknight, line 153: Should the `checkCancellation` happen inside `withTaskCancellationHandler` to match the sourcekitd logic? Seems like there could be a potential race condition where the task is cancelled between the calls to `checkCancellation` and `withTaskCancellationHandler`, which would still lead to the request being sent

hamishknight: Come to think of it, is there a potential race condition if the task is cancelled before `requestIDWrapper` is assigned? 🤔

ahoppen: I think it doesn’t really matter in practice but I moved it for consistency.

> Come to think of it, is there a potential race condition if the task is cancelled before `requestIDWrapper` is assigned? 🤔

Yes, I think there is a potential race that could happen here. But since the window in which it can happen is really short and this is only cancellation, which is best-effort anyway, I decided that it’s OK to accept it. The alternative would be yet another dispatch queue that blocks `onCancel` from executing until the request ID has been set and that seemed a little bit like overkill to me.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/LanguageServerProtocol/Connection.swift
bnbarham: How would we actually cancel before the client has the request ID back anyway?

ahoppen:
@swift-ci Please test Windows

Sources/LanguageServerProtocol/Connection.swift
ahoppen: In LSP, the client specifies the request ID

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#contentPart

bnbarham: Ah I see, unfortunate.

hamishknight: > The alternative would be yet another dispatch queue that blocks `onCancel` from executing until the request ID has been set and that seemed a little bit like overkill to me

I guess one other alternative would be to check again for cancellation afterwards, e.g something like:

```swift
extension ThreadSafeBox {
  func takeValue<U>() -> T where U? == T {
    lock.withLock {
      guard let value = self.value else { return nil }
      self.value = nil
      return value
    }
  }
}

func withCancellableThrowingContinuation<Success, CancelToken>(
  _ body: (CheckedContinuation<Success, any Error>) -> CancelToken,
  onCancel cancel: (CancelToken) -> Void
) async throws -> Success {
  let cancelWrapper = ThreadSafeBox<CancelToken?>(initialValue: nil)

  @Sendable
  func cancelFn() {
    if let token = cancelWrapper.takeValue() {
      cancel(token)
    }
  }

  return try await withTaskCancellationHandler(operation: {
    try Task.checkCancellation()
    return try await withCheckedThrowingContinuation { cont in
      cancelWrapper.value = body(cont)
      if Task.isCancelled {
        cancelFn()
      }
    }
  }, onCancel: cancelFn)
}
```

Not sure if that's worth it though.

ahoppen:
This is currently blocked because we open the documents in sourcekitd semantically, not with `syntax_only`. We thus build an AST for the test file in `testCancellation`, which takes a couple of minutes, blocking sourckitd and makes us run into timeouts in later test cases that need semantic sourcekitd functionality. This should be fixed by changing open/edit to be syntactic only (kind of tracked by https://github.com/apple/sourcekit-lsp/issues/882). I’ll let this

Sources/LanguageServerProtocol/Connection.swift
```diff
@@ -136,3 +137,44 @@ extension LocalConnection: Connection {
...
+        requestIDWrapper.value = requestID
+
+        // Check if the task was cancelled. This ensures we send a
+        // CancelNotification even if the task gets cancelled after we register
+        // the cancellation handler but before we set the `requestID`.
+        if Task.isCancelled {
+          sendCancelNotification()
+        }
+      }
+    }, onCancel: sendCancelNotification)
```
hamishknight, line 178: Even though it's only used twice, IMO it would still be nice to factor this logic out similar to the function in my previous comment since it seems like it's something that's easy to get wrong. I don't feel that strongly about it though.

ahoppen: You’re right. I factored it out into a separate function.

Sources/SKSupport/AsyncUtils.swift
```diff
@@ -0,0 +1,64 @@
...
+/// Allows the execution of a cancellable operation that returns the results
+/// via a completion handler.
+///
+/// `operation` must invoke the continuation's `resume` method exactly once.
+///
+/// If the task executing `withCancellableCheckedThrowingContinuation` gets
+/// cancelled, `cancel` is invoked with the handle that `operation` provided.
+public func withCancellableCheckedThrowingContinuation<Handle, Result>(
+  _ operation: (_ continuation: CheckedContinuation<Result, any Error>) -> Handle,
+  cancel: (Handle) -> Void
```
hamishknight, line 37: I think this probably ought to be marked `@Sendable` (I believe we'll warn under strict concurrency)

hamishknight: Talking of strict concurrency checking, are we planning on enabling that for this repo? I don't _think_ it's a proper package option yet, but I think you can add `.enableExperimentalFeature("StrictConcurrency")` as a build setting

ahoppen: I was planning to try and enable strict concurrency checking once I don’t have any further async migrations in my backlog. Until I do that, I don’t think I care about sendable annotations too much.

hamishknight: Fair enough

ahoppen:
@swift-ci Please test

ahoppen:
Since we open documents as syntactic-only now, we can support cancellation. 🎉🏎️

bnbarham (approved):
🥳

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #864, opened 2023-10-09
Wait for asynchronous setup in `BuildSystemTests` using an expectation instead of a semaphore

ahoppen:
Just cleaner because it will also fail if it times out.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #871, opened 2023-10-09
Don't hop on background queue to run `onExit`

ahoppen:
Since we use Swift Concurrency now, we no longer need to be worried about queues blocking each other, so the queue hopping is no longer necessary.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #888, opened 2023-10-10
Make `PullDiagnosticsTests.testCodeActions` pass with older toolchains

ahoppen:
The Fix-It message that we are testing here has changed in https://github.com/apple/swift/pull/67909. Accept both the old and the new message to make sure SourceKit-LSP tests also pass with older toolchains (e.g. if are using Xcode 15.0, which doesn’t contain the change yet).

rdar://116706363


hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #890, opened 2023-10-10
Format using swift-format

ahoppen:
Add `.swift-format` to the repo and format the repo with `swift-format`.

This PR does not add any automation to enforce formatting of sourcekit-lsp in CI. The goal of the PR is to get the majority of source changes out of the way so that the diff of actually enforcing formatting will have fewer changes or conflicts.

I scanned through the changes, made a couple of improvements, like converting long function signatures to use `some` but other than that, all of the changes are mechanical.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,17 @@
+{
+    "version": 1,
+    "lineLength": 120,
+    "indentation": {
+        "spaces": 2
+    },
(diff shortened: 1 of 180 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #894, opened 2023-10-12
Use new SwiftPM flag to remove `$ORIGIN` from installed sourcekit-lsp ELF executable runpath

finagolfin:
Also, move installation-only environment variable into `get_swiftpm_environment_variables()`.

One of the latest trunk snapshot toolchains shows five executables built by SwiftPM that incorrectly have this runpath, including `sourcekit-lsp` from this repo:
```
> readelf -d swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/*|ag "File:|runpath"|ag "ORIGIN[^/]" -B1
File: swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/docc
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib/swift/linux]
File: swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/sourcekit-lsp
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib/swift/linux]
File: swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/swift-build-sdk-interfaces
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib/swift/linux]
File: swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/swift-driver
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib/swift/linux]
File: swift-DEVELOPMENT-SNAPSHOT-2023-10-10-a-ubuntu20.04/usr/bin/swift-help
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib/swift/linux]
```
This will remove that unneeded runpath, as I just did for `swift-package` in apple/swift-package-manager#6947. I limited this to non-Darwin installs because I have no idea if this would be useful on Darwin too.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -124,6 +124,8 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
             '-Xcxx', '-I', '-Xcxx',
             os.path.join(args.toolchain, 'lib', 'swift', 'Block'),
         ]
+        if args.action == 'install':
+            swiftpm_args += ['--disable-local-rpath']
 
     if '-android' in build_target:
         swiftpm_args += [
@@ -181,6 +183,9 @@ def get_swiftpm_environment_variables(swift_exec: str, args: argparse.Namespace)
     if args.action == 'test' and args.skip_long_tests:
         env['SKIP_LONG_TESTS'] = '1'
 
+    if args.action == 'install':
+        env['SOURCEKIT_LSP_CI_INSTALL'] = "1"
+
     return env
 
 
@@ -190,8 +195,6 @@ def build_single_product(product: str, swift_exec: str, args: argparse.Namespace
     """
     swiftpm_args = get_swiftpm_options(swift_exec, args)
     additional_env = get_swiftpm_environment_variables(swift_exec, args)
-    if args.action == 'install':
-      additional_env['SOURCEKIT_LSP_CI_INSTALL'] = "1"
     cmd = [swift_exec, 'build', '--product', product] + swiftpm_args
     check_call(cmd, additional_env=additional_env, verbose=args.verbose)
 


Utilities/build-script-helper.py
```diff
@@ -190,8 +195,6 @@ def build_single_product(product: str, swift_exec: str, args: argparse.Namespace
     """
     swiftpm_args = get_swiftpm_options(swift_exec, args)
     additional_env = get_swiftpm_environment_variables(swift_exec, args)
-    if args.action == 'install':
-      additional_env['SOURCEKIT_LSP_CI_INSTALL'] = "1"
```
finagolfin, line 194: I think I put this here and not in `get_swiftpm_environment_variables()` initially either because it broke something when [invoking `swiftpm_bin_path()` also in the install action](https://github.com/apple/sourcekit-lsp/blob/4a22e0e725733d52f76ae692b6115ba89d302afe/Utilities/build-script-helper.py#L228) or I was concerned it would, I forget which. Let's try it on the CI and find out.

Utilities/build-script-helper.py
```diff
@@ -124,6 +124,8 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
             '-Xcxx', '-I', '-Xcxx',
             os.path.join(args.toolchain, 'lib', 'swift', 'Block'),
         ]
+        if args.action == 'install':
+            swiftpm_args += ['--disable-local-rpath']
```
ahoppen, line 128: Could you check if this causes sourcekit-lsp to get rebuilt if you do a build first and then an install? Because we want to avoid having to build sourcekit-lsp twice in CI, once for build and once for install.

finagolfin: It would have to get rebuilt, as the normal build includes the local rpath but the installation build doesn't. I think you're worried about if all the source gets rebuilt, rather than simply the final link step getting redone?

Considering only the latter happens for the added `linkerSettings` that I previously added in #715 and this SwiftPM flag also only affects the link arguments for executable targets, I believe this flag won't trigger a full rebuild either, but I haven't tried it on this repo locally to make sure.

ahoppen: Yes, re-linking would be totally fine because that really changes. I just want to avoid re-building all the Swift sources. That’s what I would expect as well, but you never know.

If you could check that the install only re-links, that would be great. You could also trigger a toolchain build from apple/swift using cross PR testing and see if the installation time of sourcekit-lsp increases.

finagolfin: Done, applied this pull and built on my Android phone: it did not trigger a full rebuild, only 7 items were rebuilt. It was helpful though, as it showed a mistake I made in not renaming the `additional_env` variable when moving it, which I've now fixed.

finagolfin:
Ping, a CI run and this can go in? I'll submit for 5.10 next.

finagolfin:
Ready for merge, will submit for 5.10 once this is in.

PR #899, opened 2023-10-12
Fix a race condition in code completion

### LINKED ISSUE #863: Remove client-side filtering for code completion
`SKCompletionOptions.serverSideFiltering` is `true` by default and I know of no editor that disables it. We should probably get rid of the option and always perform server-side filtering, deleting the client-side filtering code (`completionWithClientFiltering`).

ahoppen:
Each `sourcekitd` uses a single, global code completion session but we were managing code completion sessions on the `SwiftLanguageServer` level. This had two issues:
- Code completion is considered non-blocking on `SourceKitServer` and thus we could execute two concurrent code completion requests, which would conflict with each other.
- You could have multiple `SwiftLanguageServer`s that each have a connection to `sourcekitd` but share the same `sourcekitd` state. This shouldn't happen in the real world but does happen if we create multiple `SourceKitServer` instances in tests.

As a preparation for this do the following:
- Remove client-side filtering for code completion
  - `SKCompletionOptions.serverSideFiltering` is `true` by default and I know of no editor that disables it. Delete it. (Fixes #863 / rdar://116703670)
- Change `CodeCompletionSession` to not have a circular reference to `SourceKitServer`
  - This just cleans up the layering slightly.


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -25,63 +25,155 @@ import SourceKitD
...
+  // MARK: - Public static API
+
+  /// The queue on which all code completion requests are executed.
+  ///
+  /// This is needed because sourcekitd has a single, global code completion
+  /// session and we need to make sure that multiple code completion requests
+  /// don't race each other.
+  ///
+  /// Technically, we would only need one queue for each sourcekitd and different
+  /// sourcektid could serve code completion requests simultaneously.
```
hamishknight, line 38:
```suggestion
  /// sourcekitd could serve code completion requests simultaneously.
```

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -25,10 +25,128 @@ import SourceKitD
...
+      if let session = completionSessions[ObjectIdentifier(sourcekitd)], session.state == .open {
+        let isCompatible = session.snapshot.uri == snapshot.uri &&
+        session.utf8StartOffset == completionUtf8Offset &&
+        session.position == completionPosition &&
+        session.compileCommand == compileCommand &&
+        session.clientSupportsSnippets == clientSupportsSnippets
+
+        if isCompatible {
+          return try await session.update(filterText: filterText, position: cursorPosition, in: snapshot, options: options)
+        } else {
```
hamishknight, line 109: Nit: Could remove this `else` since the `if` returns

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -25,10 +25,128 @@ import SourceKitD
...
+
+  /// The code completion session for each sourcekitd instance.
+  ///
+  /// `sourcekitd` has a global code completion session, that's why we need to
+  /// have a global mapping from `sourcekitd` to its currently active code
+  /// completion session.
+  ///
+  /// Modification of code completion sessions should only happen on
+  /// `completionQueue`.
+  private static var completionSessions: [ObjectIdentifier: CodeCompletionSession] = [:]
```
hamishknight, line 53: Can't say I love this global state, is there any reason not to store it on `SourceKitDImpl` instead? FWIW the abstraction with the `SourceKitD` protocol doesn't seem super compelling to me since it seems like the only other conforming type is `FakeSourceKitD`, and that's only used to test the registry, it doesn't implement any of the other requirements. Seems like we could turn `SourceKitDRegistry` it into a generic `LibraryRegistry<T>` type or something, and then delete the `SourceKitD` protocol.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -395,10 +393,6 @@ extension SwiftLanguageServer {
   }
 
   public func shutdown() async {
-    if let session = self.currentCompletionSession {
-      await session.close()
-      self.currentCompletionSession = nil
-    }
```
hamishknight, line 401: I guess we no longer close on shutdown?

hamishknight (approved):

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
ahoppen: Removing the abstraction of `SourceKitD` an `SourceKitDImpl` is also something that’s on my mind. Expect a follow-up PR for that sometime in the future.

> Can't say I love this global state, is there any reason not to store it on `SourceKitDImpl` instead?

It’s a layering decision: `CodeCompletionSession` is tied  quite closely to the needs of LSP (it takes `DocumentSnapshot` and returns `CompletionList`, both of which are defined in `LanguageServerProtocol`) while the `SourceKitD` module is a fairly low-level wrapper around the C sourcekitd API. So, I think `CodeCompletionSession` doesn’t belong in the `SourceKitD` module and thus it can’t be a member of `SourceKitD` (unless you are suggesting that `SourceKitDImpl` has something like `codeCompletionSession: Any` and I assume that’s not what you are suggesting).

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Yes, that’s correct. Shouldn’t be an issue though. Unless you shut down the server, we keep the latest completion session alive anyway, so no memory concerns here and if you are shutting down sourcekit-lsp, then sourcekitd is about to be shut down as well, at which point it doesn’t matter if there’s still an open session since sourcekitd doesn’t do any kind of persisting when closing the session.

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
hamishknight: Hmm yeah that's fair, it's unfortunate because it seems like completion sessions are a sourcekid-specific concept. If we wanted to sink `CodeCompletionSession` down to a lower layer, it does seem like it doesn't have to return a `CompletionList`, it could return the raw response and let the language server handle transforming it. And it also seems like it doesn't need to store a full `DocumentSnapshot`, it only needs the `DocumentURI` I think (though maybe there's a use I'm missing). In particular it looks like the use in `guard snapshot.version == self.snapshot.version` is now redundant since that should be true by construction.

Either that or we could have a higher level wrapper that encapsulates both a `SourceKitD` and any session state it needs.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
tp

ahoppen:
@swift-ci Please test Windows

PR #900, opened 2023-10-12
Add more fine grained dependency tracking to `AsyncQueue`

### LINKED ISSUE #875: More fine-grained dependency management in `AsyncQueue`
We currently have two dependency levels in `AsyncQueue`: Tasks that can be executed concurrently and tasks that are barriers. But the dependency management is more fine grained. Modification to file A isn’t a barrier to request for file B. We should implement a more fine-grained dependency management here. 

The idea would be that each task in an `AsyncQueue` has metadata associated with it. When enqueuing a new task, a closure would get invoked with all the metadata of tasks that are currently pending in the queue and you would need to answer whether this task is a dependency of yours or not.

For example, the metadata of the queue in `SourceKitServer` could look like the following

```swift
enum QueueMetadata {
  case globalConfigurationChange
  case documentUpdate(DocumentURI)
  case documentRequest(DocumentURI)
}
```

ahoppen:
Instead of just having barriers and non-barriers, this allows `AsyncQueue` to track dependencies between tasks at a more fine-grained level.

For example, we can now specify that requests that affect one document only depend on edits to that same document and are not blocked by edits to any other document. As a consequence, a busy `sourcekitd` will not block requests from `clangd` to be executed and vice versa.

Resolves #875
rdar://116705652

### DIFF
--- a/Sources/LanguageServerProtocol/AsyncQueue.swift
+++ b/Sources/LanguageServerProtocol/AsyncQueue.swift
@@ -33,31 +33,33 @@ extension NSLock {
   }
 }
 
-/// A queue that allows the execution of asyncronous blocks of code.
-public final class AsyncQueue {
-  public enum QueueKind {
-    /// A queue that allows concurrent execution of tasks.
-    case concurrent
+/// A type that is able to track dependencies between tasks.
+public protocol DependencyTracker {
+  /// Whether the task described by `self` needs to finish executing before
+  /// `other` can start executing.
+  func isDependency(of other: Self) -> Bool
+}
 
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #903, opened 2023-10-15
Dependencies could not be resolved because root depends on 'swift-crypto' 2.5.0..<3.0.0 and 'swift-package-manager' depends on 'swift-crypto' 3.0.0..<3.1.0.


joehsieh:
i was following the [guideline](https://github.com/apple/sourcekit-lsp/blob/main/Documentation/Development.md#building-sourcekit-lsp) to build the project sourcekit-lsp, but the error `error: Dependencies could not be resolved because root depends on 'swift-crypto' 2.5.0..<3.0.0 and 'swift-package-manager' depends on 'swift-crypto' 3.0.0..<3.1.0.` came up to me.

After checking the Package.swift in both repo swift-package-manager and sourcekit-lsp, i think there is a conflict.
Perhaps, the same [revert](https://github.com/apple/swift-package-manager/pull/6988) should be applied on the main branch of swift-package-manager as well. 

swift-package-manager
[.package(url: "https://github.com/apple/swift-crypto.git", .upToNextMinor(from: "3.0.0")),](https://github.com/apple/swift-package-manager/blob/main/Package.swift#L739)

sourcekit-lsp
[ .package(url: "https://github.com/apple/swift-crypto.git", from: "2.5.0"),](https://github.com/apple/sourcekit-lsp/blob/main/Package.swift#L273)

ahoppen:
Tracked in Apple’s issue tracker as rdar://116982771

MaxDesiatov:
This revert is unlikely to be applied to the `main` branch of SwiftPM. SourceKit-LSP should update to swift-crypto 3.0 instead.

joehsieh:
ah, okie. if so, let me create a PR to bump the version of swift-crypto to 3.0.0

PR #904, opened 2023-10-16
Bump swift-crypto to 3.0.0 in Package.swift

joehsieh:
To address the error
```
Dependencies could not be resolved because root depends on 'swift-crypto' 2.5.0..<3.0.0 and 'swift-package-manager' 
depends on 'swift-crypto' 3.0.0..<3.1.0. 
```
related issue: https://github.com/apple/sourcekit-lsp/issues/903

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -270,7 +270,7 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     .package(url: "https://github.com/apple/swift-tools-support-core.git", branch: relatedDependenciesBranch),
     .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2"),
     .package(url: "https://github.com/apple/swift-syntax.git", branch: relatedDependenciesBranch),
-    .package(url: "https://github.com/apple/swift-crypto.git", from: "2.5.0"),
+    .package(url: "https://github.com/apple/swift-crypto.git", from: "3.0.0"),
   ]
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -270,7 +270,7 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     .package(url: "https://github.com/apple/swift-tools-support-core.git", branch: relatedDependenciesBranch),
     .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2"),
     .package(url: "https://github.com/apple/swift-syntax.git", branch: relatedDependenciesBranch),
-    .package(url: "https://github.com/apple/swift-crypto.git", from: "2.5.0"),
+    .package(url: "https://github.com/apple/swift-crypto.git", from: "3.0.0"),
```
joehsieh, line 273: [Question] what's the logic to decide the range of versions?
should we align it with the setting in [swift-package-manager](https://github.com/apple/swift-package-manager/blob/main/Package.swift#L739)
`.upToNextMinor(from: "3.0.0")`

MaxDesiatov: This one being broad is fine, means we'll only need to bump it in SwiftPM in the future and keep it here as is until 4.0

MaxDesiatov:
@swift-ci test

ahoppen:
Thank you for fixing the issue @joehsieh!

PR #905, opened 2023-10-16
Structure Package.swift in the same way that we structure it in swift-syntax

### LINKED ISSUE #872: Structure Package.swift in the same way that we structure it in swift-syntax
We have some more structure in swift-syntax’s Package.swift than we currently do in sourcekit-lsp (like `// MARK:` annotations).

We should structure sourcekit-lsp’s Package.swift in the same way that we structure the one in swift-syntax.

ahoppen:
Generally makes it easier to read IMO if targets and test targets are grouped.

While at it, add a few missing dependency declarations in Package.swift in a separate commit. I think technically we we were fine because these were already included as transitive dependencies but I like to explicitly declare a dependency for every module that is being imported with an `import` statement.

Resolves #872
rdar://116705650

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -51,7 +51,10 @@ let package = Package(
     .executableTarget(
       name: "sourcekit-lsp",
       dependencies: [
+        "LanguageServerProtocol",
         "LanguageServerProtocolJSONRPC",
+        "SKCore",
+        "SKSupport",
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #906, opened 2023-10-16
Use NonDarwinLogger when running tests instead of os_log

ahoppen:
'swift test' doesn't print os_log output to the command line. Use the  `NonDarwinLogger` that prints to stderr so we can view the log output in CI test runs.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,6 +9,12 @@ func hasEnvironmentVariable(_ name: String) -> Bool {
   return ProcessInfo.processInfo.environment[name] != nil
 }
 
+/// Use the `NonDarwinLogger` even if `os_log` can be imported.
+///
+/// This is useful when running tests using `swift test` because xctest will not display the output from `os_log` on the
+/// command line.
+let forceNonDarwinLogger = hasEnvironmentVariable("SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER")
+
 // When building the toolchain on the CI, don't add the CI's runpath for the
 // final build before installing.
 let installAction = hasEnvironmentVariable("SOURCEKIT_LSP_CI_INSTALL")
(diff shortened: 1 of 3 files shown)


Utilities/build-script-helper.py
```diff
@@ -202,6 +202,16 @@ def run_tests(swift_exec: str, args: argparse.Namespace) -> None:
...
+    # 'swift test' doesn't print os_log output to the command line. Use the 
+    # `NonDarwinLogger` that prints to stderr so we can view the log output in CI test 
+    # runs.
+    additional_env['SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER'] = '1'
+
+    # CI doesn't contain any sensitive information. Log everything.
+    additional_env['SOURCEKITLSP_LOG_PRIVACY_LEVEL'] = 'sensitive'
+
+    # Log with the highest log level to simplify debugging of CI failures.
+    additional_env['SOURCEKITLSP_LOG_LEVEL'] = 'default'
```
bnbarham, line 214: Might want `debug` there :)

bnbarham (approved):

ahoppen:
@swift-ci Please test

hamishknight (approved):

Package.swift
```diff
@@ -9,6 +9,12 @@ func hasEnvironmentVariable(_ name: String) -> Bool {
   return ProcessInfo.processInfo.environment[name] != nil
 }
 
+/// Use the `NonDarwinLogger` even if `os_log` can be imported.
+///
+/// This is useful when running tests using `swift test` because xctest will not display the output from `os_log` on the
+/// command line.
+let forceNonDarwinLogger = hasEnvironmentVariable("SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER")
```
benlangmuir, line 16: Won't this mean build + test + install require 3 builds instead of 1?

ahoppen: Since none of the dependencies need to be rebuilt, the recompile when flipping the default is fairly fast (20s on my machine) and I think that’s quite acceptable to get log output when running tests.

PR #907, opened 2023-10-17
A couple of logging improvements

ahoppen:
The key changes here are:
- Set the request ID as the category for all log messages created while handling it (using `withLoggingScope`)
- Pretty-print requests and notifications as JSON
- Log build settings. If they are too long to be logged in a single log message, split them over multiple messages.

### DIFF
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -72,6 +72,22 @@ public actor BuildSystemManager {
   }
 }
 
+/// Splits `message` on newline characters such that each chunk is at most `maxChunkSize` bytes long.
+///
+/// The intended use case for this is to split compiler arguments into multiple chunks so that each chunk doesn't exceed
+/// the maximum message length of `os_log` and thus won't get truncated.
+private func splitLongMultilineMessage(message: String, maxChunkSize: Int) -> [String] {
+  var chunks: [String] = [""]
+  for line in message.split(separator: "\n", omittingEmptySubsequences: false) {
+    if chunks.last!.utf8.count + line.utf8.count < maxChunkSize {
+      chunks[chunks.count - 1] += "\n" + line
+    } else {
+      chunks.append(String(line))
+    }
+  }
+  return chunks
+}
+
 extension BuildSystemManager {
   public var delegate: BuildSystemDelegate? {
     get { _delegate }
@@ -136,6 +152,25 @@ extension BuildSystemManager {
       // If the main file isn't the file itself, we need to patch the build settings
       // to reference `document` instead of `mainFile`.
       buildSettings?.buildSettings = settings.patching(newFile: document.pseudoPath, originalFile: mainFile.pseudoPath)
+    }
+    if let buildSettings {
+      let log = """
+        Compiler Arguments:
+        \(buildSettings.buildSettings.compilerArguments.joined(separator: "\n"))
+
+        Working directory:
+        \(buildSettings.buildSettings.workingDirectory ?? "<nil>")
+        """
+
+      let chunks = splitLongMultilineMessage(message: log, maxChunkSize: 800)
+      for (index, chunk) in chunks.enumerated() {
+        logger.log(
+          """
+          Build settings for \(document.forLogging) (\(index + 1)/\(chunks.count))
+          \(chunk)
+          """
+        )
+      }
     }
     return buildSettings
   }
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -76,10 +76,13 @@ public actor BuildSystemManager {
 ///
 /// The intended use case for this is to split compiler arguments into multiple chunks so that each chunk doesn't exceed
 /// the maximum message length of `os_log` and thus won't get truncated.
+///
+///  - Note: This will only split along newline boundary. If a single line is longer than `maxChunkSize`, it won't be
+///    split. This is fine for compiler argument splitting since a single argument is rarely longer than 800 characters.
 private func splitLongMultilineMessage(message: String, maxChunkSize: Int) -> [String] {
-  var chunks: [String] = [""]
+  var chunks: [String] = []
   for line in message.split(separator: "\n", omittingEmptySubsequences: false) {
-    if chunks.last!.utf8.count + line.utf8.count < maxChunkSize {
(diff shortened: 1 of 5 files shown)


ahoppen:
https://github.com/apple/swift/pull/69230

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -540,97 +563,112 @@ extension SourceKitServer: MessageHandler {
...
+    let cancellationToken = CancellationToken()
+
+    let request = Request(
+      params,
+      id: id,
+      clientID: clientID,
+      cancellation: cancellationToken,
+      reply: { [weak self] result in
+        reply(result)
+        logger.log("Replied. Took \(-startDate.timeIntervalSinceNow * 1000)ms")
```
bnbarham, line 590: Could we just pass in the time as part of `logResponse`?

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -83,26 +83,22 @@ extension SourceKitD {
...
-        logResponse(resp)
-
-        guard let dict = resp.value else {
-          continuation.resume(throwing: resp.error!)
-          return
-        }
-
-        continuation.resume(returning: dict)
+      api.send_request(req.dict, &handle) { _resp in
+        continuation.resume(returning: SKDResponse(_resp, sourcekitd: self))
```
hamishknight, line 90: Is dropping the weak-strong dance intentional here?

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,6 +72,22 @@ public actor BuildSystemManager {
...
 
+/// Splits `message` on newline characters such that each chunk is at most `maxChunkSize` bytes long.
+///
+/// The intended use case for this is to split compiler arguments into multiple chunks so that each chunk doesn't exceed
+/// the maximum message length of `os_log` and thus won't get truncated.
+private func splitLongMultilineMessage(message: String, maxChunkSize: Int) -> [String] {
+  var chunks: [String] = [""]
+  for line in message.split(separator: "\n", omittingEmptySubsequences: false) {
+    if chunks.last!.utf8.count + line.utf8.count < maxChunkSize {
+      chunks[chunks.count - 1] += "\n" + line
```
hamishknight, line 83: Is it expected that the first chunk will start with a newline?

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -72,6 +72,22 @@ public actor BuildSystemManager {
   }
 }
 
+/// Splits `message` on newline characters such that each chunk is at most `maxChunkSize` bytes long.
```
hamishknight, line 75: This doesn't quite seem accurate to me since chunks can exceed `maxChunkSize` bytes for lines in the input that already exceeded that limit. Or should we be splitting long lines too?

hamishknight (approved):

Sources/SourceKitD/SourceKitD.swift
hamishknight: Actually I suppose the old implementation was wrong because it could have lead to the task being suspended indefinitely

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: The problem is that more time could have elapsed until `_logResponse` gets executed (which is only on a background task because describing the response could take a while). If we wanted to log it in `_logReponse` we would still need to capture the date here, then pass the date to `_logResponse` and at that point, we might as well just log here.

Sources/SourceKitD/SourceKitD.swift
ahoppen: Yeah, I looked at it and decided that it isn’t needed to prevent reference cycles. 

Good catch with `ClangLanguageServerShim.shutdown`, I’ll clean that up as well.

Sources/SKCore/BuildSystemManager.swift
ahoppen: Oh, no, that wasn’t intentional. Fixed it

Sources/SKCore/BuildSystemManager.swift
ahoppen: Yeah, I know… It’s good enough for a `fileprivate` function for command line argument splitting though. I just added a comment now.

```
This will only split along newline boundary. If a single line is longer than `maxChunkSize`, it won't be split. This is fine for compiler argument splitting since a single argument is rarely longer than 800 characters.
```

Sources/SourceKitLSP/SourceKitServer.swift
bnbarham: > If we wanted to log it in _logReponse we would still need to capture the date here

Yeah, that's what I meant.

> we might as well just log here

The downside as I see it is the extra log, when we already log the response 🤷. I'd just prefer to keep the amount of log spam down as much as we can.

ahoppen: I inlined the `_log*` functions and unified the duration + response logging.

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
bnbarham: Okay, that works too 🙇‍♂️. Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -592,159 +592,184 @@ public actor SourceKitServer {
...
+      clientID: clientID,
+      cancellation: cancellationToken,
+      reply: { [weak self] result in
+        reply(result)
+        let endDate = Date()
+        if let self {
+          Task {
+            logger.debug(
+              """
+              Sending response (took \(endDate.timeIntervalSince(startDate) * 1000)ms):
```
bnbarham, line 687: Same comment as usual with the `:` but super nitpicky and I don't really care, we can change at any time.

ahoppen: Changed it to just be a newline

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #861, opened 2023-10-09
Include `isFallback` in `FileBuildSettings` instead of returning a tuple


ahoppen:
Instead of having `BuildSystemManager.buildSettings` return a tuple of `FileBuildSettings` and `isFallback: Bool`, we should make `isFallback` part of `FileBuildSettings`.

ahoppen:
Tracked in Apple’s issue tracker as rdar://116703679

joehsieh:
Hello, I'd like to take this issue to onboard myself to this project.

ahoppen:
Thank you. I’m looking forward to a PR 👍🏽

joehsieh:
here is the PR for the refactoring.
https://github.com/apple/sourcekit-lsp/pull/908

ahoppen:
Fixed by https://github.com/apple/sourcekit-lsp/pull/908

PR #908, opened 2023-10-18
Move isFallback to FileBuildSettings

joehsieh:
A refactoring to move isFallback into FileBuildSettings

issue: https://github.com/apple/sourcekit-lsp/issues/861

### DIFF
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -115,7 +115,7 @@ extension BuildSystemManager {
   private func buildSettings(
     for document: DocumentURI,
     language: Language
-  ) async -> (buildSettings: FileBuildSettings, isFallback: Bool)? {
+  ) async -> FileBuildSettings? {
     do {
       // FIXME: (async) We should only wait `fallbackSettingsTimeout` for build
       // settings and return fallback afterwards. I am not sure yet, how best to
@@ -123,19 +123,21 @@ extension BuildSystemManager {
       // For now, this should be fine because all build systems return
       // very quickly from `settings(for:language:)`.
       if let settings = try await buildSystem?.buildSettings(for: document, language: language) {
-        return (buildSettings: settings, isFallback: false)
+        return settings
       }
     } catch {
       logger.error("Getting build settings failed: \(error.forLogging)")
     }
-    if let settings = fallbackBuildSystem?.buildSettings(for: document, language: language) {
+    guard var settings = fallbackBuildSystem?.buildSettings(for: document, language: language) else {
+      return nil
+    }
+    if buildSystem == nil {
       // If there is no build system and we only have the fallback build system,
       // we will never get real build settings. Consider the build settings
       // non-fallback.
-      return (buildSettings: settings, isFallback: buildSystem != nil)
-    } else {
-      return nil
+      settings.isFallback = false
     }
+    return settings
   }
 
   /// Returns the build settings for the given document.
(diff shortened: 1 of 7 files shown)


Sources/SKCore/FallbackBuildSystem.swift
```diff
@@ -51,15 +51,18 @@ public final class FallbackBuildSystem {
 
   public var indexPrefixMappings: [PathPrefixMapping] { return [] }
 
-  public func buildSettings(for uri: DocumentURI, language: Language) -> FileBuildSettings? {
+  public func buildSettings(for uri: DocumentURI, language: Language, isFallback: Bool = true) -> FileBuildSettings? {
+    var fileBuildSettings: FileBuildSettings?
```
ahoppen, line 55: Instead of passing `isFallback` here, I would prefer to set it to `true` here, since it is coming from a fallback build system and overriding it to `false` in `BuildSystem` if we don’t have a real build system.

Sources/SKCore/FileBuildSettings.swift
```diff
@@ -25,9 +25,13 @@ public struct FileBuildSettings: Equatable {
   /// The working directory to resolve any relative paths in `compilerArguments`.
   public var workingDirectory: String? = nil
 
-  public init(compilerArguments: [String], workingDirectory: String? = nil) {
+  /// The isFallback to know if the fileBuildSetting is specified or the default one.
```
ahoppen, line 28: I would phrase this slightly differently.

```suggestion
  /// Whether the build settings were computed from a real build system or whether they are synthesized fallback arguments while the build system is still busy computing build settings.
```

ahoppen (commented):
Nice. Thank you.

Sources/SKCore/FallbackBuildSystem.swift
joehsieh: right, by doing so, the func buildSettings in FallbackBuildSystem would be more clean.

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -96,27 +96,27 @@ extension BuildSystemManager {
...
     }
-    if let settings = fallbackBuildSystem?.buildSettings(for: document, language: language) {
-      // If there is no build system and we only have the fallback build system,
-      // we will never get real build settings. Consider the build settings
-      // non-fallback.
-      return (buildSettings: settings, isFallback: buildSystem != nil)
-    } else {
+    // If there is no build system and we only have the fallback build system,
+    // we will never get real build settings. Consider the build settings
+    // non-fallback.
```
ahoppen, line 114: This comment now applies to the `settings.isFallback = buildSystem != nil` line, which, just to make it even more readable would write as

```swift
if buildSystem == nil {
  // If there is no build system and we only have the fallback build system,
  // we will never get real build settings. Consider the build settings
  // non-fallback.
  settings.isFallback = false
}

**Review by ahoppen — COMMENTED** [review]
One last minor comment, then it’s ready to ship.

Also: Could you squash your commits? It makes for a nicer git history.

File: Sources/SKCore/BuildSystemManager.swift (hunk 3 above)
Comment (joehsieh) [reply Sources/SKCore/BuildSystemManager.swift:133]: agree, let me update the code.

File: Sources/SKCore/BuildSystemManager.swift [hunk 4]
```diff
@@ -148,34 +150,34 @@ extension BuildSystemManager {
...
-        )
-      }
+    let chunks = splitLongMultilineMessage(message: log, maxChunkSize: 800)
+    for (index, chunk) in chunks.enumerated() {
+      logger.log(
+        """
+        Build settings for \(document.forLogging) (\(index + 1)/\(chunks.count))
+        \(chunk)
+        """
+      )
```
joehsieh, line 178: note: this part is a new change for solving the conflicts after rebasing.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #910, opened 2023-10-19
Open documents using `testClient.openDocument` instead of constructing a `DidOpenDocumentNotification`

ahoppen:
Instead of constructing a `DidOpenTextDocumentNotification` inside the tests, add a `TestSourceKitLSPClient.openDocument` method that opens a document. In addition to being shorter to write than the `DidOpenTextDocumentNotification`, it does two smart things: 
- It infers the language of the test file from its file extension
- It removes location markers (like `1️⃣`) and returns their positions, which means that we don’t need to write fragile line-column based positions in the test files. Migrating all the tests from line-column based `Position` initializers to use location markers is quite a big change and not part of this PR.

### DIFF
--- a/Sources/SKTestSupport/LocationMarkers.swift
+++ b/Sources/SKTestSupport/LocationMarkers.swift
@@ -1,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Finds all marked ranges in the given text, see `Marker`.
+fileprivate func findMarkedRanges(text: String) -> [Marker] {
+  var markers = [Marker]()
+  while let marker = nextMarkedRange(text: text, from: markers.last?.range.upperBound ?? text.startIndex) {
+    markers.append(marker)
+  }
+  return markers
+}
+
+extension Character {
+  var isMarkerEmoji: Bool {
+    switch self {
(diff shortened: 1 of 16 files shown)


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/LocationMarkers.swift
```diff
@@ -0,0 +1,69 @@
...
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Finds all marked ranges in the given text, see `Marker`.
+fileprivate func findMarkedRanges(text: String) -> [Marker] {
```
bnbarham, line 14: I haven't checked, is this basically just a copy paste from swift-syntax 😅?

Sources/SKTestSupport/Utils.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+
+extension Language {
+  var fileExtension: String {
+    switch self {
+    case .objective_c: return "m"
```
bnbarham, line 19: `objective_cpp` could probably use a value here as well, though I assume we don't have tests with it?

bnbarham (approved):
Nice!

Sources/SKTestSupport/LocationMarkers.swift
ahoppen: Yes, this is a copy-paste from swift-syntax 😢

Sources/SKTestSupport/Utils.swift
ahoppen: Yeah, I wanted to add a case here once/if we ever have an Objective-C++ test case.

PR #911, opened 2023-10-20
Automatically initialize the SourceKit-LSP server when creating a `TestSourceKitLSPClient`

ahoppen:
Sending the `InitializeRequest` was always unnecessarily verbose. If we automatically initialize the server when creating the test client, the code becomes sufficiently concise that we can set up the client + server in the test method itself and no longer need the `setUp` methods, making the tests easier to read.

### DIFF
--- a/Tests/SourceKitLSPTests/LocalSwiftTests.swift
+++ b/Tests/SourceKitLSPTests/LocalSwiftTests.swift
@@ -24,44 +24,20 @@ import XCTest
 typealias Notification = LanguageServerProtocol.Notification
 
 final class LocalSwiftTests: XCTestCase {
-
-  /// The mock client used to communicate with the SourceKit-LSP server.
-  ///
-  /// - Note: Set before each test run in `setUp`.
-  private var testClient: TestSourceKitLSPClient! = nil
-
-  override func setUp() async throws {
-    testClient = TestSourceKitLSPClient()
-    _ = try await self.testClient.send(
-      InitializeRequest(
-        processId: nil,
-        rootPath: nil,
-        rootURI: nil,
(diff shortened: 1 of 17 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -24,44 +24,20 @@ import XCTest
...
-              codeActionLiteralSupport: .init(
-                codeActionKind: .init(valueSet: [.quickFix])
-              )
-            ),
-            publishDiagnostics: .init(codeDescriptionSupport: true)
-          )
-        ),
-        trace: .off,
-        workspaceFolders: nil
+  private let quickFixCapabilities = ClientCapabilities(
```
bnbarham, line 27: Could be static. I don't know if we want eg. a separate support enum that has the common ones we use, but quickFix seems to be used in a few places outside this test file as well.

bnbarham (approved):

Tests/SourceKitLSPTests/LocalSwiftTests.swift
ahoppen: I would like to leave these local to the test cases. In my experience what would happen otherwise is that we need another capability in addition to `quickFix` in some other test case and it just gets added to the support enum. Also, at the moment it’s only duplicated once, so not too bad.

Regarding static: I’d prefer to leave it as a member because otherwise we need to refer to it as `Self.quickFixCapabilities` which doesn’t read as nice. And it’s not like it makes a huge performance difference.

PR #912, opened 2023-10-20
Load semantic tokens from a document using a sourcekitd request instead of the 0,0 edit

ahoppen:
This allows us to cancel the semantic tokens request. It is also the last step to allow us to open and edit documents in syntactic-only mode.

It also means that we no longer need to send a `WorkspaceSemanticTokensRefreshRequest` to the client after sourcekitd has updated its semantic tokens since with the new design the `DocumentSemanticTokensRangeRequest` will simply return the results once it has the updated semantic token.


ahoppen:
https://github.com/apple/swift/pull/69294

@swift-ci Please test

Sources/SourceKitLSP/Swift/SemanticTokens.swift
```diff
@@ -15,8 +15,33 @@ import LanguageServerProtocol
...
+    skreq[keys.compilerargs] = buildSettings.compilerArgs
+
+    let dict = try await sourcekitd.send(skreq)
+
+    guard let skTokens: SKDResponseArray = dict[keys.semantic_tokens] else {
+      return nil
+    }
+    let tokenParser = SyntaxHighlightingTokenParser(sourcekitd: sourcekitd)
+    var tokens: [SyntaxHighlightingToken] = []
+    tokenParser.parseTokens(skTokens, in: snapshot, into: &tokens)
```
bnbarham, line 41: I assume there was (maybe still is?) some reason for this, but could `parseTokens` just return an array instead?

bnbarham (commented):
Exciting, so close to enabling cancellation!

Sources/SourceKitLSP/Swift/SemanticTokens.swift
```diff
@@ -15,8 +15,27 @@ import LanguageServerProtocol
...
+    
+    let skreq = SKDRequestDictionary(sourcekitd: self.sourcekitd)
+    skreq[keys.request] = requests.semantic_tokens
+    skreq[keys.sourcefile] = snapshot.uri.pseudoPath
+
+    // FIXME: SourceKit should probably cache this for us.
+    skreq[keys.compilerargs] = buildSettings.compilerArgs
+
+    let dict = try await sourcekitd.send(skreq)
+    print(dict.description)
```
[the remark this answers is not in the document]
ahoppen: Already removed (probably in my second commit).

Sources/SourceKitLSP/Swift/SemanticTokens.swift
ahoppen: `parseTokens` calls recursively into itself for substructures. and it’s easier + more efficient to accumulate tokens in an `inout Array`. At least, that’s what I think what the reasoning was.

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SemanticTokens.swift
bnbarham: That could be internal to `SyntaxHighlightingToken`, I don't see why callers have to care about that.

ahoppen: Since there’s only one caller, I didn’t care too much. But I made the `inout` functions private and added an `internal` entry point that returns the tokens.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #914, opened 2023-10-20
Add infrastructure to define indexed single-file workspaces inside the tests

ahoppen:
The new approach has a few advantages over the olde TIBS-based approach:
1. The source file being tested is defined within the test case itself and not in a separate file, which makes it easier to understand the test case since the auxiliaury file doesn’t need to be opened. Finding it inside `Sources/SKTestSupport/INPUTS` is already hard for people that are not familiar with the codebase.
2. The build setup is significantly simpler since it doesn’t rely on `ninja`. It is thus easier to understand what is run during the test.
3. We can use the emoji location markers to refer to test locations, like we do for files that are opened using `TestSourceKitLSPClient.openDocument`.

This commit only migrates call hierarchy testing to the new design. If we like it, I’ll migrate the other test workspaces as well.


bnbarham (commented):
I prefer source next to the test, so in general I'm in favor of this. I'd be interested in seeing a multi-file test or one that has additional args though.

Sources/SKTestSupport/ProcessRunner.swift
```diff
@@ -0,0 +1,189 @@
...
+    if let executableURL = executableURL {
+      message += executableURL.path
+    }
+
+    if let arguments = arguments {
+      message += " \(arguments.joined(separator: " "))"
+    }
+
+    return message
+  }
```
hamishknight, line 188: Do we need to worry about escaping here, or does it not matter since it looks like this is currently only used for verbose printing? Would it be worth naming it `commandDescription` or something to make it clear that it's only meant for printing?

hamishknight (approved):

Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public let testClient: TestSourceKitLSPClient
+  public let fileURI: DocumentURI
+  public let positions: DocumentPositions
+
+  public init(
+    _ markedText: String,
+    testName: String = #function
+  ) async throws{
```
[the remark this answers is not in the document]
ahoppen: Incidentally, that’s on of the motivation for this PR. So we can format the entire repository without the need for a `format.py` equivalent that determines which files should be formatted. Currently we can’t do that because we don’t want to format the `INPUTS` folder.

Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
```diff
@@ -0,0 +1,125 @@
...
+          commandLine: [swiftc.path] + compilerArguments
+        )
+      ]
+    )
+    let encoder = JSONEncoder()
+    encoder.outputFormatting = .prettyPrinted
+    try encoder.encode(compilationDatabase).write(to: testWorkspaceDirectory.appendingPathComponent("compile_commands.json"))
+
+    // Run swiftc to build the index store
+    try ProcessRunner(executableURL: swiftc, arguments: compilerArguments)
```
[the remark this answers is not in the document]
ahoppen: My thoughts were: Running a full compile shouldn’t be noticeably slower in these reduced test cases than just type checking and it’s closer to what a user might do + easier to understand because fewer `swiftc` arguments. That’s what I chose to just do a full compile here. Happy to change it if you disagree, though.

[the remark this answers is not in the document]
bnbarham: I was going to say that `-typecheck` is closer to what we actually do, but I suppose that's not true for LSP since we only have index while building. So doesn't really bother me either way.

Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
```diff
@@ -0,0 +1,125 @@
...
+  ///
+  /// - Note: We need this because `URL.resolvingSymlinksInPath()` not only resolves symlinks but also standardizes the
+  ///   path by stripping away `private` prefixes. Since sourcekitd is not performing this standardization, using
+  ///   `resolvingSymlinksInPath` can lead to slightly mismatched URLs between the sourcekit-lsp response and the test
+  ///   assertion.
+  var realpath: URL {
+    return self.path.withCString { path in
+      guard let realpath = Darwin.realpath(path, nil) else {
+        return self
+      }
```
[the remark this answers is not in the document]
ahoppen: Ah, forgot about that again. Luckily, we can just use `resolvingSymlinksInPath()` on those platforms because it doesn’t have the `private`-stripping issue

Sources/SKTestSupport/ProcessRunner.swift
```diff
@@ -0,0 +1,189 @@
...
+        group.leave()
+      } else {
+        addData(newData)
+      }
+    }
+  }
+}
+
+/// The exit code and output (if redirected) from a subprocess that has
+/// terminated
```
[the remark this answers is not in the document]
ahoppen: Oh, yeah. Probably pre-dates an older version of `ProcessResult` that still stored an exit code.

Sources/SKTestSupport/ProcessRunner.swift
ahoppen: Yes, it’s only used for verbose printing an it’s been good enough so far. I’ll add a comment that it’s only meant for printing.

Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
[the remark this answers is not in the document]
ahoppen: Let’s keep it with building for now.

ahoppen:
> I'd be interested in seeing a multi-file test or one that has additional args though.

I migrated a bunch more tests from the TIBS-based test inputs to define the test inputs inline. I haven’t really needed to make any adjustments to the test workspaces for the last few migrations, so I think the design is more or less complete and it’s now just a matter of converting the remaining test cases to the new infrastructure.

ahoppen:
And today I learned about `Process.checkNonZeroExit` in swift-tools-support-core. That’s sufficient for the needs of this repo so at least we don’t need to introduce another copy of `ProcessRunner.swift` here 🎉

ahoppen:
@swift-ci Please test

Sources/SKTestSupport/MultiFileTestWorkspace.swift
```diff
@@ -0,0 +1,120 @@
...
+      workspaceFolders: workspaces(scratchDirectory),
+      cleanUp: { [scratchDirectory] in
+        try? FileManager.default.removeItem(at: scratchDirectory)
+      }
+    )
+  }
+
+  /// Opens the document with the given file name in the SourceKit-LSP server.
+  ///
+  /// - Returns: The URI as which the document is opened and the positions of the location markers.
```
bnbarham, line 104:
```suggestion
  /// - Returns: The URI for the opened document and the positions of the location markers.
```

Sources/SKTestSupport/MultiFileTestWorkspace.swift
```diff
@@ -0,0 +1,120 @@
...
+  public func openDocument(_ fileName: String) throws -> (uri: DocumentURI, positions: DocumentPositions) {
+    guard let fileData = self.fileData[fileName] else {
+      throw Error.fileNotFound
+    }
+    let positions = testClient.openDocument(fileData.markedText, uri: fileData.uri)
+    return (fileData.uri, positions)
+  }
+
+  /// Returns the UIR of the file with the given name.
```
bnbarham, line 113:
```suggestion
  /// Returns the URI of the file with the given name.
```

Sources/SKTestSupport/SwiftPMTestWorkspace.swift
```diff
@@ -0,0 +1,70 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+import SKCore
+import TSCBasic
+
+public class SwiftPMTestWorkspace: MultiFileTestWorkspace {
```
bnbarham, line 18: Does this end up being significantly slower than the single file case? If not we could just use SwiftPMTestWorkspace for all the things (with a simpler constructor for the single file case).

Tests/SourceKitLSPTests/LocalClangTests.swift
```diff
@@ -181,15 +181,20 @@ final class LocalClangTests: XCTestCase {
   }
 
   func testCodeAction() async throws {
-    guard let ws = try await staticSourceKitTibsWorkspace(name: "CodeActionCxx") else { return }
-    if ToolchainRegistry.shared.default?.clangd == nil { return }
-
-    let loc = ws.testLoc("SwitchColor")
-    let endLoc = ws.testLoc("SwitchColor:end")
+    let testClient = try await TestSourceKitLSPClient()
```
bnbarham, line 184: What does clangd do with arguments? I assume we'd end up having to create a `compile_commands.json` here at some point if we ended up wanting to have more C/C++ related tests.

Sources/SKTestSupport/MultiFileTestWorkspace.swift
```diff
@@ -0,0 +1,120 @@
...
+
+  /// The file's name.
+  fileprivate let fileName: String
+
+  public init(directories: [String] = [], _ fileName: String) {
+    self.directories = directories
+    self.fileName = fileName
+  }
+
+  public init(stringLiteral value: String) {
```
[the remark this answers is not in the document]
ahoppen: Very good idea 👍🏽

Sources/SKTestSupport/SwiftPMTestWorkspace.swift
ahoppen: Unfortunately, yes. It’s about 1.3s vs 300ms, IIRC (measured yesterday). IMO it’s OK to pay that price for tests that actually need cross-file functionality and so we don’t have to use the complexity of TIBS but I don’t want to pay it for every test case.

Tests/SourceKitLSPTests/LocalClangTests.swift
ahoppen: I don’t know but I assume it just does the equivalent of `clang /path/to/file.c`.

We can always use a `MultiFileTestWorkspace` to define a compilation DB. And if that becomes common, we can always consider adding a subclass for file + compilation DB similar to how we have `SwiftPMTestWorkspace`.

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #915, opened 2023-10-20
Add support for a compile database file not at the root of a workspace

GeorgeLyon:
Introduce a new command-line flag (`--compile-db-search-path`) to `sourcekit-lsp`, which can be used to specify which subdirectories in a workspace should be searched for a clangd-style compile database (`compile_commands.json` or `compile_flags.txt`). Originally, I was going to add a `clangd`-like `--compile-commands-dir` flag, but upon further investigation decided this wasn't a good approach, as sourcekit-lsp supports multiple workspaces and `clangd` [does not](https://github.com/clangd/clangd/issues/1549). In a world with multiple workspaces it does not make sense to specify a single absolute path to a compile-database, so instead I opted for a set of relative search paths, which apply to all databases managed by a particular instance of `sourcekit-lsp`.

### DIFF
--- a/Sources/SKCore/CompilationDatabase.swift
+++ b/Sources/SKCore/CompilationDatabase.swift
@@ -15,6 +15,7 @@ import SKSupport
 
 import struct TSCBasic.AbsolutePath
 import protocol TSCBasic.FileSystem
+import struct TSCBasic.RelativePath
 import var TSCBasic.localFileSystem
 import func TSCBasic.resolveSymlinks
 
@@ -67,14 +68,28 @@ public protocol CompilationDatabase {
   var allCommands: AnySequence<Command> { get }
 }
 
-/// Loads the compilation database located in `directory`, if any.
+/// Loads the compilation database located in `directory`, if one can be found in `additionalSearchPaths` or in the default search paths of "." and "build".
 public func tryLoadCompilationDatabase(
   directory: AbsolutePath,
+  additionalSearchPaths: [RelativePath] = [],
   _ fileSystem: FileSystem = localFileSystem
 ) -> CompilationDatabase? {
+  let searchPaths =
+    additionalSearchPaths + [
+      // These default search paths match the behavior of `clangd`
+      try! RelativePath(validating: "."),
+      try! RelativePath(validating: "build"),
+    ]
   return
-    (try? JSONCompilationDatabase(directory: directory, fileSystem))
-    ?? (try? FixedCompilationDatabase(directory: directory, fileSystem))
+    try! searchPaths
+    .lazy
+    .map { directory.appending($0) }
+    .compactMap {
+      try
+        (JSONCompilationDatabase(directory: $0, fileSystem)
+        ?? FixedCompilationDatabase(directory: $0, fileSystem))
+    }
+    .first
 }
 
 /// Fixed clang-compatible compilation database (compile_flags.txt).
@@ -99,13 +114,21 @@ public struct FixedCompilationDatabase: CompilationDatabase, Equatable {
 }
 
 extension FixedCompilationDatabase {
-  public init(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database located in `directory`, if any.
+  /// - Returns: `nil` if `compile_flags.txt` was not found
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     let path = directory.appending(component: "compile_flags.txt")
     try self.init(file: path, fileSystem)
   }
 
-  public init(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database from `file`
+  /// - Returns: `nil` if the file does not exist
+  public init?(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     self.directory = file.dirname
+
+    guard fileSystem.exists(file) else {
+      return nil
+    }
     let bytes = try fileSystem.readFileContents(file)
 
     var fixedArgs: [String] = ["clang"]
@@ -185,12 +208,20 @@ extension JSONCompilationDatabase: Codable {
 }
 
 extension JSONCompilationDatabase {
-  public init(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database located in `directory`, if any.
+  ///
+  /// - Returns: `nil` if `compile_commands.json` was not found
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     let path = directory.appending(component: "compile_commands.json")
     try self.init(file: path, fileSystem)
   }
 
-  public init(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database from `file`
+  /// - Returns: `nil` if the file does not exist
+  public init?(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+    guard fileSystem.exists(file) else {
+      return nil
+    }
     let bytes = try fileSystem.readFileContents(file)
     try bytes.withUnsafeData { data in
       self = try JSONDecoder().decode(JSONCompilationDatabase.self, from: data)
--- a/Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
+++ b/Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
@@ -55,7 +55,7 @@ public final class SKSwiftPMTestWorkspace {
   /// Connection to the language server.
   public let testClient: TestSourceKitLSPClient
 
-  /// When `testServer` is not `nil`, the workspace will be opened in that server, otherwise a new server will be created for the workspace
+  /// When `testClient` is not `nil`, the workspace will be opened in that client's server, otherwise a new client will be created for the workspace
   public init(
     projectDir: URL,
     tmpDir: URL,
--- a/Sources/sourcekit-lsp/SourceKitLSP.swift
+++ b/Sources/sourcekit-lsp/SourceKitLSP.swift
@@ -22,6 +22,7 @@ import SKSupport
 import SourceKitLSP
 
 import struct TSCBasic.AbsolutePath
+import struct TSCBasic.RelativePath
 import var TSCBasic.localFileSystem
 
 extension AbsolutePath: ExpressibleByArgument {
@@ -45,6 +46,18 @@ extension AbsolutePath: ExpressibleByArgument {
     // This type is most commonly used to select a directory, not a file.
     // Specify '.file()' in an argument declaration when necessary.
     .directory
+  }
+}
+
+extension RelativePath: ExpressibleByArgument {
+  public init?(argument: String) {
+    let path = try? RelativePath(validating: argument)
+
+    guard let path = path else {
+      return nil
+    }
+
+    self = path
   }
 }
 
@@ -138,6 +151,14 @@ struct SourceKitLSP: ParsableCommand {
   var indexPrefixMappings = [PathPrefixMapping]()
 
   @Option(
+    name: .customLong("compilation-db-search-path"),
+    parsing: .singleValue,
+    help:
+      "Specify a relative path where sourcekit-lsp should search for `compile_commands.json` or `compile_flags.txt` relative to the root of a workspace. Multiple search paths may be specified by repeating this option."
+  )
+  var compilationDatabaseSearchPaths = [RelativePath]()
+
+  @Option(
     help: "Specify the directory where generated interfaces will be stored"
   )
   var generatedInterfacesPath = defaultDirectoryForGeneratedInterfaces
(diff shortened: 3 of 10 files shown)


Sources/SKCore/CompilationDatabase.swift
```diff
@@ -185,12 +197,18 @@ extension JSONCompilationDatabase: Codable {
 }
 
 extension JSONCompilationDatabase {
-  public init(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
```
GeorgeLyon, line 200: Previously, this would throw an error both if the file was missing or if parsing failed. I updated the behavior to simply return `nil` if the file does not exist, allowing me to more correctly implement the search behavior in `tryLoadCompilationDatabase` above. This makes things slightly more strict than they were, in the unlikely case that both types of database are present _and_ the JSON fails to parse this will result in an error rather than falling back to compile_flags.txt

Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
```diff
@@ -55,7 +55,7 @@ public final class SKSwiftPMTestWorkspace {
   /// Connection to the language server.
   public let testClient: TestSourceKitLSPClient
 
-  /// When `testServer` is not `nil`, the workspace will be opened in that server, otherwise a new server will be created for the workspace
+  /// When `testClient` is not `nil`, the workspace will be opened in that client's server, otherwise a new client will be created for the workspace
```
GeorgeLyon, line 58: Not really related, but I was confused by the wording here when browsing the codebase.

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -137,6 +150,18 @@ struct SourceKitLSP: ParsableCommand {
...
+  @Option(
+    name: .customLong("compilation-db-search-path", withSingleDash: false),
+    parsing: .unconditionalSingleValue,
+    help:
+      "Specify a relative path where SourceKit should search for `compile_commands.json` or `compile_flags.txt` relative to the root of a workspace. Multiple search paths may be specified by repeating this option."
+  )
+  var compilationDatabaseSearchPaths = try! [
+    // These defaults match the behavior of clangd
+    RelativePath(validating: "."),
+    RelativePath(validating: "build"),
```
GeorgeLyon, line 162: It may make sense to pull these out and make the search paths "additional" so the defaults persist even in the presence of custom search paths.

Sources/SKCore/CompilationDatabase.swift
ahoppen: That makes sense to me. Could you add a doc-comment to the initializer explaining when this returns `nil` and when it throws?

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -70,11 +71,19 @@ public protocol CompilationDatabase {
...
-  return
-    (try? JSONCompilationDatabase(directory: directory, fileSystem))
-    ?? (try? FixedCompilationDatabase(directory: directory, fileSystem))
+  return searchPaths
+    .lazy
+    .map { directory.appending($0) }
+    .compactMap {
+      try?
+        (JSONCompilationDatabase(directory: $0, fileSystem)
+        ?? FixedCompilationDatabase(directory: $0, fileSystem))
```
ahoppen, line 83: If the initializers now return `nil` when the file doesn’t exist, I think we no longer need to do `try?` here and can change it to `try`. That way we don’t silently ignore errors anymore.

Sources/SKTestSupport/SKSwiftPMTestWorkspace.swift
ahoppen: Oh, I recently changed `testServer` to `testClient` and looks like I forgot that comment. Thanks for updating it.

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -137,6 +150,18 @@ struct SourceKitLSP: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    name: .customLong("compilation-db-search-path", withSingleDash: false),
+    parsing: .unconditionalSingleValue,
```
ahoppen, line 155: Any reason why you decided to choose `unconditionalSingleValue` here instead of the default `singleValue`?

Tests/SourceKitLSPTests/CompilationDatabaseTests.swift
```diff
@@ -43,7 +43,9 @@ final class CompilationDatabaseTests: XCTestCase {
     let compilationDatabaseUrl = ws.builder.buildRoot.appendingPathComponent("compile_commands.json")
 
     _ = try ws.sources.edit({ builder in
-      let compilationDatabase = try JSONCompilationDatabase(file: AbsolutePath(validating: compilationDatabaseUrl.path))
+      let compilationDatabase = try JSONCompilationDatabase(
+        file: AbsolutePath(validating: compilationDatabaseUrl.path)
+      )!
       let newCommands = compilationDatabase.allCommands.map {
```
ahoppen, line 49: Could you use `XCTUnwrap` instead of `!` here so the test doesn’t crash if the database is `nil`?

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -137,6 +150,18 @@ struct SourceKitLSP: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    name: .customLong("compilation-db-search-path", withSingleDash: false),
```
ahoppen, line 154: `withSingleDash` defaults to `false`, so we don’t need to specify it.

Sources/sourcekit-lsp/SourceKitLSP.swift
ahoppen: Yes, I think that would make sense. IMO the search precedence should be user-defined search paths, `.`, `build`.

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -137,6 +150,18 @@ struct SourceKitLSP: ParsableCommand {
...
+    parsing: .unconditionalSingleValue,
+    help:
+      "Specify a relative path where SourceKit should search for `compile_commands.json` or `compile_flags.txt` relative to the root of a workspace. Multiple search paths may be specified by repeating this option."
```
ahoppen, line 157: Nitpick

```suggestion
      "Specify a relative path where sourcekit-lsp should search for `compile_commands.json` or `compile_flags.txt` relative to the root of a workspace. Multiple search paths may be specified by repeating this option."
```

ahoppen (commented):
Nice. Thanks for opening a PR for this so quickly 🤩.

I would like to think about the naming of the command line option a little bit more. I’m not unconditionally happy with it but don’t have any better ideas either, so maybe it’s fine. @bnbarham Maybe you’ve got an opinion here as well.

Sources/sourcekit-lsp/SourceKitLSP.swift
GeorgeLyon: No strong reason, though now that I think about it `singleValue` might be better as --compilation-db-search-path --some-other-flag may have a better compilation error.

bnbarham:
> I would like to think about the naming of the command line option a little bit more. I’m not unconditionally happy with it but don’t have any better ideas either, so maybe it’s fine. @bnbarham Maybe you’ve got an opinion here as well.

No particularly strong ones. `compile-commands-search-path` maybe?

Sources/sourcekit-lsp/SourceKitLSP.swift
GeorgeLyon: Done.

Sources/SKCore/CompilationDatabase.swift
GeorgeLyon: I looked into this but it is actually a bigger change than it seems as the `throws` will propagate all the way up to things like the `BuildSystem` protocol, so you may want to do that separately (if at all). In the meantime I made it `try!` since decoding errors should be very infrequent and shouldn't fail silently.

GeorgeLyon:
> No particularly strong ones. compile-commands-search-path maybe?

This confused me when I was looking into this. I was used to "compile_commands.json", but the term "compile database" is more accurate (and what is used in the `sourcekit-lsp` code itself), since you could use "compile_flags.txt" as well. If we were being 100% pedantic (and we are all Swift developers, no? 😜) it would be something like "clang-style-compile-database-search-path", but even that may not be 100% correct. I think the current form is a good balance between being specific and being correct and not being too wordy, and the help text is pretty explicit which helps.

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -67,14 +68,28 @@ public protocol CompilationDatabase {
...
   directory: AbsolutePath,
+  additionalSearchPaths: [RelativePath] = [],
   _ fileSystem: FileSystem = localFileSystem
 ) -> CompilationDatabase? {
+  let searchPaths =
+    try! additionalSearchPaths + [
+      // These default search paths match the behavior of `clangd`
+      RelativePath(validating: "."),
+      RelativePath(validating: "build"),
+    ]
```
ahoppen, line 82: Nitpick: I would move `try!` in front of `RelativePath` just to make it clear that the array creation or accessing `additionalPaths` can’t throw. This is just me being overly careful with `try!`.

```suggestion
    additionalSearchPaths + [
      // These default search paths match the behavior of `clangd`
      try! RelativePath(validating: "."),
      try! RelativePath(validating: "build"),
    ]
```

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -99,13 +114,24 @@ public struct FixedCompilationDatabase: CompilationDatabase, Equatable {
...
-  public init(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database located in `directory`, if any.
+  /// returns: `nil` if `compile_flags.txt` was not found
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     let path = directory.appending(component: "compile_flags.txt")
-    try self.init(file: path, fileSystem)
+    guard let database = try Self(file: path, fileSystem) else {
+      return nil
+    }
+    self = database
```
ahoppen, line 124: Am I missing something or could this just continue being

```suggestion
    try self.init(file: path, fileSystem)
```

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -185,12 +211,22 @@ extension JSONCompilationDatabase: Codable {
 }
 
 extension JSONCompilationDatabase {
-  public init(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database located in `directory`, if any.
+  /// returns: `nil` if `compile_commands.json` was not found
```
ahoppen, line 215: Let’s use docc markup here. Same for all the other `return:` lines

```suggestion
  ///
  /// - Returns: `nil` if `compile_commands.json` was not found
```

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -185,12 +211,22 @@ extension JSONCompilationDatabase: Codable {
...
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     let path = directory.appending(component: "compile_commands.json")
-    try self.init(file: path, fileSystem)
+    guard let database = try Self(file: path, fileSystem) else {
+      return nil
+    }
+    self = database
```
ahoppen, line 221: Also here, couldn’t this be

```suggestion
    try self.init(file: path, fileSystem)
```

Sources/sourcekit-lsp/SourceKitLSP.swift
ahoppen: That’s what I thought as well.

ahoppen:
> No particularly strong ones. `compile-commands-search-path` maybe?

I think `compilation-db-search-path`  is better because it also handles `compile_flags.txt`. For the lack of better ideas, let’s leave it for now.

ahoppen (commented):
I have a couple more minor formatting comments inline.

Sources/SKCore/CompilationDatabase.swift
GeorgeLyon: Good catch. I've had one foot outside of Swift land the year or so and forgot we can have nice things :)

Sources/SKCore/CompilationDatabase.swift
```diff
@@ -99,13 +114,21 @@ public struct FixedCompilationDatabase: CompilationDatabase, Equatable {
...
+  /// Loads the compilation database located in `directory`, if any.
+  /// returns: `nil` if `compile_flags.txt` was not found
+  public init?(directory: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
     let path = directory.appending(component: "compile_flags.txt")
     try self.init(file: path, fileSystem)
   }
 
-  public init(file: AbsolutePath, _ fileSystem: FileSystem = localFileSystem) throws {
+  /// Loads the compilation database from `file`
+  /// returns: `nil` if the file does not exist
```
ahoppen, line 125: Could you also update the other `returns:` lines to use docc markup?

ahoppen (commented):
Just one more note on the doc comments. 

Also: Could you squash your commits? It makes for a nicer git history.

GeorgeLyon:
Updated comments, and squash-force0pushed... do you guys not use GitHub's "squash and merge"?

ahoppen:
> Updated comments, and squash-force0pushed... do you guys not use GitHub's "squash and merge"?

I prefer not to squash and merge for the reasons outlined here: https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits

ahoppen (approved):
Nice. Let’s get this in 🚢

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

GeorgeLyon:
Thanks for working with me on this @ahoppen!

ahoppen:
Thanks for your contribution!

PR #916, opened 2023-10-20
Add a VSCode devcontainer definition

GeorgeLyon:
I'm not sure if this is interesting, but when putting together [this PR](https://github.com/apple/sourcekit-lsp/pull/915) I did my development in a devcontainer (instructions provided in a Readme). As I'm often managing many different environments and tools (and VSCode extensions), I really like the devcontainer workflow and on M-series Macs the performance is quite good.


GeorgeLyon:
One thing that I wasn't able to figure out is if we can add the `-Xcxx -I/usr/lib/swift -Xcxx -I/usr/lib/swift/Block` arguments somewhere global in the devcontainer, which I believe would enable using VSCode's test explorer with this configuration which would be very nice.

adam-fowler:
@GeorgeLyon Do you know about the `Add Dev Container Configuration Files...` command?
If you 
- select `From a predefined container configuration ...`
- select `Show All Definitions...`
- Type `swift` into the filter box
- Select the `Swift` devcontainer
You get the recommended swift container. You can find the template here https://github.com/swift-server/swift-devcontainer-template
- There are also a number of features you can add, including both swift formats, sqlite, foundation networking and jemalloc

When you talk about `VSCode's test explorer` which test explorer are you talking about? There are multiple test explorers.

GeorgeLyon:
Thanks for the info! This was adapted from a slightly more complicated devcontainer I had on another project, but knowing there is a default Swift template is nice. In my other project I also used the [devcontainers/ci action](https://github.com/GeorgeLyon/chisel/blob/dev/george/firrtl-bindings/.github/workflows/check-circel.yml) to run CI on the devcontainer which I really like. 

I haven't used features much, but I opted to use the Dockerfile to install most things to keep things simple and more standard (I'm pretty sure Features are a devcontainer thing not a Docker thing). That said... I think they provide some interesting functionality and would be worth looking into. 

One not on the Swift devcontainer template... I also really like using `zsh` in my devcontainers, but I do it in a slightly-bad way using a [dotfiles repo](https://github.com/GeorgeLyon/devcontainer-dotfiles) and the `"dotfiles.repository":` option. This way I can separate "things actually related to the project" and "things George likes" (the "default devcontainer extensions" option helps with this too)

GeorgeLyon:
> When you talk about VSCode's test explorer which test explorer are you talking about? There are multiple test explorers.

I'm talking about the default "beaker" one in the UI. I gather sourcekit-lsp opts in to some "I detect tests" API which populates it. Unfortunately, because you need to specify extra commandline flags for the `sourcekit-lsp` project (which I could find no way to specify for running from the test explorer), the tests fail to build correctly. 
image

Perhaps the right thing to do here would be to make a package-config file named something like "swiftToolchain" that adds these flags. Might give it a shot later today.

adam-fowler:
That Test Explorer is generated by the swift extension. Go into the settings and you can add additional build arguments in the Swift: Build Arguments section. Make sure you only do it for the current workspace.

As an alternative you can add a tasks.json entry with command line parameters. Press Cmd+Shift+B and click on cog symbol at right side of menu entry for the build you want to edit.

Both of these options create files in the `.vscode` folder a `settings.json` or a `tasks.json` respectively. Perhaps one of these should be committed to the source kit-lsp repo

GeorgeLyon:
>Go into the settings and you can add additional build arguments in the Swift: Build Arguments section
Oh man, I missed this because I only checked the sections _nested under_ "Swift" in the extensions UI. I can confirm this works, and I added this to the devcontainer "settings" section, which writes it to the "Remote Settings" section in the devcontainer. Now the Test Explorer UI works out-of-the-box in the devcontainer, which is pretty cool.

Not sure this is the best approach as it is workspace-specific. The approach I've been taking in my own projects is to [bind a different directory at `.vscode` in the devcontainer](https://github.com/GeorgeLyon/chisel/blob/8cebbe4c3fc3760409a7bf622738e0011a342221/.devcontainer/circel/devcontainer.json#L11). This works pretty well if you commit to not having per-project VSCode settings, tasks or launch jsons (though you can still achieve this with some symlinking). I like this approach because I almost always need to tweak some setting in different devcontainer environments.

GeorgeLyon:
Also all that being said, I think the `pkg-config` approach is actually the "correct" way to handle these sorts of arguments. Setting this for the whole project won't work as the path is different depending on what system you are on and what toolchain you are using. Conceptually, you are depending on a library that is installed on the system, much the same way you would depend on something like `sqlite3` if that was installed on the system. In a devcontainer this would work well since you likely would never need more than one Swift toolchain installed, and could just create a `pkg-config` for "the Swift toolchain installed on my system".
A more robust solution would be for the toolchain itself to provide a "this toolchain" `.systemLibrary` (or even a new enum, `.toolchainLibrary` or something like that) which resolves correctly even in an environment where you have multiple toolchains installed. Not something I have time to chase down at the moment, so just leaving my thoughts here.

ahoppen:
Sorry for taking a while to look at this. I think having a devcontainer is an exciting idea and it should simplify getting started with sourcekit-lsp development, especially on Linux where the setup is non-trivial.

I am not sure if I followed the entire conversation but are there any open questions left? With regards to the `Dockerfile`, I would prefer if it follows the default template because that way we have one standardized Dockerfile setting for new packages adding a container and sourcekit-lsp.

GeorgeLyon:
> With regards to the Dockerfile, I would prefer if it follows the default template

Not sure what you mean by "follows the default template". The swift template @adam-fowler was referring to is a template for adding these files (devcontainer.json, etc) to a Swift project which doesn't have them. It directly references a Docker image by name, which is insufficient for developing sourcekit-lsp. My Dockerfile uses an official image as the _base_ of the devcontainer (the [`FROM`](https://github.com/apple/sourcekit-lsp/blob/4fa705d82bd4eeaa3d2506d911a4b3464f526477/.devcontainer/default/Dockerfile#L1) command) and then installs the dependencies of sourcekit-lsp on top of that (the [`RUN`](https://github.com/apple/sourcekit-lsp/blob/4fa705d82bd4eeaa3d2506d911a4b3464f526477/.devcontainer/default/Dockerfile#L3) command).

One way to think about it is the template is a way to get a reasonable starting point when adding a new devcontainer to a project that doesn't have one, and this PR is a fleshed-out devcontainer that actually works for sourcekit-lsp.

ahoppen:
Oh, sorry. I misunderstood you there. From the discussion I thought that `Add Dev Container Configuration Files…` added a `Dockerfile`. Ignore my comment then about the `Dockerfile` then.

Are there any open questions/tasks on this PR or is this ready for review as far as you are concerned @GeorgeLyon? Just trying to make sure I’m not commenting on something that’s not fully ready yet.

GeorgeLyon:
Not from my end, I just have really liked the devcontainer experience recently and this is just the setup I used to dabble in this repo, so if it can be helpful to someone else, great!

ahoppen:
Definitely! I agree, having a standardized installation setup is great. I’d like to test the devcontainer myself but it might take me a while until I get to it. I’ll come back to to this PR once I tried it.

ahoppen:
@GeorgeLyon I am really sorry how long it took us to get back to you on this PR. I wanted to push this forward for the last few months and was only now able to set up an environment in which I could test the dev container. Would you still be interested in continuing the PR? If so, could you re-open it? If you don’t have time to work on it anymore, I understand that as well.

GeorgeLyon:
Happy to keep it open, I was just cleaning up my "open PRs" list .

ahoppen:
Sure, I totally understand. Don’t blame you for assuming that there was no interest on our side.

.devcontainer/default/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+  #
+  # Disable apt interactive prompts for this RUN command
+  export DEBIAN_FRONTEND="noninteractive" && \
+    # 
+    # Update apt package list
+    apt-get update && \
+    # 
+    # Install sourcekit-lsp dependencies
+    apt-get install -y \
+      libsqlite3-dev libncurses5-dev python3 ninja-build && \
```
ahoppen, line 13: It’s been a while since the PR was opened and `ninja-build` is no longer a dependency. So I think we can remove that.

.devcontainer/default/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    #
+    # Install swift-format
+    apt-get install -y \
+      curl && \
+    mkdir /swift-format && \
+    curl -L https://github.com/apple/swift-format/archive/refs/heads/release/5.9.tar.gz | tar xz --strip-component=1 -C /swift-format && \
+    swift build --package-path /swift-format
+
+ENV \
+  PATH="/swift-format/.build/debug:${PATH}"
```
ahoppen, line 23: With https://github.com/apple/sourcekit-lsp/pull/1096, the recommended way of running swift-format on the repo is to use the _Run swift-format_ task, so we shouldn’t need this anymore.

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+  "capAdd": [ "SYS_PTRACE" ],
+  "securityOpt": [ "seccomp=unconfined" ],
+
+  "mounts": [
+    // Use a named volume for the build products for optimal performance (https://code.visualstudio.com/remote/advancedcontainers/improve-performance?WT.mc_id=javascript-14373-yolasors#_use-a-targeted-named-volume)
+    "source=${localWorkspaceFolderBasename}-build,target=${containerWorkspaceFolder}/build,type=volume"
+  ],
+  "remoteEnv": {
+    // `/usr/bin` must be first: https://github.com/apple/sourcekit-lsp/blob/main/Documentation/Development.md#linux
+    "PATH": "/usr/bin:${containerWorkspaceFolder}/.build/bin:${containerEnv:PATH}" },
```
ahoppen, line 16: This also isn’t necessary anymore. SourceKit-LSP now prefers a toolchain if it is a strict superset of another toolchain: https://github.com/apple/sourcekit-lsp/pull/1034

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "sswg.swift-lang",
+        "vknabel.vscode-apple-swift-format"
```
ahoppen, line 21: Similar to my comment above, swift-format should be run using the *Run swift-format* action, so this shouldn’t be needed anymore.

Sources/SKTestSupport/INPUTS/.gitignore
ahoppen, line 1: I think this file also shouldn’t be needed anymore because we moved away from the test fixtures defined in `SKTestSupport/INPUTS`

.devcontainer/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+      - "Choose file sharing implementation for your containers": VirtioFS (better IO performance)
+    - Resources:
+      - CPUs: Allow docker to use most or all of your CPUs
+      - Memory: Allow docker to use most or all of your memory
+4. Open up this repository in VSCode
+  - VSCode has an "Install 'code' command in PATH" command which installs a helpful tool to open VSCode from the commandline (`code <path-to-this-repo>`)
+5. Select "Reopen in Container" in the notification that pops up
+  - If there is no popup you can manually open the dev container by selecting "Dev Containers: Rebuild and Reopen in Container" from the command palette (Command-Shift-P on macOS)
+  - Occasionally, after pulling from git, VSCode may prompt you to rebuild the container if the container definition has changed
+6. Wait for the container to be built
```
ahoppen, line 23: I would prefer linking to a tutorial (or similar) instead of duplicating information about how to set up VS Code with dev containers in this repository because such documentation always has the tendency to get out-of-date. What do you think about linking to https://code.visualstudio.com/docs/devcontainers/tutorial?

I think the only specific things not mentioned in the tutorial would be the following:

```md
  - Recommended settings for macOS (some of these are defaults):
    - General:
      - "Choose file sharing implementation for your containers": VirtioFS (better IO performance)
    - Resources:
      - CPUs: Allow docker to use most or all of your CPUs
      - Memory: Allow docker to use most or all of your memory
```

GeorgeLyon: Links are also sometimes broken, which is why I both linked to it and wrote up the short version (this was actually just copy-pasta from another repo of mine). Happy to just use your copy, but I don't the current copy will be problematic.

.devcontainer/default/devcontainer.json
GeorgeLyon: This extension uses `swift-format` from the project if it is a dependency. I like including it because it makes other VSCode functionality (like format-on-save) work correctly.

.devcontainer/default/Dockerfile
GeorgeLyon: I think you from the future disagrees :) but yes since `swift-format` is now a dependency this is not strictly necessary
https://github.com/apple/sourcekit-lsp/commit/7c9a6e47e1a999890b88a345f02d31d3e71820a8

.devcontainer/default/devcontainer.json
adam-fowler: You can get format on save to work using SourceKit-LSP in Swift 6.0. The vscode extension shouldn't be needed anymore.

GeorgeLyon: You are correct! Sorry I didn't realize this was added.

.devcontainer/Readme.md
ahoppen: My preference would be to just reference the tutorial and just add the SourceKit-LSP-specific details here. My opinions is that we shouldn’t try to re-do work that somebody else has already done.

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+// Reference: https://containers.dev/implementors/json_reference/
+{
+  "name": "SourceKit LSP",
```
ahoppen, line 3: Nitpick 

```suggestion
  "name": "SourceKit-LSP",
```

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+  "dockerFile": "Dockerfile",
+
+  // Allow the processes in the container to attach a debugger
+  "capAdd": [ "SYS_PTRACE" ],
+  "securityOpt": [ "seccomp=unconfined" ],
+
+  "mounts": [
+    // Use a named volume for the build products for optimal performance (https://code.visualstudio.com/remote/advancedcontainers/improve-performance?WT.mc_id=javascript-14373-yolasors#_use-a-targeted-named-volume)
+    "source=${localWorkspaceFolderBasename}-build,target=${containerWorkspaceFolder}/build,type=volume"
```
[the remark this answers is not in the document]
GeorgeLyon: Yes, good catch!

.devcontainer/default/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    # 
+    # Update apt package list
+    apt-get update && \
+    # 
+    # Install sourcekit-lsp dependencies
+    apt-get install -y \
+      libsqlite3-dev libncurses5-dev python3
+
+ENV \
+  PATH="/swift-format/.build/debug:${PATH}"
```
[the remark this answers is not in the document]
GeorgeLyon: Yup

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,33 @@
...
+{
+  "name": "SourceKit-LSP",
+  "dockerFile": "Dockerfile",
+
+  // Allow the processes in the container to attach a debugger
+  "capAdd": [ "SYS_PTRACE" ],
+  "securityOpt": [ "seccomp=unconfined" ],
+
+  "mounts": [
+    // Use a named volume for the build products for optimal performance (https://code.visualstudio.com/remote/advancedcontainers/improve-performance?WT.mc_id=javascript-14373-yolasors#_use-a-targeted-named-volume)
```
ahoppen, line 11: Do you know where the `WT.mc_id=javascript-14373-yolasors` in the URL is coming from? Removing it seems to work as well

```suggestion
    // Use a named volume for the build products for optimal performance (https://code.visualstudio.com/remote/advancedcontainers/improve-performance#_use-a-targeted-named-volume)
```

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,33 @@
...
+    "source=${localWorkspaceFolderBasename}-build,target=${containerWorkspaceFolder}/.build,type=volume",
+    // Do the same for experimental background indexing
+    "source=${localWorkspaceFolderBasename}-build,target=${containerWorkspaceFolder}/.index-build,type=volume"
```
ahoppen, line 14: Does this volume need to have a different name? At the moment it seems like `.build` and `.index-build` are mounted to the same volume, which means that they have the same contents.

```suggestion
    "source=${localWorkspaceFolderBasename}-index-build,target=${containerWorkspaceFolder}/.index-build,type=volume"
```

.devcontainer/default/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+  #
+  # Disable apt interactive prompts for this RUN command
+  export DEBIAN_FRONTEND="noninteractive" && \
+    # 
+    # Update apt package list
+    apt-get update && \
+    # 
+    # Install sourcekit-lsp dependencies
+    apt-get install -y \
+      libsqlite3-dev libncurses5-dev python3
```
ahoppen, line 13: A little nitpicky but I think it would be more readable with a little fewer newlines. And flushing it to the left seems to match VS Code’s formatter for Dockerfiles.

```suggestion
RUN \
  # Disable apt interactive prompts for this RUN command
  export DEBIAN_FRONTEND="noninteractive" && \
  # Update apt package list
  apt-get update && \
  # Install sourcekit-lsp dependencies
  apt-get install -y libsqlite3-dev libncurses5-dev python3
```

ahoppen (commented):
Nice. I have three more minor comments, otherwise, let’s get this shipped 🚢

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,33 @@
...
+    // Do the same for experimental background indexing
+    "source=${localWorkspaceFolderBasename}-index-build,target=${containerWorkspaceFolder}/.index-build,type=volume"
+  ],
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "sswg.swift-lang"
+      ],
+      "settings": {
+        "lldb.launch.expressions": "native",
```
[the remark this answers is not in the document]
GeorgeLyon: I don't know if this has changed, but when the extension does it for you it prompts you where to store the setting, with one of the option (I believe the default) being "Global" and the other being "Workspace". *Both* will conflict with the setting needed for opening the workspace locally on macOS, and setting it here has it end up in the "remote" settings which is the correct place for it. If this setting is not what the extension expect I believe it will still prompt you to change it.

.devcontainer/default/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM swiftlang/swift:nightly-main-jammy
```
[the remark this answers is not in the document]
GeorgeLyon: I don't have strong thoughts, but when I put this together I believe the guidance was that I should be using a "recent nightly". This should follow whatever the guidance is currently.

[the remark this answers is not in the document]
adam-fowler: @ahoppen can make the call. I guess if it is for development maybe it should stay at main

.devcontainer/default/devcontainer.json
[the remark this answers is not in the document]
adam-fowler: The extension prompts you but it'll have already prompted you outside of the devcontainer.

.devcontainer/default/devcontainer.json
```diff
@@ -0,0 +1,33 @@
+// Reference: https://containers.dev/implementors/json_reference/
+{
+  "name": "SourceKit-LSP",
+  "dockerFile": "Dockerfile",
+
```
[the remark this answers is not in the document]
adam-fowler: I guess it's a security issue. All of Microsoft's language images have a vscode user. And they use the common-utils feature to set it up. So there shouldn't be any issues. I guess in theory you shouldn't need the git feature as git is already in the swift image.

.devcontainer/default/Dockerfile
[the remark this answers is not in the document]
ahoppen: We should use the latest `main` because the `main` branch of SourceKit-LSP might be using features from sourcekitd, that are only available in `main`. When running SourceKit-LSP tests with an older version of sourcekitd, those tests would be skipped.

.devcontainer/default/devcontainer.json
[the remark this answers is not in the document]
ahoppen: I would prefer to keep the `.devcontainer.json` file minimal. You will be running everything as `root` if you’re using the `swift:nightly-main-jammy` docker image, so I don’t think there’s any inherent harm in also running as `root` inside the devcontainer.

.devcontainer/default/devcontainer.json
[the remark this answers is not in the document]
ahoppen: Let’s remove it then. Fewer settings is always good.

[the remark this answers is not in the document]
GeorgeLyon: OK, for the record I'm not removing the setting below because it actually needs to be different inside the container vs out.

.devcontainer/default/devcontainer.json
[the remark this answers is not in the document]
GeorgeLyon: I'm not an expert, but AFAICT the "not running as root" stuff is more to do with software in the container making bad assumptions and not running properly as root than anything to do with security.

ahoppen (approved):
Wohoooo. Let’s get this in 🎉

ahoppen:
@swift-ci Please test

.devcontainer/default/devcontainer.json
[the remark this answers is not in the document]
adam-fowler: Yes the setting below is required

Issue #868, opened 2023-10-09
Convert `ManualMainFilesProvider` and `BSMDelegate` to be actors


ahoppen:
Both `ManualMainFilesProvider` and `BSMDelegate` in `BuildSystemManagerTests.swift` are classes with a queue. We should be able to convert both of them to actors.

ahoppen:
Tracked in Apple’s issue tracker as rdar://116703668

joehsieh:
may i take this issue if it's still unassigned?

ahoppen:
Sure, thank you!

joehsieh:
**[Discussion]**
Hello @ahoppen i'd like to discuss an issue that i found after the investigation.
There are two classes conforming protocol `MainFilesProvider`. One (ManualMainFilesProvider) is for the testing, another one is for the implementation (IndexStoreDB). I noticed the serial queue in ManualMainFilesProvider is just only for the purpose of testing not for the implementation. (can check Investigation result below)
Hence, once we convert the queue to be an actor. For the same purpose of testing, the func mainFilesContainingFile in the protocol MainFilesProvider would be changed ([the draft PR](https://github.com/apple/sourcekit-lsp/pull/917))
```
from 
func mainFilesContainingFile(_: DocumentURI) -> Set<DocumentURI>
to 
func mainFilesContainingFile(_: DocumentURI) async -> Set<DocumentURI>
```
to allow actors to conform, which is unexpected because `async` is unnecessary for the implementation. 

Instead of converting ManualMainFilesProvider to be an actor, i think we can make mainFiles immutable and update related code in BuildSystemManagerTests. (i'll give it a try soon) What do you think? 

**Investigation result:**
- class ManualMainFilesProvider

mainFiles is a member variable, and we need a serial queue to maintain it make sure it's thread-safe since mainFiles is possibly updated with the setter of mainFiles in unit testing.

```
private final class ManualMainFilesProvider: MainFilesProvider {
  let lock: DispatchQueue = DispatchQueue(label: "\(ManualMainFilesProvider.self)-lock")
  private var _mainFiles: [DocumentURI: Set<DocumentURI>] = [:]
  var mainFiles: [DocumentURI: Set<DocumentURI>] {
    get { lock.sync { _mainFiles } }
    set { lock.sync { _mainFiles = newValue } }
  }

  func mainFilesContainingFile(_ file: DocumentURI) -> Set<DocumentURI> {
    if let result = mainFiles[file] {
      return result
[...]

PR #917, opened 2023-10-21
Convert ManualMainFilesProvider and BSMDelegate to be actors

joehsieh:
issue: https://github.com/apple/sourcekit-lsp/issues/868

- Convert ManualMainFilesProvider to be an actor
- Remove an unused queue from BSMDelegate

### DIFF
--- a/Tests/SKCoreTests/BuildSystemManagerTests.swift
+++ b/Tests/SKCoreTests/BuildSystemManagerTests.swift
@@ -25,13 +25,14 @@ final class BuildSystemManagerTests: XCTestCase {
     let c = DocumentURI(string: "bsm:c")
     let d = DocumentURI(string: "bsm:d")
 
-    let mainFiles = ManualMainFilesProvider()
-    mainFiles.mainFiles = [
-      a: Set([c]),
-      b: Set([c, d]),
-      c: Set([c]),
-      d: Set([d]),
-    ]
+    let mainFiles = ManualMainFilesProvider(
+      [
+        a: [c],
+        b: [c, d],
+        c: [c],
+        d: [d],
+      ]
+    )
 
     let bsm = await BuildSystemManager(
       buildSystem: nil,
@@ -55,13 +56,9 @@ final class BuildSystemManagerTests: XCTestCase {
     await assertEqual(bsm._cachedMainFile(for: c), c)
     await assertEqual(bsm._cachedMainFile(for: d), d)
 
-    mainFiles.mainFiles = [
-      a: Set([a]),
-      b: Set([c, d, a]),
-      c: Set([c]),
-      d: Set([d]),
-    ]
-
+    await mainFiles.updateMainFiles(for: a, to: [a])
+    await mainFiles.updateMainFiles(for: b, to: [c, d, a])
+  
     await assertEqual(bsm._cachedMainFile(for: a), c)
     await assertEqual(bsm._cachedMainFile(for: b), bMain)
     await assertEqual(bsm._cachedMainFile(for: c), c)
@@ -92,8 +89,7 @@ final class BuildSystemManagerTests: XCTestCase {
 
   func testSettingsMainFile() async throws {
     let a = DocumentURI(string: "bsm:a.swift")
-    let mainFiles = ManualMainFilesProvider()
-    mainFiles.mainFiles = [a: Set([a])]
+    let mainFiles = ManualMainFilesProvider([a: [a]])
     let bs = ManualBuildSystem()
     let bsm = await BuildSystemManager(
       buildSystem: bs,
@@ -116,8 +112,7 @@ final class BuildSystemManagerTests: XCTestCase {
 
   func testSettingsMainFileInitialNil() async throws {
     let a = DocumentURI(string: "bsm:a.swift")
-    let mainFiles = ManualMainFilesProvider()
-    mainFiles.mainFiles = [a: Set([a])]
+    let mainFiles = ManualMainFilesProvider([a: [a]])
     let bs = ManualBuildSystem()
     let bsm = await BuildSystemManager(
       buildSystem: bs,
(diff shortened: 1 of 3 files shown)


Tests/SKCoreTests/BuildSystemManagerTests.swift
```diff
@@ -264,7 +263,13 @@ final class BuildSystemManagerTests: XCTestCase {
...
     assertEqual(await bsm.buildSettingsInferredFromMainFile(for: h, language: .c), bs.map[cpp1]!)
 
-    mainFiles.mainFiles[h] = Set([cpp2])
+    await mainFiles.setMainFiles(
+      mainFiles: [
+        h: Set([cpp2]),
+        cpp1: Set([cpp1]),
+        cpp2: Set([cpp2]),
+      ]
+    )
```
ahoppen, line 272: We shouldn’t need to re-state the main files for `cpp1` and `cpp2` here. What you think of having the following function instead of `setMainFiles`?

```swift
func updateMainFiles(for file: DocumentURI, to mainFiles: Set<DocumentURI>)
```

Tests/SKCoreTests/BuildSystemManagerTests.swift
```diff
@@ -342,8 +360,7 @@ final class BuildSystemManagerTests: XCTestCase {
     let a = DocumentURI(string: "bsm:a.swift")
     let b = DocumentURI(string: "bsm:b.swift")
     let c = DocumentURI(string: "bsm:c.swift")
-    let mainFiles = ManualMainFilesProvider()
-    mainFiles.mainFiles = [a: Set([a]), b: Set([b]), c: Set([c])]
+    let mainFiles = ManualMainFilesProvider(mainFiles: [a: Set([a]), b: Set([b]), c: Set([c])])
```
ahoppen, line 363: I think (haven’t checked) that we don’t need to add `Set` here. We should be able to infer that the elements are a set from the parameter of `ManualMainFilesProvider`. Same in a few other locations.

```suggestion
    let mainFiles = ManualMainFilesProvider(mainFiles: [a: [a], b: [b], c: [c]])
```

Tests/SKCoreTests/BuildSystemManagerTests.swift
```diff
@@ -412,15 +428,18 @@ final class BuildSystemManagerTests: XCTestCase {
...
+  init(mainFiles: [DocumentURI: Set<DocumentURI>]) {
+    self.mainFiles = mainFiles
+  }
+
+  func setMainFiles(mainFiles: [DocumentURI: Set<DocumentURI>]) async {
+    self.mainFiles = mainFiles
   }
 
-  func mainFilesContainingFile(_ file: DocumentURI) -> Set<DocumentURI> {
+  func mainFilesContainingFile(_ file: DocumentURI) async -> Set<DocumentURI> {
```
ahoppen, line 442: Does this function have to be `async`. I don’t see any `await` calls within it.

ahoppen (commented):
Thank you! Just a few suggestions to make the code more readable.

Tests/SKCoreTests/BuildSystemManagerTests.swift
joehsieh: true, let me update the code

Tests/SKCoreTests/BuildSystemManagerTests.swift
joehsieh: it was copied from the protocol. let me remove it.

Tests/SKCoreTests/BuildSystemManagerTests.swift
```diff
@@ -412,16 +404,19 @@ final class BuildSystemManagerTests: XCTestCase {
...
 
 /// A simple `MainFilesProvider` that wraps a dictionary, for testing.
-private final class ManualMainFilesProvider: MainFilesProvider {
-  let lock: DispatchQueue = DispatchQueue(label: "\(ManualMainFilesProvider.self)-lock")
-  private var _mainFiles: [DocumentURI: Set<DocumentURI>] = [:]
-  var mainFiles: [DocumentURI: Set<DocumentURI>] {
-    get { lock.sync { _mainFiles } }
-    set { lock.sync { _mainFiles = newValue } }
+private final actor ManualMainFilesProvider: MainFilesProvider {
+  private var mainFilesDictionary: [DocumentURI: Set<DocumentURI>]
```
ahoppen, line 408: One last comment: `Dictionary` is superfluous here because the type is already a dictionary. I would just name it `mainFiles`.

Tests/SKCoreTests/BuildSystemManagerTests.swift
```diff
@@ -412,16 +404,19 @@ final class BuildSystemManagerTests: XCTestCase {
...
+
+  init(mainFilesDictionary: [DocumentURI: Set<DocumentURI>]) {
+    self.mainFilesDictionary = mainFilesDictionary
+  }
```
ahoppen, line 412: And I think we can make the parameter unnamed here. Since we are constructing a `ManualMainFilesProvider`, it’s clear that we are passing the main files, IMO. What do you think?

ahoppen (commented):
I have one more comment on the naming of `mainFilesDictionary`. Sorry for not catching that in the first review. Otherwise, looks very good to me 👍🏽

Tests/SKCoreTests/BuildSystemManagerTests.swift
joehsieh: The intention for the renaming is to avoid the confusion between `mainFiles: [DocumentURI: Set<DocumentURI>]`
`mainFiles: Set<DocumentURI>` https://github.com/apple/sourcekit-lsp/pull/917/files#diff-4e5dc27e47b431520ccbd13ad2e859517644e3437c73157b2193a7b14ec30570R414

it was a trade-off when i named it, but yeah the data type already convey itself, so let me rename it.

Tests/SKCoreTests/BuildSystemManagerTests.swift
joehsieh: agree. let me update the code.

Tests/SKCoreTests/BuildSystemManagerTests.swift
ahoppen: I think that’s fine. `updateMainFiles` also is a really short methods and I don’t think it should cause too much confusion.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #918, opened 2023-10-23
Validate ClangWorkspaceSettings before preferring it over unknown WorkspaceSettingsChanges

JCWasmx86:
Basically everything will be decoded as ClangWorkspaceSettings, even if it has nothing to do wit hit, yielding for example:
```
clangd(LanguageServerProtocol.ClangWorkspaceSettings(compilationDatabasePath: nil, compilationDatabaseChanges: nil))
```
This breaks at least my case of using WorkspaceSettingsChange. This patch adds a - maybe bandaid - fix for that.

Furthermore I removed one test, as it - AFAIU - broke the invariant that none of the attributes of ClangWorkspaceSettings may be nil.

I'm not sure whether there would be a better solution, but this patch works for my use.

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/WorkspaceSettings.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/WorkspaceSettings.swift
@@ -19,10 +19,7 @@ public enum WorkspaceSettingsChange: Codable, Hashable {
   case unknown(LSPAny)
 
   public init(from decoder: Decoder) throws {
-    // FIXME: doing trial deserialization only works if we have at least one non-optional unique
-    // key, which we don't yet.  For now, assume that if we add another kind of workspace settings
-    // it will rectify this issue.
-    if let settings = try? ClangWorkspaceSettings(from: decoder) {
+    if let settings = try? ClangWorkspaceSettings(from: decoder), settings.isValid {
       self = .clangd(settings)
     } else {
       let settings = try LSPAny(from: decoder)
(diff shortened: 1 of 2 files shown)


Sources/LanguageServerProtocol/SupportTypes/WorkspaceSettings.swift
```diff
@@ -59,6 +56,11 @@ public struct ClangWorkspaceSettings: Codable, Hashable {
     self.compilationDatabasePath = compilationDatabasePath
     self.compilationDatabaseChanges = compilationDatabaseChanges
   }
+
+  internal func isValid() -> Bool {
```
ahoppen, line 60: This can be a computed property, making the code a little more swifty.

```suggestion
  var isValid: Bool {
```

Sources/LanguageServerProtocol/SupportTypes/WorkspaceSettings.swift
```diff
@@ -59,6 +56,11 @@ public struct ClangWorkspaceSettings: Codable, Hashable {
...
+    return (self.compilationDatabasePath != nil && self.compilationDatabaseChanges == nil)
+            || (self.compilationDatabasePath == nil && self.compilationDatabaseChanges != nil)
```
ahoppen, line 62: How do you feel about making this a `switch` so we don’t need to repeat the property names?

```suggestion
    switch (compilationDatabasePath, compilationDatabaseChanges) {
      case (nil, .some), (.some, nil): return true
      default: return false
    }
```

ahoppen (commented):
The change looks good to me. Just two suggestions on code style.

ahoppen (approved):
Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
Oh, sorry, thanks for the ping. I forgot to set myself a reminder to check again after CI passed. There’s nothing use from merging it.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
Yes, the failure seems unrelated. Let’s try again.

@swift-ci Please test Linux

ahoppen:
Sorry about missing the PR testing a few times and that it took a while to get this change in.

PR #920, opened 2023-10-24
Migrate the remaining TIBS tests to the new testing infrastructure that defines sources inline

ahoppen:
With this PR in, there are only main files test remaining that still use the TIBS test infrastructure. These tests are currently disabled because of rebranch failures. I’d like to fix them first before migrating them, which I’ll do in another PR.


Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
```diff
@@ -46,7 +53,7 @@ public struct IndexedSingleSwiftFileWorkspace {
     var compilerArguments: [String] = [
       testFileURL.path,
       "-index-store-path", indexURL.path,
-      "-index-ignore-system-modules",
```
bnbarham, line 49: Did we end up needing system modules in a test? Could pass that in to avoid the extra time when it isn't needed.

Sources/SKTestSupport/Utils.swift
```diff
@@ -51,7 +51,6 @@ public func testScratchDir(testName: String = #function) throws -> URL {
   let url = FileManager.default.temporaryDirectory
     .realpath
     .appendingPathComponent("sourcekit-lsp-test-scratch")
-    .appendingPathComponent(UUID().uuidString)
```
bnbarham, line 54: If we're not adding the UUID, it might be worth deleting the folder if it exists (probably in each workspace?)

Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
ahoppen: Good idea.

Sources/SKTestSupport/Utils.swift
ahoppen: Oh, good catch. I forgot to add that.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Testing found a request ordering race condition that I’m fixing in https://github.com/apple/sourcekit-lsp/pull/920/commits/96a9a05b08f8917504eb495490a6b79578bd4ccb. If you could review that commit, I’d appreciate it.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitDTests/CrashRecoveryTests.swift
```diff
@@ -48,39 +48,22 @@ final class CrashRecoveryTests: XCTestCase {
...
-          print("Hello world")
-        }
-        """
-    )
-    ws.testClient.send(
-      DidChangeTextDocumentNotification(
-        textDocument: VersionedTextDocumentIdentifier(loc.docUri, version: 2),
-        contentChanges: [addFuncChange]
-      )
-    )
```
hamishknight, line 78: Do we not need to test for this anymore?

Sources/SKTestSupport/Utils.swift
```diff
@@ -51,8 +51,8 @@ public func testScratchDir(testName: String = #function) throws -> URL {
...
     .appendingPathComponent(String(testBaseName))
```
hamishknight, line 54: Are there any concerns about name clashes when running tests in parallel here?

hamishknight (approved):

Tests/SourceKitDTests/CrashRecoveryTests.swift
ahoppen: Since the entire file is opened in-memory and there’s no on-disk file (because we just use a `TestSourceKitLSPClient` and not e.g. an `IndexedSingleFileSwiftWorkspace`), there’s no on-disk version of the file that sourcekit-lsp could fall back to. So no, need for this anymore.

Sources/SKTestSupport/Utils.swift
ahoppen: You have a point. I was hoping that the test method should be unique but it turns out that we do have two `testMainFilesChanged` methods that will probably both create a scratch directory.

bnbarham: Could just add the class name to the name here? Though I suppose even then *technically* we could still clash 😅

ahoppen: And how would you do that?

bnbarham: Would have to pass it in, ie. `String(describing: Self.self)` or similar.

ahoppen: Yes but we don’t have `self` here because this is a top-level function. And I don’t want to add another `self` parameter to `IndexedSingleFileWorkspace` etc. just to get the test class name.

bnbarham: Sure, like I said, would have to pass it in. I'm fine if we just want to do UUID instead, though in that case I'd make the test name part of the name with the UUID, rather than a subfolder.

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

Sources/SKTestSupport/Utils.swift
ahoppen: I filed https://github.com/apple/sourcekit-lsp/issues/929 to improve test scratch directories in general.

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #921, opened 2023-10-25
Rename `TestClient` to `TestMessageHandler`

ahoppen:
`TestClient` was confusing because we have a bunch of variables `testClient: TestSourceKitLSPClient`. `TestMessageHandler` should describe the type’s job a lot better.

There’s probably room for more cleanup in the entire JSONRPC connection testing area but I’m not touching that for now.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #924, opened 2023-10-25
Capture stderr from clangd and forward it to the logger

ahoppen:
Previously, we were forwarding clangd’s stderr to sourcekit-lsp’s stderr. We should pipe it through the logger so that it ends up in OSLog on Darwin.

### DIFF
--- a/Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
+++ b/Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
@@ -32,6 +32,26 @@ extension NSLock {
   }
 }
 
+/// Gathers data from clangd's stderr pipe. When it has accumulated a full line, writes the the line to the logger.
+fileprivate class ClangdStderrLogForwarder {
+  private var buffer = Data()
+
+  func handle(_ newData: Data) {
+    self.buffer += newData
+    while let newlineIndex = self.buffer.firstIndex(of: UInt8(ascii: "\n")) {
+      // Output a separate log message for every line in clangd's stderr.
+      // The reason why we don't output multiple lines in a single log message is that 
+      //  a) os_log truncates log messages at about 1000 bytes. The assumption is that a single line is usually less
+      //     than 1000 bytes long but if we merge multiple lines into one message, we might easily exceed this limit.
+      //  b) It might be confusing why sometimes a single log message contains one line while sometimes it contains 
+      //     multiple.
+      let logger = Logger(subsystem: subsystem, category: "clangd-stderr")
+      logger.info("\(String(data: self.buffer[...newlineIndex], encoding: .utf8) ?? "<invalid UTF-8>")")
+      buffer = buffer[buffer.index(after: newlineIndex)...]
+    }
+  }
+}
+
 /// A thin wrapper over a connection to a clangd server providing build setting handling.
 ///
 /// In addition, it also intercepts notifications and replies from clangd in order to do things
@@ -203,6 +223,17 @@ actor ClangLanguageServerShim: ToolchainLanguageServer, MessageHandler {
 
     process.standardOutput = clangdToUs
     process.standardInput = usToClangd
+    let logForwarder = ClangdStderrLogForwarder()
+    let stderrHandler = Pipe()
+    stderrHandler.fileHandleForReading.readabilityHandler = { fileHandle in
+      let newData = fileHandle.availableData
+      if newData.count == 0 {
+        stderrHandler.fileHandleForReading.readabilityHandler = nil
+      } else {
+        logForwarder.handle(newData)
+      }
+    }
+    process.standardError = stderrHandler
     process.terminationHandler = { [weak self] process in
       logger.log(
         level: process.terminationReason == .exit ? .default : .error,


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -203,6 +224,17 @@ actor ClangLanguageServerShim: ToolchainLanguageServer, MessageHandler {
 
     process.standardOutput = clangdToUs
     process.standardInput = usToClangd
+    let logForwarder = ClangdStderrLogForwarder()
+    let stderrHandler = Pipe()
+    stderrHandler.fileHandleForReading.readabilityHandler = { fileHandle in
+      let newData = fileHandle.readData(ofLength: 1)
```
bnbarham, line 230: Can't we just use `availableData` here?

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -203,6 +224,17 @@ actor ClangLanguageServerShim: ToolchainLanguageServer, MessageHandler {
...
+      if newData.count == 0 {
+        stderrHandler.fileHandleForReading.readabilityHandler = nil
```
rintaro, line 232: Doesn't this cause a cyclic reference?

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
```diff
@@ -32,6 +32,27 @@ extension NSLock {
...
+    self.buffer += newData
+    if newData.contains(UInt8(ascii: "\n")) {
+      self.flushBuffer()
+    }
+  }
+
+  private func flushBuffer() {
+    guard let newlineIndex = self.buffer.firstIndex(of: UInt8(ascii: "\n")) else {
+      return
+    }
```
hamishknight, line 49: Would it be better to use `lastIndex` here? I guess it doesn't happen with the current implementation since we're reading a byte at a time, but if we weren't and e.g if multiple lines got read at once, presumably we'd want to forward all of them here. Plus the newline is more likely to be nearer the end than the start in general.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: According to you on https://github.com/apple/swift-stress-tester/commit/4250ceb3d67d80d0f57943aa10122e026935c383

> Also using `readData` rather than `availableData` as various online comments have mentioned `availableData` causing various issues.

Do you remember what these *various issues* are? Or should we give `availableData` a try again?

And I can’t use `fileHandle.readData(ofLength: Int.max)` because that will block until it reaches `Int.max` bytes or the maximum size of `NSData`, which is usually more than a single line. Thus, `readData(ofLength: 1)` is the best I came up with.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: Once the end of the `stderr` pipe is reached (indicated by `newData.count == 0`), we remove the `readabilityHandler` from the `stderrHandler` again, which should break the cycle.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: The quest is whether we want a separate os_log message for every line from clangd or if we are fine with merging multiple clangd stderr lines into a single os_log message. 

I don’t have super strong opinions here but one thing to keep in mind is that os_log has a limit to how long the log message can be (something around 1KB). So, if we merge too many lines together os_log might truncate. That’s why I chose to only have one line per os_log message, with the assumption that usually a single line will be less than 1024 characters long.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
bnbarham: > Do you remember what these various issues are?

Hah, I knew I remembered something about this when I posted this comment. I believe it was something to do with `availableData` being empty even though there was more to read, so the `readabilityHandler` would be discarded before it actually should be. I really should have just added the various URLs there 😅 

Anyway, I vote we try `availableData` and come back to it if we see any problems.

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
hamishknight: Makes sense, my only real concern here is accidentally ending up in a situation where we don't fully flush the buffer. If we do switch to `availableData`, should we be making an effort to split lines here?

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: Changed it to `availableData`

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Clang/ClangLanguageServer.swift
ahoppen: I changed it to a `while` loop so that we flush out all the lines to separate log messages and added a comment explaining why every line is its own message.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #925, opened 2023-10-25
Implement the document color request using the SwiftSyntax tree

ahoppen:


bnbarham (approved):
Could also look for strings matching `#[a-zA-Z0-9]{6}` :P

What does this actually show up as in eg. VSCode right now? Does picking a color insert the actual hex value instead of what we expect (ie. `#colorLiteral(...)`)?

ahoppen:
@swift-ci Please test

PR #926, opened 2023-10-26
Implement DocumentSymbolsRequest using SwiftSyntax

ahoppen:
Before doing that, I refactored the document symbol tests to be more modular so that if one fails, you can actually tell what the failure is. It’s worth noting the slight differences in results introduced by the second commit that starts using swift-syntax. IMO all changes are improvements


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/DocumentSymbolTests.swift
```diff
@@ -299,23 +299,23 @@ final class DocumentSymbolTests: XCTestCase {
   }
 
   func testMultiplePatternsInVariable() async throws {
-    try await assertDocumentSymbols("1️⃣var 2️⃣varA3️⃣: Int, 4️⃣varB5️⃣ = 06️⃣") { positions in
+    try await assertDocumentSymbols("var 1️⃣varA2️⃣: Int,3️⃣ 4️⃣varB5️⃣ = 06️⃣") { positions in
       [
         DocumentSymbol(
           name: "varA",
-          detail: "Int",
+          detail: nil,
```
hamishknight, line 306: Is `detail` always `nil` now?

hamishknight (approved):

Tests/SourceKitLSPTests/DocumentSymbolTests.swift
ahoppen: Yes. We could try to get the type from a variable declaration if it is explicitly specified but I think it’s more confusing if you have it for `let x: Int = 1` but not for `let x = 1`, so I chose to just always set it to `nil`.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/DocumentSymbols.swift
```diff
@@ -0,0 +1,241 @@
...
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SwiftSyntax
+import LSPLogging
+
+extension SwiftLanguageServer {
+  public func documentSymbol(_ req: DocumentSymbolRequest) async throws -> DocumentSymbolResponse? {
+    guard let snapshot = self.documentManager.latestSnapshot(req.textDocument.uri) else {
+      logger.error("failed to find snapshot for url \(req.textDocument.uri.forLogging)")
```
bnbarham, line 20: We have this in so many places. Is requesting a snapshot and not finding it ever *not* an error? ie. should we just always log?

Sources/SourceKitLSP/Swift/DocumentSymbols.swift
```diff
@@ -0,0 +1,241 @@
...
+    let rangeEnd =
+      if let parameterClause = node.parameterClause {
+        parameterClause.endPositionBeforeTrailingTrivia
+      } else {
+        node.name.endPositionBeforeTrailingTrivia
+      }
+
+    return record(
+      node: node,
+      name: node.qualifiedDeclName,
```
bnbarham, line 124: `qualified` to me suggests eg. `Type.member` rather than just the full name (including parentheses).

Tests/SourceKitLSPTests/DocumentSymbolTests.swift
```diff
@@ -17,598 +17,632 @@ import SourceKitLSP
...
-    let text = """
-      struct Foo { }
-      """
-    let symbols = try await performDocumentSymbolRequest(text: text)
-
-    XCTAssertEqual(
-      symbols,
-      .documentSymbols([
+  func testUnicode1() async throws {
+    try await assertDocumentSymbols("1️⃣struct 2️⃣Żółć3️⃣ { }4️⃣") { positions in
```
bnbarham, line 27: Nice 👍

Sources/SourceKitLSP/Swift/DocumentSymbols.swift
ahoppen: Good idea. Here we go: https://github.com/apple/sourcekit-lsp/pull/943

Sources/SourceKitLSP/Swift/DocumentSymbols.swift
ahoppen: Ah, yeah, qualified really isn’t the term here. Renaming to `declName` in https://github.com/apple/sourcekit-lsp/pull/944

PR #927, opened 2023-10-26
Minor improvements to SwiftLanguageServer.swift

ahoppen:
- Introduce a new sub-logging scope for the implicitly triggered publish diagnostics notification
  - Just makes it clearer why some logs are being triggered if you know that it’s for the publish diagnostics notification.
- Don’t log a fault if the publish diagnostics notification is cancelled
  - Cancelling the publish diagnostics notificaiton because there was an edit is totally normal and nothing to be logged at the `fault` level.
- Format SwiftLanguageServer.swift


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #932, opened 2023-10-26
Only log build settings once unless they change

ahoppen:
This makes the logs a lot less spammy.

### DIFF
--- a/Sources/SKCore/BuildSystemManager.swift
+++ b/Sources/SKCore/BuildSystemManager.swift
@@ -72,25 +72,6 @@ public actor BuildSystemManager {
   }
 }
 
-/// Splits `message` on newline characters such that each chunk is at most `maxChunkSize` bytes long.
-///
-/// The intended use case for this is to split compiler arguments into multiple chunks so that each chunk doesn't exceed
-/// the maximum message length of `os_log` and thus won't get truncated.
-///
-///  - Note: This will only split along newline boundary. If a single line is longer than `maxChunkSize`, it won't be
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SKCore/BuildSystemManager.swift
```diff
@@ -334,3 +299,55 @@ extension BuildSystemManager {
...
   }
 }
+
+// MARK: - Build settings logger
+
+/// Shared logger that only logs build settings for a file once unless they change
+fileprivate actor BuildSettingsLogger {
+  static let shared = BuildSettingsLogger()
+
+  private var loggedSettings: [DocumentURI: FileBuildSettings] = [:]
```
bnbarham, line 309: I know we likely already keep a document -> settings in memory for all the workspaces, but it would be nice to not (potentially) double it. Are you worried about some race if this was removed on close?

ahoppen: AFAIK we don’t keep that mapping around anywhere. We used to but I removed the caches in `SwiftLanguageServer` and `ClangLanguageServer`. 

`SwiftPMWorkspace`, for example, gets the build settings from the build graph and doesn’t have any file -> build settings cache.

hamishknight (approved):

Sources/SKCore/BuildSystemManager.swift
ahoppen: I’m merging this now. If you think this should be implemented otherwise, I can address it in a follow-up PR.

PR #933, opened 2023-10-26
When sourcekitd crashes, log the file contents with which it crashed and the request

ahoppen:
This should make it a lot easier to reproduce sourcekitd crashes.

### DIFF
--- a/Sources/SourceKitD/SourceKitD.swift
+++ b/Sources/SourceKitD/SourceKitD.swift
@@ -65,7 +65,11 @@ extension SourceKitD {
 
   // MARK: - Convenience API for requests.
 
-  public func send(_ req: SKDRequestDictionary) async throws -> SKDResponseDictionary {
+  /// - Parameters:
+  ///   - req: The request to send to sourcekitd.
+  ///   - fileContents: The contents of the file that the request operates on. If sourcekitd crashes, the file contents
+  ///     will be logged.
+  public func send(_ req: SKDRequestDictionary, fileContents: String?) async throws -> SKDResponseDictionary {
     logRequest(req)
 
     let signposter = logger.makeSignposter()
@@ -86,6 +90,24 @@ extension SourceKitD {
 
     guard let dict = sourcekitdResponse.value else {
       signposter.endInterval("sourcekitd-request", signposterState, "Error")
+      if sourcekitdResponse.error == .connectionInterrupted {
+        let log = """
+          Request:
+          \(req.description)
+
+          File contents:
+          \(fileContents ?? "<nil>")
+          """
+        let chunks = splitLongMultilineMessage(message: log)
+        for (index, chunk) in chunks.enumerated() {
+          logger.fault(
+            """
+            sourcekitd crashed (\(index + 1)/\(chunks.count))
+            \(chunk)
+            """
+          )
+        }
+      }
       throw sourcekitdResponse.error!
     }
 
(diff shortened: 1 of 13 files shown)


Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -620,12 +620,7 @@ extension SourceKitServer: MessageHandler {
   private func handleImpl(_ params: some NotificationType, from clientID: ObjectIdentifier) async {
     let notification = Notification(params, clientID: clientID)
 
-    logger.log(
-      """
-      Received notification
-      \(notification.forLogging)
-      """
-    )
+    logger.log("Received notification: \(notification.forLogging)")
```
bnbarham, line 623: Presumably you found this looks nicer without the newline?

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -107,9 +129,7 @@ extension SourceKitD {
 }
 
 private func logRequest(_ request: SKDRequestDictionary) {
-  // FIXME: Ideally we could log the request key here at the info level but the dictionary is
-  // readonly.
-  logger.log(
+  logger.info(
```
bnbarham, line 132: What's default vs info? The docs weren't particularly illuminating.

bnbarham (commented):
Should we log similar for clangd?

Sources/SourceKitD/SourceKitD.swift
ahoppen: https://github.com/apple/sourcekit-lsp/blob/78c28f5dec294ffa87ffcbd91fb9bbeae9d8cbd4/Sources/LSPLogging/Logging.swift#L13-L18

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Yes, my use case for this was that I could grep for `Received notification` and see the types of all the notifications that got sent.

ahoppen:
@swift-ci Please test

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -96,6 +100,24 @@ extension SourceKitD {
...
+          """
+        let chunks = splitLongMultilineMessage(message: log)
+        for (index, chunk) in chunks.enumerated() {
+          logger.fault(
+            """
+            sourcekitd crashed (\(index + 1)/\(chunks.count))
+            \(chunk)
+            """
+          )
+        }
```
hamishknight, line 119: If OSLog treats this as private (which I believe is the default?), will we just log multiple `sourcekitd crashed` lines with redacted chunks?

Sources/SourceKitD/SourceKitD.swift
bnbarham: Ah, so persisted vs not 👍

Sources/SourceKitD/SourceKitD.swift
ahoppen: Yes, we will. It’s a big unfortunate but I don’t have any better idea.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #869, opened 2023-10-09
Remove non-async version of `SourceKitD.sendSync`


ahoppen:
The synchronous version of `SoruceKitD.sendSync` doesn’t support cancellation and is an anti-pattern. All of the calls should be migrated to use the async version and the sync version should be removed.

ahoppen:
Tracked in Apple’s issue tracker as rdar://116703689

PR #934, opened 2023-10-26
Remove `SourceKitD.sendSync`

ahoppen:
Everybody should be calling the `async` version of `send`. Most callers could be migrated easily. `SwiftLanguageServer.changeDocument` was a little bit more tricky because the callback of `DocumentManager.edit` can’t be async but that’s easily fixed by gathering the edits in an array and sending them to `sourcekitd` outside after the document manager is done.

Resolves https://github.com/apple/sourcekit-lsp/issues/869
rdar://116703689

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -422,46 +422,66 @@ extension SwiftLanguageServer {
     cancelInFlightPublishDiagnosticsTask(for: note.textDocument.uri)
 
     let keys = self.keys
-    var edits: [IncrementalEdit] = []
+    struct Edit {
+      let offset: Int
+      let length: Int
+      let replacement: String
+    }
+
+    var edits: [Edit] = []
 
     let editResult = self.documentManager.edit(note) {
       (before: LineTable, edit: TextDocumentContentChangeEvent) in
-      let req = SKDRequestDictionary(sourcekitd: self.sourcekitd)
-      req[keys.request] = self.requests.editor_replacetext
-      req[keys.name] = note.textDocument.uri.pseudoPath
-      req[keys.syntactic_only] = 1
-
       if let range = edit.range {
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -422,46 +422,66 @@ extension SwiftLanguageServer {
...
-    var edits: [IncrementalEdit] = []
+    struct Edit {
+      let offset: Int
+      let length: Int
+      let replacement: String
+    }
+
+    var edits: [Edit] = []
 
     let editResult = self.documentManager.edit(note) {
```
bnbarham, line 433: I haven't looked at this too much, but is the callback from `edit` even really needed any more? Or should we just do the TextDocumentContentChangeEvent -> [Edit] outside it?

bnbarham (approved):

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: It is, because the callback is transforming the concurrent edits from the LSP request into sequential edits that we can send to sourcekitd.

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
bnbarham: Okay, it could also just return those rather than run the callback. It would mean another top level `Edit` though, which we don't have enough of :P (Edit/SourceEdit/IncrementalEdit all in swift-syntax).

ahoppen: Filed https://github.com/apple/sourcekit-lsp/issues/939

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

PR #935, opened 2023-10-26
Remove the fallback value from `SourceKitServer.handleRequest`

### LINKED ISSUE #862: Remove the fallback value from `SourceKitServer.handleRequest`
It’s not entire clear why we have a fallback value if no document service can serve a request and I checked that VS Code behaves if we return empty results for jump to definition or if we return an error. To simplify the codebase, we should just throw an error here and remove the `fallback` parameter from `handleRequest`.

ahoppen:
It’s not entire clear why we have a fallback value if no document service can serve a request and I checked that VS Code behaves if we return empty results for jump to definition or if we return an error. To simplify the codebase, we should just throw an error here and remove the `fallback` parameter from `handleRequest`.

Resolves #862
rdar://116703662

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -395,8 +395,7 @@ public actor SourceKitServer {
 
   private func handleRequest<RequestType: TextDocumentRequest>(
     for request: Request<RequestType>,
-    requestHandler: @escaping (RequestType, Workspace, ToolchainLanguageServer) async throws -> RequestType.Response,
-    fallback: RequestType.Response
+    requestHandler: @escaping (RequestType, Workspace, ToolchainLanguageServer) async throws -> RequestType.Response
   ) async {
     await handleRequest(request) { request in
       let doc = request.textDocument.uri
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -395,16 +395,15 @@ public actor SourceKitServer {
...
+    requestHandler: @escaping (RequestType, Workspace, ToolchainLanguageServer) async throws -> RequestType.Response
   ) async {
     await handleRequest(request) { request in
       let doc = request.textDocument.uri
       guard let workspace = await self.workspaceForDocument(uri: request.textDocument.uri) else {
         throw ResponseError.workspaceNotOpen(request.textDocument.uri)
       }
       guard let languageService = workspace.documentService[doc] else {
-        return fallback
+        throw ResponseError.unknown("No language service for '\(request.textDocument.uri)' founds")
```
bnbarham, line 406:
```suggestion
        throw ResponseError.unknown("No language service for '\(request.textDocument.uri)' found")
```

bnbarham (approved):
How... interesting. I wonder if other clients don't handle errors as well?

ahoppen:
> How... interesting. I wonder if other clients don't handle errors as nicely?

@benlangmuir couldn’t remember why we didn’t throw an error here. And IMO in case we don’t have a language it’s better to be explicit about it and return an error than silently dropping the issue.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #931, opened 2023-10-26
Remove the `Notification` type

ahoppen:
`Notification` is just a wrapper around `NotificationType` and there’s no reason for it to exist anymore.

Resolves #880
rdar://116705670

PR #937, opened 2023-10-27
Change the `Request` type to be a fileprivate `RequestAndReply` in `SourceKitServer` 🚥 #931

### LINKED ISSUE #881: Remove `Request` type
AFAICT there are two more uses of the `Request` type: Inside `SourceKitServer` and inside `appendOneShotRequestHandler`. 

We should change `appendOneShotRequestHandler` to be an async function that returns the request result asynchronously.

I’m still unsure if we can get rid of `Request` inside `SourceKitServer.handleRequest` without sacrificing type safety, but we should at least be able to make `Request` a local type inside `SourceKitServer.handleRequest`.

### LINKED ISSUE #936: Remove dependency of `LanguageServerProtocol` on `LSPLogging`
The dependency of `LanguageServerProtocol` isn’t nice. `LanguageServerProtocol` should be usable in any context without any dependency on SourceKit-LSP specific logic. Move the conformances of types in `LanguageServerProtocol` to `CustomLogStringConvertible` into some other module and remove the dependency.

ahoppen:
Depends on https://github.com/apple/sourcekit-lsp/pull/931

---

Since asyncifying all the request handling, `Request` only had a few lingering uses. Remove most of them and shrink it down to a type that only contains the request’s parameters and the reply block of the corresponding type.

And while we’re doing this, also move `NotificationType.forLogging` out of the `LanguageServerProtocol` module to remove the dependency from `LanguageServerProtocol` on `LSPLogging`.

Resolves #881
Resolves #936
rdar://116705662
rdar://117562587


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/LoggableMessageTypes.swift
```diff
@@ -12,66 +12,11 @@
...
-  public func reply(_ result: Response) {
-    reply(.success(result))
-  }
-}
+import LanguageServerProtocol
 
 fileprivate extension Encodable {
   var prettyPrintJSON: String {
     let encoder = JSONEncoder()
-    encoder.outputFormatting.insert(.sortedKeys)
```
hamishknight, line 74: Did you mean to remove this?

hamishknight (approved):

Sources/SourceKitLSP/LoggableMessageTypes.swift
ahoppen: I did not, must have been caused by rebasing. Thanks for catching.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #941, opened 2023-10-27
Report codeActions for fixIts produced by swift syntax

joehsieh:
~still WIP, i'm trying to figure out how to write tests for this change.~

Implemented an initializer of CodeAction with the FixIt from swift-syntax and added a test for it.

### DIFF
--- a/Sources/SourceKitLSP/Swift/Diagnostic.swift
+++ b/Sources/SourceKitLSP/Swift/Diagnostic.swift
@@ -44,22 +44,10 @@ extension CodeAction {
     if let fromNote = fromNote {
       title = fromNote
     } else {
-      guard let startIndex = snapshot.index(of: edits[0].range.lowerBound),
-        let endIndex = snapshot.index(of: edits[0].range.upperBound),
-        startIndex <= endIndex,
-        snapshot.text.indices.contains(startIndex),
-        endIndex <= snapshot.text.endIndex
-      else {
-        logger.fault("position mapped, but indices failed for edit range \(String(reflecting: edits[0]))")
+      guard let generatedTitle = Self.title(for: edits, in: snapshot) else {
         return nil
       }
-      let oldText = String(snapshot.text[startIndex..<endIndex])
-      let description = Self.fixitTitle(replace: oldText, with: edits[0].newText)
-      if edits.count == 1 {
-        title = description
-      } else {
-        title = description + "..."
-      }
+      title = generatedTitle
     }
 
     self.init(
@@ -71,10 +59,43 @@ extension CodeAction {
   }
 
   init?(_ fixIt: FixIt, in snapshot: DocumentSnapshot) {
-    // FIXME: Once https://github.com/apple/swift-syntax/pull/2226 is merged and
-    // FixItApplier is public, use it to compute the edits that should be
-    // applied to the source.
-    return nil
+    var textEdits = [TextEdit]()
+    for edit in fixIt.edits {
+      guard let startPosition = snapshot.position(of: edit.range.lowerBound),
+        let endPosition = snapshot.position(of: edit.range.upperBound)
+      else {
+        continue
+      }
+      textEdits.append(TextEdit(range: startPosition..<endPosition, newText: edit.replacement))
+    }
+
+    self.init(
+      title: fixIt.message.message.withFirstLetterUppercased(),
+      kind: .quickFix,
+      diagnostics: nil,
+      edit: WorkspaceEdit(changes: [snapshot.uri: textEdits])
+    )
+  }
+
+  private static func title(for edits: [TextEdit], in snapshot: DocumentSnapshot) -> String? {
+    if edits.isEmpty {
+      return nil
+    }
+    guard let startIndex = snapshot.index(of: edits[0].range.lowerBound),
+      let endIndex = snapshot.index(of: edits[0].range.upperBound),
+      startIndex <= endIndex,
+      snapshot.text.indices.contains(startIndex),
+      endIndex <= snapshot.text.endIndex
+    else {
+      logger.fault("position mapped, but indices failed for edit range \(String(reflecting: edits[0]))")
+      return nil
+    }
+    let oldText = String(snapshot.text[startIndex..<endIndex])
+    let description = Self.fixitTitle(replace: oldText, with: edits[0].newText)
+    if edits.count == 1 {
+      return description
+    }
+    return description + "..."
   }
 
   /// Describe a fixit's edit briefly.
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -425,4 +425,51 @@ final class CodeActionTests: XCTestCase {
 
     try await fulfillmentOfOrThrow([editReceived])
   }
+
+  func testCodeActionForFixItsProducedBySwiftSyntax() async throws {
+    let ws = try await MultiFileTestWorkspace(files: [
+      "test.swift": "protocol 1️⃣Multi 2️⃣ident 3️⃣{}",
+      "compile_commands.json": "[]",
+    ])
+
+    let (uri, positions) = try ws.openDocument("test.swift")
+
+    let report = try await ws.testClient.send(DocumentDiagnosticsRequest(textDocument: TextDocumentIdentifier(uri)))
+    guard case .full(let fullReport) = report else {
+      XCTFail("Expected full diagnostics report")
+      return
+    }
+
+    XCTAssertEqual(fullReport.items.count, 1)
+    let diagnostic = try XCTUnwrap(fullReport.items.first)
+    let codeActions = try XCTUnwrap(diagnostic.codeActions)
+
+    let expectedCodeActions = [
+      CodeAction(
+        title: "Join the identifiers together",
+        kind: .quickFix,
+        edit: WorkspaceEdit(
+          changes: [
+            uri: [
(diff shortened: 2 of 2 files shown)


ahoppen:
Having an idea of how to test this is definitely not trivial, so here’s a snippet. You should then be able to write assertions to test `diagnostics`.

```swift
// Create a workspace with compile commands that don't contain the file we are getting diagnostics for.
// That way we have a workspace that has a build system (and thus doesn't promote settings from the fallback build
// system to be non-fallback settings) but doesn't have build settings for the file being tested. Thus, we generate
// diagnostics from the built-in SwiftParser.
let ws = try await MultiFileTestWorkspace(files: [
  "test.swift": "let var = 2",
  "compile_commands.json": "[]"
])

let (uri, _) = try ws.openDocument("test.swift")

let diagnostics = try await ws.testClient.send(DocumentDiagnosticsRequest(textDocument: TextDocumentIdentifier(uri)))
```

ahoppen:
And thanks a lot for working on this @joehsieh!

Package.swift
```diff
@@ -337,7 +337,8 @@ if useLocalDependencies {
     .package(url: "https://github.com/apple/swift-package-manager.git", branch: relatedDependenciesBranch),
     .package(url: "https://github.com/apple/swift-tools-support-core.git", branch: relatedDependenciesBranch),
     .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2"),
-    .package(url: "https://github.com/apple/swift-syntax.git", branch: relatedDependenciesBranch),
+    // .package(url: "https://github.com/apple/swift-syntax.git", branch: relatedDependenciesBranch),
+    .package(url: "https://github.com/joehsieh/swift-syntax", branch: "add-fix-it-extesion"),
```
joehsieh, line 341: this change will be reverted once this [PR](https://github.com/apple/swift-syntax/pull/2314) is merged.

Sources/SourceKitLSP/Swift/Diagnostic.swift
```diff
@@ -71,10 +59,46 @@ extension CodeAction {
...
+
+    self.init(
+      title: generatedTitle,
+      kind: .quickFix,
+      diagnostics: nil,
+      edit: WorkspaceEdit(changes: [snapshot.uri: textEdits])
+    )
+  }
+
+  private static func generateTitle(with edits: [TextEdit], snapshot: DocumentSnapshot) -> String? {
```
ahoppen, line 83: Nitpick, I would name this without `generate`. Since this is side-effect free, I don’t think the function name needs a result and `generate` basically just says that the function is returning something, which is obvious from the result type.
```suggestion
  private static func title(for edits: [TextEdit], in snapshot: DocumentSnapshot) -> String? {
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -425,4 +425,54 @@ final class CodeActionTests: XCTestCase {
...
+    }
+
+    let expectedCodeActions = [
+      CodeAction(
+        title: "Replace \'Multi \' with \'Multiident \'...", 
+        kind: .quickFix, 
+        edit: WorkspaceEdit(
+          changes: [
+            uri : [
+              TextEdit(range: Position(line: 0, utf16index: 9)..<Position(line: 0, utf16index: 15), newText: "Multiident "),
```
ahoppen, line 457: Could you use location markers here instead of hardcoded line-column offsets. We aren’t very good about it in the codebase yet as I only introduced the concept of location markers in the last couple of weeks but it’s generally proffered because of the same reasons that I mentioned in your swift-syntax PR.

Here’s an example of how to use the location markers.

https://github.com/apple/sourcekit-lsp/blob/9f7ff0891a115f1ae056348744abffd0452806e5/Tests/SourceKitLSPTests/LocalClangTests.swift#L246-L264

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -425,4 +425,54 @@ final class CodeActionTests: XCTestCase {
...
+    XCTAssertEqual(fullReport.items.count, 1)
+    let diagnostic = try XCTUnwrap(fullReport.items.first)
+    guard let codeActions = diagnostic.codeActions else {
+      XCTFail("Expected non-empty codeActions")
+      return
+    }
+
+    let expectedCodeActions = [
+      CodeAction(
+        title: "Replace \'Multi \' with \'Multiident \'...",
```
ahoppen, line 452: You don’t need to escape the `'` in Swift string literals.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -425,4 +425,54 @@ final class CodeActionTests: XCTestCase {
...
+      XCTFail("Expected full diagnostics report")
+      return
+    }
+
+    XCTAssertEqual(fullReport.items.count, 1)
+    let diagnostic = try XCTUnwrap(fullReport.items.first)
+    guard let codeActions = diagnostic.codeActions else {
+      XCTFail("Expected non-empty codeActions")
+      return
+    }
```
ahoppen, line 448: You could just use `XCTUnwrap`

```suggestion
    let codeActions = try XCTUnwrap(diagnostic.codeAction)
```

Sources/SourceKitLSP/Swift/Diagnostic.swift
joehsieh: you're right. let me update it.

Tests/SourceKitLSPTests/CodeActionTests.swift
joehsieh: got it, it makes sense to use location markers instead. i'll update the code.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -425,4 +425,51 @@ final class CodeActionTests: XCTestCase {
 
     try await fulfillmentOfOrThrow([editReceived])
   }
+
+  func testCodeActionForFixItsProducedBySwiftSyntax() async throws {
+    let ws = try await MultiFileTestWorkspace(files: [
+      "test.swift": "protocol 0️⃣Multi 1️⃣ident 2️⃣{}",
```
ahoppen, line 431: Nitpick similar to the PR in swift-syntax: We usually start counting the location markers at 0. Would be nice to be consistent here.

Sources/SourceKitLSP/Swift/Diagnostic.swift
```diff
@@ -71,10 +59,46 @@ extension CodeAction {
...
+      guard let startPosition = snapshot.position(of:edit.range.lowerBound),
+        let endPosition = snapshot.position(of:edit.range.upperBound) else {
+        continue
+      }
+      textEdits.append(TextEdit(range: startPosition..<endPosition, newText: edit.replacement))
+    }
+
+    guard let generatedTitle = Self.title(for: textEdits, in: snapshot) else {
+      return nil
+    }
```
ahoppen, line 73: Fix-Its from swift-syntax always have a message (`fixIt.message.message` I think) and I think we should be using that instead of synthesizing a title here. I think the extraction of `title(for:in:)` makes sense regardless though.

joehsieh: i tried to use `fixIt.message.message` instead of `title(for:in:)`, but it failed the test, so we might not be able to do it. 
```
Expected: 
"Replace 'Multi ' with 'Multiident '..."
"Replace 'Multi ' with 'MultiIdent '..."

Actual:
"join the identifiers together",
"join the identifiers together with camel-case"
```

ahoppen: Can’t you just update the test case to expect `join the identifiers together`?

Also, it would be good to uppercase the first character of the diagnostic, similar to what I did here: https://github.com/apple/sourcekit-lsp/pull/784

joehsieh: you're right, i put the code `protocol Multi ident {}` in a swift playground.
the error message was 
```
Join the identifiers together
Join the identifiers together with camel-case
```
let me correct it.

ahoppen (commented):
Looks good to me. We just need to wait to get https://github.com/apple/swift-syntax/pull/2314 in and revert the change to the package manifest.

When you do that, could you squash your commits? It makes for a nicer Git history.

joehsieh:
Got it, i'll do it.
btw, is there a way/workaround to format the code for sourcekit-lsp? i'm just curious about it.
i know we haven't supported it yet. https://github.com/apple/sourcekit-lsp#status.

ahoppen:
I was thinking about integrating the swift-format command plugin into sourcekit-lsp. That way you should be able to run swift-format e.g. from Xcode. But I haven’t had time to look into it yet. If you want, you could try that.

joehsieh:
thanks for the reminder. i found a vscode plugin for swift-format.
https://marketplace.visualstudio.com/items?itemName=vknabel.vscode-apple-swift-format&ssr=false#overview
even though i need to click a button to format the code by myself...

for the plugin of Xcode, it looks like there is a good one. https://github.com/nicklockwood/SwiftFormat#what-is-this
but I haven't tried it.

ahoppen (approved):
Let’s get this in 🎉

ahoppen:
@swift-ci Please test

ahoppen:
Windows failure is unrelated to your PR.

shahmishal:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #942, opened 2023-10-27
Make logs less noisy

ahoppen:
Make logs less noisy in two ways:

- Set the build system to `nil` if no compilation database can be loaded
  - When we couldn’t start a build server or find a SwiftPM package, we currently always create a `CompilationDatabaseBuildSystem`, even if no `compile_commands.json` or `compile_flags.txt` exits. Every request for build settings would then log an error that the compilation database can’t be opened, which was very spammy. Instead, if the compilation database can’t be loaded, just set the build system to `nil` and log a single error message.
- Reduce SwiftPM log message to be at most at the `default` level
  - SwiftPM would output warnings for every compilation warning it hits while building the package manifest, which usually doesn’t affect sourcekit-lsp. Logging these at the `error` level is very noisy. Reduce the log level to `default`.

### DIFF
--- a/Sources/SKCore/CompilationDatabaseBuildSystem.swift
+++ b/Sources/SKCore/CompilationDatabaseBuildSystem.swift
@@ -73,12 +73,16 @@ public actor CompilationDatabaseBuildSystem {
     return nil
   }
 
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #947, opened 2023-10-27
Skip semantic tokens tests if sourcekitd doesn't support the semantic tokens request

### LINKED ISSUE #940: Fix failed SemanticTokensTests
Got below message after running `swift test` on main.
I was wondering if it's something wrong with my env, so i git-bisect the commits, and it seems this tests were failed after we merged this [PR](https://github.com/apple/sourcekit-lsp/pull/912) on Oct, 25.
But why it pass on CI, but failed on my side.

```
Test Case '-[SourceKitLSPTests.SemanticTokensTests testSemanticTokens]' failed (0.109 seconds).
Test Case '-[SourceKitLSPTests.SemanticTokensTests testSemanticTokensForStaticMethods]' failed (0.113 seconds).
...
```

ahoppen:
When running sourcekit-lsp’s tests using Xcode 15, they fail because the sourcekitd in Xcode 15 does not contain the semantic tokens request. The intended workaround/fix is to dowload a recent Swift development snapshot from swift.org and set it as `SOURCEKIT_TOOLCHAIN_PATH` when running tests. But that’s not a great experience for new contributors. Instead, if sourcekitd doesn’t support the semantic tokens request, simply skip the test.

This also changes the implementation of the semantic tokens LSP request slightly: When the sourcekitd request to get semantic tokens fails, we now fail the entire LSP request, instead of returning the tokens from the synax tree. I think that’s reasonable because the editor did ask for semantic tokens, not syntactic tokens.

Fixes #940
rdar://117590581

### DIFF
--- a/Sources/SourceKitLSP/Swift/SemanticTokens.swift
+++ b/Sources/SourceKitLSP/Swift/SemanticTokens.swift
@@ -50,9 +50,9 @@ extension SwiftLanguageServer {
   private func mergedAndSortedTokens(
     for snapshot: DocumentSnapshot,
     in range: Range<Position>? = nil
-  ) async -> [SyntaxHighlightingToken] {
+  ) async throws -> [SyntaxHighlightingToken] {
     async let tree = syntaxTreeManager.syntaxTree(for: snapshot)
-    async let semanticTokens = await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+    async let semanticTokens = semanticHighlightingTokens(for: snapshot)
 
     let range =
       if let range = range.flatMap({ $0.byteSourceRange(in: snapshot) }) {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

bnbarham:
> This also changes the implementation of the semantic tokens LSP request slightly: When the sourcekitd request to get semantic tokens fails, we now fail the entire LSP request, instead of returning the tokens from the synax tree. I think that’s reasonable because the editor did ask for semantic tokens, not syntactic tokens.

LSP doesn't have a way to ask for syntactic tokens though, so IMO giving syntactic as a fallback is reasonable enough. Otherwise we should probably keep the old support in (which I'd really prefer not doing but...).

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Linux

PR #943, opened 2023-10-27
Make `DocumentManager.latestSnapshot` throw if no snapshot exists for the URI

ahoppen:
This simplifies most calls that would log an error + return an empty response if no document was found.

PR #949, opened 2023-10-27
Remove assertion that we can't register for change notifications twice in `SwiftPMWorkspace`

ahoppen:
There’s no harm in registering for change notifications for the same file twice. Currently, this can happen if you open two files that have the same main file.

 `BuildSystemManager` checks that there are no more files referencing a main file before calling `unregisterForChangeNotifications`, so this is where the “complicated” logic lives that checks if sourcekit-lsp really isn’t interested in a file anymore.

While looking through the code I noticed that I could remove the `language` parameter from `BuildSystem.registerForChangeNotifications` because it isn’t used anywhere.

rdar://117603105

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1186,9 +1186,6 @@ extension SourceKitServer {
   }
 
   private func openDocument(_ note: DidOpenTextDocumentNotification, workspace: Workspace) async {
-    if documentManager.latestSnapshot(note.textDocument.uri) != nil {
-      logger.fault("Document '\(note.textDocument.uri.forLogging)' is already open!")
-    }
     // Immediately open the document even if the build system isn't ready. This is important since
     // we check that the document is open when we receive messages from the build system.
     documentManager.open(note)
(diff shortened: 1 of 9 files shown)


Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1185,6 +1185,9 @@ extension SourceKitServer {
   }
 
   private func openDocument(_ note: DidOpenTextDocumentNotification, workspace: Workspace) async {
+    if documentManager.latestSnapshot(note.textDocument.uri) != nil {
+      logger.fault("Document '\(note.textDocument.uri.forLogging)' is already open!")
+    }
```
hamishknight, line 1190: Doesn't DocumentManager already log this?

hamishknight (approved):

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Only after https://github.com/apple/sourcekit-lsp/pull/943

ahoppen: Oh, yes it does. Removing it again.

ahoppen:
@swift-ci Please test

PR #950, opened 2023-10-27
Don't call into the `MainFileProvider` for Swift files

ahoppen:
Swift doesn’t have header and main files. A Swift file is always its own main file. Skip the call into the main file provider.


bnbarham (approved):

hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #951, opened 2023-10-27
Emit signposts for request handling

ahoppen:
This allows you to trace sourcekit-lsp in Instruments to get a feeling for how long requests take to be handled. I’m not sure how useful this will be in practice but we can’t really know until we have it in a build that we see an issue with, so I’d propose we put it in and if we never use it in the next year or two, we can rip it back out again.

### DIFF
--- a/Sources/LSPLogging/Logging.swift
+++ b/Sources/LSPLogging/Logging.swift
@@ -27,9 +27,17 @@ import os  // os_log
 
 public typealias LogLevel = os.OSLogType
 public typealias Logger = os.Logger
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):
Seems reasonable to me

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Issue #879, opened 2023-10-09
Change `staticSourceKitTibsWorkspace` to define the sources in-line with the test instead of in a directory


ahoppen:
Having the source of the test cases defined out-of-line in `LSPTestSupport/INPUTS` makes test cases a lot harder to read. It would be much nicer if we could specify the contents of the source files inside the test case itself.

While doing this, we should also consider if we can simplify `staticSourceKitTibsWorkspace` further (I’m not sure if we can or, if we can, how).

ahoppen:
Tracked in Apple’s issue tracker as rdar://116705679

PR #952, opened 2023-10-30
Change main file tests to define sources inline and remove tibs testing infrastructure

ahoppen:
Migrate the last tests that used the tibs workspace to define sources inline and remove the testing infrastructure that relied on TIBS.

Resolves https://github.com/apple/sourcekit-lsp/issues/879


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKTestSupport/SwiftPMTestWorkspace.swift
```diff
@@ -78,6 +78,8 @@ public class SwiftPMTestWorkspace: MultiFileTestWorkspace {
       "-Xswiftc", "-index-ignore-system-modules",
       "-Xcc", "-index-ignore-system-symbols",
     ]
-    try await Process.checkNonZeroExit(arguments: arguments)
+    var environment = ProcessEnv.vars
+    environment["SWIFTPM_ENABLE_CLANG_INDEX_STORE"] = "1"
```
bnbarham, line 82: Could you add a comment with the issue/radar number for fixing this in SPM?

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
```diff
@@ -19,78 +19,197 @@ import SourceKitLSP
...
-        #include "bridging.h"
-        void d_new(void) { bridging(); }
+  func testMainFileForHeaderInPackageTarget() async throws {
+    let ws = try await SwiftPMTestWorkspace(
+      files: [
+        "MyLibrary/include/MyLibrary.h": """
+        void bridging(void) {
+          int VARIABLE_NAME = 1;
+        }
+        """,
```
bnbarham, line 29: Is this how swift-format formats this? I'd personally prefer indented once (or twice)

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
```diff
@@ -19,78 +19,197 @@ import SourceKitLSP
...
+        // swift-tools-version: 5.7
+
+        import PackageDescription
+
+        let package = Package(
+          name: "MyLibrary",
+          targets: [
+            .target(
+              name: "MyLibrary",
+              cSettings: [.define("VARIABLE_NAME", to: "fromMyLibrary"), .unsafeFlags(["-Wunused-variable"])]
```
bnbarham, line 44: Took me a little bit to realize why we were doing this - worth a comment IMO (ie. that we're defining `VARIABLE_NAME` here to make sure that the build settings from the main file are being used for the header).

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
```diff
@@ -19,78 +19,197 @@ import SourceKitLSP
...
+        """,
+      build: false
+    )
+
+    _ = try ws.openDocument("shared.h", language: .c)
 
-    mainFilesDelegate.expectation = expectation(description: "main files changed after edit")
-    try ws.buildAndIndex()
+    // Before we build, we shouldn't have an index and thus we don't infer the build setting for 'shared.h' form
+    // 'MyLibrary.c'. Hence, there's we don't have the '-Wunused-variable' build setting and thus no diagnostics.
```
bnbarham, line 93:
```suggestion
    // 'MyLibrary.c'. Hence, we don't have the '-Wunused-variable' build setting and thus no diagnostics.
```

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
```diff
@@ -19,78 +19,197 @@ import SourceKitLSP
...
+        "Sources/shared.h": """
+        void bridging(void) {
+          int VARIABLE_NAME = 1;
+        }
+        """,
+        "Sources/MyLibrary/include/MyLibrary.h": "",
+        "Sources/MyLibrary/MyLibrary.c": """
+        #include "$TEST_DIR/Sources/shared.h"
+        """,
+        "Sources/MyFancyLibrary/include/MyFancyLibraryh": "",
```
bnbarham, line 167: This file is unused isn't it?

Documentation/Development.md
```diff
@@ -134,81 +134,6 @@ Where "N" configures the log verbosity and is one of the following numbers: 0 (e
 
 As much as is practical, all code should be covered by tests. New tests can be added under the `Tests` directory and should use `XCTest`. The rest of this section will describe the additional tools available in the `SKTestSupport` module to make it easier to write good and efficient tests.
```
bnbarham, line 136: I assume your opinion here is that the tests are now obvious enough without any description?

hamishknight (approved):

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
ahoppen: Yes, this is formatted using swift-format.

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
ahoppen: I mentioned that we use `VARIABLE_NAME` to communicate the build settings in all test cases now.

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
ahoppen: SwiftPM refuses to build if a C target does not have an `include` directory. And in the current infrastructure, the only way to create a directory is to put a file in there.

Documentation/Development.md
ahoppen: My opinion here is that we should write more documentation that’s up-to-date.

This paragraph in particular doesn’t add any value IMO. It’s mostly just: There should be tests and we use XCTest, which, I think is obvious.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitLSPTests/MainFilesProviderTests.swift
bnbarham: I see. That's... okay. Could just add a `dummy` or something though (to make that more clear)

ahoppen: Good idea. Renamed to `dummy.h`.

Documentation/Development.md
ahoppen: OK, for those, I would say that 

> As much as is practical, all code should be covered by tests. New tests can be added under the `Tests` directory and should use `XCTest`.

is obvious and not something that needs to be mentioned in documentation.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #953, opened 2023-10-30
Fix issue that `SOURCEKITLSP_LOG_LEVEL` environment variable was not honored

ahoppen:
The initializer of `NonDarwinLogLevel` always returned `nil`, even if a valid value was passed. Thus, the log level was always `default`, independent of the value passed to `SOURCEKITLSP_LOG_LEVEL`.

While at it, change the default log level to be `debug` (opposed to `default`) when sourcekit-lsp is built in debug mode. That way you get full log output when running tests on Linux by default, which makes debugging easier.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #954, opened 2023-10-30
Improve test scratch directory creation

### LINKED ISSUE #929: Improve test scratch directory creation
There are a couple of ways in which we should improve the creation of test scratch directory creation:
- We should shorten the name of the directory by not including a full UUID, a few characters should be sufficient to ensure there are no conflicts
- Instead of having a `<UUID>/<testName>` sub-directory structure we should have a single folder `<testName>-<UUID>`. That also ensures that the UUID directories don’t stay around after clean-up.
- Add an environment variable to not clean up the test directories.

ahoppen:
Addresses the comments in #929 

> - We should shorten the name of the directory by not including a full UUID, a few characters should be sufficient to ensure there are no conflicts
> - Instead of having a `<UUID>/<testName>` sub-directory structure we should have a single folder `<testName>-<UUID>`. That also ensures that the UUID directories don’t stay around after clean-up.
> - Add an environment variable to not clean up the test directories.

Resolves #929
rdar://117547610

### DIFF
--- a/Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
+++ b/Sources/SKTestSupport/IndexedSingleSwiftFileWorkspace.swift
@@ -38,7 +38,7 @@ public struct IndexedSingleSwiftFileWorkspace {
     _ markedText: String,
     indexSystemModules: Bool = false,
     workspaceDirectory: URL? = nil,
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #956, opened 2023-10-30
Reduce logging level for unknown notifications or requests

ahoppen:
We are, for example, ignoring the `setTrace` notification that VS Code sends to use. That’s a totally valid thing to do and we shouldn’t log an error for that.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -176,26 +176,56 @@ fileprivate enum TaskMetadata: DependencyTracker {
 
   init(_ notification: any NotificationType) {
     switch notification {
-    case is InitializedNotification:
-      self = .globalConfigurationChange
-    case is CancelRequestNotification:
+    case is CancelRequestNotification: 
       self = .freestanding
-    case is ExitNotification:
-      self = .globalConfigurationChange
-    case let notification as DidOpenTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
-    case let notification as DidCloseTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
-    case let notification as DidChangeTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
-    case is DidChangeWorkspaceFoldersNotification:
-      self = .globalConfigurationChange
-    case is DidChangeWatchedFilesNotification:
+    case is CancelWorkDoneProgressNotification: 
       self = .freestanding
-    case let notification as WillSaveTextDocumentNotification:
+    case is DidChangeConfigurationNotification: 
+      self = .globalConfigurationChange
+    case let notification as DidChangeNotebookDocumentNotification: 
+      self = .documentUpdate(notification.notebookDocument.uri)
+    case let notification as DidChangeTextDocumentNotification: 
       self = .documentUpdate(notification.textDocument.uri)
-    case let notification as DidSaveTextDocumentNotification:
+    case is DidChangeWatchedFilesNotification: 
+      self = .freestanding
+    case is DidChangeWorkspaceFoldersNotification: 
+      self = .globalConfigurationChange
+    case let notification as DidCloseNotebookDocumentNotification: 
+      self = .documentUpdate(notification.notebookDocument.uri)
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -197,7 +197,7 @@ fileprivate enum TaskMetadata: DependencyTracker {
     case let notification as DidSaveTextDocumentNotification:
       self = .documentUpdate(notification.textDocument.uri)
     default:
-      logger.error(
+      logger.log(
```
bnbarham, line 200: For cases we're specifically ignoring, would it be worth having a case here and then keeping the error for truly unknown cases?

ahoppen: Yeah, might be the better option. I added cases for all known notifications and requests now.

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -209,32 +214,43 @@ fileprivate enum TaskMetadata: DependencyTracker {
...
-      self = .globalConfigurationChange
-    case is WorkspaceSymbolsRequest:
-      self = .freestanding
-    case is BarrierRequest:
-      self = .globalConfigurationChange
-    case is PollIndexRequest:
-      self = .globalConfigurationChange
-    case let request as ExecuteCommandRequest:
+    case let request as any TextDocumentRequest: self = .documentRequest(request.textDocument.uri)
+    case is ApplyEditRequest: fatalError("T")
```
bnbarham, line 218: `T` 😅?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -176,26 +176,31 @@ fileprivate enum TaskMetadata: DependencyTracker {
...
-      self = .freestanding
-    case let notification as WillSaveTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
-    case let notification as DidSaveTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
+    case is CancelRequestNotification: self = .freestanding
+    case is CancelWorkDoneProgressNotification: self = .freestanding
+    case is DidChangeConfigurationNotification: self = .globalConfigurationChange
+    case let noti as DidChangeNotebookDocumentNotification: self = .documentUpdate(noti.notebookDocument.uri)
+    case let notification as DidChangeTextDocumentNotification: self = .documentUpdate(notification.textDocument.uri)
```
hamishknight, line 183: Nit: Could we make these names consistent?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -176,26 +176,31 @@ fileprivate enum TaskMetadata: DependencyTracker {
...
-      self = .globalConfigurationChange
-    case is DidChangeWatchedFilesNotification:
-      self = .freestanding
-    case let notification as WillSaveTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
-    case let notification as DidSaveTextDocumentNotification:
-      self = .documentUpdate(notification.textDocument.uri)
+    case is CancelRequestNotification: self = .freestanding
+    case is CancelWorkDoneProgressNotification: self = .freestanding
+    case is DidChangeConfigurationNotification: self = .globalConfigurationChange
```
hamishknight, line 181: Nit: IMO the switch is easier to read with the case bodies on a new line, but I don't feel too strongly about it.

hamishknight (approved):

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: I don’t fee strongly about it either but had a slight preference for the single-line formatting. But since you prefer the multi-line and we need to abbreviate `notification` to make it fit on a single line, I changed it back to multi-line.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #957, opened 2023-10-30
Always require a prefix for `orLog` calls

ahoppen:
Otherwise it’s very easy to end up with log messages like `requestCancelled` for which it’s not clear where they are coming from.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #960, opened 2023-11-01
Skip all semantic tokens requests if sourcekitd doesn't support it

ahoppen:
I was mistakingly only catching the `unknown request: source.request.semantic_tokens` in `openAndPerformSemanticTokensRequest`. We should be catching it in `performSemanticTokensRequest`, which gets invoked directly from some tests.


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #922, opened 2023-10-25
Cache diagnostics so that we don’t need to re-request them on every codeAction request


ahoppen:
We currently request diagnostics for every `CodeAction` request even if we already computed diagnostics for the same snapshot of the file.

Even though sourcekitd caches the AST, getting the diagnostics takes about 2ms. Given that VS Code sends a code action request on every cursor move, this adds up. We should check if we can improve the performance of the code action request by caching the diagnostics in SourceKit-LSP.

ahoppen:
Tracked in Apple’s issue tracker as rdar://117447328

ahoppen:
If anyone is interested in this, we are requesting the diagnostics here

https://github.com/apple/sourcekit-lsp/blob/182deb75bb7bf55ba5ad8daaad9feae2c44c4b87/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift#L962-L966

And we should implement a cache that is similar to `SyntaxTreeManager`.

ahoppen:
Definitely!

PR #965, opened 2023-11-09
Add cache to avoid requesting diagnostics from sourcekitd frequently.

joehsieh:
I'm following the way we cached computations in SyntaxTreeManager.
https://github.com/apple/sourcekit-lsp/issues/922#issuecomment-1783343524

### DIFF
--- a/Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
+++ b/Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
@@ -1,0 +1,180 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+import SourceKitD
+import SwiftParserDiagnostics
+
+actor DiagnosticReportManager {
+  /// A task to produce diagnostics, either from a diagnostics request to `sourcektid` or by using the built-in swift-syntax.
+  private typealias ReportTask = Task<RelatedFullDocumentDiagnosticReport, Error>
+
+  private let sourcekitd: SourceKitD
+  private let syntaxTreeManager: SyntaxTreeManager
+  private let documentManager: DocumentManager
+  private let clientHasDiagnosticsCodeDescriptionSupport: Bool
+
+  private nonisolated var keys: sourcekitd_keys { return sourcekitd.keys }
+  private nonisolated var requests: sourcekitd_requests { return sourcekitd.requests }
+
+  /// The cache that stores reportTasks for snapshot id and buildSettings
+  ///
+  /// Conceptually, this is a dictionary. To prevent excessive memory usage we
+  /// only keep `cacheSize` entries within the array. Older entries are at the
+  /// end of the list, newer entries at the front.
+  private var reportTaskCache:
+    [(
+      snapshotID: DocumentSnapshot.ID,
+      buildSettings: SwiftCompileCommand?,
+      reportTask: ReportTask
+    )] = []
+
+  /// The number of reportTasks to keep
+  ///
+  /// - Note: This has been chosen without scientific measurements.
+  private let cacheSize = 5
+
+  init(
+    sourcekitd: SourceKitD,
+    syntaxTreeManager: SyntaxTreeManager,
+    documentManager: DocumentManager,
+    clientHasDiagnosticsCodeDescriptionSupport: Bool
+  ) {
+    self.sourcekitd = sourcekitd
+    self.syntaxTreeManager = syntaxTreeManager
+    self.documentManager = documentManager
+    self.clientHasDiagnosticsCodeDescriptionSupport = clientHasDiagnosticsCodeDescriptionSupport
+  }
+
+  func diagnosticReport(
+    for snapshot: DocumentSnapshot,
+    buildSettings: SwiftCompileCommand?
+  ) async throws -> RelatedFullDocumentDiagnosticReport {
+    if let reportTask = reportTask(for: snapshot.id, buildSettings: buildSettings) {
+      return try await reportTask.value
+    }
+    let reportTask: Task<RelatedFullDocumentDiagnosticReport, Error>
+    if let buildSettings, !buildSettings.isFallback {
+      reportTask = Task {
+        return try await requestReport(with: snapshot, compilerArgs: buildSettings.compilerArgs)
+      }
+    } else {
+      logger.log(
+        "Producing syntactic diagnostics from the built-in swift-syntax because we \(buildSettings != nil ? "have fallback build settings" : "don't have build settings", privacy: .public))"
+      )
+      // If we don't have build settings or we only have fallback build settings,
+      // sourcekitd won't be able to give us accurate semantic diagnostics.
+      // Fall back to providing syntactic diagnostics from the built-in
+      // swift-syntax. That's the best we can do for now.
+      reportTask = Task {
+        return try await requestFallbackReport(with: snapshot)
+      }
+    }
+    setReportTask(for: snapshot.id, buildSettings: buildSettings, reportTask: reportTask)
+    return try await reportTask.value
+  }
+
+  func removeItemsFromCache(with uri: DocumentURI) async {
+    for item in reportTaskCache {
+      if item.snapshotID.uri == uri {
+        item.reportTask.cancel()
+      }
+    }
+    reportTaskCache.removeAll(where: { $0.snapshotID.uri == uri })
+  }
+
+  private func requestReport(
+    with snapshot: DocumentSnapshot,
+    compilerArgs: [String]
+  ) async throws -> LanguageServerProtocol.RelatedFullDocumentDiagnosticReport {
+    try Task.checkCancellation()
+
+    let keys = self.keys
+
+    let skreq = sourcekitd.dictionary([
+      keys.request: requests.diagnostics,
+      keys.sourcefile: snapshot.uri.pseudoPath,
+      keys.compilerargs: compilerArgs as [SKDValue],
+    ])
+
+    let dict = try await self.sourcekitd.send(skreq, fileContents: snapshot.text)
+
+    try Task.checkCancellation()
+    guard (try? documentManager.latestSnapshot(snapshot.uri).id) == snapshot.id else {
+      // Check that the document wasn't modified while we were getting diagnostics. This could happen because we are
+      // calling `fullDocumentDiagnosticReport` from `publishDiagnosticsIfNeeded` outside of `messageHandlingQueue`
+      // and thus a concurrent edit is possible while we are waiting for the sourcekitd request to return a result.
+      throw ResponseError.unknown("Document was modified while loading diagnostics")
+    }
+
+    let diagnostics: [Diagnostic] =
+      dict[keys.diagnostics]?.compactMap({ diag in
+        Diagnostic(
+          diag,
+          in: snapshot,
+          useEducationalNoteAsCode: self.clientHasDiagnosticsCodeDescriptionSupport
+        )
+      }) ?? []
+
+    return RelatedFullDocumentDiagnosticReport(items: diagnostics)
+  }
+
+  private func requestFallbackReport(
+    with snapshot: DocumentSnapshot
+  ) async throws -> LanguageServerProtocol.RelatedFullDocumentDiagnosticReport {
+    // If we don't have build settings or we only have fallback build settings,
+    // sourcekitd won't be able to give us accurate semantic diagnostics.
+    // Fall back to providing syntactic diagnostics from the built-in
+    // swift-syntax. That's the best we can do for now.
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let swiftSyntaxDiagnostics = ParseDiagnosticsGenerator.diagnostics(for: syntaxTree)
+    let diagnostics = swiftSyntaxDiagnostics.compactMap { (diag) -> Diagnostic? in
+      if diag.diagnosticID == StaticTokenError.editorPlaceholder.diagnosticID {
+        // Ignore errors about editor placeholders in the source file, similar to how sourcekitd ignores them.
+        return nil
+      }
+      return Diagnostic(diag, in: snapshot)
+    }
+    return RelatedFullDocumentDiagnosticReport(items: diagnostics)
+  }
+
+  /// The reportTask for the given document snapshot and buildSettings.
+  private func reportTask(
+    for snapshotID: DocumentSnapshot.ID,
+    buildSettings: SwiftCompileCommand?
+  ) -> ReportTask? {
+    return reportTaskCache.first(where: { $0.snapshotID == snapshotID && $0.buildSettings == buildSettings })?
+      .reportTask
+  }
+
+  /// Set the reportTask for the given document snapshot and buildSettings.
+  ///
+  /// If we are already storing `cacheSize` many reports, the oldest one
+  /// will get discarded.
+  private func setReportTask(
+    for snapshotID: DocumentSnapshot.ID,
+    buildSettings: SwiftCompileCommand?,
+    reportTask: ReportTask
+  ) {
+    reportTaskCache.insert((snapshotID, buildSettings, reportTask), at: 0)
+
+    // Remove any reportTasks for old versions of this document.
+    reportTaskCache.removeAll(where: { $0.snapshotID < snapshotID })
+
+    // If we still have more than `cacheSize` reportTasks, delete the ones that
+    // were produced last. We can always re-request them on-demand.
+    while reportTaskCache.count > cacheSize {
+      reportTaskCache.removeLast()
+    }
+  }
+}
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -139,6 +139,8 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
 
   private var stateChangeHandlers: [(_ oldState: LanguageServerState, _ newState: LanguageServerState) -> Void] = []
 
+  private let diagnosticReportManager: DiagnosticReportManager
+
   /// Creates a language server for the given client using the sourcekitd dylib specified in `toolchain`.
   /// `reopenDocuments` is a closure that will be called if sourcekitd crashes and the `SwiftLanguageServer` asks its parent server to reopen all of its documents.
   /// Returns `nil` if `sourcektid` couldn't be found.
@@ -157,6 +159,12 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
     self.state = .connected
     self.generatedInterfacesPath = options.generatedInterfacesPath.asURL
     try FileManager.default.createDirectory(at: generatedInterfacesPath, withIntermediateDirectories: true)
+    self.diagnosticReportManager = DiagnosticReportManager(
+      sourcekitd: self.sourcekitd,
+      syntaxTreeManager: syntaxTreeManager,
+      documentManager: documentManager,
+      clientHasDiagnosticsCodeDescriptionSupport: capabilityRegistry.clientHasDiagnosticsCodeDescriptionSupport
+    )
   }
 
   /// - Important: For testing only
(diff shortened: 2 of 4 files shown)


Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -139,6 +139,22 @@ public actor SwiftLanguageServer: ToolchainLanguageServer {
...
+  private var reportCache:
+    [(
+      snapshotID: DocumentSnapshot.ID,
+      report: RelatedFullDocumentDiagnosticReport
+    )] = []
+
+  /// The number of reports to keep
+  ///
+  /// - Note: This has been chosen without scientific measurements.
+  private let cacheSize = 5
```
joehsieh, line 156: i tentatively set the size of cache to 5 because i'm not sure if there is a better magic-number for it.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -1065,9 +1082,13 @@ extension SwiftLanguageServer {
   }
 
   private func fullDocumentDiagnosticReport(
-    _ req: DocumentDiagnosticsRequest
+    _ req: DocumentDiagnosticsRequest,
+    useCache: Bool = false
   ) async throws -> RelatedFullDocumentDiagnosticReport {
     let snapshot = try documentManager.latestSnapshot(req.textDocument.uri)
+    if useCache, let report = report(for: snapshot.id) {
+      return report
```
joehsieh, line 1090: i was trying to add a test for the code changes, but i didn't find a way to do it.
at least this change doesn't break any existing tests.

ahoppen (commented):
Instead of implicitly setting the cached reports in `SwiftLanguageServer`, I would prefer to have a similar design to `SyntaxTreeManager`. `SyntaxTreeManager` is an encapsulated class that has a single function

```swift
func diagnosticReport(for snapshot: DocumentSnapshot, buildSettings: SwiftCompileCommand) throws -> RelatedFullDocumentDiagnosticReport
```

And then the manager class will decide whether it has diagnostics cached already and, if not, sends a request to `sourcekitd` to compute them.

joehsieh:
ah, right. in this way, we can make the code testable.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,98 @@
...
+  /// Notify the delegate that compute a report with a given snapshot
+  func reportWithSnapshotReceived(_ snapshot: DocumentSnapshot, compilerArgs: [String]) async throws
+    -> RelatedFullDocumentDiagnosticReport
+
+  /// Notify the delegate that compute a report with a given snapshot for fallback
+  func fallbackReportWithSnapshotReceived(_ snapshot: DocumentSnapshot) async throws
+    -> RelatedFullDocumentDiagnosticReport
+}
+
+actor DiagnosticReportManager {
```
joehsieh, line 26: note: I'll add tests for this new actor.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,98 @@
...
+  /// The cache that stores reports for snapshot ids
+  ///
+  /// Conceptually, this is a dictionary. To prevent excessive memory usage we
+  /// only keep `cacheSize` entries within the array. Older entries are at the
+  /// end of the list, newer entries at the front.
+  private var reportCache:
+    [(
+      snapshotID: DocumentSnapshot.ID,
```
ahoppen, line 34: The build setting should also be part of the key. If the build settings change, the diagnostics need to be re-computed. With the current implementation, we would just return the diagnostics for the document with the old build settings.

joehsieh: good catch 👍 , let me fix it.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,98 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+
+protocol DiagnosticReportManagerDelegate: AnyObject {
```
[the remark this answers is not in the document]
joehsieh: i was considering the the way you mentioned above before, but regarding the testability of DiagnosticReportManager, i decided to have a delegate to make the testing easier.
it makes sense to remove the delegate to avoid playing ping-pong (let me update the code for it first), but i was wondering if there is a way to test it.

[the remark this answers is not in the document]
ahoppen: I think it’s fine to test it by just sending the diagnostic requests followed by a code action request (which I think we already have a test for). That should trigger the path that computes diagnostics and also the one that re-used cached diagnostics.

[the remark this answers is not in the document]
joehsieh: ah, okie. then, let me make this PR ready for review.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+    clientHasDiagnosticsCodeDescriptionSupport: Bool
+  ) {
+    self.sourcekitd = sourcekitd
+    self.syntaxTreeManager = syntaxTreeManager
+    self.documentManager = documentManager
+    self.clientHasDiagnosticsCodeDescriptionSupport = clientHasDiagnosticsCodeDescriptionSupport
+  }
+
+  func diagnosticReport(for snapshot: DocumentSnapshot, buildSettings: SwiftCompileCommand?, useCache: Bool = false)
+    async throws -> RelatedFullDocumentDiagnosticReport
```
ahoppen, line 57: Nitpick: I prefer to break this on the arguments since it allows for easier extensions if we add more arguments in the future. Also, it looks nicer IMO.

```suggestion
  func diagnosticReport(
    for snapshot: DocumentSnapshot, 
    buildSettings: SwiftCompileCommand?, 
    useCache: Bool = false
  ) async throws -> RelatedFullDocumentDiagnosticReport
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+  {
+    if useCache, let report = report(for: snapshot.id, buildSettings: buildSettings) {
+      return report
+    }
+    guard let buildSettings = buildSettings else {
```
ahoppen, line 62: Can be shortened

```suggestion
    guard let buildSettings else {
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+      // Fall back to providing syntactic diagnostics from the built-in
+      // swift-syntax. That's the best we can do for now.
+      let report = try await requestFallbackReport(with: snapshot)
+      setReport(for: snapshot.id, buildSettings: nil, report: report)
+      return report
+    }
+    guard !buildSettings.isFallback else {
+      logger.log(
+        "Producing syntactic diagnostics from the built-in swift-syntax because we have fallback arguments"
+      )
```
ahoppen, line 77: Do we need to replicate the logic to call `requestFallbackReport`. Couldn’t we just have the following above? We would maybe only just need to check `buildSettings != nil` again to determine which message to log.
```swift
guard let buildSettings, !buildSettings.isFallback else {

File: Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift [hunk 9]
```diff
@@ -0,0 +1,175 @@
...
+    }
+    guard let buildSettings = buildSettings else {
+      logger.log(
+        "Producing syntactic diagnostics from the built-in swift-syntax because we don't have buildSettings"
+      )
+      // If we don't have build settings,
+      // sourcekitd won't be able to give us accurate semantic diagnostics.
+      // Fall back to providing syntactic diagnostics from the built-in
+      // swift-syntax. That's the best we can do for now.
+      let report = try await requestFallbackReport(with: snapshot)
```
ahoppen, line 70: There’s currently a (kind of theoretical) race condition and duplicated computation here: If we request two diagnostic reports for the same file and build settings, both of them trigger `requestFallbackReport` and then both of them call `setReport`. While `requestFallbackReport` should be returning the same data for both calls, if it didn’t, we would get inconsistent results for the two calls.

To avoid the duplicated computation and the race condition here, `SyntaxTreeManager` has the `syntaxTreeComputations` dictionary. Because we map each `snapshotID` to a `Task`, we can insert a value to it immediately when the syntax tree is requested and thus we can guarantee that every syntax tree is only every built once. I think we should do the same here (in particular because getting diagnostics from sourcekitd can be a lot more expensive the building a syntax tree).

The same applies to the `requestReport` call below.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+    let report = try await requestReport(with: snapshot, compilerArgs: buildSettings.compilerArgs)
+    setReport(for: snapshot.id, buildSettings: buildSettings, report: report)
+    return report
+  }
+}
+
+private extension DiagnosticReportManager {
+  func requestReport(with snapshot: DocumentSnapshot, compilerArgs: [String]) async throws
+    -> LanguageServerProtocol.RelatedFullDocumentDiagnosticReport
```
ahoppen, line 94: Similar to above, I would prefer to break this as

```suggestion
  func requestReport(
    with snapshot: DocumentSnapshot, 
    compilerArgs: [String]
  ) async throws -> LanguageServerProtocol.RelatedFullDocumentDiagnosticReport
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+    skreq[keys.compilerargs] = compilerArgs
+
+    let dict = try await self.sourcekitd.send(skreq, fileContents: snapshot.text)
+
+    try Task.checkCancellation()
+    guard (try? documentManager.latestSnapshot(snapshot.uri).id) == snapshot.id else {
+      // Check that the document wasn't modified while we were getting diagnostics. This could happen because we are
+      // calling `fullDocumentDiagnosticReport` from `publishDiagnosticsIfNeeded` outside of `messageHandlingQueue`
+      // and thus a concurrent edit is possible while we are waiting for the sourcekitd request to return a result.
+      throw ResponseError.unknown("Document was modified while loading document")
```
ahoppen, line 114: Probably not your typo but mine when I originally wrote this, but might as well fix it.

```suggestion
      throw ResponseError.unknown("Document was modified while loading diagnostics")
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    let supportsCodeDescription = self.clientHasDiagnosticsCodeDescriptionSupport
```
ahoppen, line 117: I don’t think we need this variable. We can just pass `self.clientHasDiagnosticsCodeDescriptionSupport` below.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
joehsieh: sorry, i didn't revert this change. let me do it.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
joehsieh: i thought the cache in an actor would be thread-safe, but i can see we're using Task in SyntaxTreeManager.
let me follow the way we're doing first.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
joehsieh: you're right.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+      if let diag = Diagnostic(diag, in: snapshot, useEducationalNoteAsCode: supportsCodeDescription) {
+        diagnostics.append(diag)
+      }
+      return true
+    }
+
+    return RelatedFullDocumentDiagnosticReport(items: diagnostics)
+  }
+
+  func requestFallbackReport(with snapshot: DocumentSnapshot) async throws
```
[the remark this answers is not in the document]
joehsieh: i put all the private funcs in a `private extension`, but there is still a compile error. `method must be declared private because its result uses a private type`. let me move all the funcs back to the actor and mark them private.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,175 @@
...
+    documentManager: DocumentManager,
+    clientHasDiagnosticsCodeDescriptionSupport: Bool
+  ) {
+    self.sourcekitd = sourcekitd
+    self.syntaxTreeManager = syntaxTreeManager
+    self.documentManager = documentManager
+    self.clientHasDiagnosticsCodeDescriptionSupport = clientHasDiagnosticsCodeDescriptionSupport
+  }
+
+  func diagnosticReport(for snapshot: DocumentSnapshot, buildSettings: SwiftCompileCommand?, useCache: Bool = false)
```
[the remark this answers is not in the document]
joehsieh: right, after second thought, what i should do is to reset cache instead of skip caching it in some scenarios.
i've corrected the logic. can check it again. thanks!

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,182 @@
...
+      return report
+    }
+    let reportTask: Task<RelatedFullDocumentDiagnosticReport, Error>
+    if let buildSettings, !buildSettings.isFallback {
+      reportTask = Task {
+        return try await requestReport(with: snapshot, compilerArgs: buildSettings.compilerArgs)
+      }
+    } else {
+      logger.log(
+        "Producing syntactic diagnostics from the built-in swift-syntax because we \(buildSettings != nil ? "have fallback build settings" : "don't have build settings")"
```
ahoppen, line 73: I think strings are redacted from logs by default. Since this doesn’t contain any private information, we should set it to public

```suggestion
        "Producing syntactic diagnostics from the built-in swift-syntax because we \(buildSettings != nil ? "have fallback build settings" : "don't have build settings", privacy: .public)"
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
ahoppen: Actors are thread-safe for low-level data races (where multiple threads try to access the same variable simultaneously) but they can still have high-level data races. These occur if you store some kind of state in local variables but that state can change at suspension points, aka. `await` calls. That’s why you explicitly spell out `await` in the call.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,182 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+import SourceKitD
+import SwiftParserDiagnostics
+
+actor DiagnosticReportManager {
+  /// A task that gets report from sourcekitd
```
ahoppen, line 19:
```suggestion
  /// A task to produce diagnostics, either from a diagnostics request to `sourcektid` or by using the built-in swift-syntax.
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,182 @@
...
+    self.clientHasDiagnosticsCodeDescriptionSupport = clientHasDiagnosticsCodeDescriptionSupport
+  }
+
+  func diagnosticReport(
+    for snapshot: DocumentSnapshot,
+    buildSettings: SwiftCompileCommand?
+  ) async throws -> RelatedFullDocumentDiagnosticReport {
+    if let report = try await reportTask(for: snapshot.id, buildSettings: buildSettings)?.value {
+      return report
+    }
```
ahoppen, line 65: I would rewrite this slightly, just to make sure we only do an `await` if we already have a `reportTask`. That way you can verify that `diagnosticReport` runs up to its `return` statement without a suspension point (aka `await` call) and thus you don’t need to be concerned with high-level data races.

```suggestion
    if let reportTask = reportTask(for: snapshot.id, buildSettings: buildSettings) {
      return try await report.value
    }
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,182 @@
...
+      // swift-syntax. That's the best we can do for now.
+      reportTask = Task {
+        return try await requestFallbackReport(with: snapshot)
+      }
+    }
+    setReportTask(for: snapshot.id, buildSettings: buildSettings, reportTask: reportTask)
+    return try await reportTask.value
+  }
+
+  func resetCache() async {
```
ahoppen, line 87: Why do we ever need to reset the cache? The snapshot ID + build settings should be sufficient that we re-request diagnostic whenever they might have changed. Or am I missing something?

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -86,7 +86,7 @@ final class LocalSwiftTests: XCTestCase {
 
     testClient.send(
       DidChangeTextDocumentNotification(
-        textDocument: .init(uri, version: 14),
+        textDocument: .init(uri, version: 15),
```
ahoppen, line 89: Oh, good catch 👍🏽

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
joehsieh: there are two tests will be failed without reseting cache.
```
Test Case '-[SourceKitLSPTests.SKTests testDependenciesUpdatedSwift]' failed (5.170 seconds).
Test Case '-[SourceKitLSPTests.LocalSwiftTests testDiagnosticsReopen]' failed (1.400 seconds).
```

i think that's because the cache for the file becomes dirty if the file is reopened with changed dependencies or content.

ahoppen: Ah, I see. That makes sense. But we should not need to reset the entire cache, we should only need to clear the diagnostics for the document that we are opening/closing/reopening.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,183 @@
...
+    reportTaskCache.removeAll(where: { $0.snapshotID.uri == uri })
+  }
+
+  private func requestReport(
+    with snapshot: DocumentSnapshot,
+    compilerArgs: [String]
+  ) async throws -> LanguageServerProtocol.RelatedFullDocumentDiagnosticReport {
+    try Task.checkCancellation()
+
+    let keys = self.keys
```
ahoppen, line 102: I don’t think this is necessary. `keys` should already refer to `self.keys` everywhere below.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
```diff
@@ -0,0 +1,183 @@
...
+    dict[keys.diagnostics]?.forEach { _, diag in
+      if let diag = Diagnostic(
+        diag,
+        in: snapshot,
+        useEducationalNoteAsCode: self.clientHasDiagnosticsCodeDescriptionSupport
+      ) {
+        diagnostics.append(diag)
+      }
+      return true
+    }
```
ahoppen, line 130: I recently introduced `map` and `compactMap` on `SKDResponseArray`, which we could use here

```suggestion
    let diagnostics: [Diagnostic] = dict[keys.diagnostics]?.compactMap({ diag in
      Diagnostic(
        diag,
        in: snapshot,
        useEducationalNoteAsCode: self.clientHasDiagnosticsCodeDescriptionSupport
      )
    }) ?? []
```

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
ahoppen: No need to apologize., thank you for fixing this issue!

ahoppen (commented):
I’ve got two nitpicks, otherwise looks good to merge for me.

Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
joehsieh: i checked the var keys the implementations is 
```
private nonisolated var keys: sourcekitd_keys { return sourcekitd.keys }
```
i think the purpose is to ensure the value of keys won't change in func requestReport

the same pattern is happening in other files. 
e.g. SwiftLanguageServer.swift, SemanticRefactoring.swift, etc.

ahoppen: Ah, OK. Fair enough.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #972, opened 2023-11-22
Don't cancel cursor info and related identifiers requests on subsequent requests

ahoppen:
SourceKit-LSP supports explicit cancellation and thus we don’t need to do the implicit cancellation of cursor info and related identifiers on subsequent requests.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #975, opened 2023-11-23
Support cursor info-based requests when cursor is placed at the end of an identifier

### LINKED ISSUE #820: Cursor position at end of symbol, doesn't capture symbol
Not sure if this is a VSCode issue or one SourceKit-LSP needs to solve. If you double click on a symbol and then use the context menu to look up references it will fail to find any references. This appears to be because the cursor position sent to Sourcekit-LSP is the end of the symbol and not the beginning.

See https://github.com/swift-server/vscode-swift/issues/602#issuecomment-1721197331 for more details.

ahoppen:
VS Code considers the position after an identifier as part of an identifier. Ie. if you have `let foo| = 1`, then it considers the cursor to be positioned at the identifier. This scenario is hit, when selecting an identifier by double-clicking it and then eg. performing jump-to-definition. In that case VS Code will send the position after the identifier. `sourcekitd`, on the other hand, does not consider the position after the identifier as part of the identifier. To bridge the gap here, normalize any positions inside, or directly after, an identifier to the identifier's start.

Fixes #820
rdar://115557453

### DIFF
--- a/Sources/SourceKitLSP/CMakeLists.txt
+++ b/Sources/SourceKitLSP/CMakeLists.txt
@@ -29,6 +29,7 @@ target_sources(SourceKitLSP PRIVATE
   Swift/SourceKitD+ResponseError.swift
   Swift/SwiftCommand.swift
   Swift/SwiftLanguageServer.swift
+  Swift/SymbolInfo.swift
   Swift/SyntaxHighlightingToken.swift
   Swift/SyntaxHighlightingTokenParser.swift
   Swift/SyntaxTreeManager.swift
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #977, opened 2023-11-23
Show overriden functions when performing jump-to-definition on a dynamic call

### LINKED ISSUE #809: Better jump to definition for implementations of a protocol
Consider this code:
```swift
protocol TestProtocol {
    func doThing()
}

struct TestImpl: TestProtocol { 
    func doThing() { }
}

func anyTestProtocol(value: any TestProtocol) {
    value.doThing()
}
```

Using VSCode with SSWG Swift extension, in `value.doThing()`, when you right click on `doThing()` to jump to definition, it'll just automatically jump to the protocol. This seems to be a SourceKit-LSP limitation.

Xcode/sourcekitd though is more smart about it and lets you choose between the protocol or one of the implementations:   

<kbd> Screenshot 2023-09-01 at 6 04 36 PM </kbd>

ahoppen:
Fixes #809
rdar://114864256

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1667,7 +1667,7 @@ extension SourceKitServer {
     symbols: [SymbolDetails],
     index: IndexStoreDB?,
     useLocalFallback: Bool = false,
-    extractOccurrences: (String, IndexStoreDB) -> [SymbolOccurrence]
+    extractOccurrences: (SymbolDetails, IndexStoreDB) -> [SymbolOccurrence]
   ) -> [(occurrence: SymbolOccurrence?, location: Location)] {
     guard let symbol = symbols.first else {
       return []
@@ -1680,11 +1680,11 @@ extension SourceKitServer {
       fallback = []
     }
 
-    guard let usr = symbol.usr, let index = index else {
+    guard let index = index else {
       return fallback
     }
 
-    let occurs = extractOccurrences(usr, index)
+    let occurs = extractOccurrences(symbol, index)
     let resolved = occurs.compactMap { occur in
       indexToLSPLocation(occur.location).map {
         (occurrence: occur, location: $0)
@@ -1725,11 +1725,23 @@ extension SourceKitServer {
     }
 
     let resolved = self.extractIndexedOccurrences(symbols: symbols, index: index, useLocalFallback: true) {
-      (usr, index) in
+      (symbolDetails, index) in
+      guard symbolDetails.isDynamic ?? false else {
+        // If the symbol isn't dynamic, we won't get more information from the index compared to what we already have
+        // in the symbolDetails response. Return an empty array so that `extractIndexedOccurrences` falls back to the
+        // location in the symbol details response.
+        return []
+      }
+      guard let usr = symbolDetails.usr else { return [] }
       logger.info("performing indexed jump-to-def with usr \(usr)")
       var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
       if occurs.isEmpty {
         occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
+      }
+      if symbolDetails.isDynamic ?? false {
+        occurs += occurs.flatMap {
+          index.occurrences(relatedToUSR: $0.symbol.usr, roles: .overrideOf)
+        }
       }
       return occurs
     }
@@ -1783,7 +1795,8 @@ extension SourceKitServer {
       )
     )
     let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
-    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (usr, index) in
+    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (symbolDetails, index) in
+      guard let usr = symbolDetails.usr else { return [] }
       var occurs = index.occurrences(ofUSR: usr, roles: .baseOf)
       if occurs.isEmpty {
         occurs = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
@@ -1806,7 +1819,8 @@ extension SourceKitServer {
       )
     )
     let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
-    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (usr, index) in
+    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (symbolDetails, index) in
+      guard let usr = symbolDetails.usr else { return [] }
       logger.info("performing indexed jump-to-def with usr \(usr)")
       var roles: SymbolRole = [.reference]
       if req.context.includeDeclaration {
@@ -1852,8 +1866,9 @@ extension SourceKitServer {
     )
     let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
     // For call hierarchy preparation we only locate the definition
-    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (usr, index) in
-      index.occurrences(ofUSR: usr, roles: [.definition, .declaration])
+    let extractedResult = self.extractIndexedOccurrences(symbols: symbols, index: index) { (symbolDetails, index) in
+      guard let usr = symbolDetails.usr else { return [] }
+      return index.occurrences(ofUSR: usr, roles: [.definition, .declaration])
     }
     return extractedResult.compactMap { info -> CallHierarchyItem? in
       guard let occurrence = info.occurrence else {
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1725,12 +1725,18 @@ extension SourceKitServer {
     }
 
     let resolved = self.extractIndexedOccurrences(symbols: symbols, index: index, useLocalFallback: true) {
-      (usr, index) in
+      (symbolDetails, index) in
```
bnbarham, line 1728: Is this the swift-format formatting?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1725,12 +1725,18 @@ extension SourceKitServer {
...
+      guard let usr = symbolDetails.usr else { return [] }
       logger.info("performing indexed jump-to-def with usr \(usr)")
       var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
       if occurs.isEmpty {
         occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
       }
+      if symbolDetails.isDynamic {
+        occurs += occurs.flatMap {
+          index.occurrences(relatedToUSR: $0.symbol.usr, roles: .overrideOf)
```
bnbarham, line 1737: Ideally we should filter by receivers here, ie.
  - Get the set of types to filter (receivers + their subtypes)
  - For each override check if it's a child of one of the types in the filter

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1725,12 +1725,18 @@ extension SourceKitServer {
     }
 
     let resolved = self.extractIndexedOccurrences(symbols: symbols, index: index, useLocalFallback: true) {
```
bnbarham, line 1727: We could just skip this if the symbol isn't dynamic - the index will be at most up to date as the location in the module, or worse. Though that would change if background indexing was added (or preparation, if we go that route instead).

bnbarham: Ah, I just noticed this is on SourceKitServer and not the language service. Would still be nice to skip for Swift IMO. As for clangd not returning dynamic, that's something we could change (the request we're using is one we specifically added for this use case - https://reviews.llvm.org/D54799). But in the meantime, I'd default to returning overrides if `isDynamic` isn't set.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Yes

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Added a filter by receiver types.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Changed it so that we don’t do an index lookup if `SymbolRequest` returns a location.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
-      logger.info("performing indexed jump-to-def with usr \(usr)")
-      var occurs = index.occurrences(ofUSR: usr, roles: [.definition])
-      if occurs.isEmpty {
-        occurs = index.occurrences(ofUSR: usr, roles: [.declaration])
+    guard let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index else {
+      return []
+    }
+    guard let usr = symbol.usr else { return [] }
+    logger.info("performing indexed jump-to-def with usr \(usr)")
+    var occurances = index.occurrences(ofUSR: usr, roles: [.definition])
```
bnbarham, line 1707:
```suggestion
    var occurrences = index.occurrences(ofUSR: usr, roles: [.definition])
```

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
     let symbols = try await languageService.symbolInfo(
       SymbolInfoRequest(
         textDocument: req.textDocument,
         position: req.position
       )
     )
-    let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
+    guard let symbol = symbols.first else {
+      return []
+    }
```
bnbarham, line 1683: I realize the old code only handled a single result, but is there any reason not to just handle multiple for all of these?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
-      return try await self.definitionInInterface(
-        req,
-        moduleName: moduleName,
-        symbolUSR: firstResolved.occurrence?.symbol.usr,
-        languageService: languageService
-      )
+    return try await occurances.asyncCompactMap { occurance in
+      if URL(fileURLWithPath: occurance.location.path).pathExtension == "swiftinterface" {
+        // if first resolved location is in `.swiftinterface` file. Use moduleName to return
+        // textual interface
```
bnbarham, line 1730: What does "first" here mean?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
+  ) async throws -> LocationsOrLocationLinksResponse? {
+    let indexBasedResponse = try await indexBasedDefinition(req, workspace: workspace, languageService: languageService)
     // If we're unable to handle the definition request using our index, see if the
     // language service can handle it (e.g. clangd can provide AST based definitions).
-    if locs.isEmpty {
-      return try await languageService.definition(req)
+    if indexBasedResponse.isEmpty {
+      do {
+        return try await languageService.definition(req)
+      } catch {}
```
bnbarham, line 1753: Why are we ignoring the error here?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
+      lazy var transitiveReceiverUsrs: [String] = transitiveSubtypeClosure(
+        ofUsrs: symbol.receiverUsrs ?? [],
+        index: index
+      )
+      occurances += occurances.flatMap {
+        let overrides = index.occurrences(relatedToUSR: $0.symbol.usr, roles: .overrideOf)
+        // Only contain overrides that are children of one of the receiver types or their subtypes.
+        return overrides.filter { override in
+          override.relations.contains(where: {
+            $0.roles.contains(.childOf) && transitiveReceiverUsrs.contains($0.symbol.usr)
```
bnbarham, line 1721:
```suggestion
            $0.roles.contains(.childOf) && (transitiveReceiverUsrs.isEmpty || transitiveReceiverUsrs.contains($0.symbol.usr))
```

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1702,73 +1662,112 @@ extension SourceKitServer {
...
         textDocument: req.textDocument,
         position: req.position
       )
     )
-    let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
+    guard let symbol = symbols.first else {
+      return []
+    }
+
+    if let symbol = symbols.first, let bestLocalDeclaration = symbol.bestLocalDeclaration, !(symbol.isDynamic ?? false)
```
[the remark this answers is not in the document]
ahoppen: Yes, you’re right. I added a test case for this.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: There is not. I thought we had an issue for this but looks like we don’t. Filed rdar://119163908

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Oh, it used to refer to the fact that everything was based on `symbols.first`. Rephrased the comment slightly.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Changed to `orLog` and added the following comment:

> We are on only calling the language service's `definition` function if your index-based lookup failed.  If this fallback request fails, its error is usually not very enlightening. For example the `SwiftLanguageServer` will always respond with `unsupported method`. Thus, only log such a failure instead of returning it to the client.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Just noticed that as well. Changed it slightly so that `transitiveReceiverUsrs` is `nil` if not receivers were passed.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #980, opened 2023-11-26
Decode RegistrationOptions

krzyzanowskim:
Decode `RegistrationOptions` from `LSPAny`


Sources/LanguageServerProtocol/SupportTypes/SKCompletionOptions.swift
```diff
@@ -15,10 +15,15 @@
...
+  /// Whether to use server-side filtering or to return all results and let the
+  /// client handle all filtering.
+  public var serverSideFiltering: Bool
+
   /// The maximum number of completion results to return, or `nil` for unlimited.
   public var maxResults: Int?
 
-  public init(maxResults: Int? = 200) {
+  public init(serverSideFiltering: Bool = true, maxResults: Int? = 200) {
+    self.serverSideFiltering = serverSideFiltering
```
ahoppen, line 26: I removed the `serverSideFiltering` option a few weeks ago. Is it intentional that you add it back?

ahoppen (commented):
What’s the use case for decoding `RegistrationOptions` from `LSPAny`?

ahoppen (commented):
Ideally, we should make these types conform to `LSPAnyCodable`.

krzyzanowskim:
> Ideally, we should make these types conform to LSPAnyCodable.

You're not wrong. It's so much more work. I'll try

ahoppen:
Thank you 🙏🏽 I hope it’s not too much more

krzyzanowskim:
The `LSPAnyCodable` is here https://github.com/apple/sourcekit-lsp/pull/980/commits/d718f46776452a108e9ae60c25f626b206d6a876

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -37,9 +28,20 @@ public struct TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
...
-    guard let documentSelector = documentSelector else { return }
-    dict["documentSelector"] = documentSelector.encodeToLSPAny()
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    var documentSelector: DocumentSelector?
+
+    if let value = dictionary["documentSelector"] {
+      documentSelector = DocumentSelector(fromLSPArray: value)
+    }
+
+    self.documentSelector = documentSelector
```
ahoppen, line 38: I think this can be slightly simplified.

```suggestion
    if let value = dictionary["documentSelector"] {
      self.documentSelector = DocumentSelector(fromLSPArray: value)
    } else {
      self.documentSelector = nil
    }
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -459,6 +477,54 @@ public struct CompletionOptions: WorkDoneProgressOptions, Codable, Hashable {
...
+      dict["allCommitCharacters"] = allCommitCharacters.encodeToLSPAny()
+    }
+
+    if let completionItem {
+      dict["completionItem"] = completionItem.encodeToLSPAny()
+    }
+
+    if let workDoneProgress {
+      dict["workDoneProgress"] = .bool(workDoneProgress)
+    }
```
ahoppen, line 524: Just for my understanding: We didn’t use to include `completionItem` and `workDoneProgress` in the `LSPAny` encoded value, right? I think including it is a good but want to make sure it’s a deliberate choice.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -813,6 +903,77 @@ public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable
...
+
+    if let range {
+      let encodedRange: LSPAny
+      switch range {
+      case .bool(let value):
+        encodedRange = .bool(value)
+      case .value(let rangeOptions):
+        encodedRange = rangeOptions.encodeToLSPAny()
+      }
+      dict["range"] = encodedRange
```
ahoppen, line 957: We could use a switch expression here, I think

```suggestion
      dict["range"] =  switch range {
      case .bool(let value):
        .bool(value)
      case .value(let rangeOptions):
        rangeOptions.encodeToLSPAny()
      }
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -773,21 +839,45 @@ public struct DocumentLinkOptions: WorkDoneProgressOptions, Codable, Hashable {
...
+public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable, LSPAnyCodable {
 
-  public struct SemanticTokensRangeOptions: Equatable, Hashable, Codable {
+  public struct SemanticTokensRangeOptions: Equatable, Hashable, Codable, LSPAnyCodable {
     public init() {
       // Empty in the LSP 3.16 spec.
     }
+
+    public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+      nil
```
ahoppen, line 850: Shouldn’t this return an empty `SemanticTokensRangeOptions` so that encoding and decoding from `LSPAny` round-trips?

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -813,6 +903,77 @@ public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable
...
+    }
+
+    if let full {
+      let encodedFull: LSPAny
+      switch full {
+      case .bool(let value): encodedFull = .bool(value)
+      case .value(let fullOptions):
+        encodedFull = fullOptions.encodeToLSPAny()
+      }
+      dict["full"] = encodedFull
```
ahoppen, line 967: Similar here, could be a `switch` expression

```suggestion
      dict["full"] = switch full {
      case .bool(let value): .bool(value)
      case .value(let fullOptions): fullOptions.encodeToLSPAny()
      }
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -813,6 +903,77 @@ public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable
...
+    }
+
+    if let workDoneProgress {
+      dict["workDoneProgress"] = .bool(workDoneProgress)
+    }
```
ahoppen, line 972: Just checking: We didn’t use to include this, right. And thanks for adding it.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -773,21 +839,45 @@ public struct DocumentLinkOptions: WorkDoneProgressOptions, Codable, Hashable {
...
+      self.delta = nil
+
+      if case .bool(let value) = dictionary["delta"] {
+        self.delta = value
+      }
+    }
+
+    public func encodeToLSPAny() -> LSPAny {
+      guard let delta else {
+        return .null
```
ahoppen, line 876: Shouldn’t we return an empty dictionary here instead of `null`? Mostly so that encoding + decoding to `LSPAny` round-trips.

Sources/LanguageServerProtocol/SupportTypes/RegistrationOptions.swift
```diff
@@ -37,9 +28,20 @@ public struct TextDocumentRegistrationOptions: RegistrationOptions, Hashable {
...
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    guard let documentSelector = documentSelector else { return .null }
```
ahoppen, line 42: Should we return an empty dictionary here so that encoding + decoding `TextDocumentRegistrationOptions` round-trips?

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -429,9 +429,27 @@ public struct CompletionItemOptions: Codable, Hashable {
   ) {
     self.labelDetailsSupport = labelDetailsSupport
   }
+
+  public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+    guard case .bool(let labelDetailsSupport) = dictionary["labelDetailsSupport"] else {
+      return nil
```
ahoppen, line 435: I think we should just set `labelDetailsSupport` to `nil` here instead of returning `nil` so that encoding + decoding round-trips.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -813,6 +903,77 @@ public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable
...
+
+    if case .bool(let value) = dictionary["range"] {
+      self.range = .bool(value)
+    }
+
+    if case .dictionary(let dict) = dictionary["range"],
+       let value = SemanticTokensRangeOptions(fromLSPDictionary: dict)
+    {
+      self.range = .value(value)
+    }
```
ahoppen, line 926: What about changing this to a `switch` so that it’s clear that `.bool` and `.dictionary` are distinct branches?

```suggestion
    switch dictionary["range"] {
    case .bool(let value):
      self.range = .bool(value)
    case .dictionary(let dict):
      self.range = .value(SemanticTokensRangeOptions(fromLSPDictionary: dict))
    default:
      self.range = nil
    }
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -813,6 +903,77 @@ public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable
...
+    self.full = nil
+    if case .bool(let value) = dictionary["full"] {
+      self.full = .bool(value)
+    }
+
+    if case .dictionary(let dict) = dictionary["range"],
+       let value = SemanticTokensFullOptions(fromLSPDictionary: dict)
+    {
+      self.full = .value(value)
+    }
```
ahoppen, line 937: Similar here, I would make this a `switch`.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -904,18 +1065,48 @@ public struct DiagnosticOptions: WorkDoneProgressOptions, Codable, Hashable {
...
+
+    self.interFileDependencies = false
+    if case .bool(let value) = dictionary["interFileDependencies"] {
+      self.interFileDependencies = value
+    }
+
+    self.workspaceDiagnostics = false
+    if case .bool(let value) = dictionary["workspaceDiagnostics"] {
+      self.workspaceDiagnostics = value
+    }
```
ahoppen, line 1081: Since `interFileDependencies` and `workspaceDiagnostics` are required fields, I think we should return `nil` if they aren’t present. This is what you do for the required fields in the `init(fromLSPDictionary:)` for the other structs as well.

ahoppen (commented):
Thank you for making the types conform to `LSPAnyCodable` 🙏🏽 Much cleaner. 

I have some comments regarding consistency of de/encoding.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: yes, I added it

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: you're so fancy with this new swift :)

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: yes. you're right.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: ups. that does not compile.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: hm... maybe you're right.
```swift
self = SemanticTokensRangeOptions()
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
krzyzanowskim: or to `false` 🤔  the initializer default to `false`, not `nil`

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
ahoppen: This is the project where we can actually use all of it because we don’t need to be compatible with older Swift versions ✨

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
```diff
@@ -773,21 +837,45 @@ public struct DocumentLinkOptions: WorkDoneProgressOptions, Codable, Hashable {
...
+public struct SemanticTokensOptions: WorkDoneProgressOptions, Codable, Hashable, LSPAnyCodable {
 
-  public struct SemanticTokensRangeOptions: Equatable, Hashable, Codable {
+  public struct SemanticTokensRangeOptions: Equatable, Hashable, Codable, LSPAnyCodable {
     public init() {
       // Empty in the LSP 3.16 spec.
     }
+
+    public init?(fromLSPDictionary dictionary: [String : LSPAny]) {
+      self = SemanticTokensRangeOptions()
```
ahoppen, line 848: Nit

```suggestion
      self.init()
```

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
ahoppen: Oh, interesting that the initializer defaults to `false`. At least for decoding I think `nil` is the correct value though.

Sources/LanguageServerProtocol/SupportTypes/ServerCapabilities.swift
ahoppen: Oh, I typed it in GitHub so I forgot that `SemanticTokensRangeOptions(fromLSPDictionary:)` returned an optional.

ahoppen (approved):
Thank you! I have one minor comment left, otherwise LGTM.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

krzyzanowskim:
I don’t think I have the power to re-run macOS test. It failed without a reason @ahoppen

ahoppen:
Ah, macOS CI failed because the code isn’t formatted correctly. Could you run swift-format on it? 

Sorry, we don’t have any documentation for that at the moment, I usually build [swift-format](http://github.com/apple/swift-format) locally and then run

```bash
cd /path/to/sourcekit-lsp
/path/to/swift-format/.build/arm64-apple-macosx/release/swift-format format -i --recursive Sources/ Tests/ Package.swift
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #981, opened 2023-11-27
Never return error for diagnostics request

ahoppen:
VS Code does not request diagnostics again for a document if the diagnostics request failed. Since sourcekit-lsp usually recovers from failures (e.g. after sourcekitd crashes), this is undesirable. Instead of returning an error, return empty results.

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -1062,7 +1062,20 @@ extension SwiftLanguageServer {
   }
 
   public func documentDiagnostic(_ req: DocumentDiagnosticsRequest) async throws -> DocumentDiagnosticReport {
-    return try await .full(fullDocumentDiagnosticReport(req))
+    do {
+      return try await .full(fullDocumentDiagnosticReport(req))
+    } catch {
+      // VS Code does not request diagnostics again for a document if the diagnostics request failed.
+      // Since sourcekit-lsp usually recovers from failures (e.g. after sourcekitd crashes), this is undesirable.
+      // Instead of returning an error, return empty results.
+      logger.error(
+        """
+        Loading diagnostic failed with the following error. Returning empty diagnostics. 
+        \(error.forLogging)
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -1062,7 +1062,20 @@ extension SwiftLanguageServer {
   }
 
   public func documentDiagnostic(_ req: DocumentDiagnosticsRequest) async throws -> DocumentDiagnosticReport {
-    return try await .full(fullDocumentDiagnosticReport(req))
+    do {
+      return try await .full(fullDocumentDiagnosticReport(req))
+    } catch {
+      // VS Code does not request diagnostics again for a document if the diagnostics request failed. 
+      // Since sourcekit-lsp usually recovers from failures (e.g. after sourcekitd crashes), this is undesirable. 
+      // Instead of returning an error, return empty results.
```
bnbarham, line 1070: Is this a VS Code thing or an extension thing? Because IMO it's useful to distinguish between having results and not (consider eg. deciding whether to replace previous issues or not).

ahoppen: AFAIK it’s a VS Code thing. @adam-fowler to confirm.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
adam-fowler: This is a VSCode thing. The extension doesn't have control over when VSCode asks for diagnostics. 

Out of interest the "failed" diagnostics request, does that mean no response was returned or that an error response was returned.

ahoppen: By “failed” I meant an error response.

PR #983, opened 2023-11-29
Add semantic highlighting for function parameter labels

ahoppen:
This highlights `arg` in the following as a `parameter`. 

```swift
func foo(arg: Int) {}
foo(arg: 1)
```

LSP doesn’t explicitly state what `parameter` means. `clangd` uses `parameter` for parameter declarations, Python also uses `parameter` for parameter labels.

Here’s how syntax coloring looks with this patch in VS Code with the GitHub light theme (underlined are the changed tokens).

Screenshot 2023-11-29 at 10 16 08

Companion of https://github.com/apple/swift-syntax/pull/2375.

### DIFF
--- a/Tests/SourceKitLSPTests/SemanticTokensTests.swift
+++ b/Tests/SourceKitLSPTests/SemanticTokensTests.swift
@@ -430,7 +430,7 @@ final class SemanticTokensTests: XCTestCase {
       [
         Token(line: 0, utf16index: 0, length: 4, kind: .keyword),
         Token(line: 0, utf16index: 5, length: 1, kind: .identifier),
-        Token(line: 0, utf16index: 7, length: 1, kind: .identifier),
+        Token(line: 0, utf16index: 7, length: 1, kind: .function),
         Token(line: 0, utf16index: 10, length: 3, kind: .struct, modifiers: .defaultLibrary),
         Token(line: 0, utf16index: 17, length: 1, kind: .identifier),
         Token(line: 0, utf16index: 20, length: 6, kind: .struct, modifiers: .defaultLibrary),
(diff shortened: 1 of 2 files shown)


ahoppen:
Good idea with just considering the labels part of the function itself.

> Is the screenshot you included just from the syntax tree, or does it include the semantic tokens?

It includes semantic tokens. The two states of syntax highlighting that we have in LSP land are
1. Inside the editor based on TextMate grammars
2. Semantic highlighting from the LSP server overlayed on top of the TextMate grammar, which, for Swift is highlighting from swift-syntax + semantic tokens from sourcekitd.

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -833,13 +833,34 @@ final class SemanticTokensTests: XCTestCase {
...
+        Token(line: 5, utf16index: 4, length: 1, kind: .function),
         Token(line: 5, utf16index: 7, length: 7, kind: .actor),
-        Token(line: 6, utf16index: 4, length: 1, kind: .identifier),
+        Token(line: 6, utf16index: 4, length: 1, kind: .function),
         Token(line: 6, utf16index: 7, length: 8, kind: .struct),
       ]
     )
   }
+
+  func testArgumentLabels() async throws {
```
bnbarham, line 844: Do we have a first + second name test at all? Is the second name a parameter or an identifier?

ahoppen: We do have a test for it in swift-syntax: https://github.com/apple/swift-syntax/pull/2375/files#diff-0f770f38aaec515fed644ee83824ed0dcfb2e86ec67a6da535f2de9fef15f9f3R531

ahoppen:
https://github.com/apple/swift-syntax/pull/2375

@swift-ci Please test

Tests/SourceKitLSPTests/SemanticTokensTests.swift
bnbarham: Fair enough, I'd expect one here too since the sourcekitd response could add more information if it wanted to (eg. all params + references could be "parameters").

ahoppen: Added a test case

ahoppen:
https://github.com/apple/swift-syntax/pull/2375

@swift-ci Please test Windows

PR #974, opened 2023-11-23
Add support for different arguments per workspace

ahoppen:
Fixes #663
rdar://101815704

@adam-fowler Do you think these new options on `WorkspaceFolder` would work for you? You should be able to specify them on both the `initialize` and the `workspace/didChangeWorkspaceFolders` request.

PR #988, opened 2023-12-06
Allow configuring workspace type via flag

### LINKED ISSUE #984: Option to prefer compilation database build system
We are migrating to a CMake build, but need to continue maintaining our SPM build in parallel for some period of time. One of the blockers for us to support lsp for this project is that sourcekit-lsp will always pick up a `Package.swift` file if it's there, ignoring the `compile_commands.json`.

https://github.com/apple/sourcekit-lsp/blob/49d4db32d3122191053b0c0eab12ffe14cfb5fde/Sources/SourceKitLSP/Workspace.swift#L114-L126

I think there's been some offline discussion about adding a command line flag to support this use-case. Assuming this is the right solution, I'm happy to make a PR to add this. I know less about the `BuildServerBuildSystem`; should there be a flag to prefer each of the defined build systems?

z2oh:
Fixes #984

I opted to log an error and continue with no build system set if the selected workspace type fails to create. A reasonable alternative might be to fall back to default selection logic, but this might make it harder to tell if something is wrong.

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/WorkspaceFolder.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/WorkspaceFolder.swift
@@ -18,12 +18,22 @@ public enum BuildConfiguration: Hashable, Codable {
   case release
 }
 
+/// The type of workspace; default workspace type selection logic can be overridden.
+///
+/// **(LSP Extension)**
+public enum WorkspaceType: Hashable, Codable {
+  case buildserver, compdb, swiftpm
+}
+
 /// Build settings that should be used for a workspace.
 ///
 /// **(LSP Extension)**
 public struct WorkspaceBuildSetup: Hashable, Codable {
   /// The configuration that the workspace should be built in.
   public let buildConfiguration: BuildConfiguration?
+
+  /// The default workspace type to use for this workspace.
+  public let defaultWorkspaceType: WorkspaceType?
 
   /// The build directory for the workspace.
   public let scratchPath: DocumentURI?
@@ -42,6 +52,7 @@ public struct WorkspaceBuildSetup: Hashable, Codable {
 
   public init(
     buildConfiguration: BuildConfiguration? = nil,
+    defaultWorkspaceType: WorkspaceType? = nil,
     scratchPath: DocumentURI? = nil,
     cFlags: [String]? = nil,
     cxxFlags: [String]? = nil,
@@ -49,6 +60,7 @@ public struct WorkspaceBuildSetup: Hashable, Codable {
     swiftFlags: [String]? = nil
   ) {
     self.buildConfiguration = buildConfiguration
+    self.defaultWorkspaceType = defaultWorkspaceType
     self.scratchPath = scratchPath
     self.cFlags = cFlags
     self.cxxFlags = cxxFlags
--- a/Sources/LanguageServerProtocol/SupportTypes/WorkspaceFolder.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/WorkspaceFolder.swift
@@ -22,7 +22,7 @@ public enum BuildConfiguration: Hashable, Codable {
 ///
 /// **(LSP Extension)**
 public enum WorkspaceType: Hashable, Codable {
-  case buildserver, compdb, swiftpm
+  case buildServer, compilationDatabase, swiftPM
 }
 
 /// Build settings that should be used for a workspace.
--- a/Sources/SKSupport/WorkspaceType.swift
+++ b/Sources/SKSupport/WorkspaceType.swift
@@ -1,0 +1,17 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public enum WorkspaceType: String {
+  case buildserver
+  case compdb
+  case swiftpm
+}
--- a/Sources/SKSupport/WorkspaceType.swift
+++ b/Sources/SKSupport/WorkspaceType.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 public enum WorkspaceType: String {
-  case buildserver
-  case compdb
-  case swiftpm
+  case buildServer
+  case compilationDatabase
+  case swiftPM
 }
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1015,6 +1015,7 @@ extension SourceKitServer {
       capabilityRegistry: capabilityRegistry,
       toolchainRegistry: self.toolchainRegistry,
       buildSetup: self.options.buildSetup.merging(workspaceBuildSetup),
+      workspaceType: self.options.workspaceType,
       compilationDatabaseSearchPaths: self.options.compilationDatabaseSearchPaths,
       indexOptions: self.options.indexOptions,
       reloadPackageStatusCallback: { status in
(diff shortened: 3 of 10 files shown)


Sources/SourceKitLSP/Workspace.swift
```diff
@@ -103,14 +107,53 @@ public final class Workspace {
...
+              buildSystem = compdb
+            } else {
+              logger.error(
+                "Could not set up a compilation database build system for workspace at '\(rootUri.forLogging)'"
+              )
+              buildSystem = nil
+            }
+        }
+      } else if let buildServer = await BuildServerBuildSystem(projectRoot: rootPath, buildSetup: buildSetup) {
+          buildSystem = buildServer
```
ahoppen, line 156: I would prefer not to duplicate this code. It is bound to happen that we update one of these and then forgot to update the copy as well.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -103,14 +107,53 @@ public final class Workspace {
...
+            ) {
+              buildSystem = compdb
+            } else {
+              logger.error(
+                "Could not set up a compilation database build system for workspace at '\(rootUri.forLogging)'"
+              )
+              buildSystem = nil
+            }
+        }
+      } else if let buildServer = await BuildServerBuildSystem(projectRoot: rootPath, buildSetup: buildSetup) {
```
ahoppen, line 155: If we can’t create a build server from the specified workspace type, I think we should still fall back to all the other selections. That way you can e.g. specify your default workspace type to be a compilation database in the VS Code settings but if you open a SwiftPM package, it would still work.

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -169,6 +175,12 @@ struct SourceKitLSP: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    name: .customLong("workspace-type"),
+    help: "Override default workspace type selection; one of 'swiftpm', 'compdb', or 'buildserver'"
+  )
+  var workspaceType: WorkspaceType?
```
ahoppen, line 182: I would prefer to name this `defaultWorkspaceType`, related to my comment that we should still fall back to other workspace types.

ahoppen (commented):
Thank you @z2oh! A couple comments inline. I think it would also be good to support configuration of workspace types on a per-workspace basis, similar to how you can specify compiler arguments per workspace since https://github.com/apple/sourcekit-lsp/pull/974.

z2oh:
Thank you for the quick review @ahoppen! I've made local changes per your inline suggestions. I'm slightly unclear on the correct path to supporting per-workspace configuration. I've added the relevant config to `WorkspaceFolder`, which I believe is created by the VSCode extension here: https://github.com/swift-server/vscode-swift/blob/4822e95cde97928a176f05a08e4e7ff581297ee7/src/sourcekit-lsp/LanguageClientManager.ts#L430C94-L430C94 . I don't see any existing way for user-configuration to be specified on a per-workspace basis in the extension, but it's very possible I've overlooked something. Am I on the right track here?

z2oh:
Thanks for the quick review! I've made local changes per your inline suggestions. To support per-workspace configuration, am I correct in understanding this would be a configuration setting in the Swift VSCode extension? (Most likely extending this interface https://github.com/swift-server/vscode-swift/blob/4822e95cde97928a176f05a08e4e7ff581297ee7/src/configuration.ts#L45 )

ahoppen:
> To support per-workspace configuration, am I correct in understanding this would be a configuration setting in the Swift VSCode extension?

The goal here is to expose these configuration options from sourcekit-lsp. Adding configuration to the VS Code Swift extension would be a second step. AFAIK the VS Code Swift extension doesn’t have the capability to set any of the workspace-specific parameter offered by https://github.com/apple/sourcekit-lsp/pull/974 yet but it’s something that we want to add in the future. And since the release cycles of sourcekit-lsp are longer than those of the VS Code extension (since it’s tied to the Swift releases), it’s good to design for these configuration options.

Sources/SKSupport/WorkspaceType.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public enum WorkspaceType: String {
+  case buildserver
+  case compdb
+  case swiftpm
```
ahoppen, line 16: A little nitpicking

```suggestion
  case buildServer
  case compilationDatabase
  case swiftPM
```

Sources/LanguageServerProtocol/SupportTypes/WorkspaceFolder.swift
```diff
@@ -18,13 +18,23 @@ public enum BuildConfiguration: Hashable, Codable {
   case release
 }
 
+/// The type of workspace; default workspace type selection logic can be overridden.
+///
+/// **(LSP Extension)**
+public enum WorkspaceType: Hashable, Codable {
+  case buildserver, compdb, swiftpm
```
ahoppen, line 25: Nitpick, let’s go for more descriptive names.
```suggestion
  case buildServer, compilationDatabase, swiftPM
```

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -974,6 +974,17 @@ extension LanguageServerProtocol.BuildConfiguration {
   }
 }
 
+extension LanguageServerProtocol.WorkspaceType {
```
ahoppen, line 977: Let’s keep this localized and make it `private`.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -108,20 +108,40 @@ public final class Workspace {
...
+      return await BuildServerBuildSystem(projectRoot: rootPath, buildSetup: buildSetup)
+    }
+
+    if let rootUrl = rootUri.fileURL, let rootPath = try? AbsolutePath(validating: rootUrl.path) {
+      if let defaultWorkspaceType = buildSetup.defaultWorkspaceType, let defaultBuildSystem: BuildSystem? =
+        switch defaultWorkspaceType {
+          case .buildserver: await buildBuildServerWorkspace(rootPath: rootPath)
+          case .compdb: buildCompDBWorkspace(rootPath: rootPath)
+          case .swiftpm: await buildSwiftPMWorkspace(rootUrl: rootUrl)
+      } {
```
ahoppen, line 138: I’m not a fan of nesting a `switch` expression inside an `if` condition. I think it hurts readability. I would prefer something like

```suggestion
      let defaultBuildSystem: BuildSystem? =
        switch defaultWorkspaceType {
          case .buildserver: await buildBuildServerWorkspace(rootPath: rootPath)
          case .compdb: buildCompDBWorkspace(rootPath: rootPath)
          case .swiftpm: await buildSwiftPMWorkspace(rootUrl: rootUrl)
          case nil: nil
      }
      if let defaultBuildSystem  {
```

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -169,6 +175,12 @@ struct SourceKitLSP: ParsableCommand {
   )
   var indexPrefixMappings = [PathPrefixMapping]()
 
+  @Option(
+    name: .customLong("default-workspace-type"),
```
ahoppen, line 179: Do we need to specify a `customLong` here? Shouldn’t it infer `default-workspace-type` based on the variable?

Sources/SourceKitLSP/Workspace.swift
z2oh: Much better!

ahoppen (approved):
Thank you 🙏🏽

ahoppen:
@swift-ci Please test

z2oh:
Okay, I think I fixed the tests. I'm developing on Windows with limited testing support, but I was able to get things set up in Windows subsystem for Linux so that I could compile the tests. Thanks for your patience 🙏

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #989, opened 2023-12-06
Don’t use the best local declaration when jumping to definition of a C function

ahoppen:
For non-Swift symbols, we need to perform an index lookup because the best local declaration will point to a header file but jump-to-definition should prefer the implementation (there's the declaration request to jump to the function's declaration).

rdar://119243893


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #991, opened 2023-12-08
Execute sourcekit-lsp tests in parallel and re-run them serially only if a test fails

ahoppen:
This speeds up testing time for successful runs and minimizes log output if all tests succeed while still producing actionable output on failure.


ahoppen:
@swift-ci Please test

bnbarham (approved):
Interesting idea!

PR #996, opened 2023-12-13
Add an XCTest that performs the same testing logic as `test-sourcekit-lsp.py` in `swift-integration-tests`

ahoppen:
This has two advantages: (1) We catch failures in this test setup more quickly and (2) it’s easier to debug because you can just attach a debugger to the test execution.

The only difference at the moment is that this test case uses a tools version of 5.5 while the integration test uses 5.1 because 5.1 since is causing semantic functionality to fail. That’s because we hit https://github.com/apple/swift-package-manager/blob/64b7ca9465242b256fbfab912e03e53d31b5bf39/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift#L115-L116 and thus `moduleOutputPath` is `tempsPath` instead of `modulesPath` and thus the include path computation introduced by https://github.com/apple/sourcekit-lsp/pull/987 is incorrect. @neonichu is looking into that.

### DIFF
--- a/Tests/SourceKitLSPTests/WorkspaceTests.swift
+++ b/Tests/SourceKitLSPTests/WorkspaceTests.swift
@@ -513,4 +513,146 @@ final class WorkspaceTests: XCTestCase {
     let diag = try XCTUnwrap(fullReport.items.first)
     XCTAssertEqual(diag.message, "Cannot convert value of type 'String' to specified type 'Int'")
   }
+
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
Very cool 👍 Is the idea that we only have this and not the integration test? Mostly wondering about keeping the two in-sync

PR #1000, opened 2023-12-14
Wrap request and notification ID for the purpose of logging scopes

ahoppen:
OSLog only allows the creation of 4000 logger objects, which means that logging will crash on the 4000th request to sourcekit-lsp (which would create the 4000th os_log object because it is the 4000th distinct category in which we are logging).

To circumvent this problem, only use the last two digits of the request and notification ID for the logging scope and wrap around afterwards. This should still allow us to tell apart log messages from concurrently handled requests/notifications while only using 200 os_log objects, which means we stay well clear of the 4000 limit.

rdar://119221355

### DIFF
--- a/Sources/LSPLogging/LoggingScope.swift
+++ b/Sources/LSPLogging/LoggingScope.swift
@@ -22,36 +22,20 @@ public final class LoggingScope {
   }
 }
 
-/// Append `subscope` to the current scope name, if any exists, otherwise return
-/// `subscope`.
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1001, opened 2023-12-14
Add a tasks.json file with a default build configuration and commands to run tests in parallel

ahoppen:


.vscode/tasks.json
```diff
@@ -0,0 +1,61 @@
+{
+	"tasks": [
+		{
+			"args": [
+				"build",
+				"--build-tests"
+			],
+			"cwd": "",
+			"detail": "swift build --build-tests",
+			"disableTaskQueue": true,
```
bnbarham, line 10: Is there more documentation on this than https://code.visualstudio.com/docs/editor/tasks-appendix? I don't see `detail` or `disableTaskQueue` in there and `type` seems to only be `process` or `shell`.

adam-fowler: Each workspace folder has a task queue which tasks like resolve, update, list tests are queued on. `disableTaskQueue` means while running this build task do not run any tasks that are on the task queue. It is there to stop multiple swiftPM processes running at the same time.

Given the extension already provides build tasks what it the reason for having a default build task here.

adam-fowler: How do you run a release build? Also on Linux and Windows it doesn't include the required parameters

ahoppen: My thought was that you go through the *Configure Default Build Task* workflow for that. 

That being said, maybe adding a build task is not the right approach here, all I wanted to achieve was that I could trigger a build with a single keystroke and there might be better ways to do so.

ahoppen:
I downscoped the PR to only add tasks to run tests until we figure out how to correctly write the default build command.

ahoppen:
@swift-ci Please test

PR #1006, opened 2023-12-20
Support the prepare rename request

ahoppen:
This allows us to return the current compound decl name when renaming a function, which will get populated as the default value for the text field in which the user enters the new symbol name.

rdar://118995649


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1007, opened 2023-12-23
Add a `concurrentMap` function to `Collection`

ahoppen:
I find this a lot more intuitive to use than `TaskGroup`.

### DIFF
--- a/Sources/SKSupport/AsyncUtils.swift
+++ b/Sources/SKSupport/AsyncUtils.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
+
 public extension Task {
   /// Awaits the value of the result.
   ///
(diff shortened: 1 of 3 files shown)


Sources/SKSupport/AsyncUtils.swift
```diff
@@ -65,3 +65,27 @@ public func withCancellableCheckedThrowingContinuation<Handle, Result>(
...
 }
+
+extension Collection {
+  /// Transforms all elements in the collection concurrently and returns the transformed collection.
+  public func concurrentMap<TransformedElement>(
+    _ transform: @escaping (
+      Element
+    ) async -> TransformedElement
+  ) async -> [TransformedElement] {
+    return await withTaskGroup(of: (Int, TransformedElement).self) { taskGroup in
```
rauhul, line 76: I'd recommend limiting this to some bounded number of tasks, otherwise your tasks may start thrashing the scheduler

adam-fowler: Yeah have you considered doing something like this instead. 
https://github.com/adam-fowler/async-collections/blob/cf064405329667bbeb66d55563bbf14048faed2c/Sources/AsyncCollections/Map.swift#L90

It adds a limit to the number of tasks that can be spawn and removes the requirement for `sort`

ahoppen: Good idea 👍🏽

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1008, opened 2023-12-23
Support creation of SKDRequest(Dictionary|Array) from literals

ahoppen:
IMO this makes it a lot clearer which keys are present in the request dictionaries because we are no longer mutating them on the fly.

### DIFF
--- a/Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
+++ b/Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
@@ -191,16 +191,15 @@ class CodeCompletionSession {
       throw ResponseError(code: .invalidRequest, message: "open must use the original snapshot")
     }
 
-    let req = SKDRequestDictionary(sourcekitd: sourcekitd)
-    req[keys.request] = sourcekitd.requests.codecomplete_open
-    req[keys.offset] = utf8StartOffset
-    req[keys.name] = uri.pseudoPath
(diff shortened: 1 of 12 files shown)


Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -191,16 +191,15 @@ class CodeCompletionSession {
...
-    req[keys.request] = sourcekitd.requests.codecomplete_open
-    req[keys.offset] = utf8StartOffset
-    req[keys.name] = uri.pseudoPath
-    req[keys.sourcefile] = uri.pseudoPath
-    req[keys.sourcetext] = snapshot.text
-    req[keys.codecomplete_options] = optionsDictionary(filterText: filterText, options: options)
-    if let compileCommand = compileCommand {
-      req[keys.compilerargs] = compileCommand.compilerArgs
-    }
+    let req = [
```
bnbarham, line 194: Not a huge fan of `skd`. What if SKArray/Dictionary both had an initializer that took a sequence/dictionary?

ahoppen: Changed it to `sourcekitd.dictionary(<#dict literal#>)`

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1011, opened 2023-12-25
Fix decoding Array of LSPAnyCodable

krzyzanowskim:
Fixes decoding array of LSPAnyCodable that wasn't handled properly and decoded only dictionary values

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
@@ -152,12 +152,38 @@ extension Array: LSPAnyCodable where Element: LSPAnyCodable {
     guard case .array(let array) = array else {
       return nil
     }
+
     var result = [Element]()
-    for case .dictionary(let editDict) in array {
-      guard let element = Element.init(fromLSPDictionary: editDict) else {
-        return nil
+    for element in array {
+      switch element {
+      case .dictionary(let dict):
+        if let value = Element(fromLSPDictionary: dict) {
+          result.append(value)
+        }
+      case .array(let value):
+        if let value = value as? [Element] {
+          result.append(contentsOf: value)
+        }
+      case .string(let value):
+        if let value = value as? Element {
+          result.append(value)
+        }
+      case .int(let value):
+        if let value = value as? Element {
+          result.append(value)
+        }
+      case .double(let value):
+        if let value = value as? Element {
+          result.append(value)
+        }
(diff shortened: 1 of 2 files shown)


compnerd (commented):
Could you add a test case for this please?

krzyzanowskim:
is `LSPAnyCodable` tested at all? can't find

krzyzanowskim:
Can we get this fixed?

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
```diff
@@ -152,12 +152,38 @@ extension Array: LSPAnyCodable where Element: LSPAnyCodable {
...
+          result.append(value)
+        }
+      case .double(let value):
+        if let value = value as? Element {
+          result.append(value)
+        }
+      case .bool(let value):
+        if let value = value as? Element {
+          result.append(value)
+        }
```
ahoppen, line 182: Should we return `nil` if one of the elements can’t get decoded to indicate failure instead of silently dropping it?

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
```diff
@@ -152,12 +152,38 @@ extension Array: LSPAnyCodable where Element: LSPAnyCodable {
...
+      case .null:
+        // skip array of null
+        break
```
ahoppen, line 185: Shouldn’t we append `nil` to `result` if `Element` is `Optional`?

krzyzanowskim: the `Element: LSPAnyCodable` here is not optional

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
krzyzanowskim: right, probably we should. Changed

Sources/LanguageServerProtocol/SupportTypes/LSPAny.swift
ahoppen: Oh, stupid me. 🤦🏽

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1015, opened 2024-01-09
Ensure rename tests don’t fail if running with a sourcekitd that doesn’t support rename yet

ahoppen:
Just another case of making sure that sourcekit-lsp’s tests don’t fail after checking out the repo and running its tests using eg. Xcode 15.1 without an open source toolchain snapshot.


ahoppen:
@swift-ci Please test

PR #1017, opened 2024-01-09
Report secondary cursor info results

ahoppen:
rdar://119163908

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1706,49 +1706,24 @@ extension SourceKitServer {
     return try await languageService.declaration(req)
   }
 
-  /// Returns the result of a `DefinitionRequest` by running a `SymbolInfoRequest`, inspecting
-  /// its result and doing index lookups, if necessary.
-  ///
-  /// In contrast to `definition`, this does not fall back to sending a `DefinitionRequest` to the
-  /// toolchain language server.
-  private func indexBasedDefinition(
-    _ req: DefinitionRequest,
-    workspace: Workspace,
+  /// Return the locations for jump to definition from the given `SymbolDetails`.
+  private func definitionLocations(
+    for symbol: SymbolDetails,
+    in uri: DocumentURI,
     languageService: ToolchainLanguageServer
   ) async throws -> [Location] {
-    let symbols = try await languageService.symbolInfo(
-      SymbolInfoRequest(
-        textDocument: req.textDocument,
-        position: req.position
-      )
-    )
-    guard let symbol = symbols.first else {
-      return []
-    }
-
-    if let bestLocalDeclaration = symbol.bestLocalDeclaration,
-      !(symbol.isDynamic ?? true),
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1797,16 +1772,48 @@ extension SourceKitServer {
...
+    }) {
+      // If we have a known non-dynamic symbol within Swift, we don't need to do an index lookup.
+      // For non-Swift symbols, we need to perform an index lookup because the best local declaration will point to
+      // a header file but jump-to-definition should prefer the implementation (there's the declaration request to jump
+      // to the function's declaration).
+      return symbols.compactMap { $0.bestLocalDeclaration }
+    }
+
+    return try await symbols.asyncMap {
+      try await self.definitionLocations(for: $0, in: req.textDocument.uri, languageService: languageService)
```
bnbarham, line 1813: Rather than do the `allSatisfy` above and then mapping all here, could we just map each as required? Eg. there's no reason to run `definitionLocations` on non-dynamic Swift symbols.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -509,35 +509,42 @@ extension SwiftLanguageServer {
   public func hover(_ req: HoverRequest) async throws -> HoverResponse? {
     let uri = req.textDocument.uri
     let position = req.position
-    guard let cursorInfo = try await cursorInfo(uri, position..<position) else {
-      return nil
-    }
+    // LSP's hover can't display multiple results. Don't display any results if the cursor info response is
```
bnbarham, line 512: Did you change your mind here and concat instead of failing if there were multiple 😅?

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Good idea 👍🏽

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Yes. I think only showing the primary result when it could be ambiguous can be actively misleading.

bnbarham: Totally fine, in case it wasn't clear my comment was about how the comment is old.

```suggestion
    // LSP's hover can't display multiple results. Concatenate any alternate results together.
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
ahoppen: Ah, I didn’t see that when going through the review.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1021, opened 2024-01-11
Guard usages of `@retroactive` by Swift 5.11, not 5.10

ahoppen:
I incorrectly thought that `@retroactive` gets introduced by Swift 5.10 but it’s only available in Swift 5.11.


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #958, opened 2023-10-31
SourceKit-LSP not reporting diagnostics


adam-fowler:
If I compile a SwiftPM project and then type code referencing a symbol that doesn't exist, or the wrong parameter to a function it will not report a diagnostic for these. eg
```swift
let s = String("test", doesNotExist: true)
```
should produce a diagnostic `extra argument 'doesNotExist' in call`. This does not happen.

This was working in Swift 5.8, it is no longer working in 5.9

ahoppen:
Tracked in Apple’s issue tracker as rdar://117732149

ahoppen:
I have been able to reproduce this once using Swift 5.9 but not using Swift 5.10 or a recent `main` snapshot. Also, when I tried reproducing it using Swift 5.9 again, I couldn’t.

Does this reproduce consistently for you? If so, does it also reproduce using a recent `main` development snapshot? If it does reproduce on `main`, could you gather logs using the following command after hitting the issue and attach the logs?

```bash
log show --last 10m --predicate 'subsystem CONTAINS "sourcekit-lsp"' > /tmp/log.txt
```

adam-fowler:
It does appear to be fixed on main. It would be helpful if we could find out why it is an issue in 5.9. This significantly damages the quality of support vscode can supply. Swift 5.10 is not going to be released for a significant amount of time.

ahoppen:
Does it reproduce reliably for you in 5.9?

adam-fowler:
Yes

ahoppen:
That’s really weird. I just tried reproducing it for quite a while now and couldn’t. 

Could you set `SOURCEKIT_LOGGING` to 99 (I always forget what the real maximum log level is) in `swift.swiftEnvironmentVariables`, reproduce the issue attach the log output from the `SourceKit Language Server` output? I hope that I can tell what’s going wrong from there.

ahoppen:
Ah, I think I have found a hypothesis (hit this on `main` just now):

In 5.9 we switched to the pull model for diagnostics (https://github.com/apple/sourcekit-lsp/pull/751) instead of always pushing diagnostics to the client. But it appears that whenever a diagnostic request fails (in the case I hit, sourcekitd crashed), VS Code won’t request diagnostics for that document again, which is surprising to me. I would have expected diagnostics to be requested again, maybe after a delay.

It would be interesting to see if that is the same issue that you are hitting, @adam-fowler. If it is, that should be obvious from the log.

[...]

PR #1022, opened 2024-01-13
[5.10] Never return error for diagnostics request 

ahoppen:
* **Explanation**: VS Code does not request diagnostics again for a document if the diagnostics request failed. Since sourcekit-lsp usually recovers from failures (e.g. after sourcekitd crashes), this is undesirable. Instead of returning an error, return empty results. This causes VS Code to re-request diagnostics after an edit is made, which will succeed after sourcekitd has been restarted.
* **Scope**: The SourceKit-LSP diagnostics request after sourcekitd crashed
* **Risk**: Very low, just returns empty results where it used to return an error
* **Testing**: Manual testing of the change in https://github.com/apple/sourcekit-lsp/pull/981. This is functionally equivalent and trivial
* **Issue**: #958 / rdar://117732149
* **Reviewer**:  @bnbarham

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
@@ -1476,9 +1476,10 @@ extension SwiftLanguageServer {
           req.reply(.full(.init(items: diagnostics)))
 
         case .failure(let error):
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #1023, opened 2024-01-14
sourcekit-lsp no longer starts because ParsableCommand is not implemented properly


benjiwolff:
In [this PR](https://github.com/apple/sourcekit-lsp/pull/1018) the signature of the SourceKitLSP command was changed from
```
func run() throws
```
to
```
func run() async throws
```

Unfortunately, this does not result in a build error. Instead, the `run` function is now simply not recognized. When starting the server from the command-line exits with code 0 and simply shows the CLI manual.

ahoppen:
Tracked in Apple’s issue tracker as rdar://120976054

ahoppen:
Thanks for the report, @benjiwolff. I’m fixing it in https://github.com/apple/sourcekit-lsp/pull/1024

PR #1024, opened 2024-01-17
Make ToolchainRegistry initializer synchronous

ahoppen:
This requires a bit of a restructuring of `ToolchainRegistry` because calling `scanForToolchain` on `self` is a suspension point. Instead, we need to implement the entire toolchain discovery in the initializer itself, which really isn’t that bad anyway.

This allows us to make the `SourceKitLSP.run` function synchronous again.

Fixes https://github.com/apple/sourcekit-lsp/issues/1023
rdar://120976054

### DIFF
--- a/Sources/SKCore/ToolchainRegistry.swift
+++ b/Sources/SKCore/ToolchainRegistry.swift
@@ -18,6 +18,7 @@ import struct TSCBasic.AbsolutePath
 import protocol TSCBasic.FileSystem
 import class TSCBasic.Process
 import enum TSCBasic.ProcessEnv
+import struct TSCBasic.ProcessEnvironmentKey
 import func TSCBasic.getEnvSearchPaths
 import var TSCBasic.localFileSystem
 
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

compnerd:

compnerd:

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):
As is I suppose this is no worse. Though ideally all the IO would be async in which case this would have to become async again. Would be nice to file an issue/radar to take a look again after swift-argument-parser is fixed (or until we bump to the version it's fixed? I didn't look closely enough)

ahoppen:
swift-argument-parser is not the only issue here. At the end of `run` we call `dispatchMain`, which parks the main thread to handle any items that get put on `DisptachQueue.main`. `dispatchMain` must be called on the main thread and calls `pthread_exit` when it finishes. Because it needs to be called on the main thread, we can’t call it from an async function and we can’t call it as `await Task { @MainActor in dispatchMain() }.value` either because Swift concurrency and `pthread_exit` don’t like each other and it results in a crash on process exit (I don’t exactly know what’s causing this issue but it makes sense to me on some level).

etcwilde:
The async-main will implicitly call `dispatchMain` or `CFRunLoopRun` on the main thread at the first suspension point. Everything up to that suspension is run synchronously. You might be able to restructure some things so that the synchronous setup work happens before that first suspension and then let the generated main implicitly call it for you?

etcwilde:
Also, did `AsyncParsableCommand` not work?

ahoppen:
That’s good to know if we introduce `await` calls into our `main` function again. For now, I think the new implementation is actually better because it properly marks the properties of `ToolchainRegistry` as constants, which they are. There wasn’t really any asynchronous stuff happening in its initializer anyway.

PR #1028, opened 2024-01-17
Add logic to split command line arguments on Windows

### LINKED ISSUE #1020: Windows path separators in `compile_commands.json` not parsed correctly
Windows path separators in `compile_commands.json` are not parsed correctly in [`splitShellEscapedCommand`](https://github.com/apple/sourcekit-lsp/blob/e8e4fb210616e8d41f5803f19dd6427fc27e247e/Sources/SKCore/CompilationDatabase.swift#L281). Following the comment there to clang's [`unescapeCommandLine()`](https://github.com/llvm/llvm-project/blob/bb8f2585c6eab263916757435d71df16d92de4a8/clang/lib/Tooling/JSONCompilationDatabase.cpp#L135), it seems clang has implemented Windows specific parsing.

There's a separate bug in Swift CMake that generates incorrect path separators on Windows, but after that is fixed I expect the entries in the compilation database to look like:
```
{
  "directory": "C:\\path\\to\\build",
  "command": "C:\\path\\to\\swiftc.exe <snip>,
  "file": "C:\\path\\to\\build\\filename.swift",
  "output": "path\\to\\filename.swift.obj"
}
```
These are single-escaped JSON backslashes, which are unescaped twice (once by JSON parsing, once in the function linked above), resulting paths like `C:pathtobuildfilename.swift` being passed to `sourcekitd`. If I manually change these paths in `compile_commands.json` to double-escaped backslashes (`\\\\`) or forward slashes, the paths are interpreted as expected.

ahoppen:
Previously, we were splitting command line arguments on Windows using the same rules as on Unix, which was incorrect, most importantly because backslashes in the first component of a Windows command line invocation are not escaping anything but interpreted verbatim.

The unix command line splitting remains completely unchanged, I just moved it to a different files. Similarly, the assertions for Unix command line splitting remain the same, I just moved the test cases and restructured them slightly.

Fixes #1020
rdar://120809063

### DIFF
--- a/Tests/SKCoreTests/SplitShellCommandTests.swift
+++ b/Tests/SKCoreTests/SplitShellCommandTests.swift
@@ -1,0 +1,119 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SKCore
+import XCTest
+
+/// Assert that splitting `str` into its command line components results in `expected`.
+///
+/// By default assert that escaping using Unix and Windows rules results in the same split. If `windows` is specified,
+/// assert that escaping with Windows rules produces `windows` and escaping using Unix rules results in `expected`.
+///
+/// If set `initialCommandName` gets passed to the Windows split function.
+func assertEscapedCommand(
+  _ str: String,
+  _ expected: [String],
+  windows: [String]? = nil,
+  initialCommandName: Bool = false,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) {
+  XCTAssertEqual(
+    splitShellEscapedCommand(str),
+    expected,
+    "Splitting Unix command line arguments",
+    file: file,
+    line: line
+  )
+  XCTAssertEqual(
+    splitWindowsCommandLine(str, initialCommandName: initialCommandName),
+    windows ?? expected,
+    "Splitting Windows command line arguments",
+    file: file,
+    line: line
+  )
+}
+
+final class SplitShellCommandTests: XCTestCase {
+  func testSplitShellEscapedCommandBasic() {
+    assertEscapedCommand("", [])
+    assertEscapedCommand("    ", [])
+    assertEscapedCommand("a", ["a"])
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

MaxDesiatov:
@swift-ci test windows

ahoppen:
@swift-ci Please test Windows

Tests/SKCoreTests/SplitShellCommandTests.swift
```diff
@@ -0,0 +1,119 @@
...
+      initialCommandName: true
+    )
+    assertEscapedCommand(
+      #"C:\ swift.exe"#,
+      [#"C: swift.exe"#],
+      windows: [#"C:\"#, #"swift.exe"#],
+      initialCommandName: false
+    )
+    assertEscapedCommand(#"C:\"swift.exe""#, [#"C:"swift.exe"#], windows: [#"C:\swift.exe"#], initialCommandName: true)
+    assertEscapedCommand(#"C:\"swift.exe""#, [#"C:"swift.exe"#], windows: [#"C:"swift.exe"#], initialCommandName: false)
```
bnbarham, line 112: Is it correct that the last quote just gets dropped? It's the start of a quoted string, so probably the most reasonable thing to do, still weird though.

Tests/SKCoreTests/SplitShellCommandTests.swift
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  func testSplitShellEscapedCommandWindowsTwoDoubleQuotes() {
+    assertEscapedCommand(#"" test with "" quote""#, [" test with  quote"], windows: [#" test with " quote"#])
+    assertEscapedCommand(#"" test with "" quote""#, [" test with  quote"], windows: [#" test with " quote"#])
```
bnbarham, line 117: Identical?

bnbarham (approved):
Going to be honest, I mostly glazed over the actually implementation (especially because most of it is red due to the syntax highlighting currently). The tests looked mostly reasonable from the comments though.

Tests/SKCoreTests/SplitShellCommandTests.swift
ahoppen: I think it’s as reasonable as anything we could do. 🤷🏽 Alternatively, we could also completely error but I’m not sure if that’s any better.

Tests/SKCoreTests/SplitShellCommandTests.swift
ahoppen: No, Unix has `with  quote`, Windows has `with " quote` (double `""` results in an escaped `"` on Windows and an empty String on Unix)

PR #1031, opened 2024-01-23
Support cross-file rename for clang languages

ahoppen:
This uses the indexed rename request I added to clangd to perform global rename in clang’s language using SourceKit-LSP’s index: SourceKit-LSP’s index is used to find the locations to rename and the indexed rename request to clangd is used to translate the rename locations to edits.

rdar://118996369

### DIFF
--- a/Sources/LanguageServerProtocol/Requests/IndexedRenameRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/IndexedRenameRequest.swift
@@ -1,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Rename all occurrences of a symbol named `oldName` to `newName` at the
+/// given `positions`.
+///
+/// The use case of this method is for when the positions to rename are already
+/// known, eg. from an index lookup outside of clangd's built-in index. In
+/// particular, it determines the edits necessary to rename multi-piece
+/// Objective-C selector names.
+///
+/// `textDocument` is used to determine the language options for the symbol to
+/// rename, eg. to decide whether `oldName` and `newName` are Objective-C
+/// selectors or normal identifiers.
+///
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/Requests/IndexedRenameRequest.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Rename all occurrences of a symbol named `oldName` to `newName` at the
+/// given `positions`.
+///
+/// The use case of this method is the positions to rename are already known
```
bnbarham, line 16:
```suggestion
/// The use case of this method is for when the positions to rename are already known
```

Looks like a copy paste of the clangd comment, so you could update it from there (assuming you fixed that).

Tests/SourceKitLSPTests/RenameTests.swift
```diff
@@ -759,4 +761,74 @@ final class RenameTests: XCTestCase {
...
+
+        @implementation Foo
+        - (int)performNewAction:(int)action by:(int)value {
+          return [self performNewAction:action by:value];
+        }
+        @end
+        """,
+      ]
+    )
+  }
```
bnbarham, line 833: Could we add a cross-language test? Or multiple even, eg. swift <-> objc, swift <-> c, and swift <-> c++?

bnbarham (approved):

Tests/SourceKitLSPTests/RenameTests.swift
ahoppen: Ha, no. This PR is about adding global rename for clang languages, not doing cross-language rename yet. For that we need name translation, which I’ll add in a follow-up PR (and which is currently blocked by rdar://120066209).

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitLSPTests/RenameTests.swift
bnbarham: IIUC only ObjC is blocked by that right 😅? But cool, didn't realize we didn't have name translation yet, thanks 👍

ahoppen: Yeah, but ObjC is really the interesting candidate for name translation. And it’s easiest to design it with all name translations in mind. There will be a follow-up PR soon-ish.

PR #1033, opened 2024-01-25
Clean up rpaths of installed sourcekit-lsp

ahoppen:
- Don’t explicitly add rpaths to `/usr/lib/swift`, `@executable_path/../lib/swift/macosx` and `@executable_path/../lib/swift-5.5/macosx` on Darwin. I don’t know what they were needed for but they don’t seem to be necessary. Also standard Swift command line tools created from Xcode don’t contain these rpaths.

rdar://121400644


ahoppen:
@swift-ci Please test

PR #1034, opened 2024-01-25
Prefer a toolchain if it is a proper superset of another toolchain

ahoppen:
This fixes the issue that you need to set the `PATH` environment variable to have `/usr/bin` as the first entry to test sourcekit-lsp on Linux.

rdar://78525669

### DIFF
--- a/Sources/SKCore/Toolchain.swift
+++ b/Sources/SKCore/Toolchain.swift
@@ -86,6 +86,31 @@ public final class Toolchain {
     self.sourcekitd = sourcekitd
     self.libIndexStore = libIndexStore
   }
+
+  /// Returns `true` if this toolchain has strictly more tools than `other`.
+  ///
+  /// ### Examples
+  /// - A toolchain that contains both `swiftc` and  `clangd` is a superset of one that only contains `swiftc`.
+  /// - A toolchain that contains only `swiftc`, `clangd` is not a superset of a toolchain that contains `swiftc` and
+  ///   `libIndexStore`. These toolchains are not comparable.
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SKCore/Toolchain.swift
```diff
@@ -86,6 +86,25 @@ public final class Toolchain {
...
+  /// - A toolchain that contains only `swiftc`, `clangd` is not a superset of a toolchain that contains `swiftc` and
+  ///   `libIndexStore`. These toolchains are not comparable.
+  func isSuperset(of other: Toolchain) -> Bool {
+    func isSuperset(for tool: KeyPath<Toolchain, AbsolutePath?>) -> Bool {
+      if self[keyPath: tool] == nil && other[keyPath: tool] != nil {
+        // This toolchain doesn't contain the tool but the other toolchain does. It is not a superset.
+        return false
+      } else {
+        return true
+      }
```
bnbarham, line 103: `return self[keyPath: tool] != nil || other[keyPath: tool] == nil`?

ahoppen: I usually prefer an `if` if I put a comment in there because that way I can describe one particular branch.

bnbarham: Up to you, in this case I think describing the `true` case makes it more clear anyway. ie. it's a superset if either this toolchain has the tool, or the other doesn't.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1043, opened 2024-02-01
[5.10] Add logic to split command line arguments on Windows

z2oh:
* **Explanation**: This change implemented Windows-specific command line argument splitting for compilation database fields; previously, the Windows path separator was being interpreted as an escape character.

* **Scope**: Without this change, sourcekit-lsp does not work with compilation databases on Windows. This change refactors command line argument splitting logic into Unix and Windows variants (the Windows variant being new code), with the callsite selecting between them via `#if os(Windows)`.

* **Issue**: https://github.com/apple/sourcekit-lsp/issues/1020

* **Risk**: Low; the change only affects Windows. As sourcekit-lsp was not working with compilation database on Windows before this change, there is no possible regression on this platform.

* **Testing**: I have built and tested the 5.10 toolchain on Windows with this cherry-pick. CI will build and test for other platforms.

* **Reviewer**: @ahoppen, who authored this change and @bnbarham on https://github.com/apple/sourcekit-lsp/pull/1028

### DIFF
--- a/Sources/SKCore/CompilationDatabase.swift
+++ b/Sources/SKCore/CompilationDatabase.swift
@@ -221,7 +221,11 @@ extension CompilationDatabase.Command: Codable {
     if let arguments = try container.decodeIfPresent([String].self, forKey: .arguments) {
       self.commandLine = arguments
     } else if let command = try container.decodeIfPresent(String.self, forKey: .command) {
+      #if os(Windows)
+      self.commandLine = splitWindowsCommandLine(command, initialCommandName: true)
+      #else
       self.commandLine = splitShellEscapedCommand(command)
+      #endif
     } else {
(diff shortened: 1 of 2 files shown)


ahoppen:
Could you reduce the change to not move `splitShellEscapedCommand` to a different file? Since we are fairly late in 5.10’s release cycle, this should be the minimal possible diff to fix the issue.

z2oh:
Good point; I moved the parsing logic back into `CompilationDatabase.swift` and folded the tests into `CompilationDatabaseTests.swift`.

ahoppen:
Updated the PR to be purely additive in a way that is guaranteed to only affect Windows.

airspeedswift (approved):

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

PR #1046, opened 2024-02-02
Add infrastructure for skipping tests if the host toolchain doesn't support them

ahoppen:
Up until now we have skipped tests on a very ad-hoc basis and some of these skips have lingered around much longer than they were needed.

Add infrastructure to check if a given host toolchain has a feature required by a test. The key ideas here are:
- The skip check specifies a Swift version in which the required feature was introduced. Any newer Swift versions are always considered to support this feature. This allows us to remove the checks once the minimum required version to test sourcekit-lsp exceeds the specified version.
- If the toolchain version is the same as the Swift version that introduces the feature, we execute a closure that performs a fine-grained check to see if the feature is indeed supported. This differentiates toolchain snapshots which contain the feature from those that don’t.
- Tests are never skipped in Swift CI. This ensures that we don’t get passing tests because we hit the skip condition even though we are not expecting to hit it in Swift CI, which should have built an up-to-date toolchain.
- Use this new infrastructure to skip tests that aren’t passing using older Swift versions.

rdar://121911039


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -0,0 +1,252 @@
...
+  /// Throw an `XCTSkip` if any of the following conditions hold
+  ///  - The Swift version of the toolchain used for testing (`ToolchainRegistry.forTesting.default`) is older than
+  ///    `swiftVersion`
+  ///  - The Swift version of the toolchain used for testing is equal to `swiftVersion` and `featureCheck` returns
+  ///    `false`. This is used for features that are introduced in `swiftVersion` but are not present in all toolchain
+  ///    snapshots.
+  ///
+  /// Having the version check indicates when the check tests can be removed (namely when the minimum required version
+  /// to test sourcekit-lsp is above `swiftVersion`) and it ensures that tests can’t stay in the skipped state over
+  /// multiple releases.
```
bnbarham, line 46: Heh, thanks for the comment. I wrote a comment asking what the version is for :).

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -0,0 +1,252 @@
...
+      }
+      return true
+    }
+  }
+
+  /// SwiftPM moved the location where it stores Swift modules to a subdirectory in
+  /// https://github.com/apple/swift-package-manager/pull/7103.
+  ///
+  /// sourcekit-lsp uses the built-in SwiftPM to synthesize compiler arguments and cross-module tests fail if the host
+  /// toolchain’s SwiftPM stores the Swift modules on the top level but we sythesize compiler arguments expecting the
```
bnbarham, line 147:
```suggestion
  /// toolchain’s SwiftPM stores the Swift modules on the top level but we synthesize compiler arguments expecting the
```

Having said that, can we not just check this? ie. check for a `Modules` folder and otherwise assume top level?

bnbarham (approved):
Nice cleanup 👍

Sources/SKTestSupport/SkipUnless.swift
ahoppen: Like, if SwiftPM is called from SourceKit-LSP, check whether the build folder looks like it was generated by a previous SwiftPM version and if so, generate the old-style compiler arguments?

CC @neonichu

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SKTestSupport/SkipUnless.swift
bnbarham: Sorry, what is generating the arguments here? By synthesizing I assumed that meant LSP, in which case, can't we just check if the folder exists? Or is it SwiftPM generating them?

ahoppen: SwiftPM is generating them.

bnbarham: Ah I see, `synthesize` just meant getting them. In that case... 💩 this is probably good enough then, suppose it's only for people on old toolchains anyway.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -0,0 +1,252 @@
...
+// MARK: - Skip checks
+
+/// Namespace for functions that are used to skip unsupported tests.
+public enum SkipUnless {
+  private enum FeatureCheckResult {
+    case featureSupported
+    case featureUnsupported(skipMessage: String)
+  }
+
+  /// For any feature that has already been evaluated, the result of wheather or not it should be skipped.
```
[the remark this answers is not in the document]
A-Alanbri: > ```suggestion
> 
>   /// For any feature that has already been evaluated, the result of whether or not it should be skipped.
> 
> ```

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -0,0 +1,252 @@
...
+      file: file,
+      line: line
+    ) {
+      let testClient = try await TestSourceKitLSPClient()
+      let uri = DocumentURI.for(.swift)
+      testClient.openDocument("func test() {}", uri: uri)
+      do {
+        _ = try await testClient.send(DocumentSemanticTokensRequest(textDocument: TextDocumentIdentifier(uri)))
+      } catch let error as ResponseError {
+        print(error.message)
```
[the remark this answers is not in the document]
A-Alanbri: > Leftover?

Sources/SKTestSupport/SkipUnless.swift
A-Alanbri: > Heh, thanks for the comment. I wrote a comment asking what the version is for then had to delete it after reading more :).

Sources/SKTestSupport/SkipUnless.swift
A-Alanbri: > ```suggestion
> 
>   /// toolchain’s SwiftPM stores the Swift modules on the top level but we synthesize compiler arguments expecting the
> 
> ```
> 
> 
> 
> Having said that, can we not just check this? ie. check for a `Modules` folder and otherwise assume top level?

PR #1052, opened 2024-02-14
Skip tests that require the workspace/indexedRename request if clangd doesn’t have support for it

ahoppen:


ahoppen:
@swift-ci Please test

PR #1057, opened 2024-02-16
Add missing dependencies of `SKSwiftPMWorkspaceTests`

ahoppen:
Looks like these dependency declarations were simply missing.


bnbarham (approved):
How was this working 🤔? They just happened to be already built normally?

ahoppen:
I assume (didn’t check) that it might be fine because all/many of these are transitive dependencies of `SKTestSupport`.

PR #1058, opened 2024-02-16
Rename `SKDValue` to `SKDRequestValue`

ahoppen:
`SKDValue` is only used for requests and the name should reflect that.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1060, opened 2024-02-17
Skip formatting tests if host toolchain doesn’t contain swift-format

ahoppen:


ahoppen:
@swift-ci Please test

PR #1061, opened 2024-02-17
Add `SwiftDriver` dependency to `SKSwiftPMWorkspace`

MaxDesiatov:
In https://github.com/apple/swift-package-manager/pull/7258 we need to use `SwiftDriver` explicitly to make sure that `package` access modifier and `-package-name` support can be detected properly. Somehow this dependency needs to be propagate to everyone depending `SourceKitLSPAPI`, even though the dependency was added as `PRIVATE` in SwiftPM. Adding it as `PRIVATE` here too.


ahoppen (approved):

ahoppen:
Looks like this is no longer needed because https://github.com/apple/swift-package-manager/pull/7258 has been merged without requiring this PR.

PR #1063, opened 2024-02-19
Log when a notification gets sent to the LSP client

ahoppen:


ahoppen:
@swift-ci Please test

PR #1064, opened 2024-02-19
Increase test timeout

ahoppen:
In some CI environments, some tests exceed the 2 min timeout.


ahoppen:
@swift-ci Please test

PR #1065, opened 2024-02-19
Skip `testSwiftInterfaceAcrossModules` if the host SwiftPM doesn’t store the built modules in a subdirectory

ahoppen:
I missed this in https://github.com/apple/sourcekit-lsp/pull/1046.


ahoppen:
@swift-ci Please test

PR #1066, opened 2024-02-19
Don’t display MARK comments as children of the nodes they are attached to

ahoppen:
We used to walk the leading and trailing trivia when walking the token that’s a child of the node. This caused us to eg. report the `MARK` comment in the following as a child of `myFunc` instead of a as a sibling of `myFunc`.

```swift
struct Foo {
  // MARK: Marker
  func myFunc() {}
}
```

rdar://121600714


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1067, opened 2024-02-19
Fix jumbled formatting results

### LINKED ISSUE #1059: Formatting via sourcekit-lsp sometimes produces syntax errors (Windows)
This file:
```
public extension Example {
    func function() { }
}
```
is formatted to:
```
extension Example {
    pfuubnlci cf unction() {
}
}
```
when formatting in VSCode via `sourcekit-lsp` (`Alt+Shift+f`).

If I invoke `swift-format` on the file directly, the output looks correct. I don't have any configuration defined, so I assume that the manual invocation and `sourcekit-lsp` invocation are both using the default config. I'm on Windows and don't have access to a Mac to determine if this is platform-specific. The problem reproduces with LF or CLRF (`\n` or `\r\n`).

Here are the `sourcekit-lsp` trace logs: https://pastebin.com/raw/u2zCan2Z
(Apologies for the external link; the log exceeded GitHub's comment character count)

ahoppen:
Fixing an error that computes the edits to perform to get from unformatted to the formatted source.

Fixes #1059
rdar://123108170

### DIFF
--- a/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
+++ b/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
@@ -116,7 +116,7 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
     // Adjust the index offset based on changes that `Collection.difference` expects to already have been applied.
     var adjustment: Int = 0
     for edit in edits {
-      if edit.range.upperBound < change.offset {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1068, opened 2024-02-19
Cancel the sourcekitd request to retrieve diagnostics if noone depends on it anymore

ahoppen:
After a document has been modified, the corresponding publish diagnsotics request gets cancelled. But that cancellation didn’t propagate through `DiagnosticReportManager`.

rdar://121323969


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1069, opened 2024-02-19
Add `-sdk` to request yaml in diagnose tests

ahoppen:
Otherwise, these tests fail when run with an Xcode toolchain.


ahoppen:
@swift-ci Please test

PR #1074, opened 2024-02-21
Only report a single item in prepareCallHierarchy

ahoppen:
Returning multiple doesn't make sense because they will all have the same USR (because we query them by USR) and will thus expand to the exact same call hierarchy.

Also, VS Code doesn't seem to like multiple call hiearchy items being returned and fails to display any results if they are, failing with `Cannot read properties of undefined (reading 'map')`.

rdar://118957893

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1752,10 +1752,7 @@ extension SourceKitServer {
     }
     guard let usr = symbol.usr else { return [] }
     logger.info("performing indexed jump-to-def with usr \(usr)")
-    var occurrences = index.occurrences(ofUSR: usr, roles: [.definition])
-    if occurrences.isEmpty {
-      occurrences = index.occurrences(ofUSR: usr, roles: [.declaration])
-    }
+    var occurrences = index.definitionOrDeclarationOccurances(ofUSR: usr)
     if symbol.isDynamic ?? true {
       lazy var transitiveReceiverUsrs: [String]? = {
         if let receiverUsrs = symbol.receiverUsrs {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2255,6 +2258,18 @@ private let maxWorkspaceSymbolResults = 4096
 
 public typealias Diagnostic = LanguageServerProtocol.Diagnostic
 
+fileprivate extension IndexStoreDB {
+  /// If there are any definition occurances of the given USR, return these.
+  /// Otherwise return declaration occurances.
+  func definitionOrDeclarationOccurances(ofUSR usr: String) -> [SymbolOccurrence] {
```
bnbarham, line 2264:
```suggestion
  func definitionOrDeclarationOccurences(ofUSR usr: String) -> [SymbolOccurrence] {
```

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1963,17 +1960,23 @@ extension SourceKitServer {
...
-          moduleName: info.location.moduleName,
-          location: location
-        )
-      }
+
+    // Only return a single call hierarchy item. Returning multiple doesn't make sense because they will all have the
+    // same USR (because we query them by USR) and will thus expand to the exact same call hierarchy.
+    // Also, VS Code doesn't seem to like multiple call hiearchy items being returned and fails to display any results
+    // if they are, failing with `Cannot read properties of undefined (reading 'map')`.
+    guard let definition = index.definitionOrDeclarationOccurances(ofUSR: usr).first else {
```
bnbarham, line 1968: Should we sort by location/something so this is stable?

Tests/SourceKitLSPTests/CallHierarchyTests.swift
```diff
@@ -183,4 +183,45 @@ final class CallHierarchyTests: XCTestCase {
       ]
     )
   }
+
+  func testReportSingleItemInPrepareCallHierarchy() async throws {
```
bnbarham, line 187: Another test here (though not sure how easy it is with our infrastructure) would be two functions wrapped in `#if #else #endif` - they would both have the same USR but different locations.

bnbarham (approved):

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: That applies to quite a few locations. I filed rdar://123487052 and will take care of it in a follow-up PR.

Tests/SourceKitLSPTests/CallHierarchyTests.swift
ahoppen: Discussed offline that it doesn’t make sense to test this since one file can currently on have one set of compiler arguments in sourcekit-lsp.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1075, opened 2024-02-21
Don’t request syntactic information from sourcekitd on open and edit

ahoppen:
Just means that sourcekitd doesn’t need to build a syntax tree, doesn’t have to transfer information that we discard anyway and also doesn’t run `SyntaxModelContext` and thus can’t hit assertion failures within.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1077, opened 2024-02-22
De-dupe locations returned from the defintions request

### LINKED ISSUE #1055: `textDocument/definition` returns duplicate results when called on constructors
Similar to #1047, sourcekit-lsp returns duplicate results for `textDocument/definition` when called on a constructor. This happens for both value and reference types. The same does not happen when called on the type itself e. g. in a variable declaration.

image

ahoppen:
We might end up with duplicate locations when performing a definition request on eg. `MyStruct()` when no explicit initializer is declared. In this case we get two symbol infos, one for the declaration of the `MyStruct` type and one for the initializer, which is implicit and thus has the location of  the `MyStruct` declaration itself.

rdar://123036565
Fixes #1055

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1818,7 +1818,7 @@ extension SourceKitServer {
         position: req.position
       )
     )
-    return try await symbols.asyncMap { (symbol) -> [Location] in
+    let locations = try await symbols.asyncMap { (symbol) -> [Location] in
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

wojciech-kulik:
In which version this fix is expected?

ahoppen:
This is included in SourceKit-LSP bundled with Swift 6.0.

PR #1080, opened 2024-02-23
If sourcekitd doesn't support the semantic tokens request, return results from swift-syntax

### LINKED ISSUE #1009: DocumentSemanticTokensRequest always fail
Requesting semanic tokens does not work, however it used to work

Sending `DocumentSemanticTokensRequest` to the sourcekit-lsp ends up with `unknown request: source.request.semantic_tokens` (which is silenced in tests) with `XCTSkip("semantic tokens request not supported by sourcekitd")`. That is today's state. 

In the past, probably before the https://github.com/apple/sourcekit-lsp/pull/912 server send `WorkspaceSemanticTokensRefreshRequest` with tokens, but also sending `DocumentSemanticTokensRequest` worked with tokens.

It seems something is missing that it no longer works (or what am I missing?)

I confirmed it works with sourcekit-lsp shipped with Xcode 15.1 (15C65), so that looks like regression to me.

ahoppen:
This offers a better fallback than not returning any results at all.

Fixes #1009
rdar://120124113

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -12,9 +12,11 @@
 
 import Foundation
 import LSPLogging
+import LSPTestSupport
 import LanguageServerProtocol
 import RegexBuilder
 @_spi(Testing) import SKCore
+import SourceKitLSP
 import XCTest
 
 import enum PackageLoading.Platform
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1083, opened 2024-02-24
Show message in the client if no compiler arguments could be found for a file

ahoppen:
This should help users debug issues when functionality is not working because compiler arguments are missing.

Note that when opening a standalone Swift file, we consider that file as having non-fallback build settings, even though the build settings are synthesized from the fallback build system. Thus, this will only pop up when the user is in a context that has some build system (like a compilation database or a Swift package) but the current file isn’t handled by that.

rdar://119741960

### DIFF
--- a/Sources/SKTestSupport/TestSourceKitLSPClient.swift
+++ b/Sources/SKTestSupport/TestSourceKitLSPClient.swift
@@ -201,17 +201,28 @@ public final class TestSourceKitLSPClient: MessageHandler {
   public func nextDiagnosticsNotification(
     timeout: TimeInterval = defaultTimeout
   ) async throws -> PublishDiagnosticsNotification {
-    struct CastError: Error, CustomStringConvertible {
-      let actualType: any NotificationType.Type
+    return try await nextNotification(ofType: PublishDiagnosticsNotification.self, timeout: timeout)
+  }
 
-      var description: String { "Expected a publish diagnostics notification but got '\(actualType)'" }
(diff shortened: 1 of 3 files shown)


bnbarham (approved):
How is the notification displayed?

adam-fowler:
Will this message appear for new files that haven't been saved. A new file doesn't have a file name so I assume would be considered standalone. 

Also how this affect swift scripts?

ahoppen:
Oh, good thinking with the newly created files. I didn’t think about that. Ensured that we don’t get the warning for these. 

Swift scripts don’t show this message. That’s what I meant by this comment in the PR description. Definitely wasn’t phrased very understandably 😉 
> Note that when opening a standalone Swift file, we consider that file as having non-fallback build settings, even though the build settings are synthesized from the fallback build system. Thus, this will only pop up when the user is in a context that has some build system (like a compilation database or a Swift package) but the current file isn’t handled by that.

ahoppen:
The displayed message looks as follows
Screenshot 2024-02-23 at 23 37 29
And if you click on it, it expands to
Screenshot 2024-02-23 at 23 37 32

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1090, opened 2024-02-24
Remove logging dependency from `SourceKitD`

ahoppen:
Make logging protocol requirements on `SourceKitD` so that implementers of the protocol can decide how to log requests. This allows us to add different kinds of sourcekitd connections that behave differently in the future, for example one that dynamically links against sourcekitd instead of loading it using `dlopen`.

Also rename `SourceKitDImpl` to `DynamicallyLoadedSourceKitD`, for the reasons described above.

### DIFF
--- a/Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
+++ b/Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
@@ -21,7 +21,7 @@ import struct TSCBasic.AbsolutePath
 ///
 /// Users of this class should not call the api functions `initialize`, `shutdown`, or
 /// `set_notification_handler`, which are global state managed internally by this class.
-public final class SourceKitDImpl: SourceKitD {
+public final class DynamicallyLoadedSourceKitD: SourceKitD {
 
   /// The path to the sourcekitd dylib.
   public let path: AbsolutePath
@@ -49,7 +49,7 @@ public final class SourceKitDImpl: SourceKitD {
 
   public static func getOrCreate(dylibPath: AbsolutePath) throws -> SourceKitD {
     try SourceKitDRegistry.shared
-      .getOrAdd(dylibPath, create: { try SourceKitDImpl(dylib: dylibPath) })
+      .getOrAdd(dylibPath, create: { try DynamicallyLoadedSourceKitD(dylib: dylibPath) })
   }
 
   init(dylib path: AbsolutePath) throws {
--- a/Sources/SourceKitD/SourceKitD.swift
+++ b/Sources/SourceKitD/SourceKitD.swift
@@ -112,10 +112,6 @@ extension SourceKitD {
 
     return dict
   }
-
-  public func log(request: SKDRequestDictionary) {}
-  public func log(response: SKDResponse) {}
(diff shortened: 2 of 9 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

MaxDesiatov:
@swift-ci test windows

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -83,56 +96,26 @@ extension SourceKitD {
...
-  logger.log(
-    level: (response.error == nil || response.error == .requestCancelled) ? .debug : .error,
-    """
-    Received sourcekitd response:
-    \(response.forLogging)
-    """
-  )
+  public func log(request: SKDRequestDictionary) {}
+  public func log(response: SKDResponse) {}
+  public func log(crashedRequest: SKDRequestDictionary, fileContents: String?) {}
```
bnbarham, line 118: Is it worth having empty defaults?

Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
```diff
@@ -20,7 +21,7 @@ import struct TSCBasic.AbsolutePath
 ///
 /// Users of this class should not call the api functions `initialize`, `shutdown`, or
 /// `set_notification_handler`, which are global state managed internally by this class.
-public final class SourceKitDImpl: SourceKitD {
+public final class DynamicallyLoadedSourceKitD: SourceKitD {
```
bnbarham, line 24: `Impl` be gone! Much clearer 👍

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -70,11 +87,7 @@ extension SourceKitD {
   ///   - fileContents: The contents of the file that the request operates on. If sourcekitd crashes, the file contents
   ///     will be logged.
   public func send(_ req: SKDRequestDictionary, fileContents: String?) async throws -> SKDResponseDictionary {
-    logRequest(req)
-
-    let signposter = logger.makeSignposter()
-    let signpostID = signposter.makeSignpostID()
-    let signposterState = signposter.beginInterval("sourcekitd-request", id: signpostID, "Start")
```
bnbarham, line 77: Did you decide the signpost wasn't useful? You could have `logRequest` return some opaque context that `log(response:)` and `log(crashedRequest:)` take if we want to keep it.

bnbarham (approved):

Sources/SourceKitD/SourceKitD.swift
ahoppen: Yeah. It was a mix of not useful and harder to implement. Just don’t make sense to design something with abstraction layers when it’s doubtful that it provides any real-world value.

Sources/SourceKitD/SourceKitD.swift
ahoppen: True, probably better to not have them and force adopters to make decisions about whether they support logging or not.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1093, opened 2024-02-27
Sleep sourcekit-lsp’s main thread for 10 years instead of leaking a continuation

ahoppen:
The previous implementation of never fulfilling a continuation printed the following message to stderr:

```
SWIFT TASK CONTINUATION MISUSE: run() leaked its continuation!
```


ahoppen:
@swift-ci Please test

bnbarham (approved):
Oh that's disappointing :(

PR #1095, opened 2024-02-28
Sort results returned by the index

ahoppen:
rdar://123487052

### DIFF
--- a/Sources/SourceKitLSP/SourceKitServer.swift
+++ b/Sources/SourceKitLSP/SourceKitServer.swift
@@ -1961,18 +1961,19 @@ extension SourceKitServer {
         position: req.position
       )
     )
-    guard let symbol = symbols.first,
-      let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
-    else {
+    guard let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index else {
       return nil
     }
-    guard let usr = symbol.usr else { return nil }
-    var occurrences = index.occurrences(ofUSR: usr, roles: .baseOf)
-    if occurrences.isEmpty {
-      occurrences = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
-    }
+    let locations = symbols.flatMap { (symbol) -> [Location] in
+      guard let usr = symbol.usr else { return [] }
+      var occurrences = index.occurrences(ofUSR: usr, roles: .baseOf)
+      if occurrences.isEmpty {
+        occurrences = index.occurrences(relatedToUSR: usr, roles: .overrideOf)
+      }
 
-    return .locations(occurrences.compactMap { indexToLSPLocation($0.location) }.sorted())
+      return occurrences.compactMap { indexToLSPLocation($0.location) }
+    }
+    return .locations(locations.sorted())
   }
 
   func references(
@@ -1986,17 +1987,19 @@ extension SourceKitServer {
         position: req.position
       )
     )
-    guard let symbol = symbols.first, let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index
-    else {
+    guard let index = await self.workspaceForDocument(uri: req.textDocument.uri)?.index else {
       return []
     }
-    guard let usr = symbol.usr else { return [] }
-    logger.info("performing indexed jump-to-def with usr \(usr)")
-    var roles: SymbolRole = [.reference]
-    if req.context.includeDeclaration {
-      roles.formUnion([.declaration, .definition])
+    let locations = symbols.flatMap { (symbol) -> [Location] in
+      guard let usr = symbol.usr else { return [] }
+      logger.info("Finding references for USR \(usr)")
+      var roles: SymbolRole = [.reference]
+      if req.context.includeDeclaration {
+        roles.formUnion([.declaration, .definition])
+      }
+      return index.occurrences(ofUSR: usr, roles: roles).compactMap { indexToLSPLocation($0.location) }
     }
-    return index.occurrences(ofUSR: usr, roles: roles).compactMap { indexToLSPLocation($0.location) }.sorted()
+    return locations.sorted()
   }
 
   private func indexToLSPCallHierarchyItem(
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2030,7 +2030,7 @@ extension SourceKitServer {
         fromRanges: [location.range]
       )
     }
-    return calls
```
bnbarham, line 2033: The call above is still picking a definition at random right? Given the sorting of SymbolOccurrence, should this also be by location and then name?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1551,25 +1551,24 @@ extension SourceKitServer {
...
-        //
-        // Until then, take the first `maxWorkspaceSymbolResults` symbols to limit the impact of
-        // queries which match many symbols.
-        return symbolOccurrenceResults.count < maxWorkspaceSymbolResults
+        return true
       }
+      try Task.checkCancellation()
     }
-    return symbolOccurrenceResults
+    return symbolOccurrenceResults.sorted()
```
bnbarham, line 1565: From looking at indexstore-db this is by location, then roles, then symbol. And the symbol is by usr and then by name.

Maybe that makes sense? The usr then name seems a bit odd to me. It would look pretty random to a user.

ahoppen: Honestly, given that it sorts by location first, I don’t really care about the remaining sorting. As long as we get something that’s deterministic from it, I’m fine.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Added a commit to the PR that gets rid of all the cases where we use `.first` on index results. Meant to do that anyway, so now is as good a time as any.

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -1986,17 +1987,19 @@ extension SourceKitServer {
...
       return []
     }
-    guard let usr = symbol.usr else { return [] }
-    logger.info("performing indexed jump-to-def with usr \(usr)")
-    var roles: SymbolRole = [.reference]
-    if req.context.includeDeclaration {
-      roles.formUnion([.declaration, .definition])
+    let locations = symbols.flatMap { (symbol) -> [Location] in
+      guard let usr = symbol.usr else { return [] }
+      logger.info("performing indexed jump-to-def with usr \(usr)")
```
bnbarham, line 1995: Was this an old copy paste from somewhere? This is a references right?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2031,29 +2034,34 @@ extension SourceKitServer {
...
-    let callHierarchyItem = self.indexToLSPCallHierarchyItem(
-      symbol: definition.symbol,
-      moduleName: definition.location.moduleName,
-      location: location
-    )
-    return [callHierarchyItem]
+
+    // Ideally, we should show multiple symbols. But VS Code fails to display call hierarchies with multiple root items,
+    // failing with `Cannot read properties of undefined (reading 'map')`. Pick the first one.
+    callHierarchyItems = Array(callHierarchyItems.prefix(1))
```
bnbarham, line 2061: `[callHierarchyItems.first]` would be shorter, any reason to prefer prefix?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2031,29 +2034,34 @@ extension SourceKitServer {
...
+    return
+      callHierarchyItems
+      .sorted(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
```
bnbarham, line 2064: Isn't this max length 1?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2196,26 +2204,38 @@ extension SourceKitServer {
...
+        return nil
+      }
+      return self.indexToLSPTypeHierarchyItem(
+        symbol: info.symbol,
+        moduleName: info.location.moduleName,
+        location: location,
+        index: index
+      )
+    }
+    .sorted(by: { $0.name < $1.name })
```
bnbarham, line 2238: Does type hierarchy allow multiple roots when call doens't?

Sources/SourceKitLSP/SourceKitServer.swift
```diff
@@ -2289,14 +2314,19 @@ extension SourceKitServer {
 
     // Convert occurrences to type hierarchy items
     let types = occurs.compactMap { occurrence -> TypeHierarchyItem? in
-      guard let location = indexToLSPLocation(occurrence.location),
-        let related = occurrence.relations.first
+      if occurrence.relations.count > 1 {
+        // An occurrence with a `baseOf` of `extendedBy` relation is an occurrence inside an inheritance clause.
```
bnbarham, line 2318: `of` -> `or` (I'm in a commit so can't suggest a replacement)

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: Oh yes. Changed it.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: It would be `[callHierarchyItem.first].compactMap { $0 }`. Otherwise you get an array with a single optional not an array with 0 or 1 items containing non-optionals. And then I thought `.prefix` is nicer.

Sources/SourceKitLSP/SourceKitServer.swift
ahoppen: My idea was that this continues to work if we remove the `.prefix(1)` workaround at some point.

ahoppen: Also, I realized just now that I wanted to sort before taking the first item.

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1096, opened 2024-02-28
Add swift-format to the dependencies so the `Format Swift Code` command plugin can be used to format the codebase

ahoppen:
I deliberately didn’t add it to the local dependencies so we notice when a build dependency on swift-format is being introduced.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham:

PR #1099, opened 2024-02-28
Support rename of function parameters from variables that reference them

ahoppen:
When renaming a function parameter from inside the function body, we expect that to be a local rename and thus we shouldn’t update the external parameter name. We should thus introduce the new renamed variable name as the parameter’s second name.

rdar://123536538


bnbarham (approved):
Interesting idea!

ahoppen:
@swift-ci Please test

PR #1101, opened 2024-02-29
Add user default to log full sourcekitd request

ahoppen:
This is useful during debugging if a sourcekit-lsp developer knows which sourcekitd request is causing an issue and wants to access the request's YAML.

This option is intended to be turned off as soon as the request in question has been retrieved.

rdar://121322828


ahoppen:
@swift-ci Please test

ahoppen:
Not adding this for now. Instead, we re-order the sourcekitd keys to print compilerargs and source text last, which should make the important parts of the request end up in the log without any modification of log settings.

PR #1103, opened 2024-02-29
Show calls to satisfied protocol requirements in call hierarchy

ahoppen:
For example in the following, we should show `proto.foo()` as a call when computing the call hierarchy of `MyStruct.foo`. Otherwise `MyStruct.foo` does not have any calls, which is misleading.

```swift
protocol MyProtocol {
  func foo()
}
struct MyStruct: MyProtocol {
  func foo() {}
}
func test(proto: MyProtocol) {
  proto.foo()
}
```

rdar://123837232

### DIFF
--- a/Tests/SourceKitLSPTests/CallHierarchyTests.swift
+++ b/Tests/SourceKitLSPTests/CallHierarchyTests.swift
@@ -447,4 +447,102 @@ final class CallHierarchyTests: XCTestCase {
       ]
     )
   }
+
+  func testIncomingCallHierarchyLooksThroughProtocols() async throws {
+    let ws = try await IndexedSingleSwiftFileWorkspace(
+      """
+      protocol MyProtocol {
+        func foo()
+      }
+      struct MyStruct: MyProtocol {
+        func 1️⃣foo() {}
+      }
+      struct Unrelated: MyProtocol {
+        func foo() {}
+      }
+      func 2️⃣test(proto: MyProtocol) {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/CallHierarchyTests.swift
```diff
@@ -447,4 +447,53 @@ final class CallHierarchyTests: XCTestCase {
       ]
     )
   }
+
+  func testIncomingCallHierarchyLooksThroughProtocols() async throws {
```
bnbarham, line 451: I don't expect it to fail, but you could have a test for the class hierarchy equivalent as well. That could also have the case of `A -> B -> C` where `foo` is on each and callers of `B.foo()` shouldn't find `C.foo()`.

A test case for properties would be good too.

bnbarham (approved):

Tests/SourceKitLSPTests/CallHierarchyTests.swift
ahoppen: Added a test case. Passed

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

PR #1105, opened 2024-02-29
Refactor willEditDocument in favor of SourceEdits

### LINKED ISSUE #939: Instead of having a `willEditDocument` closure in `DocumentManager.edit` return an array of edits
See https://github.com/apple/sourcekit-lsp/pull/934/files#r1373963800

Instead of calling a `willEditDocument` callback from `DocumentManager.edit`, return an array of edits. `SwiftRefactor.SourceEdit` seems like it could be a good candidate to use for the edit struct.

mustiikhalil:
The following PR refactors the willEditDocument in the `SwiftLanguageServer` into the `DocumentManager` focusing on using SourceEdit.

Things I am not sure of:

- Is SourceEdit really the correct way to go about it?
- Naming convention and if it aligns with the repository's demands

Closes #939

### DIFF
--- a/Sources/SourceKitLSP/DocumentManager.swift
+++ b/Sources/SourceKitLSP/DocumentManager.swift
@@ -14,6 +14,7 @@ import Dispatch
 import LSPLogging
 import LanguageServerProtocol
 import SKSupport
+import SwiftSyntax
 
 /// An immutable snapshot of a document at a given time.
 ///
@@ -140,27 +141,25 @@ public final class DocumentManager {
   ///   - newVersion: The new version of the document. Must be greater than the
   ///     latest version of the document.
   ///   - edits: The edits to apply to the document
-  ///   - willEditDocument: Optional closure to call before each edit. Will be
-  ///     called multiple times if there are multiple edits.
-  /// - Returns: The snapshot of the document before the edit and the snapshot
-  ///   of the document after the edit.
+  /// - Returns: The snapshot of the document before the edit, the snapshot
+  ///   of the document after the edit, and the edits. The edits are sequential, ie.
+  ///   the edits are expected to be applied in order and later values in this array
+  ///   assume that previous edits are already applied.
   @discardableResult
   public func edit(
     _ uri: DocumentURI,
     newVersion: Int,
-    edits: [TextDocumentContentChangeEvent],
-    willEditDocument: ((_ before: LineTable, TextDocumentContentChangeEvent) -> Void)? = nil
-  ) throws -> (preEditSnapshot: DocumentSnapshot, postEditSnapshot: DocumentSnapshot) {
+    edits: [TextDocumentContentChangeEvent]
+  ) throws -> (preEditSnapshot: DocumentSnapshot, postEditSnapshot: DocumentSnapshot, edits: [SourceEdit]) {
     return try queue.sync {
       guard let document = documents[uri] else {
         throw Error.missingDocument(uri)
       }
       let preEditSnapshot = document.latestSnapshot
 
+      var sourceEdits: [SourceEdit] = []
       for edit in edits {
-        if let willEditDocument {
-          willEditDocument(document.latestLineTable, edit)
-        }
+        sourceEdits.append(SourceEdit(edit: edit, lineTableBeforeEdit: document.latestLineTable))
 
         if let range = edit.range {
           document.latestLineTable.replace(
@@ -180,7 +179,7 @@ public final class DocumentManager {
         logger.error("Document version did not increase on edit from \(document.latestVersion) to \(newVersion)")
       }
       document.latestVersion = newVersion
-      return (preEditSnapshot, document.latestSnapshot)
+      return (preEditSnapshot, document.latestSnapshot, sourceEdits)
     }
   }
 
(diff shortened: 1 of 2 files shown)


Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -192,6 +194,35 @@ public final class DocumentManager {
...
+
+  private func willEditDocument(_ before: LineTable, edit: TextDocumentContentChangeEvent) -> SourceEdit {
+    guard let range = edit.range else {
+      return SourceEdit(
+        range: Range(
+          uncheckedBounds: (
+            AbsolutePosition(utf8Offset: 0), 
+            upper: AbsolutePosition(utf8Offset: before.content.utf8.count)
+          )
+        ),
```
ahoppen, line 206: Since you’re touching this anyway, let’s not use `uncheckedBounds`. If the bounds aren’t valid, it will just blow up later and make it a lot harder to debug the code. And I don’t think the bounds check is a performance bottleneck.

```suggestion
        range: AbsolutePosition(utf8Offset: 0)..<AbsolutePosition(utf8Offset: before.content.utf8.count)
```

Same a little further below.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
```diff
@@ -1129,3 +1112,10 @@ extension sourcekitd_api_uid_t {
     }
   }
 }
+
+extension SourceEdit {
+  var positiveLength: Int {
+    // Required since SourceEdit returns a negative length
+    -length.utf8Length
+  }
```
ahoppen, line 1120: Oh, we should fix this in SwiftSyntax. Could you open a PR in swift-syntax to change that?

ahoppen (commented):
The overall design looks good to me. I’ve got a few mostly stylistic comments.

Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift
mustiikhalil: Yeah of course. However the comment there was a tiny bit misleading which made me think that it should be zero if the value is positive

```swift
  /// Length of the original source range that this edit applies to. Zero if
  /// this is an addition.
  public var length: SourceLength {
    return SourceLength(utf8Length: range.lowerBound.utf8Offset - range.upperBound.utf8Offset)
  }
 ```

Comment (ahoppen) [reply Sources/SourceKitLSP/Swift/SwiftLanguageServer.swift:1115]: The `Zero if this is an addition` meant that we represent an assertion at offset 5 by a replacement of range `5..<5` by the new text.

File: Sources/SourceKitLSP/DocumentManager.swift [hunk 3]
```diff
@@ -140,27 +141,25 @@ public final class DocumentManager {
   ///   - newVersion: The new version of the document. Must be greater than the
   ///     latest version of the document.
   ///   - edits: The edits to apply to the document
-  ///   - willEditDocument: Optional closure to call before each edit. Will be
-  ///     called multiple times if there are multiple edits.
+  ///   - requiresUpcomingEdits: Flag to return upcoming editst. Will be
+  ///     calling ``willEditDocument`` for each edit made
```
ahoppen, line 145: This parameter no longer exists. 😉

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -214,20 +214,40 @@ extension DocumentManager {
     }
   }
 
-  /// Convenience wrapper for `edit(_:newVersion:edits:willEditDocument:updateDocumentTokens:)`
+  /// Convenience wrapper for `edit(_:newVersion:edits:requiresUpcomingEdits:updateDocumentTokens:)`
```
ahoppen, line 217: Same here `requiresUpcomingEdits` doesn’t exist anymore.

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -140,27 +141,25 @@ public final class DocumentManager {
...
-  ) throws -> (preEditSnapshot: DocumentSnapshot, postEditSnapshot: DocumentSnapshot) {
+    edits: [TextDocumentContentChangeEvent]
+  ) throws -> (preEditSnapshot: DocumentSnapshot, postEditSnapshot: DocumentSnapshot, edits: [SourceEdit]) {
     return try queue.sync {
       guard let document = documents[uri] else {
         throw Error.missingDocument(uri)
       }
       let preEditSnapshot = document.latestSnapshot
 
+      var upcomingEdits: [SourceEdit] = []
```
ahoppen, line 160: I’m not a huge fan of the `upcomingEdits` naming. Maybe `sourceEdits` is a better name? At least that describes that the differentiator between this array and `edits` is that it has a different type of edit.

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -140,27 +141,25 @@ public final class DocumentManager {
   ///   - newVersion: The new version of the document. Must be greater than the
   ///     latest version of the document.
   ///   - edits: The edits to apply to the document
-  ///   - willEditDocument: Optional closure to call before each edit. Will be
-  ///     called multiple times if there are multiple edits.
+  ///   - requiresUpcomingEdits: Flag to return upcoming editst. Will be
+  ///     calling ``willEditDocument`` for each edit made
   /// - Returns: The snapshot of the document before the edit and the snapshot
-  ///   of the document after the edit.
+  ///   of the document after the edit, including all edits that were made.
```
ahoppen, line 147: I would like to add whether the edits are concurrent or sequential to the doc comment. (see `ConcurrentEdits` in swift-syntax, it has some comments on the difference).

I would add 

```
The edits are sequential, ie. the edits are expected to be applied in order and later values in this array assume that previous edits are already applied.
```

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -192,6 +191,7 @@ public final class DocumentManager {
       return document.latestSnapshot
     }
   }
+
```
ahoppen, line 194: Nit: Superfluous empty line

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -214,20 +214,40 @@ extension DocumentManager {
...
-        edits: note.contentChanges,
-        willEditDocument: willEditDocument
+        edits: note.contentChanges
       )
     }
   }
 }
+
+fileprivate extension SourceEdit {
+
```
ahoppen, line 234: Nit: Superfluous empty line

Sources/SourceKitLSP/DocumentManager.swift
```diff
@@ -214,20 +214,40 @@ extension DocumentManager {
...
+        replacement: edit.text
+      )
+    } else {
+      self.init(
+        range: AbsolutePosition(utf8Offset: 0)..<AbsolutePosition(utf8Offset: lineTableBeforeEdit.content.utf8.count),
+        replacement: edit.text
+      )
+    }
+  }
+
```
ahoppen, line 252: Nit: Superfluous empty line

ahoppen (commented):
I’ve got a couple more nitpicky comments. Other than that we should use https://github.com/apple/swift-syntax/pull/2517 once that’s in and the PR is good to go 🚢

MaxDesiatov:
@swift-ci test

ahoppen:
@mustiikhalil Could you run swift-format on your changes? If you rebase it on top of current `main`, you should be able to run it by right-clicking on the Package in Xcode.

Also, would you mind squashing your commits?

mustiikhalil:
Sure! and done. 

I've noticed a weird behavior that the command `format-source-code`, it doesn't really update the source code if ran through Xcode. It might be my setup.

ahoppen:
Hmm, weird. I’ve recently been running it from VS Code, which worked fine and I think I also saw it working in Xcode at some point. Maybe you didn’t give it permission to write to your source directory. Anyway, thanks for updating.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1108, opened 2024-03-01
Add a progress bar to the diagnose subcommand

ahoppen:
Gives some feedback about how long diagnostic bundle generation still takes. As always with progress indicators, everything is an estimate but I feel that giving some progress estimate gives some valuable feedback that something is happening. It also allows us to show the current state inline without accumulating a huge log of previous states.

Screenshot 2024-03-01 at 09 15 21


bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1109, opened 2024-03-02
Don’t log error if workspace root doesn’t contain Swift package manifest

ahoppen:
We try creating a `SwiftPMWorkspace` on every workspace to check if it is a SwiftPM package. The workspace root not having a package manifest is not an error because it might contain compile_commands.json. Remove the log message.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
@@ -188,8 +188,7 @@ public actor SwiftPMWorkspace {
         buildSetup: buildSetup,
         reloadPackageStatusCallback: reloadPackageStatusCallback
       )
-    } catch Error.noManifest(let path) {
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SKSwiftPMWorkspace/SwiftPMWorkspace.swift
```diff
@@ -188,8 +188,7 @@ public actor SwiftPMWorkspace {
         buildSetup: buildSetup,
         reloadPackageStatusCallback: reloadPackageStatusCallback
       )
-    } catch Error.noManifest(let path) {
-      logger.error("could not find manifest, or not a SwiftPM package: \(path)")
```
bnbarham, line 192: An info log/similar is probably still useful?

ahoppen: I don’t think so

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1114, opened 2024-03-02
Don’t include internal parameter names in LSP snippets

ahoppen:
When translating editor placeholders to snippets, we should not include internal parameter names into the snippet. The problem is that once you leave snippet editing mode (essentially by moving the cursor to any place that’s not the snippet, eg. by typing a string literal for one of the parameters or moving the cursor), all snippets become verbatim text. And then it’s impossible to tell whether `paramName: ` was an annotation inside the snippet describing the internal parameter’s name or the external parameter label.

And it turns out that we can do this by just using the placeholder parsing from swift-syntax instead of having our own in SourceKit-LSP.

rdar://123772474


bnbarham (approved):

ahoppen:
https://github.com/apple/swift-syntax/pull/2519

@swift-ci Please test

PR #1117, opened 2024-03-04
Add Sendable annotations to LanguageServerProtocol, LSPLogging and SKSupport

ahoppen:
- LanguageServerProtocol was just about slapping `Sendable` on basically everything because all the request and response types are value types and sendable
- LSPLogging was also pretty trivial
- SKSupport required a little bit of restructuring in `AsyncQueue` to get sendable checks to pass. The change actually localize potential races, so it’s for the better.

The most controversial change might be that I’m raising the package description version to 5.8 and adding `.enableExperimentalFeature("StrictConcurrency")` to the audited modules. But I think it’s fine. I checked that the package still builds with Xcode 15.0.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.8
 
 import Foundation
 import PackageDescription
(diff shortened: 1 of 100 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1118, opened 2024-03-05
Use logic in swift-syntax to translate sequential edits to concurrent edits

ahoppen:
Depends on https://github.com/apple/swift-syntax/pull/2527

Turns out that translation of sequential to concurrent edits is not trivial to implement. Use the logic in swift-syntax that is already well-tested.

rdar://123880677


ahoppen:
https://github.com/apple/swift-syntax/pull/2527

@swift-ci Please test

bnbarham (approved):

PR #1120, opened 2024-03-05
Fix a race condition in `TestMessageHandler`

ahoppen:
Not entirely sure what introduced the race condition. Found by thread sanitizer. Probably wouldn’t have been an issue if we had migrated this code to strict concurrency checking.


ahoppen:
@swift-ci Please test

ahoppen:
Merged to unblock CI. If there are review comments, I'll address them in a follow-up PR.

PR #1128, opened 2024-03-12
Change `#if swift(<5.11)` to `#if compiler(<5.11)`

ahoppen:
We want to check the compiler version to decide whether to use `@retroactive`, not the effective language version (which is 5.10 for a Swift 6 compiler).

### DIFF
--- a/Sources/LSPTestSupport/TestJSONRPCConnection.swift
+++ b/Sources/LSPTestSupport/TestJSONRPCConnection.swift
@@ -189,7 +189,7 @@ private let testMessageRegistry = MessageRegistry(
   notifications: [EchoNotification.self]
 )
 
-#if swift(<5.11)
+#if compiler(<5.11)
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/LSPTestSupport/TestJSONRPCConnection.swift
```diff
@@ -189,7 +189,7 @@ private let testMessageRegistry = MessageRegistry(
   notifications: [EchoNotification.self]
 )
 
-#if swift(<5.11)
+#if compiler(<5.11)
```
bnbarham, line 192: `<=5.10` might make more sense?

ahoppen: We can only check `>=` or `<` so that’s not possible.

PR #1130, opened 2024-03-15
Don’t print log test log output when running tests in parallel

ahoppen:
`swift test --parallel --verbose` prints the entire log output. We were assuming that it did not and re-run `sourcekit-lsp` tests in serial with log output if the parallel test run failed.

Suppress `--verbose` on `swift test`. This will cause it to not log any stderr logging output when running in parallel and log all stderr logging output when running in serial.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -90,7 +90,7 @@ def get_build_target(swift_exec: str, args: argparse.Namespace, cross_compile: b
 # Build SourceKit-LSP
 
 
-def get_swiftpm_options(swift_exec: str, args: argparse.Namespace) -> List[str]:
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -197,7 +197,7 @@ def run_tests(swift_exec: str, args: argparse.Namespace) -> None:
     """
     Run all tests in the package
     """
-    swiftpm_args = get_swiftpm_options(swift_exec, args)
+    swiftpm_args = get_swiftpm_options(swift_exec, args, suppress_verbose=True)
```
bnbarham, line 200: Need to calculate args for the serial case below right?

bnbarham (approved):

PR #1134, opened 2024-03-17
Don’t fail OSLogScraper if a single crashed request could not be parsed

ahoppen:
If we fail to extract the failed request from one crashed sourcekitd request, we should not fail sourcekitd request reduction altogether. Instead, we should just continue searching for the next crashed requested and see if we can reduce that one.

### DIFF
--- a/Sources/Diagnose/OSLogScraper.swift
+++ b/Sources/Diagnose/OSLogScraper.swift
@@ -82,7 +82,12 @@ struct OSLogScraper {
   /// Name is a human readable name that identifies the crash.
   func getCrashedRequests() throws -> [(name: String, info: RequestInfo)] {
     let crashedRequests = try crashedSourceKitLSPRequests().reversed()
-    return try crashedRequests.map { ($0.name, try requestInfo(for: $0.logCategory)) }
+    return crashedRequests.compactMap { (name: String, logCategory: String) -> (name: String, info: RequestInfo)? in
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/Diagnose/OSLogScraper.swift
```diff
@@ -82,7 +82,12 @@ struct OSLogScraper {
   /// Name is a human readable name that identifies the crash.
   func getCrashedRequests() throws -> [(name: String, info: RequestInfo)] {
     let crashedRequests = try crashedSourceKitLSPRequests().reversed()
-    return try crashedRequests.map { ($0.name, try requestInfo(for: $0.logCategory)) }
+    return crashedRequests.compactMap { (name: String, logCategory: String) -> (name: String, info: RequestInfo)? in
+      guard let requestInfo = try? requestInfo(for: logCategory) else {
+        return nil
```
bnbarham, line 87: Worth printing it failed, or not useful?

bnbarham (approved):

Sources/Diagnose/OSLogScraper.swift
ahoppen: Printing messes with the progress animation. And in general we don’t want to bother users with output here.

PR #1138, opened 2024-03-17
Don’t fail rename if there are multiple index entries for the same USR

ahoppen:
The test case that I’m adding right now is that we have a lingering index entry from a file that has been moved on disk. While the proper fix here would be to ignore such an entry from the index, we should also be more resilient and not fail rename altogether just because there are multiple entries for a USR – the chances are very high that they all translate to the same Swift and Objective-C name.

rdar://123937371

### DIFF
--- a/Sources/SourceKitLSP/Rename.swift
+++ b/Sources/SourceKitLSP/Rename.swift
@@ -554,12 +554,14 @@ extension SourceKitServer {
     index: IndexStoreDB
   ) async throws -> CrossLanguageName? {
     let definitions = index.occurrences(ofUSR: usr, roles: [.definition])
-    guard let definitionSymbol = definitions.only else {
-      if definitions.isEmpty {
-        logger.error("no definitions for \(usr) found")
-      } else {
-        logger.error("Multiple definitions for \(usr) found")
-      }
+    if definitions.count > 1 {
+      logger.log("Multiple definitions for \(usr) found")
+    }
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

Sources/SourceKitLSP/Rename.swift
```diff
@@ -554,12 +554,14 @@ extension SourceKitServer {
...
+    if definitions.count > 1 {
+      logger.log("Multiple definitions for \(usr) found")
+    }
+    // There might be multiple definitions of the same symbol eg. in different `#if` branches. In this case pick any of
+    // them because with very high likelihood they all translate to the same clang and Swift name. Sort the entries to
+    // ensure that we deterministically pick the same entry every time.
+    guard let definitionSymbol = definitions.sorted().first else {
+      logger.error("no definitions for \(usr) found")
       return nil
     }
```
bnbarham, line 566: Could try the next if the picked definition isn't able to be translated?

bnbarham (approved):

Sources/SourceKitLSP/Rename.swift
ahoppen: Good idea 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1144, opened 2024-03-22
Remove --sync option

ahoppen:
With the `test-sourcekit-lsp.py` integration test rewritten to not use `--sync`, there are no more users of this option. Remove it since we don’t really test it anyway.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1148, opened 2024-03-27
Implement a syntactic test discovery fallback for XCTests written in Swift

ahoppen:
If the index for a given file is not up-to-date, perform a syntactic scan for tests within it.

This should allow editors to run the `textDocument/tests` request while the user is editing a file to scan it for test cases.

### DIFF
--- a/Sources/SourceKitLSP/IndexOutOfDateChecker.swift
+++ b/Sources/SourceKitLSP/IndexOutOfDateChecker.swift
@@ -1,0 +1,101 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import IndexStoreDB
+import LSPLogging
+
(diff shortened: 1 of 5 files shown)


ahoppen:
https://github.com/apple/indexstore-db/pull/185

@swift-ci Please test

Sources/SourceKitLSP/IndexOutOfDateChecker.swift
```diff
@@ -0,0 +1,101 @@
...
+      let sourceFileModificationDate = try modificationDate(of: symbolLocation.path)
+      switch sourceFileModificationDate {
+      case .fileDoesNotExist:
+        return false
+      case .date(let sourceFileModificationDate):
+        return sourceFileModificationDate <= symbolLocation.timestamp
+      }
+    } catch {
+      logger.fault("Unable to determine if SymbolLocation is up-to-date: \(error.forLogging)")
+      return true
```
bnbarham, line 77: The only error at the moment is no mtime in attributes. Seems like we should just assume it isn't up to date in that case? I doubt it really matters though.

bnbarham (approved):

Sources/SourceKitLSP/IndexOutOfDateChecker.swift
ahoppen: My thinking was that if sourcekit-lsp is running on some platform that for some reason doesn’t have modification time or if Foundation doesn’t know how to extract it (not sure if those exist), it’s better to assume all files are up-to-date than to completely ignore index results. But as you said, I don’t think it really makes a difference.

bnbarham: Fair enough, that's fine with me 👍

PR #1149, opened 2024-03-28
Return tests discovered by `workspace/tests` and `textDocument/tests` in a hierarchical format

ahoppen:
This ways the client doesn’t need to create a hierarchical structure using the container names. It is also more flexible and allows nesting of test suites + the addition of labels and tags for swift-testing.

The data structure for `TestItem` has been heavily inspired by VS Code’s `TestItem` for the test explorer, which should make it fairly straightforward to integrate these results into the VS Code test explorer.  

A known limitation is that the test IDs are currently not qualified by the module name. I’m planning to implement this soon via a follow-up PR (https://github.com/apple/sourcekit-lsp/issues/1150).

<details><summary>Definition of TestItem in VS Code</summary>
<p>

```typescript
/**
 * An item shown in the "test explorer" view.
 *
 * A `TestItem` can represent either a test suite or a test itself, since
 * they both have similar capabilities.
 */
export interface TestItem {
    /**
     * Identifier for the `TestItem`. This is used to correlate
     * test results and tests in the document with those in the workspace
     * (test explorer). This cannot change for the lifetime of the `TestItem`,
     * and must be unique among its parent's direct children.
     */
    readonly id: string;

    /**
     * URI this `TestItem` is associated with. May be a file or directory.
     */
    readonly uri: Uri | undefined;

    /**
     * The children of this test item. For a test suite, this may contain the
     * individual test cases or nested suites.
     */
    readonly children: TestItemCollection;

    /**
     * The parent of this item. It's set automatically, and is undefined
     * top-level items in the {@link TestController.items} and for items that
     * aren't yet included in another item's {@link TestItem.children children}.
     */
    readonly parent: TestItem | undefined;

    /**
     * Tags associated with this test item. May be used in combination with
     * {@link TestRunProfile.tags}, or simply as an organizational feature.
     */
    tags: readonly TestTag[];

    /**
     * Indicates whether this test item may have children discovered by resolving.
     *
     * If true, this item is shown as expandable in the Test Explorer view and
     * expanding the item will cause {@link TestController.resolveHandler}
     * to be invoked with the item.
     *
     * Default to `false`.
     */
    canResolveChildren: boolean;

    /**
     * Controls whether the item is shown as "busy" in the Test Explorer view.
     * This is useful for showing status while discovering children.
     *
     * Defaults to `false`.
     */
    busy: boolean;

    /**
     * Display name describing the test case.
     */
    label: string;

    /**
     * Optional description that appears next to the label.
     */
    description?: string;

    /**
     * A string that should be used when comparing this item
     * with other items. When `falsy` the {@link TestItem.label label}
     * is used.
     */
    sortText?: string | undefined;

    /**
     * Location of the test item in its {@link TestItem.uri uri}.
     *
     * This is only meaningful if the `uri` points to a file.
     */
    range: Range | undefined;

    /**
     * Optional error encountered while loading the test.
     *
     * Note that this is not a test result and should only be used to represent errors in
     * test discovery, such as syntax errors.
     */
    error: string | MarkdownString | undefined;
}

/**
 * Tags can be associated with {@link TestItem TestItems} and
 * {@link TestRunProfile TestRunProfiles}. A profile with a tag can only
 * execute tests that include that tag in their {@link TestItem.tags} array.
 */
export class TestTag {
    /**
     * ID of the test tag. `TestTag` instances with the same ID are considered
     * to be identical.
     */
    readonly id: string;

    /**
     * Creates a new TestTag instance.
     * @param id ID of the test tag.
     */
    constructor(id: string);
}

```

</p>
</details>

### DIFF
--- a/Sources/LanguageServerProtocol/CMakeLists.txt
+++ b/Sources/LanguageServerProtocol/CMakeLists.txt
@@ -124,6 +124,7 @@ add_library(LanguageServerProtocol STATIC
   SupportTypes/SKCompletionOptions.swift
   SupportTypes/StringOrMarkupContent.swift
   SupportTypes/SymbolKind.swift
+  SupportTypes/TestItem.swift
   SupportTypes/TextDocumentContentChangeEvent.swift
   SupportTypes/TextDocumentEdit.swift
   SupportTypes/TextDocumentIdentifier.swift
--- a/Sources/LanguageServerProtocol/Requests/DocumentTestsRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/DocumentTestsRequest.swift
@@ -15,7 +15,7 @@
 /// **(LSP Extension)**
 public struct DocumentTestsRequest: TextDocumentRequest, Hashable {
   public static let method: String = "textDocument/tests"
-  public typealias Response = [WorkspaceSymbolItem]?
+  public typealias Response = [TestItem]
 
   public var textDocument: TextDocumentIdentifier
 
--- a/Sources/LanguageServerProtocol/Requests/WorkspaceTestsRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/WorkspaceTestsRequest.swift
@@ -15,7 +15,7 @@
 /// **(LSP Extension)**
 public struct WorkspaceTestsRequest: RequestType, Hashable {
   public static let method: String = "workspace/tests"
-  public typealias Response = [WorkspaceSymbolItem]?
+  public typealias Response = [TestItem]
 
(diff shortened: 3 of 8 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1151, opened 2024-03-28
Implement test discovery for swift-testing tests for the `textDocument/tests` request

ahoppen:
This allows us to return swift-testing tests within a single document. It does not look for swift-testing tests workspace-wide (the `workspace/tests` request), which will be a follow-up PR.

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
@@ -1,0 +1,378 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+import SwiftSyntax
+
+// MARK: - Attribute parsing
+
+/// Get the traits applied to a testing attribute.
+///
+/// - Parameters:
+///   - testAttribute: The attribute to inspect.
+///
+/// - Returns: An array of `ExprSyntax` instances representing the traits
+///   applied to `testAttribute`. If the attribute has no traits, the empty
+///   array is returned.
+private func traits(ofTestAttribute testAttribute: AttributeSyntax) -> [ExprSyntax] {
+  guard let argument = testAttribute.arguments, case let .argumentList(argumentList) = argument else {
+    return []
+  }
+
+  // Skip the display name if present.
+  var traitArgumentsRange = argumentList.startIndex..<argumentList.endIndex
+  if let firstArgument = argumentList.first,
+    firstArgument.label == nil,
+    firstArgument.expression.is(StringLiteralExprSyntax.self)
+  {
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):
+ @grynspan and @stmontgomery for the swift-testing scanning.

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
```diff
@@ -0,0 +1,378 @@
...
+        // Ignore disabled traits which have a trailing closure since
+        // they're conditional.
+        if functionCall.trailingClosure != nil {
+          return false
+        }
+
+        return true
+      }
+
+    self.isHidden = traitArguments.lazy
```
grynspan, line 129: We probably don't need to scan for `.hidden`: it's an internal-only trait we use for our test fixtures, not one external developers can use.

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
```diff
@@ -0,0 +1,378 @@
...
+  /// For example, when scanning for tests inside `Bar` in the following, this is `["Foo", "Bar"]`
+  ///
+  /// ```swift
+  /// struct Foo {
+  ///   struct Bar {
+  ///     @Test func myTest() {}
+  ///   }
+  /// }
+  /// ```
+  private let parentTypeNames: [String]
```
grynspan, line 166: What happens if I say:

```swift
extension Foo.Bar {
  @Test func myTest() {}
}
```

?

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
```diff
@@ -0,0 +1,378 @@
...
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let visitor = SyntacticSwiftTestingTestScanner(snapshot: snapshot, allTestsDisabled: false, parentTypeNames: [])
+    visitor.walk(syntaxTree)
+    return visitor.result
+  }
+
+  /// Visit a class/struct/... or extension declaration.
+  ///
+  /// `typeNames` is the name of the class struct or, if this is an extension, an array containing the components of the
+  /// extended type. For example, `extension Foo.Bar {}` passes `["Foo", "Bar"]` as `typeNames`.
```
grynspan, line 192: _oic_

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -15,6 +15,11 @@ import LSPLogging
 import LanguageServerProtocol
 import SwiftSyntax
 
+public enum TestStyle {
```
grynspan, line 18: We call this `Library` in SwiftPM, FWIW. Consistency may be a good thing.

ahoppen: I was hoping that we could maybe upstream this request to the LSP spec, at which point it would become language-agnostic. And at that point, I think it’s better to have a general term like *Style* instead of *Library*, which has more implications about build/etc structure.

grynspan: I don't feel strongly about either term. :)

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
ahoppen: If we decide to not support `.hidden`, I’ll change it in a follow-up PR to unblock https://github.com/apple/sourcekit-lsp/pull/1175

PR #1153, opened 2024-03-29
Inherit actor isolation of closure passed to `orLog`

ahoppen:
This allows us to express that `body` will run on the same actor isolation domain as the caller of `orLog`, which effectively makes `orLog` usable from actors again.


ahoppen:
@swift-ci Please test

PR #1158, opened 2024-04-04
Adjustment for the removal of `IncrementalEdit(offset:length:replacement:)` 🤝 swift-syntax#2588

ahoppen:
This initializer was only introduced very recently since `IncrementalEdit` stores the replacement bytes instead of just the replacement length. We are now changing `IncrementalEdit` store the range it’s replacing as `Range<AbsolutePosition>` instead of `ByteSourceRange`.

Companion of https://github.com/apple/swift-syntax/pull/2588


ahoppen:
https://github.com/apple/swift-syntax/pull/2588

@swift-ci Please test

bnbarham (approved):

PR #1161, opened 2024-04-04
Unify logging of errors during position conversions

ahoppen:
Instead of logging errors in position translation ad-hoc at the caller’s side (and ofter forgetting to do so), log these errors in `LineTable`. To be able to debug where the position conversion error is coming from, also log the file name and line number of the caller.

### DIFF
--- a/Sources/SKSupport/LineTable.swift
+++ b/Sources/SKSupport/LineTable.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import LSPLogging
+
 public struct LineTable: Hashable {
   @usableFromInline
   var impl: [String.Index]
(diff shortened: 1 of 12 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKSupport/LineTable.swift
```diff
@@ -123,113 +125,242 @@ extension LineTable {
...
+    line: Int,
+    utf16Column: Int,
+    callerFile: StaticString = #fileID,
+    callerLine: UInt = #line
+  ) -> String.Index? {
     guard line < count else {
-      // Line out of range.
+      logger.fault(
+        """
+        Unable to get string index for \(line):\(utf16Column) because line is out of range \
```
bnbarham, line 150: Probably obvious enough from the caller, but could help to add whether it's utf8 vs utf16 in all these messages. It's not super necessary for the line case, so maybe just add `utf<n>` in front of `column` for the column cases?

bnbarham (approved):
Much cleaner 👍

Sources/SKSupport/LineTable.swift
ahoppen: Good idea 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1163, opened 2024-04-04
Make sure the range returned by `PrepareRenameRequest` always contains the position from which the rename was initiated

ahoppen:
VS Code rejects the placeholder name returned by the `PrepareRenameRequest`. When initiating rename on `x` in the following, this means that VS Code picks its own internal symbol name as the placeholder text, opening the rename textbox with `x`.

```swift
func foo(
  x: Int
) {}
```

Users then enter `y` in the expectation to rename only the parameter but sourcekit-lsp expects a function name and thus renames `foo` to `y(x:)`, which is unexpected.

rdar://125551489


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1165, opened 2024-04-04
Show container name as the detail of a call hierarchy item

ahoppen:
The container name, showing the class a method is defined on, is more useful than the module name.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2039,14 +2039,14 @@ extension SourceKitLSPServer {
 
   private func indexToLSPCallHierarchyItem(
     symbol: Symbol,
-    moduleName: String?,
+    containerName: String?,
     location: Location
   ) -> CallHierarchyItem {
     CallHierarchyItem(
       name: symbol.name,
       kind: symbol.kind.asLspSymbolKind(),
       tags: nil,
-      detail: moduleName,
+      detail: containerName,
       uri: location.uri,
       range: location.range,
       selectionRange: location.range,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2472,14 +2472,28 @@ extension IndexSymbolKind {
   }
 }
 
+fileprivate extension IndexSymbolKind {
+  /// Whether this symbol should be considered as a container in `SymbolOccurrence.containerName`
+  var isContainer: Bool {
```
bnbarham, line 2477: Feels like this should have something to do with type in its name (though also includes module/namespace...) given it doesn't include functions. Same with `containerName`.

Also `union` should probably be included in the `true` case either way.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2143,7 +2143,7 @@ extension SourceKitLSPServer {
           return CallHierarchyIncomingCall(
             from: indexToLSPCallHierarchyItem(
               symbol: related.symbol,
-              moduleName: definitionSymbolLocation?.moduleName,
+              containerName: definition?.containerName,
```
bnbarham, line 2146: Should module be the fallback if there is no container?

Sources/SourceKitLSP/SourceKitLSPServer.swift
hamishknight: Would probably be best to avoid using the word `container` too, to avoid confusion with the `containedBy` relation. Maybe something like `hasChildren`?

ahoppen: Really this is just: Should this be considered for `containerName` and I think it’s actually easier to just have the `switch` inside the `filter` of `containerName`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: No, I think it’s better to leave the `containerName` empty for top-level functions. That way it’s also easier to tell top-level functions apart from member functions.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #1166, opened 2024-04-04
Don’t repeat a function in `incomingCalls` if it contains multiple calls to the same function

ahoppen:
Eg. if we have the following, and we get the call hierarchy of `foo`, we only want to show `bar` once, with multiple `fromRanges` instead of having two entries for `bar` in the call hierarchy.

```swift
func foo() {}
func bar() {
  foo()
  foo()
}
```

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2126,27 +2126,44 @@ extension SourceKitLSPServer {
     callableUsrs += index.occurrences(ofUSR: data.usr, roles: .overrideOf).flatMap { occurrence in
       occurrence.relations.filter { $0.roles.contains(.overrideOf) }.map(\.symbol.usr)
     }
+    // callOccurrences are all the places that any of the USRs in callableUsrs is called.
+    // We also load the `calledBy` roles to get the method that contains the reference to this call.
     let callOccurrences = callableUsrs.flatMap { index.occurrences(ofUSR: $0, roles: .calledBy) }
-    let calls = callOccurrences.flatMap { occurrence -> [CallHierarchyIncomingCall] in
-      guard let location = indexToLSPLocation(occurrence.location) else {
-        return []
-      }
-      return occurrence.relations.filter { $0.symbol.kind.isCallable }
-        .map { related in
-          // Resolve the caller's definition to find its location
-          let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: related.symbol.usr)
-          let definitionSymbolLocation = definition?.location
-          let definitionLocation = definitionSymbolLocation.flatMap(indexToLSPLocation)
 
-          return CallHierarchyIncomingCall(
-            from: indexToLSPCallHierarchyItem(
-              symbol: related.symbol,
-              containerName: definition?.containerName,
-              location: definitionLocation ?? location  // Use occurrence location as fallback
-            ),
-            fromRanges: [location.range]
-          )
-        }
+    // Maps functions that call a USR in `callableUSRs` to all the called occurrences of `callableUSRs` within the
+    // function. If a function `foo` calls `bar` multiple times, `callersToCalls[foo]` will contain two call
+    // `SymbolOccurrence`s.
+    // This way, we can group multiple calls to `bar` within `foo` to a single item with multiple `fromRanges`.
+    var callersToCalls: [Symbol: [SymbolOccurrence]] = [:]
+
+    for call in callOccurrences {
+      // Callers are all `calledBy` relations of a call to a USR in `callableUsrs`, ie. all the functions that contain a
+      // call to a USR in callableUSRs. In practice, this should always be a single item.
+      let callers = call.relations.filter { $0.roles.contains(.calledBy) }.map(\.symbol)
+      for caller in callers {
+        callersToCalls[caller, default: []].append(call)
+      }
+    }
+
+    let calls = callersToCalls.compactMap { (caller: Symbol, calls: [SymbolOccurrence]) -> CallHierarchyIncomingCall? in
+      // Resolve the caller's definition to find its location
+      let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: caller.usr)
+      let definitionSymbolLocation = definition?.location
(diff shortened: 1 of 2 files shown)


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2128,27 +2128,44 @@ extension SourceKitLSPServer {
     callableUsrs += index.occurrences(ofUSR: data.usr, roles: .overrideOf).flatMap { occurrence in
       occurrence.relations.filter { $0.roles.contains(.overrideOf) }.map(\.symbol.usr)
     }
+    // callOccurrences are all the places that any of the USRs in callableUsrs is called.
+    // We also load the `calledBy` roles to get the method that contains the reference to this call.
```
bnbarham, line 2132: Sort of unrelated to this PR, but it'd be good to add a bunch of test cases for properties:
```
struct Foo {
  var member: Int = {
    return bar()
  }();

  func bar() {}
}
```

Call hierarchy for `bar` probably should give `member` there. `member` isn't *really* the caller, but... I think it's what people would expect to see in the UI.

Also just a regular getter:
```
struct Foo {
  var member: Int {
    return bar()
  }

  func bar() {}
}
```

Plus explicit getter/setter/etc:
```
struct Foo {
  var member: Int {
    get {
      return bar()
    }
  }

  func bar() {}
}
```

Also CC @hamishknight as he's been looking at this a bit recently (both to comment on the above and also a review of this PR in general).

ahoppen: We do already have a test for valuable getters FWIW: https://github.com/apple/sourcekit-lsp/blob/4ed62392e7b4f889d44c09e4d6857a0eb28975c3/Tests/SourceKitLSPTests/CallHierarchyTests.swift#L414-L452

I agree that the rest would be good to test

bnbarham: Extending that with a use of `foo` seems like a good thing to do as well.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2128,27 +2128,44 @@ extension SourceKitLSPServer {
...
+    // Maps functions that call a USR in `callableUSRs` to all the called occurrences of `callableUSRs` within the
+    // function. If a function `foo` calls `bar` multiple times, `callersToCalls[foo]` will contain two call
+    // `SymbolOccurrence`s.
+    // This way, we can group multiple calls to `bar` within `foo` to a single item with multiple `fromRanges`.
+    var callersToCalls: [Symbol: [SymbolOccurrence]] = [:]
+
+    for call in callOccurrences {
+      // Callers are all `calledBy` relations of a call to a USR in `callableUsrs`, ie. all the functions that contain a
+      // call to a USR in callableUSRs. In practice, this should always be a single item.
+      let callers = call.relations.filter { $0.symbol.kind.isCallable }.map(\.symbol)
```
hamishknight, line 2144: I think it would be better to check the role here, with https://github.com/apple/swift/pull/72930 it ought to be sufficient to check for `.containedBy`, which would handle the case where the parent is a property (note that `isCallable` would be `false` in that case). Though for compatibility with older toolchains it may be better to check if the role contains `.calledBy`, and if that's empty, fallback to checking for `.containedBy`.

ahoppen: Good suggestion with checking `$0.roles`.

There is a subtle difference between `calledBy` and `containedBy` relations.

```swift
func foo() {}

var testVar: Int {
  get {
    let myVar = foo()
    return 2
  }
}

func test() {
  _ = testVar
}
```

The call to `foo` in `let myVar = foo()` is called by `s:4test0A3VarSivg` (demangled `test.testVar.getter : Swift.Int`, display name `getter:testVar`) while it is contained by `s:4test0A3VarSivg02myB0L_ytvp` (demangled `myVar #1 : () in test.testVar.getter : Swift.Int` / display name `testVar`). But `s:4test0A3VarSivg02myB0L_ytvp` doesn’t have any callers while `s:4test0A3VarSivg` does show the call inside `test()`. So I’ll go with just checking the `calledBy` relation. Or do you see any advantage of `containedBy` over `calledBy`?

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
hamishknight: > while it is contained by `s:4test0A3VarSivg02myB0L_ytvp`

That should have been fixed by https://github.com/apple/swift/pull/72930, is it still an issue? The main advantage of `containedBy` is that it works for cases like:

```swift
struct Foo {
  var member: Int = {
    return bar()
  }()

  func bar() {}
}
```

Even though `member` isn't a callable symbol, it probably makes sense to consider it the caller.

ahoppen: Oh, right, sorry, you mentioned that. I was using an open source toolchain snapshot.

Let me try again once use have a toolchain snapshot with https://github.com/apple/swift/pull/72930 because I can’t be bothered to build a toolchain myself.

bnbarham: That seems like a bug to me, the container there should be the same as the calledBy relation. The main advantage is that it handles one of the cases I sent above, ie.
```
func foo() -> Int {}

var testVar: Int = {
  foo()
}()
```

`foo` isn't really called by `testVar`, but IMO it's what a user would expect to see. Alternatively maybe we should have a `calledBy` to the init, but then that could also be very confusing. If we do the former I don't think we should continue the call hierarchy (though open to opinions there). The latter would make sense to continue though.

bnbarham: :( I wish GitHub would tell me to refresh.

bnbarham: Should we stop the traversal if we use the container instead? Does it make sense to continue from `member` there? Or should we have a calledBy to the `init`?

ahoppen:
@swift-ci Please test Linux

PR #1170, opened 2024-04-08
Fix a race condition during the computation of `uriToWorkspaceCache`

ahoppen:
See comment in `workspaceForDocument` that explains the race.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -408,6 +408,10 @@ public actor SourceKitLSPServer {
   /// request's task before handling any cancellation request for it.
   private let cancellationMessageHandlingQueue = AsyncQueue<Serial>()
 
+  /// The queue on which all modifications of `uriToWorkspaceCache` happen. This means that the value of
+  /// `workspacesAndIsImplicit` and `uriToWorkspaceCache` can't change while executing a closure on `workspaceQueue`.
+  private let workspaceQueue = AsyncQueue<Serial>()
+
   /// The connection to the editor.
   public let client: Connection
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SKSupport/AsyncQueue.swift
```diff
@@ -169,6 +169,15 @@ extension AsyncQueue where TaskMetadata == Serial {
     return self.async(priority: priority, metadata: Serial(), operation: operation)
   }
 
+  /// Executes the given `operation` and awaits its result.
+  @discardableResult
+  public func sync<Success: Sendable>(
```
bnbarham, line 174:
```suggestion
  public func asyncResult<Success: Sendable>(
```
Maybe? Seems weird to call it `sync` since it is really still async.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -517,54 +528,59 @@ public actor SourceKitLSPServer {
...
-      if fileHandlingCapability > bestWorkspace.fileHandlingCapability {
-        bestWorkspace = (workspace, fileHandlingCapability)
+    // Execute the computation of the workspace on `workspaceQueue` to ensure that the file handling capabilities of the
+    // workspaces don't change during the computation. Otherwise, we could run into a race condition like the following:
+    //  1. We don't have an entry for file `a.swift` in `uriToWorkspaceCache` and start the computation
+    //  2. We find that the first workspace in `self.workspaces` can handle this file.
+    //  3. During the `await ... .fileHandlingCapability` for a second workspace the file handling capabilities for the
+    //    first workspace change, meaning it can no longer handle the document. This resets `uriToWorkspaceCache`
+    //    assuming that the URI to workspace relation will get re-computed.
+    //  4. But we then set `uriToWorkspaceCache[uri]` to the workspace found in step (2), caching in out-of-date result.
```
bnbarham, line 542:
```suggestion
    //  4. But we then set `uriToWorkspaceCache[uri]` to the workspace found in step (2), caching an out-of-date result.
```

bnbarham (approved):

Sources/SKSupport/AsyncQueue.swift
ahoppen: I removed it and added `await <xxx>.value` instead. Turns out that one of the three calls benefited from using `await <xxx>.valuePropagatingCancellation`

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1176, opened 2024-04-16
Set time zone to the user's local time zone when using date formatters

ahoppen:
Otherwise, the date formatters always output the time in UTC, which has confused me multiple times. Setting the user’s local time zone will emit the date in the user’s local time zone and include the time zone offset to GMT in the date as well.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1178, opened 2024-04-16
Instead of returning an empty array in `prepareTypeHierarchy`, sourcekit-lsp should return `nil`

ahoppen:
Eg. when requesting type hierarchy of a class when the project hasn’t been built, sourcekit-lsp returns an empty array. That causes VS Code to fail with very ambiguous error messages
- MISSING provider
- Cannot read properties of null (reading 'kind')

To work around this, instead of returning an empty array, return `nil` instead.

rdar://126228814

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2291,6 +2291,13 @@ extension SourceKitLSPServer {
     }
     .sorted(by: { $0.name < $1.name })
 
+    if typeHierarchyItems.isEmpty {
+      // When returning an empty array, VS Code fails with the following two errors. Returning `nil` works around those
+      // VS Code-internal errors showing up
+      //  - MISSING provider
+      //  - Cannot read properties of null (reading 'kind')
+      return nil
+    }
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2291,6 +2291,13 @@ extension SourceKitLSPServer {
     }
     .sorted(by: { $0.name < $1.name })
 
+    if typeHierarchyItems.isEmpty {
+      // When returning an empty array, VS Code fails with the following two errors. Returning `nil` works around those
```
bnbarham, line 2295: I assume this is because it always expects at least the original element itself? Does call hierarchy have the same issue?

bnbarham (approved):

dschaefer2:
You're ahead of me. :). Saw that today too. Thanks!

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Call hierarchy doesn’t seem to have that issue.

PR #1179, opened 2024-04-17
Introduce new refactoring code actions based on the Swift syntax tree.

DougGregor:
This change includes a number of new refactoring code actions that build on the syntax refactorings for the SwiftRefactor module of swift-syntax:

  * Add digit separators to an integer literal, e.g., `1000000` -> `1_000_000`.
  * Remove digit separators from an integer literal, e.g., 1_000_000 -> 1000000.
  * Format a raw string literal, e.g., `"Hello \#(world)"` -> `##"Hello\#(world)"##`
  * Migrate to new if let syntax, e.g., `if let x = x { ... }` -> `if let x { ... }`
  * Replace opaque parameters with generic parameters, e.g., `func f(p: some P)` --> `func f<T1: P>(p: T1)`.

This is generally easy to do, requiring one conformance to provide a name for the refactoring:

    extension AddSeparatorsToIntegerLiteral: SyntaxRefactoringCodeActionProvider {
      public static var title: String { "Add digit separators" }
    }

### DIFF
--- a/Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
@@ -1,0 +1,76 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SwiftRefactor
+import SwiftSyntax
+
+// TODO: Make the type IntegerLiteralExprSyntax.Radix conform to CaseEnumerable
+// in swift-syntax.
+
+extension IntegerLiteralExprSyntax.Radix {
+  static var allCases: [Self] = [.binary, .octal, .decimal, .hex]
+}
+
+/// Syntactic code action provider to convert integer literals between
+/// different bases.
+struct ConvertIntegerLiteral: SyntaxCodeActionProvider {
+  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard
+      let token = scope.firstToken,
+      let integerExpr = token.parent?.as(IntegerLiteralExprSyntax.self),
+      let integerValue = Int(
+        integerExpr.split().value.filter { $0 != "_" },
+        radix: integerExpr.radix.size
+      )
+    else {
+      return []
+    }
+
+    var actions = [CodeAction]()
+    let currentRadix = integerExpr.radix
+    for radix in IntegerLiteralExprSyntax.Radix.allCases {
+      guard radix != currentRadix else {
+        continue
+      }
+
+      //TODO: Add this to swift-syntax?
+      let prefix: String
+      switch radix {
+      case .binary:
+        prefix = "0b"
+      case .octal:
+        prefix = "0o"
+      case .hex:
+        prefix = "0x"
+      case .decimal:
+        prefix = ""
+      }
+
+      let convertedValue: ExprSyntax =
+        "\(raw: prefix)\(raw: String(integerValue, radix: radix.size))"
+      let edit = TextEdit(
+        range: scope.snapshot.range(of: integerExpr),
+        newText: convertedValue.description
+      )
+      actions.append(
+        CodeAction(
+          title: "Convert \(integerExpr) to \(convertedValue)",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(changes: [scope.snapshot.uri: [edit]])
+        )
+      )
+    }
+
+    return actions
+  }
+}
--- a/Sources/SourceKitLSP/Swift/CodeActions/SyntaxCodeActionProvider.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/SyntaxCodeActionProvider.swift
@@ -1,0 +1,65 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+import SwiftRefactor
+import SwiftSyntax
+
+/// Describes types that provide one or more code actions based on purely
+/// syntactic information.
+protocol SyntaxCodeActionProvider {
+  /// Produce code actions within the given scope. Each code action
+  /// corresponds to one syntactic transformation that can be performed, such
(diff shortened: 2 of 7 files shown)


DougGregor:
@swift-ci please test

DougGregor:
Speculatively merging, and can address comments in subsequent PRs

Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
```diff
@@ -0,0 +1,76 @@
...
+import LanguageServerProtocol
+import SwiftRefactor
+import SwiftSyntax
+
+// TODO: Make the type IntegerLiteralExprSyntax.Radix conform to CaseEnumerable
+// in swift-syntax.
+
+extension IntegerLiteralExprSyntax.Radix {
+  static var allCases: [Self] = [.binary, .octal, .decimal, .hex]
+}
```
ahoppen, line 22: As I commented in my other PR, I would prefer to just do this instead of having a FIXME in the codebase.

Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
```diff
@@ -0,0 +1,76 @@
...
+      switch radix {
+      case .binary:
+        prefix = "0b"
+      case .octal:
+        prefix = "0o"
+      case .hex:
+        prefix = "0x"
+      case .decimal:
+        prefix = ""
+      }
```
ahoppen, line 57: Similar here, it would be good to actually sink this to swift-syntax.

Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
```diff
@@ -0,0 +1,76 @@
...
+// TODO: Make the type IntegerLiteralExprSyntax.Radix conform to CaseEnumerable
+// in swift-syntax.
+
+extension IntegerLiteralExprSyntax.Radix {
+  static var allCases: [Self] = [.binary, .octal, .decimal, .hex]
+}
+
+/// Syntactic code action provider to convert integer literals between
+/// different bases.
+struct ConvertIntegerLiteral: SyntaxCodeActionProvider {
```
ahoppen, line 26: Could you add a test case for this, checking that the refactoring do actually show up when retrieving code actions? Ie. a test in https://github.com/apple/sourcekit-lsp/blob/main/Tests/SourceKitLSPTests/CodeActionTests.swift

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -699,22 +699,31 @@ extension SwiftLanguageService {
...
       codeActions: codeActions,
       clientCapabilities: codeActionCapabilities
     )
     return response
   }
 
-  func retrieveCodeActions(_ req: CodeActionRequest, providers: [CodeActionProvider]) async throws -> [CodeAction] {
+  func retrieveCodeActions(_ req: CodeActionRequest, providers: [CodeActionProvider]) async throws
+    -> [CodeAction]
+  {
```
ahoppen, line 726: I’d prefer to format this as

```swift
  func retrieveCodeActions(
    _ req: CodeActionRequest, 
    providers: [CodeActionProvider]
  ) async throws -> [CodeAction]
  {
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -1091,6 +1111,18 @@ extension DocumentSnapshot {
...
 
+  /// Extracts the range of the given syntax node in terms of positions within
+  /// this source file.
+  func range(
+    of node: some SyntaxProtocol,
+    callerFile: StaticString = #fileID,
+    callerLine: UInt = #line
+  ) -> Range<Position> {
+    let lowerBound = self.position(of: node.position)
+    let upperBound = self.position(of: node.endPosition)
```
ahoppen, line 1122: I think you should pass `callerFile` and `callerLine` into `position(of:)` so that the logger fault message contains the file and line number that is the real originator if position conversion fails.

Sources/SourceKitLSP/Swift/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -0,0 +1,65 @@
...
+    self.snapshot = snapshot
+    self.request = request
+    self.file = tree
+
+    let start = snapshot.absolutePosition(of: request.range.lowerBound)
+    let end = snapshot.absolutePosition(of: request.range.upperBound)
+    let left = file.token(at: start)
+    let right = file.token(at: end)
+    let leftOff = left?.position ?? AbsolutePosition(utf8Offset: 0)
+    let rightOff = right?.endPosition ?? leftOff
```
ahoppen, line 57: I’d prefer to fail the creation of `SyntaxCodeActionScope` and emit a `logger.fault` if we can’t find a token at a specific position. It’s the kind of thing where things start failing and then it’s hard to track down what’s going wrong if there’s no logging.

Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -0,0 +1,75 @@
...
+  static var title: String { get }
+}
+
+/// SyntaxCodeActionProviders with a \c Void context can automatically be
+/// adapted provide a code action based on their refactoring operation.
+extension SyntaxRefactoringCodeActionProvider where Self.Context == Void {
+  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard
+      let token = scope.firstToken,
+      let node = token.parent?.as(Input.self)
```
ahoppen, line 29: Should we traverse the parents until we find a node of type `Input`? That way you don’t need to start eg. a `MigrateToNewIfLetSyntax` refactoring on the `if` token but you could also start it from within the condition or the body (not sure if we want the latter though, maybe we should stop the traversal at some pre-defined nodes like `CodeBlockItemListSyntax` and `MemberBlockItemListSyntax` to prevent that).

Similarly, the `OpaqueParameterToGeneric` refactoring action can’t currently be invoked from one of the `some` types, which might be confusing.

ahoppen (commented):
Thanks for bridging this gap, Doug! Great to see these refactorings in sourcekit-lsp. I think a few tests to check that the refactorings actually do show up would be good though.

Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
DougGregor: https://github.com/apple/swift-syntax/pull/2626

Sources/SourceKitLSP/Swift/CodeActions/ConvertIntegerLiteral.swift
DougGregor: yeah, it's here: https://github.com/apple/swift-syntax/pull/2626

Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
DougGregor: I'm wondering if this is something that should be handled by the `MigrateToNewIfLetSyntax`, `OpaqueParameterToGeneric`, etc., refactorings themselves, because we don't necessarily know how far out in the tree we should go.

ahoppen: That seems reasonable.

PR #1180, opened 2024-04-17
Reload a file when other files within the same module or a `.swiftmodule` file has been changed

### LINKED ISSUE #620: Jump to definition requires a reload after build

### LINKED ISSUE #1116: Diagnostics are not refreshed after build when a new file is added
Steps to reproduce:
1. Add a new file with some class.
2. Reference this class within the same target.
3. Build the project using `xcodebuild` command.

Actual result:
The build succeeds, LSP doesn't print any errors in the log, LSP still shows issues:
image

Expected result:
LSP should be automatically refreshed and errors should disappear.

Workaround:
Restart LSP (in Neovim `:LspRestart`).

Version:
Sourcekit-lsp provided by Xcode Version 15.2 (15C500b)

Related issue:
https://github.com/wojciech-kulik/xcodebuild.nvim/issues/56

Notes:
I reproduced it both in Neovim and Visual Studio Code, so I guess it's not the LSP client's fault.

ahoppen:
When the client sends us `workspace/didChangeWatchedFiles` notification of an updated `.swift` file, we should refresh the other open files in that module since they might be referencing functions from that updated file.

If a `.swiftmodule` file has been updated, we refresh all the files within the package since they might import that module. Technically, we would only need to refresh files that are in module that are downstream of the updated module but we don’t currently have that information easily available from SwiftPM. Also, usually, if the client has a file from a low-level module open, he’ll be working on that module which means that such an optimization won’t help. The real solution here is to wait for us to finish preparation (which we would exactly know when it finishes since sourcekit-lsp would schedule it) but for that we need to implement background preparation.

Fixes #620
Fixes #1116
rdar://99329579
rdar://123971779

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
@@ -99,6 +99,16 @@ public actor SwiftPMBuildSystem {
   /// This callback is informed when `reloadPackage` starts and ends executing.
   var reloadPackageStatusCallback: (ReloadPackageStatus) async -> Void
 
+  /// Debounces calls to `delegate.filesDependenciesUpdated`.
+  ///
+  /// This is to ensure we don't call `filesDependenciesUpdated` for the same file multiple time if the client does not
+  /// debounce `workspace/didChangeWatchedFiles` and sends a separate notification eg. for every file within a target as
+  /// it's being updated by a git checkout, which would cause other files within that target to receive a
+  /// `fileDependenciesUpdated` call once for every updated file within the target.
+  ///
+  /// Force-unwrapped optional because initializing it requires access to `self`.
+  var fileDependenciesUpdatedDebouncer: Debouncer<Set<DocumentURI>>! = nil
+
   /// Creates a build system using the Swift Package Manager, if this workspace is a package.
   ///
   /// - Parameters:
@@ -165,6 +175,19 @@ public actor SwiftPMBuildSystem {
 
     self.modulesGraph = try ModulesGraph(rootPackages: [], dependencies: [], binaryArtifacts: [:])
     self.reloadPackageStatusCallback = reloadPackageStatusCallback
+
+    // The debounce duration of 500ms was chosen arbitrarily without scientific research.
+    self.fileDependenciesUpdatedDebouncer = Debouncer(
+      debounceDuration: .milliseconds(500),
+      combineResults: { $0.union($1) }
+    ) {
+      [weak self] (filesWithUpdatedDependencies) in
+      guard let delegate = await self?.delegate else {
+        logger.fault("Not calling filesDependenciesUpdated because no delegate exists in SwiftPMBuildSystem")
+        return
+      }
+      await delegate.filesDependenciesUpdated(filesWithUpdatedDependencies)
+    }
 
     try await reloadPackage()
   }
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -189,6 +198,17 @@ public actor SwiftLanguageService: LanguageService {
     self.generatedInterfacesPath = options.generatedInterfacesPath.asURL
     try FileManager.default.createDirectory(at: generatedInterfacesPath, withIntermediateDirectories: true)
     self.diagnosticReportManager = nil  // Needed to work around rdar://116221716
+
+    // The debounce duration of 500ms was chosen arbitrarily without scientific research.
```
bnbarham, line 202:
```suggestion
    // The debounce duration of 500ms was chosen arbitrarily without any real benchmarking.
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -339,13 +363,15 @@ extension SwiftLanguageService {
   }
 
   public func documentDependenciesUpdated(_ uri: DocumentURI) async {
-    guard let snapshot = try? self.documentManager.latestSnapshot(uri) else {
-      return
+    // FIXME: Debounce, otherwise it will get called over and over
```
bnbarham, line 366: Any reason not to debounce it now?

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -368,6 +391,35 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+        continue
+      }
+      filesWithUpdatedDependencies.formUnion(target.sources.map { DocumentURI($0) })
+    }
+
+    // If a `.swiftmodule` file is updated, this means that we have performed a build / are
+    // performing a build and files that depend on this module have updated dependencies.
+    // We don't have access to the build graph from the SwiftPM API offered to SourceKit-LSP to figure out which files
+    // depend on the updated module, so assume that all files have updated dependencies. Currently, this does not matter
+    // because `SwiftLanguageService` sends a `dependencyUpdated` request to sourcekitd, which is global anyway.
```
bnbarham, line 413: FWIW that only seems to impact completion, not eg. cursor info.

bnbarham (approved):

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
ahoppen: Oh, this was just an in-development comment that I forgot to remove. We `fileDependenciesUpdated` on the `BuildSystem` level now, so this isn’t needed. I just put it here at some point to remind myself that we need to debounce somewhere.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: Ah, we do also reopen the file. I updated the comment to just remove the

> Currently, this does not matter because `SwiftLanguageService` sends a `dependencyUpdated` request to sourcekitd, which is global anyway.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1183, opened 2024-04-18
Make hover popover prettier

ahoppen:
Essentially two changes:
1. Remove the symbol name from the hover. This name is already included in the documentation
2. Make the `Alternative Result` heading smaller.

Changes it from looking like this
Screenshot 2024-04-17 at 21 41 27
to
Screenshot 2024-04-17 at 21 40 42

rdar://126488098

### DIFF
--- a/Sources/SourceKitLSP/Swift/CommentXML.swift
+++ b/Sources/SourceKitLSP/Swift/CommentXML.swift
@@ -77,7 +77,7 @@ private struct XMLToMarkdown {
       newlineIfNeeded(count: 2)
       out += "```swift\n"
       toMarkdown(node.children)
-      out += "\n```\n\n---\n"
+      out += "\n```\n"
 
     case "Name", "USR", "Direction":
       break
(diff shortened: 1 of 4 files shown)


bnbarham:
Any thoughts on removing the horizontal rule after the declaration name and instead moving it prior to `Alternative result`? Needs something else too, since even with that I think I'd still start reading the bottom doc as the doc for the first result. Just... not sure what.

ahoppen:
I did experiment with horizontal rulers to separate the results quite a bit and was never satisfied. FWIW, here’s how it looks like with a horizontal ruler before `Alternative Result`

Screenshot 2024-04-17 at 22 00 27

And here without any rulers

Screenshot 2024-04-17 at 22 01 22

ahoppen:
And here’s a version that uses `### &nbsp;` to add some spacing between the results (best way I have found to add spacing in VS Code so far).

Screenshot 2024-04-17 at 22 06 48

ahoppen:
Next try. Heading `Multiple Results` if there are multiple results and horizontal rulers between the results.

Screenshot 2024-04-19 at 22 37 48

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham:
Nice, I think that's better. How about H3 still and with a horizontal line after the heading?

ahoppen:
I prefer H2 and no line after the heading. Here’s how H3 and line after the heading looks like.

Screenshot 2024-04-22 at 10 50 35

ahoppen:
@swift-ci Please test

bnbarham:

bnbarham (approved):
I still prefer H3 + line, but either is better than what's there now IMO. So happy to merge either way.

PR #1012, opened 2023-12-26
Refactor SemanticTokenTypes, SemanticTokenModifiers

krzyzanowskim:
Refactor `SyntaxHighlightingToken.Kind` and `SyntaxHighlightingToken.Modifiers` into `SemanticTokenTypes` and `SemanticTokenModifiers`. This change moves types from `SourceKitLSP` to `LanguageServerProtocol` where it belongs.

`SemanticTokenTypes` is extended with additional types in `SyntaxHighlightingToken.swift` that is then used by the server.

Resolves https://github.com/apple/sourcekit-lsp/issues/1010

PR #1184, opened 2024-04-18
Update Editors/README.md

yaroslavyaroslav:
In Sublime Text setup part, to reflect recent LSP package API changes.

### DIFF
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -29,7 +29,7 @@ To configure SourceKit-LSP additionally, open the SourceKit-LSP package's settin
 ```json
 {
     "command": [
-        "<sourcekit-lsp command>"
+        "sourcekit-lsp"
     ],
     "languages": [
         {
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -22,9 +22,11 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 ## Sublime Text
 
-Before using SourceKit-LSP with Sublime Text, you will need to install the LSP package from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
+Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
 
 To configure SourceKit-LSP additionally, open the SourceKit-LSP package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
+
+To make your journey even more pleasant it's recomended to install [Swift-Next](https://github.com/Swift-Next/Swift-Next) package that provides advanced swift syntax highlighting support.
 
 ```json
 {
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -24,7 +24,7 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
 
-To configure SourceKit-LSP additionally, open the SourceKit-LSP package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
+To configure LSP-SourceKit additionally, open the LSP-SourceKit package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
 
 To make your journey even more pleasant it's recomended to install [Swift-Next](https://github.com/Swift-Next/Swift-Next) package that provides advanced swift syntax highlighting support.
 
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -33,23 +33,7 @@ To make your journey even more pleasant it's recomended to install [Swift-Next](
     "command": [
         "sourcekit-lsp"
     ],
-    "languages": [
-        {
-            "languageId": "c"
-        },
-        {
-            "languageId": "c++"
-        },
-        {
-            "languageId": "objc"
-        },
-        {
-            "languageId": "objc++"
-        },
-        {
-            "languageId": "swift"
-        },
-    ]
+    "selector": "source.c | source.c++ | source.objc | source.objc++ | source.swift "
 }
 ```
 
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -26,7 +26,7 @@ Before using SourceKit-LSP with Sublime Text, you will need to install the LSP a
 
 To configure LSP-SourceKit additionally, open the LSP-SourceKit package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
 
-To make your journey even more pleasant it's recomended to install [Swift-Next](https://github.com/Swift-Next/Swift-Next) package that provides advanced swift syntax highlighting support.
+Also install [Swift-Next](https://github.com/Swift-Next/Swift-Next) syntax package to use Swift in Sublime Text 4.
 
 ```json
 {
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -22,20 +22,7 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 ## Sublime Text
 
-Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
-
-To configure LSP-SourceKit additionally, open the LSP-SourceKit package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
-
-Also install [Swift-Next](https://github.com/Swift-Next/Swift-Next) syntax package to use Swift in Sublime Text 4.
-
-```json
-{
-    "command": [
-        "sourcekit-lsp"
-    ],
-    "selector": "source.c | source.c++ | source.objc | source.objc++ | source.swift"
-}
-```
+Before using SourceKit-LSP with Sublime Text, you will need to install the [LSP](https://packagecontrol.io/packages/LSP), [LSP-SourceKit](https://github.com/sublimelsp/LSP-SourceKit) and [Swift-Next](https://github.com/Swift-Next/Swift-Next) packages from Package Control. Then toggle the server on by typing in command palette `LSP: Enable Language Server Globally` or `LSP: Enable Language Server in Project`.
 
 ## Emacs
 
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -22,56 +22,32 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 ## Sublime Text
 
-Before using SourceKit-LSP with Sublime Text, you will need to install the LSP package from Package Control. To configure SourceKit-LSP, open the LSP package's settings. The following snippet should be enough to get started with Swift.
+Before using SourceKit-LSP with Sublime Text, you will need to install the LSP package from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
 
-You will need the path to the `sourcekit-lsp` executable for the "command" section.
+To configure SourceKit-LSP additionally, open the SourceKit-LSP package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
 
 ```json
 {
-  "clients":
-  {
-    "SourceKit-LSP":
-    {
-      "enabled": true,
-      "command": [
+    "command": [
         "<sourcekit-lsp command>"
-      ],
-      "env": {
-        // To override the toolchain, uncomment the following:
-        // "SOURCEKIT_TOOLCHAIN_PATH": "<path to toolchain>",
-      },
-      "languages": [
+    ],
+    "languages": [
         {
-          "scopes": ["source.swift"],
-          "syntaxes": [
-            "Packages/Swift/Syntaxes/Swift.tmLanguage",
-            "Packages/Decent Swift Syntax/Swift.sublime-syntax",
-          ],
-          "languageId": "swift"
+            "languageId": "c"
         },
         {
-          "scopes": ["source.c"],
-          "syntaxes": ["Packages/C++/C.sublime-syntax"],
-          "languageId": "c"
+            "languageId": "c++"
         },
         {
-          "scopes": ["source.c++"],
-          "syntaxes": ["Packages/C++/C++.sublime-syntax"],
-          "languageId": "cpp"
+            "languageId": "objc"
         },
         {
-          "scopes": ["source.objc"],
-          "syntaxes": ["Packages/Objective-C/Objective-C.sublime-syntax"],
-          "languageId": "objective-c"
+            "languageId": "objc++"
         },
         {
-          "scopes": ["source.objc++"],
-          "syntaxes": ["Packages/Objective-C/Objective-C++.sublime-syntax"],
-          "languageId": "objective-cpp"
+            "languageId": "swift"
         },
-      ]
-    }
-  }
+    ]
 }
 ```
 
--- a/Editors/README.md
+++ b/Editors/README.md
@@ -33,7 +33,7 @@ To make your journey even more pleasant it's recomended to install [Swift-Next](
     "command": [
         "sourcekit-lsp"
     ],
-    "selector": "source.c | source.c++ | source.objc | source.objc++ | source.swift "
+    "selector": "source.c | source.c++ | source.objc | source.objc++ | source.swift"
 }
 ```
 


Editors/README.md
```diff
@@ -22,56 +22,32 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
...
 ```json
 {
-  "clients":
-  {
-    "SourceKit-LSP":
-    {
-      "enabled": true,
-      "command": [
+    "command": [
         "<sourcekit-lsp command>"
```
ahoppen, line 32: I think we can just use `sourcekit-lsp` here, which should work if you have `sourcekit-lsp` in your PATH, which you should have if you have Xcode installed. We do the same for the `vim` instructions below.

Editors/README.md
```diff
@@ -22,56 +22,32 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
...
-      },
-      "languages": [
+    ],
+    "languages": [
         {
-          "scopes": ["source.swift"],
-          "syntaxes": [
-            "Packages/Swift/Syntaxes/Swift.tmLanguage",
-            "Packages/Decent Swift Syntax/Swift.sublime-syntax",
-          ],
```
ahoppen, line 49: Why don’t we need to specify `syntax` files anymore? sourcekit-lsp only implements semantic highlighting and while it does take that job very far and returns classifications for most tokens, I think it’s good to have an in-process syntactic highlighting as well, eg. when the `textDocument/semanticTokens` requests takes a while to reply.

ahoppen (commented):
Thanks for updating the instructions @yaroslavyaroslav. Two questions/comments inline

Editors/README.md
yaroslavyaroslav: It's because of LSP update itself, not sure for how long, but definitely quite a while LSP plugin relies on sublime text scopes, rather than on selected syntax, e.g. `source.swift`, and semantic highlight feature toggles generally now.. But anyway the current sourcekit-lsp wrapper doesn't support that feature yet. I'll try to update it properly someday I guess.

Here's the references:
1. general LSP related setting [link](https://github.com/sublimelsp/LSP/blob/5891ff41deaa781783effeb97c978b3e50bd0a26/LSP.sublime-settings#L216)
2. kinda golden standard pyright LSP client with the absence of syntaxes settings [link](https://github.com/sublimelsp/LSP-pyright/blob/7a31a89b247b788793e4383e1548345d6f842858/LSP-pyright.sublime-settings#L58)
3. current settings of a sourcekit-lsp client with the very limited setup options [link](https://github.com/sublimelsp/LSP-SourceKit/blob/master/LSP-SourceKit.sublime-settings)

Editors/README.md
yaroslavyaroslav: I've updated, but isn't this man should be cross-platform? I mean the default client side command is `xcrun sourcekit-lsp` that should just work in one have Xcode installed, therefore this setting is redundant at all. But I'm not sure about the picture on Linux and Windows.

ahoppen: My understanding is that if you install Xcode, you get `/usr/bin/sourcekit-lsp` (which is in your PATH), which is essentially a trampoline for `xcrun sourcekit-lsp`. So, if we use `sourcekit-lsp` as the command, it uses that on macOS. On Linux, we expect the toolchain to be in your path as well (which it usually is if you install it to `/usr/bin`, so then you also have `sourcekit-lsp` in your path). Given that this should work for most users, I think it’s good to require one less customization step from users.

Editors/README.md
ahoppen: Regarding the LSP link: I just want to state that LSP supports semantic highlighting, not syntactic highlighting. Semantic highlighting might take a while (we’re ideally talking 100ms - 1s because we need to build an AST to figure out all the types and whether identifiers are locals, members, system symbols, …) here, my understanding of the LSP spec has always been that clients should have fast syntactic highlighting and enrich that with the semantic highlighting from LSP as it comes in.

But I don’t have enough experience how the current best practices in Sublime-Land are and will leave the decision up to you.

yaroslavyaroslav: Yep, this is exactly the case. The syntactic highlighting handled by a `*.sublime-syntax` files powered by a st engine. But the catch is that those settings are completely unrelated to lsp functioning.

The reason why those settings were here before was that it was the way how the lsp plug-in mapped which server to call on a given file, but isn't true now.

Editors/README.md
yaroslavyaroslav: I have nothing to complain here. I'm fine with either way. Just mentioning, that this block could be skipped completely if the minimalism matters.

Editors/README.md
yaroslavyaroslav: Btw, speaking about syntactic highlighting, those that were here were really limited in their implementation, finally there is a way better one now implemented. [this one](https://forums.swift.org/t/swift-next-swift-syntax-package-in-async-await-era-for-sublime-text-4/58227). So just in case, if you had an idea where to put a highlight of it to make swift st users life a little less painful, I'd really appreciate that.

ahoppen: Oh, that’s great. I have personally been waiting for some syntactic highlighting package that contains recent Swift syntax for Sublime. Thanks 🙏🏽 

How about you mention it in this document?

yaroslavyaroslav: Ah, you just can't imagine how I was thrilled when found it occasionally. 

Done. Looking forward to it being merged, if there're no issues left.

yaroslavyaroslav: Ah, the last one, FYI: not working semantic syntax highlighting for a swift was a Sublime LSP plugin side bug, and it's almost [fixed](https://github.com/sublimelsp/LSP/pull/2453) now.

rchl: `languages` and `languageId` is deprecated in ST LSP. Instead `"selector": "source.swift | source.c"' (etc.) should be used.

ST3 doesn't support `selector` but ST3 has not been relevant for many years now. ST4 is what matters.

Editors/README.md
```diff
@@ -22,56 +22,34 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 ## Sublime Text
 
-Before using SourceKit-LSP with Sublime Text, you will need to install the LSP package from Package Control. To configure SourceKit-LSP, open the LSP package's settings. The following snippet should be enough to get started with Swift.
+Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
 
-You will need the path to the `sourcekit-lsp` executable for the "command" section.
+To configure SourceKit-LSP additionally, open the SourceKit-LSP package's settings by typing in command palette `Preferences: LSP-SourceKit Settings`. The following snippet should be enough to get started with Swift and Objective-C/C++ and the custom path to `sourcekit-lsp` executable.
```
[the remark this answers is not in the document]
yaroslavyaroslav: 1. Fair point, fixed.
2. I'd rather leave this one to @ahoppen decision.
3. This one is interesting. AFAIK `xcrun` is a part of xcodebuild tools only, thus it provided on macOS only, so in this case, since ST is focused on being crossplatfrom, maybe to change the default setup from `xcrun sourcekit-lsp` up to `sourcekit-lsp` could be a better choice. Anyway, it was discussed just right [here](https://github.com/apple/sourcekit-lsp/pull/1184#discussion_r1571461864), and Alex insisted to keep this line in anyway.

Editors/README.md
```diff
@@ -22,56 +22,34 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
...
+
+To make your journey even more pleasant it's recomended to install [Swift-Next](https://github.com/Swift-Next/Swift-Next) package that provides advanced swift syntax highlighting support.
```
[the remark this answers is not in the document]
yaroslavyaroslav: I believe we have this talk before, when Swift-Next got included in Package Control [initially](https://github.com/wbond/package_control_channel/pull/8843#issuecomment-1814248902). So after all folks decided to not replace any existing ST swift syntax package with this one, but to add tests and add it into the default [packages list](https://github.com/sublimehq/Packages) instead. In short the main reason was that this package is ST4 compatible only, which is not the case for the "default" [Swift](https://github.com/quiqueg/Swift-Sublime-Package) package provides ST3 and ST2 support as well. So here we are. The only thing I can to this: someday I'll complete the tests task, but the day is not today.

The same reasoning is true when we're speaking about the replacement of [Decent Swift Syntax](https://packagecontrol.io/packages/Decent%20Swift%20Syntax) with [Swift-Next](https://github.com/Swift-Next/Swift-Next) in LSP-SourceKit package readme. If it's acceptable to suggest the ST4 only option as the default option for this package — i'm up for it, and would love to update its readme as well. But if you prefer to keep the one that provides better backward capability, so there's nothing to do with LSP-SourceKit's readme.

Editors/README.md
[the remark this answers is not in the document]
rchl: If `sourcekit-lsp` is a better command than `xcrun sourcekit-lsp` then this should be updated in the `LSP-SourceKit` package and then this override could be removed here.

Editors/README.md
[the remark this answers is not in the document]
rchl: Yes, ST3 is not relevant anymore.
And also `LSP-SourceKit` is ST4-only so if "Swift-Next" is better, it should just recommend that.

But what I'm mainly tried to say is that from the perspective of someone who just reads intructions here, there is not even a mention of installing any syntax so this is why I thought the wording here is not ideal. Many users will likely not read `LSP-SourceKit` readme separately if they are installing it based on those instructions.

Editors/README.md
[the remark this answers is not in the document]
yaroslavyaroslav: Again it's on Alex duty to decide

Editors/README.md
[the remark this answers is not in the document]
yaroslavyaroslav: Yeah, I see your point, but personally I think that it's unlikely that a user would bother to install an lsp plugin for language while intentionally skipping its syntax highlighting part. So this is why I chose this form rather than more strict one.

Editors/README.md
```diff
@@ -22,56 +22,18 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
 
 ## Sublime Text
 
-Before using SourceKit-LSP with Sublime Text, you will need to install the LSP package from Package Control. To configure SourceKit-LSP, open the LSP package's settings. The following snippet should be enough to get started with Swift.
+Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `xcrun` in the `$PATH`.
```
ahoppen, line 25: I would change this to `sourcekit-lsp` instead of `xcrun` because we don’t actually invoke `xcrun`.

```suggestion
Before using SourceKit-LSP with Sublime Text, you will need to install the LSP and LSP-SourceKit packages from Package Control. And you're ready to go if you have `sourcekit-lsp` in the `$PATH`. This is the case if you have Xcode installed on macOS or a [Swift Toolchain](https://www.swift.org/install/) installed on Linux.
```

Editors/README.md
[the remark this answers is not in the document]
ahoppen: > If sourcekit-lsp is a better command than xcrun sourcekit-lsp then this should be updated in the LSP-SourceKit package and then this override could be removed here.

I agree, it would be better to update that package. Also, do we even need to specify the `selector` here or could that also be sunk down into the `LSP-SourceKit` package, in which case no additional configuration would be necessary?

Editors/README.md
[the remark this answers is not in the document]
ahoppen: I agree that we should make the suggestion for the syntax highlighting package consistent between the documentation of _LSP-SourceKit_ and this REAMDE. I take it that there is consensus that _Swift Next_ is superior to _Decent Swift Syntax_ (especially because there is a plan to get _Swift Next_ into the default packages) except for the part that _Swift Next_ is not ST4-compatible. But if _LSP-SourceKit_ isn’t either, I don’t think that’s a blocker. Could we

- Change _LSP-SourceKit_ to also refer to _Swift Next_?
- I agree that we should rephrase this sentence to be a little less optional, on the lines of: Also install _Swift Next_ to use Swift in ST.
- I am not familiar with ST’s package infrastructure. But would it be possible to make _LSP-SourceKit_ depend on _Swift Next_ so users get the entire bundle when installing _LSP-SourceKit_?

Editors/README.md
ahoppen: I think the right course of action would be to:
- Change _LSP-SourceKit_ to run `sourcekit-lsp` instead of `xcrun sourcekit-lsp`. `/usr/bin/sourcekit-lsp` is just a shim around `xcrun sourcekit-lsp`.
- Then we don’t even need to include the `command` here.

Editors/README.md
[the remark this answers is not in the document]
rchl: > Change LSP-SourceKit to also refer to Swift Next?

It's updated now - https://github.com/sublimelsp/LSP-SourceKit/pull/6

> I am not familiar with ST’s package infrastructure. But would it be possible to make LSP-SourceKit depend on Swift Next so users get the entire bundle when installing LSP-SourceKit?

It's not technically possible right now for one package to depend on another package. There is a concept of dependencies but those are not packages and can't be installed on its own.

Editors/README.md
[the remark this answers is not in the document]
rchl: The LSP-SourceKit package by default only enables server for swift files. Does it make sense to enable it for c/c++/objc by default? I guess it could be problematic to have it running in projects that are not really swift projects?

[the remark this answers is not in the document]
ahoppen: Oh, I see. The VS Code extension specifies support for C/C++/ObjC/ObjC++ by default, so I think that is a default that we can use here as well. That way, they are aligned.

Editors/README.md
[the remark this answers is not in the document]
ahoppen: > It's updated now - [sublimelsp/LSP-SourceKit#6](https://github.com/sublimelsp/LSP-SourceKit/pull/6)

Thanks!

> It's not technically possible right now for one package to depend on another package. There is a concept of dependencies but those are not packages and can't be installed on its own.

Ah, OK, I wasn’t aware of that. Let’s keep the suggestion in here then.

Editors/README.md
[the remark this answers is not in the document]
yaroslavyaroslav: > The VS Code extension specifies support

It's some kind of similar to suggesting a feature to macOS team by referencing to Windows design decision 😀 It could work somewhat with zed though.

Anyway I'd rather agree with that, there's a loosing chance that some one who working on plain c/c++ codebase would occasionally install and turn on globally LSP-SourceKit that'll brings to a lsp servers conflict on an ST side.

So if @rchl has nothing against that, I can provide another LSP-SourceKit that (a) extends its scope to the list above and (b) replaces a given `xcrun sourcekit-lsp` command to `/usr/bin/sourcekit-lsp` or whatever that we would decide in the thread above.

Editors/README.md
yaroslavyaroslav: What about Windows users? it's obviously would be a different path there.

Editors/README.md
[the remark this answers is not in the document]
rchl: I think that replacing `xcrun sourcekit-lsp` with `sourcekit-lsp` should be done in the LSP-SourceKit package. At least as long as it's tested on mac.

As for enabling for other file types... I suppose it could be a double edged sword. On one hand it would potentially enable server in projects that don't use swift. On the other hand, not enabling for other file types probably makes the server less effective in swift projects.

So one way of handling that could be to:
1. Disable LSP-SourceKit server by default (have `enabled: false` in its default settings
2. Add instructions in its readme that if the user wants to enable it, he/she has to run "LSP: Enable language server in project" from the command palette.
So users would explicitly opt-in to it per-project.

Editors/README.md
ahoppen: My thinking here was that Windows users will need to update the path in both versions, so I thought it’s preferable if it at least works out-of-the-box for those on macOS and Linux.

That being said, `sourcekit-lsp.exe` should be in `PATH` on Windows as well if a Swift toolchain is installed. Is there a way in ST to make the executable name dependent on the OS? This is about adding the `.exe` suffix or not.

Editors/README.md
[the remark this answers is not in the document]
ahoppen: That sounds like a great idea to me.

Also @rchl: Thanks for engaging in this discussion and sharing your knowledge of ST packages. That’s really valuable. 🙏🏽

Editors/README.md
rchl: > > Is there a way in ST to make the executable name dependent on the OS?
> 
> AFAIK, no.

Yes, it is. The plugin code of LSP-SourceKit can dynamically modify command based on OS.

ahoppen: > Yes, it is. The plugin code of LSP-SourceKit can dynamically modify command based on OS.
But I think that we don't have to care about that for now. LSP-SourceKit is anyway only available for Mac and Linux currently.

I think it would be good to consider adding support for LSP-SourceKit on Windows as well but that that’s a separate topic. 

> @ahoppen What's about this PR? What do we do now? Do we skipping the whole ST part or replacing it with a placeholder text or leaving it as it is just for information purpose?

I think the key pieces of information we should include in the REAME are:
- To use SourceKit-LSP in Sublime, install the LSP-SourceKit package
- Install Swift-Next for syntactic highlighting
- Note that clang-family languages can be enabled in SourceKit-LSP as well, as discussed here https://github.com/apple/sourcekit-lsp/pull/1184#discussion_r1576129038
> Add instructions in its readme that if the user wants to enable it, he/she has to run "LSP: Enable language server in project" from the command palette.
So users would explicitly opt-in to it per-project.

And yes, I think then we don’t need the snippet at all anymore, which is a great simplification in my opinion.

yaroslavyaroslav: > Note that clang-family languages can be enabled in SourceKit-LSP as well

They are enabled by default as rchl suggested here https://github.com/apple/sourcekit-lsp/pull/1184#discussion_r1576129038 so I'd rather skip this part completely.

So to sum things up a bit:

LSP-SourceKit Sublime text plugin's default setup https://github.com/sublimelsp/LSP-SourceKit/pull/7:
- command `sourcekit-lsp`, (plain, not `/usr/bin/sourcekit-lsp`)
- ``enabled`: false` this server disabled globally in ST, so if user installed it he/she should to toggle it on in advance per command palette or by overriding global LSP-SourceKit settings.
- `scope: c | c++ | objc | objc++ | swift` which means that it's handling broader cland-family language scope (therefore it's global disabled by default).

So considering all of this I'll update this readme with the mention of the LSP-SourceKit itself and a Swift-Next syntax to install, and trowing off all the code snippet.

ahoppen: Oh, my understanding was that we enabled it by default for Swift but disabled it by default for all the clang languages. But I assume that’s not possible?

yaroslavyaroslav: Nah, I'd rather leave this to answer to @rchl he definitely better in this 😀

rchl: What I meant before is what @yaroslavyaroslav said and did.

To have it enabled by default for Swift and for the user to be able to opt-in into other file types, could be done but it would be less user friendly. The user would have to go into LSP-SourceKit settings and override the `selector` manually (we could leave commented-out, alternative selector to make it easier). So essentially what this guide said initially. But if the user does that manually, then he/she will still have the same problem that it might trigger in non-swift projects. Unless user overrides `selector` in project-specific settings (which is also possible) but again, less user friendly than just using Command Palette.

So not sure what's better. I would have to leave the decision to you two. I don't even code in Swift.

yaroslavyaroslav: So again, let's sum things up:

1. The scope could be declared only once in a settings config, so it could be either all or nothing.
2. If we declare the full scope of the server, we should disable it by default, due to potential conflicts with `clang` one.
3. We can declare some limited scope as default (for instance objc/objc++/swift), but then it would take more effort from a user to expand it with c/c++ on his side rather than just send a single command in command palette. The other downside of this decision, is that all objc header files identifies as `c` scope in ST by default, so it's likely would be a cause of an inconsistency issue in a UX with the sourcekit-lsp.

So after all, I believe that this is the best option that we got, to make it less error prone on one hand and to try to keep UX at its best on the other.

yaroslavyaroslav: @ahoppen FYI: the PR https://github.com/sublimelsp/LSP-SourceKit/pull/7 is merged, so it's safe to merge this one now as well.

ahoppen (approved):
Looks great to me. Would like to get a second review from @bnbarham as well.

Editors/README.md
```diff
@@ -22,58 +22,7 @@ The [Swift for Visual Studio Code extension](https://marketplace.visualstudio.co
...
-          "scopes": ["source.objc++"],
-          "syntaxes": ["Packages/Objective-C/Objective-C++.sublime-syntax"],
-          "languageId": "objective-cpp"
-        },
-      ]
-    }
-  }
-}
-```
+Before using SourceKit-LSP with Sublime Text, you will need to install the [LSP](https://packagecontrol.io/packages/LSP), [LSP-SourceKit](https://github.com/sublimelsp/LSP-SourceKit) and [Swift-Next](https://github.com/Swift-Next/Swift-Next) packages from Package Control. Then toggle the server on by typing in command palette `LSP: Enable Language Server Globally` or `LSP: Enable Language Server in Project`.
```
bnbarham, line 25: IIUC LSP-SourceKit is a small amount of configuration to enable sourcekit-lsp (which could be done in the LSP configuration)? And Swift-Next adds grammar files (vs the old https://github.com/quiqueg/Swift-Sublime-Package and https://github.com/colinta/decent-swift-syntax)?

If so, I think we should re-word this slightly to something like:
> Before using SourceKit-LSP with Sublime Text, you will need to install the [LSP](https://packagecontrol.io/packages/LSP) from Package Control. From there you can either [manually configure](https://lsp.sublimetext.io/client_configuration/) Swift and C-family languages to use `sourcekit-lsp`, or use a community-provided package to set this up automatically (eg. [LSP-SourceKit](https://github.com/sublimelsp/LSP-SourceKit)).

For syntax highlighting, how does this interact with the semantic highlighting provided by LSP? Is it merged after the response comes back? Something else? Depending on that we could also add eg.
> [Swift-Next](https://github.com/Swift-Next/Swift-Next) (another community-provided package) can also be added for syntax highlighting prior to Sublime receiving a response from the LSP server.

Not for you in this PR, but we probably need to got through this doc and clear it up for the other editors in general 😅.

yaroslavyaroslav: # Swift docs update

Ah sheesh, here we go again 😅

So let's break it down from top to bottom in one place

## LSP-SourceKit package purpose

> LSP-SourceKit is a small amount of configuration to enable `sourcekit-lsp` (which could be done in the LSP configuration)? 

It's no. It was so a while ago, but at least for a few years for now it has handled by a separate `LSP-ServerName` packages. This mentioned in a note on a link that you provided in your suggested change [snippet](https://lsp.sublimetext.io/client_configuration/):

> The external LSP-* helper packages already come with their setting file and a client configuration and you don't need to add anything to the global LSP settings. This section is only relevant if you want to add a new client configuration for a server that doesn't have a corresponding helper package.

So since we're having LSP-SourceKit helper package this is the main way to interact with a `sourcekit-lsp` server for ST4 user.

## LSP-SourceKit maintaince

> or use a community-provided package to set this up automatically (eg. LSP-SourceKit).

Technically speaking all the LSP stack for Sublime Text is community provided, including the main [LSP](https://github.com/sublimelsp/LSP) itself. The main LSP package and in our case LSP-SourceKit helper are stored within the same group (?) on github so it's make sense to treat them both as either semi-official of community-driven at once.

## C family language support

> either ... Swift and C-family languages to use `sourcekit-lsp`

This is true, the default configuration of LSP-SourceKit package is set to support all `c`, `c++`, `objc`, `objc++` and `swift`. And at the same time this is the reason why the plugin **installs in it's disabled state**. We had long discussion about how to make it best and conclude that this is the best scenario from UX perspective.

To elaborate that, most ST users unaware of `sourcekit-lsp` capabilities to support c family languages, and they can use `clang` for that purpose in theirs swift unrelated projects. So This bit says no to either include `c`, `c++` into LSP-SourceKit default scope or toggle it on by default.

From the other perspective, if to exclude those scopes from the default setup, ST barely could differentiate `c`/`c++` header file from `objc` one, which leads to it will not call this server for `objc` headers if they are appeared in project without additional user's effort to setup scope properly. And this is requires more effort that just run a single command from within command palette to explicitly toggle this server for a project or globally.

This is how we went here. To include the full list of scopes that `sourcekit-lsp` supports in the default setup, but to toggle them off on installation, to not break other plain c/cpp projects.

## Miscellaneous Syntax highlighting questions

### Various ST swift syntax packages

> And Swift-Next adds grammar files (vs the old https://github.com/quiqueg/Swift-Sublime-Package and https://github.com/colinta/decent-swift-syntax)?

That's right. ST suffered from a very limited swift syntax highlight support until Swift-Next release, which is finally quite mature in just this. Now this package is on its (long) way to being included in predefined [packages list](https://github.com/sublimehq/Packages)

As my college claimed above the one grammar file is strictly required to make LSP helper package work. So me and ST team are suggesting this one among the others.

### Semantic syntax highlight support

> For syntax highlighting, how does this interact with the semantic highlighting provided by LSP? Is it merged after the response comes back?

Yep, grammar loads first, almost instantly on a file open, then, it could be expanded with semantic syntax tokens provided by a server. Now this feature disabled by default (to the LSP package in general), I've tried to enable this for swift sources and have to say it's pretty useless for ST for now, as it's syntax highlight just worst comparing to a grammar one. Didn't dig deep enough yet, but I believe the cause is that there is token naming mismatch between those that server sending and ST are working with.

@rchl could you please fix me if I'm wrong in the message above, if you'll have a time to look at this.

bnbarham: > Didn't dig deep enough yet, but I believe the cause is that there is token naming mismatch between those that server sending and ST are working with.

Ah, I wonder if this was while we had an extra kind inbetween the built-in ones. *Technically* this is allowed by the spec, but most clients seem to ignore the mapping 😅. That was changed in https://github.com/apple/sourcekit-lsp/pull/1012.

bnbarham (approved):
Thank you for the detailed summary @yaroslavyaroslav, that was super helpful!

bnbarham:
@swift-ci please test

Editors/README.md
yaroslavyaroslav: Heh, yeah, it was defiant after that PR, like a few weeks ago. It's because before that LSP package had a bug that broke swift semantics syntax highlight completely 😅. 

Now it's fixed, but yet if to enable it's literally colors almost all the tokens in the same color for a vast majority of ST color schemes that I've tried. 

I saw an option to remap tokens names right within LSP helper scope, so I think it could be a solution to that.

ahoppen:
@yaroslavyaroslav It looks like you deleted your repository and thus we can’t merge this PR. Would you mind setting it up again? Otherwise, I can also re-open the PR in your name.

yaroslavyaroslav:
@ahoppen oh my bad, I still have in on my laptop will recreate PR today.

yaroslavyaroslav:
Must be fixed now. Sorry about the mess that I brought, was today's old when knew that forks on GH implemented in reference like behaviour 😅

ahoppen:
Thanks for re-opening the PR.

ahoppen:
@swift-ci Please test macOS

PR #1185, opened 2024-04-18
Fix jump-to-definition to methods in swift interfaces that are in synthesized extensions

ahoppen:
For example when trying to go-to-definition to `filter` on `Array`, we get a USR `s:s14_ArrayProtocolPsE6filterySay7ElementQzGSbAEKXEKF::SYNTHESIZED::s:Sa`. We were trying to look it up in the index, which failed because synthesized extension methods are not indexed.

Instead, consult the `module` and `groupName` that `sourcekitd` returns in the cursor info request to decide which module to jump to.

rdar://126240558


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1186, opened 2024-04-18
Truncate log message in `NonDarwinLogger` at 10.000 characters

ahoppen:
Truncate log message after 10.000 characters to avoid flooding the log with huge log messages (eg. from a sourcekitd response). 10.000 characters wasn't chosen because it seems to fit the result of most sourcekitd responses that are not generated interface or global completion results (which are a lot bigger).

### DIFF
--- a/Sources/LSPLogging/NonDarwinLogging.swift
+++ b/Sources/LSPLogging/NonDarwinLogging.swift
@@ -299,11 +299,20 @@ public struct NonDarwinLogger: Sendable {
     guard level >= self.logLevel else { return }
     let date = Date()
     loggingQueue.async {
+      // Truncate log message after 10.000 characters to avoid flooding the log with huge log messages (eg. from a
+      // sourcekitd response). 10.000 characters was chosen because it seems to fit the result of most sourcekitd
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/LSPLogging/NonDarwinLogging.swift
```diff
@@ -299,11 +299,20 @@ public struct NonDarwinLogger: Sendable {
     guard level >= self.logLevel else { return }
     let date = Date()
     loggingQueue.async {
+      // Truncate log message after 10.000 characters to avoid flooding the log with huge log messages (eg. from a
+      // sourcekitd response). 10.000 characters wasn't chosen because it seems to fit the result of most sourcekitd
```
bnbarham, line 303:
```suggestion
      // sourcekitd response). 10.000 characters wasn't chosen because it seems to fit the result of most sourcekitd
```

"was" 😅?

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1192, opened 2024-04-19
Slightly simplify code completion logic

ahoppen:
Fallout from trying to fix rdar://126708088 (No closure expansion in call pattern completions) which I eventually gave up on because it required too much redesign of `CodeCompletionSession` for too little value for too much risk right now.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1193, opened 2024-04-20
Add refactoring actions for adding a test targets and products to a package manifest

DougGregor:
Leverage the newly-introduced package manifest editing tools in SwiftPM
to create a package editing refactoring operation. This operation can
be triggered from the a target in the manifest itself, e.g.,

    .target(name: "MyLib")

and will add a test target to the package manifest that depends on
this target, i.e.,

    .testTarget(
      name: "MyLibTests",
      dependencies: [ "MyLib" ]
    )

It will also create a new source file `Tests/MyLibTests/MyLibTests.swift`
that that imports both MyLib and XCTest, and contains an XCTestCase subclass
with one test to get you started.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -309,6 +309,7 @@ let package = Package(
         .product(name: "SwiftRefactor", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
+        .product(name: "SwiftPM-auto", package: "swift-package-manager"),
       ],
       exclude: ["CMakeLists.txt"]
     ),
--- a/Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
@@ -52,31 +52,4 @@ public struct TextDocumentEdit: Hashable, Codable, Sendable {
     self.textDocument = textDocument
     self.edits = edits
   }
-
-  public enum CodingKeys: String, CodingKey {
-    case kind
-    case textDocument
-    case edits
-  }
-
-  public func encode(to encoder: Encoder) throws {
-    var container = encoder.container(keyedBy: CodingKeys.self)
-    try container.encode("textDocumentEdit", forKey: .kind)
-    try container.encode(self.textDocument, forKey: .textDocument)
-    try container.encode(self.edits, forKey: .edits)
-  }
-
-  public init(from decoder: Decoder) throws {
-    let container = try decoder.container(keyedBy: CodingKeys.self)
-    let kind = try container.decode(String.self, forKey: .kind)
-    guard kind == "textDocumentEdit" else {
-      throw DecodingError.dataCorruptedError(
-        forKey: .kind,
-        in: container,
-        debugDescription: "Kind of TextDocumentEdit is not 'textDocumentEdit'"
-      )
-    }
-    self.textDocument = try container.decode(OptionalVersionedTextDocumentIdentifier.self, forKey: .textDocument)
-    self.edits = try container.decode([Edit].self, forKey: .edits)
-  }
 }
--- a/Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
@@ -52,4 +52,31 @@ public struct TextDocumentEdit: Hashable, Codable, Sendable {
     self.textDocument = textDocument
     self.edits = edits
   }
+
+  public enum CodingKeys: String, CodingKey {
+    case kind
+    case textDocument
+    case edits
+  }
+
+  public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode("textDocumentEdit", forKey: .kind)
+    try container.encode(self.textDocument, forKey: .textDocument)
+    try container.encode(self.edits, forKey: .edits)
+  }
+
+  public init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    let kind = try container.decode(String.self, forKey: .kind)
+    guard kind == "textDocumentEdit" else {
+      throw DecodingError.dataCorruptedError(
+        forKey: .kind,
+        in: container,
+        debugDescription: "Kind of TextDocumentEdit is not 'textDocumentEdit'"
+      )
+    }
+    self.textDocument = try container.decode(OptionalVersionedTextDocumentIdentifier.self, forKey: .textDocument)
+    self.edits = try container.decode([Edit].self, forKey: .edits)
+  }
 }
--- a/Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
@@ -1,0 +1,166 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import PackageModel
+import PackageModelSyntax
+import SwiftRefactor
+import SwiftSyntax
+
+/// Syntactic code action provider to provide refactoring actions that
+/// edit a package manifest.
+struct PackageManifestEdits: SyntaxCodeActionProvider {
+  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard
+      let token = scope.firstToken,
+      let call = token.findEnclosingCall()
+    else {
+      return []
+    }
+
+    var actions = [CodeAction]()
+
+    // If there's a target name, offer to create a test target derived from it.
+    if let targetName = call.findStringArgument(label: "name") {
+        do {
+            let target = try TargetDescription(
+                name: "\(targetName)Tests",
+                dependencies: [ .byName(name: targetName, condition: nil) ],
+                type: .test
+            )
+
+            let edits = try AddTarget.addTarget(target, to: scope.file)
+            actions.append(
+              CodeAction(
+                title: "Add test target for this target",
+                kind: .refactor,
+                edit: edits.asWorkspaceEdit(snapshot: scope.snapshot)
+              )
+            )
+        } catch {
+            // nothing to do
+        }
+    }
+
+    return actions
+  }
+}
+
+fileprivate extension PackageEditResult {
+    /// Translate package manifest edits into a workspace edit. This can
+    /// involve both modifications to the manifest file as well as the creation
+    /// of new files.
+    func asWorkspaceEdit(snapshot: DocumentSnapshot) -> WorkspaceEdit {
+        // The edits to perform on the manifest itself.
+        let manifestTextEdits = manifestEdits.map { edit in
+            TextEdit(
+                range: snapshot.range(of: edit.range),
+                newText: edit.replacement
+            )
+        }
+
+        // If we couldn't figure out the manifest directory, or there are no
+        // files to add, the only changes are the manifest edits. We're done
+        // here.
+        let manifestDirectoryURI = snapshot.uri.fileURL?
+            .deletingLastPathComponent()
+        guard let manifestDirectoryURI, !auxiliaryFiles.isEmpty else {
+            return WorkspaceEdit(
+                changes: [snapshot.uri: manifestTextEdits]
+            )
+        }
+
+        // Use the more full-featured documentChanges, which takes precedence
+        // over the individual changes to documents.
+        var documentChanges: [WorkspaceEditDocumentChange] = []
+
+        // Put the manifest changes into the array.
+        documentChanges.append(
+            .textDocumentEdit(
+                .init(
+                    textDocument: .init(snapshot.uri, version: nil),
+                    edits: manifestTextEdits.map { .textEdit($0) }
+                )
+            )
+        )
+
+        // Create an populate all of the auxiliary files.
+        for (relativePath, contents) in auxiliaryFiles {
+            guard let uri = URL(
+                string: relativePath.pathString,
+                relativeTo: manifestDirectoryURI
+            ) else {
+                continue
+            }
+
+            let documentURI = DocumentURI(uri)
+            let createFile = CreateFile(
+                uri: documentURI
+            )
+
+            let zeroPosition = Position(line: 0, utf16index: 0)
+            let edit = TextEdit(
+                range: zeroPosition..<zeroPosition,
+                newText: contents.description
+            )
+
+            documentChanges.append(.createFile(createFile))
+            documentChanges.append(
+                .textDocumentEdit(
+                    .init(
+                        textDocument: .init(documentURI, version: nil),
+                        edits: [ .textEdit(edit) ]
+                    )
+                )
+            )
+        }
+
+        return WorkspaceEdit(
+            changes: [ snapshot.uri: manifestTextEdits ],
+            documentChanges: documentChanges
+        )
+    }
+}
+
+fileprivate extension SyntaxProtocol {
+    // Find an enclosing call syntax expression.
+    func findEnclosingCall() -> FunctionCallExprSyntax? {
+        var current = Syntax(self)
+        while true {
+            if let call = current.as(FunctionCallExprSyntax.self) {
+                return call
+            }
+
+            if let parent = current.parent {
+                current = parent
+                continue
+            }
+
+          return nil
+      }
+  }
+}
+
+fileprivate extension FunctionCallExprSyntax {
+    /// Find an argument with the given label that has a string literal as
+    /// its argument.
+    func findStringArgument(label: String) -> String? {
+        for arg in arguments {
+            if arg.label?.text == label {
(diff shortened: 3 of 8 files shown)


DougGregor:
This depends on the SwiftPM changes in https://github.com/apple/swift-package-manager/pull/7474

DougGregor:
https://github.com/apple/swift-package-manager/pull/7474

DougGregor:
https://github.com/apple/swift-package-manager/pull/7474

@swift-ci please test

Package.swift
```diff
@@ -309,6 +309,7 @@ let package = Package(
         .product(name: "SwiftRefactor", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
+        .product(name: "SwiftPM-auto", package: "swift-package-manager"),
```
ahoppen, line 312: I think you need to add the corresponding dependency to the CMake build as well.

Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
```diff
@@ -52,4 +52,31 @@ public struct TextDocumentEdit: Hashable, Codable, Sendable {
...
+  public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode("textDocumentEdit", forKey: .kind)
+    try container.encode(self.textDocument, forKey: .textDocument)
+    try container.encode(self.edits, forKey: .edits)
+  }
+
+  public init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    let kind = try container.decode(String.self, forKey: .kind)
```
ahoppen, line 71: Where is the `kind` and `textDocumentEdit` coming from? I can’t see it mentioned in the LSP spec anywhere.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
ahoppen, line 1: Could you add test cases for these refactorings?

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,166 @@
...
+      let token = scope.firstToken,
+      let call = token.findEnclosingCall()
+    else {
+      return []
+    }
+
+    var actions = [CodeAction]()
+
+    // If there's a target name, offer to create a test target derived from it.
+    if let targetName = call.findStringArgument(label: "name") {
```
ahoppen, line 33: Should we also check that we are indeed at a `target`? For example I think we shouldn’t offer this when invoking the refactoring on the call to the `Package` initializer or on a `testTarget`.

Package.swift
DougGregor: I must be missing something. The CMake targets for other things that use SwiftPM don't list a specific dependency on anything in SwiftPM.

Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
DougGregor: Ah, good catch. I added these based on `CreateFile`, but the actual issue was somewhere else ([OptionalVersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#optionalVersionedTextDocumentIdentifier)). I've removed these.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: Yeah, I want to validate that we're on a kind of target we know how to build a test for.

DougGregor: Done in https://github.com/apple/sourcekit-lsp/pull/1193/commits/4c1e119197f5997af6d9cc6e6f5a404d09e4e470

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: Done in https://github.com/apple/sourcekit-lsp/pull/1193/commits/4c1e119197f5997af6d9cc6e6f5a404d09e4e470

Sources/LanguageServerProtocol/SupportTypes/TextDocumentEdit.swift
DougGregor: Done in https://github.com/apple/sourcekit-lsp/pull/1193/commits/4c1e119197f5997af6d9cc6e6f5a404d09e4e470

DougGregor:
@swift-ci please test

DougGregor:
https://github.com/apple/swift-package-manager/pull/7477

@swift-ci please test

DougGregor:
https://github.com/apple/swift-package-manager/pull/7477

@swift-ci please test

DougGregor:
https://github.com/apple/swift-package-manager/pull/7477

@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift-package-manager/pull/7477

@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

Package.swift
ahoppen: If it builds, whatever. I don’t know how this CMake build works.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+
+    return actions
+  }
+
+  /// Produce code actions to add test target(s) if we are currently on
+  /// a target for which we know how to create a test.
+  static func maybeAddTestTargetActions(
+    call: FunctionCallExprSyntax,
+    in scope: SyntaxCodeActionScope
+  ) -> [CodeAction] {
```
ahoppen, line 49: It’s not really “adding” actions to anything right? What about just adding `addTestTargetTestActions`?

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+    actions.append(
+      contentsOf: maybeAddTestTargetActions(call: call, in: scope)
+    )
+
+    // Add product(s).
+    actions.append(
+      contentsOf: maybeAddProductActions(call: call, in: scope)
+    )
+
+    return actions
```
ahoppen, line 41: I think this could be simplified to just

```suggestion
    return maybeAddTestTargetActions(call: call, in: scope) + maybeAddProductActions(call: call, in: scope)
```

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+
+    do {
+      let type: ProductType =
+        calledMember == "executableTarget"
+        ? .executable
+        : .library(.automatic)
+
+      // Describe the target we are going to create.
+      let product = try ProductDescription(
+        name: "\(targetName)",
```
ahoppen, line 106: `targetName` is a `String`, right? So this can just be

```suggestion
        name: targetName,
```

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+      )
+
+      let edits = try AddTarget.addTarget(target, to: scope.file)
+      return [
+        CodeAction(
+          title: "Add test target",
+          kind: .refactor,
+          edit: edits.asWorkspaceEdit(snapshot: scope.snapshot)
+        )
+      ]
```
ahoppen, line 72: Would it make sense to do some rough check whether a test target with that name already exists? I think even if we just do a textual search for `"\(targetName)Tests"` (ie the target name wrapped in quotes), that would cover most cases and would prevent us from showing a code action to add a test target when it obviously already exists.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+        : .library(.automatic)
+
+      // Describe the target we are going to create.
+      let product = try ProductDescription(
+        name: "\(targetName)",
+        type: type,
+        targets: [targetName]
+      )
+
+      let edits = try AddProduct.addProduct(product, to: scope.file)
```
ahoppen, line 111: Similar here, do you think there’s a way to check if we already have a product for this target? I think textual search wouldn’t work here.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+      TextEdit(
+        range: snapshot.range(of: edit.range),
+        newText: edit.replacement
+      )
+    }
+
+    // If we couldn't figure out the manifest directory, or there are no
+    // files to add, the only changes are the manifest edits. We're done
+    // here.
+    let manifestDirectoryURI = snapshot.uri.fileURL?
```
ahoppen, line 147: Nitpick: This is a `URL`, not `DocumentUri`, so the variable should be named `manifestDirectoryURL` instead of `manifestDirectoryURI`

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+  static let targetsThatAllowProducts: Set<String> = [
+    "executableTarget",
+    "target",
+  ]
+}
+
+fileprivate extension PackageEditResult {
+  /// Translate package manifest edits into a workspace edit. This can
+  /// involve both modifications to the manifest file as well as the creation
+  /// of new files.
```
ahoppen, line 134:
```suggestion
  /// of new files.
  /// `snapshot` is the latest snapshot of the `Package.swift` file.
```

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+    // over the individual changes to documents.
+    var documentChanges: [WorkspaceEditDocumentChange] = []
+
+    // Put the manifest changes into the array.
+    documentChanges.append(
+      .textDocumentEdit(
+        .init(
+          textDocument: .init(snapshot.uri, version: nil),
+          edits: manifestTextEdits.map { .textEdit($0) }
+        )
```
ahoppen, line 165: Since we have an open snapshot for `Package.swift`, we should also send the version. From the LSP spec

> If an optional versioned text document identifier is sent from the server to the client and the file is not open in the editor (the server has not received an open notification before) the server can send `null` to indicate that the version is known and the content on disk is the master (as specified with document content ownership).

Which I read like you should only specify `nil` as the `version` if the file isn’t open in the editor, but in this case it is.

Also, nitpick: I generally dislike `.init` because while it’s easy to write, it’s harder to read, not knowing which type gets created.

```suggestion
        TextDocumentEdit(
          textDocument: OptionalVersionedTextDocumentIdentifier(snapshot.uri, version: snapshot.version),
          edits: manifestTextEdits.map { .textEdit($0) }
        )
```

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+      )
+    )
+
+    // Create an populate all of the auxiliary files.
+    for (relativePath, contents) in auxiliaryFiles {
+      guard
+        let uri = URL(
```
ahoppen, line 172: Again, nitpick: This is a `URL`, not a `DocumentURI`, so it should be named `url` instead of `uri`.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,254 @@
...
+        newText: contents.description
+      )
+
+      documentChanges.append(.createFile(createFile))
+      documentChanges.append(
+        .textDocumentEdit(
+          .init(
+            textDocument: .init(documentURI, version: nil),
+            edits: [.textEdit(edit)]
+          )
```
ahoppen, line 197: Similar nitpick here: I’d prefer to not use `.init`.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -482,4 +489,47 @@ final class CodeActionTests: XCTestCase {
...
+        return action.title == "Add test target"
+      }
+    )
+
+    XCTAssertTrue(
+      codeActions.contains { action in
+        return action.title == "Add product to export this target"
+      }
+    )
+  }
```
ahoppen, line 534: I think it would also be good to have tests for:
- Getting code actions on a `.testTarget` and ensure that we don’t get a code action to add a test target for it
- I think it makes sense to test for at least one code action that we get edits (so we test `asWorkspaceEdit`)
- I think we shouldn’t be giving these code actions in files that aren’t a package manifest, eg.
```swift
struct Foo {
  func target(name: String) {}
  func test() {
    self.target(name: "test")
  }
}
```

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: Eh, I guess so.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: We might regret it if we end up with a bunch of these, but okay

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: If we're going to do that check, it should happen down in the SwiftPM code. Personally, I'm not sure... someone might want to use the action and then go rename the text. Are we helping them by taking the action away?

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: I agree on the `TextDocumentEdit`, but `OptionalVersionedTextDocumentIdentifier`... yuck.

Tests/SourceKitLSPTests/CodeActionTests.swift
DougGregor: SwiftPM does the checking to make sure we have a legitimate package file; we don't need to do it here.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift-package-manager/pull/7498

@swift-ci please test Windows

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
ahoppen: That’s a good point. Let’s leave it for now 👍🏽

DougGregor:
https://github.com/apple/swift/pull/73267
https://github.com/apple/swift-package-manager/pull/7498

@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift/pull/73267
https://github.com/apple/swift-package-manager/pull/7502

@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift/pull/73267
https://github.com/apple/swift-package-manager/pull/7502

@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift-package-manager/pull/7498
https://github.com/apple/swift/pull/73267

@swift-ci please test Windows

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,243 @@
...
+          edit: edits.asWorkspaceEdit(snapshot: scope.snapshot)
+        )
+      ]
+    } catch {
+      return []
+    }
+  }
+
+  /// A list of target kinds that allow the creation of tests.
+  static let targetsThatAllowTests: Set<String> = [
```
karwa, line 67: I know this isn't particularly performance-sensitive, but this should probably be an Array. A Set is overkill.

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
```diff
@@ -0,0 +1,243 @@
...
+          edit: edits.asWorkspaceEdit(snapshot: scope.snapshot)
+        )
+      ]
+    } catch {
+      return []
+    }
+  }
+
+  /// A list of target kinds that allow the creation of tests.
+  static let targetsThatAllowProducts: Set<String> = [
```
karwa, line 113: Same here. An Array would probably be better.

DougGregor:
https://github.com/apple/swift-package-manager/pull/7509

@swift-ci clean test Windows

DougGregor:
https://github.com/apple/swift-syntax/pull/2628/
https://github.com/apple/swift-package-manager/pull/7505

@swift-ci clean test Windows

Sources/SourceKitLSP/Swift/CodeActions/PackageManifestEdits.swift
DougGregor: Yes, you're right. I'll clean it up in a subsequent PR, because I'm drowning in unmerged PRs right now

PR #1194, opened 2024-04-20
Filter index entries for deleted source files

ahoppen:
This introduces an abstraction layer around `IndexStoreDB` (yes, I known another one …) that consults the underlying `IndexStoreDB` and then checks whether the entries are up-to-date with respect to some `IndexCheckLevel`. Requests can specify how picky they want to be about declaring results from the index out-of-date.

The default choice right now is to not include index entries from source files that have been deleted (because those are definitely from lingering unit files) but include results even if the source file has been modified after it was last indexed. We do include results from files that have been modified since they were last indexed because: When a file gets modified, it's likely that some of the line:column locations in it are still correct – eg. if only one line is modified and if lines are inserted/deleted all locations above are still correct. For locations that are out of date, showing stale results is one of the best ways of communicating to the user that the index is out-of-date and that they need to rebuild. We might want to reconsider this default when we have background indexing.

rdar://125230833
rdar://126622963


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -257,15 +254,14 @@ extension SourceKitLSPServer {
 
     let syntacticTests = try await languageService.syntacticDocumentTests(for: req.textDocument.uri)
 
-    if let index = workspace.index {
+    if let index = workspace.index(checkedFor: .imMemoryModifiedFiles(documentManager)) {
```
bnbarham, line 257:
```suggestion
    if let index = workspace.index(checkedFor: .inMemoryModifiedFiles(documentManager)) {
```
Weirdly this is the first time I noticed the misspelling (despite there being many before this).

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

PR #1195, opened 2024-04-20
Enhance syntactic test discovery with information from the semantic index

ahoppen:
When the semantic index is out-of-date, we currently purely rely on the syntactic index to discover tests and completely ignore data from the semantic index. This may lead to confusing behavior. For example if you have

```
class MightInheritFromXCTestCaseOrNot {}

class MyClass: MightInheritFromXCTestCaseOrNot {
  func testStuff() {}
}
```

Then we don’t return any tests when the semantic index is up-to-date. But once the file is modified (either on disk or in-memory), we purely rely on the syntactic index, which reports `testStuff` as a test method. After a build / background indexing finishes, the test method disappears again.

We can mitigate this problem as follows: If we have stale semantic index data for the test file, for every test method found by the syntactic index, check if we have an entry for this method in the semantic index. If we do, but that entry is not marked as a test class/method, we know that the semantic index knows about this method but decided that it’s not a test method for some reason. So we should ignore it.

rdar://126492948

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/TestItem.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/TestItem.swift
@@ -12,7 +12,7 @@
 
 public struct TestTag: Codable, Equatable, Sendable {
   /// ID of the test tag. `TestTag` instances with the same ID are considered to be identical.
-  public let id: String
+  public var id: String
 
   public init(id: String) {
     self.id = id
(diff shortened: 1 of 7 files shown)


bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1198, opened 2024-04-23
Minor improvements to the diagnose subcommand

ahoppen:
- Change name of request in reproducer bundle from `request.json` to `request.yml` because it’s a YAML file, not a JSON file
- Change names of diagnose components to use dashes instead of camel case


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1200, opened 2024-04-23
Skip a few hover tests if sourcekitd doesn’t return raw documentation in cursor info

ahoppen:
The formatting of these hover responses contains a newline if we don’t have a raw documentation response. Skip these tests if sourcekitd doesn’t return the raw documentation in cursor info, like we do with other tests in this file.

This ensures that all tests are passing in Xcode 15.4 again.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1203, opened 2024-04-24
Skip new tests that use Swift modules if the host’s SwiftPM doesn’t store `.swiftmodule` files in a subdirectory

ahoppen:
This fixes a test failure when running tests using Xcode 15.3.


ahoppen:
@swift-ci Please test

PR #1204, opened 2024-04-24
Elevate the log level for messages that record which toolchain was used for a document

ahoppen:
This is some fairly essential information for troubleshooting and should be logged at the same level as the build settings for a file.


bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1205, opened 2024-04-26
Add a syntactic "Add Codable structs from JSON" code action

DougGregor:
Add a syntactic action that takes JSON pasted into a Swift file or
placed in a string literal, then turns it into a set of Codable
structs that can represent the JSON. Our typical example starts like
this:

```
{
    "name": "Produce",
    "shelves": [
        {
            "name": "Discount Produce",
            "product": {
                "name": "Banana",
                "points": 200,
                "description": "A banana that's perfectly ripe."
            }
        }
    ]
}
```

and turns into this:

```swift
struct JSONValue: Codable {
    var name: String
    var shelves: [Shelves]

    struct Shelves: Codable {
        var name: String
        var product: Product

        struct Product: Codable {
            var description: String
            var name: String
            var points: Double
        }
    }
}
```

The refactoring itself would live down in the swift-syntax package if
not for its dependency on Foundation. We'll move as appropriate.

### DIFF
--- a/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
@@ -88,7 +88,8 @@ public struct ConvertJSONToCodableStruct: EditRefactoringProvider {
 
     // Try to process this as JSON.
     guard
-      let object = try? JSONSerialization.jsonObject(with: text.data(using: .utf8)!),
+      let data = text.data(using: .utf8),
+      let object = try? JSONSerialization.jsonObject(with: data),
       let dictionary = object as? [String: Any]
     else {
       return []
@@ -131,12 +132,10 @@ public struct ConvertJSONToCodableStruct: EditRefactoringProvider {
       ]
     }
   }
-}
 
-extension ConvertJSONToCodableStruct {
   /// The result of preflighting a syntax node to try to find potential JSON
   /// in it.
-  enum Preflight {
+  private enum Preflight {
     /// A closure, which is what a JSON dictionary looks like when pasted
     /// into Swift.
     case closure(ClosureExprSyntax)
@@ -150,7 +149,7 @@ extension ConvertJSONToCodableStruct {
   }
 
   /// Look for either a closure or a string literal that might have JSON in it.
-  static func preflightRefactoring(_ syntax: Syntax) -> Preflight? {
+  private static func preflightRefactoring(_ syntax: Syntax) -> Preflight? {
     // Preflight a closure.
     //
     // A blob of JSON dropped into a Swift source file will look like a
@@ -166,7 +165,7 @@ extension ConvertJSONToCodableStruct {
     }
 
     // We found a string literal; its contents might be JSON.
-    if let stringLit = syntax.as(StringLiteralExprSyntax.self) {
+    if let stringLiteral = syntax.as(StringLiteralExprSyntax.self) {
       // Look for an enclosing context and prefer that, because we might have
       // a string literal that's inside a closure where the closure itself
       // is the JSON.
@@ -176,11 +175,11 @@ extension ConvertJSONToCodableStruct {
         return enclosingPreflight
       }
 
-      guard let text = stringLit.representedLiteralValue else {
+      guard let text = stringLiteral.representedLiteralValue else {
         return nil
       }
 
-      return .stringLiteral(stringLit, text)
+      return .stringLiteral(stringLiteral, text)
     }
 
     // Look further up the syntax tree.
@@ -216,9 +215,7 @@ fileprivate struct JSONObject {
   /// a JSON object
   func merging(with other: JSONObject) -> JSONObject {
     // Collect the set of all keys from both JSON objects.
-    var allKeys: Set<String> = []
-    allKeys.formUnion(fields.keys)
-    allKeys.formUnion(other.fields.keys)
+    let allKeys = Set(fields.keys).union(other.fields.keys)
 
     // Form a new JSON object containing the union of the fields
     let newFields = allKeys.map { key in
@@ -235,7 +232,7 @@ fileprivate struct JSONObject {
     let sortedFields = fields.sorted(by: { $0.key < $1.key })
 
     // Collect the nested types
-    let nestedTypes: [(String, JSONObject)] = sortedFields.compactMap { (name, type) in
+    let nestedTypes: [(name: String, type: JSONObject)] = sortedFields.compactMap { (name, type) in
       guard let object = type.innerObject else {
         return nil
       }
@@ -255,7 +252,7 @@ fileprivate struct JSONObject {
       // Print any nested types.
       for (typeName, object) in nestedTypes {
         MemberBlockItemSyntax(
-          leadingTrivia: (typeName == nestedTypes.first?.0) ? .newlines(2) : .newline,
+          leadingTrivia: (typeName == nestedTypes.first?.name) ? .newlines(2) : .newline,
           decl: object.asDeclSyntax(name: typeName)
         )
       }
@@ -298,16 +295,13 @@ fileprivate enum JSONType {
   init(value: Any) {
     switch value {
     case let string as String:
-      if string == "true" || string == "false" {
-        self = .boolean
-      } else {
-        self = .string
+      switch string {
+      case "true", "false": self = .boolean
+      default: self = .string
       }
     case is NSNumber:
       self = .number
-    case is NSArray:
-      let array = value as! [Any]
-
+    case let array as [Any]:
       // Use null as a fallback for an empty array.
       guard let firstValue = array.first else {
         self = .array(.null)
@@ -324,8 +318,8 @@ fileprivate enum JSONType {
 
     case is NSNull:
       self = .null
-    case is NSDictionary:
-      self = .object(JSONObject(dictionary: value as! [String: Any]))
+    case let dictionary as [String: Any]:
+      self = .object(JSONObject(dictionary: dictionary))
     default:
       self = .string
     }
--- a/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
@@ -78,6 +78,9 @@ public struct ConvertJSONToCodableStruct: EditRefactoringProvider {
       /// We're only going to look at the text of the closure to see if we
       /// have JSON in there.
       text = closure.trimmedDescription
+    case let .endingClosure(closure, unexpected):
+      text = closure.trimmedDescription + unexpected.description
+
     case .stringLiteral(_, let literalText):
       /// A string literal that could contain JSON within it.
       text = literalText
@@ -108,6 +111,15 @@ public struct ConvertJSONToCodableStruct: EditRefactoringProvider {
       return [
         SourceEdit(range: closure.trimmedRange, replacement: decls.description)
       ]
+    case .endingClosure(let closure, let unexpected):
+      // Closures are replaced entirely, since they were invalid code to
+      // start with.
+      return [
+        SourceEdit(
+          range: closure.positionAfterSkippingLeadingTrivia..<unexpected.endPosition,
+          replacement: decls.description
+        )
+      ]
     case .stringLiteral(let literal, _):
       /// Leave the string literal in place (it might be there for testing
       /// purposes), and put the newly-created structs afterward.
@@ -129,6 +141,10 @@ extension ConvertJSONToCodableStruct {
     /// into Swift.
     case closure(ClosureExprSyntax)
 
+    /// A closure with a bunch of unexpected nodes following it, which is what
+    /// a big JSON dictionary looks like when pasted into Swift.
+    case endingClosure(ClosureExprSyntax, UnexpectedNodesSyntax)
+
     /// A string literal that may contain JSON.
     case stringLiteral(StringLiteralExprSyntax, String)
   }
@@ -141,6 +157,11 @@ extension ConvertJSONToCodableStruct {
     // closure due to the curly braces. The internals might be a syntactic
     // disaster, but we don't actually care.
     if let closure = syntax.as(ClosureExprSyntax.self) {
+      if let file = closure.parent?.parent?.parent?.as(SourceFileSyntax.self),
+        let unexpected = file.unexpectedBetweenStatementsAndEndOfFileToken
+      {
+        return .endingClosure(closure, unexpected)
+      }
       return .closure(closure)
     }
 
--- a/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
@@ -1,0 +1,389 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+import SwiftBasicFormat
+import SwiftRefactor
+import SwiftSyntax
+
+/// Convert JSON literals into corresponding Swift structs that conform to the
+/// `Codable` protocol.
+///
+/// ## Before
+///
+/// ```javascript
+/// {
+///   "name": "Produce",
+///   "shelves": [
+///     {
+///       "name": "Discount Produce",
+///       "product": {
+///         "name": "Banana",
+///         "points": 200,
+///         "description": "A banana that's perfectly ripe."
+///       }
+///     }
+///   ]
+/// }
+/// ```
+///
+/// ## After
+///
+/// ```swift
+/// struct JSONValue: Codable {
+///   var name: String
+///   var shelves: [Shelves]
+///
+///   struct Shelves: Codable {
+///     var name: String
+///     var product: Product
+///
+///     struct Product: Codable {
+///       var description: String
+///       var name: String
+///       var points: Double
+///     }
+///   }
+/// }
+/// ```
+@_spi(Testing)
+public struct ConvertJSONToCodableStruct: EditRefactoringProvider {
+  @_spi(Testing)
+  public static func textRefactor(
+    syntax: Syntax,
+    in context: Void
+  ) -> [SourceEdit] {
+    // Dig out a syntax node that looks like it might be JSON or have JSON
+    // in it.
+    guard let preflight = preflightRefactoring(syntax) else {
+      return []
+    }
+
+    // Dig out the text that we think might be JSON.
+    let text: String
+    switch preflight {
+    case let .closure(closure):
+      /// The outer structure of the JSON { ... } looks like a closure in the
+      /// syntax tree, albeit one with lots of ill-formed syntax in the body.
+      /// We're only going to look at the text of the closure to see if we
+      /// have JSON in there.
+      text = closure.trimmedDescription
+    case .stringLiteral(_, let literalText):
+      /// A string literal that could contain JSON within it.
+      text = literalText
+    }
+
+    // Try to process this as JSON.
+    guard
+      let object = try? JSONSerialization.jsonObject(with: text.data(using: .utf8)!),
+      let dictionary = object as? [String: Any]
+    else {
+      return []
+    }
+
+    // Create the top-level object.
+    let topLevelObject = JSONObject(dictionary: dictionary)
+
+    // Render the top-level object as a struct.
+    let indentation = BasicFormat.inferIndentation(of: syntax)
+    let format = BasicFormat(indentationWidth: indentation)
+    let decls = topLevelObject.asDeclSyntax(name: "JSONValue")
+      .formatted(using: format)
+
+    // Render the change into a set of source edits.
+    switch preflight {
+    case .closure(let closure):
+      // Closures are replaced entirely, since they were invalid code to
+      // start with.
+      return [
+        SourceEdit(range: closure.trimmedRange, replacement: decls.description)
+      ]
+    case .stringLiteral(let literal, _):
+      /// Leave the string literal in place (it might be there for testing
+      /// purposes), and put the newly-created structs afterward.
+      return [
+        SourceEdit(
+          range: literal.endPosition..<literal.endPosition,
+          replacement: "\n" + decls.description
+        )
+      ]
+    }
+  }
+}
+
+extension ConvertJSONToCodableStruct {
+  /// The result of preflighting a syntax node to try to find potential JSON
+  /// in it.
+  enum Preflight {
+    /// A closure, which is what a JSON dictionary looks like when pasted
+    /// into Swift.
+    case closure(ClosureExprSyntax)
+
+    /// A string literal that may contain JSON.
+    case stringLiteral(StringLiteralExprSyntax, String)
+  }
+
+  /// Look for either a closure or a string literal that might have JSON in it.
+  static func preflightRefactoring(_ syntax: Syntax) -> Preflight? {
+    // Preflight a closure.
+    //
+    // A blob of JSON dropped into a Swift source file will look like a
+    // closure due to the curly braces. The internals might be a syntactic
+    // disaster, but we don't actually care.
+    if let closure = syntax.as(ClosureExprSyntax.self) {
+      return .closure(closure)
+    }
+
+    // We found a string literal; its contents might be JSON.
+    if let stringLit = syntax.as(StringLiteralExprSyntax.self) {
+      // Look for an enclosing context and prefer that, because we might have
+      // a string literal that's inside a closure where the closure itself
+      // is the JSON.
+      if let parent = syntax.parent,
+        let enclosingPreflight = preflightRefactoring(parent)
+      {
+        return enclosingPreflight
+      }
+
+      guard let text = stringLit.representedLiteralValue else {
+        return nil
+      }
+
+      return .stringLiteral(stringLit, text)
+    }
+
+    // Look further up the syntax tree.
+    if let parent = syntax.parent {
+      return preflightRefactoring(parent)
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Linux

DougGregor:
@swift-ci please test Linux

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+///         {
+///             "name": "Discount Produce",
+///             "product": {
+///                 "name": "Banana",
+///                 "points": 200,
+///                 "description": "A banana that's perfectly ripe."
+///             }
+///         }
+///     ]
+/// }
```
ahoppen, line 36: Could we make this two spaces to match the rest?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+    // in it.
+    guard let preflight = preflightRefactoring(syntax) else {
+      return []
+    }
+
+    // Dig out the text that we think might be JSON.
+    let text: String
+    switch preflight {
+    case let .closure(closure):
+      text = closure.trimmedDescription
```
ahoppen, line 75: I think it would be good to have a short comment about how the JSON struct is represented as a closure in the syntax tree since it’s not obvious why we are operating on a `ClosureExprSyntax` here when we’re doing JSON to struct refactoring.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+      let object = try? JSONSerialization.jsonObject(with: text.data(using: .utf8)!),
+      let serial = object as? Dictionary<String, Any>
+    else {
+      return []
+    }
+
+    // Build the declarations from the JSON dictionary.
+    let decls = build(from: serial)
+
+    // Render the change into
```
ahoppen, line 93: Stoped typing the comment midway? 😉

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+    case let .tail(closure, unexpected):
+      text = closure.trimmedDescription + unexpected.description
+    case .stringLiteral(_, let literalText):
+      text = literalText
+    }
+
+    // Try to process this as JSON.
+    guard
+      let object = try? JSONSerialization.jsonObject(with: text.data(using: .utf8)!),
+      let serial = object as? Dictionary<String, Any>
```
ahoppen, line 85: Why `serial`? Wouldn’t something like `dictionary` make more sense?


Also, I’d use `[String: Any]`

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+        SourceEdit(
+          range: literal.endPosition..<literal.endPosition,
+          replacement: "\n" + decls.description
+        )
+      ]
+    }
+  }
+}
+
+extension ConvertJSONToCodableStruct {
```
ahoppen, line 115: Let’s just shove them all together instead of having `extension`s of the same type within the same file. 

```suggestion

```

Same once more a few lines below.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+    // Preflight a closure.
+    //
+    // A blob of JSON dropped into a Swift source file will look like a
+    // closure due to the curly braces. The internals might be a syntactic
+    // disaster, but we don't actually care.
+    if let closure = syntax.as(ClosureExprSyntax.self) {
+      if let file = closure.parent?.parent?.parent?.as(SourceFileSyntax.self),
+        let unexpected = file.unexpectedBetweenStatementsAndEndOfFileToken
+      {
+        return .tail(closure, unexpected)
```
ahoppen, line 141: Do you have an example where this is relevant? Shouldn’t the braces from the JSON object match the braces of the closure in this case as well?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+}
+
+extension ConvertJSONToCodableStruct: SyntaxRefactoringCodeActionProvider {
+  static var title = "Create Codable structs from JSON"
+}
+
+extension ConvertJSONToCodableStruct {
+  fileprivate static let indentation = 4
+
+  private static func build(from jsonDictionary: Dictionary<String, Any>) -> DeclSyntax {
```
ahoppen, line 172:
```suggestion
  private static func build(from jsonDictionary: [String: Any]) -> DeclSyntax {
```
 
Comment for `Dictionary<String, Any>` applies to a few more places in the file as well.

---

Also: Do we need this function at all? Doesn’t seem like it’s doing a whole lot and I think it’s only called once.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+    return
+      """
+      \(raw: self.buildStruct(from: jsonDictionary))
+      """
+  }
+
+  private static func buildStruct(
+    named name: String = "JSONValue",
+    at depth: Int = 0,
```
ahoppen, line 181: I would make these parameters non-defaulted. We only use the default values in the initial entry point, which could specify them manually and then we don’t run into the problem of accidentally using the defaulted values where we don’t want to, just because we forget to specify the argument.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  private static func buildStructMembers(
+    at depth: Int,
+    from jsonDictionary: Dictionary<String, Any>
+  ) -> [String] {
+    var members: [String] = []
+
+    var uniquer: Set<String> = []
+    var nestedTypes: [(String, Dictionary<String, Any>)] = []
```
ahoppen, line 200: Why can’t we use a `[String: [String: Any]]` here instead of `uniquer` and an array containing a tuple? We do sort the keys from the `jsonDictionary` anyway, so I think there’s no order we need to preserver (if we sort by the dictionary keys in the end).

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+      guard uniquer.insert(name).inserted else {
+        return
+      }
+      nestedTypes.append((name, object))
+    }
+
+    for key in jsonDictionary.keys.sorted() {
+      guard let value = jsonDictionary[key] else {
+        continue
+      }
```
ahoppen, line 211: I think this can be simplified

```suggestion
    for (key, value) in jsonDictionary.sorted(\.key) {
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+          addNestedType(elementType, object: subObject)
+        }
+      }
+
+      let member = "\(String(repeating: " ", count: depth * indentation))var \(key): \(type.stringValue)"
+      members.append(member)
+    }
+
+    if !nestedTypes.isEmpty {
+      members.append("")
```
ahoppen, line 230: Having to append an empty string to members to (presumably) bet an empty line between the members seems wrong. I think we should add an empty line between all the members instead.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  indirect enum JSONType {
+    case string
+    case double
+    case array(JSONType)
+    case null
+    case object(String)
+
+    var outermostObject: (Int, String)? {
```
ahoppen, line 245: Instead of returning `unwraps` here and then calling `unwrap` with it, wouldn’t it be easier to just return the name and `JSONType.object` here?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+      case let .object(name):
+        return name
+      }
+    }
+  }
+
+  private static func jsonType(of value: Any, suggestion name: String) -> JSONType {
+    switch Swift.type(of: value) {
+    case is NSString.Type:
+      return .string
```
ahoppen, line 279: Do we need to use `Swift.type(of:)` here? Can’t we just use

```suggestion
    switch value {
    case is NSString.self:
      return .string
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+    case is NSNumber.Type:
+      return .double
+    case is NSArray.Type:
+      guard let firstValue = (value as! [Any]).first else {
+        return .array(.null)
+      }
+      let innerType = self.jsonType(of: firstValue, suggestion: name)
+      return .array(innerType)
```
ahoppen, line 287: Hmm, I think it’s unfortunate that we don’t allow merging the keys from multiple objects in the array. I don’t think it’s uncommon for JSON to omit keys for null values. We should be able to take the union of all the keys in the array. And I think it would also be good to infer optionality of a value eg. from

```json
[
  {
    "value": 1
  },
  {
    "value": null
  }
]
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+      return .null
+    case is NSDictionary.Type:
+      return .object(name.capitalized)
+    default:
+      return .string
+    }
+  }
+}
+
+extension Array where Element == Any {
```
ahoppen, line 298:
```suggestion
extension Array<Any> {
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,318 @@
...
+  fileprivate func unwrap(_ depth: Int) -> Dictionary<String, Any>? {
```
ahoppen, line 299: Could you add a comment of what `unwrap` does?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -292,6 +292,48 @@ final class CodeActionTests: XCTestCase {
     XCTAssertTrue(codeActions.contains(expectedCodeAction))
   }
 
+  func testJSONCodableCodeActionResult() async throws {
+    let testClient = try await TestSourceKitLSPClient(capabilities: clientCapabilitiesWithCodeActionSupport())
+    let uri = DocumentURI.for(.swift)
+    let positions = testClient.openDocument(
+      """
+         1️⃣{
```
ahoppen, line 300: Nitpick

```suggestion
     1️⃣{
```

Also, 2 spaces indentation would be nice here for consistency.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -0,0 +1,332 @@
...
+///   - line: The line number on which failure occurred. Defaults to the line number on which this
+///     function was called.
+public func assertStringsEqualWithDiff(
+  _ actual: String,
+  _ expected: String,
+  _ message: String = "",
+  additionalInfo: @autoclosure () -> String? = nil,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) {
```
ahoppen, line 210: If we’re adding the string diff functions, could you put them in `SKTestSupport` so that other tests can also use them?

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -0,0 +1,332 @@
...
+  let stringComparison: String
+
+  // Use `CollectionDifference` on supported platforms to get `diff`-like line-based output. On
+  // older platforms, fall back to simple string comparison.
+  if #available(macOS 10.15, *) {
```
ahoppen, line 272: The `#available` check isn’t needed since sourcekit-lsp targets macOS 13.

ahoppen (commented):
Copied a few review comments from https://github.com/apple/sourcekit-lsp/pull/1169 😉

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Yeah, fixed.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Sure, done.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Sure

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: I want the protocol-conforming extensions separate, but other than that---sure

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: I'm having trouble triggering this, so I'm going to remove it outright

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Nah, we can remove it.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
I just went ahead and rewrote this with a completely new structure that does merging of like elements in the whole tree. When merging a field w/ a missing value (or one explicitly marked `null`), we introduce an optional in Swift. This also handles Boolean values specially.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+    //
+    // A blob of JSON dropped into a Swift source file will look like a
+    // closure due to the curly braces. The internals might be a syntactic
+    // disaster, but we don't actually care.
+    if let closure = syntax.as(ClosureExprSyntax.self) {
+      return .closure(closure)
+    }
+
+    // We found a string literal; its contents might be JSON.
+    if let stringLit = syntax.as(StringLiteralExprSyntax.self) {
```
ahoppen, line 169: Nitpick

```suggestion
    if let stringLiteral = syntax.as(StringLiteralExprSyntax.self) {
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -299,6 +299,48 @@ final class CodeActionTests: XCTestCase {
...
+         "name": "Produce",
+         "shelves": [
```
ahoppen, line 309: I think right now it’s only possible to invoke the refactoring from the opening `{`. I think it would be a lot more discoverable if you could invoke it from anywhere within the JSON.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+    /// A closure, which is what a JSON dictionary looks like when pasted
+    /// into Swift.
+    case closure(ClosureExprSyntax)
+
+    /// A string literal that may contain JSON.
+    case stringLiteral(StringLiteralExprSyntax, String)
+  }
+
+  /// Look for either a closure or a string literal that might have JSON in it.
+  static func preflightRefactoring(_ syntax: Syntax) -> Preflight? {
```
ahoppen, line 153: Could be `private`, right?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+        )
+      ]
+    }
+  }
+}
+
+extension ConvertJSONToCodableStruct {
+  /// The result of preflighting a syntax node to try to find potential JSON
+  /// in it.
+  enum Preflight {
```
ahoppen, line 139: Could be `fileprivate`, right?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
ahoppen: This is no longer a protocol-conforming extension 😜 Also, I’m not generally a fan of putting protocol conformances in extensions. I think it makes it harder to reason about which protocols a type conforms to.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+        JSONType(value: firstValue)
+      ) { (result, value) in
+        result.merging(with: JSONType(value: value))
+      }
+      self = .array(elementType)
+
+    case is NSNull:
+      self = .null
+    case is NSDictionary:
+      self = .object(JSONObject(dictionary: value as! [String: Any]))
```
ahoppen, line 328: Could we fail gracefully if the key is not a `String`? Emitting a `logger.fault` and either falling back to some value or failing the refactoring.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+      // Merging two arrays merges the element types
+      return .array(left.merging(with: right))
+
+    case (.object(let left), .object(let right)):
+      // Merging two arrays merges the element types
+      return .object(left.merging(with: right))
+
+    // Merging a string with a Boolean means we misinterpreted "true" or
+    // "false" as Boolean when it was meant as a string.
+    case (.string, .boolean), (.boolean, .string): return .string
```
ahoppen, line 362: The same is true for `.string` and `.null`, right? In case we mis-interpreted `"null"`.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+    fields = dictionary.mapValues { JSONType(value: $0) }
+  }
+
+  /// Merge the fields of this JSON object with another JSON object to produce
+  /// a JSON object
+  func merging(with other: JSONObject) -> JSONObject {
+    // Collect the set of all keys from both JSON objects.
+    var allKeys: Set<String> = []
+    allKeys.formUnion(fields.keys)
+    allKeys.formUnion(other.fields.keys)
```
ahoppen, line 221: Could be simplified to the following, right?

```suggestion
    var allKeys = Set(fields.keys).union(other.fields.keys)
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+
+  /// A value that is optional, for example because it is missing or null in
+  /// other cases.
+  indirect case optional(JSONType)
+
+  /// Determine the type of a JSON value.
+  init(value: Any) {
+    switch value {
+    case let string as String:
+      if string == "true" || string == "false" {
```
ahoppen, line 301: Interesting that `JSONSerialization` doesn’t deserialize those to an `NSNumber`. That does make me wonder: Does `JSONSerialization` deserialize `"null"` as `NSNull` or also as a string with value `null`? I don’t think we’ve got a test for that.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+        )
+      }
+
+      // Print any nested types.
+      for (typeName, object) in nestedTypes {
+        MemberBlockItemSyntax(
+          leadingTrivia: (typeName == nestedTypes.first?.0) ? .newlines(2) : .newline,
+          decl: object.asDeclSyntax(name: typeName)
+        )
+      }
```
ahoppen, line 261: I know this is a subjective opinion, but I would always put the nested types before the members.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -0,0 +1,389 @@
...
+        MemberBlockItemSyntax(
+          leadingTrivia: .newline,
+          decl: "var \(raw: fieldName): \(fieldType.asTypeSyntax(name: fieldName))" as DeclSyntax
+        )
+      }
+
+      // Print any nested types.
+      for (typeName, object) in nestedTypes {
+        MemberBlockItemSyntax(
+          leadingTrivia: (typeName == nestedTypes.first?.0) ? .newlines(2) : .newline,
```
ahoppen, line 258: A little nitpicky: If you added labels to the declaration of `nestedTypes` the `.0` could be a proper label, which would be more readable.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

ahoppen (commented):
That’s a lot shorter and simpler and even implements merging of structs 🤩 

I have a few, mostly nitpicky, comments.

Tests/SourceKitLSPTests/CodeActionTests.swift
DougGregor: No, the refactoring can be invoked from elsewhere in the closure. The test harness doesn't make it easy to check for places other than the start, though.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: or `private`

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Yes. I want to kill the NS* type references here, too, as much as I can.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: Sure, then I can make it a `let`.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: It leaves them all as strings. A `"null"` is a dreadful thing because it messes with the merging logic.

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
DougGregor: I'm going to leave `"null"` as a string. We don't want to merge a `.null` and a `.string` as `.string`; it needs to be `.optional(.string)`.

Tests/SourceKitLSPTests/CodeActionTests.swift
DougGregor: Hmm, I see what's going on. The parser isn't treating the whole blob as a single closure, so we sometimes end up in the "unexpected" area. I don't have a great fix at the moment.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Tests/SourceKitLSPTests/CodeActionTests.swift
ahoppen: Could you file an issue for that?

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
ahoppen: Ah, fair point 👍🏽

Sources/SourceKitLSP/Swift/CodeActions/ConvertJSONToCodableStruct.swift
ahoppen: Is there any case that does produce an `NSNull` then?

PR #1208, opened 2024-04-26
Add a task scheduler for background indexing and preparation

ahoppen:
The scheduler isn’t actually being used yet but it’s complex and generic enough that it’s possible to review it by itself.

### DIFF
--- a/Sources/SKCore/TaskScheduler.swift
+++ b/Sources/SKCore/TaskScheduler.swift
@@ -1,0 +1,491 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import CAtomics
+import Foundation
+import LSPLogging
+
+/// See comment on ``TaskDescriptionProtocol/dependencies(to:taskPriority:)``
+public enum TaskDependencyAction<TaskDescription: TaskDescriptionProtocol> {
+  case waitAndElevatePriorityOfDependency(TaskDescription)
+  case cancelAndRescheduleDependency(TaskDescription)
+}
+
+public protocol TaskDescriptionProtocol: Identifiable, Sendable, CustomLogStringConvertible {
+  /// Execute the task.
+  ///
+  ///  - Important: This should only be called from `TaskScheduler` and never be called manually.
+  func execute() async
+
+  /// When a new task is picked for execution, this determines how the task should behave with respect to the tasks that
+  /// are already running.
+  ///
+  /// Options are the following (see doc comment on `TaskScheduler` for examples):
+  ///  1. Not add any `TaskDependencyAction` for a currently executing task. This means that the two tasks can run in
+  ///     parallel.
+  ///  2. Declare a `waitAndElevatePriorityOfDependency` dependency. This will prevent execution of this task until
+  ///     the other task has finished executing. It will elevate the priority of the dependency to the same priority as
+  ///     this task. This ensures that we don't get into a priority inversion problem where a high-priority task is
+  ///     waiting for a low-priority task.
+  ///  3. Declare a `cancelAndRescheduleDependency`. If the task dependency is idempotent and has a priority that's not
+  ///     higher than the this task's priority, this causes the task dependency to be cancelled, so that this task can
+  ///     execute. The canceled task will be scheduled to re-run at a later point.
+  ///     - Declaring a `cancelAndRescheduleDependency` dependency on a task that is not idempotent will change the
+  ///       dependency to a `waitAndElevatePriorityOfDependency` dependency and log a fault.
+  ///       A `cancelAndRescheduleDependency` dependency should never be emitted for a task that's not idempotent.
+  ///     - If the task that should be canceled and re-scheduled has a higher priority than this task, the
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SKCore/TaskScheduler.swift
```diff
@@ -0,0 +1,459 @@
...
+  Logger(subsystem: subsystem + ".task-scheduler", category: LoggingScope.scope)
+}
+
+/// See comment on ``TaskDescriptionProtocol/dependencies(to:taskPriority:)``
+public enum TaskDependencyAction<TaskDescription: TaskDescriptionProtocol> {
+  case waitAndElevatePriorityOfDependency(TaskDescription)
+  case suspendDependency(TaskDescription)
+}
+
+public protocol TaskDescriptionProtocol: Identifiable, Sendable, CustomLogStringConvertible {
```
ktoso, line 27: I'd love for you to file an issue on github.com/apple/swift so we can enable making such a thing as TaskExecutor, rather than having to resort to trickery and unstructured tasks like this. I believe we could allow such things but today are missing pieces in the custom executors infra I think -- like querying a job for its details etc, like priority, maybe some hooks to know a job "finished" etc.

Sources/SKCore/TaskScheduler.swift
```diff
@@ -0,0 +1,459 @@
...
+  }
+
+  /// If the priority of this task is less than `targetPriority`, elevate the priority to `targetPriority` by spawning
+  /// a new task that depends on it. Otherwise a no-op.
+  nonisolated func elevatePriority(to targetPriority: TaskPriority) {
+    if priority < targetPriority {
+      logger.debug(
+        "Elevating priority of \(self.description.forLogging) from \(self.priority.rawValue) to \(targetPriority.rawValue)"
+      )
+      Task.detached(priority: targetPriority) {
```
ktoso, line 228: probably doesn't need to detach -- plain task with priority should be fine -- and get task locals copied then

ktoso: If I remember dispatch behavior this will not necessarily be 100% reliable... even if you passed highest priority it may decide to use lower I think... but the general idea is okey I suppose.

Please file an issue about an API for `__priorityBoost(task, priority)` even if underscored I think we should offer this API.

ktoso (approved):
Overall I don't have big concerns here, let's start with that and see how it works out in practice. I'm going to give it another proper read though, and we can keep polishing if we find anything.

Sources/SKCore/TaskScheduler.swift
```diff
@@ -0,0 +1,459 @@
...
+    } else {
+      return SuspendableTaskFinishStatus.terminated
+    }
+  }
+
+  /// Try suspending this task.
+  ///
+  /// Suspension might not do anything if the task has already finished, ie. even though `suspend` was called, the
+  /// task might still
+  func suspend() {
```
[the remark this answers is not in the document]
ahoppen: Good call. Changed to `cancelToBeRescheduled`.

Sources/SKCore/TaskScheduler.swift
```diff
@@ -0,0 +1,459 @@
...
+    self.updatePriorityContinuation = updatePriorityContinuation
+
+    var suspendableTaskCreatedContinuation: AsyncStream<Task<SuspendableTaskFinishStatus, Never>>.Continuation!
+    let suspendableTaskCreatedStream = AsyncStream {
+      suspendableTaskCreatedContinuation = $0
+    }
+    self.suspendableTaskCreatedContinuation = suspendableTaskCreatedContinuation
+
+    self.resultTask = Task.detached(priority: priority) {
+      await withTaskGroup(of: Void.self) { taskGroup in
```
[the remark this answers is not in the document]
ahoppen: I don’t think that works because I need to cancel the first task in the task group (the one that is updating `self.priority`) when the second task in the task group (the one that runs tasks from `executionTaskCreatedStream`) finishes.

Sources/SKCore/TaskScheduler.swift
ahoppen: Filed https://github.com/apple/swift/issues/73369

Sources/SKCore/TaskScheduler.swift
ahoppen: Filed https://github.com/apple/swift/issues/73372. It’s a little vague, feel free to refine it with the things that you actually think are relevant and doable.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

Tests/SKCoreTests/TaskSchedulerTests.swift
```diff
@@ -0,0 +1,411 @@
...
+
+        for i in 0..<10 {
+          let id = TaskID.highPriority(i)
+          await scheduler.schedule(priority: .high, id: id) {
+            await taskExecutionRecorder.run(taskID: id)
+          }
+        }
+      },
+      validate: { (recordings: [Set<TaskID>]) in
+        // Check that all high-priority tasks get executed before the low-priority tasks
```
bnbarham, line 39: We *could* be executing some high priority and low priority at the same time in the actual check (eg. 2 high -> high + low -> low + low would/should pass).

bnbarham (approved):
Heh, I was always imagining just having separate queues (and pausing when there's high priority work + not caring if there's still low priority running at the same time). But this does give us more control. May not need that much in the end, we shall see.

ahoppen:
@swift-ci Please test Windows

Sources/SKCore/TaskScheduler.swift
ktoso: That's good, thank you -- I'll fill in the bits, and we'll figure out what is actionable :)

PR #1216, opened 2024-05-03
Implement initial background indexing of a project

ahoppen:
Implements an initial background index when the project is opened.

The following will be implemented in follow-up PRs:
- Resolving package dependencies
- Preparing dependent modules
- Watching for file updates

### DIFF
--- a/Sources/SKSupport/SwitchableProcessResultExitStatus.swift
+++ b/Sources/SKSupport/SwitchableProcessResultExitStatus.swift
@@ -1,0 +1,45 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// We need to import all of TSCBasic because otherwise we can't refer to Process.ExitStatus (rdar://127577691)
+import struct TSCBasic.ProcessResult
+
+/// Same as `ProcessResult.ExitStatus` in tools-support-core but has the same cases on all platforms and is thus easier
+/// to switch over
+public enum SwitchableProcessResultExitStatus {
+  /// The process was terminated normally with a exit code.
+  case terminated(code: Int32)
+  /// The process was terminated abnormally.
+  case abnormal(exception: UInt32)
+  /// The process was terminated due to a signal.
+  case signalled(signal: Int32)
+}
+
+extension ProcessResult.ExitStatus {
+  public var exhaustivelySwitchable: SwitchableProcessResultExitStatus {
+    #if os(Windows)
+    switch self {
+    case .terminated(let code):
+      return .terminated(code: code)
+    case .abnormal(let exception):
+      return .abnormal(exception: exception)
+    }
+    #else
+    switch self {
+    case .terminated(let code):
+      return .terminated(code: code)
+    case .signalled(let signal):
+      return .signalled(signal: signal)
+    }
+    #endif
+  }
+}
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -1,0 +1,167 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LSPLogging
+import LanguageServerProtocol
+import SKCore
+
+/// Describes the state of indexing for a single source file
+private enum FileIndexStatus {
+  /// The index is up-to-date.
+  case upToDate
+  /// The file is being indexed by the given task.
+  case inProgress(Task<Void, Never>)
+}
+
+/// Schedules index tasks and keeps track of the index status of files.
+public final actor SemanticIndexManager {
+  /// The underlying index. This is used to check if the index of a file is already up-to-date, in which case it doesn't
+  /// need to be indexed again.
+  private let index: CheckedIndex
+
+  /// The build system manager that is used to get compiler arguments for a file.
+  private let buildSystemManager: BuildSystemManager
+
+  /// The index status of the source files that the `SemanticIndexManager` knows about.
+  ///
+  /// Files that have never been indexed are not in this dictionary.
+  private var indexStatus: [DocumentURI: FileIndexStatus] = [:]
+
+  /// The `TaskScheduler` that manages the scheduling of index tasks. This is shared among all `SemanticIndexManager`s
+  /// in the process, to ensure that we don't schedule more index operations than processor cores from multiple
+  /// workspaces.
+  private let indexTaskScheduler: TaskScheduler<UpdateIndexStoreTaskDescription>
+
+  /// Callback that is called when an index task has finished.
+  ///
+  /// Currently only used for testing.
+  private let indexTaskDidFinish: (@Sendable (UpdateIndexStoreTaskDescription) -> Void)?
+
+  // MARK: - Public API
+
+  public init(
+    index: UncheckedIndex,
+    buildSystemManager: BuildSystemManager,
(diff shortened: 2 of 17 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

SolaWing:
is it support other BSP or only Swift Package Manager?

ahoppen:
To start off, I’m only aiming to support SwiftPM. I just opened https://github.com/apple/sourcekit-lsp/issues/1226 from my local to-do list so we can share ideas there. If you could comment there which BSP server you are using with SourceKit-LSP, that would be valuable.

ahoppen:
Merging this to unblock further work on background indexing. Will address review comments in follow-up PRs.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  private func updateIndexStoreForSingleFile(_ uri: DocumentURI) async {
+    guard let url = uri.fileURL else {
+      // The URI is not a file, so there's nothing we can index.
+      return
+    }
+    guard !index.hasUpToDateUnit(for: url) else {
+      // We consider a file's index up-to-date if we have any up-to-date unit. Changing build settings does not
+      // invalidate the up-to-date status of the index.
```
hamishknight, line 126: I guess you have to `touch` the file or something if you want to index it with up-to-date build settings?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,311 @@
...
+      buildSettings.compilerArguments,
+      fileToIndex: uri
+    )
+
+    guard let swiftc = toolchain.swiftc else {
+      logger.error(
+        "Not updating index store for \(uri.forLogging) because toolchain \(toolchain.identifier) does not contain a Swift compiler"
+      )
+      return
+    }
```
hamishknight, line 189: Could swap these

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -0,0 +1,167 @@
...
+      if case .upToDate = indexStatus[$0] {
+        return false
+      }
+      return true
+    }
+
+    var indexTasks: [Task<Void, Never>] = []
+
+    // TODO (indexing): Group index operations by target when we support background preparation.
+    for files in outOfDateFiles.partition(intoNumberOfBatches: ProcessInfo.processInfo.processorCount * 5) {
```
hamishknight, line 131: Worth a comment explaining the batching size? Or does it not matter since this will change once we support preparation?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,311 @@
...
+    result.append(argument)
+  }
+  result += [
+    "-index-file",
+    "-index-file-path", fileToIndex.pseudoPath,
+    // batch mode is not compatible with -index-file
+    "-disable-batch-mode",
+    // Fake an output path so that we get a different unit file for every Swift file we background index
+    "-o", fileToIndex.pseudoPath + ".o",
```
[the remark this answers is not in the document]
hamishknight: Also just noticed this doesn't seem to respect IndexOptions's `indexStorePath`, is that intentional?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -0,0 +1,167 @@
...
+  public func waitForUpToDateIndex(for uris: some Collection<DocumentURI>) async {
+    logger.info(
+      "Waiting for up-to-date index for \(uris.map { $0.fileURL?.lastPathComponent ?? $0.stringValue }.joined(separator: ", "))"
+    )
+    let filesWithOutOfDateIndex = uris.filter { uri in
+      switch indexStatus[uri] {
+      case .inProgress, nil: return true
+      case .upToDate: return false
+      }
+    }
```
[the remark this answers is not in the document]
ahoppen: Good catch 👍🏽

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: I am actually already changing this to one index file per task in https://github.com/apple/sourcekit-lsp/pull/1273. 

https://github.com/apple/sourcekit-lsp/issues/1268 is when the batching really starts to make sense.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,302 @@
...
+      if let nextArgument = iterator.next() {
+        if removeFrontendFlags.contains(nextArgument) {
+          continue
+        }
+        result += [argument, nextArgument]
+        continue
+      }
+      result.append(argument)
+    }
+    result.append(argument)
```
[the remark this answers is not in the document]
ahoppen: Oh, yes. Looks like we should just remove the inner `result.append(argument)`

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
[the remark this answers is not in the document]
ahoppen: > I don't think it really matters, but I think you could use `-index-unit-output-path` here.

Good idea. That does seem cleaner.

> Also just noticed this doesn't seem to respect IndexOptions's indexStorePath, is that intentional?

Yes, that’s intentional. `IndexOption.indexStorePath` is an artifact from index-while-building. Background indexing stores all its artifacts independently of the build folder, which ensures that it doesn’t conflict with real builds for now. It should thus also use a different index store path. Unifying background indexing and index-while building is part of https://github.com/apple/sourcekit-lsp/issues/1270

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Still an open question: https://github.com/apple/sourcekit-lsp/issues/1251

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: 👍🏽 Good idea

Sources/SKSupport/SwitchableProcessResultExitStatus.swift
```diff
@@ -0,0 +1,45 @@
...
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// We need to import all of TSCBasic because otherwise we can't refer to Process.ExitStatus (rdar://127577691)
+import struct TSCBasic.ProcessResult
```
bnbarham, line 14: Comment says we're importing all, but actually only importing ProcessResult 🤔?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,311 @@
...
+      for: uri,
+      language: language,
+      logBuildSettings: false
+    )
+    guard let buildSettings else {
+      logger.error("Not indexing \(uri.forLogging) because it has no compiler arguments")
+      return
+    }
+    guard !buildSettings.isFallback else {
+      // Only index with real build settings. Indexing with fallback arguments could result in worse results than not
```
bnbarham, line 143: When could this actually happen? We're not using index-while-building here, so it seems like an odd case. Presumably having *something* is better than nothing, but even if you added a "do we have anything in the index" check, that would still be odd as then we'd index it the first time and not again after that.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -0,0 +1,311 @@
...
+        BuildSettingsLogger.log(level: .error, settings: buildSettings, for: uri)
+      }
+    }
+  }
+}
+
+/// Adjust compiler arguments that were created for building to compiler arguments that should be used for indexing.
+///
+/// This removes compiler arguments that produce output files and adds arguments to index the file.
+private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
```
bnbarham, line 240: This all makes me sad, we really need a "please don't output anything mode". Though... how that would then work with "oh but actually do output the index" is certainly a reasonable question 😅

Sources/SKSupport/SwitchableProcessResultExitStatus.swift
ahoppen: Oh, yeah. I was just stupid when I wrote the comment.

PR #1220, opened 2024-05-03
Rename diagnose bundle created on disk from `sourcekitd-reproducer` to `sourcekit-lsp-diagnose`

ahoppen:
The name still dated back to when `sourcekit-lsp diagnose` would only reduce sourcekitd crashes. Now it does quite a bit more.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1224, opened 2024-05-06
Add GitHub issue templates

ahoppen:
Most importantly, the bug report template is now asking for the user’s OS, Swift version and logging.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -1,0 +1,53 @@
+name: Bug Report
+description: Something isn't working as expected
+labels: [bug]
+body:
+  - type: input
+    id: version
+    attributes:
+      label: Swift version
+      description: Which version of Swift are you using? If you are unsure, insert the output of `path/to/swift --version`
+      placeholder: Eg. swiftlang-5.10.0.13, swift-DEVELOPMENT-SNAPSHOT-2024-05-01-a
+  - type: input
+    id: platform
+    attributes:
+      label: Platform
+      description: What operating system are you seeing the issue on?
+      placeholder: Eg. Ubuntu 22.04, Windows 11, macOS 14
+  - type: input
+    id: editor
+    attributes:
+      label: Editor
+      description: Which text editor are you using?
+      placeholder: Eg. Visual Studio Code with Swift plugin 1.9.0, neovim
+  - type: dropdown
+    id: reproduces-with-swift-6
+    attributes:
+      label: Does the issue reproduce with Swift 6?
+      description: |
+        Does the issue also reproduce using a [recent Swift 6 Development Snapshot](https://www.swift.org/download/#swift-60-development)?
+
+        We have made significant changes to SourceKit-LSP in Swift 6 and the issue might have already been fixed. If you didn’t try, that is fine.
+      options:
+        - "Yes"
+        - "No"
+        - I didn’t try
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: |
+        A short description of the incorrect behavior.
+        If you think this issue has been recently introduced and did not occur in an earlier version, please note that. If possible, include the last version that the behavior was correct in addition to your current version.
+  - type: textarea
+    id: steps-to-reproduce
+    attributes:
+      label: Steps to Reproduce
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,53 @@
...
+    attributes:
+      label: Platform
+      description: What operating system are you seeing the issue on?
+      placeholder: Eg. Ubuntu 22.04, Windows 11, macOS 14
+  - type: input
+    id: editor
+    attributes:
+      label: Editor
+      description: Which text editor are you using?
+      placeholder: Eg. Visual Studio Code with Swift plugin 1.9.0, neovim
```
bnbarham, line 22:
```suggestion
      description: Which text editor are you using (and LSP extension/plugin if applicable)?
      placeholder: Eg. Visual Studio Code with Swift extension 1.9.0, Neovim, etc.
```

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,53 @@
...
+    attributes:
+      label: Steps to Reproduce
+      description: If you have steps that reproduce the issue, please add them here. If you can share a project that reproduces the issue, please attach it.
+  - type: textarea
+    id: logging
+    attributes:
+      label: Logging
+      description: |
+        If you are using SourceKit-LSP from Swift 6, running `sourcekit-lsp diagnose` in terminal and attaching the generated bundle helps us diagnose the issue.
+        The generated bundle might contain portions of your source code, so please only attach it if you feel comfortable sharing it.
```
bnbarham, line 53:
```suggestion
        The generated bundle may contain paths to files on disk as well as portions of your source code. This greatly helps in reproducing issues, but you should only attach it if you feel comfortable doing so.
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,12 @@
+# This source file is part of the Swift.org open source project
+ #
+ # Copyright (c) 2024 Apple Inc. and the Swift project authors
+ # Licensed under Apache License v2.0 with Runtime Library Exception
+ #
+ # See https://swift.org/LICENSE.txt for license information
+ # See https://swift.org/CONTRIBUTORS.txt for Swift project authors
```
bnbarham, line 7: Why copyright on this and not the others?

bnbarham (approved):

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
bnbarham: Do we/should we relativize paths based on the workspace root if we're able to?

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
ahoppen: I think `etc.` isn’t necessary because we already have `E.g.` at the start of the placeholder.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
ahoppen: I don’t think relativizing paths makes much sense. The last path component already contains the file’s name, which might be as sensitive as other parts of the path. And relativizing them might make it harder to diagnose issues. So, I would say nearly no benefit with potential loss of diagnosability, so not worth it.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
bnbarham: Was more the extension/plugin. I don't mind if you add the etc or not :)

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
bnbarham: It could be that the project itself is opensource (so there's no concerns about source), but eg. absolute paths could have their home directory name which could be sensitive.

ahoppen: I doubt that’s a common. You are already sharing your name by filing a GitHub issue and in most cases I expect your GitHub name to be somewhat similar to your username. So, I don’t think it’s something to optimize for until the issue comes up.

PR #1206, opened 2024-04-26
Add "Add documentation" code action to stub out documentation for a function

DougGregor:
This code action takes an undocumented function declaration like

    func refactor(syntax: DeclSyntax, in context: Void) -> DeclSyntax?

and adds stub documentation for the parameters / result / etc., like this:

    /// A description
    /// - Parameters:
    ///   - syntax:
    ///   - context:
    ///
    /// - Returns:

PR #1225, opened 2024-05-06
Use `TokenSyntax.indentationOfLine` from SwiftBasicFormat instead of duplicating indentation inferring logic

ahoppen:
Follow-up to https://github.com/apple/sourcekit-lsp/pull/1206. `SwiftBasicFormat` already has `TokenSyntax.indentationOfLine` that also takes into account trailing trivia of the previous token, so we don’t need to re-implement it here.

Also use `as(DeclSyntaxEnum.self)` instead of force-unwrapping.


DougGregor (approved):
Ah, very good!

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1229, opened 2024-05-07
Prevent rename of argument labels for enum cases

ahoppen:
Renaming an enum case currently caused invalid code to be generated because we would rename eg. `myCase(String)` to `myNewCase(_ String)`.

Fixing the underlying issue requires changes to `sourcekitd`, that are out-of-scope at the moment. For now, just suppress argument label rename for enum cases in SourceKit-LSP and avoid generating invalid code even if just the base name is modified.

rdar://127248157

### DIFF
--- a/Sources/SourceKitLSP/Rename.swift
+++ b/Sources/SourceKitLSP/Rename.swift
@@ -995,6 +995,24 @@ extension SwiftLanguageService {
     return nil
   }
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1230, opened 2024-05-07
Don’t report no-op rename edits

ahoppen:
When renaming `func test(foo: Int) {}` to `test2(foo:)`, rename used to report an edit from `foo` to `foo`, which clutters the refactor preview view. We shouldn’t report edits if no text is actually changed.

rdar://127291815


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1232, opened 2024-05-07
Return `fileHandlingCapability.handled` for all files that `SwiftPMBuildSystem` has build settings for

### LINKED ISSUE #1210: Symbol completion for Package.swift from SwiftPM project in a subfolder doesn't work
If you have a SwiftPM project in a subfolder of a VSCode workspace, SourceKit-LSP does not provide symbol completion, hover documentation for the Package.swift file even though a `workspace/didChangeWorkspaceFolders` request that adds the subfolder has been sent.

See https://github.com/swift-server/vscode-swift/issues/768

ahoppen:
Previously, `SwiftPMBuildSystem` would report `.unhandled` as the file handling capability for `Package.swift`. In single workspace-folder setups, this was not an issue because we would always default to using the build system of the first/only workspace folder, which was the only `SwiftPMBuildSystem` and then returned compiler arguments despite saying that it couldn’t handle `Package.swift`.

This, however, breaks if you have a setup with multi workspace folders, where we can’t rely on the first workspace folder being able to provide build settings for the package manifest.

To fix this, report `.handled` file handling capability for all files that `SwiftPMBuildSystem` can compute build settings for.

Fixes #1210
rdar://127351614

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
@@ -315,6 +315,11 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
   public var indexPrefixMappings: [PathPrefixMapping] { return [] }
 
   public func buildSettings(for uri: DocumentURI, language: Language) throws -> FileBuildSettings? {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1237, opened 2024-05-07
Address my own review comments to #1179

ahoppen:
Addresses a few minor comments and the following major ones:
- Add test cases for the syntax refactorings
- Don’t report code actions for refactorings that don’t actually modify the source
- Instead of just looking at the parent of the token of the selected range, walk up the syntax tree to find the syntax node to refactor. This makes the refactorings available in a lot more locations.

### DIFF
--- a/Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
@@ -18,29 +18,35 @@ import SwiftSyntax
 /// swift-syntax) into a SyntaxCodeActionProvider.
 protocol SyntaxRefactoringCodeActionProvider: SyntaxCodeActionProvider, EditRefactoringProvider {
   static var title: String { get }
+
+  /// Returns the node that the syntax refactoring should be performed on, if code actions are requested for the given
+  /// scope.
+  static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> Input?
 }
 
 /// SyntaxCodeActionProviders with a \c Void context can automatically be
 /// adapted provide a code action based on their refactoring operation.
 extension SyntaxRefactoringCodeActionProvider where Self.Context == Void {
   static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
-    guard
-      let token = scope.firstToken,
-      let node = token.parent?.as(Input.self)
-    else {
+    guard let node = nodeToRefactor(in: scope) else {
       return []
     }
 
     let sourceEdits = Self.textRefactor(syntax: node)
-    if sourceEdits.isEmpty {
-      return []
-    }
 
-    let textEdits = sourceEdits.map { edit in
-      TextEdit(
+    let textEdits = sourceEdits.compactMap { (edit) -> TextEdit? in
+      let edit = TextEdit(
         range: scope.snapshot.range(of: edit.range),
         newText: edit.replacement
       )
+      if edit.isNoOp(in: scope.snapshot) {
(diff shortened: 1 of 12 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -42,6 +42,26 @@ final class SyntaxRefactorTests: XCTestCase {
     )
   }
 
+  func testAddDocumentationRefactorSingleParameter() throws {
+    try assertRefactor(
+      """
+        1️⃣func 2️⃣refactor(syntax: DeclSyntax) { }
```
bnbarham, line 48: Is on the label valid? Parameter?

bnbarham (approved):

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
ahoppen: It doesn't work in the `assertRefactor` tests because those invoke the refactoring directly from the syntax node. I did add a new code action test case that tests invoking documentation adding from the parameters (because that does the new logic of walking up parents to find the node on which to invoke the refactoring action).

ahoppen:
Fixed a few more bugs:
- Documentation was added at the start of a declaration’s trivia. This meant that if there were two newlines before the declaration, the documentation would be separated to the declaration by one newline and if the declaration was at the start of a line, the declaration would be on the same line as the doc comment, effectively making the documentation part of a comment.
- Don’t show `Add documentation` if the declaration is not on a new line, like `var x = 1; var y = 2`
- Add tests to check that the code action request also works when selecting ranges

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -18,29 +18,35 @@ import SwiftSyntax
...
-    }
 
-    let textEdits = sourceEdits.map { edit in
-      TextEdit(
+    let textEdits = sourceEdits.compactMap { (edit) -> TextEdit? in
+      let edit = TextEdit(
         range: scope.snapshot.range(of: edit.range),
         newText: edit.replacement
       )
+      if edit.isNoOp(in: scope.snapshot) {
```
DougGregor, line 42: Ah, good idea. I was thinking of fixing up the various refactors individually to not produce no-ops, but this will work more consistently.

Sources/SourceKitLSP/Swift/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -57,22 +63,77 @@ extension SyntaxRefactoringCodeActionProvider where Self.Context == Void {
...
+    return scope.innermostNodeContainingRange?.findParentOfSelf(
+      ofType: IntegerLiteralExprSyntax.self,
+      stoppingIf: { $0.is(CodeBlockSyntax.self) || $0.is(MemberBlockSyntax.self) }
+    )
+  }
+}
+
+extension Syntax {
+  /// Finds the innermost parent of the given type while not walking outside of nodes that satisfy `stoppingIf`.
+  func findParentOfSelf<ParentType: SyntaxProtocol>(
```
DougGregor, line 126: Nice abstraction

DougGregor (approved):
Thank you!

ahoppen:
@swift-ci Please test Windows

PR #1249, opened 2024-05-08
Introduce a notion of `ConfiguredTargets` into the build system

ahoppen:
Instead of asking for build settings of a file, the build system manager asks for the targets of a file and then asks for the build settings of that file in a specific target. This has two advantages:
- We know about targets and can prepare the targets for background indexing
- Once we support build systems in which a single file can be part of multiple targets, we can have a centralized place that picks preferred targets for a file, eg. based on user configuration 

Depends on https://github.com/apple/swift-package-manager/pull/7540


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift-package-manager/pull/7540

@swift-ci Please test

Sources/SKCore/BuildSystem.swift
```diff
@@ -43,6 +43,27 @@ public struct SourceFileInfo: Sendable {
...
+  ///
+  /// The run destination's ID should be generated by the build system that handles the target and only interpreted by
+  /// that build system.
+  public let runDestinationID: String
+
+  public init(targetID: String, runDestinationID: String) {
+    self.targetID = targetID
+    self.runDestinationID = runDestinationID
+  }
+}
```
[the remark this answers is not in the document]
ahoppen: Looks like we had similar thought. I tried that as well, had too many issues in `BuildSystemManager` and then decided to just go with strings.

Sources/SKCore/BuildSystem.swift
```diff
@@ -43,6 +43,27 @@ public struct SourceFileInfo: Sendable {
...
+  ///
+  /// The target's ID should be generated by the build system that handles the target and only interpreted by that
+  /// build system.
+  public let targetID: String
+
+  /// An opaque string that represents the run destination.
+  ///
+  /// The run destination's ID should be generated by the build system that handles the target and only interpreted by
+  /// that build system.
+  public let runDestinationID: String
```
bnbarham, line 59: Do we need a "configuration" as well, eg. debug/release/etc? Run destination is fairly general, so you could definitely include platform/sdk/architecture/etc in that, but I'm not sure configuration makes much sense there.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -314,43 +323,72 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+    guard let buildTarget = buildTargets.first else {
+      if buildTargets.isEmpty {
+        logger.error("Did not find target with name \(configuredTarget.targetID)")
+      }
+      return nil
     }
 
-    if path.extension == "h" {
-      return try settings(forHeader: path)
+    if url.pathExtension == "h", let substituteFile = buildTarget.sources.first {
```
bnbarham, line 351: `hpp` is pretty common as well. Should this instead call `defaultLanguage` and then check if it ends with `-header`? (We'd have to implement something small for `defaultLanguage` here)

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -314,43 +323,72 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+      return FileBuildSettings(
+        compilerArguments: try buildTarget.compileArguments(for: substituteFile),
+        workingDirectory: workspacePath.pathString
+      ).patching(newFile: try resolveSymlinks(path).pathString, originalFile: substituteFile.absoluteString)
```
bnbarham, line 355: Do we actually want to only compile the header? Has pros and cons - I imagine there's a bunch of codebases where the header doesn't actually compile in isolation, but it would generally be much faster.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -314,43 +323,72 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+      return [ConfiguredTarget(targetID: target.name, runDestinationID: "dummy")]
+    }
+
+    if path.basename == "Package.swift" {
+      // We use an empty target name to represent the package manifest since an empty target name is not valid for any
+      // user-defined target.
+      return [ConfiguredTarget(targetID: "", runDestinationID: "dummy")]
+    }
+
+    if url.pathExtension == "h", let target = try? target(forHeader: path) {
```
bnbarham, line 385: Could use `defaultLanguage` here too.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -314,43 +323,72 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
   }
 
+  public func configuredTargets(for uri: DocumentURI) -> [ConfiguredTarget] {
+    guard let url = uri.fileURL, let path = try? AbsolutePath(validating: url.path) else {
+      // We can't determine targets for non-file URIs.
+      return []
+    }
+
+    if let target = try? buildTarget(for: path) {
+      return [ConfiguredTarget(targetID: target.name, runDestinationID: "dummy")]
```
bnbarham, line 376: I realize I'm well down the stack of PRs, so it's possible this has already changed. But there is now multiple destinations (host/target) in SwiftPM today.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2471,7 +2471,11 @@ fileprivate extension CheckedIndex {
   /// If the USR has an ambiguous definition, the most important role of this function is to deterministically return
   /// the same result every time.
   func primaryDefinitionOrDeclarationOccurrence(ofUSR usr: String) -> SymbolOccurrence? {
-    return definitionOrDeclarationOccurrences(ofUSR: usr).sorted().first
+    let result = definitionOrDeclarationOccurrences(ofUSR: usr).sorted().first
+    if result == nil {
+      logger.error("Failed to find definition of \(usr) in index")
```
bnbarham, line 2476: I assume you added this after something happened and this would have helped, out of interest... what was that?

Sources/SKCore/BuildSystem.swift
ahoppen: Discussed offline that we don’t need it for now because the configuration is picked globally when launching sourcekit-lsp. If we do allow changing the configuration while sourcekit-lsp is running, we’ll need to re-consider whether to add configuration here.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: I changed to just check if `url` is part of `sources` and if not, assume that it’s a header file. That removes the need for file extension checking, which should be more general.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: This will be handled by https://github.com/apple/sourcekit-lsp/pull/1288

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I think it was just trying to figure out why the test failed and I wanted to know whether LSP got to a USR and couldn’t find the USR, or if it never found the USR, or I wanted to check which USR it was actually looking for and if it was looking for the wrong USR. I don’t remember the details.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: We don’t actually compile the headers. The primary use case of this function is to provide the compiler arguments for clangd for semantic functionality. During indexing, we always index a main file. Not sure if that code existed when you reviewed the PR though.

https://github.com/apple/sourcekit-lsp/blob/490871b21f90bc1270fb4c57978b2ab0b14073e9/Sources/SemanticIndex/SemanticIndexManager.swift#L236-L241

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: Same here, just assume that a file is a header if it’s not part of the sources. Even if the file is not a header, I think this gives reasonable results. 

Say, there was a new `.cpp` file in a target and for some reason the `SwiftPMBuildSystem` doesn’t know about it. Then we would infer the target based on the file’s location on disk and generate compiler arguments for it by picking a source file in that target, getting its compiler arguments and then patching up the compiler arguments by replacing the substitute file with the `.cpp` file.

bnbarham: Did you also rename `target(forHeader:)`? Maybe `inferTarget(for:)`?

PR #1276, opened 2024-05-13
Make all non-test modules except for SourceKitLSP build in Swift 6 mode

ahoppen:
Depends on https://github.com/apple/swift-package-manager/pull/7553

---

This would have caught a race condition in background indexing that was caused by accessing `CheckedIndex` from multiple threads despite it not being thread-safe.

### DIFF
--- a/Tests/SKCoreTests/ToolchainRegistryTests.swift
+++ b/Tests/SKCoreTests/ToolchainRegistryTests.swift
@@ -36,36 +36,6 @@ final class ToolchainRegistryTests: XCTestCase {
     await assertTrue(tr.default === tr.toolchain(identifier: "a"))
   }
 
-  func testDefaultDarwin() async throws {
-    let prevPlatform = Platform.current
-    defer { Platform.current = prevPlatform }
-    Platform.current = .darwin
-
-    let tr = ToolchainRegistry(
-      toolchains: [
-        Toolchain(identifier: "a", displayName: "a", path: nil),
-        Toolchain(identifier: ToolchainRegistry.darwinDefaultToolchainIdentifier, displayName: "a", path: nil),
-      ]
-    )
-    await assertEqual(tr.default?.identifier, ToolchainRegistry.darwinDefaultToolchainIdentifier)
-  }
-
-  func testUnknownPlatform() throws {
-    let prevPlatform = Platform.current
-    defer { Platform.current = prevPlatform }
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Tests/SKCoreTests/ToolchainRegistryTests.swift
```diff
@@ -36,36 +36,6 @@ final class ToolchainRegistryTests: XCTestCase {
     await assertTrue(tr.default === tr.toolchain(identifier: "a"))
   }
 
-  func testDefaultDarwin() async throws {
-    let prevPlatform = Platform.current
```
ahoppen, line 40: These tests are no longer supported because `Platform.current` isn’t mutable anymore. We should still have coverage fro this because we’re running tests on macOS and Linux.

Sources/SKCore/Platform+CurrentConcurrencySafe.swift
```diff
@@ -0,0 +1,71 @@
...
+import protocol TSCBasic.FileSystem
+import class TSCBasic.Process
+import var TSCBasic.localFileSystem
+
+// Copy pasted from PackageLoading/Platform.swift to make it concurrency-safe
+
+extension Platform {
+  // This is not just a computed property because the ToolchainRegistryTests
+  // change the value.
+  public nonisolated(unsafe) static let currentConcurrencySafe: Platform? = {
```
hamishknight, line 24: Can't we fix this in SwiftPM? From the sounds of the comment, the SwiftPM one only needed to be a var for the test that's being removed here

ahoppen: Oh, indeed. I just assumed that tests in SwiftPM would also modify it. But we are indeed the only ones that require it to be a variable.

ahoppen:
https://github.com/apple/swift-package-manager/pull/7553

@swift-ci Please test

hamishknight (approved):

ahoppen:
https://github.com/apple/swift-package-manager/pull/7553

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift-package-manager/pull/7553

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-package-manager/pull/7553

@swift-ci Please test Windows

PR #1289, opened 2024-05-14
Index clang files

### LINKED ISSUE #1253: Indexing of C-family languages
We need to support indexing of C-family languages.

rdar://127474135

ahoppen:
Fixes #1253
rdar://127474135

### DIFF
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -163,9 +163,12 @@ public struct UpdateIndexStoreTaskDescription: TaskDescriptionProtocol {
         BuildSettingsLogger.log(settings: buildSettings, for: uri)
       }
     case .c, .cpp, .objective_c, .objective_cpp:
-      // TODO (indexing): Support indexing of clang files, including headers.
-      // https://github.com/apple/sourcekit-lsp/issues/1253
-      break
+      do {
+        try await updateIndexStore(forClangFile: uri, buildSettings: buildSettings, toolchain: toolchain)
+      } catch {
+        logger.error("Updating index store for \(uri) failed: \(error.forLogging)")
+        BuildSettingsLogger.log(settings: buildSettings, for: uri)
+      }
     default:
       logger.error(
         "Not updating index store for \(uri) because it is a language that is not supported by background indexing"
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -304,3 +346,56 @@ private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
...
+    // Disable writing of a depfile
+    "-M",
+    "-MD",
+    "-MMD",
+    "-MG",
+    "-MM",
+    "-MV",
+    // Don't create phony targets
+    "-MP",
+    // Don't writ out compilation databases
```
bnbarham, line 367: `write` (and thanks for adding the comments 🙏)

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -304,3 +346,56 @@ private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
...
+    "-MJ",
+    // Continue in the presence of errors during indexing
```
bnbarham, line 369: Is this the reason? Or is it just that we don't have a decent timestamp to use for the "build session" here? Or does having this enable verification where we wouldn't otherwise?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -304,3 +346,56 @@ private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
...
+    "-MQ",
+    // Don't write serialized diagnostic files
+    "--serialize-diagnostics",
+  ]
+
+  var result: [String] = []
+  result.reserveCapacity(compilerArguments.count)
+  var iterator = compilerArguments.makeIterator()
+  while let argument = iterator.next() {
+    if removeFlags.contains(argument) || argument.starts(with: "-fbuild-session-file=") {
```
bnbarham, line 388: Does `serailize-diagnostics` also allow `=`? Does `-fbuild-session-file` allow a separate arg rather than `=`?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Honestly, I don’t know. I was trying to infer it but maybe I was wrong. I just removed the explanation for now.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: So, `serailize-diagnostics` does not allow `=` and `-fbuild-session-file` requires a `=`. But eg. `-MT` doesn’t require a space. And `serialize-diagnostics` can be spelled with a single or multiple spaces. 

I ended up generalizing the entire command line argument matching in https://github.com/apple/sourcekit-lsp/pull/1314

PR #1291, opened 2024-05-14
Don’t rely on working directory during index preparation

ahoppen:
Amazon Linux and CentOS don’t support working directory. We can work around this by passing `--package-path` to `swift build`.

rdar://128037023


ahoppen:
@swift-ci Please test

PR #1300, opened 2024-05-14
Change compiler’s diagnostic provider name to `SourceKit`

ahoppen:
The fact that they are coming from a service named `sourcekitd` should be an implentation detail of SourceKit-LSP and shouldn’t be exposed to users. Use the generic `SourceKit` term, which is vague about which SourceKit the diagnostics are coming from.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1302, opened 2024-05-15
Update index as files are modified on disk

### LINKED ISSUE #1251: Decide which files we should re-index when the package is changed
When the build system is modified, we need to do two things:
- Index all new files
- For all other files, we have three options:
  - Trigger a re-index for them (most correct)
  - Diff build settings for the files and only re-index if build settings have changed (slightly correct but might miss indexing changes if a new file changes overload resolution without build settings changing)
  - Assume that the updated package doesn’t affect the index and don’t re-index

rdar://127476161

ahoppen:
Fixes #1251
rdar://127476161

### DIFF
--- a/Documentation/Files_To_Reindex.md
+++ b/Documentation/Files_To_Reindex.md
@@ -1,0 +1,33 @@
+# Which files to re-index when a file is modified
+
+## `.swift`
+
+Obviously affects the file itself, which we do re-index.
+
+If an internal declaration was changed, all files within the module might be also affected. If a public declaration was changed, all modules that depend on it might be affected. The effect can only really be in three ways:
+1. It might change overload resolution in another file, which is fairly unlikely
+2. A new declaration is introduced in this file that is already referenced by another file
+3. A declaration is removed in this file that was referenced from other files. In those cases the other files now have an invalid reference.
+
+We decided to not re-index any files other than the file itself because naively re-indexing all files that might depend on the modified file requires too much processing power that will likely have no or very little effect on the index – we are trading accuracy for CPU time here.
+We mark the targets of the changed file as well as any dependent targets as out-of-date. The assumption is that most likely the user will go back to any of the affected files shortly afterwards and modify them again. When that happens, the affected file will get re-indexed and bring the index back up to date.
+
+Alternatives would be:
+- We could we check the file’s interface hash and re-index other files based on whether it has changed. But at that point we are somewhat implementing a build system. And even if a new public method was introduced it’s very likely that the user hasn’t actually used it anywhere yet, which means that re-indexing all dependent modules would still be doing unnecessary work.
+- To cover case (2) from above, we could re-index only dependencies that previously indexed with errors. This is an alternative that hasn’t been fully explored.
+
+## `.h`
+
+Affects all files that include the header (including via other headers). For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and so we don’t re-index any file. If it is, it’s likely that the user will modify the file that includes the new header file shortly after, which will re-index it.
+
+## `.c` / `.cpp` / `.m`
+
+This is the easy case since only the file itself is affected.
+
+## Compiler settings (`compile_commands.json` / `Package.swift`)
+
+Ideally, we would considered like a file change for all files whose compile commands have changed, if they changed in a meaningful way (ie. in a way that would also trigger re-compilation in an incremental build). Non-meaningful changes would be:
+- If compiler arguments, like include paths are shuffled around. We could have a really quick check for compiler arguments equality by comparing them unordered. Any really compiler argument change will most likely do more than rearranging the arguments.
+- The addition of a new Swift file to a target is equivalent to that file being modified and shouldn’t trigger a re-index of the entire target.
+
+At the moment, unit files don’t include information about the compiler arguments with which they were created, so it’s impossible to know whether the compiler arguments have changed when a project is opened. Thus, for now, we don’t re-index any files on compiler settings changing.
--- a/Sources/SKSupport/Sequence+AsyncMap.swift
+++ b/Sources/SKSupport/Sequence+AsyncMap.swift
@@ -12,7 +12,7 @@
 
 extension Sequence {
   /// Just like `Sequence.map` but allows an `async` transform function.
-  func asyncMap<T>(
+  public func asyncMap<T>(
     @_inheritActorContext _ transform: @Sendable (Element) async throws -> T
   ) async rethrows -> [T] {
     var result: [T] = []
@@ -26,7 +26,7 @@ extension Sequence {
   }
 
   /// Just like `Sequence.compactMap` but allows an `async` transform function.
-  func asyncCompactMap<T>(
+  public func asyncCompactMap<T>(
     @_inheritActorContext _ transform: @Sendable (Element) async throws -> T?
   ) async rethrows -> [T] {
     var result: [T] = []
--- a/Sources/SemanticIndex/CheckedIndex.swift
+++ b/Sources/SemanticIndex/CheckedIndex.swift
@@ -104,10 +104,6 @@ public final class CheckedIndex {
     }
   }
 
-  public func symbolProvider(for sourceFilePath: String) -> SymbolProviderKind? {
-    return index.symbolProvider(for: sourceFilePath)
-  }
-
   public func symbols(inFilePath path: String) -> [Symbol] {
     guard self.hasUpToDateUnit(for: URL(fileURLWithPath: path, isDirectory: false)) else {
       return []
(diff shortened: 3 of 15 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Documentation/Files_To_Reindex.md
```diff
@@ -0,0 +1,32 @@
...
+1. It might change overload resolution in another file, which is fairly unlikely
+2. A new declaration is introduced in this file that is already referenced by another file
+
+We decided to not re-index any files other than the file itself because naively re-indexing all files that might depend on the modified file requires too much processing power that will likely have no or very little affect on the index – we are trading accuracy for CPU time here.
+We mark the targets of the changed file as well as any dependent targets as out-of-date. The assumption is that most likely the user will go back to any of the affected files shortly afterwards and modify them again. When that happens, the affected file will get re-indexed and bring the index back up to date.
+
+Alternatives would be:
+- We could we check the file’s interface hash and re-index other files based on whether it has changed. But at that point we are somewhat implementing a build system. And even if a new public method was introduced it’s very likely that the user hasn’t actually used it anywhere yet, which means that re-indexing all dependent modules would still be doing unnecessary work.
```
hamishknight, line 15: I wonder, with fine enough dependencies and swift-driver support, whether some kind of incremental-mode-based indexing could make sense. But that's way off in the future 😄

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -170,6 +184,42 @@ public final actor SemanticIndexManager {
...
+    await scheduleBackgroundIndex(files: filesToReIndex)
+  }
+
+  /// Returns the files that should be re-indexed if the given files have been modified.
+  ///
+  /// - SeeAlso: The `Documentation/Files_To_Reindex.md` file.
+  private func filesToReIndex(changedFiles: [DocumentURI]) async -> Set<DocumentURI> {
+    let sourceFiles = await buildSystemManager.sourceFiles().map(\.uri)
+    let filesToReIndex = await changedFiles.asyncMap { (uri) -> [DocumentURI] in
+      if sourceFiles.contains(uri) {
```
hamishknight, line 203: I would use a `Set` for `sourceFiles` to avoid potentially quadratic runtime

Sources/SemanticIndex/CheckedIndex.swift
```diff
@@ -120,6 +116,13 @@ public final class CheckedIndex {
     return index.unitTests(referencedByMainFiles: mainFilePaths).filter { checker.isUpToDate($0.location) }
   }
 
+  /// Returns all the files that (transitively) include the file at the given path.
+  ///
+  /// If `crossLanguage` is set to `true`, Swift files that import a header will through a module will also be reported.
+  public func mainFilesContainingFile(path: String, crossLanguage: Bool = false) -> [String] {
+    index.mainFilesContainingFile(path: path, crossLanguage: crossLanguage)
+  }
```
hamishknight, line 124: Is it expected this doesn't do any up-to-date checking?

hamishknight (approved):

Sources/SemanticIndex/CheckedIndex.swift
ahoppen: Ah, good call. Guess we should check.

Documentation/Files_To_Reindex.md
```diff
@@ -0,0 +1,32 @@
+# Which files to re-index when a file is modified
+
+## `.swift`
+
+Obviously affects the file itself, which we do re-index.
+
+If an internal declaration was changed, all files within the module might be also affected. If a public declaration was changed, all modules that depend on it might be affected. The effect can only really be in two ways:
+1. It might change overload resolution in another file, which is fairly unlikely
+2. A new declaration is introduced in this file that is already referenced by another file
```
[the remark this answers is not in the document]
ahoppen: Yes, I think so. But added the following for completeness sake:

> 3. A declaration is removed in this file that was referenced from other files. In those cases the other files now have an invalid reference.

Documentation/Files_To_Reindex.md
ahoppen: Even if we had it, I’m not sure if we really want to do a whole downstream re-index. Say you add a new public function in a fairly low-level module then we would need to re-index the entire project based on incremental information because it might affect overload resolution. And that’s still quite a bit of work, considering that most likely nothing really changed.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Documentation/Files_To_Reindex.md
hamishknight: It depends how fine grained the dependencies are, in theory if you're adding a function called `foo`, you shouldn't need to rebuild anything that didn't do a lookup for `foo`

hamishknight (approved):

Documentation/Files_To_Reindex.md
ahoppen: That’s a dream for a far and bright future ✨

Documentation/Files_To_Reindex.md
```diff
@@ -0,0 +1,33 @@
...
+
+Alternatives would be:
+- We could we check the file’s interface hash and re-index other files based on whether it has changed. But at that point we are somewhat implementing a build system. And even if a new public method was introduced it’s very likely that the user hasn’t actually used it anywhere yet, which means that re-indexing all dependent modules would still be doing unnecessary work.
+- To cover case (2) from above, we could re-index only dependencies that previously indexed with errors. This is an alternative that hasn’t been fully explored.
+
+## `.h`
+
+Affects all files that include the header (including via other headers). For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and so we don’t re-index any file. If it is, it’s likely that the user will modify the file that includes the new header file shortly after, which will re-index it.
```
bnbarham, line 21: There's another PR that changes this to only index one main file (which IMO is what you want) - should update this doc

Documentation/Files_To_Reindex.md
```diff
@@ -0,0 +1,33 @@
...
+
+## `.c` / `.cpp` / `.m`
+
+This is the easy case since only the file itself is affected.
+
+## Compiler settings (`compile_commands.json` / `Package.swift`)
+
+Ideally, we would considered like a file change for all files whose compile commands have changed, if they changed in a meaningful way (ie. in a way that would also trigger re-compilation in an incremental build). Non-meaningful changes would be:
```
bnbarham, line 29: `would be considered` rather than `we would considered`?

Documentation/Files_To_Reindex.md
```diff
@@ -0,0 +1,33 @@
...
+- If compiler arguments, like include paths are shuffled around. We could have a really quick check for compiler arguments equality by comparing them unordered. Any really compiler argument change will most likely do more than rearranging the arguments.
```
bnbarham, line 30: `really` -> `real`? FWIW shuffling of includes *could* be meaningful since it would change the search order.

Sources/SKSupport/Sequence+AsyncMap.swift
```diff
@@ -12,7 +12,7 @@
 
 extension Sequence {
   /// Just like `Sequence.map` but allows an `async` transform function.
-  func asyncMap<T>(
+  public func asyncMap<T>(
```
bnbarham, line 15: Do these need to be public or can they be package (or SPI)? Don't love adding public extensions onto `Sequence`.

Sources/SemanticIndex/CheckedIndex.swift
```diff
@@ -120,6 +116,19 @@ public final class CheckedIndex {
     return index.unitTests(referencedByMainFiles: mainFilePaths).filter { checker.isUpToDate($0.location) }
   }
 
+  /// Returns all the files that (transitively) include the header file at the given path.
+  ///
+  /// If `crossLanguage` is set to `true`, Swift files that import a header will through a module will also be reported.
```
bnbarham, line 121: Drop the first `will`

Sources/SourceKitLSP/Rename.swift
```diff
@@ -500,7 +500,7 @@ extension SourceKitLSPServer {
   ) async -> (swiftLanguageService: SwiftLanguageService, snapshot: DocumentSnapshot, location: SymbolLocation)? {
     var reference: SymbolOccurrence? = nil
     index.forEachSymbolOccurrence(byUSR: usr, roles: renameRoles) {
-      if index.symbolProvider(for: $0.location.path) == .swift {
+      if index.unchecked.symbolProvider(for: $0.location.path) == .swift {
```
bnbarham, line 503: Doesn't the reference here need to be up to date?

Documentation/Files_To_Reindex.md
ahoppen: Already done in https://github.com/apple/sourcekit-lsp/pull/1309

Sources/SKSupport/Sequence+AsyncMap.swift
ahoppen: Adopting `package` in SourceKit-LSP would be good in general. We currently still use `public` for anything that’s  used cross-module. I filed https://github.com/apple/sourcekit-lsp/issues/1315 to adopt `package` in general.

Sources/SourceKitLSP/Rename.swift
ahoppen: `index.forEachSymbolOccurrence` only returns valid references because `index` is a `CheckedIndex`. 

Once we have the reference, it doesn’t make sense to find the symbol provider for a document with respect to a `CheckedIndex`:
- Checking for `.deletedFile` doesn’t make sense because we wouldn’t even have a reference from within the file. And if you ask for the symbol provider of a modified file: Sure, why not just return the last language that we know about.
- Checking for `.modifiedFile` or `.inMemoryModifiedFiles` also doesn’t really make sense because it’s better to get the last known symbol provider than no symbol provider at all. A source file’s symbol provider doesn’t change just because the file is modified on disk.

PR #1306, opened 2024-05-16
Don’t re-index all files that include a header when a header is modified

ahoppen:
When a header is modified, we don’t we want to re-index all main files that include it. Instead, we just want to index one main to effectively re-index the header itself.

I originally implemented re-indexing of all files that include the header but on second thought, headers are like Swift modules, where we also don’t re-index all dependencies either. And if you change a low-level header that’s included by the entire project, you probably don’t want the indexer to go off and re-index the entire project.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -115,7 +115,7 @@ public final actor SemanticIndexManager {
   /// Returns immediately after scheduling that task.
   ///
   /// Indexing is being performed with a low priority.
-  public func scheduleBackgroundIndex(files: some Collection<DocumentURI>) async {
+  private func scheduleBackgroundIndex(files: some Collection<DocumentURI>) async {
     await self.index(files: files, priority: .low)
   }
 
@@ -127,7 +127,7 @@ public final actor SemanticIndexManager {
     generateBuildGraphTask = Task(priority: .low) {
       await orLog("Generating build graph") { try await self.buildSystemManager.generateBuildGraph() }
       let index = index.checked(for: .modifiedFiles)
-      let filesToIndex = await self.buildSystemManager.sourceFiles().map(\.uri)
+      let filesToIndex = await self.buildSystemManager.sourceFiles().lazy.map(\.uri)
         .filter { uri in
           guard let url = uri.fileURL else {
             // The URI is not a file, so there's nothing we can index.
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -279,20 +290,20 @@ public final actor SemanticIndexManager {
...
 
     // Sort the targets in topological order so that low-level targets get built before high-level targets, allowing us
     // to index the low-level targets ASAP.
-    var filesByTarget: [ConfiguredTarget: [DocumentURI]] = [:]
+    var filesByTarget: [ConfiguredTarget: [FileToIndex]] = [:]
     for file in outOfDateFiles {
-      guard let target = await buildSystemManager.canonicalConfiguredTarget(for: file) else {
-        logger.error("Not indexing \(file.forLogging) because the target could not be determined")
+      guard let target = await buildSystemManager.canonicalConfiguredTarget(for: file.uri) else {
+        logger.error("Not indexing \(file.uri.forLogging) because the target could not be determined")
```
hamishknight, line 306: It probably(?) doesn't matter, but IMO passing the main file here makes more sense since that's the thing we're indexing

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -185,31 +185,41 @@ public final actor SemanticIndexManager {
...
+        .sorted(by: { $0.stringValue < $1.stringValue }).first
+      guard let mainFile else {
+        return nil
       }
-      // Otherwise, see if it is a header file. If so, re-index all the files that import it.
-      // We don't want to re-index `.swift` files that depend on a header file, similar to how we don't re-index a Swift
-      // module if one of its dependent modules has changed.
-      return index.checked(for: .deletedFiles).mainFilesContainingFile(uri: uri, crossLanguage: false)
-    }.flatMap { $0 }
-    return Set(filesToReIndex)
```
hamishknight, line 212: Given we're no longer uniquing, is there any concern that if e.g 5 headers get modified, and those headers are all included by a single main file, that we could end up running 5 concurrent indexing jobs for that main file?

hamishknight (approved):

ahoppen:
Will address review comments in a follow-up PR.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -185,31 +185,41 @@ public final actor SemanticIndexManager {
   }
 
   public func filesDidChange(_ events: [FileEvent]) async {
-    let filesToReIndex = await filesToReIndex(changedFiles: events.map(\.uri))
-
+    // We only re-index the files that were changed and don't re-index any of their dependencies. See the
+    // `Documentation/Files_To_Reindex.md` file.
```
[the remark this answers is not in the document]
ahoppen: Yeah, thought about that on my way home last night as well and wanted to put up a PR to change that anyway. But well spotted 👍🏽

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: For `canonicalConfiguredTarget(for:)`: Yes, you’re right, we should use `file.mainFile ?? file.uri`

For `logger.error`: Added both `file.uri` and `file.mainFile` to the logging message.

Sources/SemanticIndex/SemanticIndexManager.swift
hamishknight: I also didn't really think about the scheduling logic here :) Looking at it again, we'll block the requests on each other, and then they should all be considered up-to-date

ahoppen: We don’t actually, because they all have different `FileToIndex.uri` and thus they would actually be executing in parallel. Fixing that now.

Issue #1258, opened 2024-05-09
Track which targets are up-to-date


ahoppen:
We currently always prepare a target when a file is scheduled for indexing. We should track which targets are up-to-date and don’t schedule a prepare for up-to-date targets, avoiding the delay of a null build. 

rdar://127475948

Issue #1252, opened 2024-05-09
Prepare target when a file is opened


ahoppen:
When a file is opened, we should prepare its target so that all the dependent types are available in code completion etc.

rdar://127474003

PR #1307, opened 2024-05-16
When interacting with a document, prepare the target it belongs to

ahoppen:
Whenever we get request for a document, open it or edit it, trigger a preparation of its target, but don’t block any interaction based on it. This should ensure that the target is usually prepared when the user is interacting with it.

We need to track the preparation status of targets somewhat accurately in `SemanticIndexManager`, so we don’t unnecessarily re-prepare a target. When updating the index store, it is acceptable to schedule another `UpdateIndexStoreTaskDescription` because it will early exit based on an `mtime` check with the unit file. Null builds of a target take significantly longer and thus we want to avoid them.

Fixes https://github.com/apple/sourcekit-lsp/issues/1252
Fixes https://github.com/apple/sourcekit-lsp/issues/1258
rdar://127474003
rdar://127475948
rdar://127473987


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -233,67 +244,101 @@ public final actor SemanticIndexManager {
...
+        // re-create new `PreparationTaskDescription`s for every preparation request. Instead, all the preparation
+        // requests await the same task. At the same time, if we have a multi-file preparation request and then get a
+        // single-file preparation request, we will override the preparation of that target with the single-file
+        // preparation task, ensuring that the task gets prepared as quickly as possible.
+        if configuredTargets.count <= targets.count {
+          preparationTasksToAwait.append(task)
+        }
+        fallthrough
+      case nil:
+        targetsToPrepare.append(target)
```
hamishknight, line 284: I might be misunderstanding, but won't this mean that we'd still schedule the preparation again for the same target if we already have one running?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -994,6 +994,14 @@ extension SourceKitLSPServer: MessageHandler {
 
     logger.log("Received request \(id): \(params.forLogging)")
 
+    if let textDocumentRequest = params as? any TextDocumentRequest {
+      // When we are requesting information from a document, poke preparation of its target. We don't want to wait for
+      // the preparation to finish because that would cause too big a delay.
+      // In practice, if the user is working on a file, we'll
```
hamishknight, line 1000: Nit: Incomplete comment

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -233,67 +244,101 @@ public final actor SemanticIndexManager {
     return filesToReIndex
   }
 
+  /// Schedule preparation of the target that contains the given URI, building all modules that the file depends on.
+  ///
+  /// This is intended to be called when the user is interacting with the document at the given URI.
+  public func schedulePreparation(of uri: DocumentURI, priority: TaskPriority? = nil) async {
```
hamishknight, line 250: Nit: This doesn't have to be `async`

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Yes, it does and that’s intentional, in the context of multi-target preparation. Say that while background indexing you are preparing targets `A`, `B`, and `C` in a single SwiftPM invocation (like a hypothetical `swift prepare --target A --target B --target C`), which has the benefit that it allows concurrent building of the three targets if they don’t depend on each other. 

Now you are trying to work on a file in `A`, which requires target `A` to be prepared. Now, you don’t want to wait for preparation of all three targets to finish, but want to build target `A` as quickly as possible. So, you schedule preparation of `A` with high priority, which will cancel and reschedule the preparation of `A`, `B` and `C` in `TaskScheduler`. Once preparation of `A` has finished, the joint preparation of `A`, `B`, and `C` is re-scheduled, which will effectively only need to prepare `B` and `C` because the build system realizes that `A` is already up-to-date.

Does that make sense?

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Oops, it was getting late last night. Finished the comment to

```
      // In practice, while the user is working on a file, we'll get a text document request for it on a regular basis,
      // which prepares the files. For files that are open but aren't being worked on (eg. a different tab), we don't
      // get requests, ensuring that we don't unnecessarily prepare them.
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SemanticIndex/SemanticIndexManager.swift
hamishknight: Yeah that makes sense. I was mainly thinking of the case where you have `A` being prepared and get a new request to prepare `A`, we'll schedule it again. Though I wasn't really thinking about the scheduling logic, now that I think about it, we'll make the subsequent request wait on the former, and then when evaluating it should be a no-op, so I guess this isn't the end of the world.

hamishknight: This doc comment doesn't quite seem true though:

```
        // target's preparation status with a longer-running task. The key benefit here is that when we get many
        // preparation requests for the same target (eg. one for every text document request sent to a file), we don't
        // re-create new `PreparationTaskDescription`s for every preparation request. 
```

ahoppen: Ah, I was just missing a `break` here.

PR #1308, opened 2024-05-16
Add devcontainer for building SK-LSP

adam-fowler:
- Uses swiftlang/swift:nightly-6.0-jammy as a base image
- Installs `sqlite3`, `libsqlite3-dev`, `libncurses5-dev` and `python3`.
- Adds build arguments `-Xcxx -I/usr/lib/swift -Xcxx -I/usr/lib/swift/Block`
- Includes swift extension

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,10 @@
+FROM swiftlang/swift:nightly-6.0-jammy
+
+RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true \
+    && apt-get -q update \
+    && apt-get -q dist-upgrade -y \
+    && apt-get -q install -y \
+      sqlite3 \
+      libsqlite3-dev \
+      libncurses5-dev \
+      python3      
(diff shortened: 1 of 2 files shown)


ahoppen:
Thanks for the PR @adam-fowler. There is already a PR up to add a dev container to SourceKit-LSP, let’s iterate on that: https://github.com/apple/sourcekit-lsp/pull/916.

I would appreciate your feedback on that PR. I think the pieces that differ are 
- You run `apt-get dist-uprgrade`. I am curious if that’s a recommended thing to do, I don’t do it on Dockerfiles I run at home. Is there any reason to not stick with the distro that comes with the base swiftlang Docker image?
- You set the following options, I’m not sure if they are necessary?
```json
        "ghcr.io/devcontainers/features/common-utils:2": {
            "installZsh": "false",
            "username": "vscode",
            "userUid": "1000",
            "userGid": "1000",
            "upgradePackages": "false"
        },
        "ghcr.io/devcontainers/features/git:1": {
            "version": "os-provided",
            "ppa": "false"
        },
```
- You set `"seccomp=unconfined"`. Do you remember what that’s needed for?
- You set `"remoteUser": "vscode"`, is that needed?

PR #1309, opened 2024-05-16
Address review comments to #1306

ahoppen:
A few minor fixes and the following bigger change

### Improve handling of main file vs header file during indexing

Essentially fix two issues in updating the index store:
1. If there was one task to index `HeaderA.h` through `main.c` and one to index `HeaderB.h` through `main.c`, we would not declare a dependency between them in the task scheduler, which meant that we could have two concurrent and racing index tasks for `main.c`. Declare a dependency between any two files that have the same main file
2. `UpdateIndexStoreTaskDescription` was computing the target to index a file in independently of `SemanticIndexManager`. While they currently always line up, we should pass the target in which to index a file to the `UpdateIndexStoreTaskDescription`. Only this way can we guarantee that we actually prepared the target that the file will be indexed in

### DIFF
--- a/Documentation/Files_To_Reindex.md
+++ b/Documentation/Files_To_Reindex.md
@@ -18,7 +18,9 @@ Alternatives would be:
 
 ## `.h`
 
-Affects all files that include the header (including via other headers). For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and so we don’t re-index any file. If it is, it’s likely that the user will modify the file that includes the new header file shortly after, which will re-index it.
+All files that include the header (including via other headers) might be affected by the change, similar to how all `.swift` files that import a module might be affected. Similar to modules, we choose to not re-index all files that include the header because of the same considerations mentioned above.
+
+To re-index the header, we pick one main file that includes the header and re-index that, which will effectively update the index for the header. For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and thus don't re-index it. If the header is included from somewhere lese, it’s likely that the user will modify the file that includes the new header file shortly after, which will index the header and establish the header to main file connection.
 
 ## `.c` / `.cpp` / `.m`
 
--- a/Documentation/Files_To_Reindex.md
+++ b/Documentation/Files_To_Reindex.md
@@ -20,7 +20,7 @@ Alternatives would be:
 
 All files that include the header (including via other headers) might be affected by the change, similar to how all `.swift` files that import a module might be affected. Similar to modules, we choose to not re-index all files that include the header because of the same considerations mentioned above.
 
-To re-index the header, we pick one main file that includes the header and re-index that, which will effectively update the index for the header. For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and thus don't re-index it. If the header is included from somewhere lese, it’s likely that the user will modify the file that includes the new header file shortly after, which will index the header and establish the header to main file connection.
+To re-index the header, we pick one main file that includes the header and re-index that, which will effectively update the index for the header. For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and thus don't index it. If the user wrote an include to the new header before creating the header itself, we don't know about that include from the existing index. But it’s likely that the user will modify the file that includes the new header file shortly after, which will index the header and establish the header to main file connection.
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

Documentation/Files_To_Reindex.md
```diff
@@ -18,7 +18,9 @@ Alternatives would be:
...
+All files that include the header (including via other headers) might be affected by the change, similar to how all `.swift` files that import a module might be affected. Similar to modules, we choose to not re-index all files that include the header because of the same considerations mentioned above.
+
+To re-index the header, we pick one main file that includes the header and re-index that, which will effectively update the index for the header. For existing headers, we know which files import a header from the existing index. For new header files, we assume that it hasn’t been included in any file yet and thus don't re-index it. If the header is included from somewhere lese, it’s likely that the user will modify the file that includes the new header file shortly after, which will index the header and establish the header to main file connection.
```
bnbarham, line 23: > For new header files, we assume that it hasn’t been included in any file yet and thus don't re-index it
re-index -> index given we haven't actually indexed it yet?

> If the header is included from somewhere lese
lese -> else. But also, is the case you're referring to here if it's no longer included in the original file?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -350,21 +352,21 @@ public final actor SemanticIndexManager {
       switch newState {
       case .executing:
         for file in files {
-          if case .scheduled((taskID, let task)) = self.indexStatus[file.uri] {
-            self.indexStatus[file.uri] = .executing((taskID, task))
+          if case .scheduled((taskID, let task)) = self.indexStatus[file.file.sourceFile] {
```
bnbarham, line 355: `file.file.sourceFile` is... amusing 😅. I don't have a great suggestion here outside of naming `sourceFile` -> `sourceUri` and then maybe the variable `file` -> `fileWithTarget`. Or just leave them but add a `let sourceFile = file.file.sourceFile` so we don't have the many references 🤷

bnbarham (approved):

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -104,7 +150,9 @@ public struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
     to currentlyExecutingTasks: [UpdateIndexStoreTaskDescription]
   ) -> [TaskDependencyAction<UpdateIndexStoreTaskDescription>] {
     return currentlyExecutingTasks.compactMap { (other) -> TaskDependencyAction<UpdateIndexStoreTaskDescription>? in
-      guard !other.filesToIndex.intersection(filesToIndex).isEmpty else {
+      guard
+        !other.filesToIndex.hasIntersection(with: filesToIndex, where: { $0.file.mainFile == $1.file.mainFile })
+      else {
```
hamishknight, line 155: Any concerns about this becoming quadratic as batch sizes grow? Or will the batch sizes always be limited by some constant?

hamishknight (approved):

Documentation/Files_To_Reindex.md
ahoppen: > lese -> else. But also, is the case you're referring to here if it's no longer included in the original file?

Changed it to 

> If the user wrote an include to the new header before creating the header itself, we don't know about that include from the existing index.

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Changed it to `fileAndTarget.file.sourceFile`. It’s at least a little better.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: I was thinking that building up sets for every `currentlyExecutingTasks` isn’t much cheaper either but on second thought I realized that we can just build up `selfMainFiles` at the start, which seems like it’s beneficial, especially in the worst case.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1313, opened 2024-05-16
When a file is changed, only mark targets that depend on it as out-of-date

ahoppen:


ahoppen:
@swift-ci Please test

Sources/SKCore/BuildSystem.swift
```diff
@@ -147,6 +147,15 @@ public protocol BuildSystem: AnyObject, Sendable {
...
   func topologicalSort(of targets: [ConfiguredTarget]) async -> [ConfiguredTarget]?
 
+  /// Returns the list of targets that might depend on the given target and that need to be re-prepared when a file in
+  /// `target` is modified.
+  ///
+  /// The returned list can be an over-approximation, in which case the indexer will perform more work than strictly
+  /// necessary by scheduling re-preparation of a target where it isn't necessary.
+  ///
+  /// Returning `nil` indicates that all targets should be considered depending on the given target.
+  func targets(dependingOn targets: [ConfiguredTarget]) async -> [ConfiguredTarget]?
```
bnbarham, line 157: Would taking `Target` make more sense here? We don't really care about the possibly large set of configured targets as input. I'd argue that's even the case for output, but we do need the configured target eventually since that's what we track preparation of.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -209,11 +209,16 @@ public final actor SemanticIndexManager {
...
-    // might be affected by the changed files.
-    preparationStatus.removeAll()
+    let targets = await changedFiles.asyncMap { await buildSystemManager.configuredTargets(for: $0) }.flatMap { $0 }
+    if let dependentTargets = await buildSystemManager.targets(dependingOn: targets) {
+      for dependentTarget in dependentTargets {
+        preparationStatus[dependentTarget] = nil
+      }
+    } else {
+      // We couldn't determine which targets depend on the modified targets. Be conservative and assume all of them do.
+      preparationStatus = [:]
```
bnbarham, line 219: We could not remove `targets` in this case, not that it's really expected to happen anyway.

bnbarham (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -209,11 +209,16 @@ public final actor SemanticIndexManager {
...
-    // we rely on a null build during preparation to fast re-preparation.
-    // Ideally, we would have more fine-grained dependency management here and only mark those targets out-of-date that
-    // might be affected by the changed files.
-    preparationStatus.removeAll()
+    let targets = await changedFiles.asyncMap { await buildSystemManager.configuredTargets(for: $0) }.flatMap { $0 }
+    if let dependentTargets = await buildSystemManager.targets(dependingOn: targets) {
+      for dependentTarget in dependentTargets {
+        preparationStatus[dependentTarget] = nil
+      }
+    } else {
```
hamishknight, line 217: The dependent targets won't include the target itself, right? Should we be setting that to `nil` too?

hamishknight (approved):
Same question as Ben regarding whether the dependencies should work with ConfiguredTargets, but otherwise LGTM

Sources/SemanticIndex/SemanticIndexManager.swift
bnbarham: A change to the target itself shouldn't re-trigger preparation for that target - we have those files in memory on the sourcekit side. Hence my other comment about:

> We could not remove targets in this case, not that it's really expected to happen anyway.

Even if we don't know what the dependencies are, we know that there's no need to prepare the target the file is in.

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: The only case where we’d be able to not remove the target itself is if there is only a single target. Say you have a project two targets `A` and `B` and you change one file in `A` and one file in `B` then you don’t know if they depend on each other and you need to assume that they do. So you need to invalidate both.

I don’t think it’s worth optimizing for the single module case. Really, the build system should provide more accurate information here.

Sources/SKCore/BuildSystem.swift
bnbarham: Ah, Alex brought up a very valid point here. It's entirely possible that while the concept of a "target" may contain the file, the underlying target + platform may not - consider settings for including/excluding files per platform. So ConfiguredTarget really is the right abstraction here, thanks for remembering that one :).

ahoppen: Added the following comment to `SemanticIndexManager.filesDidChange`

```swift
    // Note that configured targets are the right abstraction layer here (instead of a non-configured target) because a
    // build system might have targets that include different source files. Hence a source file might be in target T
    // configured for macOS but not in target T configured for iOS.
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SemanticIndex/SemanticIndexManager.swift
bnbarham: Single target is the *main* case though 😅 I agree the buildsystem should, but asking for preparation on every edit is really pretty bad.

ahoppen:
@swift-ci Please test Windows

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: In one of my follow-up PRs (not up yet), I’m re-implementing the entire up-to-date tracking of targets and that allows us to early-exit if we know that the target is up-to-date without any build system interaction.

PR #1318, opened 2024-05-17
Move `LocalConnection` to `LSPTestSupport`

ahoppen:
We weren’t logging requests sent to a `TestSourceKitLSPClient` because we were assuming that `JSONRPCConnection` logs those requests in `SourceKitLSPServer`. But no logging happens in `LocalConnection`, which `TestSourceKitLSPClient` uses.

Since `LangaugeServerProtcol` can’t depend on `LSPLogging`, move the type to `LSPTestsSupport`.

### DIFF
--- a/Sources/LSPTestSupport/LocalConnection.swift
+++ b/Sources/LSPTestSupport/LocalConnection.swift
@@ -1,0 +1,133 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
Seems like it might be worth making a new module that a simple client could use, eg. you could imagine someone depending on sourcekit-lsp wanting an in-process client.

ahoppen:
> Seems like it might be worth making a new module that a simple client could use, eg. you could imagine someone depending on sourcekit-lsp wanting an in-process client.

I will probably do exactly that today for a debug utility ;-)

PR #1319, opened 2024-05-17
Remove status logging from `TaskScheduler`

ahoppen:
The logging from `TaskScheduler` was really verbose (and somewhat useful during its development) but it turns out that it’s just noise in real-world debugging. So, let’s remove it.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #1322, opened 2024-05-21
Split up-to-date status tracking and index progress tracking

ahoppen:
We were mixing the up-to-date status and in-progress status of an index task in `SemanticIndexManager`. This meant that a single `QueuedTask` in the task scheduler could be needed for eg. both preparation for editor functionality in a file of that target and to re-index a file in that target. This dual ownership made it unclear, which caller would be entitled to cancel the task. Furthermore, we needed to duplicate some logic from the preparation task dependencies in `SemanticIndexManager.prepare`.

To simplify things:
- Split the up-to-date status and the in-progress status into two different data structures
- Make the caller of `prepare` and `scheduleIndex` responsible for cancellation of the task it has scheduled. `TaskScheduler` might receive more scheduled tasks this way but the additional tasks should all be no-ops because the status is known to be up-to-date when they execute.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -15,40 +15,44 @@ import LSPLogging
 import LanguageServerProtocol
 import SKCore
 
-/// Describes the state of indexing for a single source file
-private enum IndexStatus<T> {
-  /// The index is up-to-date.
-  case upToDate
-  /// The file or target is not up to date. We have scheduled a task to update the index store for the file / prepare
-  /// the target, but that index operation hasn't been started yet.
-  case scheduled(T)
-  /// We are currently actively updating the index store for the file / preparing the target, ie. we are running a
-  /// subprocess that updates the index store / prepares a target.
-  case executing(T)
+/// A wrapper around `QueuedTask` that only allows equality comparison and inspection whether the `QueuedTask` is
+/// currently executing.
+///
+/// This way we can store `QueuedTask` in the `inProgress` dictionaries while guaranteeing that whoever created the
+/// queued task still has exclusive ownership of the task and can thus control the task's cancellation.
+private struct OpaqueQueuedIndexTask: Equatable {
+  private let task: QueuedTask<AnyIndexTaskDescription>
 
-  var description: String {
-    switch self {
-    case .upToDate:
-      return "upToDate"
-    case .scheduled:
-      return "scheduled"
-    case .executing:
-      return "executing"
-    }
+  var isExecuting: Bool {
+    task.isExecuting
+  }
+
+  init(_ task: QueuedTask<AnyIndexTaskDescription>) {
+    self.task = task
+  }
+
+  static func == (lhs: OpaqueQueuedIndexTask, rhs: OpaqueQueuedIndexTask) -> Bool {
+    return lhs.task === rhs.task
   }
 }
 
-/// See `SemanticIndexManager.preparationStatus`
-private struct PreparationTaskStatusData {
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

etcwilde:
https://github.com/apple/swift/pull/73785

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -488,7 +468,10 @@ public final actor SemanticIndexManager {
         // setting it to `.scheduled` because we don't have an `await` call between the creation of `indexTask` and
         // this loop, so we still have exclusive access to the `SemanticIndexManager` actor and hence `updateIndexStore`
         // can't execute until we have set all index statuses to `.scheduled`.
-        indexStatus[file.sourceFile] = .scheduled((taskID, indexTask))
+        inProgressIndexTasks[file.sourceFile] = .waitingForPreparation(
+          preparationTaskID: preparationTaskID,
+          indexTask: indexTask
+        )
```
hamishknight, line 474: Could we potentially run into a case?

1. We kick off indexing for a single file
2. We then schedule the indexing for a whole bunch of files that include that file
3. `inProgressIndexTasks` gets set to the new preparation task id for the large group of files
4. The indexing job for the single file gets completed, but that won't be reflected in `inProgressIndexTasks`, so `inProgressIndexFiles` will still list it as running
5. We then have to wait for the large file job to complete before the file is marked as up-to-date

hamishknight (commented):
LGTM

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -216,21 +230,26 @@ public final actor SemanticIndexManager {
...
     let targets = await changedFiles.asyncMap { await buildSystemManager.configuredTargets(for: $0) }.flatMap { $0 }
     if let dependentTargets = await buildSystemManager.targets(dependingOn: targets) {
-      for dependentTarget in dependentTargets {
-        preparationStatus[dependentTarget] = nil
-      }
+      await preparationUpToDateStatus.markOutOfDate(dependentTargets)
     } else {
       // We couldn't determine which targets depend on the modified targets. Be conservative and assume all of them do.
-      preparationStatus = [:]
+      await indexStoreUpToDateStatus.markOutOfDate(changedFiles)
```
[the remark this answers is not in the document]
ahoppen: Indeed

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Yes, we can run into that case but I don’t think it matters because the two uses of `inProgressIndexFiles` are:
- Index progress reporting: The effect of this will be that the numerator of the progress will only be incremented to reflect that the single file is up-to-date only when the batch is indexed, ie. it’s lower by 1 for a little while longer. Since we’re also waiting for other files to be indexed, I don’t think it actually matters
- Waiting for an up-to-date index: In that case we want to wait for all index tasks to finish anyway, including the batch.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -216,21 +230,26 @@ public final actor SemanticIndexManager {
...
+
+      await preparationUpToDateStatus.markAllOutOfDate()
+      // `markAllOutOfDate` only marks targets out-of-date that have been indexed before. Also mark all targets with
```
bnbarham, line 246: `markAllKnownOutOfDate`?

PR #1325, opened 2024-05-22
Fix issue that caused topological order to be incorrect

ahoppen:


ahoppen:
@swift-ci Please test

PR #1326, opened 2024-05-22
Produce an index log for the client

### LINKED ISSUE #1265: Send trace message to editor if indexing of a file fails
We should investigate if it makes sense to report errors during preparation and index store updates to the editor eg via the trace messages. Maybe we can implement something similar to the index log in Xcode this way.

rdar://127474136

### LINKED ISSUE #1299: Show build graph generation in work done progress
While we’re generating the build graph (aka resolving packages in SwiftPM) during background, we don’t show any progress update, which makes it seem like background indexing isn’t doing anything. We should show the build graph generation as a work done progress.

ahoppen:
This allows a user of SourceKit-LSP to inspect the result of background indexing. I think this gives useful insights into what SourceKit-LSP is indexing and why/how it fails, if it fails, also for users of SourceKit-LSP.

rdar://127474136
rdar://128071435
Fixes #1265
Fixes #1299

### DIFF
--- a/Sources/SourceKitLSP/IndexProgressManager.swift
+++ b/Sources/SourceKitLSP/IndexProgressManager.swift
@@ -11,19 +11,13 @@
 //===----------------------------------------------------------------------===//
 
 import LanguageServerProtocol
+import SKCore
 import SKSupport
 import SemanticIndex
 
 /// Listens for index status updates from `SemanticIndexManagers`. From that information, it manages a
 /// `WorkDoneProgress` that communicates the index progress to the editor.
 actor IndexProgressManager {
-  /// A queue on which `indexTaskWasQueued` and `indexStatusDidChange` are handled.
-  ///
-  /// This allows the two functions two be `nonisolated` (and eg. the caller of `indexStatusDidChange` doesn't have to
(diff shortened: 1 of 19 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/IndexProgressManager.swift
```diff
@@ -70,23 +64,47 @@ actor IndexProgressManager {
...
       workDoneProgress = nil
       return
     }
 
-    let finishedTasks = queuedIndexTasks - scheduled.count - executing.count
-    let message = "\(finishedTasks) / \(queuedIndexTasks)"
+    // We can get into a situation where queuedIndexTasks < indexTasks.count if we haven't processed all
+    // `indexTasksWereScheduled` calls yet but the semantic index managers already track them in their in-progress tasks.
+    // Clip the finished tasks to 0 because showing a negative number there looks stupid.
+    let finishedTasks = max(queuedIndexTasks - indexTasks.count, 0)
```
hamishknight, line 91: Just realized that `queuedIndexTasks` will be artificially high if you schedule a re-index for a file multiple times, right?

PR #1333, opened 2024-05-22
Fix a race condition that caused `IndexProgressManager` to create two `WorkDoneProgress`

ahoppen:


ahoppen:
@swift-ci Please test

PR #1334, opened 2024-05-22
Respect failed work done progress creation

ahoppen:
When the client replies with an error to the `CreateWorkDoneProgressRequest`, we shouldn’t be sending updates for it.

### DIFF
--- a/Sources/SourceKitLSP/WorkDoneProgressManager.swift
+++ b/Sources/SourceKitLSP/WorkDoneProgressManager.swift
@@ -22,6 +22,19 @@ final class WorkDoneProgressManager {
   private let token: ProgressToken
   private let queue = AsyncQueue<Serial>()
   private let server: SourceKitLSPServer
+  /// `true` if the client returned without an error from the `CreateWorkDoneProgressRequest`.
+  ///
+  /// Since all work done progress reports are being sent on `queue`, we never access it in a state where the
+  /// `CreateWorkDoneProgressRequest` is still in progress.
+  ///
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

Sources/SourceKitLSP/WorkDoneProgressManager.swift
```diff
@@ -22,6 +22,19 @@ final class WorkDoneProgressManager {
...
+  /// `CreateWorkDoneProgressRequest` is still in progress.
+  ///
+  /// Must be a reference because `deinit` captures it and wants to observe changes to it from `init` eg. in the
+  /// following:
+  ///  - `init` is called
+  ///  - `deinit` is called
+  ///  - The task from `init` gets executed
+  ///  - The task from `deinit` gets executed
+  ///    - This should have `workDoneProgressCreated == true` so that it can send the work progress end.
+  private let workDoneProgressCreated: ThreadSafeBox<Bool> & AnyObject = ThreadSafeBox<Bool>(initialValue: false)
```
hamishknight, line 37: Isn't `& AnyObject` redundant here?

hamishknight (approved):

Sources/SourceKitLSP/WorkDoneProgressManager.swift
ahoppen: That’s to catch any issues that might arise if `ThreadSafeBox` became a non-reference type in the future.

> Must be a reference because `deinit` captures it and wants to observe changes to it from `init`

PR #1338, opened 2024-05-23
Fix a non-deterministic failure of `testBackgroundIndexingStatusWorkDoneProgress`

ahoppen:
We didn’t take into account that there are also work done progress notifications for package reloading. Only listen for


ahoppen:
@swift-ci Please test

hamishknight (approved):
Thanks!

PR #1339, opened 2024-05-23
Don’t cancel build when closing a document

ahoppen:
In SourceKit-LSP, we can get into the following situation:
1. We open A.swift
2. We issue a request for A.swift, the request takes a while to execute
3. The dependencies of A.swift are updated, which causes us to reopen the document in sourcekitd, so that the AST is rebuilt
4. This shouldn’t cause the request from (2) to be cancelled. We should continue executing it and only re-open the document after the request from (2) has finished

While at it, also make sure that we send `syntacticOnly` when re-opening a document.

rdar://127475366

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -335,13 +335,7 @@ extension SwiftLanguageService {
     ])
     _ = try? await self.sourcekitd.send(closeReq, fileContents: nil)
 
-    let openReq = sourcekitd.dictionary([
-      keys.request: self.requests.editorOpen,
-      keys.name: path,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -382,6 +382,7 @@ extension SwiftLanguageService {
       keys.enableDiagnostics: 0,
       keys.syntacticOnly: 1,
       keys.compilerArgs: compileCommand?.compilerArgs as [SKDRequestValue]?,
+      keys.cancelBuilds: 0,
```
hamishknight, line 385: This should be set for the close rather than the open, right?

PR #1340, opened 2024-05-23
Watch for changes to source files

ahoppen:
We used to only watch for file creation and deletion because that might modify build settings but for background indexing, we also need to watch for changes to the files, so we can invalidate the up-to-date status of the target.


ahoppen:
@swift-ci Please test

hamishknight (approved):

bnbarham (approved):

PR #1346, opened 2024-05-24
On macOS, reveal diagnostic bundle in Finder after it has been created

ahoppen:
Removes a step to copy paste the bundle path to find it in Finder.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1347, opened 2024-05-24
Set `cancelBuilds` to 0 on close instead of open

ahoppen:
Addresses https://github.com/apple/sourcekit-lsp/pull/1339#discussion_r1612353958

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -326,13 +326,7 @@ extension SwiftLanguageService {
     cancelInFlightPublishDiagnosticsTask(for: snapshot.uri)
     await diagnosticReportManager.removeItemsFromCache(with: snapshot.uri)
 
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -430,6 +429,7 @@ extension SwiftLanguageService {
     let req = sourcekitd.dictionary([
       keys.request: self.requests.editorClose,
       keys.name: uri.pseudoPath,
+      keys.cancelBuilds: 0,
```
hamishknight, line 432: Wasn't the intention to do it for `reopenDocument`? (or should we do it for both?)

bnbarham: Yeah, I'd say we really only want for re-open

ahoppen:
@swift-ci Please test

PR #1352, opened 2024-05-25
Don’t re-index files in languages that cannot be indexed

ahoppen:
Swift packages can have source files that we can’t index (like assembly files). When re-opening, we schedule indexing for those files, which requires the targets to be re-prepared. We skip them earlier.

rdar://128711633


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight:
LGTM

PR #1354, opened 2024-05-25
Time out updating the index store after 2 minutes

ahoppen:
Time out updating of the index store after 2 minutes. We don't expect any single file compilation to take longer than 2 minutes in practice, so this indicates that the compiler has entered some kind of loop. We will try indexing the file again when it is edited or when the project is re-opened.

I did not add a timeout for preparation because it’s much harder to tell what an upper bound could be here. Preparation of a high-level target might take quite a while if none of the lower-level targets have been prepared yet.

rdar://128732571

### DIFF
--- a/Sources/SKSupport/AsyncUtils.swift
+++ b/Sources/SKSupport/AsyncUtils.swift
@@ -166,3 +166,28 @@ extension Collection where Element: Sendable {
     }
   }
 }
+
+public struct TimeoutError: Error, CustomStringConvertible {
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #1356, opened 2024-05-25
Show work done progress while a source file is being prepared for editor functionality

ahoppen:
While `SemanticIndexManager.inProgressPrepareForEditorTask` is not `nil`, show a work done progress in the editor that the current file is being prepared for editor functionality.

I decided to use the indexing progress indicator for this for now because I think it’s too spammy if SourceKit-LSP has two different progress indicators for preparation and indexing. I’ll need to see how this feels like in practice.

rdar://128722609


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1357, opened 2024-05-25
A couple minor priority elevation bug fixes in `TaskScheduler`

ahoppen:
Fix two issue:
1. If a task’s priority got elevated, we would emit the `Elevating priority` log message repeatedly until the elevated priority is propagated to the underlying task via `withTaskPriorityChangedHandler`
2. Explicitly set the priority of `QueuedTask`, which ensures that we create the detached `resultTask` inside of it with that priority. I’m not sure if this was an actual issue but seems like a good cleanup regardless.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1362, opened 2024-05-28
Fix a memory leak in `BackgroundIndexingTests.testPrepareTargetAfterEditToDependency`

ahoppen:
The closure that handled the `DiagnosticsRefreshRequest` was retaining `project`, which constituted a retain cycle.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1365, opened 2024-05-28
Add paths from compiler arguments spelled with `=` to the diagnose bundle

ahoppen:
Eg. if the crash contained `-fmodule-map-file=/path/to/module.modulemap`, we weren’t including `/path/to/module.modulemap` in the diagnose bundle.


ahoppen:
@swift-ci Please test

Sources/Diagnose/ReproducerBundle.swift
```diff
@@ -47,14 +58,14 @@ func makeReproducerBundle(for requestInfo: RequestInfo, toolchain: Toolchain, bu
...
-      try? FileManager.default.createDirectory(at: dest.deletingLastPathComponent(), withIntermediateDirectories: true)
-      try? FileManager.default.copyItem(at: URL(fileURLWithPath: compilerArg), to: dest)
+    if compilerArg.hasPrefix("/") {
+      copyPathToBundleIfNecessary(compilerArg, bundlePath: bundlePath)
+    } else if let firstEqual = compilerArg.firstIndex(of: "=") {
+      // Copy files from eg. `-fmodule-map-file=/path/to/module.modulemap`
+      let possiblePath = String(compilerArg[compilerArg.index(after: firstEqual)...])
+      if possiblePath.hasPrefix("/") {
+        copyPathToBundleIfNecessary(possiblePath, bundlePath: bundlePath)
+      }
```
hamishknight, line 68: Do we care about cases like `-I/foo/bar`?

hamishknight (approved):

Sources/Diagnose/ReproducerBundle.swift
ahoppen: Let’s just look for the first `/` to start the path. Eventually I’ll need to think about how this should work on Windows as well but right now the entire diagnose bundle pretty heavily relies on oslog scraping, which is only applicable for macOS anyway.

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

PR #1366, opened 2024-05-29
Support running tests that require building with a Swift 5.10 toolchain

ahoppen:
When we detect that we’re running using a 5.10 toolchain, adjust the compiler arguments that we received from SwiftPM to drop any `/Modules` suffixes in the build directory, to account for the fact that 5.10 stores the modules one directory higher up (https://github.com/apple/swift-package-manager/pull/7103).

### DIFF
--- a/Sources/SKCore/Toolchain.swift
+++ b/Sources/SKCore/Toolchain.swift
@@ -12,12 +12,50 @@
 
 import LSPLogging
 import LanguageServerProtocol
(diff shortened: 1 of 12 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #1368, opened 2024-05-29
Relax assertion around `testIntegrationTest`

ahoppen:
Depending on whether clangd has finished indexing/built an AST/something, we get different code completion results. The actual integration test is only checking that we have a result with `insertText` `clib_func`, which is satisfied independently of whether clangd’s progress. Adopt the same assertion in this test case.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #1371, opened 2024-05-31
Make tests pass with Xcode 15.4

ahoppen:
We are no longer skipping these tests, which means that we need to make them account for different error messages emitted by Swift 5.10.

I added warnings behind `#if compiler(>=6.1)` to give us a reminder that we can remove these checks when we no longer support running SourceKit-LSP with SwiftPM from Swift 5.10. Swift 6.1 doesn’t have to be this cut-off point but it’s the most likely candidate for now if we want to support the current and last Swift version from tip SourceKit-LSP.


bnbarham (approved):

hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1372, opened 2024-05-31
Relax `testMainFileChangesIfIncludeIsAdded` to allow `clangd` to return diagnostics from old build settings

ahoppen:
`clangd` may return diagnostics from the old build settings sometimes (I believe when it's still building the preamble for shared.h when the new build settings come in). Check that it eventually returns the correct diagnostics and allow it to return outdated diagnostics for a short while.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

bnbarham:

PR #1376, opened 2024-05-31
Flush stdout after issuing message that we re-run tests in serial

ahoppen:
For some reason the log message that we’re re-running tests in serial always showed up after the serial run in the build log output. Test if flushing stdout fixes the issue.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test macOS

PR #1377, opened 2024-05-31
Replace an assertion by a fault log in `WorkDoneProgressState`

ahoppen:
We didn’t hit this assertion but I noticed it and the best practice in sourcekit-lsp is to log a fault instead of throwing an assertion for any kind of recoverable error.


bnbarham (approved):

hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1379, opened 2024-05-31
Don’t show a “Preparing current file” progress while determining target of file

ahoppen:
Every time a request regarding a text document is sent, we call `schedulePreparationForEditorFunctionality` on the `SemanticIndexManager` with the assumption that it’s a no-op if the target is already prepared. We did, however, show the “Preparing current file” status while we were determining the file‘s target to determine that the file’s target is up-to-date.

### DIFF
--- a/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
+++ b/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
@@ -868,4 +868,23 @@ final class BackgroundIndexingTests: XCTestCase {
     let message = try await project.testClient.nextNotification(ofType: ShowMessageNotification.self)
     XCTAssert(message.message.contains("Background indexing"), "Received unexpected message: \(message.message)")
   }
+
+  func testNoPreparationStatusIfTargetIsUpToDate() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Lib.swift": ""
+      ],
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -868,4 +868,23 @@ final class BackgroundIndexingTests: XCTestCase {
...
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Lib.swift": ""
+      ],
+      capabilities: ClientCapabilities(window: WindowClientCapabilities(workDoneProgress: true)),
+      enableBackgroundIndexing: true
+    )
+
+    // Opening the document prepares it for editor functionality. Its target is already prepared, so we shouldn't show
+    // a work done progress for it.
```
bnbarham, line 882: What make sit already prepared?

bnbarham (approved):

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
ahoppen: Creating a `SwiftPMTestProject` polls the index (`pollIndex` parameter in the initializer defaults to `true`), so just creating the project will ensure that the index is up-to-date.

hamishknight (commented):
LGTM (given you added the `defer` in another PR)

PR #1381, opened 2024-05-31
Don’t catch `CancellationError` for document diagnostics

ahoppen:
We were catching errors during diagnostic generation because VS Code will not request diagnostics again if SourceKit-LSP returns an error to any diagnostic request. We were, however, a little over-eager when doing this and also caught cancellation errors, which means that if the user made multiple edits in quick succession (which would cancel the diagnostics request from the first edit), we would return empty diagnostics, clearing the displayed diagnostics.


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

hamishknight (commented):
LGTM

PR #1384, opened 2024-05-31
Add a few more arguments to allow errors during indexing

ahoppen:
I noticed that we were missing a bunch more arguments (in particular for clang index invocations), while diagnosing rdar://129071600, which is fixed by passing `-experimental-allow-module-with-compiler-errors` to the Swift index invocations.

rdar://129071600

### DIFF
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -430,11 +430,6 @@ private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
     .option("output-file-map", [.singleDash], [.separatedBySpace, .separatedByEqualSign]),
   ]
 
-  let removeFrontendFlags = [
-    "-experimental-skip-non-inlinable-function-bodies",
-    "-experimental-skip-all-function-bodies",
-  ]
-
   var result: [String] = []
   result.reserveCapacity(compilerArguments.count)
   var iterator = compilerArguments.makeIterator()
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -430,11 +430,6 @@ private func adjustSwiftCompilerArgumentsForIndexStoreUpdate(
     .option("output-file-map", [.singleDash], [.separatedBySpace, .separatedByEqualSign]),
   ]
 
-  let removeFrontendFlags = [
-    "-experimental-skip-non-inlinable-function-bodies",
-    "-experimental-skip-all-function-bodies",
-  ]
-
```
hamishknight, line 437: Don't we want to remove these for indexing?

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -887,4 +887,47 @@ final class BackgroundIndexingTests: XCTestCase {
     _ = try project.openDocument("Lib.swift")
     _ = try await project.testClient.send(BarrierRequest())
   }
+
+  func testImportPreparedModuleWithFunctionBodiesSkipped() async throws {
+    // This test case was crashing the indexing compiler invocation for Client if Lib was built for index preparation
+    // (using `-enable-library-evolution -experimental-skip-all-function-bodies -experimental-lazy-typecheck`) but x
```
bnbarham, line 893: `x`?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
bnbarham: They would presumably only be there if we added them, which we're currently. not. Probably doesn't hurt to keep anyway though.

PR #1385, opened 2024-05-31
Improve folding ranges if editor only supports line folding

ahoppen:
If the folding range doesn't end at the end of the last line, exclude that line from the folding range since the end line gets folded away. This means if we reported `end.line`, we would eg. fold away the `}` that matches a `{`, which looks surprising.

If the folding range does end at the end of the line we are in cases that don't have a closing indicator (like comments), so we can fold the last line as well.

### DIFF
--- a/Sources/SourceKitLSP/Swift/FoldingRange.swift
+++ b/Sources/SourceKitLSP/Swift/FoldingRange.swift
@@ -12,6 +12,7 @@
 
 import LSPLogging
 import LanguageServerProtocol
+import SKSupport
 import SwiftSyntax
 
 fileprivate final class FoldingRangeFinder: SyntaxAnyVisitor {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

xedin (approved):
Works great, thank you!

bnbarham (approved):

Sources/SourceKitLSP/Swift/FoldingRange.swift
```diff
@@ -264,3 +277,14 @@ extension SwiftLanguageService {
...
 }
+
+fileprivate extension LineTable {
+  func isAtEndOfLine(_ position: Position) -> Bool {
+    guard position.line >= 0, position.line < self.count else {
+      return false
+    }
+    let line = self[position.line]
+    let suffixAfterPositionColumn = line[line.utf16.index(line.startIndex, offsetBy: position.utf16index)...]
+    return suffixAfterPositionColumn.allSatisfy(\.isNewline)
```
hamishknight, line 288: Should we care about trailing whitespace too?

ahoppen: I don’t think it matters in practice because this is really only for comment folding, which would include the whitespace in the comment’s trivia and thus point to the position right before the newline. 

Let’s leave it for now because I don’t have a clean name for this function if we also consider whitespaces.

PR #1386, opened 2024-06-01
Forward all errors that translate to `.cancelled` from document diagnostics request

ahoppen:
We were only forwarding `CancellationError` but depending on when cancellation happens, we would get `SKDError.cancelled`, which we would translate to empty diagnostics, causing the test to fail.


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

hamishknight (commented):
LGTM

PR #1387, opened 2024-06-01
Skip testImportPreparedModuleWithFunctionBodiesSkipped if toolchain doesn’t support rename

ahoppen:
The test uses rename and should thus be skipped if sourcekitd doesn’t support rename.

This makes the tests pass with Xcode 15.4.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (commented):
LGTM

PR #1388, opened 2024-06-01
Set `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER` for parallel test execution

ahoppen:
Set `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER` to 1 so we generate a log to stderr if parallel test execution fails, which helps diagnose the issue.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1389, opened 2024-06-01
Add a general notion of experimental features to sourcekit-lsp

ahoppen:
Background indexing probably won’t be the last experimental feature in sourcekit-lsp that we want to gate behind a feature flag. Instead of adding new parameters ad-hoc, introduce a general notion of experimental features.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer+Options.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer+Options.swift
@@ -49,6 +49,9 @@ extension SourceKitLSPServer {
     /// notification when running unit tests.
     public var swiftPublishDiagnosticsDebounceDuration: TimeInterval
 
+    /// Experimental features that are enabled.
+    public var experimentalFeatures: [ExperimentalFeature]
+
     public var indexTestHooks: IndexTestHooks
 
     public init(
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

gottesmm:
@ahoppen I am pretty sure this broke the linux PR build: https://ci.swift.org/job/oss-swift-pr-test-ubuntu-20_04/

Sources/SourceKitLSP/SourceKitLSPServer+Options.swift
```diff
@@ -49,6 +49,9 @@ extension SourceKitLSPServer {
     /// notification when running unit tests.
     public var swiftPublishDiagnosticsDebounceDuration: TimeInterval
 
+    /// Experimental features that are enabled.
+    public var experimentalFeatures: [ExperimentalFeature]
```
hamishknight, line 53: Any reason not to make a `Set`?

hamishknight (commented):
LGTM

Sources/SourceKitLSP/SourceKitLSPServer+Options.swift
hamishknight: Yeah I wasn't really thinking of performance, more just semantically it feels like `Set` better fits. But I don't feel strongly at all 🤷

ahoppen: Neither do I. It’s in as an array now, let’s just leave it.

PR #1394, opened 2024-06-01
Require `ThreadSafeBox.T` to be Sendable

ahoppen:
Otherwise, I think `ThreadSafeBox` might still have data races. This also requires us to make `TestSourceKitLSPClient.RequestHandler` sendable.

rdar://128572489

### DIFF
--- a/Sources/SKSupport/ThreadSafeBox.swift
+++ b/Sources/SKSupport/ThreadSafeBox.swift
@@ -24,7 +24,7 @@ extension NSLock {
 /// A thread safe container that contains a value of type `T`.
 ///
 /// - Note: Unchecked sendable conformance because value is guarded by a lock.
-public class ThreadSafeBox<T>: @unchecked Sendable {
+public class ThreadSafeBox<T: Sendable>: @unchecked Sendable {
   /// Lock guarding `_value`.
   private let lock = NSLock()
 
@@ -40,6 +40,11 @@ public class ThreadSafeBox<T>: @unchecked Sendable {
       lock.withLock {
         _value = newValue
       }
+    }
+    _modify {
+      lock.lock()
+      defer { lock.unlock() }
+      yield &_value
     }
   }
 
--- a/Sources/SKTestSupport/TestSourceKitLSPClient.swift
+++ b/Sources/SKTestSupport/TestSourceKitLSPClient.swift
@@ -34,7 +34,7 @@ extension SourceKitLSPServer.Options {
 /// that the server sends to the client.
 public final class TestSourceKitLSPClient: MessageHandler {
   /// A function that takes a request and returns the request's response.
-  public typealias RequestHandler<Request: RequestType> = (Request) -> Request.Response
+  public typealias RequestHandler<Request: RequestType> = @Sendable (Request) -> Request.Response
 
   /// The ID that should be assigned to the next request sent to the `server`.
   /// `nonisolated(unsafe)` is fine because `nextRequestID` is atomic.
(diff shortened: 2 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

Sources/SKTestSupport/TestSourceKitLSPClient.swift
```diff
@@ -72,7 +72,7 @@ public final class TestSourceKitLSPClient: MessageHandler {
   ///
   /// `isOneShort` if the request handler should only serve a single request and should be removed from
   /// `requestHandlers` after it has been called.
-  private nonisolated(unsafe) var requestHandlers: ThreadSafeBox<[(requestHandler: Any, isOneShot: Bool)]> =
+  private nonisolated(unsafe) var requestHandlers: ThreadSafeBox<[(requestHandler: Sendable, isOneShot: Bool)]> =
```
hamishknight, line 75: One thing I've just noticed here is that when we're doing `requestHandlers.value.append`, there's a gap between where we read the array and write back the array where e.g the request handler could remove one of the handlers, but then we'd immediately write it back

Sources/SKSupport/ThreadSafeBox.swift
```diff
@@ -24,7 +24,7 @@ extension NSLock {
 /// A thread safe container that contains a value of type `T`.
 ///
 /// - Note: Unchecked sendable conformance because value is guarded by a lock.
-public class ThreadSafeBox<T>: @unchecked Sendable {
```
hamishknight, line 27: Out of curiosity, what issues are being caused by it not being `Sendable`?

hamishknight (approved):

Sources/SKSupport/ThreadSafeBox.swift
ahoppen: I think the main issue that I noticed was that some of the request handles in `TestSourceKitLSPClient` weren’t sendable. I didn’t technically hit any bug from it but I was wondering why I could modify local variables from the surrounding context from within the request handler, which generally isn’t safe.

Sources/SKTestSupport/TestSourceKitLSPClient.swift
ahoppen: Oh, good catch. It’s only test code, so a little bit of 🤷🏽 but I changed them to use `withLock`. Do you have an idea if there’s a way to prevent this in the type system? I couldn’t come up with anything but maybe you have ideas. If not: Whatever, should be fine.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKTestSupport/TestSourceKitLSPClient.swift
hamishknight: I _think_ using `_modify` should allow it to be mutated while covered by the lock.

ahoppen: Wohooo. That seems to have worked.

ahoppen:
@swift-ci Please test

Sources/SKSupport/ThreadSafeBox.swift
```diff
@@ -41,6 +41,11 @@ public class ThreadSafeBox<T>: @unchecked Sendable {
         _value = newValue
       }
     }
+    _modify {
+      lock.lock()
+      yield &_value
+      lock.unlock()
```
hamishknight, line 47: Nit: best to use a `defer` here since the code after `yield` won't be executed if the caller throws, e.g `try x.value.someThrowingFn()`, `yield` is more like `try yield` in that case

ahoppen: Oh, that’s subtle

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1395, opened 2024-06-01
Use `DocumentURI` instead of `URL` in more locations

ahoppen:
The originator for this change was that I noticed that we weren’t able to index files that contain a `+` because we wouldn’t get build settings for them. This is because `SwiftPMBuildSystem` had a URL comparison to check if the file that we are requesting build settings for is part of the target’s source files. Now, the target’s source files represented the `+` as a literal `+` while the requested file name had it escaped as `%2B`, so we incorrectly inferred that the source file was not part of the target.

`DocumentURI` accounts for this and compares the scheme + normalized path separately. So, the solution is to work in terms of `DocumentURI` in `SwiftPMBuildSystem`.

While doing this, I also changed other files to use `DocumentURI` instead of `URL` in case they have similar issues.

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
@@ -146,8 +146,8 @@ public actor SwiftPMBuildSystem {
     }
   }
 
-  var fileToTarget: [AbsolutePath: SwiftBuildTarget] = [:]
-  var sourceDirToTarget: [AbsolutePath: SwiftBuildTarget] = [:]
+  var fileToTarget: [DocumentURI: SwiftBuildTarget] = [:]
+  var sourceDirToTarget: [DocumentURI: SwiftBuildTarget] = [:]
 
   /// Maps configured targets ids to their SwiftPM build target as well as an index in their topological sorting.
   ///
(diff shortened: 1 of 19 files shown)


ahoppen:
@swift-ci Please test

Tests/SKSwiftPMWorkspaceTests/SwiftPMBuildSystemTests.swift
```diff
@@ -179,6 +183,52 @@ final class SwiftPMBuildSystemTests: XCTestCase {
...
+        workspacePath: packageRoot,
+        toolchainRegistry: tr,
+        fileSystem: localFileSystem,
+        buildSetup: SourceKitLSPServer.Options.testDefault.buildSetup,
+        isForIndexBuild: false
+      )
+      try await swiftpmBuildSystem.generateBuildGraph(allowFileSystemWrites: false)
+
+      let aPlusSomething = packageRoot.appending(components: "Sources", "lib", "a+something.swift")
+      print(aPlusSomething.asURL)
```
hamishknight, line 215: Left over from debugging?

Tests/SKSwiftPMWorkspaceTests/SwiftPMBuildSystemTests.swift
```diff
@@ -179,6 +183,52 @@ final class SwiftPMBuildSystemTests: XCTestCase {
...
+      assertEqual(await swiftpmBuildSystem.buildPath, build)
+      assertNotNil(await swiftpmBuildSystem.indexStorePath)
+      let arguments = try await unwrap(
+        swiftpmBuildSystem.buildSettings(
+          for: DocumentURI(URL(string: "file://\(aPlusSomething.asURL.path.replacing("+", with: "%2B"))")!),
+          language: .swift
+        )
+      )
+      .compilerArguments
+      print(arguments)
```
hamishknight, line 228: Left over from debugging?

hamishknight (approved):

Tests/SKSwiftPMWorkspaceTests/SwiftPMBuildSystemTests.swift
ahoppen: Yes, I guess I should add an actual check 🙈

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -449,9 +462,10 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
       return nil
     }
 
-    if !buildTarget.sources.contains(url),
+    if !buildTarget.sources.lazy.map(DocumentURI.init).contains(uri),
       let substituteFile = buildTarget.sources.sorted(by: { $0.path < $1.path }).first
     {
+      logger.info("😅 Getting compiler arguments for \(url) using substitute file \(substituteFile)")
```
bnbarham, line 468: Is the emoji intentional?

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1396, opened 2024-06-01
Show container name in call hierarchy item name instead of detail field

ahoppen:
The container name isn’t displayed very prominently and it’s easy to miss. I think the call hierarchy looks nicer when we prepend the container name with language-specific logic.

Before
Screenshot 2024-06-01 at 16 53 23

After
Screenshot 2024-06-01 at 16 59 05

rdar://128396648


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1397, opened 2024-06-02
Show `deinit` in document symbols

ahoppen:
rdar://128715297

### DIFF
--- a/Sources/SourceKitLSP/Swift/DocumentSymbols.swift
+++ b/Sources/SourceKitLSP/Swift/DocumentSymbols.swift
@@ -170,6 +170,16 @@ fileprivate final class DocumentSymbolsFinder: SyntaxAnyVisitor {
     return .skipChildren
   }
 
+  override func visit(_ node: DeinitializerDeclSyntax) -> SyntaxVisitorContinueKind {
+    return record(
+      node: node,
+      name: node.deinitKeyword.text,
+      symbolKind: .null,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

Sources/SourceKitLSP/Swift/DocumentSymbols.swift
```diff
@@ -170,6 +170,16 @@ fileprivate final class DocumentSymbolsFinder: SyntaxAnyVisitor {
     return .skipChildren
   }
 
+  override func visit(_ node: DeinitializerDeclSyntax) -> SyntaxVisitorContinueKind {
+    return record(
+      node: node,
+      name: node.deinitKeyword.text,
+      symbolKind: .null,
```
bnbarham, line 177: Kind of weird, but could make constructor?

ahoppen: I thought about it as well but then felt that `constructor` is weird because it’s not a constructor… But I don’t have strong opinions here.

ahoppen: I just checked and VS Code at least uses the same icon for constructor and function, so it’s odd if `deinit` uses a different icon. `clangd` uses `constructor`, so let’s just do that as well.


Screenshot 2024-06-03 at 09 07 41

PR #1398, opened 2024-06-02
Add an extension to the LogMessageNotification to add a message to a specific log in the client

ahoppen:
If the editor has support for this LSP extension, it can create a separate output view for the index log.

For example, in VS Code, we could have one output view for the standard LSP logging (which are the messages between VS Code and SourceKit-LSP if verbose logging is enabled in VS Code) and a separate log that shows information about background indexing.

rdar://128572032


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (commented):
LGTM

PR #1400, opened 2024-06-02
Show “Preparing targets” as index status if we only have preparation tasks and no index tasks

ahoppen:
I was wondering for a while why we were showing “Indexing 0 / 0” as the index progress. I think it’s if we only have a preparation task but no index tasks running. I’m not entirely sure how this happens because preparation should only happen for two reasons:
- We are preparing a target so we can index a files -> We should have an active index task
- We are preparing a file for editor functionality of the current file -> we should have a `inProgressPrepareForEditorTask`

Maybe I’ll understand it once we have this change in. It seems like a worthwhile change in any way.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (commented):
LGTM

PR #1403, opened 2024-06-02
Share module cache between test projects

ahoppen:
This improves serial test execution time of eg. `DocumentTestDiscoveryTests` from 36s to 22s because we don’t need to re-build the XCTest module from its interface when using an open source toolchain.

This also uncovered that we weren‘t passing the build setup flags to the prepare command.

rdar://126493151

### DIFF
--- a/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
+++ b/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
@@ -574,6 +574,30 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
       "--disable-index-store",
       "--target", target.targetID,
     ]
+    if self.toolsBuildParameters.configuration != self.destinationBuildParameters.configuration {
+      logger.fault(
+        """
+        Preparation is assuming that tools and destination are built using the same configuration, \
+        got tools: \(String(describing: self.toolsBuildParameters.configuration), privacy: .public), \
+        destination: \(String(describing: self.destinationBuildParameters.configuration), privacy: .public)
+        """
+      )
+    }
+    arguments += ["-c", self.destinationBuildParameters.configuration.rawValue]
+    if self.toolsBuildParameters.flags != self.destinationBuildParameters.flags {
+      logger.fault(
+        """
+        Preparation is assuming that tools and destination are built using the same build flags, \
+        got tools: \(String(describing: self.toolsBuildParameters.flags)), \
+        destination: \(String(describing: self.destinationBuildParameters.configuration))
+        """
+      )
+    }
+    arguments += self.destinationBuildParameters.flags.cCompilerFlags.flatMap { ["-Xcc", $0] }
+    arguments += self.destinationBuildParameters.flags.cxxCompilerFlags.flatMap { ["-Xcxx", $0] }
+    arguments += self.destinationBuildParameters.flags.swiftCompilerFlags.flatMap { ["-Xswiftc", $0] }
+    arguments += self.destinationBuildParameters.flags.linkerFlags.flatMap { ["-Xlinker", $0] }
+    arguments += self.destinationBuildParameters.flags.xcbuildFlags?.flatMap { ["-Xxcbuild", $0] } ?? []
     if await swiftBuildSupportsPrepareForIndexing {
       arguments.append("--experimental-prepare-for-indexing")
     }
--- a/Sources/SKTestSupport/Utils.swift
+++ b/Sources/SKTestSupport/Utils.swift
@@ -112,3 +112,15 @@ fileprivate extension URL {
     #endif
   }
 }
+
+var globalModuleCache: URL? = {
+  if let customModuleCache = ProcessInfo.processInfo.environment["SOURCEKIT_LSP_TEST_MODULE_CACHE"] {
+    if customModuleCache.isEmpty {
+      return nil
+    }
+    return URL(fileURLWithPath: customModuleCache)
+  }
+  return FileManager.default.temporaryDirectory.realpath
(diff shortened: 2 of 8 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -574,6 +574,31 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+        destination: \(String(describing: self.destinationBuildParameters.configuration))
+        """
+      )
+      logger.fault("Preparation is assuming that tools and destination to have the same build flags")
+    }
+    arguments += self.destinationBuildParameters.flags.cCompilerFlags.flatMap { ["-Xcc", $0] }
+    arguments += self.destinationBuildParameters.flags.cxxCompilerFlags.flatMap { ["-Xcxx", $0] }
+    arguments += self.destinationBuildParameters.flags.swiftCompilerFlags.flatMap { ["-Xswiftc", $0] }
+    arguments += self.destinationBuildParameters.flags.linkerFlags.flatMap { ["-Xlinker", $0] }
+    arguments += self.destinationBuildParameters.flags.xcbuildFlags?.flatMap { ["-Xxcbuild", $0] } ?? []
```
bnbarham, line 601: Why are these needed? The prepare should already have these shouldn't it?

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
```diff
@@ -574,6 +574,31 @@ extension SwiftPMBuildSystem: SKCore.BuildSystem {
...
+    arguments += ["-c", self.destinationBuildParameters.configuration.rawValue]
+    if self.toolsBuildParameters.flags != self.destinationBuildParameters.flags {
+      logger.fault(
+        """
+        Preparation is assuming that tools and destination are built using the same build flags, \
+        got tools: \(String(describing: self.toolsBuildParameters.flags)), \
+        destination: \(String(describing: self.destinationBuildParameters.configuration))
+        """
+      )
+      logger.fault("Preparation is assuming that tools and destination to have the same build flags")
```
bnbarham, line 595: Duplicate, but also this is going to be fairly spammy.

Sources/SKTestSupport/Utils.swift
```diff
@@ -112,3 +112,16 @@ fileprivate extension URL {
     #endif
   }
 }
+
+var globalModuleCache: URL? = {
+  if let customModuleCache = ProcessInfo.processInfo.environment["SOURCEKIT_LSP_TEST_MODULE_CACHE"] {
+    if customModuleCache.isEmpty {
+      return nil
+    } else {
```
bnbarham, line 120: No need for the else

Utilities/build-script-helper.py
```diff
@@ -228,15 +229,17 @@ def run_tests(swift_exec: str, args: argparse.Namespace) -> None:
...
-    try:
-        check_call(cmd + ['--parallel'], additional_env=additional_env, verbose=args.verbose)
-    except:
-        print('--- Running tests in parallel failed. Re-running tests serially to capture more actionable output.')
-        sys.stdout.flush()
-        check_call(cmd, additional_env=additional_env, verbose=args.verbose)
-        # Return with non-zero exit code even if serial test execution succeeds.
-        raise SystemExit(1)
+    with tempfile.TemporaryDirectory() as test_module_cache:
+        additional_env['SOURCEKIT_LSP_TEST_MODULE_CACHE'] = f"{test_module_cache}/module-cache"
```
bnbarham, line 233: Should we delete the module cache after the tests have run?

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: These are flags that you pass when launching SourceKit-LSP and that are equivalent to the `-Xswiftc` flags passed to `swift build`. For example, if you need to build using `swift build -Xcxx -I/usr/lib/swift -Xcxx -I/usr/lib/swift/Block` then you should launch sourcekit-lsp using `sourcekit-lsp -Xcxx -I/usr/lib/swift -Xcxx -I/usr/lib/swift/Block`, which will then now also pass them to the `swift build` commands it runs as part of background indexing.

Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift
ahoppen: I don’t expect this ever to be logged because we set the same build flags for the tools and destination build parameters here: https://github.com/apple/sourcekit-lsp/blob/b9c74387f297147d82716195974306384ca59c11/Sources/SKSwiftPMWorkspace/SwiftPMBuildSystem.swift#L245-L259

The intention of the log is that if that assumption is broken in the future, we’ll have a log message that’s pointing us to it.

Utilities/build-script-helper.py
ahoppen: We do, it’s a `tempfile.TemporaryDirectory` and I’m pretty sure that those get deleted when they get destroyed.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Utilities/build-script-helper.py
bnbarham: Heh, I somehow missed the `with` here 👍

hamishknight (approved):

PR #1404, opened 2024-06-02
Add tests to rename C++ symbols exposed to Swift

ahoppen:
This exposes an issue where we wouldn’t rename a symbol on the clang side f a symbol was only defined in a header and not a C/C++/... file. In that case we failed to determine the language of the header file because there was no unit file that contained the header file. The cleaner solution here is to ask for the symbol provider of each occurrence directly.

rdar://127391127


ahoppen:
https://github.com/apple/indexstore-db/pull/200

@swift-ci Please test

hamishknight (approved):

ahoppen:
https://github.com/apple/indexstore-db/pull/200

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1405, opened 2024-06-03
Log when a sourcekitd request gets cancelled

ahoppen:
I saw a failure in CI where a sourcekitd diagnostic request got cancelled without an LSP cancellation request. I am suspecting that there is some implicit cancellation going on in sourcekitd despite `key.cancel_builds: 0`, which doesn’t make sense to me. Adding some explicit logging to sourcekit-lsp to check if we cancel the request for some reason.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #1408, opened 2024-06-03
Classify deinitializers as constructors in document symbols

ahoppen:
LSP doesn't have a destructor kind. constructor is the closest match and also what clangd for destructors.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1412, opened 2024-06-03
Wait for cancellation to propagate in `testDontReturnEmptyDiagnosticsIfDiagnosticRequestIsCancelled`

ahoppen:
I saw a few non-deterministic test failures. I think the issue was that handling of the `CancelRequestNotification` is done asynchronously, which left a short window in which the sourcekitd diagnostics request could run and return results instead of being cancelled. Wait for the diagnostic request to actually be cancelled before running it for real.

While doing this, also introduce proper sourcekitd test hooks instead of relying on the preparation test hooks, which just got run as a side effect.

### DIFF
--- a/Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
+++ b/Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
@@ -25,6 +25,14 @@ extension sourcekitd_api_keys: @unchecked Sendable {}
 extension sourcekitd_api_requests: @unchecked Sendable {}
 extension sourcekitd_api_values: @unchecked Sendable {}
 
+public struct SourceKitDTestHooks: Sendable {
+  public var sourcekitdRequestDidStart: (@Sendable (SKDRequestDictionary) -> Void)?
+
+  public init(sourcekitdRequestDidStart: (@Sendable (SKDRequestDictionary) -> Void)? = nil) {
+    self.sourcekitdRequestDidStart = sourcekitdRequestDidStart
+  }
+}
+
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
```diff
@@ -104,6 +119,8 @@ public actor DynamicallyLoadedSourceKitD: SourceKitD {
   }
 
   public nonisolated func log(request: SKDRequestDictionary) {
+    // `log(request:)` is called in `SourceKitD.send` before the request is sent, so we're hooking in here.
+    testHooks.sourcekitdRequestDidStart?(request)
```
hamishknight, line 123: Any reason not to put this in `func send`? I guess to avoid exposing it in the protocol?

hamishknight (approved):

Sources/SourceKitD/DynamicallyLoadedSourceKitD.swift
ahoppen: I was thinking because protocols can’t have stored properties. But somehow didn’t occur to me before that we could just make the test hooks a protocol requirement… 🤦🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1419, opened 2024-06-04
Populate inProgressRequest with the same priority that we cancel requests

ahoppen:
Others, this could lead to interesting priority inversions: We would handle the cancellation with higher priority than the population of inProgressRequest`, which means that we wouldn’t actually cancel the request, logging `Cannot cancel request <request-id> because it hasn't been scheduled for execution yet or because the request already returned a response`.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -612,7 +612,7 @@ extension SourceKitLSPServer: MessageHandler {
     // Keep track of the ID -> Task management with low priority. Once we cancel
     // a request, the cancellation task runs with a high priority and depends on
     // this task, which will elevate this task's priority.
-    cancellationMessageHandlingQueue.async(priority: .background) {
+    cancellationMessageHandlingQueue.async(priority: .high) {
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -612,7 +612,7 @@ extension SourceKitLSPServer: MessageHandler {
     // Keep track of the ID -> Task management with low priority. Once we cancel
     // a request, the cancellation task runs with a high priority and depends on
     // this task, which will elevate this task's priority.
-    cancellationMessageHandlingQueue.async(priority: .background) {
+    cancellationMessageHandlingQueue.async(priority: .high) {
```
hamishknight, line 615: `cancellationMessageHandlingQueue` is an `AsyncQueue`, so shouldn't that enforce that this runs before handling a subsequent cancellation?

ahoppen: Now that you mention it: I did think about this when I wrote the code originally. And also looking at the logs again, it seems like the issue is that we just returned results. I’ll open a new PR that fixes that issue instead.

PR #1420, opened 2024-06-04
Demote two log messages issued on the `error` level

ahoppen:
Just two cases where we were logging an error that wasn’t strictly necessary and was just spamming the log at the high log levels.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test macOS

PR #1421, opened 2024-06-04
Keep track of recently finished requests

ahoppen:
This is only used so we don't log an error when receiving a `CancelRequestNotification` for a request that has just returned a response.


ahoppen:
@swift-ci Please test

hamishknight (approved):

bnbarham (approved):

PR #1431, opened 2024-06-05
Fix a race condition in `LocalConnection`

ahoppen:
The comment on `LocalConnection` assumed that all mutable state was guarded by `queue` but `state` and `handler` actually weren’t. This was found by the thread sanitizer.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1432, opened 2024-06-05
Only report log collection progress in `sourcekit-lsp diagnose` if we’ve made meaningful progress

ahoppen:
Especially when we didn’t have a redrawing progress indicator, we would spam the output with out making progress that was visible to the user because the overall progress just shows a percentage without any decimal digits. To fix this, only report progress when we’ve made at least 1% of progress in log collection.

### DIFF
--- a/Sources/Diagnose/DiagnoseCommand.swift
+++ b/Sources/Diagnose/DiagnoseCommand.swift
@@ -25,6 +25,12 @@ import class TSCUtility.PercentProgressAnimation
 @MainActor
 private var progressBar: PercentProgressAnimation? = nil
 
+/// The last progress that was reported on the progress bar. This ensures that when the progress indicator uses the
+/// `MultiLinePercentProgressAnimation` (eg. because stderr is redirected to a file) we don't emit status updates
+/// without making any real progress.
+@MainActor
+private var lastProgress: (Int, String)? = nil
+
 /// A component of the diagnostic bundle that's collected in independent stages.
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/Diagnose/DiagnoseCommand.swift
```diff
@@ -222,7 +223,12 @@ public struct DiagnoseCommand: AsyncParsableCommand {
             // The log is larger than we expected. Halt at 100%
             progress = 1
           }
-          reportProgress(.collectingLogMessages(progress: progress), message: "Collecting log messages")
+          if progress - lastReportedProgress > 0.01 {
+            // Only call `reportProgress` when we've made at least 1% of progress (which will get scaled down further
+            // based on the other components). This avoids us spamming the console with non-meaningful progress updates.
```
hamishknight, line 228: IMO it seems like this is something that should be handled by `reportProgress`, but I don't feel strongly about it

hamishknight (approved):

Sources/Diagnose/DiagnoseCommand.swift
ahoppen: Yes, that’s probably the better design. Updated.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1434, opened 2024-06-05
Set the priority of processes launched for background indexing

ahoppen:
Unfortunately, `setpriority` only allows reduction of a process’s priority and doesn’t support priority elevation (unless you are a super user). I still think that it’s valuable to set the process’s priority based on the task priority when it is launched because many indexing processes never get their priority escalated and should thus run in the background.

On Windows, we can elevate the process’s priority.

rdar://127474245


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

Issue #755, opened 2023-06-09
Show expanded code of freestanding macro expansions in hover


adam-fowler:
Xcode has a nice feature where you can expand macros to see the code they generate. It'd be cool if you could do this with VSCode as well. Could we add a CodeAction or something similar to provide the code expansion?

ahoppen:
Tracked in Apple’s issue tracker as rdar://110516432

ahoppen:
I think we do have a code action that does this. If you create a new macro template and build it, you can right click on `stringify`  in main.swift and select *Refactor…* -> *Inline Macro*.

It’s not quite what Xcode does with the code roll. If you know if VSCode has an option to show the expanded source code inline, we should be able to figure something out how to give you the necessary expansion information.

ahoppen:
I suspect that that would probably require dedicated support in the Swift VSCode extension. If you figure out how to display those inline text hint, I’m happy to create an LSP request that gives you the text to display.

ahoppen:
AFAICT there’s nothing actionable for SourceKit-LSP here, so I’m closing this issue. Could you file another issue once you’ve got support in the VSCode plugin to display the macro expansion and then we can discuss what kind of SourceKit-LSP request you need?

adam-fowler:
What's your thoughts on including the macro expansion in the `textDocument/hover` output. eg
```
### Expands to
(x + y, "x + y")
```

ahoppen:
I don’t think that scales. In a lot of cases the generated code will span over multiple lines and I don’t think that makes for a good hover response. Also, for attached macros, you would want to show where the generated code is being inserted (e.g. in case of a peer macro) and that also doesn’t fit with the design of hover.

adam-fowler:
I don't think spanning over multiple lines is a huge issue. The discussion part of a hover comment can cover many lines. Obviously the positioning is a different issue.

We might not be given the UI capabilities in VSCode to insert read only text sections into a text document. So this maybe the only option.

ahoppen:
Do you know which other plugins do show read-only code inline and how they do it? Is it only supported for first-party plugins by Microsoft?

adam-fowler:
This is how cpp-tools does it 
![](https://user-images.githubusercontent.com/51793967/251263415-fa8e14b8-407e-4149-bf7a-789f05cc04d6.png)
ie as part of the hover window

[...]

PR #1436, opened 2024-06-05
Add LSP support for showing Macro Expansions

lokesh-tr:
**Note: This Feature is Experimental 🧪. You have to first enable this feature by passing `--experimental-feature show-macro-expansions` to `sourcekit-lsp`**

This PR aims to bring a streamlined way of showing macro expansions through the LSP.

This allows users to pick a code action on a macro which performs an `ExecuteCommandRequest` with an `ExpandMacroCommand`. The server then generates the macro expansion (using semantic refactoring request) and stores it in a temporary directory. The server also performs a `ShowDocumentRequest` ([LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument)) on the client. Thus, displaying the expansion to the users.

`ShowDocumentRequest` being a part of the LSP spec will allow us to make macro expansions accessible across wide range of editors.

The result of this PR is that macro expansions work identical to that of Xcode with the only difference being Xcode showing the expansions inline, while we don't have a standardised LSP spec for displaying inline. (Although some customisation is possible for *VS Code* alone through their API).

#### How it works?
1. `sourcekitd` returns an "Inline Macro" Code Action when the editor requests for the available code actions on a macro.
2. We inject an "Expand Macro" Code Action having a `ExpandMacroCommand` before passing the response of `sourcekitd` to the editor.
3. If the user decides to pick the "Expand Macro" Code Action, then the editor sends an `ExecuteCommandRequest` containing the `ExpandMacroCommand`.
4. Upon receiving the `ExecuteCommandRequest` containing the `ExpandMacroCommand`, We then ask `sourcekitd` for the expansion of the given macro by making a semantic refactoring request.
5. The response is interpreted as a `MacroExpansion`, and for each edit (expansion / replacement) it suggested, we make a temporary file as follows:
`<NSTemporaryDirectory>/sourcekit-lsp/GeneratedMacroExpansions/<edit.bufferName>/<SourceFileName>_<LineColumnRange>.<SourceFileExtension>`

Here,
- `<edit.bufferName>` will be the buffer name associated with each edit, it looks like `@__swiftmacro_..._.swift` without the ".swift" extension at last.
- `<LineColumnRange>` will look like `LaCb-LcCd` where a & c represent Line number of start and end of expansion and b & d represent Column number of start and end of expansion.
6. Once, the temporary file has been created, we make a `ShowDocumentRequest`, containing the URI of temporary file, to the editor.
7. Now, the editor shows the temporary file containing the macro expansion to the user. 🎉

#### Follow-up Work & Future Directions:
1. The current implementation lays foundation for macro expansions of any kind, but `@attached` macro expansions is little buggy. LSP support for `@attached` macro expansions will be fixed in the following PR: https://github.com/apple/sourcekit-lsp/pull/1479
2. This implementation is **not** 1:1 with Xcode's "Expand Macro" feature. The only difference being the fact that we open separate documents for each expansion edit, whereas Xcode shows them all inline. This is a known limitation due to the LSP specifications having no support for showing anything inline the document.
**But for people using VS Code, we are looking forward to implement some custom functionality to show expansions inline with the help of VS Code API.** ~This will be implemented in a future PR.~
(Implemented in 
https://github.com/swiftlang/sourcekit-lsp/pull/1479
https://github.com/swiftlang/vscode-swift/pull/945)

#### References:
 - https://github.com/apple/sourcekit-lsp/issues/755 
 - https://github.com/apple/sourcekit-lsp/pull/892
 - https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument
 - https://docs.swift.org/swift-book/documentation/the-swift-programming-language/macros/
 - https://github.com/apple/swift-evolution/blob/main/proposals/0415-function-body-macros.md
 - https://github.com/apple/swift/pull/66295
 
 Fixes https://github.com/apple/sourcekit-lsp/issues/755
 rdar://110516432
 
### DIFF
--- a/Sources/SourceKitLSP/Swift/ExpandMacroCommand.swift
+++ b/Sources/SourceKitLSP/Swift/ExpandMacroCommand.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -13,22 +13,27 @@
 import LanguageServerProtocol
 import SourceKitD
 
-public struct SemanticRefactorCommand: RefactorCommand {
-  typealias Response = SemanticRefactoring
+public struct ExpandMacroCommand: RefactorCommand {
+  typealias Response = MacroExpansion
 
-  public static let identifier: String = "semantic.refactor.command"
+  public static let identifier: String = "expand.macro.command"
 
   /// The name of this refactoring action.
-  public var title: String
+  public var title = "Expand Macro"
 
   /// The sourcekitd identifier of the refactoring action.
-  public var actionString: String
+  public var actionString = "source.refactoring.kind.expand.macro"
 
-  /// The range to refactor.
+  /// The range to expand.
   public var positionRange: Range<Position>
 
   /// The text document related to the refactoring action.
   public var textDocument: TextDocumentIdentifier
+
+  public init(positionRange: Range<Position>, textDocument: TextDocumentIdentifier) {
+    self.positionRange = positionRange
+    self.textDocument = textDocument
+  }
 
   public init?(fromLSPDictionary dictionary: [String: LSPAny]) {
     guard case .dictionary(let documentDict)? = dictionary[CodingKeys.textDocument.stringValue],
@@ -43,6 +48,7 @@ public struct SemanticRefactorCommand: RefactorCommand {
     else {
       return nil
     }
+
     self.init(
       title: title,
       actionString: actionString,
@@ -51,8 +57,12 @@ public struct SemanticRefactorCommand: RefactorCommand {
     )
   }
 
-  public init(title: String, actionString: String, positionRange: Range<Position>, textDocument: TextDocumentIdentifier)
-  {
+  public init(
+    title: String,
+    actionString: String,
+    positionRange: Range<Position>,
+    textDocument: TextDocumentIdentifier
+  ) {
     self.title = title
     self.actionString = actionString
     self.positionRange = positionRange
@@ -66,42 +76,5 @@ public struct SemanticRefactorCommand: RefactorCommand {
       CodingKeys.positionRange.stringValue: positionRange.encodeToLSPAny(),
       CodingKeys.textDocument.stringValue: textDocument.encodeToLSPAny(),
     ])
-  }
-}
-
-extension Array where Element == SemanticRefactorCommand {
-  init?(
-    array: SKDResponseArray?,
-    range: Range<Position>,
-    textDocument: TextDocumentIdentifier,
-    _ keys: sourcekitd_api_keys,
-    _ api: sourcekitd_api_functions_t
-  ) {
-    guard let results = array else {
-      return nil
-    }
-    var commands = [SemanticRefactorCommand]()
-    results.forEach { _, value in
-      if let name: String = value[keys.actionName],
-        let actionuid: sourcekitd_api_uid_t = value[keys.actionUID],
-        let ptr = api.uid_get_string_ptr(actionuid)
-      {
-        let actionName = String(cString: ptr)
-        guard !actionName.hasPrefix("source.refactoring.kind.rename.") else {
-          // TODO: Rename.
-          return true
-        }
-        commands.append(
-          SemanticRefactorCommand(
-            title: name,
-            actionString: actionName,
-            positionRange: range,
-            textDocument: textDocument
-          )
-        )
-      }
-      return true
-    }
-    self = commands
   }
 }
--- a/Sources/SourceKitLSP/Swift/MacroExpansion.swift
+++ b/Sources/SourceKitLSP/Swift/MacroExpansion.swift
@@ -1,0 +1,132 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LSPLogging
+import LanguageServerProtocol
+import SourceKitD
+
+/// Detailed information about the result of a macro expansion operation.
+///
+/// Wraps the information returned by sourcekitd's `semantic_refactoring`
+/// request, such as the necessary macro expansion edits.
+struct MacroExpansion: RefactoringResponse {
+  /// The title of the refactoring action.
+  var title: String
+
+  /// The URI of the file where the macro is used
+  var uri: DocumentURI
+
+  /// The resulting array of `RefactoringEdit` of a semantic refactoring request
+  var edits: [RefactoringEdit]
+
+  init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit]) {
+    self.title = title
+    self.uri = uri
+    self.edits = refactoringEdits.compactMap { refactoringEdit in
+      if refactoringEdit.bufferName == nil && !refactoringEdit.newText.isEmpty {
+        logger.fault("Unable to retrieve some parts of the expansion")
+        return nil
+      }
+
+      return refactoringEdit
+    }
+  }
+}
+
+extension SwiftLanguageService {
+  /// Handles the `ExpandMacroCommand`.
+  ///
+  /// Makes a request to sourcekitd and wraps the result into a `MacroExpansion`
+  /// and then makes a `ShowDocumentRequest` to the client side for each
+  /// expansion to be displayed.
+  ///
+  /// - Parameters:
+  ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this request.
+  ///
+  /// - Returns: A `[RefactoringEdit]` with the necessary edits and buffer name as a `LSPAny`
+  func expandMacro(
+    _ expandMacroCommand: ExpandMacroCommand
+  ) async throws -> LSPAny {
+    guard let sourceKitLSPServer else {
+      // `SourceKitLSPServer` has been destructed. We are tearing down the
+      // language server. Nothing left to do.
+      throw ResponseError.unknown("Connection to the editor closed")
+    }
+
+    guard let sourceFileURL = expandMacroCommand.textDocument.uri.fileURL else {
+      throw ResponseError.unknown("Given URI is not a file URL")
+    }
+
+    let expansion = try await self.refactoring(expandMacroCommand)
+
+    for macroEdit in expansion.edits {
+      if let bufferName = macroEdit.bufferName {
+        // buffer name without ".swift"
+        let macroExpansionBufferDirectoryName =
+          bufferName.hasSuffix(".swift")
+          ? String(bufferName.dropLast(6))
+          : bufferName
+
+        let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
+          .appendingPathComponent(macroExpansionBufferDirectoryName)
+        do {
+          try FileManager.default.createDirectory(
+            at: macroExpansionBufferDirectoryURL,
+            withIntermediateDirectories: true
+          )
+        } catch {
+          throw ResponseError.unknown(
+            "Failed to create directory for macro expansion buffer at path: \(macroExpansionBufferDirectoryURL.path)"
+          )
+        }
+
+        // name of the source file
+        let macroExpansionFileName = sourceFileURL.deletingPathExtension().lastPathComponent
+
+        // github permalink notation for position range
+        let macroExpansionPositionRangeIndicator =
+          "L\(macroEdit.range.lowerBound.line)C\(macroEdit.range.lowerBound.utf16index)-L\(macroEdit.range.upperBound.line)C\(macroEdit.range.upperBound.utf16index)"
+
+        let macroExpansionFilePath =
+          macroExpansionBufferDirectoryURL
+          .appendingPathComponent(
+            "\(macroExpansionFileName)_\(macroExpansionPositionRangeIndicator).\(sourceFileURL.pathExtension)"
+          )
+
+        do {
+          try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: .utf8)
+        } catch {
+          throw ResponseError.unknown(
+            "Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\""
+          )
+        }
+
+        Task {
+          let req = ShowDocumentRequest(uri: DocumentURI(macroExpansionFilePath), selection: macroEdit.range)
+
+          let response = await orLog("Sending ShowDocumentRequest to Client") {
+            try await sourceKitLSPServer.sendRequestToClient(req)
+          }
+
+          if let response, !response.success {
+            logger.error("client refused to show document for \(expansion.title, privacy: .public)")
+          }
+        }
+      } else if !macroEdit.newText.isEmpty {
+        logger.fault("Unable to retrieve some parts of macro expansion")
+      }
+    }
+
+    return expansion.edits.encodeToLSPAny()
+  }
+}
--- a/Sources/SourceKitLSP/Swift/Refactoring.swift
+++ b/Sources/SourceKitLSP/Swift/Refactoring.swift
@@ -1,0 +1,129 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LSPLogging
+import LanguageServerProtocol
+import SourceKitD
+
+protocol RefactoringResponse {
+  init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit])
+}
+
+extension RefactoringResponse {
+  /// Create an instance of `RefactoringResponse` from a sourcekitd semantic
+  /// refactoring response dictionary, if possible.
+  ///
+  /// - Parameters:
+  ///   - title: The title of the refactoring action.
+  ///   - dict: Response dictionary to extract information from.
+  ///   - snapshot: The snapshot that triggered the `semantic_refactoring` request.
+  ///   - keys: The sourcekitd key set to use for looking up into `dict`.
+  init?(_ title: String, _ dict: SKDResponseDictionary, _ snapshot: DocumentSnapshot, _ keys: sourcekitd_api_keys) {
+    guard let categorizedEdits: SKDResponseArray = dict[keys.categorizedEdits] else {
+      logger.fault("categorizedEdits doesn't exist in response dictionary")
+      return nil
+    }
+
+    var refactoringEdits = [RefactoringEdit]()
+
+    categorizedEdits.forEach { _, categorizedEdit in
+      guard let edits: SKDResponseArray = categorizedEdit[keys.edits] else {
+        logger.fault("edits doesn't exist in categorizedEdit dictionary")
+        return true
+      }
+      edits.forEach { _, edit in
+        guard let startLine: Int = edit[keys.line],
+          let startColumn: Int = edit[keys.column],
+          let endLine: Int = edit[keys.endLine],
+          let endColumn: Int = edit[keys.endColumn],
+          let text: String = edit[keys.text]
+        else {
+          logger.fault("Failed to deserialise edit dictionary containing values: \(edit)")
+          return true  // continue
+        }
+
+        // The LSP is zero based, but semantic_refactoring is one based.
+        let startPosition = snapshot.positionOf(
+          zeroBasedLine: startLine - 1,
+          utf8Column: startColumn - 1
+        )
+        let endPosition = snapshot.positionOf(
+          zeroBasedLine: endLine - 1,
+          utf8Column: endColumn - 1
+        )
+        // Snippets are only supported in code completion.
+        // Remove SourceKit placeholders in refactoring actions because they
+        // can't be represented in the editor properly.
+        let textWithSnippets = rewriteSourceKitPlaceholders(in: text, clientSupportsSnippets: false)
+        refactoringEdits.append(
+          RefactoringEdit(
+            range: startPosition..<endPosition,
+            newText: textWithSnippets,
+            bufferName: edit[keys.bufferName]
+          )
+        )
+        return true
+      }
+      return true
+    }
+
+    guard !refactoringEdits.isEmpty else {
+      logger.error("No refactoring edits found")
+      return nil
+    }
+
+    self.init(title: title, uri: snapshot.uri, refactoringEdits: refactoringEdits)
+  }
+}
+
+extension SwiftLanguageService {
+  /// Provides detailed information about the result of a specific refactoring
+  /// operation.
+  ///
+  /// Wraps the information returned by sourcekitd's `semantic_refactoring`
+  /// request, such as the necessary edits and placeholder locations.
+  ///
+  /// - Parameters:
+  ///   - refactorCommand: The semantic `RefactorCommand` that triggered this request.
+  /// - Returns: The response of the refactoring
+  func refactoring<T: RefactorCommand>(
+    _ refactorCommand: T
+  ) async throws -> T.Response {
+    let keys = self.keys
+
+    let uri = refactorCommand.textDocument.uri
+    let snapshot = try self.documentManager.latestSnapshot(uri)
+    let line = refactorCommand.positionRange.lowerBound.line
+    let utf16Column = refactorCommand.positionRange.lowerBound.utf16index
+    let utf8Column = snapshot.lineTable.utf8ColumnAt(line: line, utf16Column: utf16Column)
+
+    let skreq = sourcekitd.dictionary([
+      keys.request: self.requests.semanticRefactoring,
+      // Preferred name for e.g. an extracted variable.
+      // Empty string means sourcekitd chooses a name automatically.
+      keys.name: "",
+      keys.sourceFile: uri.pseudoPath,
+      // LSP is zero based, but this request is 1 based.
+      keys.line: line + 1,
+      keys.column: utf8Column + 1,
+      keys.length: snapshot.utf8OffsetRange(of: refactorCommand.positionRange).count,
+      keys.actionUID: self.sourcekitd.api.uid_get_from_cstr(refactorCommand.actionString)!,
+      keys.compilerArgs: await self.buildSettings(for: snapshot.uri)?.compilerArgs as [SKDRequestValue]?,
+    ])
+
+    let dict = try await self.sourcekitd.send(skreq, fileContents: snapshot.text)
+    guard let refactor = T.Response(refactorCommand.title, dict, snapshot, self.keys) else {
+      throw SemanticRefactoringError.noEditsNeeded(uri)
+    }
+    return refactor
+  }
+}
--- a/Sources/SourceKitLSP/Swift/RefactoringEdit.swift
+++ b/Sources/SourceKitLSP/Swift/RefactoringEdit.swift
@@ -1,0 +1,74 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SourceKitD
+
+/// Represents an edit from semantic refactor response. Notionally, a subclass of `TextEdit`
+@_spi(Testing) public struct RefactoringEdit: Hashable, Sendable, Codable {
+  /// The range of text to be replaced.
+  @CustomCodable<PositionRange>
+  public var range: Range<Position>
+
+  /// The new text.
+  public var newText: String
+
+  /// If the new text of the edit should not be applied to the original source
+  /// file but to a separate buffer, a fake name for that buffer. For example
+  /// for expansion of macros, this is @ followed by the mangled name of the
+  /// macro expansion, followed by .swift.
+  public var bufferName: String?
+
+  public init(range: Range<Position>, newText: String, bufferName: String?) {
+    self._range = CustomCodable<PositionRange>(wrappedValue: range)
+    self.newText = newText
+    self.bufferName = bufferName
+  }
+}
+
+extension RefactoringEdit: LSPAnyCodable {
+  public init?(fromLSPDictionary dictionary: [String: LSPAny]) {
+    guard case .dictionary(let rangeDict) = dictionary[CodingKeys.range.stringValue],
+      case .string(let newText) = dictionary[CodingKeys.newText.stringValue]
+    else {
+      return nil
+    }
+
+    guard let range = Range<Position>(fromLSPDictionary: rangeDict) else {
+      return nil
+    }
+
+    self._range = CustomCodable<PositionRange>(wrappedValue: range)
+    self.newText = newText
+
+    if case .string(let bufferName) = dictionary[CodingKeys.bufferName.stringValue] {
+      self.bufferName = bufferName
+    } else {
+      self.bufferName = nil
+    }
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
+    guard let bufferName = bufferName else {
+      return .dictionary([
+        CodingKeys.range.stringValue: range.encodeToLSPAny(),
+        CodingKeys.newText.stringValue: .string(newText),
+      ])
+    }
+
+    return .dictionary([
+      CodingKeys.range.stringValue: range.encodeToLSPAny(),
+      CodingKeys.newText.stringValue: .string(newText),
+      CodingKeys.bufferName.stringValue: .string(bufferName),
+    ])
+  }
+}
--- a/Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
+++ b/Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
@@ -9,13 +9,16 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
+
+import LSPLogging
 import LanguageServerProtocol
 import SourceKitD
 
 /// Detailed information about the result of a specific refactoring operation.
 ///
-/// Wraps the information returned by sourcekitd's `semantic_refactoring` request, such as the necessary edits and placeholder locations.
-struct SemanticRefactoring {
+/// Wraps the information returned by sourcekitd's `semantic_refactoring`
+/// request, such as the necessary edits and placeholder locations.
+struct SemanticRefactoring: RefactoringResponse {
 
   /// The title of the refactoring action.
   var title: String
@@ -28,57 +31,11 @@ struct SemanticRefactoring {
     self.edit = edit
   }
 
-  /// Create a `SemanticRefactoring` from a sourcekitd response dictionary, if possible.
-  ///
-  /// - Parameters:
-  ///   - title: The title of the refactoring action.
-  ///   - dict: Response dictionary to extract information from.
-  ///   - snapshot: The snapshot that triggered the `semantic_refactoring` request.
-  ///   - keys: The sourcekitd key set to use for looking up into `dict`.
-  init?(_ title: String, _ dict: SKDResponseDictionary, _ snapshot: DocumentSnapshot, _ keys: sourcekitd_api_keys) {
-    guard let categorizedEdits: SKDResponseArray = dict[keys.categorizedEdits] else {
-      return nil
-    }
-
-    var textEdits = [TextEdit]()
-
-    categorizedEdits.forEach { _, value in
-      guard let edits: SKDResponseArray = value[keys.edits] else {
-        return false
-      }
-      edits.forEach { _, value in
-        // The LSP is zero based, but semantic_refactoring is one based.
-        guard let startLine: Int = value[keys.line],
-          let startColumn: Int = value[keys.column],
-          let endLine: Int = value[keys.endLine],
-          let endColumn: Int = value[keys.endColumn],
-          let text: String = value[keys.text]
-        else {
-          return true  // continue
-        }
-        let startPosition = snapshot.positionOf(
-          zeroBasedLine: startLine - 1,
-          utf8Column: startColumn - 1
-        )
-        let endPosition = snapshot.positionOf(
-          zeroBasedLine: endLine - 1,
-          utf8Column: endColumn - 1
-        )
-        // Snippets are only supported in code completion.
-        // Remove SourceKit placeholders in refactoring actions because they can't be represented in the editor properly.
-        let textWithSnippets = rewriteSourceKitPlaceholders(in: text, clientSupportsSnippets: false)
-        textEdits.append(TextEdit(range: startPosition..<endPosition, newText: textWithSnippets))
-        return true
-      }
-      return true
-    }
-
-    guard textEdits.isEmpty == false else {
-      return nil
-    }
-
+  init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit]) {
     self.title = title
-    self.edit = WorkspaceEdit(changes: [snapshot.uri: textEdits])
+    self.edit = WorkspaceEdit(changes: [
+      uri: refactoringEdits.map { TextEdit(range: $0.range, newText: $0.newText) }
+    ])
   }
 }
 
@@ -103,41 +60,41 @@ extension SemanticRefactoringError: CustomStringConvertible {
 }
 
 extension SwiftLanguageService {
-  /// Provides detailed information about the result of a specific refactoring operation.
+
+  /// Handles the `SemanticRefactorCommand`.
   ///
-  /// Wraps the information returned by sourcekitd's `semantic_refactoring` request, such as the necessary edits and placeholder locations.
+  /// Sends a request to sourcekitd and wraps the result into a
+  /// `SemanticRefactoring` and then makes an `ApplyEditRequest` to the client
+  /// side for the actual refactoring.
   ///
   /// - Parameters:
-  ///   - refactorCommand: The semantic refactor `Command` that triggered this request.
+  ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
+  ///
+  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
   func semanticRefactoring(
-    _ refactorCommand: SemanticRefactorCommand
-  ) async throws -> SemanticRefactoring {
-    let keys = self.keys
-
-    let uri = refactorCommand.textDocument.uri
-    let snapshot = try self.documentManager.latestSnapshot(uri)
-    let line = refactorCommand.positionRange.lowerBound.line
-    let utf16Column = refactorCommand.positionRange.lowerBound.utf16index
-    let utf8Column = snapshot.lineTable.utf8ColumnAt(line: line, utf16Column: utf16Column)
-
-    let skreq = sourcekitd.dictionary([
-      keys.request: self.requests.semanticRefactoring,
-      // Preferred name for e.g. an extracted variable.
-      // Empty string means sourcekitd chooses a name automatically.
-      keys.name: "",
-      keys.sourceFile: uri.pseudoPath,
-      // LSP is zero based, but this request is 1 based.
-      keys.line: line + 1,
-      keys.column: utf8Column + 1,
-      keys.length: snapshot.utf8OffsetRange(of: refactorCommand.positionRange).count,
-      keys.actionUID: self.sourcekitd.api.uid_get_from_cstr(refactorCommand.actionString)!,
-      keys.compilerArgs: await self.buildSettings(for: snapshot.uri)?.compilerArgs as [SKDRequestValue]?,
-    ])
-
-    let dict = try await self.sourcekitd.send(skreq, fileContents: snapshot.text)
-    guard let refactor = SemanticRefactoring(refactorCommand.title, dict, snapshot, self.keys) else {
-      throw SemanticRefactoringError.noEditsNeeded(uri)
+    _ semanticRefactorCommand: SemanticRefactorCommand
+  ) async throws -> LSPAny {
+    guard let sourceKitLSPServer else {
+      // `SourceKitLSPServer` has been destructed. We are tearing down the
+      // language server. Nothing left to do.
+      throw ResponseError.unknown("Connection to the editor closed")
     }
-    return refactor
+
+    let semanticRefactor = try await self.refactoring(semanticRefactorCommand)
+
+    let edit = semanticRefactor.edit
+    let req = ApplyEditRequest(label: semanticRefactor.title, edit: edit)
+    let response = try await sourceKitLSPServer.sendRequestToClient(req)
+    if !response.applied {
+      let reason: String
+      if let failureReason = response.failureReason {
+        reason = " reason: \(failureReason)"
+      } else {
+        reason = ""
+      }
+      logger.error("client refused to apply edit for \(semanticRefactor.title, privacy: .public) \(reason)")
+    }
+
+    return edit.encodeToLSPAny()
   }
 }
--- a/Sources/SourceKitLSP/Swift/SwiftCommand.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftCommand.swift
@@ -15,8 +15,11 @@ import LanguageServerProtocol
 ///
 /// All commands from the Swift LSP should be listed here.
 public let builtinSwiftCommands: [String] = [
-  SemanticRefactorCommand.self
-].map { $0.identifier }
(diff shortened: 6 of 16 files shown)


Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,124 @@
...
+
+typealias RefactoringEdit = (startPosition: Position, endPosition: Position, newText: String, bufferName: String?)
+
+extension Refactoring {
+  /// Create an instance of `Refactoring` from a sourcekitd semantic refactoring response dictionary, if possible. Passes the response to `init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit])` in a neat manner
+  ///
+  /// - Parameters:
+  ///   - title: The title of the refactoring action.
+  ///   - dict: Response dictionary to extract information from.
+  ///   - url: The client URL that triggered the `semantic_refactoring` request.
```
ahoppen, line 31: I think I fixed this comment mismatch last week 😉 

```suggestion
  ///   - snapshot: The snapshot that triggered the `semantic_refactoring` request.
```

Sources/SourceKitLSP/Swift/ExpandMacroCommand.swift
```diff
@@ -0,0 +1,74 @@
...
+    self.init(
+      title: title,
+      actionString: actionString,
+      positionRange: positionRange,
+      textDocument: textDocument
+    )
+  }
+
+  public init(title: String, actionString: String, positionRange: Range<Position>, textDocument: TextDocumentIdentifier)
+  {
```
ahoppen, line 59: Nitpick, I prefer to break this on the parameters.

```suggestion
  public init(
    title: String, 
    actionString: String, 
    positionRange: Range<Position>, 
    textDocument: TextDocumentIdentifier
  ) {
```

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,124 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SourceKitD
+
+protocol Refactoring {
```
ahoppen, line 16: Would `RefactoringResult` or `RefactoringResponse` make more sense as the name here? `Refactoring` sounds more like what you implemented as `RefactorCommand` to me.

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,124 @@
...
+  /// Provides detailed information about the result of a specific refactoring operation.
+  ///
+  /// Wraps the information returned by sourcekitd's `semantic_refactoring` request, such as the necessary edits and placeholder locations.
+  ///
+  /// - Parameters:
+  ///   - command: The semantic `RefactorCommand` that triggered this request.
+  ///   - responseType: The response type `T.Type` of the particular command
+  /// - Returns:
+  ///   - The `Refactoring` as `T`
+  func refactoring<T: Refactoring>(_ refactorCommand: any RefactorCommand, _ responseType: T.Type) async throws -> T {
```
ahoppen, line 95: Instead of explicitly passing in the response type here, what do you think about adding a 

```swift
associatedtype Result: Refactoring
```

to `RefactorCommand` and then changing the function definition to


```suggestion
  func refactoring<T: RefactoringCommand>(_ refactorCommand: T, _ responseType: T.Result) async throws -> T.Result {
```

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -1,3 +1,4 @@
+import LSPLogging
```
ahoppen, line 1: Could you move the import after the copyright header?

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -193,6 +193,11 @@ struct SourceKitLSP: AsyncParsableCommand {
   )
   var generatedInterfacesPath = defaultDirectoryForGeneratedInterfaces
 
+  @Option(
+    help: "Specify the directory where generated macro expansions will be stored"
+  )
+  var generatedMacroExpansionsPath = defaultDirectoryForGeneratedMacroExpansions
+
```
ahoppen, line 200: Instead of adding options for every new kind of generated file directories, what do you think about generalizing `generatedInterfacesPath` to `generatedFilesPath` and then placing the generated interfaces and the macro expansions in those subdirectories?

Sources/SourceKitLSP/Swift/SwiftCommand.swift
```diff
@@ -15,15 +15,20 @@ import LanguageServerProtocol
...
+}
 
 /// A `Command` that should be executed by Swift's language server.
 public protocol SwiftCommand: Codable, Hashable, LSPAnyCodable {
   static var identifier: String { get }
   var title: String { get set }
 }
 
+public typealias SwiftCommandHandler = (any SwiftCommand) async throws -> LSPAny
+
```
ahoppen, line 31: I think we only use this once so I’d just use the function type inline. Just means that you don’t need to look up what `SwiftCommandHandler`. Just means that you don’t need to look up what `SwiftCommandHandler ` is when reading the definition of `commandsAndHandlers`.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -797,15 +802,27 @@ extension SwiftLanguageService {
...
         let lspCommand = try $0.asCommand()
+        canInlineMacro = $0.title == "Inline Macro"
         return CodeAction(title: $0.title, kind: .refactor, command: lspCommand)
       } catch {
         logger.log("Failed to convert SwiftCommand to Command type: \(error.forLogging)")
         return nil
       }
     }
+
+    if (canInlineMacro) {
```
ahoppen, line 818:
```suggestion
    if canInlineMacro {
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -797,15 +802,27 @@ extension SwiftLanguageService {
       additionalParameters: additionalCursorInfoParameters
     )
 
-    return cursorInfoResponse.refactorActions.compactMap {
+    var canInlineMacro = false
+
+    var refactorActions = cursorInfoResponse.refactorActions.compactMap {
       do {
         let lspCommand = try $0.asCommand()
+        canInlineMacro = $0.title == "Inline Macro"
```
ahoppen, line 810: Instead of checking the macro title, what do you think about comparing the `actionString`. That should be more stable than the title.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -797,15 +802,27 @@ extension SwiftLanguageService {
...
         return CodeAction(title: $0.title, kind: .refactor, command: lspCommand)
       } catch {
         logger.log("Failed to convert SwiftCommand to Command type: \(error.forLogging)")
         return nil
       }
     }
+
+    if (canInlineMacro) {
+      let expandMacroCommand = try! ExpandMacroCommand(positionRange: params.range, textDocument: params.textDocument)
```
ahoppen, line 819: We shouldn’t use `try!` because we don’t want to crash the entire SourceKit-LSP server just because some request failed. Checking really quickly, I think we can just make `asCommand` not throw. If that’s not possible, we should use `orLog` to log the error and then just not add the code action.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -922,29 +939,19 @@ extension SwiftLanguageService {
...
-    let edit = refactor.edit
-    let req = ApplyEditRequest(label: refactor.title, edit: edit)
-    let response = try await sourceKitLSPServer.sendRequestToClient(req)
-    if !response.applied {
-      let reason: String
-      if let failureReason = response.failureReason {
-        reason = " reason: \(failureReason)"
-      } else {
-        reason = ""
+
```
ahoppen, line 942: Superfluous newline.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -922,29 +939,19 @@ extension SwiftLanguageService {
...
+    for (command, handler) in commandsAndHandlers {
+      if let swiftCommand = req.swiftCommand(ofType: command) {
+        return try await handler(swiftCommand)
       }
-      logger.error("client refused to apply edit for \(refactor.title, privacy: .public)!\(reason)")
     }
-    return edit.encodeToLSPAny()
+
+    throw ResponseError.unknown("semantic refactoring: unknown command \(req.command)")
```
ahoppen, line 954: `ExecuteCommand` no longer is just about semantic refactorings, right? I’d just drop the `semantic refactoring: ` prefix from the log message.

Sources/SourceKitLSP/Swift/MacroExpansionEdit.swift
```diff
@@ -0,0 +1,57 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SourceKitD
+
+/// Represents a macro expansion as an edit. Notionally, a subclass of `TextEdit`
+public struct MacroExpansionEdit: ResponseType, Hashable, Sendable {
```
ahoppen, line 17: Does this type have to be `public`? Also, I don’t think it has to conform to `ResponseType` because we never return it as a response to an LSP request – we always encode it to `LSPAny` first.

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -103,41 +58,44 @@ extension SemanticRefactoringError: CustomStringConvertible {
...
-      // LSP is zero based, but this request is 1 based.
-      keys.line: line + 1,
-      keys.column: utf8Column + 1,
-      keys.length: snapshot.utf8OffsetRange(of: refactorCommand.positionRange).count,
-      keys.actionUID: self.sourcekitd.api.uid_get_from_cstr(refactorCommand.actionString)!,
-      keys.compilerArgs: await self.buildSettings(for: snapshot.uri)?.compilerArgs as [SKDRequestValue]?,
-    ])
+    guard let refactorCommand = refactorCommand as? SemanticRefactorCommand else {
+      throw ResponseError.unknown("refactorCommand is not a SemanticRefactorCommand")
+    }
```
ahoppen, line 82: Similar to my comment on the macro expansion, should we make this function take a `SemanticRefactorCommand` instead of `any SwiftCommand`?

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -922,29 +939,19 @@ extension SwiftLanguageService {
...
-      let reason: String
-      if let failureReason = response.failureReason {
-        reason = " reason: \(failureReason)"
-      } else {
-        reason = ""
+
+    let commandsAndHandlers: [(command: any SwiftCommand.Type, handler: SwiftCommandHandler)] = [
+      (command: SemanticRefactorCommand.self, handler: semanticRefactoring),
+      (command: ExpandMacroCommand.self, handler: expandMacro),
+    ]
```
ahoppen, line 946: Instead of looping here, do you think it would be simpler to just have a quick `if` series? That way we could change `semantRefactoring` and `expandMacro` to take concrete types instead of `any SwiftCommand`

Something like

```swift
if let command = req.swiftCommand(ofType: SemanticRefactorCommand.self) {
  return try await semanticRefactoring(command)
} else if let command = req.swiftCommand(ofType: ExpandMacroCommand.self) {
  return try await expandMacro(command)
}
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,94 @@
...
+
+import LSPLogging
+import LanguageServerProtocol
+import SourceKitD
+
+/// Detailed information about the result of a macro expansion operation.
+///
+/// Wraps the information returned by sourcekitd's `semantic_refactoring` request, such as the necessary macro expansion edits.
+struct MacroExpansion: Refactoring {
+
```
ahoppen, line 21: Superfluous newline

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,94 @@
...
+    self.uri = uri
+    self.edits = refactoringEdits.map { refactoringEdit in
+      MacroExpansionEdit(
+        range: refactoringEdit.startPosition..<refactoringEdit.endPosition,
+        newText: refactoringEdit.newText,
+        bufferName: refactoringEdit.bufferName ?? ""  // TODO: handle bufferName if nil
+      )
+    }
+  }
+
```
ahoppen, line 42: Superfluous newline

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,94 @@
...
+  var edits: [MacroExpansionEdit]
+
+  init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit]) {
+    self.title = title
+    self.uri = uri
+    self.edits = refactoringEdits.map { refactoringEdit in
+      MacroExpansionEdit(
+        range: refactoringEdit.startPosition..<refactoringEdit.endPosition,
+        newText: refactoringEdit.newText,
+        bufferName: refactoringEdit.bufferName ?? ""  // TODO: handle bufferName if nil
```
ahoppen, line 38: I would just log a fault and then ignore that edit.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,94 @@
...
+        try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: String.Encoding.utf8)
+      } catch {
+        throw ResponseError.unknown("Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\"")
+      }
+
+      let req = ShowDocumentRequest(uri: macroExpansionDocURI, selection: macroEdit.range)
+      let response = try await sourceKitLSPServer.sendRequestToClient(req)
+      if !response.success {
+        logger.error("client refused to show document for \(expansion.title, privacy: .public)!")
+      }
```
ahoppen, line 89: I would prefer to send this request to the client asynchronously (ie. wrap it in a `Task`) because otherwise it’s an easy way to deadlock SourceKit-LSP in the future. I don’t think it can happen right now but depending on how `MessageHandlingDependencyTracker` gets changed in the future it’s not completely unreasonable to assume that as part of the `ShowDocumentRequest` to the client, the client will send a request to SourceKit-LSP that cannot be executed until the `ExecuteCommandRequest` finishes executing. But that can’t finish executing until the `ShowDocumentRequest` has finished, so deadlock.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+            return "(\(argument), \(literal: argument.description))"
+          }
+        }
+
+        @main
+        struct MyMacroPlugin: CompilerPlugin {
+            let providingMacros: [Macro.Type] = [
+                StringifyMacro.self,
+            ]
+        }
```
ahoppen, line 168: Nitpick: I would indent this using 2 spaces for consistency.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+        "MyMacroClient/MyMacroClient.swift": """
+        @freestanding(expression)
+        public macro stringify<T>(_ value: T) -> (T, String) = #externalMacro(module: "MyMacros", type: "StringifyMacro")
+
+        func test() {
+           1️⃣#stringify2️⃣(1 + 2)
+        }
+        """,
+      ],
+      manifest: SwiftPMTestProject.macroPackageManifest
```
ahoppen, line 179: You can remove the other test case that uses `SwiftPMTestProject.macroPackageManifest` now. That test was basically a no-op that was just there to make sure that `SwiftPMTestProject.macroPackageManifest` works correctly. And we have this test to do that now.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+    )
+    try await SwiftPMTestProject.build(at: project.scratchDirectory)
+
+    let (uri, positions) = try project.openDocument("MyMacroClient.swift")
+
+    let args = ExpandMacroCommand(
+      positionRange: positions["1️⃣"]..<positions["2️⃣"],
+      textDocument: TextDocumentIdentifier(uri)
+    )
```
ahoppen, line 188: I think it would make sense to also check that we show macro expansions at the zero character ranges (aka positions) at `1️⃣#stringify` and `#1️⃣stringify`.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+
+    let result = try await project.testClient.send(request)
+
+    guard let resultArray: [MacroExpansionEdit] = Array(fromLSPArray: result ?? .null) else {
+      XCTFail("Result is not an array.")
+      return
+    }
+
+    XCTAssertEqual(resultArray.count, 1)
+    XCTAssertEqual(resultArray[0].newText, "(1 + 2, \"1 + 2\")")
```
ahoppen, line 209: We have `.only` for exactly this pattern 😉 This also prevents the test from crashing if `resultArray` is empty.

```suggestion
    XCTAssertEqual(resultArray.only?.newText, "(1 + 2, \"1 + 2\")")
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+
+    let generatedMacroExpansionsPath = SourceKitLSPServer.Options.testDefault.generatedMacroExpansionsPath
+
+    XCTAssert(
+      (try? generatedMacroExpansionsPath.appending(component: resultArray[0].bufferName).asURL
+        .checkResourceIsReachable()) ?? false
+    )
+
```
ahoppen, line 217: Instead of inspecting the internals of the macro expansion here, what do you think about the following (which would also be closer to what editors do):
- In the handler for the `ShowDocumentRequest` request, we record the URL of the generated buffer
- We use an XCTest expectation to make sure we have received the `ShowDocumentRequest` here.
- And then we inspect the contents of that file.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,95 @@ final class ExecuteCommandTests: XCTestCase {
...
+    let fileContents = try XCTUnwrap(
+      try? String(
+        contentsOf: generatedMacroExpansionsPath.appending(component: resultArray[0].bufferName).asURL,
+        encoding: .utf8
+      )
+    )
+
+    XCTAssertTrue(fileContents.contains("(1 + 2, \"1 + 2\")"))
```
ahoppen, line 225: Nit
```suggestion
    XCTAssert(fileContents.contains("(1 + 2, \"1 + 2\")"))
```

ahoppen (commented):
Very impressive. This is looking great. All my comments are minor / code style related. The overall design looks great!

yaroslavyaroslav:
Sorry to disturbing you folks in such a busy time, but just to clarify for myself does this implementation would add `ExpandSnippet` code action option in along with `ExtractMethod` on a client side for free or there's be some work on their side required?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: Yeah that makes sense

Sources/SourceKitLSP/Swift/MacroExpansionEdit.swift
lokesh-tr: Yeah right

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: Cool

Sources/SourceKitLSP/Swift/MacroExpansionEdit.swift
lokesh-tr: Wait, I realised that the tests are using the `MacroExpansionEdit`. So, it has to be `public`? or should we not mark this `public` and try some workarounds in the tests?

lokesh-tr:
> Sorry to disturbing you folks in such a busy time, but just to clarify for myself does this implementation would add `ExpandSnippet` code action option in along with `ExtractMethod` on a client side for free or there's be some work on their side required?

@yaroslavyaroslav well I don't think this implementation would add anything related to what you have asked. There's this `ExecuteCommandTests.testRangeSemanticRefactoring` which tests whether extract method works (already implemented, not specific to this PR). But am not sure if the same is implemented on the client side and am not sure sourcekitd sends an `ExpandSnippet` code action to the LSP.

Perhaps @ahoppen can answer?

@yaroslavyaroslav I guess you can open an issue and discuss this with him to not spam this PR since its not related to this PR.

Sources/SourceKitLSP/Swift/MacroExpansionEdit.swift
ahoppen: Ah, I missed that. Let’s mark it as `@_spi(Testing) public` then and `@_spi(Testing) import SourceKitLSP` in the test.

ahoppen:
Sorry, I don’t fully understand your question @yaroslavyaroslav. If we want to continue discussing it, could you please file an issue so we can have a dedicated thread for it?

lokesh-tr:
Based on the meeting which we had yesterday:

- [ ] Address Review Comments

- [ ] Wrap it in experimental feature

- [ ] Rename generated macro file name to have line numbers

- [ ] Implement custom peek requests

lokesh-tr:
I was little skeptical with my implementation of `Refactoring` (thought I would replace it with a better name at the end but forgot) and the passing around `T.Type`. I did think about using associated typesbut that didn't fit well with my array-looping based implementation of `executeCommandRequest`. Now that I changed it to if-elseif, the associate type implementation is trivial.

I have resolved all your comments. Kindly review and let me know.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,104 @@ final class ExecuteCommandTests: XCTestCase {
...
+
+      await fulfillment(of: [expectation], timeout: 10.0)
+
+      let url = try XCTUnwrap(showDocumentRequestURI?.fileURL)
+
+      XCTAssert(
+        (try? url.checkResourceIsReachable()) ?? false
+      )
+
+      let fileContents = try XCTUnwrap(try? String(contentsOf: url, encoding: .utf8))
```
ahoppen, line 231: I think you can save the `XCTUnwrap` dance here and just throw if it fails. The benefit would be that this also shows the error why the string couldn’t load the file contents when it fails.

```suggestion
      let fileContents = try String(contentsOf: url, encoding: .utf8)
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +136,104 @@ final class ExecuteCommandTests: XCTestCase {
...
+
+    let argsToBeTested = [
+      ExpandMacroCommand(
+        positionRange: positions["1️⃣"]..<positions["3️⃣"],
+        textDocument: TextDocumentIdentifier(uri)
+      ),
+      ExpandMacroCommand(
+        positionRange: positions["2️⃣"]..<positions["3️⃣"],
+        textDocument: TextDocumentIdentifier(uri)
+      ),
```
ahoppen, line 193: Could you also add tests for `1️⃣..<1️⃣` and `2️⃣..<2️⃣`? That should cover cases when you are invoking the command from a cursor without selecting text.

ahoppen (commented):
Looking great. I have a few minor comments still but everything is very local.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,98 @@
...
+        try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: String.Encoding.utf8)
+      } catch {
+        throw ResponseError.unknown("Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\"")
+      }
+
+      Task {
+        let req = ShowDocumentRequest(uri: macroExpansionDocURI, selection: macroEdit.range)
+        let response = try await sourceKitLSPServer.sendRequestToClient(req)
+        if !response.success {
+          logger.error("client refused to show document for \(expansion.title, privacy: .public)!")
```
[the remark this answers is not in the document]
lokesh-tr: I don't think that the `ShowDocumentResponse` as per the LSP spec has a property for the reason behind the error. Am I missing anything? 🤔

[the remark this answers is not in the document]
ahoppen: Oh, sorry, I was mistaking this with the client returning an error response. And looking at it again, I notice that if we get an error from `sendRequestToClient`, we just swallow it. We should log that by wrapping `sendRequestToClient` in `orLog`.

lokesh-tr:
@ahoppen I have addressed the review comments. When possible, let me know if there is anything else which I should fix. 😌 

I have also made a new commit.
Instead of creating generated files with buffer name in `GeneratedMacroExpansions` directory, this will now create a new subdirectory for each buffer within the `GeneratedMacroExpansions`. Within that subdirectory, will exist a file with the name as original source file containing the macro expansion.

This is inline with what we discussed in the meeting of adding a GitHub style name notation. The only difference is that instead of having just the starting line number in the file name like `SourceFile.swift#L3`, this implementation adds position range to the file name as `SourceFile.swift#L3C2-L4C5`. This is based on my finding that GitHub supports a url fragment which denotes the starting and ending position separated with an hyphen, each denoted by the letters L and C representing Line number and Column Number respectively.

The start and end position are equal when new lines are added while expanding. Similarly, they are unequal when new lines are replacing old lines while expanding. And thus, allows the user to just look at the name of the expanded file name to decide whether it is adding or replacing new lines.

lokesh-tr:
Based on the discussion about my findings which we had,
I changed the file naming scheme as follows: `SourceFileName_LaCb-LcCd.swift`

The benefit of the above naming is that:
1. The user can be clear about whether the macro is adding new lines or replacing with new lines just by looking at the position range.
2. No use of `#` means, no need to worry about fragments, escaping the sign, etc.
3. Instead of putting the position range indicator after the file extension, we put it before the extension, so that the code editors can understand the actual file extension.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+
+  /// The resulting array of `MacroExpansionEdit` of a semantic refactoring request
+  var edits: [MacroExpansionEdit]
+
+  init(title: String, uri: DocumentURI, refactoringEdits: [RefactoringEdit]) {
+    self.title = title
+    self.uri = uri
+    self.edits = refactoringEdits.compactMap { refactoringEdit in
+      if refactoringEdit.bufferName.isEmpty {
+        logger.error("Unable to retrieve some parts of the expansion")
```
ahoppen, line 39: This gave me an excuse to write some document about when to use which logging level. I think `fault` would be more applicable here because if we don’t have a buffer name that’s a bug in sourcekitd.

https://github.com/apple/sourcekit-lsp/blob/main/Documentation/Logging.md

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+  /// and then makes `ShowDocumentRequest` to the client side for each
+  /// expansion to be displayed.
+  ///
+  /// - Parameters:
+  ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this
+  /// request.
+  ///
+  /// - Returns:
+  ///   - an `[MacroExpansionEdit]` with the necessary edits and buffer name as
+  /// a `LSPAny`
```
ahoppen, line 61: Some formatting suggestions that should make this read a little better, I think.

```suggestion
  /// - Parameters:
  ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this request.
  ///
  /// - Returns: An `[MacroExpansionEdit]` with the necessary edits and buffer name as a `LSPAny`
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+    _ expandMacroCommand: ExpandMacroCommand
+  ) async throws -> LSPAny {
+    guard let sourceKitLSPServer else {
+      // `SourceKitLSPServer` has been destructed. We are tearing down the
+      // language server. Nothing left to do.
+      throw ResponseError.unknown("Connection to the editor closed")
+    }
+
+    guard let sourceFileURL = expandMacroCommand.textDocument.uri.fileURL else {
+      throw ResponseError.unknown("Given URL is not a file URL")
```
ahoppen, line 72: Nit: LSP communicates in URIs instead of URLs

```suggestion
      throw ResponseError.unknown("Given URI is not a file URL")
```

Sources/SourceKitLSP/Swift/RefactoringEdit.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SourceKitD
+
+/// Represents an edit from semantic refactor response. Notionally, a subclass /// of `TextEdit`
```
ahoppen, line 16:
```suggestion
/// Represents an edit from semantic refactor response. Notionally, a subclass of `TextEdit`
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    let expansion = try await self.refactoring(expandMacroCommand)
+
+    for macroEdit in expansion.edits {
+      let macroExpansionBufferDirectoryName = macroEdit.bufferName.dropLast(6)  // buffer name without ".swift"
```
ahoppen, line 78: Should we check that the buffer name has a `.swift` suffix? Otherwise I can imagine that we could get very odd failure scenarios if the buffer name does not end in `.swift`.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+
+      let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath.appendingPathComponent(
+        String(macroExpansionBufferDirectoryName)
+      )
```
ahoppen, line 82: I generally prefer to put `appendingPathComponent` on a new line because it scales if you need to add more components to URLs.

```suggestion
      let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
        .appendingPathComponent(String(macroExpansionBufferDirectoryName))
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+        "L\(macroEdit.range.lowerBound.line)C\(macroEdit.range.lowerBound.utf16index)-L\(macroEdit.range.upperBound.line)C\(macroEdit.range.upperBound.utf16index)"  // github permalink notation for position range
+
+      let macroExpansionFilePath = macroExpansionBufferDirectoryURL.appendingPathComponent(
+        "\(macroExpansionFileName)_\(macroExpansionPositionRangeIndicator).\(sourceFileURL.pathExtension)"
+      )
+
+      let macroExpansionDocURI = DocumentURI(macroExpansionFilePath)
+
+      do {
+        try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: String.Encoding.utf8)
```
ahoppen, line 102: You can drop the `String.Encoding`.

```suggestion
        try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: .utf8)
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,122 @@
...
+      } catch {
+        throw ResponseError.unknown("Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\"")
+      }
+
+      Task {
+        let req = ShowDocumentRequest(uri: macroExpansionDocURI, selection: macroEdit.range)
```
ahoppen, line 108: I think this is the only place that uses `macroExpansionDocURI`, so you could just change it to the following and remove `macroExpansionDocURI`.

```suggestion
        let req = ShowDocumentRequest(uri: DocumentURI(macroExpansionFilePath), selection: macroEdit.range)
```

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,130 @@
...
+  /// refactoring response dictionary, if possible.
+  ///
+  /// - Parameters:
+  ///   - title: The title of the refactoring action.
+  ///   - dict: Response dictionary to extract information from.
+  ///   - snapshot: The snapshot that triggered the `semantic_refactoring` request.
+  ///   - keys: The sourcekitd key set to use for looking up into `dict`.
+  init?(_ title: String, _ dict: SKDResponseDictionary, _ snapshot: DocumentSnapshot, _ keys: sourcekitd_api_keys) {
+    guard let categorizedEdits: SKDResponseArray = dict[keys.categorizedEdits] else {
+      logger.error("categorizedEdits doesn't exist in response dictionary")
```
ahoppen, line 32: Similar to my comment above, I think this could be a `fault`.

```suggestion
      logger.fault("categorizedEdits doesn't exist in response dictionary")
```

Same for the other `error`s you are introducing in this file.

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    var refactoringEdits = [RefactoringEdit]()
+
+    categorizedEdits.forEach { _, categorizedEdit in
+      guard let edits: SKDResponseArray = categorizedEdit[keys.edits] else {
+        logger.error("edits doesn't exist in categorizedEdit dictionary")
+        return true
+      }
+      edits.forEach { _, edit in
+        // The LSP is zero based, but semantic_refactoring is one based.
```
ahoppen, line 44: I think this comment makes more sense before the lines where you do the `- 1` calculations.

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,130 @@
...
+            bufferName: edit[keys.bufferName] ?? ""
+          )
+        )
+        return true
+      }
+      return true
+    }
+
+    guard !refactoringEdits.isEmpty else {
+      logger.error("No refactors found")
```
ahoppen, line 80:
```suggestion
      logger.error("No refactoring edits found")
```

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,130 @@
...
+  /// Provides detailed information about the result of a specific refactoring
+  /// operation.
+  ///
+  /// Wraps the information returned by sourcekitd's `semantic_refactoring`
+  /// request, such as the necessary edits and placeholder locations.
+  ///
+  /// - Parameters:
+  ///   - refactorCommand: The semantic `RefactorCommand` that triggered this request.
+  /// - Returns:
+  ///   - The `RefactoringResponse` as `T`
```
ahoppen, line 98: Also, I’m not entirely sure what you mean by _The `RefactoringResponse` as `T`_. I would just go for something like: _The response of the refactoring_.

---

Nit: You don’t need a sub-bullet point for the description of the return value.

```suggestion
  /// - Returns: The `RefactoringResponse` as `T`
```

Sources/SourceKitLSP/Swift/RefactoringEdit.swift
```diff
@@ -0,0 +1,57 @@
...
+@_spi(Testing) public struct RefactoringEdit: Hashable, Sendable, Codable {
+  /// The range of text to be replaced.
+  @CustomCodable<PositionRange>
+  public var range: Range<Position>
+
+  /// The new text.
+  public var newText: String
+
+  public var bufferName: String
```
ahoppen, line 25: Should we add a doc comment for `bufferName` because I think it’s not entirely obvious what it is? Something like 

> If the new text of the edit should not be applied to the original source file but to a separate buffer, a fake name for that buffer. For example for expansion of macros, this is `@` followed by the mangled name of the macro expansion, followed by `.swift`.

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -103,41 +60,42 @@ extension SemanticRefactoringError: CustomStringConvertible {
...
+  /// Makes request to sourcekitd and wraps the result into a
+  /// `SemanticRefactoring` and then makes an `ApplyEditRequest` to the client
+  /// side for the actual refactoring.
   ///
   /// - Parameters:
-  ///   - refactorCommand: The semantic refactor `Command` that triggered this request.
+  ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
+  ///
+  /// - Returns:
+  ///   - a `WorkspaceEdit` with the necessary refactors as a `LSPAny`
```
ahoppen, line 74:
```suggestion
  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +137,123 @@ final class ExecuteCommandTests: XCTestCase {
...
+      )
+
+      let metadata = SourceKitLSPCommandMetadata(textDocument: TextDocumentIdentifier(uri))
+
+      var command = args.asCommand()
+      command.arguments?.append(metadata.encodeToLSPAny())
+
+      let request = ExecuteCommandRequest(command: command.command, arguments: command.arguments)
+
+      let expectation = XCTestExpectation(description: "Handle Show Document Request")
```
ahoppen, line 206: `self.expectation` also raises in error if the expectation is never awaited, which is quite useful.

```suggestion
      let expectation = self.expectation(description: "Handle Show Document Request")
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +137,123 @@ final class ExecuteCommandTests: XCTestCase {
...
+        1,
+        "resultArray is empty. Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+      )
+      XCTAssertEqual(
+        resultArray.only?.newText,
+        "(1 + 2, \"1 + 2\")",
+        "Wrong macro expansion. Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+      )
+
+      await fulfillment(of: [expectation], timeout: 10.0)
```
ahoppen, line 235: `fulfillmentOfOrThrow` is a useful utility that uses `defaultTimeout` and throws an error if it times out, which is useful because it’s very likely that all the following assertions will fail if the expectation times out.

```suggestion
      try await fulfillmentOfOrThrow([expectation])
```

ahoppen (commented):
I have a few more minor comments inline.

lokesh-tr:
The last commit which I made tries to convert the sequential edit from sourcekitd into `ConcurrentEdits` as we discussed, but it fails the `testFreestandingMacroExpansion` test, the reason being the line numbers aren't matching after i convert them to concurrent edits.

I would like to have you help me debug this issue.

lokesh-tr:
I removed the `ConcurrentEdits` commit from here... I will wrap the current state of freestanding macro expansion as such in  experimental features and will make a seperate PR for concurrent edits and attached macro expansion

lokesh-tr:
@ahoppen I wrapped the feature under `ExperimentalFeatures`. Let me have your comments so that I can fix them by tomorrow and also squash the commits. Hope we will merge this soon.

Further, I will follow up with a new PR for the `ConcurrentEdits`.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -0,0 +1,118 @@
...
+          try await sourceKitLSPServer.sendRequestToClient(req)
+        }
+
+        if let response, !response.success {
+          logger.error("client refused to show document for \(expansion.title, privacy: .public)")
+        }
+      }
+    }
+
+    return expansion.edits.encodeToLSPAny()
```
ahoppen, line 116: Do we even need to return any edits here? I would expect that we have already handled the showing of the macro expansion buffers and the client shouldn’t need to deal with the edits returned from the `ExecuteCommandRequest`, right? Also, since it’s `LSPAny`, I would expect that the client doesn’t even know how to interpret it. Or am I missing something?

Edit: I now see that the tests rely on `ExecuteCommand` returning the edits. If we don’t need `expandMacro` and `semanticRefactoring` to return any results for editors, I think we should change this in a follow-up PR to get this one merged.

Sources/SourceKitLSP/Swift/Refactoring.swift
```diff
@@ -0,0 +1,129 @@
...
+        )
+        // Snippets are only supported in code completion.
+        // Remove SourceKit placeholders in refactoring actions because they
+        // can't be represented in the editor properly.
+        let textWithSnippets = rewriteSourceKitPlaceholders(in: text, clientSupportsSnippets: false)
+        refactoringEdits.append(
+          RefactoringEdit(
+            range: startPosition..<endPosition,
+            newText: textWithSnippets,
+            bufferName: edit[keys.bufferName] ?? ""
```
ahoppen, line 71: Do you think a cleaner design would be to make `RefactoringEdit.bufferName` an optional? Then we wouldn’t need to nil-coalesce here and I think `nil` represents the absence of a buffer name better than an empty string.

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -103,41 +60,41 @@ extension SemanticRefactoringError: CustomStringConvertible {
 }
 
 extension SwiftLanguageService {
-  /// Provides detailed information about the result of a specific refactoring operation.
+
+  /// Handles the `SemanticRefactorCommand`.
   ///
-  /// Wraps the information returned by sourcekitd's `semantic_refactoring` request, such as the necessary edits and placeholder locations.
+  /// Makes request to sourcekitd and wraps the result into a
```
ahoppen, line 66:
```suggestion
  /// Sends a request to sourcekitd and wraps the result into a
```

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -103,41 +60,41 @@ extension SemanticRefactoringError: CustomStringConvertible {
...
+  /// `SemanticRefactoring` and then makes an `ApplyEditRequest` to the client
+  /// side for the actual refactoring.
   ///
   /// - Parameters:
-  ///   - refactorCommand: The semantic refactor `Command` that triggered this request.
+  ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
+  ///
+  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
```
ahoppen, line 73: I think this comment is out-of-date, right? This function doesn’t return a `WorkspaceEdit`. It actually got me wondering: Do we need to return anything here at all? Similar to my comment on `macroExpansion`, I wouldn’t expect the client to interpret the `LSPAny` and we already handled the edits by sending `ApplyEditRequest` to the client.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -815,15 +831,31 @@ extension SwiftLanguageService {
...
-        let lspCommand = try $0.asCommand()
-        return CodeAction(title: $0.title, kind: .refactor, command: lspCommand)
-      } catch {
-        logger.log("Failed to convert SwiftCommand to Command type: \(error.forLogging)")
-        return nil
+    var canInlineMacro = false
+
+    let showMacroExpansionsIsEnabled = await self.sourceKitLSPServer?.options.experimentalFeatures.contains(
+      .showMacroExpansions
+    )
```
ahoppen, line 838: Formatting nitpick again 😉 Also, I would just nil-coalesce here so we don’t need to deal with `showMacroExpansionsIsEnabled == nil` below

```suggestion
    let showMacroExpansionsIsEnabled = await self.sourceKitLSPServer?.options.experimentalFeatures
      .contains(.showMacroExpansions) ?? false
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -815,15 +831,31 @@ extension SwiftLanguageService {
...
+
+    var refactorActions = cursorInfoResponse.refactorActions.compactMap {
+      let lspCommand = $0.asCommand()
+      canInlineMacro = $0.actionString == "source.refactoring.kind.inline.macro"
+
+      if !(showMacroExpansionsIsEnabled != nil && showMacroExpansionsIsEnabled!) {
+        canInlineMacro = false
       }
```
ahoppen, line 846: Do you think this would read cleaner as follows? I found the setting of `canInlineMacro` to `true` just to revert it back to `false` a little confusing. I also noticed that we would set `canInlineMacro` back to `false` if `source.refactoring.kind.inline.macro` was not the last refactoring action.

```suggestion
      if !canInlineMacro, showMacroExpansionsIsEnabled {
        canInlineMacro = $0.actionString == "source.refactoring.kind.inline.macro"
      }
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -136,6 +137,124 @@ final class ExecuteCommandTests: XCTestCase {
...
+        @freestanding(expression)
+        public macro stringify<T>(_ value: T) -> (T, String) = #externalMacro(module: "MyMacros", type: "StringifyMacro")
+
+        func test() {
+          1️⃣#2️⃣stringify3️⃣(1 + 2)
+        }
+        """,
+      ],
+      manifest: SwiftPMTestProject.macroPackageManifest,
+      serverOptions: .init(swiftPublishDiagnosticsDebounceDuration: 0, experimentalFeatures: [.showMacroExpansions])
```
ahoppen, line 181: I would prefer to build on top of `.testDefault` so that if we change `.testDefault`, the changes get applied here as well. The current pattern is that at the top of the test function we do something like

```swift
var serverOptions = SourceKitLSPServer.Options.testDefault
serverOptions.experimentalFeatures.insert(.showMacroExpansions)
```

ahoppen (commented):
I only have very minor comments now. Could you address them and then we can get the PR merged. My comment about making `expandMacro` and `semanticRefactoring` to not return the edits is more wide-reaching and I think we should look into that in a follow-up PR to get this one merged.

Sources/SourceKitLSP/Swift/Refactoring.swift
lokesh-tr: Sure, and I would also highlight the fact that it is ok for bufferName to be nil (i.e. not passed as a response from sourcekitd) if the refactoring edit is meant to represent the removal of the macro annotation from the original source file (here, newText will be empty). Its only a fault of sourcekitd if the refactoring edit's buffer name is nil when the refactoring edit represents an expansion or replacement and has an associated newText value which won't be empty.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: The way the semantic refactoring for inline macro command was implemented is such that the testing part utilises the result. Since it was already accepted into the codebase, I thought it was a convention and implemented macro expansion command in a similar manner.

I guess it would be better to merge this PR and make a fix for this is in a follow-up PR as you said.

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
lokesh-tr: Here, I would like to highlight the reply which I made to your comment on `macroExpansion`

Sources/SourceKitLSP/Swift/MacroExpansion.swift
ahoppen: Filed https://github.com/apple/sourcekit-lsp/issues/1498 so we don’t forget about it.

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
ahoppen: Just to double-check: You mean [this comment](https://github.com/apple/sourcekit-lsp/pull/1436#discussion_r1647371565), right? Saying that we should change this to not return the edits, at which point the comment will go away, right?

ahoppen (approved):
Looking great! Let’s get this 🚢ed.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/ExpandMacroCommand.swift
ahoppen, line 1: On Windows, we build SourceKit-LSP using CMake instead of SwiftPM. For this, these new files need to be added to [CMakeLists.txt](https://github.com/apple/sourcekit-lsp/blob/main/Sources/SourceKitLSP/CMakeLists.txt)

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1437, opened 2024-06-06
Use 3 emoji markers in the index log

ahoppen:
I found that 2 markers sometimes chooses very close color combinations like red-yellow and red-orange. Let’s add a third emoji marker to make it even easier to tell the log lanes apart.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1439, opened 2024-06-06
Add line and column to test item id

### LINKED ISSUE #1152: Include filename, line and column in test IDs
The test IDs returned by https://github.com/apple/sourcekit-lsp/pull/1151 don’t qualify the tests with the filename, line and column. Because of this, the test IDs may be ambiguous, eg. when multiple files define a private `@Test` function with the same name.

We should include the filename, line and column in the test ID as `MyModule.S/f()/Tests.swift:2:3`. 

Note that when trying to test this, the `--filter` argument takes a regex, so the parentheses need to be escaped (and technically  `.`)

kimdv:
Fixes #1152

### DIFF
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -27,18 +27,68 @@ package enum LanguageServerState {
 }
 
 package struct AnnotatedTestItem: Sendable {
-  /// The test item to be annotated
-  package var testItem: TestItem
+  /// Identifier for the `AnnotatedTestItem`.
+  ///
+  /// This identifier uniquely identifies the test case or test suite. It can be used to run an individual test (suite).
+  public var id: String
+
+  /// Display name describing the test.
+  public var label: String
+
+  /// Optional description that appears next to the label.
+  public var description: String?
+
+  /// A string that should be used when comparing this item with other items.
+  ///
+  /// When `nil` the `label` is used.
+  public var sortText: String?
+
+  /// Whether the test is disabled.
+  public var disabled: Bool
+
+  /// The type of test, eg. the testing framework that was used to declare the test.
+  public var style: String
+
+  /// The location of the test item in the source code.
+  public var location: Location
+
+  /// The children of this test item.
+  ///
+  /// For a test suite, this may contain the individual test cases or nested suites.
+  public var children: [AnnotatedTestItem]
+
+  /// Tags associated with this test item.
+  public var tags: [TestTag]
 
   /// Whether the `TestItem` is an extension.
   package var isExtension: Bool
 
+  public var ambiguousTestDifferentiator: String
+
   package init(
-    testItem: TestItem,
-    isExtension: Bool
+    id: String,
+    label: String,
+    description: String? = nil,
+    sortText: String? = nil,
+    disabled: Bool,
+    style: String,
+    location: Location,
+    children: [AnnotatedTestItem],
+    tags: [TestTag],
+    isExtension: Bool,
+    ambiguousTestDifferentiator: String
   ) {
-    self.testItem = testItem
+    self.id = id
+    self.label = label
+    self.description = description
+    self.sortText = sortText
+    self.disabled = disabled
+    self.style = style
+    self.location = location
+    self.children = children
+    self.tags = tags
     self.isExtension = isExtension
+    self.ambiguousTestDifferentiator = ambiguousTestDifferentiator
   }
 }
 
--- a/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
@@ -251,10 +251,13 @@ final class SyntacticSwiftTestingTestScanner: SyntaxVisitor {
     let range = snapshot.absolutePositionRange(
       of: node.positionAfterSkippingLeadingTrivia..<node.endPositionBeforeTrailingTrivia
     )
+
+    let location = location(range: range)
+
     // Members won't be extensions since extensions will only be at the top level.
     let testItem = AnnotatedTestItem(
       testItem: TestItem(
-        id: (parentTypeNames + typeNames).joined(separator: "/"),
+        id: (parentTypeNames + typeNames + [location]).joined(separator: "/"),
         label: attributeData?.displayName ?? typeNames.last!,
         disabled: (attributeData?.isDisabled ?? false) || allTestsDisabled,
         style: TestStyle.swiftTesting,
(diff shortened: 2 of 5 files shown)


Tests/SourceKitLSPTests/DocumentTestDiscoveryTests.swift
```diff
@@ -210,14 +210,14 @@ final class DocumentTestDiscoveryTests: XCTestCase {
       tests,
       [
         TestItem(
-          id: "MyTests",
+          id: "MyTests/test.swift:2:0",
```
ahoppen, line 213: I just checked and test suites aren’t file- and position-qualified. If you run `swift test --filter 'MyTests/test.swift:2:0'` in this project, it wouldn’t execute any tests.

Tests/SourceKitLSPTests/DocumentTestDiscoveryTests.swift
```diff
@@ -210,14 +210,14 @@ final class DocumentTestDiscoveryTests: XCTestCase {
...
           label: "MyTests",
           disabled: false,
           style: TestStyle.swiftTesting,
           location: Location(uri: uri, range: positions["1️⃣"]..<positions["4️⃣"]),
           children: [
             TestItem(
-              id: "MyTests/oneIsTwo()",
+              id: "MyTests/oneIsTwo()/test.swift:3:2",
```
ahoppen, line 220: The location of a test is the start of the `Test` identifier in `@Test`. You can see this because `swift test --filter 'MyTests/oneIsTwo\(\)/test.swift:3:2'` doesn’t actually run any tests in this project (or rather, if you create an equivalent project that actually compiles because it has a proper `swift-testing` dependency declaration).

ahoppen (commented):
Thank you, Kim!

ahoppen:
@grynspan Had a very interesting idea just now to reduce possible issues where the test index still has an old location for the test but it has been moved: We only need the file-line-column distinguisher if the test ID is otherwise ambiguous, eg. because you have two `@Test fileprivate func testSomething() {}` functions in different files. 

What we could do is:
- Change `AnnotatedTestItem` to essentially be a `TestItem` just with additional information. Differences to `TestItem` are
  - Has `isExtension` property
  - Children are also `AnnotatedTestItem`
  - Stores a `ambiguousTestDifferentiator: String` of the test (I’d prefer not to re-use the `location` property because it makes sense that the test’s range contains the `@` but the test location needs to be at `Test`). This would eg. be `MyFile.swift:4:2`.
- In `SyntacticTestIndex` add a new method `ambigousTestIDs()`. This iterates over all tests in the project and checks if there are any test IDs that are used twice
- When translating `AnnotatedTestItem` to `TestItem`, pass in the ambiguous test IDs. If the `id` of an `AnnotatedTestItem` is in that set of ambiguous IDs, we append the `ambiguousTestDifferentiator ` to its ID. 

This is a lot more involved than what I described in the original issue. Would you be interested in trying this, @kimdv? Also: Do you think our idea makes sense?

kimdv:
> Change `AnnotatedTestItem` to essentially be a `TestItem` just with additional information.

Just to be clear, does it mean I should remove `AnnotatedTestItem` and replace all usage with `TestItem`?

ahoppen:
I meant it the other way round: When computing the tests, everything should be `AnnotatedTestItem` and `AnnotatedTestItem` would look something like the following.

```swift
public struct TestItem: ResponseType, Equatable {
  public var id: String
  public var label: String
  public var description: String?
  public var sortText: String?
  public var disabled: Bool
  public var style: String
  public var location: Location
  public var children: [AnnotatedTestItem]
  public var tags: [TestTag]
  public var isExtension: Bool
  public var ambiguousTestDifferentiator: String
}
```

And only when actually creating the LSP response do we convert the `AnnotatedTestItem`s into `TestItem`s.

kimdv:
@ahoppen  it's been a while. Catching up. 

I added a commit [2e6780c](https://github.com/swiftlang/sourcekit-lsp/pull/1439/commits/2e6780c6f1aa90d25e6f64d030cead77cd63dc4b) is that something that you imagined? 
It's still WIP

Tests/SourceKitLSPTests/DocumentTestDiscoveryTests.swift
grynspan: Correct, because fully qualified type names are not ambiguous. Function names however can be ambiguous:

```swift
@Suite MyTests {
  @Test(arguments: 0 ... 100)
  func example(i: Int) { ... }

  @Test(arguments: ["a", "b", "c"])
  func example(i: Character) { ... }
}
```

The ID of a test function thus includes its fully-qualified containing type name (if applicable), the name of the function including labels (`example(i:)` in the above case), and a source location.

For convenience, the source location is not required to be specified with `swift test --filter`—if two tests have the same name, we'll just run 'em both in that case. And when you run `swift test list`, we don't show source locations for unambiguous test function names unless you pass `--verbose`.

grynspan:
> @grynspan Had a very interesting idea just now to reduce possible issues where the test index still has an old location for the test but it has been moved: We only need the file-line-column distinguisher if the test ID is otherwise ambiguous, eg. because you have two `@Test fileprivate func testSomething() {}` functions in different files.

Sorry, I just saw this now. That's basically how Swift Testing decides whether or not to include the source location in its output to `swift test list`. The ID will always contain the source location, but it's not necessary to include it if the function name is unambiguous because, in effect, the function name is itself a pseudo-suite for the purposes of test discovery/planning/running, so specifying a test ID without a source location component just says "run all the tests with this name" and if there's only one… you just run the one.

Does that make sense?

ahoppen:
> is that something that you imagined?

Yes, exactly. Now we just need

>  - Stores a `ambiguousTestDifferentiator: String` of the test (I’d prefer not to re-use the `location` property because it makes sense that the test’s range contains the `@` but the test location needs to be at `Test`). This would eg. be `MyFile.swift:4:2`.
> - In `SyntacticTestIndex` add a new method `ambigousTestIDs()`. This iterates over all tests in the project and checks if there are any test IDs that are used twice
> - When translating `AnnotatedTestItem` to `TestItem`, pass in the ambiguous test IDs. If the `id` of an `AnnotatedTestItem` is in that set of ambiguous IDs, we append the `ambiguousTestDifferentiator ` to its ID.

Tests/SourceKitLSPTests/DocumentTestDiscoveryTests.swift
grynspan: Well, as of https://github.com/apple/swift-testing/pull/589 anyway.

kimdv:
@ahoppen so far so good. 

> Stores a ambiguousTestDifferentiator: String of the test (I’d prefer not to re-use the location property because it makes sense that the test’s range contains the @ but the test location needs to be at Test). This would eg. be MyFile.swift:4:2.

If we not can use the location how do we then get the `4:2` etc? 
Maybe I'm miss understadning something? 😅

ahoppen:
I was saying that the `ambiguousTestDifferentiator` should be created by whoever calls the initializer of `AnnotatedTestItem` and that caller should decide what the differentiator is and we shouldn’t implicitly re-use `AnnotatedTestItem.location` for that.

plemarquand:
@kimdv I came across this issue independently while writing some sourcekit-lsp tests and I [sketched out a PR](https://github.com/swiftlang/sourcekit-lsp/pull/1662). Unfortunately I didn't see your work first! Would you be open to coauthoring this if I put my tests in to this PR? 

I think we could also remove the need to unwrap the testItem from the AnnotatedTestItem and just [modify the TestItem's ID directly](https://github.com/swiftlang/sourcekit-lsp/pull/1662/files#diff-fdb0d2e4e192a73ab75b193c3a9cbadda5365e471d2aa135cd346c825d7f1abdR497). @ahoppen I think this simplifies things quite a bit, what do you think?

PR #1442, opened 2024-06-06
Fix a couple issue in the `debug reduce-swift-frontend` subcommand

ahoppen:
- Strip the `-index-system-modules` flag because it doesn’t work if no output file map is passed
- Remove `-primary-file` when we’re merging all `.swift` files into one file
- Fix an issue where errors during reduction would be rendered within the progress bar.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test macOS

PR #1445, opened 2024-06-06
Inspect the build setup to determine the extra flags to pass to the `swift build` invocations during preparation

ahoppen:
The `BuildParameters` that we were previously inspecting contained some parameters that SwiftPM synthesizes and that weren’t specified on the command line. For example, it contains `-g` for all languages. This causes us to add `-Xswiftc -g -Xcc -g -Xcxx -g` to the `swift build` invocations for preparation, which is not necessary.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1447, opened 2024-06-07
Change all variables that are `Atomic*` types to not be `nonisolated(unsafe)`

ahoppen:
Since the `Atomic*` types can not be marked as `Sendable` (because they aren’t C structs), we can change the variables to constants and can remove `nonisolated(unsafe)`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1453, opened 2024-06-07
Add test to check that we get semantic functionality even if a dependency module fails to prepare

ahoppen:
rdar://128683404


ahoppen:
@swift-ci Please test

Issue #1155, opened 2024-03-30
Navigation in generated .swiftinterface files


akaralar:
Is there a way to make the LSP work with `.swiftinterface` files? For instance when I go to definition of a `SwiftUI` symbol in my project, say `Text`, I am directed to a generated `SwiftUI.swiftinterface` file located in a predefined path which can be seen by running 
```
xcrun --toolchain swift sourcekit-lsp --help
```

Later in this `SwiftUI.swiftinterface` file, if I try to go to definition of a symbol in another `Swift` module, say `String`, it doesn't navigate there and my editor (Neovim) falls back to finding occurrences in the current file. Also I don't see the LSP attached and LSP says it's running in "single file mode" when I'm in this `.swiftinterface` file.  

If I go back to my project files and try to navigate to `String` from one of the project files, it navigates me to a `Swift.String.swiftinterface` file in that same folder for generated interfaces. So I'm guessing there should be a way to send an LSP request for generating an interface even when the originating file is in that generated interfaces folder but I'm not sure how or whether LSP can be configured to do so.

akaralar:
Just a small update, I managed to get LSP to attach when browsing the `.swiftinterface` file, but it attaches in "single file mode". Whether "single file mode" is something supported by `sourcekit-lsp` itself is not clear to me, it might as well be a feature of the editor. LSP does not attach when "single file mode" is disabled, my editor seems to think a `.swiftinterface` file is not part of a project.

Attaching LSP in single file mode does not make a difference in practice, LSP navigation still does not work.

PR #1460, opened 2024-06-08
A couple of improvements for generated interfaces

### LINKED ISSUE #878: Rename Swift Interface to Generated Interface
When generating an interface for a Swift module or a file as part of the `OpenInterfaceRequest`, we currently refer to it as a Swift Interface, which isn’t technically true. `.swiftinterface` files are distributed as part of library evolution, while the generated interfaces are generated by an IDE. 

We should use correct terminology here and refer to the interfaces we produce in sourcekit-lsp as *Generated Interfaces*, not *Swift Interfaces*.

ahoppen:
- Rename methods to highlight that we’re talking about generated interfaces here, not `.swiftinterface` files
- Don’t open the generated interface in `documentManager`. Opening documents in `documentManager` should only be done by the `textDocument/didOpen` notification from the LSP client. Otherwise we might indefinitely keep the document in the document manager
- After getting the generated interface from sourcekitd, close the document in sourcekitd again. We don’t provide semantic functionality in the generated interface yet, so we can’t interact with the generated interface path. Before, we left it open in sourcekitd indefinitely.
- A couple of code simplifications.

Fixes #878
rdar://116705653

### DIFF
--- a/Sources/SourceKitLSP/Swift/OpenInterface.swift
+++ b/Sources/SourceKitLSP/Swift/OpenInterface.swift
@@ -16,82 +16,83 @@ import LanguageServerProtocol
 import SKSupport
 import SourceKitD
 
-struct InterfaceInfo {
+struct GeneratedInterfaceInfo {
   var contents: String
 }
 
 extension SwiftLanguageService {
-  public func openInterface(_ request: OpenInterfaceRequest) async throws -> InterfaceDetails? {
-    let uri = request.textDocument.uri
-    let moduleName = request.moduleName
+  public func openGeneratedInterface(
+    _ request: OpenGeneratedInterfaceRequest
+  ) async throws -> GeneratedInterfaceDetails? {
     let name = request.name
     let symbol = request.symbolUSR
     let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent("\(name).swiftinterface")
     let interfaceDocURI = DocumentURI(interfaceFilePath)
     // has interface already been generated
     if let snapshot = try? self.documentManager.latestSnapshot(interfaceDocURI) {
-      return await self.interfaceDetails(request: request, uri: interfaceDocURI, snapshot: snapshot, symbol: symbol)
-    } else {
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -16,82 +16,83 @@ import LanguageServerProtocol
...
-    let moduleName = request.moduleName
+  public func openGeneratedInterface(
+    _ request: OpenGeneratedInterfaceRequest
+  ) async throws -> GeneratedInterfaceDetails? {
     let name = request.name
     let symbol = request.symbolUSR
     let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent("\(name).swiftinterface")
     let interfaceDocURI = DocumentURI(interfaceFilePath)
     // has interface already been generated
     if let snapshot = try? self.documentManager.latestSnapshot(interfaceDocURI) {
```
hamishknight, line 32: Just to make sure I understand, once the client has opened the interface, it gets populated in the `documentManager`?

hamishknight (approved):

bnbarham (approved):
> We don’t provide semantic functionality in the generated interface yet

IIRC sourcekitd does provide cursor-info for generated interface, is this just that we don't expose that from the LSP side yet?

Sources/SourceKitLSP/Swift/OpenInterface.swift
ahoppen: Yes, the client should send a `textDocument/didOpen` notification to SourceKit-LSP, which will then add it to the document manager like any other source file.

ahoppen:
> IIRC sourcekitd does provide cursor-info for generated interface, is this just that we don't expose that from the LSP side yet?

Yes, we currently don’t provide any intelligent editor functionality in swift interfaces (https://github.com/apple/sourcekit-lsp/issues/1155).

PR #1464, opened 2024-06-08
Show protocol requirements satisfied in extensions in jump-to-definition

ahoppen:
The key issue here was that we were looking at the symbol occurrence that had the `overrideOf` role and were using this symbol occurrence to also check that it’s a child of one of the receiver types. But if the protocol requirement is satisfied in an extension, we have an implicit symbol occurrence at the location where the protocol requirement is stated but the real declaration’s occurrence is inside the extension. The fix here is to just extract the USR from the `overrideOf` relation and then do another index lookup to find the symbol’s definition.

rdar://129412428

### DIFF
--- a/Tests/SourceKitLSPTests/DefinitionTests.swift
+++ b/Tests/SourceKitLSPTests/DefinitionTests.swift
@@ -601,4 +601,28 @@ class DefinitionTests: XCTestCase {
       ])
     )
   }
+
+  func testJumpToSatisfiedProtocolRequirementInExtension() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      protocol TestProtocol {
+        func 1️⃣doThing()
+      }
+
+      struct TestImpl: TestProtocol {}
+      extension TestImpl {
+        func 2️⃣doThing() { }
+      }
+      """
+    )
+
+    let response = try await project.testClient.send(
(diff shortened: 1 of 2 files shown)


bnbarham (approved):

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/DefinitionTests.swift
```diff
@@ -543,62 +543,27 @@ class DefinitionTests: XCTestCase {
...
-  func testJumpToDefinitionOnProtocolImplementationShowsAllFulfilledRequirements() async throws {
-    let project = try await IndexedSingleSwiftFileTestProject(
-      """
-      protocol TestProtocol {
-        func 1️⃣doThing()
-      }
-
-      protocol OtherProtocol {
-        func 2️⃣doThing()
-      }
```
hamishknight, line 585: Is it not worth keeping these tests?

hamishknight (approved):

Tests/SourceKitLSPTests/DefinitionTests.swift
ahoppen: Thank you for catching. I think I accidentally lost them during a rebase.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1465, opened 2024-06-08
Rewrite how “Preparing current file” is differentiated from the generic “Preparing targets”

ahoppen:
The old implementation of inspecting the `InProgressPrepareForEditorTask` has a few race conditions including the fact that we would start the preparation task before the `inProgressPrepareForEditorTask` was set.

Instead, keep track of the purpose for each preparation task. This ends up simplifying the code as well.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -414,7 +414,6 @@ public final actor SemanticIndexManager {
   ///
   /// If file's target is known to be up-to-date, this returns almost immediately.
   public func prepareFileForEditorFunctionality(_ uri: DocumentURI) async {
-    // Should be kept in sync with `schedulePreparationForEditorFunctionality`.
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -433,13 +421,16 @@ public final actor SemanticIndexManager {
     if Task.isCancelled {
       return
     }
```
hamishknight, line 423: There's a comment above `Should be kept in sync with schedulePreparationForEditorFunctionality`, is that outdated now?

hamishknight (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Yes, that’s no longer needed. Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #1298, opened 2024-05-14
Debounce work done progress that shows the index update


ahoppen:
We currently create a work done progress in the editor as soon as a file starts to be indexed. While this is great to see exactly what’s going on indexing-wise, it might be too noisy for users if `Indexing: 0/1` shows up for fractions of a section whenever a file is saved. We should consider debouncing its display.

ahoppen:
Synced to Apple’s issue tracker as rdar://128071328

PR #1467, opened 2024-06-08
Debounce the creation of work done progress for indexing and package reloading

ahoppen:
This avoids us from showing a work done progress for short-lived index tasks that take less than a second.

Fixes https://github.com/apple/sourcekit-lsp/issues/1298
rdar://128071328


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1472, opened 2024-06-10
Android: use the right setpriority() signature for Bionic

finagolfin:
I had to apply [a similar patch to keep this repo cross-compiling for Android](https://github.com/finagolfin/swift-android-sdk/commit/80a4fee0f8d0c0e517d6e4cdd2c38218844d75cb#diff-2d4e6bc96862356bd7129ebabe486bb38286f6f16b855df629879a4d617de4ba) on my daily CI.


ahoppen (approved):
Huh, interesting that Android has the same kind of non-standard POSIX headers as Darwin.

ahoppen:
@swift-ci Please test

finagolfin:
> Huh, interesting that Android has the same kind of non-standard POSIX headers as Darwin.

Both get a lot of their libc code from BSD is my understanding. 😄

PR #1473, opened 2024-06-11
Relax assertions about diagnostics from missing protocol requirements

ahoppen:
In some Swift 6 toolchains, the diagnostic about missing protocol requirements is emitted on the protocol in the inheritance clause instead of the conforming type. Relax a few assertions in the tests.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (commented):
LGTM

PR #1479, opened 2024-06-13
Add LSP extension to show Macro Expansions (or any document) in a "peeked" editor (and some minor quality improvements)

lokesh-tr:
### Intro:

**Note: This Feature is Experimental 🧪. You have to first enable this feature by passing `--experimental-feature show-macro-expansions` to `sourcekit-lsp`**

The previous PR (https://github.com/apple/sourcekit-lsp/pull/1436) lays the foundation for adding LSP support to show Macro Expansions of any kind. This PR brings better macro expansion functionality to all editors and some custom functionality through LSP extension for VS Code (, which other editors can also utilise by custom request handling.)

*If you are interested in bringing the custom functionality to your own editor, I highly recommend going through the PR below:
Accompanying PR in "Swift for VS Code" Extension":  https://github.com/swiftlang/vscode-swift/pull/945*

### About this pull request:

This implements an LSP Extension `PeekDocumentsRequest` to let `ExpandMacroCommand` to open the macro expansions in a "peeked" editor window. For this to work, the client has to pass "workspace/peekDocuments" enabled to `ClientCapabilities.experimental` and the client should handle the `PeekDocumentsRequest` and show the expansions in a "peeked" editor window.

PR to support the above capability in the "Swift for VS Code" Extension: https://github.com/swiftlang/vscode-swift/pull/945
The "Swift for VS Code" extension cannot send the client capability, so it instead passes the same through `initializationOptions` in the `InitializeRequest`.

For editors which doesn't support this capability, `sourcekit-lsp` sends a `ShowDocumentRequest`.
The `ShowDocumentRequest` is updated to show all the macro expansions in a single generated file. Moreover, its folder structure is updated to use hex string of MD5 hash of concatenation of buffer names of expansions.

This PR also does a **lot** of quality improvements.

### List of all changes:
 - changes the shown Line and Column numbers to be One-based indexing instead of Zero-based indexing, so that users can match the macro expansions' line and column numbers with their editor's line and column numbers.
 - adds a test case for `@attached` Macro Expansion, thereby assures LSP support for `@attached` Macro Expansions.
 - fixes a bug where `ShowDocumentRequest` fails due to invalid `selection`
 
 - makes `ShowDocumentRequest` to use only a single file to display all the expansions
 - implements a general `PeekDocumentsRequest` that can be used to make the editor to open files in a peeked editor
 - if the client sets the experimental client capability `peekDocuments` (or alternatively, in the `initializationOptions`), then the `ExpandMacroCommand` sends a `PeekDocumentsRequest` instead of a `ShowDocumentRequest`. (Accompanying PR in vscode-swift extension: https://github.com/swiftlang/vscode-swift/pull/945)
 - Fixes https://github.com/swiftlang/vscode-swift/issues/564
 
 - updated all `ExecuteCommandTests` appropriately
 - Fixes https://github.com/swiftlang/sourcekit-lsp/issues/1498 (rdar://130207754)
 - Added Documentation wherever necessary
 
 - updates folder structure to utilise MD5 hash for `ShowDocumentRequest` generated macro expansions
 
 ### References:

*Previous PR which laid the foundation: https://github.com/apple/sourcekit-lsp/pull/1436*
*Accompanying PR in the Swift Extension for VS Code repository: https://github.com/swiftlang/vscode-swift/pull/945*

-----

### DIFF
--- a/Documentation/LSP Extensions.md
+++ b/Documentation/LSP Extensions.md
@@ -436,3 +436,40 @@ Users should not need to rely on this request. The index should always be update
 ```ts
 export interface TriggerReindexParams {}
 ```
+
+## `workspace/peekDocuments`
+
+Request from the server to the client to show the given documents in a "peeked" editor.
+
+This request is handled by the client to show the given documents in a "peeked" editor (i.e. inline with / inside the editor canvas).
+
+It requires the experimental client capability `"workspace/peekDocuments"` to use.
+
+- params: `PeekDocumentsParams`
+- result: `PeekDocumentsResult`
+
+```ts
+export interface PeekDocumentsParams {
+  /**
+   * The `DocumentUri` of the text document in which to show the "peeked" editor
+   */
+  uri: DocumentUri;
+
+  /**
+   * The `Position` in the given text document in which to show the "peeked editor"
+   */
+  position: Position;
+
+  /**
+   * An array `DocumentUri` of the documents to appear inside the "peeked" editor
+   */
+  locations: DocumentUri[];
+}
+
+/**
+ * Response to indicate the `success` of the `PeekDocumentsRequest`
+ */
+export interface PeekDocumentsResult {
+  success: boolean;
+}
+```
--- a/Sources/LanguageServerProtocol/Requests/PeekDocumentsRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/PeekDocumentsRequest.swift
@@ -1,0 +1,57 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Request from the server to the client to show the given documents in a "peeked" editor **(LSP Extension)**
+///
+/// This request is handled by the client to show the given documents in a
+/// "peeked" editor (i.e. inline with / inside the editor canvas). This is
+/// similar to VS Code's built-in "editor.action.peekLocations" command.
+///
+/// - Parameters:
+///   - uri: The DocumentURI of the text document in which to show the "peeked" editor
+///   - position: The position in the given text document in which to show the "peeked editor"
+///   - locations: The DocumentURIs of documents to appear inside the "peeked" editor
+///
+/// - Returns: `PeekDocumentsResponse` which indicates the `success` of the request.
+///
+/// ### LSP Extension
+///
+/// This request is an extension to LSP supported by SourceKit-LSP.
+/// It requires the experimental client capability `"workspace/peekDocuments"` to use.
+/// It also needs the client to handle the request and present the "peeked" editor.
+public struct PeekDocumentsRequest: RequestType {
+  public static let method: String = "workspace/peekDocuments"
+  public typealias Response = PeekDocumentsResponse
+
+  public var uri: DocumentURI
+  public var position: Position
+  public var locations: [DocumentURI]
+
+  public init(
+    uri: DocumentURI,
+    position: Position,
+    locations: [DocumentURI]
+  ) {
+    self.uri = uri
+    self.position = position
+    self.locations = locations
+  }
+}
+
+/// Response to indicate the `success` of the `PeekDocumentsRequest`
+public struct PeekDocumentsResponse: ResponseType {
+  public var success: Bool
+
+  public init(success: Bool) {
+    self.success = success
+  }
+}
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -956,7 +956,25 @@ extension SourceKitLSPServer {
   }
 
   func initialize(_ req: InitializeRequest) async throws -> InitializeResult {
-    capabilityRegistry = CapabilityRegistry(clientCapabilities: req.capabilities)
+    // If the client can handle `PeekDocumentsRequest`, they can enable the
+    // experimental client capability `"workspace/peekDocuments"` through the `req.capabilities.experimental`.
+    //
+    // The below is a workaround for the vscode-swift extension since it cannot set client capabilities.
+    // It passes "workspace/peekDocuments" through the `initializationOptions`.
+    var clientCapabilities = req.capabilities
+    if case .dictionary(let initializationOptions) = req.initializationOptions,
+      let peekDocuments = initializationOptions["workspace/peekDocuments"]
+    {
+      if case .dictionary(var experimentalCapabilities) = clientCapabilities.experimental {
+        experimentalCapabilities["workspace/peekDocuments"] = peekDocuments
+        clientCapabilities.experimental = .dictionary(experimentalCapabilities)
+      } else {
+        clientCapabilities.experimental = .dictionary(["workspace/peekDocuments": peekDocuments])
+      }
+    }
+
+    capabilityRegistry = CapabilityRegistry(clientCapabilities: clientCapabilities)
+
     self.options = SourceKitLSPOptions.merging(
       base: self.options,
       override: orLog("Parsing SourceKitLSPOptions", { try SourceKitLSPOptions(fromLSPAny: req.initializationOptions) })
--- a/Sources/SourceKitLSP/Swift/MacroExpansion.swift
+++ b/Sources/SourceKitLSP/Swift/MacroExpansion.swift
@@ -95,7 +95,7 @@ extension SwiftLanguageService {
 
         // github permalink notation for position range
         let macroExpansionPositionRangeIndicator =
-          "L\(macroEdit.range.lowerBound.line)C\(macroEdit.range.lowerBound.utf16index)-L\(macroEdit.range.upperBound.line)C\(macroEdit.range.upperBound.utf16index)"
+          "L\(macroEdit.range.lowerBound.line + 1)C\(macroEdit.range.lowerBound.utf16index + 1)-L\(macroEdit.range.upperBound.line + 1)C\(macroEdit.range.upperBound.utf16index + 1)"
 
         let macroExpansionFilePath =
           macroExpansionBufferDirectoryURL
@@ -112,7 +112,7 @@ extension SwiftLanguageService {
         }
 
         Task {
-          let req = ShowDocumentRequest(uri: DocumentURI(macroExpansionFilePath), selection: macroEdit.range)
+          let req = ShowDocumentRequest(uri: DocumentURI(macroExpansionFilePath))
 
           let response = await orLog("Sending ShowDocumentRequest to Client") {
             try await sourceKitLSPServer.sendRequestToClient(req)
--- a/Sources/SourceKitLSP/Swift/MacroExpansion.swift
+++ b/Sources/SourceKitLSP/Swift/MacroExpansion.swift
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Crypto
 import Foundation
 import LSPLogging
 import LanguageServerProtocol
@@ -46,17 +47,15 @@ struct MacroExpansion: RefactoringResponse {
 extension SwiftLanguageService {
   /// Handles the `ExpandMacroCommand`.
   ///
-  /// Makes a request to sourcekitd and wraps the result into a `MacroExpansion`
-  /// and then makes a `ShowDocumentRequest` to the client side for each
-  /// expansion to be displayed.
+  /// Makes a `PeekDocumentsRequest` or `ShowDocumentRequest`, containing the
+  /// location of each macro expansion, to the client depending on whether the
+  /// client supports the `experimental["workspace/peekDocuments"]` capability.
   ///
   /// - Parameters:
   ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this request.
-  ///
-  /// - Returns: A `[RefactoringEdit]` with the necessary edits and buffer name as a `LSPAny`
   func expandMacro(
     _ expandMacroCommand: ExpandMacroCommand
-  ) async throws -> LSPAny {
+  ) async throws {
     guard let sourceKitLSPServer else {
       // `SourceKitLSPServer` has been destructed. We are tearing down the
       // language server. Nothing left to do.
@@ -69,6 +68,10 @@ extension SwiftLanguageService {
 
     let expansion = try await self.refactoring(expandMacroCommand)
 
+    var completeExpansionFileContent = ""
+    var completeExpansionDirectoryName = ""
+
+    var macroExpansionFilePaths: [URL] = []
     for macroEdit in expansion.edits {
       if let bufferName = macroEdit.bufferName {
         // buffer name without ".swift"
@@ -79,6 +82,9 @@ extension SwiftLanguageService {
 
         let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
           .appendingPathComponent(macroExpansionBufferDirectoryName)
+
+        completeExpansionDirectoryName += "\(bufferName)-"
+
         do {
           try FileManager.default.createDirectory(
             at: macroExpansionBufferDirectoryURL,
@@ -111,22 +117,95 @@ extension SwiftLanguageService {
           )
         }
 
-        Task {
-          let req = ShowDocumentRequest(uri: DocumentURI(macroExpansionFilePath))
+        macroExpansionFilePaths.append(macroExpansionFilePath)
 
-          let response = await orLog("Sending ShowDocumentRequest to Client") {
-            try await sourceKitLSPServer.sendRequestToClient(req)
-          }
+        let editContent =
+          """
+          // \(sourceFileURL.lastPathComponent) @ \(macroEdit.range.lowerBound.line + 1):\(macroEdit.range.lowerBound.utf16index + 1) - \(macroEdit.range.upperBound.line + 1):\(macroEdit.range.upperBound.utf16index + 1)
+          \(macroEdit.newText)
 
-          if let response, !response.success {
-            logger.error("client refused to show document for \(expansion.title, privacy: .public)")
-          }
-        }
+
+          """
+        completeExpansionFileContent += editContent
       } else if !macroEdit.newText.isEmpty {
         logger.fault("Unable to retrieve some parts of macro expansion")
       }
     }
 
-    return expansion.edits.encodeToLSPAny()
+    // removes superfluous newline
+    if completeExpansionFileContent.hasSuffix("\n\n") {
+      completeExpansionFileContent.removeLast()
+    }
+
+    if completeExpansionDirectoryName.hasSuffix("-") {
+      completeExpansionDirectoryName.removeLast()
+    }
+
+    var completeExpansionFilePath =
+      self.generatedMacroExpansionsPath.appendingPathComponent(
+        Insecure.MD5.hash(
+          data: Data(completeExpansionDirectoryName.utf8)
+        )
+        .map { String(format: "%02hhx", $0) }  // maps each byte of the hash to its hex equivalent `String`
+        .joined()
+      )
+
+    do {
+      try FileManager.default.createDirectory(
+        at: completeExpansionFilePath,
+        withIntermediateDirectories: true
+      )
+    } catch {
+      throw ResponseError.unknown(
+        "Failed to create directory for complete macro expansion at path: \(completeExpansionFilePath.path)"
+      )
+    }
+
+    completeExpansionFilePath =
+      completeExpansionFilePath.appendingPathComponent(sourceFileURL.lastPathComponent)
+    do {
+      try completeExpansionFileContent.write(to: completeExpansionFilePath, atomically: true, encoding: .utf8)
+    } catch {
+      throw ResponseError.unknown(
+        "Unable to write complete macro expansion to file path: \"\(completeExpansionFilePath.path)\""
+      )
+    }
+
+    let completeMacroExpansionFilePath = completeExpansionFilePath
+    let expansionURIs = macroExpansionFilePaths.map {
+      return DocumentURI($0)
+    }
+
+    if case .dictionary(let experimentalCapabilities) = self.capabilityRegistry.clientCapabilities.experimental,
+      case .bool(true) = experimentalCapabilities["workspace/peekDocuments"]
+    {
+      Task {
+        let req = PeekDocumentsRequest(
+          uri: expandMacroCommand.textDocument.uri,
+          position: expandMacroCommand.positionRange.lowerBound,
+          locations: expansionURIs
+        )
+
+        let response = await orLog("Sending PeekDocumentsRequest to Client") {
+          try await sourceKitLSPServer.sendRequestToClient(req)
+        }
+
+        if let response, !response.success {
+          logger.error("client refused to peek macro")
+        }
+      }
+    } else {
+      Task {
+        let req = ShowDocumentRequest(uri: DocumentURI(completeMacroExpansionFilePath))
+
+        let response = await orLog("Sending ShowDocumentRequest to Client") {
+          try await sourceKitLSPServer.sendRequestToClient(req)
+        }
+
+        if let response, !response.success {
+          logger.error("client refused to show document for macro expansion")
+        }
+      }
+    }
   }
 }
--- a/Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
+++ b/Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
@@ -69,11 +69,9 @@ extension SwiftLanguageService {
   ///
   /// - Parameters:
   ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
-  ///
-  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
   func semanticRefactoring(
     _ semanticRefactorCommand: SemanticRefactorCommand
-  ) async throws -> LSPAny {
+  ) async throws {
     guard let sourceKitLSPServer else {
       // `SourceKitLSPServer` has been destructed. We are tearing down the
       // language server. Nothing left to do.
@@ -94,7 +92,5 @@ extension SwiftLanguageService {
       }
       logger.error("client refused to apply edit for \(semanticRefactor.title, privacy: .public) \(reason)")
     }
-
-    return edit.encodeToLSPAny()
   }
 }
--- a/Tests/SourceKitLSPTests/ExecuteCommandTests.swift
+++ b/Tests/SourceKitLSPTests/ExecuteCommandTests.swift
@@ -252,7 +252,209 @@ final class ExecuteCommandTests: XCTestCase {
 
       XCTAssertEqual(
         url.lastPathComponent,
-        "MyMacroClient_L4C2-L4C19.swift",
+        "MyMacroClient_L5C3-L5C20.swift",
+        "Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+      )
+    }
+  }
+
+  func testAttachedMacroExpansion() async throws {
+    try await SkipUnless.canBuildMacroUsingSwiftSyntaxFromSourceKitLSPBuild()
+
+    let options = SourceKitLSPOptions.testDefault(experimentalFeatures: [.showMacroExpansions])
+
+    let project = try await SwiftPMTestProject(
+      files: [
+        "MyMacros/MyMacros.swift": #"""
+        import SwiftCompilerPlugin
+        import SwiftSyntax
+        import SwiftSyntaxBuilder
+        import SwiftSyntaxMacros
+
+        public struct DictionaryStorageMacro {}
+
+        extension DictionaryStorageMacro: MemberMacro {
+          public static func expansion(
+            of node: AttributeSyntax,
+            providingMembersOf declaration: some DeclGroupSyntax,
+            in context: some MacroExpansionContext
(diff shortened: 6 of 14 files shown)


Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -252,7 +252,278 @@ final class ExecuteCommandTests: XCTestCase {
...
+                }
+              }
+
+              return true
+
+            case .getter:
+              return false
+            }
+          }
+        }
```
ahoppen, line 317: I would prefer to keep the macro as simple as possible so that the long macro implementation doesn’t distract from the actual test here, which is the macro expansion. It’s totally fine if for these test cases the macro only works in the very narrow application in which we are using them in this test.

Concretely, this means that we can skip the entire `isStoredProperty` check here (because we know that all properties in our test are stored). 

Similarly, I think the `DictionaryStoragePropertyMacro` is not being used (macros should expand just fine even if the added accessor macro doesn’t exist, so we can remove it.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -252,7 +252,278 @@ final class ExecuteCommandTests: XCTestCase {
...
+        struct Point {
+          var x: Int = 1
+          var y: Int = 2
+        }
+
+        var point = Point()
+        print("Point storage begins as an empty dictionary: \(point)")
+        print("Default value for point.x: \(point.x)")
+        point.y = 17
+        print("Point storage contains only the value we set:  \(point)")
```
ahoppen, line 423: I think none of this is needed for the test

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -252,7 +252,278 @@ final class ExecuteCommandTests: XCTestCase {
...
+      )
+
+      let metadata = SourceKitLSPCommandMetadata(textDocument: TextDocumentIdentifier(uri))
+
+      var command = args.asCommand()
+      command.arguments?.append(metadata.encodeToLSPAny())
+
+      let request = ExecuteCommandRequest(command: command.command, arguments: command.arguments)
+
+      let expectations = [
```
ahoppen, line 453: You should be able to create an expectation and then set `expectation.expectedFulfillmentCount = 3` instead of an array of expectations.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -252,7 +252,278 @@ final class ExecuteCommandTests: XCTestCase {
...
+      XCTAssertEqual(
+        resultArray.count,
+        4,
+        "resultArray count is not equal to four. Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+      )
+
+      XCTAssertEqual(
+        resultArray.map {
+          $0.newText
+        }.sorted(),
```
ahoppen, line 487: Formatting nit
```suggestion
        resultArray.map { $0.newText }.sorted(),
```

ahoppen (commented):
Nice. I only have a few suggestions to simplify the test

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: I removed the `isStoredProperty` as you suggested. But, the test fails if `DictionaryStoragePropertyMacro` doesn't exist.

lokesh-tr:
@ahoppen I have also updated the PR to fix the bug where it fails to perform `ShowDocumentRequest` due to invalid `selection`.

lokesh-tr:
The last commit which I made breaks the test cases and is also a *proof-of-concept* implementation of the feature which will be implemented in a full-fledged manner.

Here's the accompanying PR https://github.com/swiftlang/vscode-swift/pull/945 in the VS Code Swift Extension Repository.

lokesh-tr:
@ahoppen I have updated everything as per last call, I will squash the commits as soon as this passes the review stage.

Documentation/LSP Extensions.md
```diff
@@ -376,3 +376,47 @@ Users should not need to rely on this request. The index should always be update
...
+
+    /**
+     * An array `Uri` of the documents to appear inside the "peeked" editor
+     */
+    locations: Uri[];
+
+    /**
+     * Presentation strategy when having multiple locations (default: "peek")
+     */
+    multiple: string;
```
ahoppen, line 413: This seems pretty VS Code-specific. I would remove this property. In my mind, if SourceKit-LSP sends this request,  it should always display the results in a peeked editor.

Documentation/LSP Extensions.md
```diff
@@ -376,3 +376,47 @@ Users should not need to rely on this request. The index should always be update
 ```ts
 export interface TriggerReindexParams {}
 ```
+
+## `sourcekit-lsp/peekDocuments`
```
ahoppen, line 380: All the other LSP extensions are named according to “normal” LSP requests, like `workspace/tests`. Let’s also name this as `workspace/peekDocuments`.

Documentation/LSP Extensions.md
```diff
@@ -376,3 +376,47 @@ Users should not need to rely on this request. The index should always be update
...
+
+Request from the server to the client to show the given documents in a "peeked" editor.
+
+This request is handled by the client to show the given documents in a "peeked" editor (i.e. inline with / inside the editor canvas).
+
+It requires the experimental client capability `"peekDocuments"` to use.
```
ahoppen, line 386: I don’t think you currently have any capability check in place. Or am I missing something?

Sources/LanguageServerProtocol/Requests/PeekDocumentsRequest.swift
```diff
@@ -0,0 +1,70 @@
...
+///   - position: The position in the given text document in which to show the
+/// "peeked editor" (default: nil, current cursor position in the active editor handled by the client)
+///   - locations: The URIs of documents to appear inside the "peeked" editor
+///   - multiple: Presentation strategy when having multiple locations (default: "peek")
+///
+/// - Returns: `PeekDocumentsResponse` which indicates the `success` of the request.
+///
+/// ### LSP Extension
+///
+/// This request is an extension to LSP supported by SourceKit-LSP and clangd.
```
ahoppen, line 31: I don’t think that clangd supports the `sourcekit-lsp/peekDocuments` request 🧐

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -958,7 +958,26 @@ extension SourceKitLSPServer {
...
+    // If the client can handle `PeekDocumentsRequest`, they can enable the
+    // experimental client capability `"peekDocuments"` through the `req.capabilities.experimental`.
+    //
+    // The below allows the client to pass the `"peekDocuments"` flag in the
+    // `req.initializationOptions` if for some reason the client cannot pass it
+    // directly through `req.capabilities.experimental` .
+    var clientCapabilities = req.capabilities
+    if case .dictionary(let initializationOptions) = req.initializationOptions,
+      let peekDocuments = initializationOptions["peekDocuments"]
+    {
```
ahoppen, line 970: Do we even need to go through `initializationOptions` here? Shouldn’t clients just declare the capabilities in `ClientCapabilities.experimental`?

Also, should we use the request name as the key here? Ie. `workspace/peekDocuments` or `sourcekit-lsp/peekDocuments`.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -21,13 +21,15 @@ import SourceKitD
...
 
   /// The URI of the file where the macro is used
-  var uri: DocumentURI
+  let uri: DocumentURI
 
   /// The resulting array of `RefactoringEdit` of a semantic refactoring request
-  var edits: [RefactoringEdit]
+  let edits: [RefactoringEdit]
+
+  var expansionURIs: [DocumentURI]?
```
ahoppen, line 32: Could you add a doc comment?

Also, out of curiosity: Why did you change all properties to `let` and kept this one as a `var`?

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -141,120 +159,362 @@ final class ExecuteCommandTests: XCTestCase {
...
+          let url = try XCTUnwrap(
+            showDocumentRequestURI.value?.fileURL,
+            "Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+          )
+
+          let fileContents = try String(contentsOf: url, encoding: .utf8)
+
+          XCTAssertEqual(
+            fileContents,
+            "// MyMacroClient.swift @ 11:3 - 11:3\n@DictionaryStorageProperty\n// MyMacroClient.swift @ 12:3 - 12:3\n@DictionaryStorageProperty\n// MyMacroClient.swift @ 13:1 - 13:1\nvar _storage: [String: Any] = [:]",
```
ahoppen, line 507: Let’s use a multi-line string literal here. I think that’s easier to understand as a human.

Also, I think it would be nice to have an empty line before every `//` comment that’s not on the first line in the file, it loosens up the file a little. Ie. something like

```suggestion
            """
            // MyMacroClient.swift @ 11:3 - 11:3
            @DictionaryStorageProperty
            
            // MyMacroClient.swift @ 12:3 - 12:3
            @DictionaryStorageProperty

            // MyMacroClient.swift @ 13:1 - 13:1
            var _storage: [String: Any] = [:]",
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -79,6 +86,10 @@ extension SwiftLanguageService {
 
         let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
           .appendingPathComponent(macroExpansionBufferDirectoryName)
+        completeExpansionFilePath =
+          completeExpansionFilePath
+          .appendingPathComponent(macroExpansionBufferDirectoryName)
```
ahoppen, line 91: Doesn’t this add another component for every macro expansion buffer and result in file names like `path/to/temp/macroBuffer1.swift/macroBuffer2.swift/macroBuffer3.swift`. If we want to concatenate them, I think we should concatenate them into a single directory name and not create nested directories.

Also, maybe we should just us a hash of all the macro expansions that can exist. I can imagine that if we concatenate the macro buffer names, we very quickly exceed the 260 character maximum file path length  on Windows. https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -69,6 +73,9 @@ extension SwiftLanguageService {
 
     let expansion = try await self.refactoring(expandMacroCommand)
 
+    var completeExpansionFileContent = ""
+    var completeExpansionFilePath = self.generatedMacroExpansionsPath
+    var macroExpansionFilePaths = [URL]()
```
ahoppen, line 78: Nitpick, the preferred spelling is

```suggestion
    var macroExpansionFilePaths: [URL] = []
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -111,22 +122,77 @@ extension SwiftLanguageService {
...
+        completeExpansionFileContent += editContent
       } else if !macroEdit.newText.isEmpty {
         logger.fault("Unable to retrieve some parts of macro expansion")
       }
     }
 
-    return expansion.edits.encodeToLSPAny()
+    if completeExpansionFileContent.hasSuffix("\n") {
+      completeExpansionFileContent.removeLast()
+    }
```
ahoppen, line 137: Why not keep the trailing newline? All the source files in SourceKit-LSP also have trailing newlines

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -40,23 +42,25 @@ struct MacroExpansion: RefactoringResponse {
...
   ///
   /// - Parameters:
   ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this request.
   ///
-  /// - Returns: A `[RefactoringEdit]` with the necessary edits and buffer name as a `LSPAny`
+  /// - Returns: nil
   func expandMacro(
     _ expandMacroCommand: ExpandMacroCommand
-  ) async throws -> LSPAny {
+  ) async throws -> LSPAny? {
```
ahoppen, line 63: Can we just remove the return value of the function if it always returns `nil`?

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -70,10 +70,10 @@ extension SwiftLanguageService {
   /// - Parameters:
   ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
   ///
-  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
+  /// - Returns: nil
   func semanticRefactoring(
     _ semanticRefactorCommand: SemanticRefactorCommand
-  ) async throws -> LSPAny {
+  ) async throws -> LSPAny? {
```
ahoppen, line 76: Can we remove the return value?

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -46,19 +46,28 @@ final class ExecuteCommandTests: XCTestCase {
...
     testClient.handleSingleRequest { (req: ApplyEditRequest) -> ApplyEditResponse in
+      applyEditTitle.value = req.label
+      applyEditWorkspaceEdit.value = req.edit
+      expectation.fulfill()
+
       return ApplyEditResponse(applied: true, failureReason: nil)
     }
 
-    let result = try await testClient.send(request)
+    _ = try await testClient.send(request)
```
ahoppen, line 61: I think with https://github.com/swiftlang/sourcekit-lsp/pull/1519 you no longer need the `_ =`.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -46,19 +46,28 @@ final class ExecuteCommandTests: XCTestCase {
...
-      return
-    }
+    try await fulfillmentOfOrThrow([expectation])
 
+    let label = try XCTUnwrap(applyEditTitle.value)
+    let edit = try XCTUnwrap(applyEditWorkspaceEdit.value)
+
+    XCTAssertEqual(label, "Localize String")
     XCTAssertEqual(
-      WorkspaceEdit(fromLSPDictionary: resultDict),
```
ahoppen, line 61: Can we remove the `WorkspaceEdit(fromLSPDictionary:)` initializer and `WorkspaceEdit.encodeToLSPAny` now?

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -141,120 +159,362 @@ final class ExecuteCommandTests: XCTestCase {
...
-        """,
-      ],
-      manifest: SwiftPMTestProject.macroPackageManifest,
-      options: options
-    )
-    try await SwiftPMTestProject.build(at: project.scratchDirectory)
+      }
 
-    let (uri, positions) = try project.openDocument("MyMacroClient.swift")
+      public struct DictionaryStoragePropertyMacro: AccessorMacro {
```
ahoppen, line 346: Picking up https://github.com/swiftlang/sourcekit-lsp/pull/1479#discussion_r1649078893 again:

Ah, I see. You need to enable background indexing (passing `enableBackgroundIndexing: true` to SwiftPMTestProject`) and remove the `try await SwiftPMTestProject.build(at: project.scratchDirectory)` call. Then it should work.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -141,120 +159,362 @@ final class ExecuteCommandTests: XCTestCase {
...
+
+          XCTAssertEqual(
+            urls.map { $0.lastPathComponent }.sorted(),
+            [
+              "MyMacroClient_L11C3-L11C3.swift",
+              "MyMacroClient_L12C3-L12C3.swift",
+              "MyMacroClient_L13C1-L13C1.swift",
+            ].sorted(),
+            "Failed for position range between \(positionMarker.start) and \(positionMarker.end)"
+          )
```
ahoppen, line 483: Why do we need to sort the URLs but not the file contents?

ahoppen (commented):
Wohooo 🚀

Sources/SourceKitLSP/SourceKitLSPServer.swift
lokesh-tr: I was not able to find a way to send client capabilities from the vscode-swift extension. I asked @adam-fowler to let me know how to let sourcekit-lsp know that we support "peekDocuments" and he said the following:

> If I remember right In LanguageClientManager.createLSPClient in the clientOptions you can set initialization options which are passed to the server.
> [11:56](https://swift-open-source.slack.com/archives/D072BNXPFMW/p1719599209024749)
> I discuss something similar here https://github.com/swiftlang/sourcekit-lsp/issues/797#issuecomment-1974750452

The current implementation allows the client to set "peekDocuments" through the client capabilities, and also allows an alternative way to set the same through `initializationOptions`.

@adam-fowler Let me know if I am missing something?

Or perhaps, we can eliminate the capabilities route and stick with just `initializationOptions`, but it doesn't stick with me design wise.

Let me have your thoughts on this @ahoppen @adam-fowler

Documentation/LSP Extensions.md
lokesh-tr: https://github.com/swiftlang/sourcekit-lsp/pull/1479#discussion_r1660913419
@ahoppen @adam-fowler

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: I first thought of concatenating the buffer names into a single directory name, and I realised the character limit issue myself.

So, I decided to make sub directories. The sub directories won't have the `.swift` extension and thus the location will be like `path/to/temp/macroBuffer1/macroBuffer2/macroBuffer3/SourceFile.swift`.

I am not sure about the "hash" idea which you are talking about, perhaps you could enlighten me on this.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Ah, I see. I’m fine with keeping the initialization options as a secondary way to enable `peekDocument`, I would just add a comment that explains that this is a workaround for VS Code, which can’t set client capabilities from an extension for future reference.

Documentation/LSP Extensions.md
ahoppen: Ah, sorry. I forgot to remove this after I found the check via `initializeOptions`.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
ahoppen: I was thinking that you essentially build up a string `macroBuffer1-macroBuffer2-macroBuffer3` and then, to shorten it, take an MD5 (or some other but MD5 is fine because it doesn’t need to be cryptographic) hash of it, similar to 

https://github.com/swiftlang/sourcekit-lsp/blob/e5d93e1ddd218811693221abe1c45230752d38b6/Sources/LSPLogging/CustomLogStringConvertible.swift#L67

That way we get a directory name of a constant length.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: With file contents, we have to ensure that the macro expands in the same order as we expect. I don't think we should sort them, since it won't allows us to test the order of expansion.

However, with the file names, whatever the order it maybe, as long as the line and character numbers are correct, we are good to go. So I `sorted` them to let the equality check to pass, as long as they contain the same elements, irrespective of the order.

Let me know if I overlooked anything here.

Sources/LanguageServerProtocol/Requests/PeekDocumentsRequest.swift
lokesh-tr: Oops, sleepy mistake 🤦🏻‍♂️

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
ahoppen: But the order of the file names is always deterministic, right? So it would be fine to compare them without sorting, right? That way we would effectively also be checking that the correct contents are in the correct file.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: Or should I allow this to be:

```
// MyMacroClient.swift @ 11:3 - 11:3
@DictionaryStorageProperty
// MyMacroClient.swift @ 12:3 - 12:3
@DictionaryStorageProperty
// MyMacroClient.swift @ 13:1 - 13:1
var _storage: [String: Any] = [:]",
            
```

Do you want me to let the macro expansion implementation to add new lines before the comments?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
ahoppen: For structs I often prefer `var`. Structs are value types so shared mutable state is not really an issue and if the properties are `let`, you can’t create a modified copy as follows.

```swift
var myCopy = someStruct
myCopy.someProperty = 2
doSomething(with: myCopy)
```

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: I guess we can.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: I will try this.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: That makes sense now.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
lokesh-tr: Makes sense 👍🏻

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
ahoppen: I’m not entirely sure what you mean.

Documentation/LSP Extensions.md
```diff
@@ -376,3 +376,40 @@ Users should not need to rely on this request. The index should always be update
...
+
+  /**
+   * The `Position` in the given text document in which to show the "peeked editor"
+   */
+  position: vscode.Position;
+
+  /**
+   * An array `DocumentUri` of the documents to appear inside the "peeked" editor
+   */
+  locations: langclient.DocumentUri[];
```
ahoppen, line 406: The LSP spec, which we’re following here, doesn’t module-qualify `DocumentUri` etc.

```suggestion
  /**
   * The `DocumentUri` of the text document in which to show the "peeked" editor
   */
  uri: DocumentUri;

  /**
   * The `Position` in the given text document in which to show the "peeked editor"
   */
  position: Position;

  /**
   * An array `DocumentUri` of the documents to appear inside the "peeked" editor
   */
  locations: DocumentUri[];
```

Sources/LanguageServerProtocol/Requests/PeekDocumentsRequest.swift
```diff
@@ -0,0 +1,57 @@
...
+///   - uri: The DocumentURI of the text document in which to show the "peeked" editor
+///   - position: The position in the given text document in which to show the "peeked editor"
+///   - locations: The DocumentURIs of documents to appear inside the "peeked" editor
+///
+/// - Returns: `PeekDocumentsResponse` which indicates the `success` of the request.
+///
+/// ### LSP Extension
+///
+/// This request is an extension to LSP supported by SourceKit-LSP.
+/// It requires the experimental client capability `"peekDocuments"` to use.
```
ahoppen, line 29: Picking up my second comment in https://github.com/swiftlang/sourcekit-lsp/pull/1479#discussion_r1660721111

Should we call the experimental client capability `workspace/peekDocuments` to match the request’s method?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Crypto
```
ahoppen, line 13: Could you add `Crypto` as a dependency of `SourceKitLSP` in the Package.swift? Ie. add 
```swift
.product(name: "Crypto", package: "swift-crypto"),
```

to the dependencies of the `SourceKitLSP` target?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -46,17 +47,17 @@ struct MacroExpansion: RefactoringResponse {
...
-  /// expansion to be displayed.
+  /// Makes a `PeekDocumentsRequest` or `ShowDocumentRequest`, containing the
+  /// location of each macro expansion, to the client depending on whether the
+  /// client supports the `experimental["peekDocuments"]` capability.
   ///
   /// - Parameters:
   ///   - expandMacroCommand: The `ExpandMacroCommand` that triggered this request.
   ///
-  /// - Returns: A `[RefactoringEdit]` with the necessary edits and buffer name as a `LSPAny`
+  /// - Returns: nil
```
ahoppen, line 57: No longer applies since the function doesn’t have a return value anymore.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -111,22 +119,96 @@ extension SwiftLanguageService {
...
+      completeExpansionDirectoryName.removeLast()
+    }
+
+    var completeExpansionFilePath =
+      self.generatedMacroExpansionsPath.appendingPathComponent(
+        Insecure.MD5.hash(
+          data: Data(completeExpansionDirectoryName.utf8)
+        )
+        .map { String(format: "%02hhx", $0) }
+        .joined()
```
ahoppen, line 152: What kind of directory name does this produce? How does this compare to the `.description` that we do in `hashForLogging`?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -111,22 +119,96 @@ extension SwiftLanguageService {
...
+    }
+
+    let completeMacroExpansionFilePath = completeExpansionFilePath
+    let expansionURIs = macroExpansionFilePaths.map {
+      return DocumentURI($0)
+    }
+
+    if case .dictionary(let experimentalCapabilities) = self.capabilityRegistry.clientCapabilities.experimental,
+      case .bool(let peekDocuments) = experimentalCapabilities["peekDocuments"],
+      peekDocuments
```
ahoppen, line 183: We can shorten this slightly.

```suggestion
      case .bool(true) = experimentalCapabilities["peekDocuments"]
```

Sources/SourceKitLSP/Swift/SemanticRefactoring.swift
```diff
@@ -70,10 +70,10 @@ extension SwiftLanguageService {
   /// - Parameters:
   ///   - semanticRefactorCommand: The `SemanticRefactorCommand` that triggered this request.
   ///
-  /// - Returns: A `WorkspaceEdit` with the necessary refactors as a `LSPAny`
+  /// - Returns: nil
```
ahoppen, line 73: Also no longer applicable because there’s no return value anymore.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: I mistook `.description` to provide some debugging description. Seems its the same as what I am trying to achieve. Will change this. 👍🏻

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: That’s cool

Sources/SourceKitLSP/Swift/MacroExpansion.swift
lokesh-tr: After playing around with it, seems I was right. `.description` indeed provides a debug description "MD5 Digest: e1c37d2fd6a2a2efdfab947c8715b5e4".
My original method, maps each byte into hex and joins them into a string. and it outputs the following "e1c37d2fd6a2a2efdfab947c8715b5e4"

I guess my method is fine.

lokesh-tr:
@ahoppen I have addressed your comments, let me know if there are any other concerns. I have also squashed the commits.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
ahoppen: Oh, interesting. Thanks for adding the comment.

ahoppen (approved):
Let’s see what CI has to say 🚀

ahoppen:
@swift-ci Please test

lokesh-tr:
@ahoppen trigger Test windows ?

ahoppen:
@swift-ci Please test Windows

lokesh-tr:
@ahoppen All the test cases pass. But, seems like a merge conflict happened in between. But the conflicting file `Sources/SourceKitLSP/Swift/Refactoring.swift` is already deleted / renamed to `RefactoringResponse.swift`. Do you want me to rebase once again and go through the tests once again? I guess it would be fine if you merge it directly by removing this file. 🤔

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1480, opened 2024-06-17
Skip `testJumpToSatisfiedProtocolRequirementInExtension` if the toolchain does not mark overridable function declarations as dynamic

ahoppen:
This test requires https://github.com/apple/swift/pull/74080 and should be skipped if the host toolchain does not have that change. 

Fixes a test failure when running SourceKit-LSP tests using Xcode 15.4.

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -233,6 +233,37 @@ public actor SkipUnless {
     }
   }
 
+  /// Checks if the toolchain contains https://github.com/apple/swift/pull/74080.
+  public static func sourcekitdReportsOverridableFunctionDefinitionsAsDynamic(
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws {
+    struct ExpectedLocationsResponse: Error {}
+
+    return try await shared.skipUnlessSupportedByToolchain(swiftVersion: SwiftVersion(6, 0), file: file, line: line) {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -233,6 +233,37 @@ public actor SkipUnless {
     }
   }
 
+  /// Checks if the toolchain contains https://github.com/apple/swift/pull/74080.
```
bnbarham, line 236: I'm not really sure what to do about this, but the skip test is really the test itself 😅

bnbarham (approved):

Sources/SKTestSupport/SkipUnless.swift
ahoppen: Yes, I know. The key point here is that it won’t skip the test if the toolchain is ≥6.1 or if we are on Swift CI (`SWIFTCI_USE_LOCAL_DEPS ` is set – https://github.com/apple/sourcekit-lsp/blob/main/Documentation/Environment%20Variables.md#build-time)

bnbarham: Well, my point was more that it would skip if the original test would fail, though depends on the failure (ie. if *more* results are returned then it wouldn't skip, but if the result is missing at all then it will). Anyway, I think it's fine 👍

PR #1481, opened 2024-06-18
Add workaround for rdar://116221716

ahoppen:
This works around a compiler crash when building SourceKit-LSP using a 6.0 Swift development snapshot.


ahoppen:
@swift-ci Please test

PR #1485, opened 2024-06-18
When setting a working directory on process launch is not supported, set it using `sh`

ahoppen:
In particular, this affects Amazon Linux 2, which has a glibc that doesn’t support `posix_spawn_file_actions_addchdir_np`.

rdar://128016626

### DIFF
--- a/Sources/SKSupport/Process+Run.swift
+++ b/Sources/SKSupport/Process+Run.swift
@@ -69,9 +69,33 @@ extension Process {
     do {
       try process.launch()
     } catch Process.Error.workingDirectoryNotSupported where workingDirectory != nil {
-      // TODO (indexing): We need to figure out how to set the working directory on all platforms.
+      return try Process.launchWithWorkingDirectoryUsingSh(
+        arguments: arguments,
+        environmentBlock: environmentBlock,
+        workingDirectory: workingDirectory!,
+        outputRedirection: outputRedirection,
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

Sources/SKSupport/Process+Run.swift
```diff
@@ -82,7 +105,16 @@ extension Process {
         loggingHandler: loggingHandler
       )
     }
-    return process
+    do {
+      return try await Process.launch(
+        arguments: [sh.path, "-c", #"cd $0; "$@""#, workingDirectory.pathString] + arguments,
```
hamishknight, line 110: Should probably quote `$0` here, also could use `exec` to avoid a fork

```suggestion
        arguments: [sh.path, "-c", #"cd "$0"; exec "$@""#, workingDirectory.pathString] + arguments,
```

hamishknight (approved):

Sources/SKSupport/Process+Run.swift
ahoppen: Great suggestions. Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1487, opened 2024-06-18
Save sourcekit-lsp diagnose bundles in subdirectories of temporary directory

ahoppen:
Finder seems to generally struggle opening a temporary directory with many subfolders and eg. hang when zipping the diagnose bundle. Place the diagnose folders in a dedicated subfolder to avoid this.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1488, opened 2024-06-18
Add comment to keep the message about keeping the list of information collected in sourcekit-lsp in sync with the VS Code plugin

ahoppen:
This command-line message needs to be kept in-sync with

https://github.com/swiftlang/vscode-swift/pull/830/files#diff-fb34ccc6e0def68748fe8adf02c9147377045d01322de282b8fafb416850cb1eR94-R106


ahoppen:
@swift-ci Please test

PR #1489, opened 2024-06-18
Remove warning message about missing compiler arguments

ahoppen:
It turns out that this message was more noise than help. For example, it would often show up when adding a new file to a SwiftPM project: The file gets added before we reload the package and thus we don’t have build settings for the new file for a short while. Since we can’t dismiss the notification we sent to the client, the notification will stick around.

Let’s just remove the message.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1490, opened 2024-06-18
Resolve symlinks in toolchain path when creating sourcekitd reproducer bundle

ahoppen:
The realpath will be more useful than eg. `/Library/Developer/Toolchains/swift-latest.xctoolchain`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1493, opened 2024-06-19
Allow prefixing of the token for a `WorkDoneProgress` with a custom string

### LINKED ISSUE #1476: Add an identifier to work done progress messages
### Description

[Work done progress messages](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress) (`$/progress`) contain a `message` and a `title`, but there is no `id` to match begin/end events up together. It is possible to use the title, but if developers want to check for specific types of work progress they'll have to key off the title and if that title ever changes in future then their code will break.

It would be nice to add an `id` field to these progress messages that is considered immutable. One catch is there is no `id` field in the [protocol definition](https://github.com/microsoft/vscode-languageserver-node/blob/1ce790dd341a38da1fbf902cddc16ce456149de0/protocol/src/common/protocol.progress.ts#L11) so this would be out of spec.

ahoppen:
This allows VS Code to detect when sourcekitd has crashed and prompt the user to gather a diagnostic report + file an issue about the crash.

I took this opportunity to refactor the WorkDoneProgress that effectively counts how many tasks of this kind are going on, into a separate type. I’m not quite happy with the name `SharedWorkDoneProgressManager`. Happy for better suggestions if anyone has some.

rdar://129678779
Fixes #1476

### DIFF
--- a/Sources/SourceKitLSP/WorkDoneProgressManager.swift
+++ b/Sources/SourceKitLSP/WorkDoneProgressManager.swift
@@ -20,7 +20,7 @@ import SwiftExtensions
 ///
 /// The work done progress is started when the object is created and ended when the object is destroyed.
 /// In between, updates can be sent to the client.
-final actor WorkDoneProgressManager {
+actor WorkDoneProgressManager {
   private enum Status: Equatable {
     case inProgress(message: String?, percentage: Int?)
     case done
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

Sources/SourceKitLSP/WorkDoneProgressManager.swift
```diff
@@ -177,3 +190,68 @@ final actor WorkDoneProgressManager {
...
+      )
+    }
+  }
+
+  func end() async {
+    if inProgressOperations > 0 {
+      inProgressOperations -= 1
+    } else {
+      logger.fault(
+        "Unbalanced calls to SharedWorkDoneProgressManager.start and end for \(tokenPrefix, privacy: .public)"
```
plemarquand, line 249: I checked this out locally but the patch failed to build because `tokenPrefix` should be `self.tokenPrefix`. After fixing this it worked as expected.

```
sourcekit-lsp/Sources/SourceKitLSP/WorkDoneProgressManager.swift:249:80: error: reference to property 'tokenPrefix' in closure requires explicit use of 'self' to make capture semantics explicit
247 |     } else {
248 |       logger.fault(
249 |         "Unbalanced calls to SharedWorkDoneProgressManager.start and end for \(tokenPrefix, privacy: .public)"
    |                                                                                |- error: reference to property 'tokenPrefix' in closure requires explicit use of 'self' to make capture semantics explicit
    |                                                                                `- note: reference 'self.' explicitly
250 |       )
251 |     }
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

PR #1478, opened 2024-06-12
Add log messages to track target preparation

ahoppen:
These should help us figure out the root cause of rdar://129698768 if we hit it again.

PR #1499, opened 2024-06-20
Fix a nondeterministic test failure of `testDontStackTargetPreparationForEditorFunctionality`

ahoppen:
We fulfilled the `allDocumentsOpened` expectation directly after dispatching the notification to open LibD.swift. This left a very short window in which LibB’s preparation could finish and LibC’s expectation could start before the open notification for LibD.swift was handled, causing an unexpected preparation of LibC.

This fixes the issue first reported as rdar://129698768 and for which I added logging in https://github.com/apple/sourcekit-lsp/pull/1478.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1500, opened 2024-06-20
Send `SIGKILL` to `swift-frontend` indexing processes

ahoppen:
We were sending `SIGINT` to `swift-frontend` processes if they didn’t terminate after 2 minutes. However, `swift-frontend` doesn’t listen to `SIGINT`.

If a task running `waitUntilExitStoppingProcessOnTaskCancellation` is cancelled and the process doesn’t terminate on a `SIGINT` after 2 seconds, kill it.

rdar://130103147

### DIFF
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -19,6 +19,7 @@ import SwiftExtensions
 
 import struct TSCBasic.AbsolutePath
 import class TSCBasic.Process
+import struct TSCBasic.ProcessResult
 
 private let updateIndexStoreIDForLogging = AtomicUInt32(initialValue: 1)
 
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -360,15 +368,21 @@ public struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
     // than 2 minutes in practice, so this indicates that the compiler has entered a loop and we probably won't make any
     // progress here. We will try indexing the file again when it is edited or when the project is re-opened.
     // 2 minutes have been chosen arbitrarily.
```
hamishknight, line 370: Should this comment be updated/moved?

hamishknight (approved):
Yeah we really ought to fix swift-frontend to respect SIGINT

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Good catch 👍🏽

ahoppen:
@swift-ci Please test

PR #1501, opened 2024-06-21
Watch for changes to `Package.resolved`

ahoppen:
We need to watch for changes to `Package.resolved` so that we can update the dependency checkouts in `.index-build` when the user runs `swift package update`.

rdar://130103181


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1502, opened 2024-06-22
Log path of the toolchain used to opening a document

ahoppen:
I don’t know why we logged the toolchain’s identifier twice. That didn’t make much sense.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1506, opened 2024-06-22
[6.0] Watch for changes to `Package.resolved`

ahoppen:
- **Explanation**: We need to watch for changes to `Package.resolved` so that we can update the dependency checkouts in `.index-build` when the user runs `swift package update`.
- **Scope**: Background indexing when the user updates package dependencies
- **Risk**: Low, this uses the same code path that we use to watch for changes to `Package.swift`
- **Issue**: rdar://130103181
- **Reviewer**:   @hamishknight on https://github.com/apple/sourcekit-lsp/pull/1501


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1508, opened 2024-06-22
Add test case that we don't reload the package if a `.swift` file get added in a folder that doesn't affect compilation

ahoppen:
Companion of https://github.com/apple/swift-package-manager/pull/7699

rdar://128573306


ahoppen:
https://github.com/apple/swift-package-manager/pull/7699

@swift-ci Please test

PR #1510, opened 2024-06-23
Translate the semantic token legend used by clangd to the semantic token legend used by SourceKit-LSP

ahoppen:
clangd uses a completely different semantic token legend than SourceKit-LSP (it doesn’t even adhere to the ordering of the pre-defined token types) but we were passing index offsets from clangd through assuming that clangd uses the same legend, which was incorrect.

When retrieving semantic tokens from clangd, translate the semantic tokens from clangd’s legend to SourceKit-LSP’s legend.

rdar://129895062

### DIFF
--- a/Sources/SourceKitLSP/Clang/SemanticTokenTranslator.swift
+++ b/Sources/SourceKitLSP/Clang/SemanticTokenTranslator.swift
@@ -1,0 +1,136 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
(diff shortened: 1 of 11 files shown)


Sources/SourceKitLSP/Clang/SemanticTokenTranslator.swift
```diff
@@ -0,0 +1,136 @@
...
+    /// clangd index to this SourceKit-LSP index.
+    case translation(UInt32)
+  }
+
+  /// For all token types whose representation in clang differs from the representation in SourceKit-LSP, maps the
+  /// index of that token type in clangd’s token type legend to the corresponding representation in SourceKit-LSP.
+  private let tokenTypeTranslations: [UInt32: Translation]
+
+  /// For all token modifiers whose representation in clang differs from the representation in SourceKit-LSP, maps the
+  /// index of that token type in clangd’s token type legend to the corresponding representation in SourceKit-LSP.
```
bnbarham, line 34:
```suggestion
  /// index of that token type in clangd’s token modifier legend to the corresponding representation in SourceKit-LSP.
```

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1512, opened 2024-06-25
[6.0] Add test case that we don't reload the package if a `.swift` file get added in a folder that doesn't affect compilation

ahoppen:
- **Explanation**: Instead of inspecting file extensions to decide whether a file might affect build settings, check which files are in directories that are affecting compilation. This is the same logic that `LLBuildManifestBuilder` uses and should be more stable. In particular, this stops us from reloading the package manifest in SourceKit-LSP when a header file is written to a `.build` directory.
- **Scope**: Only affects SourceKit-LSP
- **Risk**: Low, only affects SourceKit-LSP
- **Testing**: Added a test case
- **Issue**: rdar://128573306
- **Reviewer**:  @MaxDesiatov @xedin on https://github.com/swiftlang/swift-package-manager/pull/7699


ahoppen:
https://github.com/swiftlang/swift-package-manager/pull/7709

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-package-manager/pull/7709

@swift-ci Please test

PR #1517, opened 2024-06-25
Add documentation about each module's purpose and move some files between modules

ahoppen:
The purpose of the different modules wasn’t clearly defined, which lead to inconsistent responsibilities between the different modules. Define each module’s purpose and move a few files between modules to satisfy these definitions.

There are a few more larger changes that will need to be made for a fully consistent module structure. These are FIXMEs in the new Modules.md document and I’ll address them in follow-up PRs.


ahoppen:
@swift-ci Please test

Documentation/Modules.md
```diff
@@ -0,0 +1,121 @@
+# Modules
+
+The SourceKit-LSP package contains the following non-testing modules.
```
bnbarham, line 3: What's the "non-testing" mean here?

hamishknight (approved):

Documentation/Modules.md
ahoppen: Technically the `testTargets` are also modules. But I didn’t include them here because they aren’t really relevant for the modules structure.

bnbarham: Ah I see 👍 Not sure why this didn't make me think of `testTarget`.

PR #1524, opened 2024-06-26
Allow configuring of SourceKit-LSP’s options using configuration files

ahoppen:
The idea here is to unify the different ways in which we can currently set options on SourceKit-LSP in a scalable way: Environment variables, command line arguments to `sourcekit-lsp` and initialization options.

The idea is that a user can define a `~/.sourcekit-lsp/.sourcekit-lsp` file (we store logs in `~/.sourcekit-lsp/logs` on non-Darwin platforms), which will be used as the default configuration for all SourceKit-LSP instances. They can also place a `.sourcekit-lsp` file in the root of a workspace to configure SourceKit-LSP for that project specifically, eg. setting arguments that need to be passed to `swift build` for that project and which thus also need to be set on SourceKit-LSP.

For compatibility reasons, I’m mapping the existing command line options into the new options structure for now. I hope to delete the command line arguments in the future and solely rely on `.sourcekit-lsp` configuration files.

Environment variable will be migrated to `.sourcekit-lsp` in a follow-up commit.


ahoppen:
@swift-ci Please test

Documentation/Configuration File.md
```diff
@@ -0,0 +1,36 @@
+# `.sourcekit-lsp` Configuration File
+
+`.sourcekit-lsp` configuration files can be used to modify the behavior of SourceKit-LSP in various ways. The following locations are checked:
+- `~/.sourcekit-lsp/.sourcekit-lsp` (ie. a `.sourcekit-lsp` file in the `~/.sourcekit-lsp` directory)
+- A `.sourcekit-lsp` file in a workspace’s root
```
bnbarham, line 5: What about `.sourcekit-lsp/config` and require the directory in the workspace root as well? Seems plausible we may add more later. And IMO ideally we'd check `Library/org.swift.sourcekit-lsp/config` for macOS -> `$XDG_CONFIG_HOME/sourcekit-lsp/config` everywhere (though not sure what the standard is for Windows) -> `.sourcekit-lsp/config`.

I do wonder if we need to keep the server options still as well - consider experimental features that may only be supported by specific editors + a user that jumps between editors. Though arguably that case might not work even if you *could* configure them differently. This also makes it more difficult for editors to add support for specific options.

ahoppen: Good ideas. I modified the search locations slightly to 

- `~/.sourcekit-lsp/config.json`
- On macOS: `~/Library/Application Support/org.swift.sourcekit-lsp/config.json` from the various `Library` folders on the system
- If the `XDG_CONFIG_HOME` environment variable is set: `$XDG_CONFIG_HOME/org.swift.sourcekit-lsp/config.json`
- A `.sourcekit-lsp/config.json` file in a workspace’s root

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -989,27 +958,9 @@ extension SourceKitLSPServer {
...
-          break
-        case .some(.null):
-          self.options.completionOptions.maxResults = nil
-        case .some(.int(let maxResults)):
-          self.options.completionOptions.maxResults = maxResults
-        case .some(let invalid):
-          logger.error("Expected null or int for 'maxResults'; got \(String(reflecting: invalid))")
-        }
-      }
-    }
```
hamishknight, line 1008: I guess this isn't supported anymore?

Sources/LanguageServerProtocol/Requests/CompletionRequest.swift
```diff
@@ -36,18 +36,14 @@ public struct CompletionRequest: TextDocumentRequest, Hashable {
 
   public var context: CompletionContext?
 
-  public var sourcekitlspOptions: SKCompletionOptions?
-
   public init(
     textDocument: TextDocumentIdentifier,
     position: Position,
-    context: CompletionContext? = nil,
-    sourcekitlspOptions: SKCompletionOptions? = nil
```
hamishknight, line 45: There's a doc comment for this parameter above, I guess it should be removed?

hamishknight (approved):

Sources/SKCore/SourceKitLSPOptions.swift
```diff
@@ -0,0 +1,253 @@
...
+        maxCoresPercentageToUseForBackgroundIndexing: override?.maxCoresPercentageToUseForBackgroundIndexing
+          ?? base.maxCoresPercentageToUseForBackgroundIndexing,
+        updateIndexStoreTimeout: override?.updateIndexStoreTimeout ?? base.updateIndexStoreTimeout
+      )
+    }
+  }
+
+  public var swiftPM: SwiftPMOptions?
+  public var compilationDatabase: CompilationDatabaseOptions?
+  public var fallbackBuildSystem: FallbackBuildSystemOptions?
```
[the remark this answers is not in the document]
ahoppen: I tried that as well but then JSON decoding failed if the keys didn’t exist

Sources/SKCore/SourceKitLSPOptions.swift
```diff
@@ -0,0 +1,253 @@
...
+    /// Build configuration (debug|release).
+    public var configuration: BuildConfiguration?
+
+    /// Build artifacts directory path. If nil, the build system may choose a default value.
+    public var scratchPath: String?
+
+    public var cCompilerFlags: [String]?
+    public var cxxCompilerFlags: [String]?
+    public var swiftCompilerFlags: [String]?
+    public var linkerFlags: [String]?
```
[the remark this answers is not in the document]
ahoppen: Yes, there’s a difference between `nil` and empty here:

Say you have you global configuration file with `cCompilerFlags: ["-DFLAG"]`. And then there’s a workspace-level configuration file with `cCompilerFlags: []`. In that case `cCompilerFlags` for the workspace is empty. But if the workspace-level configuration file didn’t contain the `cCompilerFlags` key, then we’d inherit the global `cCompilerFlags` of `-DFLAG`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Yes, I’m not aware of any uses for it and I don’t see it as useful. We can always add it again if we should need it, at which point we probably wouldn’t make it in initialization option anymore either.

Sources/SKCore/SourceKitLSPOptions.swift
```diff
@@ -0,0 +1,253 @@
...
+    self.workDoneProgressDebounce = workDoneProgressDebounce
+  }
+
+  public init?(path: URL?) {
+    guard let path, let contents = try? String(contentsOf: path, encoding: .utf8) else {
+      return nil
+    }
+    guard
+      let decoded = orLog(
+        "Parsing .sourcekit-lsp",
```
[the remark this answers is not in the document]
ahoppen: Thanks for spotting. Forgot to rename.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKCore/SourceKitLSPOptions.swift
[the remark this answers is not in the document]
hamishknight: Ah yeah, I forgot Decodable doesn't utilize property initializers as default values :(

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):
IMO the XDG_CONFIG_HOME should be just sourcekit-lsp rather than reverse-domain. I'd also specify the configuration in LSP and then allow taking it in initializationOptions. But feel free to merge and fix in a follow up 👍

ahoppen:
@swift-ci Please test Windows

PR #1525, opened 2024-06-26
[6.0] Send `SIGKILL` to `swift-frontend` indexing processes

ahoppen:
- **Explanation**: We were sending `SIGINT` to `swift-frontend` processes if they didn’t terminate after 2 minutes. However, `swift-frontend` doesn’t listen to `SIGINT`.

If a task running `waitUntilExitStoppingProcessOnTaskCancellation` is cancelled and the process doesn’t terminate on a `SIGINT` after 2 seconds, kill it.
- **Scope**: Background indexing if a `swift-frontend` or `clang` process takes more than 2 minutes
- **Risk**: Low, we would mostly be stuck in indexing in these cases
- **Issue**: rdar://130103147
- **Reviewer**:   @hamishknight on https://github.com/swiftlang/sourcekit-lsp/pull/1500

### DIFF
--- a/Sources/SKSupport/Process+Run.swift
+++ b/Sources/SKSupport/Process+Run.swift
@@ -26,12 +26,30 @@ import WinSDK
 
 extension Process {
   /// Wait for the process to exit. If the task gets cancelled, during this time, send a `SIGINT` to the process.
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #1263, opened 2024-05-09
Manual re-indexing


ahoppen:
It might be good to provide a request to re-index the entire project. This would mark all units out-of-date and index all files with a high priority. This way such an option could also be used to speed up indexing.

rdar://127476221

PR #1526, opened 2024-06-27
[6.0] Add a request to re-index all files in SourceKit-LSP

ahoppen:
- **Explanation**: Users should not need to rely on this request. The index should always be updated automatically in the background. Having to invoke this request manes there is a bug in SourceKit-LSP's automatic re-indexing. It does, however, offer a workaround to re-index files when such a bug occurs where otherwise there would be no workaround.
- **Scope**: A new request, purely additive
- **Risk**: Low, purely additive
- **Issue**: rdar://127476221, #1263 
- **Reviewer**:   @bnbarham and @hamishknight on https://github.com/swiftlang/sourcekit-lsp/pull/1507

### DIFF
--- a/Documentation/LSP Extensions.md
+++ b/Documentation/LSP Extensions.md
@@ -431,3 +431,17 @@ New request that returns symbols for all the test classes and test methods withi
 ```ts
 export interface WorkspaceTestsParams {}
 ```
+
+## `workspace/triggerReindex`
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1527, opened 2024-06-27
[6.0] Translate the semantic token legend used by clangd to the semantic token legend used by SourceKit-LSP

ahoppen:
- **Explanation**: clangd uses a completely different semantic token legend than SourceKit-LSP (it doesn’t even adhere to the ordering of the pre-defined token types) but we were passing index offsets from clangd through assuming that clangd uses the same legend, which was incorrect.
When retrieving semantic tokens from clangd, translate the semantic tokens from clangd’s legend to SourceKit-LSP’s legend.
- **Scope**: Semantic tokens from clangd
- **Risk**: Low, semantic tokens for clangd are very broken without this change
- **Issue**: rdar://129895062
- **Reviewer**:   @hamishknight and @bnbarham  on https://github.com/swiftlang/sourcekit-lsp/pull/1509 and https://github.com/swiftlang/sourcekit-lsp/pull/1510

### DIFF
--- a/Documentation/LSP Extensions.md
+++ b/Documentation/LSP Extensions.md
@@ -35,7 +35,34 @@ Added field (this is an extension from clangd that SourceKit-LSP re-exposes):
 codeActions: CodeAction[]?
 ```
 
+## Semantic token modifiers
+
+Added the following cases from clangd
+
+```ts
+deduced = 'deduced'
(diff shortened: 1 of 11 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #1516, opened 2024-06-25
SourceKit-LSP is reporting wrong file for diagnostic related information


adam-fowler:
### Swift version

swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)

### Platform

macOS 14.5

### Editor

VS Code

### Does the issue reproduce with Swift 6?

Yes

### Description

The related information attached to diagnostics reference the wrong file. It references the file the diagnostic occurred and not the file the related information is in.

### Steps to Reproduce

1. In a VS Code project add file Test.swift with the following lines
```swift
@available(*, unavailable)
struct Test {}
```
2. In a second file Test2.swift reference type `Test`
```swift
func test() {
    let t = Test()
}
```
3. Notice diagnostic `'Test' is unavailable` and then notice that the related information `'Test' has been explicitly marked unavailable here` references a location in Test2.swift and not Test.swift

### Logging

Here is the related response from the SourceKit-LSP output log
```
[Trace - 15:28:16] Received response 'textDocument/diagnostic - (423)' in 5ms.
Result: {
    "kind": "full",
    "items": [
        {
            "tags": [
                1
            ],
            "range": {
                "start": {
                    "line": 1,
                    "character": 8
                },
                "end": {
                    "line": 1,
                    "character": 8
                }
            },
            "codeActions": [
                {
                    "title": "Replace 'let t' with '_'",
                    "kind": "quickfix",
                    "edit": {
                        "changes": {
                            "file:///Users/adamfowler/Temp/RelatedInformation/Sources/RelatedInformation/Test2.swift": [
                                {
                                    "newText": "_",
                                    "range": {
                                        "end": {
                                            "character": 9,
                                            "line": 1
                                        },
                                        "start": {
                                            "line": 1,
                                            "character": 4
                                        }
                                    }
                                }
                            ]
                        }
                    }
                }
            ],
            "source": "sourcekitd",
            "message": "initialization of immutable value 't' was never used; consider replacing with assignment to '_' or removing it",
            "severity": 2
        },
        {
            "relatedInformation": [
                {
                    "location": {
                        "uri": "file:///Users/adamfowler/Temp/RelatedInformation/Sources/RelatedInformation/Test2.swift",
                        "range": {
                            "start": {
                                "line": 1,
                                "character": 7
                            },
                            "end": {
                                "character": 7,
                                "line": 1
                            }
                        }
                    },
[...]

PR #1531, opened 2024-06-28
[6.0] Handle diagnostics in secondary files correctly

ahoppen:
* **Explanation**: sourcekitd sometimes returned notes in secondary source files but SourceKit-LSP assumed that all diagnostics and notes are in the primary source file, which was incorrect
* **Scope**: Diagnostics
* **Risk**: Should be fairly low, I don’t know how this would break something
* **Testing**: Added test case
* **Issue**: rdar://130503535, #1516 
* **Reviewer**:  @hamishknight on https://github.com/swiftlang/sourcekit-lsp/pull/1528


ahoppen:
@swift-ci Please test

PR #1533, opened 2024-06-28
Make background indexing a proper option in `SourceKitLSPOptions`

ahoppen:
This allows us to flip the default in the future more easily. It also allows users to disable background indexing when it’s enabled by default.

rdar://130280855


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1534, opened 2024-06-28
Allow specification of SourceKitLSPOptions in the initialize request and look for SourceKit-LSP options in `$XDG_CONFIG_HOME/sourcekit-lsp`

ahoppen:
Specifying `SourceKitLSPOptions` in the initialize request allows editors to provide UI elements to toggle SourceKit-LSP options.

The reverse-DNS notation is macOS style and doesn’t feel at home on Linux.

### DIFF
--- a/Sources/SKCore/SourceKitLSPOptions.swift
+++ b/Sources/SKCore/SourceKitLSPOptions.swift
@@ -12,7 +12,7 @@
 
 import Foundation
 import LSPLogging
-import SKCore
+import LanguageServerProtocol
 import SKSupport
 
 import struct TSCBasic.AbsolutePath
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SKCore/SourceKitLSPOptions.swift
```diff
@@ -200,6 +200,21 @@ public struct SourceKitLSPOptions: Sendable, Codable {
     self.workDoneProgressDebounce = workDoneProgressDebounce
   }
 
+  public init?(fromLSPAny lspAny: LSPAny?) throws {
```
bnbarham, line 203: Should we move this type to LSP so that clients can also construct it (though for that to actually be true we probably need to split LSP out of this repo, but still... seems like a reasonable place for it to live)?

bnbarham (approved):

Sources/SKCore/SourceKitLSPOptions.swift
ahoppen: I started moving it to the LanguageServerProtocol module and then decided against it. I can’t really tell why but it didn’t feel at home there.

PR #1542, opened 2024-07-01
feat: Add Semantic modifier to parameter labels

GrayJack:
Screenshot 2024-07-01 at 18 21 56

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/SemanticTokenModifiers.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/SemanticTokenModifiers.swift
@@ -47,6 +47,9 @@ public struct SemanticTokenModifiers: OptionSet, Hashable, Sendable {
   public static let fileScope = Self(rawValue: 1 << 19)
   public static let globalScope = Self(rawValue: 1 << 20)
 
+  // The following are our LSP extensions
+  public static let parameterLabel = Self(rawValue: 1 << 21)
+
   public var name: String? {
     switch self {
     case .declaration: return "declaration"
@@ -70,6 +73,7 @@ public struct SemanticTokenModifiers: OptionSet, Hashable, Sendable {
     case .classScope: return "classScope"
     case .fileScope: return "fileScope"
     case .globalScope: return "globalScope"
+    case .parameterLabel: return "parameterLabel"
     default: return nil
     }
   }
@@ -98,5 +102,6 @@ public struct SemanticTokenModifiers: OptionSet, Hashable, Sendable {
     .classScope,
     .fileScope,
     .globalScope,
+    .parameterLabel,
   ]
 }
--- a/Sources/SourceKitLSP/Swift/SemanticTokens.swift
+++ b/Sources/SourceKitLSP/Swift/SemanticTokens.swift
@@ -150,7 +150,7 @@ extension SyntaxClassification {
     case .docLineComment, .docBlockComment:
       return (.comment, .documentation)
     case .argumentLabel:
-      return (.function, [])
+        return (.function, [.parameterLabel])
     }
   }
 }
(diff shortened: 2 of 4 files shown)


Sources/LanguageServerProtocol/SupportTypes/SemanticTokenModifiers.swift
```diff
@@ -47,6 +47,9 @@ public struct SemanticTokenModifiers: OptionSet, Hashable, Sendable {
   public static let fileScope = Self(rawValue: 1 << 19)
   public static let globalScope = Self(rawValue: 1 << 20)
 
+  // The following are our LSP extensions
+  public static let parameterLabel = Self(rawValue: 1 << 21)
```
ahoppen, line 51: - Could you mark this as `/// **(LSP Extension)**` to be consistent with other LSP extensions we introduce?
- Could you add it to the `LSP Extensions.md` document? 
- I think we should call this `argumentLabel` to be consistent with the notion in swift-syntax.

```suggestion
  /// Argument labels in function definitions and function calls
  ///
  /// **(LSP Extension)**
  public static let argumentLabel = Self(rawValue: 1 << 21)
```

Sources/SourceKitLSP/Swift/SemanticTokens.swift
```diff
@@ -150,7 +150,7 @@ extension SyntaxClassification {
     case .docLineComment, .docBlockComment:
       return (.comment, .documentation)
     case .argumentLabel:
-      return (.function, [])
+        return (.function, [.parameterLabel])
```
ahoppen, line 153: Could you indent this using two spaces?

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -335,7 +335,7 @@ final class SemanticTokensTests: XCTestCase {
       expected: [
         TokenSpec(marker: "1️⃣", length: 4, kind: .keyword),
         TokenSpec(marker: "2️⃣", length: 1, kind: .identifier),
-        TokenSpec(marker: "3️⃣", length: 1, kind: .function),
+        TokenSpec(marker: "3️⃣", length: 1, kind: .function, modifiers: .parameterLabel),
```
ahoppen, line 338: Do we also add the `parameterLabel` modifier to labels in function calls? Shouldn’t your change also affect the `SemanticTokenTests.testArgumentLabels` test?

ahoppen (commented):
Thanks for the PR, @GrayJack 🙏🏽

Tests/SourceKitLSPTests/SemanticTokensTests.swift
GrayJack: Yes, as it is also adding `function` token on the function call as well.

Oh, I missed that test, I'll update

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -905,7 +905,7 @@ final class SemanticTokensTests: XCTestCase {
       expected: [
         TokenSpec(marker: "1️⃣", length: 4, kind: .keyword),
         TokenSpec(marker: "2️⃣", length: 3, kind: .identifier),
-        TokenSpec(marker: "3️⃣", length: 3, kind: .function),
+        TokenSpec(marker: "3️⃣", length: 3, kind: .function, modifiers: .argumentLabel),
         TokenSpec(marker: "4️⃣", length: 3, kind: .struct, modifiers: .defaultLibrary),
         TokenSpec(marker: "5️⃣", length: 3, kind: .function),
         TokenSpec(marker: "6️⃣", length: 3, kind: .function),
```
ahoppen, line 911: Shouldn’t this token also have a `.argumentLabel` modifier? 

Also in case you see the test as being skipped on your machine, you need to set `SOURCEKIT_TOOLCHAIN_PATH` to a Swift Development Snapshot as described here: https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#selecting-a-toolchain

PR #1547, opened 2024-07-02
[6.0] Add a maximum duration for sourcekitd requests

ahoppen:
- **Explanation**: VS Code does not cancel semantic tokens requests. If a source file gets into a state where an AST build takes very long, this can cause us to wait for the semantic tokens from sourcekitd for a few minutes, effectively blocking all other semantic functionality in that file.
To circumvent this problem (or any other problem where an editor might not be cancelling requests they are no longer interested in) add a maximum request duration for SourceKitD requests, defaulting to 2 minutes.
- **Scope**: sourcekitd requests that take more than 2 minutes
- **Risk**: Low, no reasonable sourcekitd request should take more than 2 minutes and users can override the timeout duration as a workaround
- **Issue**: rdar://130948453
- **Reviewer**:  @hamishknight on https://github.com/swiftlang/sourcekit-lsp/pull/1543

### DIFF
--- a/Sources/Diagnose/RunSourcekitdRequestCommand.swift
+++ b/Sources/Diagnose/RunSourcekitdRequestCommand.swift
@@ -96,6 +96,9 @@ public struct RunSourceKitdRequestCommand: AsyncParsableCommand {
     case .requestCancelled:
       print("request cancelled")
       throw ExitCode(1)
+    case .timedOut:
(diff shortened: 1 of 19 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1550, opened 2024-07-03
Import new Android overlay

finagolfin:
This was recently added in swiftlang/swift#72161 and swiftlang/swift#72634. I've been building it with this patch on my daily Android CI since then, finagolfin/swift-android-sdk#151, and natively on Android too.

@ahoppen, should be an easy review.


ahoppen (approved):
Thanks! LGTM

ahoppen:
@swift-ci Please test

PR #1551, opened 2024-07-03
Convert String Concatenation to String Interpolation

### LINKED ISSUE #1244: Code action to replace string concatenation with string interpolation
### Description

We should have a code action that replaces string concatenation, e.g.

```swift
"[" + key + ": " + value + "]"
```

with interpolation:

```swift
"[\(key): \(value)]"
```

AppAppWorks:
fixed #1244 

It's only a draft as the support for multiline string literals is not yet implemented and more tests have to be written, I also wonder if it's a good idea to also perform the action in `FormatRawStringLiteral` when performing this conversion.

### DIFF
--- a/Sources/SourceKitLSP/CMakeLists.txt
+++ b/Sources/SourceKitLSP/CMakeLists.txt
@@ -31,6 +31,7 @@ target_sources(SourceKitLSP PRIVATE
   Swift/CodeActions/AddDocumentation.swift
   Swift/CodeActions/ConvertIntegerLiteral.swift
   Swift/CodeActions/ConvertJSONToCodableStruct.swift
+  Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
   Swift/CodeActions/PackageManifestEdits.swift
   Swift/CodeActions/SyntaxCodeActionProvider.swift
   Swift/CodeActions/SyntaxCodeActions.swift
--- a/Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
+++ b/Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
@@ -1,0 +1,224 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SwiftRefactor
+import SwiftSyntax
+
+/// ConvertStringConcatenationToStringInterpolation is a code action that converts a valid string concatenation into a
+/// string interpolation.
+struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvider {
+  static func refactor(syntax: SequenceExprSyntax, in context: Void) -> SequenceExprSyntax? {
+    guard let (componentsOnly, commonPounds) = preflight(exprList: syntax.elements) else {
+      return nil
+    }
+
+    var segments: StringLiteralSegmentListSyntax = []
+    for component in componentsOnly {
+      guard let stringLiteral = component.as(StringLiteralExprSyntax.self) else {
+        segments.append(
+          .expressionSegment(
+            ExpressionSegmentSyntax(
+              pounds: commonPounds,
+              expressions: [
+                LabeledExprSyntax(expression: component.singleLineTrivia)
+              ]
+            )
+          )
+        )
+        continue
+      }
+
+      if let commonPounds, stringLiteral.openingPounds?.tokenKind != commonPounds.tokenKind {
+        segments += stringLiteral.segments.map { segment in
+          if case let .expressionSegment(exprSegment) = segment {
+            .expressionSegment(exprSegment.with(\.pounds, commonPounds))
+          } else {
+            segment
+          }
+        }
+      } else {
+        segments += stringLiteral.segments
+      }
+    }
+
+    return syntax.with(
+      \.elements,
+      [
+        ExprSyntax(
+          StringLiteralExprSyntax(
+            openingPounds: commonPounds,
+            openingQuote: .stringQuoteToken(),
+            segments: segments,
+            closingQuote: .stringQuoteToken(),
+            closingPounds: commonPounds
+          )
+        )
+      ]
+    )
+  }
+
+  /// If `exprList` is a valid string concatenation, returns 1) all elements in `exprList` with concat operators
+  /// stripped and 2) the longest pounds amongst all string literals, otherwise returns nil.
+  ///
+  /// `exprList` as a valid string concatenation must contain n >= 3 children where n is an odd number with a concat
+  /// operator `+` separating every other child, which must either be a single-line string literal or a valid
+  /// expression for string interpolation. `exprList` must also contain at least one string literal child.
+  ///
+  /// The following is a valid string concatenation.
+  /// ``` swift
+  /// "Hello " + aString + "\(1)World"
+  /// ```
+  /// The following are invalid string concatenations.
+  /// ``` swift
+  /// aString + bString // no string literals
+  ///
+  /// "Hello " * aString - "World" // non `+` operators
+  ///
+  /// """
+  /// Hello
+  /// """
+  /// + """
+  /// World
+  /// """ // multi-line string literals
+  /// ```
+  private static func preflight(
+    exprList: ExprListSyntax
+  ) -> (componentsOnly: [ExprListSyntax.Element], longestPounds: TokenSyntax?)? {
+    var iter = exprList.makeIterator()
+    guard let first = iter.next() else {
+      return nil
+    }
+
+    var hasStringComponents = false
+    var longestPounds: TokenSyntax?
+    var componentsOnly = [ExprListSyntax.Element]()
+    componentsOnly.reserveCapacity(exprList.count / 2 + 1)
+
+    if let stringLiteral = first.as(StringLiteralExprSyntax.self) {
+      guard stringLiteral.isSingleLine else {
+        return nil
+      }
+      hasStringComponents = true
+      longestPounds = stringLiteral.openingPounds
+    }
+    componentsOnly.append(first)
+
+    while let concat = iter.next(), let stringComponent = iter.next() {
+      guard let concat = concat.as(BinaryOperatorExprSyntax.self),
+        concat.operator.tokenKind == .binaryOperator("+") && !stringComponent.is(MissingExprSyntax.self)
+      else {
+        return nil
+      }
+
+      if let stringLiteral = stringComponent.as(StringLiteralExprSyntax.self) {
+        guard stringLiteral.isSingleLine else {
+          return nil
+        }
+        hasStringComponents = true
+        if let pounds = stringLiteral.openingPounds,
+          pounds.trimmedLength > (longestPounds?.trimmedLength ?? SourceLength(utf8Length: 0))
+        {
+          longestPounds = pounds
+        }
+      }
+
+      componentsOnly[componentsOnly.count - 1].trailingTrivia += concat.leadingTrivia
+      componentsOnly.append(
+        stringComponent.with(\.leadingTrivia, stringComponent.leadingTrivia + concat.trailingTrivia)
+      )
+    }
+
+    guard hasStringComponents && componentsOnly.count > 1 else {
+      return nil
+    }
+
+    return (componentsOnly, longestPounds)
+  }
+}
+
+extension ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringCodeActionProvider {
+  static let title: String = "Convert String Concatenation to String Interpolation"
+
+  static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> SequenceExprSyntax? {
+    guard let expr = scope.innermostNodeContainingRange,
+      let seqExpr = expr.findParentOfSelf(
+        ofType: SequenceExprSyntax.self,
+        stoppingIf: {
+          $0.kind == .codeBlockItem || $0.kind == .memberBlockItem
+        }
+      )
+    else {
+      return nil
+    }
+
+    return seqExpr
+  }
+}
+
+private extension String {
+  var uncommented: Substring {
+    trimmingPrefix { $0 == "/" }
+  }
+}
+
+private extension StringLiteralExprSyntax {
+  var isSingleLine: Bool {
+    openingQuote.tokenKind == .stringQuote
+  }
+}
+
+private extension SyntaxProtocol {
+  /// Modifies the trivia to not contain any newlines. This removes whitespace trivia, replaces newlines with
+  /// whitespaces in block comments and converts line comments to block comments.
+  var singleLineTrivia: Self {
+    with(\.leadingTrivia, leadingTrivia.withSingleLineComments.withCommentsOnly(isLeadingTrivia: true))
+      .with(\.trailingTrivia, trailingTrivia.withSingleLineComments.withCommentsOnly(isLeadingTrivia: false))
+  }
+}
+
+private extension Trivia {
+  /// Replaces newlines with whitespaces in block comments and converts line comments to block comments.
+  var withSingleLineComments: Self {
+    Trivia(
+      pieces: map {
+        switch $0 {
+        case let .lineComment(lineComment):
+          .blockComment("/*\(lineComment.uncommented)*/")
+        case let .docLineComment(docLineComment):
+          .docBlockComment("/**\(docLineComment.uncommented)*/")
+        case let .blockComment(blockComment), let .docBlockComment(blockComment):
+          .blockComment(blockComment.replacing("\r\n", with: " ").replacing("\n", with: " "))
+        default:
+          $0
+        }
+      }
+    )
+  }
+
+  /// Removes all non-comment trivia pieces and inserts a whitespace between each comment.
+  func withCommentsOnly(isLeadingTrivia: Bool) -> Self {
+    Trivia(
+      pieces: flatMap { piece -> [TriviaPiece] in
+        if piece.isComment {
+          if isLeadingTrivia {
+            [piece, .spaces(1)]
+          } else {
+            [.spaces(1), piece]
+          }
+        } else {
+          []
+        }
+      }
+    )
+  }
+}
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1006,6 +1006,105 @@ final class CodeActionTests: XCTestCase {
     }
   }
 
+  func testConvertStringConcatenationToStringInterpolation() async throws {
+    try await assertCodeActions(
+      #"""
+      1️⃣#"["# + 2️⃣key + ": \(3️⃣d) " + 4️⃣value + ##"]"##5️⃣
+      """#,
+      ranges: [("1️⃣", "2️⃣"), ("3️⃣", "4️⃣"), ("1️⃣", "5️⃣")],
+      exhaustive: false
+    ) { uri, positions in
+      [
+        CodeAction(
+          title: "Convert String Concatenation to String Interpolation",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["1️⃣"]..<positions["5️⃣"],
+                  newText: ###"""
+                    ##"[\##(key): \##(d) \##(value)]"##
+                    """###
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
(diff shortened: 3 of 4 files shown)


Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1005,6 +1005,33 @@ final class CodeActionTests: XCTestCase {
...
+                TextEdit(
+                  range: positions["1️⃣"]..<positions["5️⃣"],
+                  newText: "##\"[\\##(key): \\##(d) \\##(value)]\"##"
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
```
ahoppen, line 1033: Could you also add a test case to check that we don’t show this refactoring when there’s a standalone string literal.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+//  Created by Lau Chun Kai on 3/7/2024.
+//
+
+import LanguageServerProtocol
+import SwiftRefactor
+import SwiftSyntax
+
+struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvider {
+  static func refactor(syntax: ExprListSyntax, in context: Void) -> ExprListSyntax? {
```
ahoppen, line 13: I would prefer if the refactoring operated primarily on `SequenceExprSyntax` instead of `ExprListSyntax`. Currently `ExprListSyntax` is only used inside `SequenceExprSyntax` but that might change in the future, while `SequenceExprSyntax` is the dedicated node for this kind of expression + operator list.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,138 @@
...
+    
+    guard hasStringLiterals else {
+      return nil
+    }
+    
+    let commonPounds: TokenSyntax? = if let longestPoundsText {
+      .rawStringPoundDelimiter(longestPoundsText)
+    } else {
+      nil
+    }
```
ahoppen, line 102: Could we avoid this if `longestPoundsText` was of type `TokenSyntax?` from the start?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1005,6 +1005,33 @@ final class CodeActionTests: XCTestCase {
...
+      [
+        CodeAction(
+          title: "Convert String Concatenation to String Interpolation",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["1️⃣"]..<positions["5️⃣"],
+                  newText: "##\"[\\##(key): \\##(d) \\##(value)]\"##"
```
ahoppen, line 1026: I think this would be easier to read if you made it a multi-line string literal, because then you don’t need to escape the `"`.

ahoppen (commented):
Thanks for adding the refactoring @AppAppWorks.

Depending on how you want to approach it, you can add support for multi-line string literals in this PR or make the refactoring unavailable if the sequence expression contains a multi-line string literal and then think about that case in a follow-up PR.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,138 @@
...
+struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvider {
+  static func refactor(syntax: ExprListSyntax, in context: Void) -> ExprListSyntax? {
+    guard let (componentsOnly, commonPounds) = syntax.preflight() else {
+      return nil
+    }
+    
+    var ret: StringLiteralSegmentListSyntax = []
+    for component in componentsOnly {
+      if var stringLiteral = StringLiteralExprSyntax(component) {
+        stringLiteral.pounds = commonPounds
```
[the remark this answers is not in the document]
AppAppWorks: Setting the pounds here also sets the pounds for all interpolations inside the literal, maybe I should refactor the setter of `StringLiteralExprSyntax.pounds` to make the intent clearer?

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,138 @@
+//
+//  ConvertStringConcatenationToStringInterpolation.swift
+//  
+//
+//  Created by Lau Chun Kai on 3/7/2024.
+//
```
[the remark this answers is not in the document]
AppAppWorks: Sure, that was an overlook

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
AppAppWorks: I feel like I was bikeshedding between `SequenceExprSyntax` and `ExprListSyntax`, thanks for your explanation!

AppAppWorks:
@ahoppen please check again :)

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1006,6 +1006,64 @@ final class CodeActionTests: XCTestCase {
...
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testConvertStringConcatenationToStringInterpolationNotShowUp() async throws {
+    //    ""
```
ahoppen, line 1039: I think we can simplify this test case to only contain a simple string literals without pounds or string interpolation so that it focuses on one thing and is thus easier to reason about.

---

Leftover comment?

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+  ///  For example,
+  ///
+  ///      "Hello " + aString + "\(1)World"
+  ///  is a valid string concatenation.
+  ///
+  ///      aString + bString
+  ///  and
+  ///
+  ///      "Hello " * aString - "World"
+  ///  are invalid string concatenations.
```
ahoppen, line 78: I know that swift-format doesn’t enforce it but could you try to keep the comment length < 120 columns?

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  /// If `exprList` is a string concatenation, returns 1) all elements in `exprList` with concat operators stripped and 2) the longest pounds amongst all string literals,
+  /// otherwise returns nil.
+  ///
+  ///  `exprList` as a valid string concatenation must contain n >= 3 children where n is an odd number with a concat operator `+` separating every other child which must either be a string literal or a valid expression for string interpolation. `exprList` must also contain at least one string literal child.
+  ///
+  ///  For example,
+  ///
+  ///      "Hello " + aString + "\(1)World"
```
ahoppen, line 71: Let’s use markdown to format these code snippets.

```suggestion
  /// ```swift
  /// "Hello " + aString + "\(1)World"
  /// ```
```

Same below

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+  ///  is a valid string concatenation.
+  ///
+  ///      aString + bString
+  ///  and
+  ///
+  ///      "Hello " * aString - "World"
+  ///  are invalid string concatenations.
+  static func preflight(
```
ahoppen, line 79: This could be `private`, right?

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+    var longestPounds: TokenSyntax?
+    var componentsOnly = [first]
+
+    if let stringLiteral = StringLiteralExprSyntax(first) {
+      hasStringLiterals = true
+      longestPounds = stringLiteral.openingPounds
+    }
+
+    while let concat = iter.next(), let stringComponent = iter.next() {
+      guard let concat = BinaryOperatorExprSyntax(concat), case .binaryOperator("+") = concat.operator.tokenKind else {
```
ahoppen, line 97: A little nitpicky but `TokenKind` is equatable, so we don’t need to use `case` here, I think. I just think that `==` reads nicer. Also, we generally prefer to use `as` for downcasting.

```suggestion
      guard let concat = concat.as(BinaryOperatorExprSyntax.self), concat.operator.tokenKind == .binaryOperator("+") else {
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+        return nil
+      }
+
+      if let stringLiteral = StringLiteralExprSyntax(stringComponent) {
```
ahoppen, line 101:
```suggestion
      if let stringLiteral = stringComponent.as(StringLiteralExprSyntax.self) {
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,139 @@
...
+    var iter = exprList.makeIterator()
+    guard let first = iter.next() else {
+      return nil
+    }
+
+    var hasStringLiterals = false
+    var longestPounds: TokenSyntax?
+    var componentsOnly = [first]
+
+    if let stringLiteral = StringLiteralExprSyntax(first) {
```
ahoppen, line 91:
```suggestion
    if let stringLiteral = first.as(StringLiteralExprSyntax.self) {
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1006,6 +1006,64 @@ final class CodeActionTests: XCTestCase {
     }
   }
 
+  func testConvertStringConcatenationToStringInterpolation() async throws {
+    try await assertCodeActions(
+      """
+      1️⃣#"["# + 2️⃣key + ": \\(3️⃣d) " + 4️⃣value + ##"]"##5️⃣
+      """,
```
ahoppen, line 1013: If you use a raw multiline string literal (with `#"""`), you shouldn’t need to escape the `\`, which should make the test easier to reason about. Same in the expected new text.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1006,6 +1006,64 @@ final class CodeActionTests: XCTestCase {
...
+                    """
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
```
ahoppen, line 1066: Could you also add a test case for multi-line string literals? I’m fine with it either not showing up the code action (I think string concatenations with multi-line string literals are less common than with single-line string literals) or it working properly. I just want to make sure that multi-line string literals don’t produce an invalid refactoring result.

ahoppen (commented):
Very nice

AppAppWorks:
Could a follow-up action be converting concatenation of `(some CustomStringConvertible).description` into string interpolation?

e.g. `1.description + false.description` -> `"\(1)\(false)"`

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,184 @@
...
+}
+
+fileprivate extension StringLiteralExprSyntax {
+  var isSingleLine: Bool {
+    openingQuote.tokenKind == .stringQuote
+  }
+
+  func isPoundsNotEqual(to otherPounds: TokenSyntax) -> Bool {
+    openingPounds?.tokenKind != otherPounds.tokenKind
+  }
```
ahoppen, line 173: I think you’re only using this once and the implementation is quite simple as well, so I’d just inline it. I feel like it’s easier to understand the code that way because I don’t need to look up what `isPoundsNotEqual` does.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,184 @@
...
+  ///
+  ///"Hello " * aString - "World" // non `+` operators
+  ///
+  ///"""
+  ///Hello
+  ///"""
+  ///+ """
+  ///World
+  ///""" // multi-line string literals
+  ///```
```
ahoppen, line 92: Seems like you’re missing some whitespace for the code blocks to make them look nice.

```suggestion
  /// "Hello " + aString + "\(1)World"
  /// ```
  ///  The following are invalid string concatenations.
  /// ``` swift
  /// aString + bString // no string literals
  ///
  /// "Hello " * aString - "World" // non `+` operators
  ///
  /// """
  /// Hello
  /// """
  /// + """
  /// World
  /// """ // multi-line string literals
  /// ```
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,184 @@
...
+      guard stringLiteral.isSingleLine else {
+        return nil
+      }
+      hasStringLiterals = true
+      longestPounds = stringLiteral.openingPounds
+    }
+
+    while let concat = iter.next(), let stringComponent = iter.next() {
+      guard let concat = concat.as(BinaryOperatorExprSyntax.self),
+        concat.isConcat && !stringComponent.is(MissingExprSyntax.self)
```
ahoppen, line 115: Do you think that `isConcat` improves readability here? I personally think it’s easier to understand if we just inlined `isConcat` because it gives the visual cue that we’re checking for `+` here. I personally needed to check what `isConcat` does to fully understand it.

```suggestion
        concat.operator.tokenKind == .binaryOperator("+") && !stringComponent.is(MissingExprSyntax.self)
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,184 @@
...
+      else {
+        return nil
+      }
+
+      if let stringLiteral = stringComponent.as(StringLiteralExprSyntax.self) {
+        guard stringLiteral.isSingleLine else {
+          return nil
+        }
+        hasStringLiterals = true
+        if let pounds = stringLiteral.openingPounds, pounds.isLongerThanAsTrimmed(longestPounds) {
```
ahoppen, line 125: Similarly here, I think it’s actually more readable if we inlined `isLongerThanAsTrimmed`

```suggestion
        if let pounds = stringLiteral.openingPounds, pounds.trimmedLength > (longestPounds.trimmedLength ?? 0) {
```

ahoppen (commented):
Looking great! I have a few minor comments that would make the code more readable in my opinion. After that, we’re ready to go.

ahoppen:
> Could a follow-up action be converting concatenation of `(some CustomStringConvertible).description` into string interpolation?
> 
> e.g. `1.description + false.description` -> `"\(1)\(false)"`

I personally wouldn’t prioritize it because I don’t think it’s a pattern that’s used terribly often but I also wouldn’t object to a PR that implements it.

AppAppWorks:
All done, please have a look.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+  ///
+  /// "Hello " * aString - "World" // non `+` operators
+  ///
+  /// """
+  /// Hello
+  /// """
+  /// + """
+  /// World
+  /// """ // multi-line string literals
+  ///```
```
ahoppen, line 93:
```suggestion
  /// ```
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+      .with(\.trailingTrivia, trailingTrivia.inlineComments.commentsOnly(isLeadingTrivia: false))
+  }
+}
+
+fileprivate extension Trivia {
+  var inlineComments: Self {
+    Trivia(
+      pieces: map {
+        switch $0 {
+        case let .lineComment(lineComment):
```
ahoppen, line 186: Could you also cover `.docLineComment` here?

Instead of stripping off the first two characters, I would drop the leading `/` characters.

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+          .blockComment("/*\(lineComment.dropFirst(2))*/")
+        case let .blockComment(blockComment):
```
ahoppen, line 188: Same here, could you also cover `.docBlockComment`?

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+          .blockComment(blockComment.replacing(#/\n+/#, with: { match in String(repeating: " ", count: match.count) }))
```
ahoppen, line 189: I don’t think we need regex here, we can just use normal string replacement. Also, we should handle Windows-style CRLF.

```suggestion
          .blockComment(blockComment.replacing("\r\n", with: " ").replacing("\n", with: " "))
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+}
+
+fileprivate extension StringLiteralExprSyntax {
+  var isSingleLine: Bool {
+    openingQuote.tokenKind == .stringQuote
+  }
+}
+
+fileprivate extension SyntaxProtocol {
+  var commentsInlined: Self {
```
ahoppen, line 175: I think the primary action here is that the trivia gets flattened to a single line.

```suggestion
  /// Modifies the trivia to not contain any newlines. This removes whitespace trivia, removes newlines from block comments and converts line comments to block comments.
  var withSingleLineTrivia: Self {
```

Sources/SourceKitLSP/Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -0,0 +1,212 @@
...
+    with(\.leadingTrivia, leadingTrivia.inlineComments.commentsOnly(isLeadingTrivia: true))
+      .with(\.trailingTrivia, trailingTrivia.inlineComments.commentsOnly(isLeadingTrivia: false))
+  }
+}
+
+fileprivate extension Trivia {
+  var inlineComments: Self {
```
ahoppen, line 182: Similar to my comment above, I would rename this to `withSingleLineComments`

ahoppen (commented):
Nice catch with the multi-line comments.

AppAppWorks:
Revised, please check again!

ahoppen (approved):
Thank you for implementing the refactoring @AppAppWorks. Looks great!

Would you mind filing an issue to also support multi-line string literals. Just so we don’t forget about that as a follow-up thing that can be done.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ target_sources(SourceKitLSP PRIVATE
   Swift/CodeActions/AddDocumentation.swift
   Swift/CodeActions/ConvertIntegerLiteral.swift
   Swift/CodeActions/ConvertJSONToCodableStruct.swift
+  Swift/CodeActions/ConvertStringConcatenationToStringInterpolation.swift,
```
ahoppen, line 34: I think this the Windows build is failing because this shouldn’t have a comma.

AppAppWorks: My bad, I've pushed a new commit, please test again.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1552, opened 2024-07-03
Apply DeMorgan's Law

### LINKED ISSUE #1243: Code action to apply DeMorgan's law
### Description

Sometimes in code we end up with something like:

```swift
if !x && y != z { ... }
```

and it would be cleaner to apply DeMorgan's law to make this:

```swift
if !(x || y == z) { ... }
```

We should have a code action to do this correctly, which requires dealing with inverting conditions and dealing with operator precedence. The `SwiftOperators` module of swift-syntax can help here.

AppAppWorks:
fixed #1243


Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,551 @@
...
+        if let falsyExpr = PrefixOperatorExprSyntax(expr), falsyExpr.tokenKindText == negationPrefix {
+          return falsyExpr.expression
+        } else if var infixOptrExpr = InfixOperatorExprSyntax(expr),
+          var binaryOptrExpr = BinaryOperatorExprSyntax(infixOptrExpr.operator)
+        {
+          switch binaryOptrExpr.tokenKindText {
+          case "!=":
+            binaryOptrExpr.tokenKindText = "=="
+          case "!==":
+            binaryOptrExpr.tokenKindText = "==="
```
ahoppen, line 384: Should we also handle `<`, `<=`, `>` and `>=`?

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,551 @@
...
+    set(value) {
+      `operator`.tokenKind = .binaryOperator(value)
+    }
+  }
+
+  func appliedDeMorgan() -> ExprSyntax? {
+    let negationPrefix = tokenKindText
+    guard negationPrefix == "!" || negationPrefix == "~" else {
+      return nil
+    }
```
ahoppen, line 500: Instead of passing arbitrary strings around for the `negationPrefix`, would it make sense to declare an enum for the two kinds of negation prefixes that are supported?

AppAppWorks:
> Thanks for tackling this @AppAppWorks. I have left a first set of review comments inline. Correctness-wise, I think it’s mostly missing the addition of parentheses where precedence of operators flips, eg. `!(a && b || c)` should get transformed to `(!a || !b) && !c`.

This example involves a recursive DeMorgan's Law application that has not been implemented in this PR?
i.e. `!(a && b || c)` -> `!(a && b) && !c` -> `(!a || !b) && !c`

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,551 @@
...
+    )
+  }
+
+  func negatedToTruthy(
+    negationPrefix: String,
+    binaryOptrExprText: String,
+    reversedBinaryOptrExprText: String
+  )
+    -> ExprSyntax?
+  {
```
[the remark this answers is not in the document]
AppAppWorks: In my current design, only these two conditions will warrant the code action,
1. `!(expr_1 && expr_2 && ... expr_n)` or `!(expr_1 || expr_2 || ... expr_n)`
2. `neg_expr_1 && neg_expr_2 && ... neg_expr_n` or `neg_expr1 || neg_expr_2 || ... neg_expr_n`

So are you suggesting I should relax the conditions to also include,
3. `!([neg_]expr_1 && [neg_]expr_2 && ... [neg_]expr_n)` or `!([neg_]expr_1 || [neg_]expr_2 || ... [neg_]expr_n)`
4. `[neg_]expr_1 && [neg_]expr_2 && ... [neg_]expr_n` or `[neg_]expr1 || [neg_]expr_2 || ... [neg_]expr_n`

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,551 @@
...
+      } else {
+        nil
+      }
+    }
+  }
+
+  /// Recursively walks down its syntax tree if this is a tuple and has a tuple as its single child.
+  ///
+  /// For example, exposes `a` from `((((a))))` for manipulation and returns `((((b))))`if `a` is transformed to `b`.
+  func untuplify(transform: (ExprSyntax) throws -> (some ExprSyntaxProtocol)?) rethrows -> ExprSyntax? {
```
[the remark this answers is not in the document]
AppAppWorks: Taking a page from monadic operations, I suggest the following renaming,

1. `untuplify` -> `mapNested1Tuple`
2. `untuplifyWithNegationHoisted` -> `flatMapNested1TupleNegating`
3.  `untuplifyWithNegationSeperated` -> `mapNested1TupleNegating`

I'd like to further change `1Tuple` into `Single` if this wouldn't obscure the function intent.

ahoppen:
Do you mean recursive because a boolean expression is nested inside a `!`? I think support for these kinds of expressions is mandator for a DeMorgan code action.

AppAppWorks:
Do you mean that whenever the code action results in any child expression like `!(...)`, we should apply DeMorgan to such child expression recursively until no `!(...)` is found as a result of any DeMorgan action?

e.g. `!(((a && b) || c) || d)` -> `!((a && b) || c) && !d` -> `!(a && b) && !c && !d` -> `(!a || !b) && !c && !d`

Or do you mean that the code action will also apply recursively to all DeMorgan-able descendent expressions of the outermost DeMorgan-able expressions?

e.g. `!(!(!(a && b) || c) || d)` -> `(!(a && b) || c) && !d` -> `(!a || !b || c) && !d`

ahoppen:
I would expect the former. My logic would be:
- When applying DeMorgan, add a `!` to the target expression to convert (or remove it if `!` already exists), then recursively DeMorgan-negate the children
- DeMorgan-negating an expression prefixed with `!` will just drop the `!` because that simplifies the code
- DeMorgan-negating a binary operator flips the operator and then DeMorgan-negates the operands.

With those rules, the transformation of your second example would be

`!(!(!(a && b) || c) || d)` -> `!!(!(a && b) || c) && !d` -> `(!(a && b) || c) && !d`

AppAppWorks:
The latest commit is almost a complete rewrite of the code action. I've adopted a more succinct naming convention, simplified code paths, and covered more edge cases such as adding parentheses when precedence orders are flipped after conversion, and expansion on ternary expressions.

AppAppWorks:
By the way I wonder whether there is any determining factor for creating a code action under `SourceKitLSP` vs a refactoring action under `SwiftRefactor`?

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+    /// // expression of interest
+    /// !a || b is String
+    /// !a /* negated to and mapped to */ a
+    /// b is String /* negated to */ Negation(b is String) /* mapped to */ !(b is String)
+    /// // negated expression of interest
+    /// a && !(b is String)
+    /// // mapped to
+    /// !((a && !(b is String)))
+    /// ```
+    func map(negation: Negation) -> Negated {
```
ahoppen, line 411: `map` seems like a pretty ambiguous term here but I’m struggling to find anything more meaningful. My feeling is also that the `negation` parameter and `self` duplicate some information. Just a thought: Could `Negation` just store the underlying `PrefixOperatorExprSyntax` that it was created from and then offer `exprType` and `expr` as computed properties (or stored properties if that’s better for performance)?

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+  /// only the expression of interest for easier manipulation.
+  private enum Expr {
+    /// A single means a tuple containing only one element, implying that there exists at least one pair of extraneous
+    /// parentheses enclosing the expression of interest.
+    ///
+    /// - Parameters:
+    ///   - desinglified: The first non-single descendent expression. The expression of interest.
+    case single(TupleExprSyntax, desinglified: ExprSyntax)
+    /// A non-single expression at its root. The expression of interest.
+    case other(ExprSyntax)
```
ahoppen, line 212: Do we really have to have an enum with these two cases? Wouldn’t it be easier to just operate on `ExprSyntax` and have a property that looks through all parentheses? Something like 

```swift
extension ExprSyntax {
  // Equivalent of the `desinglified` associated value in `single`
  var lookingThroughParentheses: ExprSyntax {
    if let tuple = self.as(TupleExprSyntax.self), 
      let underlyingExpr = tuple.singleUnlabledExpression?.lookingThroughParentheses else {
      return underlyingExpr
    }
    return self
  }

  // Equivalent of checking for the `single` case.
  var isParenthesizedExpression: Bool {
    return self.as(TupleExprSyntax.self)?.isParentheses ?? false
  }
}
```

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+    /// `Change.orToAnd` or `Change.andToOr`.
+    ///
+    /// - Note: A valid instance of propositions must have its binary operator agree with `exprType`.
+    func negatedPropositions(exprType: ExprType) -> Negated? {
+      if let infixExpr = self.as(InfixOperatorExprSyntax.self),
+        let negatedPropositions = NegatedPropositions(infixExpr, exprType: exprType)
+      {
+        (
+          self.map(negatedExpr: negatedPropositions.expr),
+          negatedPropositions.operator.kind == .and ? .orToAnd : .andToOr
```
ahoppen, line 306: Can you add explicit `return` here and in the `else` clause because swift-syntax needs to compile with Swift 5.8, which doesn’t have `if` expressions. I think the same applies to a few other methods as well.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+          \.leadingTrivia,
+          prefixExpr.leadingTrivia + prefixExpr.expression.leadingTrivia
+        )
+      )
+    }
+  }
+
+  /// Represents the negated expression of an instance of propositions.
+  private struct NegatedPropositions {
+    struct Operator {
```
ahoppen, line 472: I feel like the `Operator` type is a little overkill here. Wouldn’t it easier to just have a function that returns the negated operator’s `TokenKind` and `ExprType`?

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+      }
+
+      let leftNegated: Expr.Negated
+      let rightNegated: Expr.Negated
+      if let exprType {
+        guard self.operator.exprType == exprType else {
+          return nil
+        }
+        leftNegated = Expr(expr: infixExpr.leftOperand).negated(exprType: self.operator.exprType)
+        rightNegated = Expr(expr: infixExpr.rightOperand).negated(exprType: self.operator.exprType)
```
ahoppen, line 573: I think this is the same in both `if` branches, so we should be able to extract it to the definition of `leftNegated`.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+        expr: prefixExpr.expression.with(
+          \.leadingTrivia,
+          prefixExpr.leadingTrivia + prefixExpr.expression.leadingTrivia
+        )
+      )
+    }
+  }
+
+  /// Represents the negated expression of an instance of propositions.
+  private struct NegatedPropositions {
```
ahoppen, line 471: Do we need the `NegatedPropositions` type at all? It seems like all we ever do with it is to pass it into `Expr.map`, so instead of having the `NegatedPropositions` type, could we just have a function that returns the negated `Expr`?

ahoppen (commented):
Nice. I think the recursive design is the correct one. I do wonder whether the implementation can be simplified a little bit by having fewer intermediate types. I left some comments inline but it could be that I missed something. Please let me know what you think.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: Isn't SourceKitLSP compiled with Swift 5.9?

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+        )
+      ]
+    }
+
+    return []
+  }
+}
+
+/// A pseudo namespace for all facilities necessary for DeMorgan conversion.
+private enum DeMorgan {
```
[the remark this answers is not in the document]
AppAppWorks: True, that namespace was more of part of my thought process during this rewrite.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,730 @@
...
+      self.complementOfNegation() ?? self.complementOfPropositions()
+    }
+
+    /// Returns the complement of the expression stored by treating it as a negation, if any.
+    private func complementOfNegation() -> ExprSyntax? {
+      if let prefixExpr = self.expr.as(PrefixOperatorExprSyntax.self),
+        let negation = Negation(prefixExpr),
+        let complement = Expr(expr: negation.expr).negatedPropositions(exprType: negation.exprType)
+      {
+        // TODO: remove parentheses?
```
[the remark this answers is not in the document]
AppAppWorks: I was thinking about the backwards conversion from a complement to which this code action has added parentheses.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: I stored the negated expr of `PrefixOperatorExprSyntax` directly into `Negation` because it's what all the intermediate data types do. It's okay to turn `exprType` and `expr` into computed properties.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: During this rewrite I've found explicit nominal types tremendously helpful for my thought process because I've been confused for so many times by conflating SwiftSyntax types with domain specific concepts. 

I'm open to remove these aids for thought at this stage ;)

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: We need to tell from negated propositions the type of its operator to determine whether we need to add extra parentheses to preserve precedence order. We can turn `operator` into a computed property though.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: We need something more than `TokenKind` here, (i.e. to tell AND or OR)

AppAppWorks:
In this rewrite it was a deliberate decision to decouple domain specific concepts from Swift Syntax's parlance, as I thought it'd pave a good foundation for future implementations of other boolean/bitwise algebraic transformations, such as association, distribution, absorption, etc.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
ahoppen: You are right. The code just uses so many swift-syntax types that I thought I was in the swift-syntax repo.

I still think that explicit returns would help readability here, though. Especially with the returned tuple.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
ahoppen: While reading through the code, they confused me more than they helped me. I usually thought: Why can’t we just return the rewritten expression? Also, the `map` functions taking both a syntax node and one of the types you defined confused me a bit because they seemed to reason about to very similar but only related, not equivalent types simultaneously.

You did spend more time thinking about this than I did and I might be missing something, but if you could find the energy to write an implementation without the intermediate types that you define, I would very much like to see that. Maybe it will convince me that the current design is the correct one and we find the correct pieces of documentation that I would have needed to reason about it.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
ahoppen: Couldn’t you do that by running the negation and then when realizing that you might need to add operators, get the transformed operator using something like `expr.as(BinaryExprSyntax.self)?.operator`? That way we wouldn’t have to pass around redundant state.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
ahoppen: Do we need a whole type for that though? Wouldn’t it be sufficient to have

```swift
// In `NegatedPropositions`
var exprType: ExprType? {
  switch expr.operator.tokenKind {
  case .binaryOperator("|"):
    return .bitwise
  case .binaryOperator("&"):
    return .bitwise
  case .binaryOperator("||"):
    return .boolean
  case .binaryOperator("&&"):
    return .boolean
  default:
    return nil
  }
}
```

And where we check for `operator.kind == .and` also just check `expr.operator.tokenKind == .binaryOperator("&") || expr.operator.tokenKind == .binaryOperator("&&")`. 

Or alternatively, if you prefer add

```swift
extension TokenKind {
  var exprType: ExprType? {
    switch expr.operator.tokenKind {
    case .binaryOperator("|"):
      return .bitwise
    case .binaryOperator("&"):
      return .bitwise
    case .binaryOperator("||"):
      return .boolean
    case .binaryOperator("&&"):
      return .boolean
    default:
      return nil
    }
  }

  var isAnd: Bool {
    switch expr.operator.tokenKind {
    case .binaryOperator("&"):
      return .bitwise
    case .binaryOperator("&&"):
      return .boolean
    default:
      return nil
    }
  }
}
```

The set of `and`/`or` operators hasn’t changed since C was introduced in the 70s (or even before in other language), so I don’t think we need an abstraction layer for it.

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: > While reading through the code, they confused me more than they helped me. I usually thought: Why can’t we just return the rewritten expression? Also, the `map` functions taking both a syntax node and one of the types you defined confused me a bit because they seemed to reason about to very similar but only related, not equivalent types simultaneously.
> 
It's a fair observation :) I'd leaned much towards adopting the terminology of the de Morgan's law when rewriting this code action because, for me, it'd provided a better conceptual mapping to the business logic by abstracting away the minute details of SwiftSyntax. I've struggled with and bikeshedded on naming everything, the picture didn't become clearer after I did a complete divorce from SwiftSyntax parlance.

> You did spend more time thinking about this than I did and I might be missing something, but if you could find the energy to write an implementation without the intermediate types that you define, I would very much like to see that. Maybe it will convince me that the current design is the correct one and we find the correct pieces of documentation that I would have needed to reason about it.

Sure, I can still expend some time doing a minor rewrite before the start of the new school term ;)

Sources/SourceKitLSP/Swift/CodeActions/ApplyDeMorganLaw.swift
AppAppWorks: > `map` seems like a pretty ambiguous term here but I’m struggling to find anything more meaningful.

I had functors in mind when choosing `map` as the name, regrettably I also struggled with finding a better name.

PR #1567, opened 2024-07-17
Allow macro expansions to be viewed through `GetReferenceDocumentRequest` instead of storing in temporary files

lokesh-tr:
This PR eliminates the storage of temporary macro expansion files for `PeekDocumentsRequest` by introducing a new LSP Extension `GetReferenceDocumentRequest`.

The implementation is made in such a way that the URI passed for macro expansions through `PeekDocumentsRequest` is customised with a new scheme, and expects the client to resolve the contents of this new URI by making a `GetReferenceDocumentRequest`.

```
Custom Scheme for Macro Expansions:
`sourcekit-lsp://swift-macro-expansion/LaCb-LcCd.swift?primaryFilePath=&fromLine=&fromColumn=&toLine=&toColumn=&bufferName=`
```

#### References:
Previous PR which implemented `PeekDocumentsRequest`: https://github.com/swiftlang/sourcekit-lsp/pull/1479
Accompanying PR in vscode-swift repository: https://github.com/swiftlang/vscode-swift/pull/971

------

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -357,7 +357,7 @@ package actor SourceKitLSPServer {
 
     // This should be created as soon as we receive an open call, even if the document
     // isn't yet ready.
-    guard let languageService = workspace.documentService.value[doc] else {
+    guard let languageService = workspace.documentService(for: doc) else {
       return
     }
 
@@ -377,7 +377,7 @@ package actor SourceKitLSPServer {
       guard let workspace = await self.workspaceForDocument(uri: request.textDocument.uri) else {
         throw ResponseError.workspaceNotOpen(request.textDocument.uri)
       }
-      guard let languageService = workspace.documentService.value[doc] else {
+      guard let languageService = workspace.documentService(for: doc) else {
         throw ResponseError.unknown("No language service for '\(request.textDocument.uri)' found")
       }
       return try await requestHandler(request, workspace, languageService)
@@ -400,7 +400,7 @@ package actor SourceKitLSPServer {
       guard let workspace = await self.workspaceForDocument(uri: documentUri) else {
         continue
       }
-      guard workspace.documentService.value[documentUri] === languageService else {
+      guard workspace.documentService(for: documentUri) === languageService else {
         continue
       }
       guard let snapshot = try? self.documentManager.latestSnapshot(documentUri) else {
@@ -518,7 +518,7 @@ package actor SourceKitLSPServer {
     _ language: Language,
     in workspace: Workspace
   ) async -> LanguageService? {
-    if let service = workspace.documentService.value[uri] {
+    if let service = workspace.documentService(for: uri) {
       return service
     }
 
@@ -536,17 +536,7 @@ package actor SourceKitLSPServer {
       """
     )
 
-    return workspace.documentService.withLock { documentService in
-      if let concurrentlySetService = documentService[uri] {
-        // Since we await the construction of `service`, another call to this
-        // function might have happened and raced us, setting
-        // `workspace.documentServices[uri]`. If this is the case, return the
-        // existing value and discard the service that we just retrieved.
-        return concurrentlySetService
-      }
-      documentService[uri] = service
-      return service
-    }
+    return workspace.setDocumentService(for: uri, service)
   }
 }
 
@@ -733,6 +723,8 @@ extension SourceKitLSPServer: MessageHandler {
       await request.reply { try await executeCommand(request.params) }
     case let request as RequestAndReply<FoldingRangeRequest>:
       await self.handleRequest(for: request, requestHandler: self.foldingRange)
+    case let request as RequestAndReply<GetReferenceDocumentRequest>:
+      await request.reply { try await getReferenceDocument(request.params) }
     case let request as RequestAndReply<HoverRequest>:
       await self.handleRequest(for: request, requestHandler: self.hover)
     case let request as RequestAndReply<ImplementationRequest>:
@@ -804,7 +796,7 @@ extension SourceKitLSPServer: BuildSystemDelegate {
         continue
       }
 
-      guard let service = await self.workspaceForDocument(uri: uri)?.documentService.value[uri] else {
+      guard let service = await self.workspaceForDocument(uri: uri)?.documentService(for: uri) else {
         continue
       }
 
@@ -828,7 +820,7 @@ extension SourceKitLSPServer: BuildSystemDelegate {
       }
       for uri in self.affectedOpenDocumentsForChangeSet(changedFilesForWorkspace, self.documentManager) {
         logger.log("Dependencies updated for opened file \(uri.forLogging)")
-        if let service = workspace.documentService.value[uri] {
+        if let service = workspace.documentService(for: uri) {
           await service.documentDependenciesUpdated(uri)
         }
       }
@@ -964,6 +956,8 @@ extension SourceKitLSPServer {
     //
     // The below is a workaround for the vscode-swift extension since it cannot set client capabilities.
     // It passes "workspace/peekDocuments" through the `initializationOptions`.
+    //
+    // Similarly, for "workspace/getReferenceDocument".
     var clientCapabilities = req.capabilities
     if case .dictionary(let initializationOptions) = req.initializationOptions {
       if let peekDocuments = initializationOptions["workspace/peekDocuments"] {
@@ -972,6 +966,15 @@ extension SourceKitLSPServer {
           clientCapabilities.experimental = .dictionary(experimentalCapabilities)
         } else {
           clientCapabilities.experimental = .dictionary(["workspace/peekDocuments": peekDocuments])
+        }
+      }
+
+      if let getReferenceDocument = initializationOptions["workspace/getReferenceDocument"] {
+        if case .dictionary(var experimentalCapabilities) = clientCapabilities.experimental {
+          experimentalCapabilities["workspace/getReferenceDocument"] = getReferenceDocument
+          clientCapabilities.experimental = .dictionary(experimentalCapabilities)
+        } else {
+          clientCapabilities.experimental = .dictionary(["workspace/getReferenceDocument": getReferenceDocument])
         }
       }
 
@@ -1143,6 +1146,7 @@ extension SourceKitLSPServer {
         "workspace/tests": .dictionary(["version": .int(2)]),
         "textDocument/tests": .dictionary(["version": .int(2)]),
         "workspace/triggerReindex": .dictionary(["version": .int(1)]),
+        "workspace/getReferenceDocument": .dictionary(["version": .int(1)]),
       ])
     )
   }
@@ -1342,7 +1346,7 @@ extension SourceKitLSPServer {
       )
       return
     }
-    await workspace.documentService.value[uri]?.reopenDocument(notification)
+    await workspace.documentService(for: uri)?.reopenDocument(notification)
   }
 
   func closeDocument(_ notification: DidCloseTextDocumentNotification, workspace: Workspace) async {
@@ -1356,7 +1360,7 @@ extension SourceKitLSPServer {
 
     await workspace.buildSystemManager.unregisterForChangeNotifications(for: uri)
 
-    await workspace.documentService.value[uri]?.closeDocument(notification)
+    await workspace.documentService(for: uri)?.closeDocument(notification)
   }
 
   func changeDocument(_ notification: DidChangeTextDocumentNotification) async {
@@ -1382,7 +1386,7 @@ extension SourceKitLSPServer {
       // Already logged failure
       return
     }
-    await workspace.documentService.value[uri]?.changeDocument(
+    await workspace.documentService(for: uri)?.changeDocument(
       notification,
       preEditSnapshot: preEditSnapshot,
       postEditSnapshot: postEditSnapshot,
@@ -1645,7 +1649,7 @@ extension SourceKitLSPServer {
     guard let workspace = await workspaceForDocument(uri: uri) else {
       throw ResponseError.workspaceNotOpen(uri)
     }
-    guard let languageService = workspace.documentService.value[uri] else {
+    guard let languageService = workspace.documentService(for: uri) else {
       return nil
     }
 
@@ -1654,6 +1658,21 @@ extension SourceKitLSPServer {
       arguments: req.argumentsWithoutSourceKitMetadata
     )
     return try await languageService.executeCommand(executeCommand)
+  }
+
+  func getReferenceDocument(_ req: GetReferenceDocumentRequest) async throws -> GetReferenceDocumentResponse {
+    let referenceDocumentURL = try ReferenceDocumentURL(from: req.uri)
+    let primaryFileURI = referenceDocumentURL.primaryFile
+
+    guard let workspace = await workspaceForDocument(uri: primaryFileURI) else {
+      throw ResponseError.workspaceNotOpen(primaryFileURI)
+    }
+
+    guard let languageService = workspace.documentService(for: primaryFileURI) else {
+      throw ResponseError.unknown("No Language Service for URI: \(primaryFileURI)")
+    }
+
+    return try await languageService.getReferenceDocument(req)
   }
 
   func codeAction(
--- a/Sources/SourceKitLSP/Swift/MacroExpansion.swift
+++ b/Sources/SourceKitLSP/Swift/MacroExpansion.swift
@@ -62,7 +62,7 @@ extension SwiftLanguageService {
       throw ResponseError.unknown("Connection to the editor closed")
     }
 
-    guard let sourceFileURL = expandMacroCommand.textDocument.uri.fileURL else {
+    guard let primaryFileURL = expandMacroCommand.textDocument.uri.fileURL else {
       throw ResponseError.unknown("Given URI is not a file URL")
     }
 
@@ -71,57 +71,26 @@ extension SwiftLanguageService {
     var completeExpansionFileContent = ""
     var completeExpansionDirectoryName = ""
 
-    var macroExpansionFilePaths: [URL] = []
+    var macroExpansionReferenceDocumentURLs: [ReferenceDocumentURL] = []
     for macroEdit in expansion.edits {
       if let bufferName = macroEdit.bufferName {
-        // buffer name without ".swift"
-        let macroExpansionBufferDirectoryName =
-          bufferName.hasSuffix(".swift")
-          ? String(bufferName.dropLast(6))
-          : bufferName
+        let macroExpansionReferenceDocumentURLData =
+          ReferenceDocumentURL.macroExpansion(
+            MacroExpansionReferenceDocumentURLData(
+              macroExpansionEditRange: macroEdit.range,
+              primaryFileURL: primaryFileURL,
+              selectionRange: expandMacroCommand.positionRange,
+              bufferName: bufferName
+            )
+          )
 
-        let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
-          .appendingPathComponent(macroExpansionBufferDirectoryName)
+        macroExpansionReferenceDocumentURLs.append(macroExpansionReferenceDocumentURLData)
 
         completeExpansionDirectoryName += "\(bufferName)-"
 
-        do {
-          try FileManager.default.createDirectory(
-            at: macroExpansionBufferDirectoryURL,
-            withIntermediateDirectories: true
-          )
-        } catch {
-          throw ResponseError.unknown(
-            "Failed to create directory for macro expansion buffer at path: \(macroExpansionBufferDirectoryURL.path)"
-          )
-        }
-
-        // name of the source file
-        let macroExpansionFileName = sourceFileURL.deletingPathExtension().lastPathComponent
-
-        // github permalink notation for position range
-        let macroExpansionPositionRangeIndicator =
-          "L\(macroEdit.range.lowerBound.line + 1)C\(macroEdit.range.lowerBound.utf16index + 1)-L\(macroEdit.range.upperBound.line + 1)C\(macroEdit.range.upperBound.utf16index + 1)"
-
-        let macroExpansionFilePath =
-          macroExpansionBufferDirectoryURL
-          .appendingPathComponent(
-            "\(macroExpansionFileName)_\(macroExpansionPositionRangeIndicator).\(sourceFileURL.pathExtension)"
-          )
-
-        do {
-          try macroEdit.newText.write(to: macroExpansionFilePath, atomically: true, encoding: .utf8)
-        } catch {
-          throw ResponseError.unknown(
-            "Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\""
-          )
-        }
-
-        macroExpansionFilePaths.append(macroExpansionFilePath)
-
         let editContent =
           """
-          // \(sourceFileURL.lastPathComponent) @ \(macroEdit.range.lowerBound.line + 1):\(macroEdit.range.lowerBound.utf16index + 1) - \(macroEdit.range.upperBound.line + 1):\(macroEdit.range.upperBound.utf16index + 1)
+          // \(primaryFileURL.lastPathComponent) @ \(macroEdit.range.lowerBound.line + 1):\(macroEdit.range.lowerBound.utf16index + 1) - \(macroEdit.range.upperBound.line + 1):\(macroEdit.range.upperBound.utf16index + 1)
           \(macroEdit.newText)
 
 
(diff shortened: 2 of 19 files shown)


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1633,6 +1635,25 @@ extension SourceKitLSPServer {
...
+    let filePath = "file:/" + givenURL.pathComponents.dropLast().joined(separator: "/")
+    let sourceFileURI = try DocumentURI(string: filePath)
+
+    guard let workspace = await workspaceForDocument(uri: sourceFileURI) else {
+      throw ResponseError.workspaceNotOpen(sourceFileURI)
+    }
+
+    guard let languageService = workspace.documentService.value[sourceFileURI] else {
+      throw ResponseError.unknown("No Language Service for URI: \(sourceFileURI)")
+    }
```
ahoppen, line 1652: You are really using implementation details of how the custom URL is formed to get the workspace for the document. We shouldn’t need to construct the `filePath`, instead we should call `workspaceForDocument(uri: req.uri)`. 

The way I would implement this is that `Workspace` gets a new method

```swift
/// Returns the primary document based on which the document at the given `uri` is based.
/// 
/// For macro expansions, this is the document from which the macro was expanded.
func primaryDocument(for uri: DocumentURI) -> DocumentURI {
  if fileURL.scheme == "swift-macro-expansion" {
    // your logic from lines 1643 and 1644.
  }
  return uri
}
```

And then in `SourceKitLSPServer.workspaceForDocument`, we use that primary file to call `fileHandlingCapability(for:)`.

That will give use the workspace that’s responsible for the macro expansion. To get the language service, we should wrap the current accesses to `workspace.documentService.value` in a function on `Workspace` that also calls `primaryDocument(for:)`. Something like the following (we can probably also mark `documentService` private then.

```swift
func documentService(for uri: DocumentURI) -> LanguageService? {
  return documentService.value[primaryDocument(for: uri)]
}
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -74,50 +74,38 @@ extension SwiftLanguageService {
...
-            "Unable to write macro expansion to file path: \"\(macroExpansionFilePath.path)\""
-          )
+        components.scheme = "swift-macro-expansion"
+        components.path +=
+          "/L\(macroEdit.range.lowerBound.line + 1)C\(macroEdit.range.lowerBound.utf16index + 1)-L\(macroEdit.range.upperBound.line + 1)C\(macroEdit.range.upperBound.utf16index + 1).swift"
+        let queryItems = [
+          URLQueryItem(name: "lowerL", value: String(expandMacroCommand.positionRange.lowerBound.line)),
+          URLQueryItem(name: "lowerC", value: String(expandMacroCommand.positionRange.lowerBound.utf16index)),
+          URLQueryItem(name: "upperL", value: String(expandMacroCommand.positionRange.upperBound.line)),
+          URLQueryItem(name: "upperC", value: String(expandMacroCommand.positionRange.upperBound.utf16index)),
```
ahoppen, line 98: We don’t really need to optimize for a few characters here, so let’s spell out `L` to `Line` and `C` to `Column`.

Also, I find `from` and `to` more intuitive names than `lower` and `upper`.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -968,6 +968,48 @@ extension SwiftLanguageService {
...
   }
+
+  public func getReferenceDocument(_ req: GetReferenceDocumentRequest) async throws -> GetReferenceDocumentResponse {
+    switch req.uri.scheme {
+    case "swift-macro-expansion":
+      guard let givenURL = URL(string: req.uri.stringValue) else {
+        throw ResponseError.unknown("Unable to parse given URL")
+      }
+
+      let filePath = "file:/" + givenURL.pathComponents.dropLast().joined(separator: "/")
```
ahoppen, line 979: Now that I read this, I think it would be good to centralize the creation and parsing of this URL scheme. I’d suggest to create a `MacroExpansionURLScheme` type that has members for all the parameters encoded in the URL scheme and that can be initialized from + converted to a URL. That way it’s easier to make sure that the URL scheme matches in the function that creates it and the functions that interpret it.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -968,6 +968,48 @@ extension SwiftLanguageService {
...
+        selectionRange: selectionRange,
+        bufferName: bufferName
+      )
+
+      return GetReferenceDocumentResponse(content: content)
+    default:
+      throw ResponseError.unknown("Invalid Scheme")
+    }
+
+    throw ResponseError.unknown("No Content to show")
```
ahoppen, line 1011: I don’t think this code will ever be executed, right?

ahoppen (commented):
Wow, this was easier than I expected.

lokesh-tr:
@ahoppen I have updated the PR and fixed the rebase issue with package specifiers. Ready for full review now. 👍🏻

Sources/SourceKitLSP/Swift/ReferenceDocument.swift
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Creates the `DocumentURI` of the source file, using which the reference document is generated, and returns it.
+  /// - Returns: The `DocumentURI` of the source file
+  func sourceDocument() -> DocumentURI
+
+  /// This generates the content of the reference document and returns it.
+  /// - Parameter languageService: The `SwiftLanguageService` using which it generates
+  /// the content of the reference document.
+  /// - Returns: The content of the reference document
+  func documentContent(using languageService: SwiftLanguageService) async throws -> String
```
ahoppen, line 50: I don’t like how `ReferenceDocument` is used to both form/parse the URL of a reference document and to get the document contents because it’s not clear what the type’s purpose is. I would prefer if we had a dedicated type to just do the URL conversions and then a method on `SwiftLanguageService` that is responsible for generating the document contents.

And since we know all the types of reference documents that SourceKit-LSP can support, it would probably be cleaner if we have an enum that represents all of them. For example, something like

```swift
enum ReferenceDocumentURL {
  case macroExpansion(MacroExpansionReferenceDocumentURLData)

  var url: URL { … }

  init(from url: URL) throws { … }

  // I don’t think any other members are needed
}
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -74,51 +74,17 @@ extension SwiftLanguageService {
     var macroExpansionFilePaths: [URL] = []
     for macroEdit in expansion.edits {
       if let bufferName = macroEdit.bufferName {
```
ahoppen, line 76: It’s a shame that we need to effectively run the expand macro refactoring request twice now: Once to get the macro’s range and buffer name and then again in the `workspace/getReferenceDocument` request. But I don’t really see a way around that. So, I guess it’s fine.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -74,51 +74,17 @@ extension SwiftLanguageService {
...
+        let macroExpansionReferenceDocument = MacroExpansionReferenceDocument(
+          macroExpansionEditRange: macroEdit.range,
+          sourceFileURL: sourceFileURL,
+          selectionRange: expandMacroCommand.positionRange,
+          bufferName: bufferName
+        )
 
-        let macroExpansionBufferDirectoryURL = self.generatedMacroExpansionsPath
-          .appendingPathComponent(macroExpansionBufferDirectoryName)
+        macroExpansionFilePaths.append(try macroExpansionReferenceDocument.url)
```
ahoppen, line 84: These are no longer file paths, right? So, this should probably be `macroReferenceDocuments`. Also, we could maintain a little bit more type information here and make this a variable of type `[ReferenceDocumentURL]`.

Sources/SourceKitLSP/Swift/ReferenceDocument.swift
```diff
@@ -0,0 +1,150 @@
...
+
+package extension ReferenceDocument {
+  var url: URL {
+    get throws {
+      let baseURL = "sourcekit-lsp://\(Self.documentType.rawValue)"
+      guard var components = URLComponents(string: baseURL) else {
+        throw ReferenceDocumentError(
+          description: "Unable to create URL for ReferenceDocument: \(baseURL)"
+        )
+      }
```
ahoppen, line 78: Wouldn’t it be cleaner if you just create an empty `URLComponents` and then set scheme and host on it?

```suggestion
var components = URLComponents()
components.scheme = "sourcekit-lsp"
components.host = "swift-macro-expansion"
```

Sources/SourceKitLSP/Swift/ReferenceDocument.swift
```diff
@@ -0,0 +1,150 @@
...
+    guard let documentType = url.host(percentEncoded: false),
+      let referenceDocumentType = ReferenceDocumentType(rawValue: documentType),
+      let type = supportedReferenceDocumentTypes[referenceDocumentType]
+    else {
+      throw ReferenceDocumentError(description: "Invalid Document Type")
+    }
+
+    return try type.init(from: url)
+  }
+}
```
ahoppen, line 142: With my proposal of `ReferenceDocumentURL`, this should be handled by the initializer.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -75,7 +75,8 @@ package final class Workspace: Sendable {
   private let documentManager: DocumentManager
 
   /// Language service for an open document, if available.
-  let documentService: ThreadSafeBox<[DocumentURI: LanguageService]> = ThreadSafeBox(initialValue: [:])
+  @_spi(Testing) public let documentService: ThreadSafeBox<[DocumentURI: LanguageService]> = ThreadSafeBox(
```
ahoppen, line 78: Why does this need to become SPI? I don’t see any uses in the tests.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -253,6 +254,10 @@ package final class Workspace: Sendable {
     await syntacticTestIndex.filesDidChange(events)
     await semanticIndexManager?.filesDidChange(events)
   }
+
+  public func documentService(for uri: DocumentURI) -> LanguageService? {
+    return documentService.value[uri]
+  }
```
ahoppen, line 260: Could this be `internal` access level?

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -199,110 +199,127 @@ final class ExecuteCommandTests: XCTestCase {
...
 
-          let urls = try XCTUnwrap(
-            peekDocumentsRequestURIs.value?.map {
-              return try XCTUnwrap(
-                $0.fileURL,
+            _ = try await project.testClient.send(request)
+
+            try await fulfillmentOfOrThrow([expectation])
+
+            let urlStrings = try XCTUnwrap(
```
ahoppen, line 251: I think we could save a bit of casting if this was just `[DocumentURI]`.

Same in the test below.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -371,125 +388,142 @@ final class ExecuteCommandTests: XCTestCase {
...
-        (start: "1️⃣", end: "3️⃣"),
-        (start: "2️⃣", end: "3️⃣"),
-      ]
-
-      for positionMarker in positionMarkersToBeTested {
-        let args = ExpandMacroCommand(
-          positionRange: positions[positionMarker.start]..<positions[positionMarker.end],
-          textDocument: TextDocumentIdentifier(uri)
+    for getReferenceDocument in [false, true] {
+      for peekDocuments in [false, true] {
```
ahoppen, line 392: I’m wondering if it’s worth saving a nesting level by doing

```suggestion
    for (getReferenceDocument, peekDocuments) in [(false, false), (false, true), (true, false), (true, true)] {
```

What do you think? There’s definitely a trade-off here.

lokesh-tr: I don't like the nesting too. I wish there was some function that goes like: `Bool.truthTableBase(variablesCount: 2)` which generates the true/false pattern for the given number of variables, It will also be less error-prone because we won't be hard coding the values, and will be more readable. That will then allow us to write the following:
```
    for (getReferenceDocument, peekDocuments) in Bool.truthTableBase(variablesCount: 2) {  
```
Even if any new test case needs similar looping with more boolean variables, It can go like:
```
    for (bool1, bool2, bool3, ..., boolN) in Bool.truthTableBase(variablesCount: N) {
```

But maybe this is over-engineering and just hard-coding the values would be fine for now.

lokesh-tr: Or this could be a neat candidate for `SwiftExtensions` package 🤔

lokesh-tr:
@ahoppen ready for review

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -255,6 +256,23 @@ package final class Workspace: Sendable {
...
     await semanticIndexManager?.filesDidChange(events)
   }
+
+  func documentService(for uri: DocumentURI) -> LanguageService? {
+    return documentService.value[uri]
+  }
+
+  /// Set a language service for a document uri only if it doesn't already exist.
+  /// - Returns: The language service of the given `uri`
+  func documentService(for uri: DocumentURI, newLanguageService: any LanguageService) -> LanguageService {
```
ahoppen, line 266: At its core, this method is a setter but its current name sounds like a getter. What do you think about?

```suggestion
  /// Set a language service for a document uri and returns if none exists already.
  /// If a language service already exists for this document, eg. because two requests start creating a language service for a document and race, `newLanguageService` is dropped and the existing language service for the document is returned.
  func setDocumentService(for uri: DocumentURI, _ newLanguageService: any LanguageService) -> LanguageService {
```

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -0,0 +1,75 @@
...
+    if let macroExpansionURLData = try? MacroExpansionReferenceDocumentURLData(from: url) {
+      self = .macroExpansion(macroExpansionURLData)
+    } else {
+      throw ReferenceDocumentURLError(description: "Invalid document type / data in url for reference document")
+    }
+  }
+
+  /// Creates the `DocumentURI` of the source file, using which the reference document is generated, and returns it.
+  /// - Returns: The `DocumentURI` of the source file
+  func sourceDocument() -> DocumentURI {
```
ahoppen, line 61: I think the primary purpose of `sourceDocument` is to decide which language service is responsible for handling this `GetReferenceDocumentRequest`. How about we follow the sourcekitd naming here and call it `primaryFile`, which also avoids ambiguity between *source* and *source code*.

Also, this can be a computed property.

```suggestion
  /// The URI of the document from which this reference document was derived. This is used to determine the workspace and language service that is used to generate the reference document.
  var sourceDocument: DocumentURI {
```

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      return url
+    }
+  }
+
+  package init(from url: URL) throws {
+    guard url.scheme == "sourcekit-lsp" else {
+      throw ReferenceDocumentURLError(description: "Invalid scheme for reference document")
+    }
```
ahoppen, line 87: Instead of repeating this scheme check for every reference document URL data that we might get in the future, what do you think about fully splitting the responsibility of the URL components between `ReferenceDocumentURL` and `MacroExpansionReferenceURLData`. 

`ReferenceDocumentURL` is responsible for scheme and host.
`MacroExpansionReferenceURLData` is responsible for display name and query parameters.

So, `MacroExpansionReferenceURLData` would look like the following

```swift
struct MacroExpansionReferenceURLData {
  var displayName: String { … }
  var queryItems: [URLQueryItem] { … }
  
  init(displayName: String, queryItems: [URLQueryItem]) { … }
}
```

And `ReferenceDocumentURL` can do the scheme check + decide which reference document URL type to parse by inspecting the host.

Also, I think we should parse `macroExpansionEditRange` from the display name so that it’s possible to round-trip `ReferenceDocumentURL` to a URL and back to a `ReferenceDocumentURL`.

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+
+import Foundation
+import LanguageServerProtocol
+
+/// Represents url of macro expansion reference document as follows:
+/// `sourcekit-lsp://swift-macro-expansion/LaCb-LcCd.swift?sourceFileURL=&fromLine=&fromColumn=&toLine=&toColumn=&bufferName=`
+///
+/// Here,
+///  - `LaCb-LcCd.swift`, the `displayName`, represents where the macro will expand to or
+/// replace in the source file (i.e. `macroExpansionEditRange`)
```
ahoppen, line 21:
```suggestion
///    replace in the source file (i.e. `macroExpansionEditRange`)
```

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+  private var displayName: String {
+    if let macroExpansionEditRange {
+      "L\(macroExpansionEditRange.lowerBound.line + 1)C\(macroExpansionEditRange.lowerBound.utf16index + 1)-L\(macroExpansionEditRange.upperBound.line + 1)C\(macroExpansionEditRange.upperBound.utf16index + 1).swift"
+    } else {
+      ""
+    }
+  }
+
+  private var queryItems: [URLQueryItem]? {
+    get {
```
ahoppen, line 52: If the getter doesn’t throw, you can remove the `get {`.

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+    self.selectionRange =
+      Position(line: fromLine, utf16index: fromColumn)..<Position(line: toLine, utf16index: toColumn)
+    self.bufferName = bufferName
+  }
+
+  package func sourceDocument() -> DocumentURI {
+    DocumentURI(sourceFileURL)
+  }
+
+  struct Parameters {
```
ahoppen, line 125: I guess this could be `private`, right?

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    self.sourceFileURL = sourceFileURL
+    self.selectionRange =
+      Position(line: fromLine, utf16index: fromColumn)..<Position(line: toLine, utf16index: toColumn)
+    self.bufferName = bufferName
+  }
+
+  package func sourceDocument() -> DocumentURI {
+    DocumentURI(sourceFileURL)
+  }
```
ahoppen, line 123: Could also be a computed property.

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    guard let urlComponents = URLComponents(string: url.absoluteString),
+      let queryItems = urlComponents.queryItems,
+      let sourceFilePath = queryItems.first { $0.name == Parameters.sourceFilePath }?.value,
+      let fromLine = Int(queryItems.first { $0.name == Parameters.fromLine }?.value ?? ""),
+      let fromColumn = Int(queryItems.first { $0.name == Parameters.fromColumn }?.value ?? ""),
+      let toLine = Int(queryItems.first { $0.name == Parameters.toLine }?.value ?? ""),
+      let toColumn = Int(queryItems.first { $0.name == Parameters.toColumn }?.value ?? ""),
+      let bufferName = queryItems.first { $0.name == Parameters.bufferName }?.value
```
ahoppen, line 102: Nitpicky, but I’d do `last` because usually the semantics are that a repeated statement of the same parameter overrides previous parameters.

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -0,0 +1,75 @@
...
+    get throws {
+      switch self {
+      case let .macroExpansion(data):
+        return try data.url
+      }
+    }
+  }
+
+  init(from uri: DocumentURI) throws {
+    guard let url = URL(string: uri.stringValue) else {
```
ahoppen, line 40: Instead of going through the `stringValue` and then parsing that as a `URL` again, we should expose `DocumentURI`. Since in almost all use cases users want `fileURL`, I would name it something that sounds a little scary, like

```swift
/// The URL representation of the URI. Note that this URL can have an arbitrary scheme and might not represent a file URL.
public var arbitrarySchemURL: URL { storage }
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -208,4 +178,33 @@ extension SwiftLanguageService {
       }
     }
   }
+
+  func expandMacro(macroExpansionURLData: inout MacroExpansionReferenceDocumentURLData) async throws -> String {
```
ahoppen, line 182: Why does the parameter need to be `inout`? As far as I can tell, this is only called from `getReferenceDocument`, which doesn’t use the modified parameter.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -208,4 +178,33 @@ extension SwiftLanguageService {
...
+    )
+
+    let expansion = try await self.refactoring(expandMacroCommand)
+
+    guard
+      let macroExpansionEdit = expansion.edits.filter({
+        $0.bufferName == macroExpansionURLData.bufferName
+      }).only
+    else {
+      throw ResponseError.unknown("Macro Expansion Edit doesn't Exist")
```
ahoppen, line 201:
```suggestion
      throw ResponseError.unknown("Macro expansion edit doesn't exist")
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -208,4 +178,33 @@ extension SwiftLanguageService {
...
+    }
+
+    macroExpansionURLData.macroExpansionEditRange = macroExpansionEdit.range
+
+    let content = macroExpansionEdit.newText
+
+    return content
```
ahoppen, line 208:
```suggestion
    return macroExpansionEdit.newText
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -980,6 +980,17 @@ extension SwiftLanguageService {
 
     return nil
   }
+
+  public func getReferenceDocument(_ req: GetReferenceDocumentRequest) async throws -> GetReferenceDocumentResponse {
```
ahoppen, line 984:
```suggestion
  package func getReferenceDocument(_ req: GetReferenceDocumentRequest) async throws -> GetReferenceDocumentResponse {
```

ahoppen (commented):
Looks much better with the split for the URL data in my opinion.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
ahoppen: I think the natural think to implement in `SwiftExtensions` would be a cartesian product, so that you would have

```swift
for (getReferenceDocument, peekDocuments) in cartesionProduct([true, false], [true, false]) {
```

lokesh-tr: I wanted to write a very general `cartesianProduct` function, but seems like we will lose the ability to unpack the values since it returns an array.

Perhaps, parameter packs can be used in place of variadic parameters to solve this. I tried it out of our project with Swift 6, but am slightly confused with the implementation.

And even then, I am using Swift 5.10 for our regular workflow.

For now, I guess this would suffice. When the project moves to Swift 6 completely, then we can update the function to use parameter packs maybe.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -198,11 +198,17 @@ final class ExecuteCommandTests: XCTestCase {
 
     let options = SourceKitLSPOptions.testDefault(experimentalFeatures: [.showMacroExpansions])
 
-    for peekDocuments in [false, true] {
+    for product in cartesianProduct([true, false], [true, false]) {
+      let getReferenceDocument = product[0]
+      let peekDocuments = product[1]
```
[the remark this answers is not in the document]
lokesh-tr: Exactly! I thought of making a proposal for some new syntax to solve this, lol. Have to do further research...

Sources/SwiftExtensions/CartesianProduct.swift
```diff
@@ -0,0 +1,49 @@
...
+/// - Returns: The cartesian product of the given `arrays`
+///
+/// Example Usage:
+/// ```swift
+/// let coinFlipOutcomes = ["Head", "Tails"]
+/// let twoCoinFlipOutcomes = cartesianProduct(coinFlipOutcomes, coinFlipOutcomes)
+/// print(twoCoinFlipOutcomes)
+/// // Prints: "[["Head", "Head"], ["Tails", "Head"], ["Head", "Tails"], ["Tails", "Tails"]]"
+///
+/// let alphaNumberPairs = cartesianProduct([1, 2, 3],  ["a", "b", "c"])
```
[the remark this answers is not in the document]
lokesh-tr: I checked it out, seems like it upcasts to `Any`

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,150 @@
...
+    static let sourceFilePath = "sourceFilePath"
+    static let fromLine = "fromLine"
+    static let fromColumn = "fromColumn"
+    static let toLine = "toLine"
+    static let toColumn = "toColumn"
+    static let bufferName = "bufferName"
+  }
+
+  private func parse(displayName: String) throws -> Range<Position> {
+    let regex = try? Regex("L\\d+C\\d+-L\\d+C\\d+\\.swift")
```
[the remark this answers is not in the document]
lokesh-tr: Wow! this is great, I was pretty good with Swift 4-ish era and then got busy with everything else. I probably missed this one while learning Swift 5.

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,150 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+
+/// Represents url of macro expansion reference document as follows:
+/// `sourcekit-lsp://swift-macro-expansion/LaCb-LcCd.swift?sourceFileURL=&fromLine=&fromColumn=&toLine=&toColumn=&bufferName=`
+///
```
lokesh-tr, line 18: @ahoppen This line exceeds the 120 character line limit. Running Swift Lint works fine. Will this cause any issues in the CI?

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
```diff
@@ -0,0 +1,147 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import LanguageServerProtocol
+import RegexBuilder
+
+/// Represents url of macro expansion reference document as follows:
+/// `sourcekit-lsp://swift-macro-expansion/LaCb-LcCd.swift?primaryFilePath=&fromLine=&fromColumn=&toLine=&toColumn=&bufferName=`
```
lokesh-tr, line 18: @ahoppen This line exceeds the 120 character line limit. Running Swift Lint works fine. Will this cause any issues in the CI?

lokesh-tr:
@ahoppen Fixed everything as you suggested 🙂

Sources/SourceKitLSP/Swift/MacroExpansionReferenceDocumentURLData.swift
ahoppen: That’s OK. swift-format can’t break comments at the moment and doesn’t complain about them either.

ahoppen (approved):
Very nice

ahoppen:
@swift-ci Please test Windows

PR #1570, opened 2024-07-17
Remove `indexPrefixMappings` from `BuildSystem`

ahoppen:
They weren’t used.


ahoppen:
@swift-ci Please test

bnbarham (commented):
Could use them instead? They're needed if the index was built with prefix mappings, which would be nice to support. Probably need to add them as options to sourcekit-lsp as well (in cases where it's not possible to retrieve from the build system).

ahoppen:
Since the current `BuildSystem.indexPrefixMappings` wasn’t doing anything, let’s remove it. I opened https://github.com/swiftlang/sourcekit-lsp/issues/1608 to properly add index prefix mapping support to build systems.

ahoppen:
p;t

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #1571, opened 2024-07-17
Add a documentation document about background indexing

ahoppen:


ahoppen:
@swift-ci Please test

Documentation/Background Indexing.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Status tracking
+
+When SourceKit-LSP is launched, all targets are considered to be out-of-date. This needs to be done because source files might have changed since SourceKit-LSP was last started – if the module wasn’t modified since the last SourceKit-LSP launch, we re-prepare the target and rely on the build system to produce a fast null build.
+
+After we have prepared a target, we mark it as being up-to-date in `SemanticIndexManager.preparationUpToDateTracker`. That way we don’t need to invoke the build system every time we want to perform semantic functionality of a source file, which saves us the time of a null build (which can hundreds of milliseconds for SwiftPM). If a source file is changed (as noted through file watching), all of its target’s dependencies are marked as out-of-date. Note that the target that the source file belongs to is not marked as out-of-date – preparation of a target builds all dependencies but does not need to build the modules itself. The next operation that requires the target to be prepared will trigger a preparation job.
```
bnbarham, line 17:
```suggestion
After we have prepared a target, we mark it as being up-to-date in `SemanticIndexManager.preparationUpToDateTracker`. That way we don’t need to invoke the build system every time we want to perform semantic functionality of a source file, which saves us the time of a null build (which can hundreds of milliseconds for SwiftPM). If a source file is changed (as noted through file watching), all of its target’s dependencies are marked as out-of-date. Note that the target that the source file belongs to is not marked as out-of-date – preparation of a target builds all dependencies but does not need to build the target's module itself. The next operation that requires the target to be prepared will trigger a preparation job.
```

Documentation/Background Indexing.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Updating the index store
+
+Updating the index store is done by invoking the compiler for the source file (either `swiftc` or `clang`) with flags that write index data about declared symbols and their occurrences (ie. locations that reference them) to the index store as unit and record files. These unit and record files are watched for by indexstore-db, which is a database on top of the raw index date, which allows an efficient lookup of the record files that contain the reference to a symbol.
```
bnbarham, line 21:
```suggestion
Updating the index store is done by invoking the compiler for the source file (either `swiftc` or `clang`) with flags that write index data about declared symbols and their occurrences (ie. locations that reference them) to the index store as unit and record files. These unit and record files are watched for by indexstore-db, which is a database on top of the raw index data that allows eg. an efficient lookup of the record files that contain the reference to a symbol.
```

Documentation/Background Indexing.md
```diff
@@ -0,0 +1,42 @@
...
+
+Other than many other LSP servers, SourceKit-LSP does not have a custom parser and type checker but re-uses the compiler to resolve symbols.
```
bnbarham, line 23:
```suggestion
Unlike many other LSP servers, SourceKit-LSP does not have a custom parser and type checker but re-uses the compiler to resolve symbols.
```

Though this feels a bit out of place here.

Documentation/Background Indexing.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Status tracking
+
+Similarly to target preparation, when SourceKit-LSP is launched, the index of all source files is considered out-of-date and the initial index of the project is triggered, which indexes every file.
+
+When a source file should be indexed, we first check if a unit file for this source file already exists and whether the date of that unit file is later than the last modification date of the source file. If this is the case, we know that the source file hasn’t been modified since it was last indexed and thus, no work needs to be done – this is essentially because it means we don’t need to launch a type checking process for every source file in the project if it is closed and re-opened. If no unit file exists or if the source file has been modified, the compiler for the source file is launched to update the file’s index.
```
bnbarham, line 29:
```suggestion
When a source file should be indexed, we first check if a unit file for this source file already exists and whether the date of that unit file is later than the last modification date of the source file. If this is the case, we know that the source file hasn’t been modified since it was last indexed and thus, no work needs to be done - saving re-indexing the entire project if it is closed and re-opened. If no unit file exists or if the source file has been modified, the compiler is launched to update the file’s index.

Note that avoiding re-running based on modification date is actually trading accuracy for performance, since it is possible that another file could eg. impact the overload that ends up being resolved.
```

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #1575, opened 2024-07-18
Fix for trivia not preserved after string interpolation conversion

AppAppWorks:
fix an issue in #1551

My bad, I should've tested the code action more thoroughly. The leading trivia of the source `SequenceExprSyntax` is not carried over to the new string literal, resulting in this:

```swift
<code block A>
"" + value + ""
<code block B>
```
->
```swift
<code block A>"\(value)"
<code block B>
```

This fix will preserve the leading trivia, i.e.
```swift
<code block A>
"\(value)"
<code block B>
```


ahoppen:
How was `// code block A` preserved before this change if the leading trivia of the code block weren’t preserved? Or did you mean for `// code block A` to be a statement?

Should we also retain the trailing trivia from `syntax` in the same way?

AppAppWorks:
Yeah,  `// code block A` was meant for a statement instead of a comment, I've changed that in the description above.

AFAIK, string literals don't have trailing trivia, and trailing trivia of any non-string-literal component will be converted and moved into the string interpolation, so the trailing trivia of `syntax` has been preserved in this way.

ahoppen:
What about `"a" + b + "c" / * some comment */`? In that case `/* some comment */` should be trailing trivia for the `"`.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1022,9 +1024,10 @@ final class CodeActionTests: XCTestCase {
             changes: [
               uri: [
                 TextEdit(
-                  range: positions["1️⃣"]..<positions["5️⃣"],
+                  range: positions["0️⃣"]..<positions["5️⃣"],
                   newText: ###"""
-                    ##"[\##(key): \##(d) \##(value)]"##
+
```
ahoppen, line 1029: Looks like this line has trailing whitespace. Could you run swift-format on your changes? https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting

AppAppWorks: Done, please test again.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1576, opened 2024-07-18
Import stdlib_h module to get `free`

ian-twilightcoder:
stdlib.h is moving to a top level module in several SDKs, and is no longer transitively imported from Csourcekitd. Exlicitly import stdlib_h where available.


ian-twilightcoder:
@swift-ci test

ian-twilightcoder:
@swift-ci test

ian-twilightcoder:
@swift-ci test

ahoppen (commented):
I would prefer to import `Darwin` because it’s more in-line with the import to Glibc, Musl etc.

https://github.com/swiftlang/sourcekit-lsp/pull/1577

ian-twilightcoder:
> I would prefer to import `Darwin` because it’s more in-line with the import to Glibc, Musl etc.
> 
> #1577

Android and musl have stdlib_h too. Darwin is just so heavyweight. But whatever you prefer 🤷‍♂️

PR #1595, opened 2024-07-30
Fix an infinite recursion when printing the redacted description of `SourceKitLSPOptions`

ahoppen:


ahoppen:
@swift-ci Please test

PR #1596, opened 2024-07-31
Remove null byte after home directory on Windows

ahoppen:
This works around https://github.com/apple/swift-corelibs-foundation/issues/5041, which caused a null byte to be incorrectly added after the user name's home directory, causing, among others, directory iteration to recurse indefinitely, always adding another level of the user's name.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

PR #1603, opened 2024-08-01
Fix build warnings

ahoppen:
Fix build warnings both when building with Swift 6.0 and with Swift from `main` on Linux and macOS.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1616, opened 2024-08-05
Type-qualify accessors in call hierarchy

ahoppen:
An accessor is a child of the corresponding variable, which is why we failed to compute the correct container for them. Look one `childOf` relation further for accessors to fix this.

rdar://129452601


hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1618, opened 2024-08-07
Add a test case that we get build settings for a SwiftPM package if package hasn't been resolved

ahoppen:
When a package had a package dependency but we haven’t resolved packages, we should still be able to get build settings for the existing files – just without the dependencies. This behavior broken by https://github.com/swiftlang/swift-package-manager/pull/7854 and is being fixed by https://github.com/swiftlang/swift-package-manager/pull/7865.


xedin:
https://github.com/swiftlang/swift-package-manager/pull/7865
@swift-ci please test

PR #1619, opened 2024-08-07
Use `firstRange` instead of implementing `firstIndex`

ahoppen:
We have `firstRange` in the stdlib now, so we don’t need to implement `firstIndex` ourselves.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1620, opened 2024-08-08
Re-arrange members in `SwiftPMBuildSystem` by groups

ahoppen:
I think this makes it a little easier to understand what `SwiftPMBuildSystem` stores.


ahoppen:
@swift-ci Please test

PR #1629, opened 2024-08-15
Implicitly cancel text document requests when the document is edited or closed

ahoppen:
As a user makes an edit to a file, these requests are most likely no longer relevant. It also makes sure that a long-running sourcekitd request can't block the entire language server if the client does not cancel all requests. For example, consider the following sequence of requests:
 - `textDocument/semanticTokens/full` for document A
 - `textDocument/didChange` for document A
 - `textDocument/formatting` for document A

If the editor is not cancelling the semantic tokens request on edit (like VS Code does), then the `didChange` notification is blocked on the semantic tokens request finishing. Hence, we also can't run the `textDocument/formatting` request. Cancelling the semantic tokens on the edit fixes the issue.

rdar://133987424

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -177,7 +177,10 @@ package actor SourceKitLSPServer {
   /// The requests that we are currently handling.
   ///
   /// Used to cancel the tasks if the client requests cancellation.
-  private var inProgressRequests: [RequestID: Task<(), Never>] = [:]
+  private var inProgressRequestsByID: [RequestID: Task<(), Never>] = [:]
+
+  /// For all currently handled text document requests a mapping from the document to the corresponding request ID.
+  private var inProgressTextDocumentRequests: [DocumentURI: Set<RequestID>] = [:]
 
   /// Up to 10 request IDs that have recently finished.
   ///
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1238,6 +1257,38 @@ extension SourceKitLSPServer {
...
+    // from performing more work, we execute it with a high priority.
+    cancellationMessageHandlingQueue.async(priority: .high) {
+      await self.cancelTextDocumentRequestsImpl(for: uri)
+    }
+  }
+
+  private func cancelTextDocumentRequestsImpl(for uri: DocumentURI) {
+    guard self.options.cancelTextDocumentRequestsOnEditAndCloseOrDefault else {
+      return
+    }
```
bnbarham, line 1285: Could just check in the outer method before queuing anything?

bnbarham (approved):

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: No, because `options` is isolated to the `SourceKitLSPServer` instance and the outer `cancelTextDocumentRequests` method is nonisolated.

PR #1636, opened 2024-08-19
Add an extra percent encoding layer when encoding DocumentURIs to LSP requests

ahoppen:
The URI standard RFC 3986 is ambiguous about whether percent encoding and their represented characters are considered equivalent. VS Code considers them equivalent and treats them the same:

```js
vscode.Uri.parse("x://a?b=xxxx%3Dyyyy").toString() -> 'x://a?b%3Dxxxx%3Dyyyy'
vscode.Uri.parse("x://a?b=xxxx%3Dyyyy").toString(/*skipEncoding=*/true) -> 'x://a?b=xxxx=yyyy'
```

This causes issues because SourceKit-LSP's macro expansion URLs encoded by URLComponents use `=` do denote the separation of a key and a value in the outer query. The value of the `parent` key may itself contain query items, which use the escaped form '%3D'. Simplified, such a URL may look like `scheme://host?parent=scheme://host?line%3D2`.

But after running this through VS Code's URI type `=` and `%3D` get canonicalized and are indistinguishable.

To avoid this ambiguity, always percent escape the characters we use to distinguish URL query parameters, producing the following URL: `scheme://host?parent%3Dscheme://host%3Fline%253D2`.

---

EDIT: This should be a temporary fix. After some URI spec reading, I’ve concluded that the URI decoding in VS Code is incorrect (https://github.com/swiftlang/vscode-swift/pull/1017#pullrequestreview-2246431175 and https://github.com/swiftlang/vscode-swift/pull/1017#issuecomment-2297944976). 

I would like to merge this PR for now to enable semantic functionality in nested macro expansions for @lokesh-tr’s GSoC project and the investigate a proper fix for URI de/encoding on the VS Code side after which we can revert this workaround.

### DIFF
--- a/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
+++ b/Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
@@ -84,7 +84,19 @@ public struct DocumentURI: Codable, Hashable, Sendable {
   }
 
   public init(from decoder: Decoder) throws {
-    try self.init(string: decoder.singleValueContainer().decode(String.self))
+    let string = try decoder.singleValueContainer().decode(String.self)
+    guard let url = URL(string: string) else {
+      throw FailedToConstructDocumentURIFromStringError(string: string)
+    }
+    if url.query() != nil, var urlComponents = URLComponents(string: url.absoluteString) {
+      // See comment in `encode(to:)`
+      urlComponents.percentEncodedQuery = urlComponents.percentEncodedQuery!.removingPercentEncoding
+      if let rewrittenQuery = urlComponents.url {
+        self.init(rewrittenQuery)
+        return
+      }
+    }
+    self.init(url)
   }
 
   /// Equality check to handle escape sequences in file URLs.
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Tests/LanguageServerProtocolTests/CodingTests.swift
```diff
@@ -1343,6 +1343,29 @@ final class CodingTests: XCTestCase {
...
     )
   }
+
+  func testDocumentUriQueryParameterCoding() throws {
+    checkCoding(
+      try DocumentURI(string: "scheme://host?parent=scheme://host?line%3D2"),
+      json: #"""
+        "scheme:\/\/host?parent%3Dscheme:\/\/host%3Fline%253D2"
+        """#
+    )
```
lokesh-tr, line 1353: Can you add one more test case where the parent also has a parent?
It would also help understand how this change will handle such a case.

Sources/LanguageServerProtocol/SupportTypes/DocumentURI.swift
```diff
@@ -97,7 +109,36 @@ public struct DocumentURI: Codable, Hashable, Sendable {
...
-    try storage.absoluteString.encode(to: encoder)
+    let urlToEncode: URL
+    if let query = storage.query(percentEncoded: true), var components = URLComponents(string: storage.absoluteString) {
+      // The URI standard RFC 3986 is ambiguous about whether percent encoding and their represented characters are
+      // considered equivalent. VS Code considers them equivalent and treats them the same:
+      //
+      // vscode.Uri.parse("x://a?b=xxxx%3Dyyyy").toString() -> 'x://a?b%3Dxxxx%3Dyyyy'
+      // vscode.Uri.parse("x://a?b=xxxx%3Dyyyy").toString(/*skipEncoding=*/true) -> 'x://a?b=xxxx=yyyy'
+      //
+      // This causes issues because SourceKit-LSP's macro expansion URLs encoded by URLComponents use `=` do denote the
```
lokesh-tr, line 123:
```suggestion
      // This causes issues because SourceKit-LSP's macro expansion URLs encoded by URLComponents use `=` to denote the
```

lokesh-tr:
@ahoppen Thanks for this fix, I will test this out and give you an update.

lokesh-tr:
@ahoppen This PR in combination with my semantic functionality [PR](https://github.com/swiftlang/sourcekit-lsp/pull/1634) fixes the issue and also makes semantic functionality to work as expected without needing any change in vscode-swift extension repo.

🎉 

Thank you so much for fixing this

Tests/LanguageServerProtocolTests/CodingTests.swift
ahoppen: Added a test case. Let me know if you think it makes sense.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/LanguageServerProtocolTests/CodingTests.swift
lokesh-tr: That makes sense to me 👍🏻

hamishknight (approved):

PR #1638, opened 2024-08-20
Fix the build using a Swift 5.10 compiler

ahoppen:
`host(percentEncoded:)` is not available using Swift 5.10 on Linux.

Bare slash regex literals are not enabled in Swift 5 mode.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1639, opened 2024-08-22
Migrate `getReferenceDocument` to upstream `textDocumentContent` request

fwcd:
Support for retrieving documents via custom URI schemes has recently been upstreamed and will be supported in the upcoming version 3.18 of the Language Server Protocol:

- https://github.com/microsoft/language-server-protocol/issues/336
- https://github.com/microsoft/language-server-protocol/pull/1994

sourcekit-lsp current uses a non-standard `workspace/getReferenceDocument` LSP extension for this, specifically for retrieving macro expansions. Since this requires either client-side support or workarounds (e.g. generation of temporary files), this PR migrates our LSP extension to the upstream [`workspace/textDocumentContent`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspace_textDocumentContent) request.

The branch compiles, but still needs testing on the client side and we'll also have to figure out how to declare the corresponding server/client capabilities for this to work, hence why I've marked the PR as draft for now.


ahoppen:
CC @lokesh-tr

lokesh-tr:
Wow! It got merged 14 days back. @fwcd You rock, bro! 🔥

fwcd:
Can confirm that this works (in conjunction with the VSCode extension PR):

image

Sources/LanguageServerProtocol/Requests/TextDocumentContentRequest.swift
```diff
@@ -10,22 +10,22 @@
...
 ///
-/// - Returns: `GetReferenceDocumentResponse` which contains the `content` to be displayed.
+/// - Returns: `TextDocumentContentResponse` which contains the `content` to be displayed.
 ///
 /// ### LSP Extension
 ///
 /// This request is an extension to LSP supported by SourceKit-LSP.
-/// Enable the experimental client capability `"workspace/getReferenceDocument"` so that the server responds with
+/// Enable the experimental client capability `"workspace/textDocumentContent"` so that the server responds with
 /// reference document URLs for certain requests or commands whenever possible.
```
lokesh-tr, line 25: I don't think that the LSP Extension Heading is no longer necessary here. Am I missing something?

Sources/SourceKitLSP/Swift/MacroExpansion.swift
```diff
@@ -218,8 +218,8 @@ extension SwiftLanguageService {
     }
 
     if case .dictionary(let experimentalCapabilities) = self.capabilityRegistry.clientCapabilities.experimental,
-      case .bool(true) = experimentalCapabilities["workspace/peekDocuments"],
-      case .bool(true) = experimentalCapabilities["workspace/getReferenceDocument"]
+      case .bool(true) = experimentalCapabilities["workspace/peekDocuments"]
+      // TODO: Check if client supports LSP 3.18's workspace/textDocumentContent
```
lokesh-tr, line 222: I guess this check should be in place before we merge the PR

lokesh-tr (commented):
I just have a few minor comments but, everything else looks good to me.

One more thought:
Given that the LSP now supports `TextDocumentContentRequest`, I think we should now remove the code that generates temporary files for editors that doesn't support `PeekDocumentsRequest`, and ensure that we generate macro expansions on the fly for that and display them by passing a `ReferenceDocumentURL` to the `ShowDocumentRequest`.

lokesh-tr:
@fwcd  As per what we discussed, I guess it would be nice to mark this as "Draft" (or atleast with some "Do NOT Merge") just to ensure that this doesn't get merged before the LSP spec gets finalised.

Sources/LanguageServerProtocol/Requests/TextDocumentContentRequest.swift
fwcd: No, you're right. Thanks!

ahoppen:
> Given that the LSP now supports `TextDocumentContentRequest`, I think we should now remove the code that generates temporary files for editors that doesn't support `PeekDocumentsRequest`, and ensure that we generate macro expansions on the fly for that and display them by passing a `ReferenceDocumentURL` to the `ShowDocumentRequest`.

I think it’s still valuable to have that logic for editors that don’t support the `workspace/textDocumentContent` request, which will probably be most editors for now. All these newer requests are optional and it’s possible for editors to not declare that capability.


> As per what we discussed, I guess it would be nice to mark this as "Draft" (or atleast with some "Do NOT Merge") just to ensure that this doesn't get merged before the LSP spec gets finalised.

Just so I understand, what was the reasoning why we want to wait for LSP 3.18 to get finalized before merging this? Would there be any harm in switching to the `workspace/textDocumentContent` request now, implement the middleware to handle in in VS Code and then if other editors start implementing LSP 3.18 early, great and if not, we aren’t off any worse than we are right now?

fwcd:
We could do that, but I wasn't sure if using a beta version of `vscode-languageclient` for the extension is acceptable (and I'm not sure if the request ships there before it is stabilized in LSP). If it is, we could do that.

ahoppen:
Oh, I was thinking that the Swift VS Code extension could continue intercepting the `workspace/textDocumentContent` request in the middleware for now with the existing logic. And once VS Code adds native support for `workspace/textDocumentContent`, we can just remove that middleware. Would that work?

fwcd:
That's a nice idea, I don't see why that wouldn't work and as far as I understand we haven't shipped a version of sourcekit-lsp with this yet, so we wouldn't need to keep the legacy request name around for compatibility.

ahoppen:
Exactly 😉

fwcd:
I'm marking the PR as ready for review again, since I have successfully tested it together with https://github.com/swiftlang/vscode-swift/pull/1050, which just adapts our existing client-side integration to the new request name/format.

Contributor Documentation/LSP Extensions.md
```diff
@@ -595,30 +595,3 @@ export interface PeekDocumentsResult {
   success: boolean;
 }
 ```
-
-## `workspace/getReferenceDocument`
```
ahoppen, line 599: The `LanguageServerProtocol` module currently represents LSP 3.17, so I would still consider  `workspace/textDocumentContent` an extension. Just one that happens to line up with 3.18. Could you add that to the `LSP Extensions` document and mention that it matches the 3.18 request?

Sources/LanguageServerProtocol/Requests/TextDocumentContentRequest.swift
```diff
@@ -0,0 +1,38 @@
...
+  public var uri: DocumentURI
+
+  public init(uri: DocumentURI) {
+    self.uri = uri
+  }
+}
+
+/// Response containing `text` of `TextDocumentContentRequest`
+public struct TextDocumentContentResponse: ResponseType {
+  public var text: String
```
ahoppen, line 33: Could you include the doc comment from LSP here?

```
/// The text content of the text document. Please note, that the content of
/// any subsequent open notifications for the text document might differ
/// from the returned content due to whitespace and line ending
/// normalizations done on the client.
```

Sources/SourceKitLSP/Swift/MacroExpansion.swift
ahoppen: Could you address this TODO? I don’t like having TODOs in the codebase without an associated issue because in practice they never get fixed.

ahoppen (commented):
Looks great. Just a few minor comments.

Sources/SourceKitLSP/Swift/MacroExpansion.swift
fwcd: I've re-added this and added a short remark to the PR description for how this check has to be updated once we wish to support the upstream client capability.

fwcd:
Tested with the VSCode extension and could successfully verify that it works. I had to re-add some of the experimental capability logic in 99836e08c3e362cae6876e6e678a531fc73fce93, but we'll need that anyway until the request is fully supported by upstream LSP.

ahoppen (commented):
Thanks!

I was wondering how we got away without the capabilities in `initializationOptions` but I took your word for it 😆 But guess it means that we do need them 🤷🏽

ahoppen (approved):

fwcd:
> I was wondering how we got away without the capabilities

By removing the capability check, which I still did under the assumption that we'd hold off with this PR until the request is landed upstream (i.e. where I wrote the todo comment) 😄

ahoppen:
@swift-ci Please test Windows

lokesh-tr:
This is great! 🚀 🚀 🚀

PR #1640, opened 2024-08-23
Print error when creation of a sourcekitd request from YAML fails in the `debug run-sourcekitd-request` command

ahoppen:


ahoppen:
@swift-ci Please test

PR #1643, opened 2024-08-28
Use SwiftPM's SDK computation logic

kabiroberai:
Using `SwiftSDK.deriveTargetSwiftSDK`, which is the same method that SwiftPM's own CLI tools use to determine the SDK from passed-in info (target `--triple`, `--swift-sdk`, and host sdk). This allows us to better uphold the contract in the [Configuration File](https://github.com/swiftlang/sourcekit-lsp/blob/d11c101ce210ad23f917cf06fb13a7d9e243872b/Documentation/Configuration%20File.md#structure) docs, namely that the `swiftSDK` param is "Equivalent to SwiftPM's `--swift-sdk` option" and similarly for `triple`.

As concrete examples of where (AFAICT) the current implementation diverges:
- Passing a `--triple` of `wasm32-unknown-wasi` to `swift-build` will use the toolchain-integrated Wasm SDK if one exists. Passing the same value to sourcekit-lsp does not do this.
- Perhaps more relevant: after landing https://github.com/swiftlang/swift-package-manager/pull/6828, this change will make it so that building for iOS is as simple as setting `"triple": "arm64-apple-ios"` in the config! Currently, it's necessary to set C/Swift flags and hardcode the sysroot. Should close https://github.com/swiftlang/sourcekit-lsp/issues/1587.

This PR depends on:

- https://github.com/swiftlang/swift-package-manager/pull/7925

But I feel that the two PRs are worth discussing as one since this use case is the primary motivating factor. Will flesh out the SwiftPM PR if the SourceKit/SwiftPM teams agree that this is the right strategy.

### DIFF
--- a/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
+++ b/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
@@ -226,25 +226,25 @@ package actor SwiftPMBuildSystem {
     let hostSDK = try SwiftSDK.hostSwiftSDK(AbsolutePath(destinationToolchainBinDir))
     let hostSwiftPMToolchain = try UserToolchain(swiftSDK: hostSDK)
 
-    var destinationSDK: SwiftSDK
-    if let swiftSDK = options.swiftPM.swiftSDK {
-      let bundleStore = try SwiftSDKBundleStore(
-        swiftSDKsDirectory: fileSystem.getSharedSwiftSDKsDirectory(
-          explicitDirectory: options.swiftPM.swiftSDKsDirectory.map { try AbsolutePath(validating: $0) }
-        ),
-        fileSystem: fileSystem,
-        observabilityScope: observabilitySystem.topScope,
-        outputHandler: { _ in }
-      )
-      destinationSDK = try bundleStore.selectBundle(matching: swiftSDK, hostTriple: hostSwiftPMToolchain.targetTriple)
-    } else {
-      destinationSDK = hostSDK
-    }
+    let bundleStore = try SwiftSDKBundleStore(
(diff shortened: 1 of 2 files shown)


kabiroberai:
GitHub won't let me request reviewers but cc @kateinoigakukun @MaxDesiatov

MaxDesiatov (approved):

MaxDesiatov:
https://github.com/swiftlang/swift-package-manager/pull/7925

@swift-ci test

ahoppen (commented):
Thank you @kabiroberai! Do you think you could write a test to make sure we don’t regress this in the future? Probably something like [`SwiftPMIntegrationTests.testWasm`](https://github.com/swiftlang/sourcekit-lsp/blob/main/Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift#L249) but that utilizes the Wasm SDK in a way that fails before your PR and succeeds after? Maybe it would be sufficient to just update that tests.

kabiroberai:
@ahoppen added a test! Confirmed that it fails prior to applying the changes in this PR. Don't think it can be an end-to-end integration test because afaik we'd need to provide a real WASI sysroot for that, but I added logic to test that we calculate the correct SDK path.

kateinoigakukun:
https://github.com/swiftlang/swift-package-manager/pull/7925

@swift-ci test

ahoppen (approved):
Thanks for the test @kabiroberai!

ahoppen:
https://github.com/swiftlang/swift-package-manager/pull/7925

@swift-ci Please test Windows

MaxDesiatov:
@swift-ci test

PR #1645, opened 2024-09-06
Log full options with which SourceKit-LSP is initialized

ahoppen:
Log the resolved options of the used `config.json` file as well as the initialize request and response. I have found that seeing these options is generally useful to debug issues with the SourceKit-LSP + Editor setup.

A nice side effect is that we can now also log the non-sensitive parts of requests and responses sent between SourceKit-LSP and the editor.

### DIFF
--- a/Sources/SKLogging/CustomLogStringConvertible.swift
+++ b/Sources/SKLogging/CustomLogStringConvertible.swift
@@ -64,7 +64,8 @@ extension String {
   /// A hash value that can be logged in a redacted description without
   /// disclosing any private information about the string.
   package var hashForLogging: String {
-    return Insecure.MD5.hash(data: Data(self.utf8)).description
+    let hash = SHA256.hash(data: Data(self.utf8)).prefix(8).map { String(format: "%02x", $0) }.joined()
+    return "<private \(hash)>"
   }
 }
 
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

Sources/SKLogging/CustomLogStringConvertible.swift
```diff
@@ -86,58 +87,51 @@ extension Optional where Wrapped: CustomLogStringConvertible {
...
-        if let value = value as? any CustomLogStringConvertible {
-          value.redactedDescription
-        } else {
-          recursiveRedactedDescription(of: value)
-        }
-      return "\(label): \(value)"
+  package var prettyPrintedRedactedJSON: String {
+    func redact(subject: Any) -> Any {
+      if let subject = subject as? [String: Any] {
+        return subject.mapValues { redact(subject: $0) }
```
hamishknight, line 109: In principle you could have a dictionary keyed by a private value e.g a file path string, I guess that's not a concern currently?

Sources/SKLogging/SplitLogMessage.swift
```diff
@@ -34,3 +36,34 @@ package func splitLongMultilineMessage(message: String, maxChunkSize: Int = 800)
   }
   return chunks
 }
+
+extension Logger {
+  /// Implementation detail of
```
hamishknight, line 41: Incomplete comment?

Sources/SKLogging/SplitLogMessage.swift
```diff
@@ -10,14 +10,16 @@
 //
 //===----------------------------------------------------------------------===//
 
+private let maxChunkSize = 800
```
hamishknight, line 13: Doesn't seem this is used outside of `splitLongMultilineMessage`, was factoring out a leftover change from a previous revision?

hamishknight (approved):

Sources/SKLogging/SplitLogMessage.swift
ahoppen: Yes, I think so. Moved it to `splitLongMultilineMessage`.

Sources/SKLogging/CustomLogStringConvertible.swift
ahoppen: Yes, I’m not aware of any requests that store sensitive data in the keys.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1646, opened 2024-09-06
Remove custom decoding function from `SourceKitLSPOptions`

ahoppen:
We forgot to decode the following keys in the custom decode function, which meant that you couldn’t set them using SourceKit-LSP’s `config.json` file.
- `backgroundPreparationMode`
- `sourcekitdRequestTimeout`
- `cancelTextDocumentRequestsOnEditAndClose`

We had the custom decoder function so that the keys weren’t required in the JSON but we could access eg. `SwiftPMOptions` without needing to deal with optionals in the codebase.

Make the accesses to these nested options structs a little more verbose but eliminate the source of the above bug, which seems like a good tradeoff.

### DIFF
--- a/Sources/SKOptions/SourceKitLSPOptions.swift
+++ b/Sources/SKOptions/SourceKitLSPOptions.swift
@@ -20,8 +20,8 @@ import struct TSCBasic.AbsolutePath
 /// Options that can be used to modify SourceKit-LSP's behavior.
 ///
 /// See `ConfigurationFile.md` for a description of the configuration file's behavior.
-public struct SourceKitLSPOptions: Sendable, Codable, CustomLogStringConvertible {
-  public struct SwiftPMOptions: Sendable, Codable, CustomLogStringConvertible {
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -259,12 +259,38 @@ public struct SourceKitLSPOptions: Sendable, Codable, CustomLogStringConvertible
     case enabled
   }
 
-  public var swiftPM: SwiftPMOptions
-  public var compilationDatabase: CompilationDatabaseOptions
-  public var fallbackBuildSystem: FallbackBuildSystemOptions
+  // Pretend that all of these keys are
```
bnbarham, line 262: Are....

ahoppen: Oh, leftover comment. Removed it.

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #1652, opened 2024-09-09
Fix a race condition that could cause the build graph to not be generated when doing initial background indexing

ahoppen:
…

We were making the initial `generateBuildGraph` call in `SourceKitLSPServer` from a `Task`. This means that `generateBuildGraph` could be executed after `waitForUpToDateBuildGraph` was called by `SemanticIndexManager`. Thus `waitForUpToDateBuildGraph` returned immediately and no files were background indexed.

Make `generateBuildGraph` immediately schedule a package reload for SwiftPM build systems instead of doing the hop through a `Task`, fixing the race condition.

rdar://135551812


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1653, opened 2024-09-10
Create `BuiltInBuildSystem` in `BuildSystemAdapter`

ahoppen:
We currently create the `BuiltInBuildSystem` while constructing a `Workspace` and then pass it into the `BuildSystemManager`, which will do some final wiring up, setting itself as the message handler and delegate of the build system. 

This PR moves the creation of the `BuiltInBuildSystem` into the `BuiltInBuildSystemAdapter`. That way, the creation of a `BuiltInBuildSystem` is more like the initialization of a BSP server. 

To do so, we need to split the creation of a build system (now happening in `BuiltInBuildSystemAdapter`) and determining what kind of build system to use for a workspace (still happening during workspace creation). I have found that this split was generally beneficial and made the code easier to reason about.

### DIFF
--- a/Sources/BuildSystemIntegration/BuildServerBuildSystem.swift
+++ b/Sources/BuildSystemIntegration/BuildServerBuildSystem.swift
@@ -15,6 +15,7 @@ import Foundation
 import LanguageServerProtocol
 import LanguageServerProtocolJSONRPC
 import SKLogging
+import SKOptions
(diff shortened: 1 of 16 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Will address review comments in follow-up PRs.

PR #1657, opened 2024-09-10
Allow overriding the timeout duration for tests

ahoppen:


ahoppen:
@swift-ci Please test

PR #1659, opened 2024-09-11
Migrate `fileHandlingCapability` and `prepare` to BSP and remove `registerForChangeNotifications`

ahoppen:
Three somewhat unrelated commits:
- Instead of having `FileHandlingCapability` for a source file, check if it belongs to any targets
- Remove `registerForChangeNotifications`
  - This was only really used for the legacy BSP integration, which we are phasing out in favor of the new `SourceKitOptions` request.
- Migrate `BuildSystem.prepare` to a BSP request


ahoppen:
@swift-ci Please test

PR #1663, opened 2024-09-12
Migrate `defaultLanguage(for:)` and `filesDependenciesUpdated` to BSP and use `InitializeRequest` from BSP to communicate static options

ahoppen:


ahoppen:
@swift-ci Please test

PR #1676, opened 2024-09-16
Only reopen files for which the build settings actually changed

ahoppen:
Whenever we got a notification that the build settings of a file might have changed (eg. because the Package manifest was reloaded), we would re-open it in sourcekitd. But we only need to re-open it if the build settings actually changed.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #1677, opened 2024-09-16
Require a Swift 6 compiler to build SourceKit-LSP

ahoppen:
Xcode 16 with Swift 6 has been released, we can drop support for building and testing SourceKit-LSP using a Swift 5.10 toolchain. This allows us to remove a number of workarounds.


hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1680, opened 2024-09-17
Mark all transitive dependents of a modified target as being unprepared

ahoppen:
We weren’t considering transitive dependencies in `BuildSystemManager.targets(dependingOn:)` before.

rdar://136039234


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1682, opened 2024-09-18
Fix an issue that caused us to not get compiler arguments when opening a SwiftPM package at a symlink

ahoppen:
When you had a package at `/pkg_real` and a symlink `/pkg` pointing to `/pkg_real`, then opened a workspace at `/pkg`, we wouldn’t get any build settings for any of the files. This was masked in tests because they still called `SwiftPMBuildSystem.targets(for:)`, which handled symlink resolution but wasn’t called in production.

Handle symlink resolution `BuildSystemManager`, remove `SwiftPMBuildSystem.targets(for:)` and its related members/methods, and migrate the tests to go through `BuildSystemManager`, which is what production code does. A nice side effect of this is that the tests will log the requests sent to the build system.

### DIFF
--- a/Sources/BuildSystemIntegration/BuildSystemManager.swift
+++ b/Sources/BuildSystemIntegration/BuildSystemManager.swift
@@ -21,6 +21,7 @@ import SwiftExtensions
 import ToolchainRegistry
 
 import struct TSCBasic.AbsolutePath
+import func TSCBasic.resolveSymlinks
 
 fileprivate typealias RequestCache<Request: RequestType & Hashable> = Cache<Request, Request.Response>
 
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -78,6 +79,18 @@ fileprivate extension BuildTarget {
   }
 }
 
+fileprivate extension DocumentURI {
+  /// If this is a file URI pointing to a symlink, return the realpath of the URI, otherwise return `nil`.
+  var symlinksResolved: DocumentURI? {
```
hamishknight, line 84: IMO `symlinksResolved` doesn't really communicate "returns `nil` if not a symlink", maybe it should be something more like `symlinkTarget`?

hamishknight (approved):

Sources/BuildSystemIntegration/BuildSystemManager.swift
ahoppen: Good idea 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1683, opened 2024-09-18
Adopt faster `traverseModules` implementation in SwiftPM

ahoppen:
Companion of https://github.com/swiftlang/swift-package-manager/pull/7977

---

Adjustments because `traverseModules` no longer returns depth

We no longer need `depth`. Adopt a faster version of `traverseModules` that doesn’t return the depth.

rdar://136107035


hamishknight (approved):

ahoppen:
https://github.com/swiftlang/swift-package-manager/pull/7977

@swift-ci Please test

PR #1684, opened 2024-09-19
Don’t send messages to the BSP server before the `build/initialize` request returns

ahoppen:
We currently allow sending requests to the BSP server before `build/initialize` returns. This is not valid in general and we should wait for the initialize result before sending any other messages.

rdar://136111469


hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1685, opened 2024-09-19
Only scan test targets for tests

ahoppen:
Don’t run the syntactic test scanner on files that we know are only part of non-test targets.

rdar://126493903

### DIFF
--- a/Sources/SwiftExtensions/Sequence+AsyncMap.swift
+++ b/Sources/SwiftExtensions/Sequence+AsyncMap.swift
@@ -81,4 +81,11 @@ extension Sequence {
 
     return nil
   }
+
+  /// Just like `Sequence.contains` but allows an `async` predicate function.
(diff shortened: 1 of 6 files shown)


Sources/SwiftExtensions/Sequence+AsyncMap.swift
```diff
@@ -81,4 +81,11 @@ extension Sequence {
 
     return nil
   }
+
+  /// Just like `Sequence.first` but allows an `async` predicate function.
```
hamishknight, line 85:
```suggestion
  /// Just like `Sequence.contains` but allows an `async` predicate function.
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1688, opened 2024-09-19
When a JSON-RPC connection is closed, send an error response to all outstanding requests

ahoppen:
Otherwise, we would never get a reply for these requests. I noticed this while developing BSP server tests that had Python errors, which caused the BSP server to crash.


ahoppen:
@swift-ci Please test

hamishknight (approved):

MaxDesiatov:
This seems to be breaking toolchain builds:

```
JSONRPCConnection.swift:168:35: error: reference to property 'outstandingRequests' in closure
requires explicit use of 'self' to make capture semantics explicit
        for outstandingRequest in outstandingRequests.values {
```

https://ci.swift.org/job/swift-PR-toolchain-macos/1519/console

PR #1691, opened 2024-09-19
Remove VS Code tasks.json to run all tests in parallel

ahoppen:
VS Code now has a command to run all tests in parallel, so we don’t need this anymore.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1689, opened 2024-09-19
Expose all BSP messages used by SourceKit-LSP to BSP servers

ahoppen:
The interaction to an out-of-process BSP server still went through the `BuildServerBuildSystem`, which doesn’t forward all messages to the build system and uses the old push-based model for build settings.

If we discover that the BSP server supports the new pull-based build settings model, we now forward all methods to it directly, without going through `BuiltInBuildSystemAdapter`, which has been renamed to `LegacyBuildServerBuildSystem`.

rdar://136106323
rdar://127606323
rdar://126493405
Fixes #1226
Fixes #1173

PR #1701, opened 2024-09-21
Restart a BSP server after it has crashed 🚥 #1689

### LINKED ISSUE #1686: Handle BSP server crashes
When a BSP server crashes, we should try to restart it, similar to how we restart clangd.

ahoppen:
Depends on #1689

---

If a BSP server crashes, we should try to relaunch it. Use a similar relaunch logic as the one we use for `clangd`: Try restarting immediately unless the BSP server has crashed within the last 30 seconds, in which case we delay the restart by 10s.

rdar://136311009
Fixes #1686


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #1702, opened 2024-09-21
Consistently use `SOURCEKITLSP_` as prefix of environment variables instead of `SOURCEKIT_LSP`

ahoppen:
We were inconsistently using `SOURCEKITLSP_` and `SOURCEKIT_LSP_` as the prefix for environment variables. Make it consistent and always use `SOURCEKITLSP_`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "swift.swiftEnvironmentVariables": {
     // Show log messages when running tests on macOS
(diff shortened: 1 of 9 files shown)


hamishknight (approved):
IMO `SOURCEKIT_LSP` is more readable (and more consistent with there being a separator for `sourcekit-lsp`), but I don't feel too strongly 🤷

ahoppen:
I was split 50/50 as well. Changed it to consistently use `SOURCEKIT_LSP_` now.

ahoppen:
@swift-ci Please test

PR #1703, opened 2024-09-21
When running tests, raise in `XCTFail` if `logger.fault` is called 🚥 #1701

### LINKED ISSUE #1669: When running tests, raise in `XCTFail` if `logger.fault` is called
`logger.fault` indicates that there was an internal error in SourceKit-LSP. We should ensure that we don’t hit any situation that causes a `logger.fault` while running tests by raising an `XCTFail` if `logger.fault` is called in `NonDarwinLogger` (which is the logger that we’re using in Swift CI).

ahoppen:
Depends on #1701

---

`logger.fault` indicates that there was an internal error in SourceKit-LSP. We should ensure that we don’t hit any situation that causes a `logger.fault` while running tests by raising an `XCTFail` if `logger.fault` is called in `NonDarwinLogger` (which is the logger that we’re using in Swift CI).

rdar://136014548
Resolves #1669

### DIFF
--- a/Documentation/Environment Variables.md
+++ b/Documentation/Environment Variables.md
@@ -5,6 +5,7 @@ The following environment variables can be used to control some behavior in Sour
 ## Build time
 
 - `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER`: Use the `NonDarwinLogger` to log to stderr, even when building SourceKit-LSP on macOS. This is useful when running tests using `swift test` because it writes the log messages to stderr, which is displayed during the `swift test` invocation.
+- `SOURCEKITLSP_XCTFAIL_ON_FAULT`: When a fault is logged, record it using `XCTFail`, verifying that no faults are logged during test execution, since faults indicate Must only be set when running tests. A `sourcekit-lsp` binary built with this flag will not launch because it can’t load XCTest. Has no effect if os_log is used (ie. when building on macOS without `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER` set).
 - `SOURCEKITLSP_CI_INSTALL`: Modifies rpaths in a way that’s necessary to build SourceKit-LSP to be included in a distributed toolchain. Should not be used locally.
 - `SWIFTCI_USE_LOCAL_DEPS`: Assume that all of SourceKit-LSP’s dependencies are checked out next to it and use those instead of cloning the repositories. Primarily intended for CI environments that check out related branches.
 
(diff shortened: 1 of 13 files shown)


Documentation/Environment Variables.md
```diff
@@ -5,6 +5,7 @@ The following environment variables can be used to control some behavior in Sour
 ## Build time
 
 - `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER`: Use the `NonDarwinLogger` to log to stderr, even when building SourceKit-LSP on macOS. This is useful when running tests using `swift test` because it writes the log messages to stderr, which is displayed during the `swift test` invocation.
+- `SOURCEKITLSP_XCTFAIL_ON_FAULT`: When a fault is logged, record it using `XCTFail`, verifying that no faults are logged during test execution, since faults indicate Must only be set when running tests. A `sourcekit-lsp` binary built with this flag will not launch because it can’t load XCTest. Has no effect if os_log is used (ie. when building on macOS without `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER` set).
```
hamishknight, line 8: Nit: "because" makes it sound unconditional

```suggestion
- `SOURCEKITLSP_XCTFAIL_ON_FAULT`: When a fault is logged, record it using `XCTFail`, verifying that no faults are logged during test execution, since faults indicate Must only be set when running tests. A `sourcekit-lsp` binary built with this flag will not launch if it can’t load XCTest. Has no effect if os_log is used (ie. when building on macOS without `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER` set).
```

Sources/SKLogging/NonDarwinLogging.swift
```diff
@@ -12,6 +12,40 @@
 
 import SwiftExtensions
 
+#if SOURCEKITLSP_XCTFAIL_ON_FAULT
+import XCTest
+
+private let allowFault = AtomicBool(initialValue: false)
```
hamishknight, line 18: Do we need to worry about race conditions if e.g one thread enters `withLoggingFaultsAllowed`, and another thread then leaves `withLoggingFaultsAllowed`, leaving this set to `false` for the first thread?

hamishknight (approved):

ahoppen:
Decided against this, mostly because it requires us to modify global state and even though https://github.com/swiftlang/sourcekit-lsp/pull/1703#discussion_r1776023254 shouldn’t pose an issue right now, I’m not sure if that might become problematic in the future.

PR #1706, opened 2024-09-22
Revert "When a JSON-RPC connection is closed, send an error response to all outstanding requests"

MaxDesiatov:
Reverts swiftlang/sourcekit-lsp#1688

This seems to be breaking toolchain builds:

```
JSONRPCConnection.swift:168:35: error: reference to property 'outstandingRequests' in closure
requires explicit use of 'self' to make capture semantics explicit
        for outstandingRequest in outstandingRequests.values {
```

https://ci.swift.org/job/swift-PR-toolchain-macos/1519/console

PR #1707, opened 2024-09-23
When a JSON-RPC connection is closed, send an error response to all outstanding requests

ahoppen:
Re-apply #1688 after it was reverted by #1706.


ahoppen:
@swift-ci Please test

PR #1711, opened 2024-09-25
Set `isExecuting` to `true` before yielding the execution task

ahoppen:
There was a very brief time when a `QueuedTask` would start executing but `isExecuting` was still sent to `false`. This caused the `preparationTask.isExecuting` check in `SemanticIndexManager.prepare` to fail after we did start preparation, causing the preparation of `LibB` in `testDontStackTargetPreparationForEditorFunctionality` to be cancelled, which wasn’t intended. Thus, the expected preparation tracker got out of sync and the test would fail.

rdar://136570212


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1716, opened 2024-09-26
Specify explicit `self` in `ExternalBuildSystemAdapter`

ahoppen:
I’m not entirely sure why but the file fails to compile with multiple arches without the `self`.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1717, opened 2024-09-26
Build sourcekit-lsp for multiple arches and lipo them

ahoppen:
Companion of https://github.com/swiftlang/swift/pull/76730

---

Instead of building SourceKit-LSP using SwiftPM's multi-arch xcbuild backend, build it for only one arch at a time and then run `lipo` to merge the two resulting binaries.

This should allow us to share build products between building installing and testing and also eliminates other quirks resulting from the xcbuild backend.


ahoppen:
https://github.com/swiftlang/swift/pull/76730

@swift-ci Please test

PR #1719, opened 2024-09-27
Explicitly shut down the build server when SourceKit-LSP is shut down

ahoppen:
The build server is automatically shut down using a background task when `BuildSystemManager` is deallocated.

This, however, leads to possible race conditions where the shutdown task might not finish before the test is done, which could result in the connection being reported as a leak. To avoid this problem, we want to explicitly shut, down the build server when the `SourceKitLSPServer` gets shut down.

### DIFF
--- a/Sources/BuildSystemIntegration/BuildSystemManager.swift
+++ b/Sources/BuildSystemIntegration/BuildSystemManager.swift
@@ -430,6 +430,23 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
     }
   }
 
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1721, opened 2024-09-27
Improve `BuildServerBuildSystemTests.testCrashRecovery`

ahoppen:
The original idea behind this test was that only opening `Crash.swift` would retrieve build settings for `Crash.swift`. But that wasn’t actually true: Scanning for tests also retrieved build settings (to get the test’s module name). This caused test failures if we set `SOURCEKIT_LSP_TEST_TIMEOUT: 10`, like we do in `settings.json`.

Improve the test to actually check that the BSP server has crashed and increase timeout to be bigger than the 10s restart delay.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1723, opened 2024-09-27
Revert "Use fallback build settings if build system doesn’t provide build settings within a timeout"

ahoppen:
This reverts commit https://github.com/swiftlang/sourcekit-lsp/pull/1700 because it was causing non-deterministic CI failure when hitting the build settings timeout when running tests.


ahoppen:
@swift-ci Please test

PR #1728, opened 2024-09-30
Update BSP connection build server config lookup path

### LINKED ISSUE #1695: Read BSP connection details from `.bsp` directory in addition to `buildServer.json`
SourceKit-LSP currently looks for a BSP server config file in `<workspaceRoot>/buildServer.json`. BSP specifies that these config files should be located in `<workspace-dir>/.bsp` (among other locations). We should adopt the `.bsp` search locations.

Rajveer100:
Resolves #1695

Adopt `<workspace_root>/.bsp` search locations in addition to `<workspace_root>/`.

### DIFF
--- a/Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
+++ b/Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
@@ -52,6 +52,10 @@ struct ExecutableNotFoundError: Error {
   let executableName: String
 }
 
+enum BuildServerNotFoundError: Error {
+  case fileNotFound
+}
+
 private struct BuildServerConfig: Codable {
   /// The name of the build tool.
   let name: String
@@ -94,7 +98,7 @@ actor ExternalBuildSystemAdapter {
   private var lastRestart: Date?
 
   static package func projectRoot(for workspaceFolder: AbsolutePath, options: SourceKitLSPOptions) -> AbsolutePath? {
-    guard localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json")) else {
+    guard getConfigPath(for: workspaceFolder) != nil else {
       return nil
     }
     return workspaceFolder
@@ -142,7 +146,10 @@ actor ExternalBuildSystemAdapter {
 
   /// Create a new JSONRPCConnection to the build server.
   private func createConnectionToBspServer() async throws -> JSONRPCConnection {
-    let configPath = projectRoot.appending(component: "buildServer.json")
+    guard let configPath = ExternalBuildSystemAdapter.getConfigPath(for: self.projectRoot) else {
+      throw BuildServerNotFoundError.fileNotFound
+    }
+
     let serverConfig = try BuildServerConfig.load(from: configPath)
     var serverPath = try AbsolutePath(validating: serverConfig.argv[0], relativeTo: projectRoot)
     var serverArgs = Array(serverConfig.argv[1...])
@@ -176,6 +183,62 @@ actor ExternalBuildSystemAdapter {
         }
       }
     ).connection
+  }
+
+  private static func getConfigPath(for workspaceFolder: AbsolutePath? = nil) -> AbsolutePath? {
+    var buildServerConfigLocations: [URL?] = []
+    if let workspaceFolder = workspaceFolder {
+      buildServerConfigLocations.append(workspaceFolder.appending(component: ".bsp").asURL)
+    }
+
+    #if os(Windows)
+    if let localAppData = ProcessInfo.processInfo.environment["LOCALAPPDATA"] {
+      buildServerConfigLocations.append(URL(fileURLWithPath: localAppData).appendingPathComponent("bsp"))
+    }
+    if let programData = ProcessInfo.processInfo.environment["PROGRAMDATA"] {
+      buildServerConfigLocations.append(URL(fileURLWithPath: programData).appendingPathComponent("bsp"))
+    }
+    #else
+    if let xdgDataHome = ProcessInfo.processInfo.environment["XDG_DATA_HOME"] {
+      buildServerConfigLocations.append(URL(fileURLWithPath: xdgDataHome).appendingPathComponent("bsp"))
+    }
+
+    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
+      buildServerConfigLocations.append(libraryUrl.appendingPathComponent("bsp"))
+    }
+
+    if let xdgDataDirs = ProcessInfo.processInfo.environment["XDG_DATA_DIRS"] {
+      buildServerConfigLocations += xdgDataDirs.split(separator: ":").map { xdgDataDir in
+        URL(fileURLWithPath: String(xdgDataDir)).appendingPathComponent("bsp")
+      }
+    }
+
+    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .systemDomainMask).first {
+      buildServerConfigLocations.append(libraryUrl.appendingPathComponent("bsp"))
+    }
+    #endif
+
+    for case let buildServerConfigLocation? in buildServerConfigLocations {
+      let jsonFiles =
+        try? FileManager.default.contentsOfDirectory(at: buildServerConfigLocation, includingPropertiesForKeys: nil)
+        .filter { $0.pathExtension == "json" }
+
+      if let configFileURL = jsonFiles?.sorted(by: { $0.lastPathComponent < $1.lastPathComponent }).first,
+        let configFilePath = AbsolutePath(validatingOrNil: configFileURL.path)
+      {
+        return configFilePath
+      }
+    }
+
+    // Pre Swift 6.1 SourceKit-LSP looked for `buildServer.json` in the project root. Maintain this search location for
+    // compatibility even though it's not a standard BSP search location.
+    if let workspaceFolder = workspaceFolder,
+      localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json"))
+    {
+      return workspaceFolder.appending(component: "buildServer.json")
+    }
+
+    return nil
   }
 
   /// Restart the BSP server after it has crashed.
--- a/Sources/SKTestSupport/BuildServerTestProject.swift
+++ b/Sources/SKTestSupport/BuildServerTestProject.swift
@@ -55,11 +55,12 @@ private let skTestSupportInputsDirectory: URL = {
 package class BuildServerTestProject: MultiFileTestProject {
   package init(
     files: [RelativeFileLocation: String],
+    buildServerConfigLocation: RelativeFileLocation = ".bsp/sourcekit-lsp.json",
     buildServer: String,
     testName: String = #function
   ) async throws {
     var files = files
-    files["buildServer.json"] = """
+    files[buildServerConfigLocation] = """
       {
         "name": "client name",
         "version": "10",
--- a/Tests/BuildSystemIntegrationTests/BuildServerBuildSystemTests.swift
+++ b/Tests/BuildSystemIntegrationTests/BuildServerBuildSystemTests.swift
@@ -305,6 +305,61 @@ final class BuildServerBuildSystemTests: XCTestCase {
       )
       return diagnostics.fullReport?.items.map(\.message) == ["DEBUG SET"]
     }
+  }
 
+  func testBuildServerConfigAtLegacyLocation() async throws {
+    let project = try await BuildServerTestProject(
+      files: [
+        "Test.swift": """
+        #if DEBUG
+        #error("DEBUG SET")
+        #else
+        #error("DEBUG NOT SET")
+        #endif
+        """
+      ],
+      buildServerConfigLocation: "buildServer.json",
+      buildServer: """
+        class BuildServer(AbstractBuildServer):
+          def workspace_build_targets(self, request: Dict[str, object]) -> Dict[str, object]:
+            return {
+              "targets": [
+                {
+                  "id": {"uri": "bsp://dummy"},
+                  "tags": [],
+                  "languageIds": [],
+                  "dependencies": [],
+                  "capabilities": {},
+                }
+              ]
+            }
+
+          def buildtarget_sources(self, request: Dict[str, object]) -> Dict[str, object]:
+            return {
+              "items": [
+                {
+                  "target": {"uri": "bsp://dummy"},
+                  "sources": [
+                    {"uri": "$TEST_DIR_URL/Test.swift", "kind": 1, "generated": False}
+                  ],
+                }
+              ]
+            }
+
+          def textdocument_sourcekitoptions(self, request: Dict[str, object]) -> Dict[str, object]:
+            return {
+              "compilerArguments": ["$TEST_DIR/Test.swift", "-DDEBUG", $SDK_ARGS]
+            }
+        """
+    )
+
+    let (uri, _) = try project.openDocument("Test.swift")
+
+    try await repeatUntilExpectedResult {
+      let diags = try await project.testClient.send(
+        DocumentDiagnosticsRequest(textDocument: TextDocumentIdentifier(uri))
+      )
+      return diags.fullReport?.items.map(\.message) == ["DEBUG SET"]
+    }
   }
 }


ahoppen (commented):
https://build-server-protocol.github.io/docs/overview/server-discovery#default-locations-for-bsp-connection-files specifies a number of other search locations that we should also adopt.

Also, I don’t see any mention of the `buildServer.json` file so we shouldn’t hardcode that. For the lack of a better idea, I think we should pick the lexicographically first file in one of the search directories that is a valid BSP configuration file.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
   private var lastRestart: Date?
 
   static package func projectRoot(for workspaceFolder: AbsolutePath, options: SourceKitLSPOptions) -> AbsolutePath? {
-    guard localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json")) else {
-      return nil
+    var buildServerConfigLocations = [String]()
+
+    #if os(Windows)
+    let workspacePath = workspaceFolder.pathString + "\\.bsp"
+    let localAppDataPath = ProcessInfo.processInfo.environment["LOCALAPPDATA"] ?? "C:\\Users\\Default\\AppData\\Local\\bsp"
```
ahoppen, line 101: We shouldn’t hardcode any paths. I doubt that a user is named `Default` anyway. If the environment variable doesn’t exist, we shouldn’t search it. Same for the other missing environment variable below.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
   private var lastRestart: Date?
 
   static package func projectRoot(for workspaceFolder: AbsolutePath, options: SourceKitLSPOptions) -> AbsolutePath? {
-    guard localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json")) else {
-      return nil
+    var buildServerConfigLocations = [String]()
+
+    #if os(Windows)
+    let workspacePath = workspaceFolder.pathString + "\\.bsp"
```
ahoppen, line 100: Instead of concatenating strings here, we should use `workspaceFolder.appending(component: ".bsp")` Same below.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
   private var lastRestart: Date?
 
   static package func projectRoot(for workspaceFolder: AbsolutePath, options: SourceKitLSPOptions) -> AbsolutePath? {
-    guard localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json")) else {
-      return nil
+    var buildServerConfigLocations = [String]()
```
ahoppen, line 97:
```suggestion
    var buildServerConfigLocations: [String] = []
```

is the preferred spelling.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
...
+
+    #if os(Windows)
+    let workspacePath = workspaceFolder.pathString + "\\.bsp"
+    let localAppDataPath = ProcessInfo.processInfo.environment["LOCALAPPDATA"] ?? "C:\\Users\\Default\\AppData\\Local\\bsp"
+    let programDataPath = ProcessInfo.processInfo.environment["PROGRAMDATA"] ?? "C:\\ProgramData\\bsp"
+
+    buildServerConfigLocations.append(contentsOf: [workspacePath, localAppDataPath, programDataPath])
+    #elseif os(Linux)
```
ahoppen, line 105: The spec says `Unix + Mac`, so I think we should search for `XDG_DATA_HOME` also on macOS and not have an `os(Linux)` clause.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
...
+    let workspacePath = workspaceFolder.pathString + "/bsp"
+    let xdgDataHomePath = ProcessInfo.processInfo.environment["XDG_DATA_HOME"] ?? FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(".local/share/bsp")
+    let xdgDataDirsPath = ProcessInfo.processInfo.environment["XDG_DATA_DIRS"] ?? "/usr/local/share:/usr/share"
+
+    buildServerConfigLocations.append(contentsOf: [workspacePath, xdgDataHomePath, xdgDataDirsPath])
+    #else
+    let workspacePath = workspaceFolder.pathString + "/bsp"
+    let libraryPath = FileManager.default.homeDirectoryForCurrentUser.appending(components: "Library", "Application Support", "bsp").absoluteString
```
ahoppen, line 113: We don’t need to stitch together the path to `Application Support`. You can use

```swift
FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)
```

or

```swift
FileManager.default.urls(for: .applicationSupportDirectory, in: .systemDomainMask)
```

instead.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -94,10 +94,34 @@ actor ExternalBuildSystemAdapter {
...
+
+    buildServerConfigLocations.append(contentsOf: [workspacePath, libraryPath])
+    #endif
+
+    for buildServerConfigLocation in buildServerConfigLocations {
+      guard let buildServerConfigLocation = try? AbsolutePath(validating: buildServerConfigLocation) else { continue }
```
ahoppen, line 119: You’re only checking if the `.bsp` directory exists, not if there’s a build server in that directory.

ahoppen (commented):
> Regarding picking the file, did you mean to say that if we had multiple `.json` files in different valid directories, we should pick the lexicographically first one?

Kind of. I’m saying that if in a single search location there are multiple `.json` files, we should pick the lexicographically first one. But we should still always prefer a build server config in the workspace `.bsp` folder over one in a system location.

---

Also, you need to update the logic that actually loads the build server config in `createConnectionToBspServer`. Could you also add test cases to make sure that we find BSP server config files in the `<workspace folder>/.bsp` location to `BuildServerBuildSystemTests`?

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: But `.appending` adds a `/` and for Windows isn't `\` the way to separate directories (although the former works in most cases)?

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: Yep.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: Yeah, much better.

Rajveer100:
I am not quite sure about the testing part, which path should be considered as the project workspace root for testing? Also, what do you recommend to invoke `getConfigPath`, do we add `@testable import BuildSystemIntegration` for accessing the private methods?

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: `appending(component:)` should take care of all platform differences in path representations.

ahoppen:
> I am not quite sure about the testing part, which path should be considered as the project workspace root for testing? Also, what do you recommend to invoke getConfigPath, do we add @testable import BuildSystemIntegration for accessing the private methods?

I would create a `MultiFileTestProject` similar to how it’s done in https://github.com/swiftlang/sourcekit-lsp/blob/01640ee68fe946b0c90fb60526c7839b08c20780/Sources/SKTestSupport/BuildServerTestProject.swift#L61-L86, just put the build server configuration in a `.bsp` folder inside the workspace root.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
     ).connection
   }
 
+  private func getConfigPath() -> AbsolutePath? {
```
ahoppen, line 188: Now you’re no longer for the folders in `projectRoot(for workspaceFolder:options:)`. You need to check in both.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
...
+    var buildServerConfigLocations: [AbsolutePath?] = []
+
+    let workspacePath = projectRoot.appending(component: ".bsp")
+    buildServerConfigLocations.append(workspacePath)
```
ahoppen, line 192: Can easily be one operation

```suggestion
    buildServerConfigLocations.append(projectRoot.appending(component: ".bsp"))
```

Same below.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
...
+
+    #if os(Windows)
+    let localAppDataPath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["LOCALAPPDATA"])
+    let programDataPath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["PROGRAMDATA"])
+
```
ahoppen, line 197: You need to search `%LOCALAPPDATA%\bsp\`, not `%LOCALAPPDATA%`. Same for `%PROGRAMDATA%`, `$XDG_DATA_HOME`, `$XDG_DATA_DIRS` and the macOS search locations.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
...
+    #else
+    let xdgDataHomePath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["XDG_DATA_HOME"])
+    let xdgDataDirsPath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["XDG_DATA_DIRS"])
+
+    buildServerConfigLocations.append(contentsOf: [xdgDataHomePath, xdgDataDirsPath])
+
+    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
+      let libraryPath = AbsolutePath(validatingOrNil: libraryUrl.absoluteString)
+      buildServerConfigLocations.append(libraryPath)
+    }
```
ahoppen, line 208: We should keep the ordering here. `$XDG_DATA_HOME/bsp/` should be before `$HOME/Library/Application Support/bsp/` and that before `$XDG_DATA_DIRS/bsp/`, just in case somebody has the `XDG` variables set on macOS.

Rajveer100: That's exactly how it's ordered, I have appended `XDG` paths before the `library` paths. So if it finds in either of them it's returned first.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: Let me know if the updated logic is fine, I will then proceed with the tests.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: You need to add `bsp` to the path returned by the environment variable, not the environment variable’s key.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: No, it’s not. According to the spec, if you have a build server in `$HOME/Library/Application Support/bsp/` and also in `$XDG_DATA_DIRS/bsp/`, we should pick the one in `$HOME/Library/Application Support/bsp/`. But with the current implementation, we pick the one in `$XDG_DATA_DIRS/bsp/`.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: Yes, this path looks fine now. I would encourage you to write tests alongside the implementation because you would have easily found this issue with a test case.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: > We should keep the ordering here. $XDG_DATA_HOME/bsp/ should be before $HOME/Library/Application Support/bsp/

I thought you meant to prefer `XDG` before `library` when you say "before"? I think I misunderstood.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
Rajveer100: Indeed.

Rajveer100:
@ahoppen
I have added the changes.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
...
+    let programDataPath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["PROGRAMDATA"])?.appending(component: "bsp")
+
+    buildServerConfigLocations.append(contentsOf: [localAppDataPath, programDataPath])
+    #else
+    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
+      let libraryPath = AbsolutePath(validatingOrNil: libraryUrl.absoluteString)?.appending(component: "bsp")
+      buildServerConfigLocations.append(libraryPath)
+    }
+
+    let xdgDataHomePath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["XDG_DATA_HOME"])?.appending(component: "bsp")
```
ahoppen, line 205:
```suggestion
    let xdgDataHomePath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["XDG_DATA_HOME"])?.appending(component: "bsp")

    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
      let libraryPath = AbsolutePath(validatingOrNil: libraryUrl.absoluteString)?.appending(component: "bsp")
      buildServerConfigLocations.append(libraryPath)
    }
```

Should be the other way round to follow the spec.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -178,6 +185,48 @@ actor ExternalBuildSystemAdapter {
...
+    let xdgDataDirsPath = AbsolutePath(validatingOrNil: ProcessInfo.processInfo.environment["XDG_DATA_DIRS"])?.appending(component: "bsp")
+
+    buildServerConfigLocations.append(contentsOf: [xdgDataHomePath, xdgDataDirsPath])
```
ahoppen, line 208: You should still search the system application support after `XDG_DATA_DIRS`

```swift
    if let libraryUrl = FileManager.default.urls(for: .applicationSupportDirectory, in: .systemDomainMask).first {
      let libraryPath = AbsolutePath(validatingOrNil: libraryUrl.absoluteString)?.appending(component: "bsp")
      buildServerConfigLocations.append(libraryPath)
    }
```

Tests/BuildSystemIntegrationTests/BuildServerBuildSystemTests.swift
```diff
@@ -303,4 +303,65 @@ final class BuildServerBuildSystemTests: XCTestCase {
...
+        #if DEBUG
+        #error("DEBUG SET")
+        #else
+        #error("DEBUG NOT SET")
+        #endif
+        """,
+        ".bsp/test.json": """
+          {
+          }
+        """
```
ahoppen, line 322: That’s some interesting JSON formatting with `{` starting at an indentation level of 2 spaces. I would just write it in one line

```suggestion
        ".bsp/test.json": "{}"
```

Sources/SKTestSupport/BuildServerTestProject.swift
```diff
@@ -59,7 +59,7 @@ package class BuildServerTestProject: MultiFileTestProject {
     testName: String = #function
   ) async throws {
     var files = files
-    files["buildServer.json"] = """
+    files[".bsp/buildServer.json"] = """
```
ahoppen, line 62: Could you also keep a test around that validates that we keep finding the build server config if it’s placed at `<workspace root>/buildServer.json`. We need to keep supporting that for compatibility reasons.

Rajveer100: Did you mean to keep the old logic:

```swift
localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json"))
```

ahoppen: Ye

Rajveer100: Currently we have multiple `.json` files in the following paths:

- `<workspace root>/buildServer.json`
- `<workspace root>/.bsp/buildServer.json`

In the `.bsp` directory I have included the following too:

```json
".bsp/test.json": "{}"
```

So, which is the preferred location the path as per the old logic or the new one? 

Since, now the test currently fails due to the presence of `test.json` in `.bsp` which is empty and even though we have one in `<workspace root>` it's not used as `getConfigPath` returns early.

ahoppen: > So, which is the preferred location the path as per the old logic or the new one?

We should prefer files in `<workspace root>/.bsp`, then look at `<workspace root>/buildServer.json` and then continue looking at the system locations

> As now the test currently fails due to the presence of test.json in .bsp which isn't a valid config file and even though we have one in <workspace root> it's not used as getConfigPath returns early.

I would say this is expected. In case you have a malformed build server config file, we don’t silently fall to system files, which would be surprising, I think.

Rajveer100: Ok, let me know if any further changes are needed.

ahoppen:
I have updated your test case to make it pass, made a few more adjustments and pushed the changes to your branch.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Rajveer100:
Thanks @ahoppen!

PR #1730, opened 2024-09-30
When `withTimeout` hits a timeout, don’t wait for cooperative cancellation before returning

### LINKED ISSUE #883: If request is cancelled, stop waiting for a response from sourcekitd/clangd
When a request is cancelled, we currently request cancellation in clangd/sourcekitd but still wait for the result. This means that we might still block the message handling queue for a request whose result we no longer care about.

What we should do instead is to immediately return a cancellation error for the cancelled request and no longer care about what sourcekitd/clangd returns to us.

ahoppen:
Previously, when we hit the timeout in `withTimeout`, we just cancelled `body` but relied on cooperative cancellation before the caller could continue. Since the caller is obviously no longer interested in the operation, we can return with `CancellationError` immediately.

Fixes #883
rdar://116705684


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

ktoso:
Heh, you might want to consider renaming this to withUnstructuredTimeout or something like that. Giving up structured concurrency should be explicit I think as you lost priority escalation here.

ahoppen:
Good catch, @ktoso. I implemented priority propagation now.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ktoso:
That does the trick with the hot loop around task.priority... tbh I'd rather just admit this drops structured concurrency, but if you've happy with this then your call.

We should provide a built-in operation for this soon, the while-true loop is very wasteful 😭

ahoppen:
I would be very happy to see native implementations of all the async utilities I have in here. But until then, I’ve got to implement them in the best way I can…

PR #1737, opened 2024-10-04
In `withTaskPriorityChangedHandler` execute the task that watches for priority changes with high priority

ahoppen:
I saw a nondeterministic test failure of `AsyncUtilsTests.testWithTimeoutEscalatesPriority` and I believe this was the cause.

Also refactor a few test helper functions on the way.


ahoppen:
@swift-ci Please test

PR #1738, opened 2024-10-05
Remove elements from `outstandingRequests` on crash

ahoppen:
Otherwise, we can have a retain cycle where the `OutstandingRequest` keeps the `JSONRPCConnection` alive.

Also move the code that handled `outstandingRequests` out of the `Task` because `outstandingRequests` should only be accessed from `queue`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1743, opened 2024-10-08
Yield `TimeoutError` before cancelling body task in `withTimeout`

ahoppen:
Otherwise, the body task might finish before we yield the `TimeoutError` and thus `withTimeout` would not actually throw a `TimeoutError`.

rdar://137171114


ahoppen:
@swift-ci Please test

PR #1747, opened 2024-10-09
Use the test’s default timeout duration for sourcekitd timeouts

ahoppen:
Otherwise, we are stuck with the timeout default of 120s and the test runner can’t increase it.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1752, opened 2024-10-11
Highlight in non-Apple extended logging that the section does not app ly to macOS

ahoppen:
Helpful if you glance over the heading.


ahoppen:
@swift-ci Please test

PR #1760, opened 2024-10-15
Make `logHandler` print to stdout instead of stderr

ahoppen:
For some reason, logging to `stderr` makes test execution hang on Windows when running tests in parallel – it appears that stdout and stderr are handled the same way but there must be some difference, I wasn’t able to reduce the issue.

When running the sourcekit-lsp binary, stdout gets redirected to stderr, so logging printing to stdout ends up with the same result as `fputs` to `stderr`.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1762, opened 2024-10-16
Use fallback build settings if build system doesn’t provide build settings within a timeout

### LINKED ISSUE #1693: Return fallback build settings if build system doesn’t return build settings within a timeout
If the build system doesn’t return build settings within a pre-defined timeout, use fallback build settings. Notify `SourceKitLSPServer` that settings have changed once we get build settings from the build system.

ahoppen:
When we receive build settings after hitting the timeout, we call `fileBuildSettingsChanged` on the delegate, which should cause the document to get re-opened in sourcekitd and diagnostics to get refreshed.

rdar://136332685
Fixes #1693

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -259,7 +259,7 @@ package actor SwiftLanguageService: LanguageService, Sendable {
     }
   }
 
-  func buildSettings(for document: DocumentURI) async -> SwiftCompileCommand? {
+  func buildSettings(for document: DocumentURI, fallbackAfterTimeout: Bool) async -> SwiftCompileCommand? {
     let primaryDocument = document.primaryFile ?? document
 
     guard let sourceKitLSPServer else {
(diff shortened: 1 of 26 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Ran the test suite 20 times locally as-is and 20 times with a 1s delay before getting build settings from the build system. This should rule out any nondeterministic failures.

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -917,7 +920,7 @@ extension SwiftLanguageService {
 
   func retrieveQuickFixCodeActions(_ params: CodeActionRequest) async throws -> [CodeAction] {
     let snapshot = try await self.latestSnapshot(for: params.textDocument.uri)
-    let buildSettings = await self.buildSettings(for: params.textDocument.uri)
+    let buildSettings = await self.buildSettings(for: params.textDocument.uri, fallbackAfterTimeout: true)
```
bnbarham, line 923: Diagnostics is false (though clangd publish diagnostics is true), do we want fixes to be true?

ahoppen: Yes, because Fix-Its are delivered as part of code actions (which include Fix-Its + refactorings). Using a fallback here allows us to deliver syntactic refactorings even if we don’t have build settings.

bnbarham (approved):

PR #1763, opened 2024-10-16
Log to stdout on non-Windows platforms

ahoppen:
We need to print to stdout on Windows because printing to stderr causes XCTest execution to deadlock when running tests in parallel on Windows.

But printing to stdout causes XCTest to interleave process output with its own output, which creates hard-to-read log. To avoid this, print to stderr on all other platforms.


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
Closing in favor of https://github.com/swiftlang/sourcekit-lsp/pull/1765

PR #1764, opened 2024-10-16
Support `workspace/diagnostic/refresh` request from `TestSourceKitLSPClient`

ahoppen:
Currently, we return a `methodNotFound` error when we get a diagnostic refresh request from SourceKit-LSP. This doesn’t do any harm but draws the attention when reading test logs, despite being the expected result. Clean up that output by returning a proper response.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1766, opened 2024-10-16
Make `DocumentManager.documents` private

ahoppen:
`documents` needs to be guarded by `queue` and thus shouldn’t be accessible from outside the type. Also uncovered on incorrect use of it.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1768, opened 2024-10-17
Support building only tests using SwiftPM

ahoppen:
This allows us to build everything required for the executable using CMake and only the test targets using SwiftPM, enabling us to run tests on Windows from build.ps1 with minimal overhead.

Analogous to https://github.com/swiftlang/swift-format/pull/851

### DIFF
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name SourceKitLSP>")
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name sourcekit_lsp>")
 add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-DRESILIENT_LIBRARIES>")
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name SourceKitLSP>")
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name sourcekit_lsp>")
```
bnbarham, line 1: Why the switch?

bnbarham (approved):

Sources/CMakeLists.txt
ahoppen: When building the tests using SwiftPM, this is the package name that SwiftPM will chose and it means that we can use declarations at the package level from the modules built using CMake.

PR #1774, opened 2024-10-21
Use `withUnsafeFileSystemRepresentation` to get the path of a URL on disk

ahoppen:
`URL.path` returns forward slashes in the path on Windows (https://github.com/swiftlang/swift-foundation/issues/973) where we expect backslashes. Work around that by defining our own `filePath` property that is backed by `withUnsafeFileSystemRepresentation`, which produces backslashes.

rdar://137963660


bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #1776, opened 2024-10-23
Lower `swift-tools-version` to 5.9 in `Package.swift`

MaxDesiatov:
This would unblock https://github.com/swiftlang/swift-tools-support-core/pull/488


ahoppen:
SourceKit-LSP does not build with Swift 5.9 and should thus not have a package version of 5.9. https://github.com/swiftlang/swift-tools-support-core/pull/489 should fix the CI failure in TSC.

PR #1778, opened 2024-10-24
Increase the duration in which we allow index cancellation to happen

ahoppen:
Slow CI systems weren’t able to resolve the package and prepare the target in 30s. Increase the timeout to 90s to make them pass.


ahoppen:
@swift-ci Please test

bnbarham (approved):
Oof.

PR #1781, opened 2024-10-25
Don’t rely on timers for BSP target / settings change tests

ahoppen:
The timers were racy and on Windows we would get the updated build targets / build settings before we expect them. Change the tests to explicitly tell the build server when to change the targets / settings it reports by sending it a `workspace/didChangeWatchedFiles` notification.

This should also speed up the tests because we don’t have a 1s timer anymore.


ahoppen:
@swift-ci Please test

PR #1782, opened 2024-10-25
Fix nondeterministic test failure for tests that await next diagnostic notification

ahoppen:
These tests were designed to poll for the next diagnostic notification and if that notification didn’t contain the expected result, try again.

But if we didn’t get any diagnostic notification in 1s, we would unconditionally fail instead of trying again, which was not intended.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

PR #1785, opened 2024-10-25
When writing an updated package manifest in tests fails, try again

ahoppen:
Package resolving can open Package.swift in exclusive mode on Windows, which prevents us from writing the new package manifest. Keep retrying until we get a successful write. This matches what a user would do.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

Issue #1788, opened 2024-10-27
Refine automatic rewriting of trailing closures


woolsweater:
## Context

Following from the rationale given in <https://github.com/swiftlang/sourcekit-lsp/issues/1787>, whether to expand a trailing closure is not a syntactical decision; it depends on the function's semantics. As discussed, predicate-based functions (`map`, `filter`) are one case where an expansion is often inappropriate. Since the introduction of `async` functions we can also see callback-passing style become less common/prominent altogether, meaning that the relative proportion of predicate-like calls is increasing.

In the case of multiple closures, it is again a stylistic choice whether to expand _all_ of them, or only some. Sometimes only the last should be expanded; e.g., this is a perfectly common style in SwiftUI code:

``` swift
Button(action: { self.viewModel.makeItDo() }) {
  Text("Tap me!")
    .other(.modifiers)
    .etc(.etc)
}
```

## Proposed Change

Since the expansion logic has limited or no knowledge of a function's semantics, I believe that it should be made much less aggressive. By default, it should always produce a closure **on the same line** as the call, with a **single** placeholder inside the closure: i.e. completion for a call to `func foo(bar: Bar, baz: @escaping () -> Void)` should expand to:

``` swift
foo(baz: <#Baz#>) { <##> }
```

This is a much better experience for the user in the case where they did _not_ want a multi-line expansion, while preserving most of the benefits of the current behavior from <https://github.com/swiftlang/sourcekit-lsp/pull/1072>. A "full expansion" configuration setting could opt the user in to current behavior.

A further refinement might be to use an allowlist for known "callback-like" methods, and produce the multi-line expansion only for them. But since users can always define their own methods with trailing closures, and the LSP server has no way to distinguish the semantics of those cases, the default should remain the single-line form.

ahoppen:
Synced to Apple’s issue tracker as rdar://138732066

PR #1789, opened 2024-10-27
Add config option for trailing closure rewriting

woolsweater:
This would resolve <https://github.com/swiftlang/sourcekit-lsp/issues/1787>

The default remains the automatic expansion introduced in <https://github.com/swiftlang/sourcekit-lsp/pull/1072>, but the new option allows the user to disable it via a configuration file.

This is expressed as an enumeration rather than a boolean flag because there is room for [other levels of rewriting][0]. E.g. for trailing closures, a "basic" level might be implemented as rewriting to a pair of brackets on the same line, with a single placeholder, rather than the "full" multi-line behavior.

A [configuration file][1] like the following will disable the expansion feature:

``` json
{
  "codeCompletion" : {
    "rewriteTrailingClosures" : "never"
  }
}
```

[0]: https://github.com/swiftlang/sourcekit-lsp/issues/1788
[1]: https://github.com/swiftlang/sourcekit-lsp/blob/f25055094789a27af9ecd3b965717bcf5008faa6/Documentation/Configuration%20File.md

### DIFF
--- a/Documentation/Configuration File.md
+++ b/Documentation/Configuration File.md
@@ -39,6 +39,8 @@ The structure of the file is currently not guaranteed to be stable. Options may 
     - `indexPrefixMap: [string: string]`: Path remappings for remapping index data for local use.
     - `maxCoresPercentageToUseForBackgroundIndexing: double`: A hint indicating how many cores background indexing should use at most (value between 0 and 1). Background indexing is not required to honor this setting
     - `updateIndexStoreTimeout: int`: Number of seconds to wait for an update index store task to finish before killing it.
+- `codeCompletion`: Dictionary with the following keys, defining options related to code completion actions
+    - `rewriteTrailingClosures: "full"|"never"`: Whether to pre-expand trailing closures when completing a function call expression
 - `logging`: Dictionary with the following keys, changing SourceKit-LSP’s logging behavior on non-Apple platforms. On Apple platforms, logging is done through the [system log](Diagnose%20Bundle.md#Enable%20Extended%20Logging). These options can only be set globally and not per workspace.
   - `logLevel: "debug"|"info"|"default"|"error"|"fault"`: The level from which one onwards log messages should be written.
   - `privacyLevel: "public"|"private"|"sensitive"`: Whether potentially sensitive information should be redacted. Default is `public`, which redacts potentially sensitive information.
--- a/Sources/SKOptions/SourceKitLSPOptions.swift
+++ b/Sources/SKOptions/SourceKitLSPOptions.swift
@@ -219,6 +219,33 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     }
   }
 
+  /// User settings controlling code completion.
+  public struct CodeCompletionOptions: Sendable, Codable, Equatable {
+    /// The extent to which a completion action should apply stylistic or
+    /// convenience transformations before passing a result to the client.
+    public enum RewriteLevel: String, Sendable, Codable {
+      case full, never
+    }
+
+    /// Whether trailing closures should be eagerly expanded by SourceKit-LSP
+    /// before being passed to the client.
+    public var rewriteTrailingClosures: RewriteLevel
+
+    public init(rewriteTrailingClosures: RewriteLevel = .full) {
+      self.rewriteTrailingClosures = rewriteTrailingClosures
+    }
+
+    static func merging(
+      base: CodeCompletionOptions,
+      override: CodeCompletionOptions?
+    ) -> CodeCompletionOptions {
+      return CodeCompletionOptions(
+        rewriteTrailingClosures: override?.rewriteTrailingClosures
+          ?? base.rewriteTrailingClosures
+      )
+    }
+  }
+
   public enum BackgroundPreparationMode: String {
     /// Build a target to prepare it
     case build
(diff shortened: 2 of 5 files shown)


woolsweater:
@swift-ci Please test

ahoppen (commented):
Thanks for the PR, @woolsweater.  I agree that we are currently using trailing closure syntax for code completion fairly aggressively and that there is a problem to be solved here, but I don’t think that a global configuration option is the right solution here. 

I think that whether you want to use a trailing closure depends less on personal preference and more on the context. For example, I think that in almost all cases a user is expecting `Task.init` to be followed by a trailing closure. For methods like `filter` it usually depends on whether the condition will be complex and span multiple lines or, as you stated in your issue, if you are intending to use a simple condition or key path.

Thus, I think the user should be given the choice of whether to use a trailing closure completion on every completion. The options I see right now are:
1. Offer two different completion options for the trailing closure and the non-trailing closure case. But this might inflate the number of code completion items and make it harder to find the one you are looking for.
2. Revert code completions to suggest snippets (like SourceKit-LSP did in 5.10 and earlier) and then offer a code action to expand snippets to trailing closure, similar to how Xcode expands placeholders to trailing closures when you press enter while they are selected. This may be tricky to implement because editors don’t represent snippets in a file’s source text, so SourceKit-LSP doesn’t know if the user is currently at a snippet that was inserted to represent a closure, unless SourceKit-LSP keeps track of that information itself, which I guess it could.
3. Make code completion expand snippets to single-line closures but not trailing closures (e.g expand to `filter(where: { ${1:Int} in ${2:Bool} })` instead of `filter { ${1:Int} in \n${2:Bool}\n }`). This would be a compromise from where you can go into either direction: (a) Deleting the closure to replace it with a function reference isn’t too much effort since it only spas a single line, (b) similarly using anonymous closure argument names (`$0` etc) is fairly straightforward, (c) to go to trailing closure syntax, we already have a code action for that, we would just need to make that refactoring more fault tolerant and be applicable in case there is a closure with snippets.
4. Continue expanding to trailing closure syntax like we do now and offer a code action to convert trailing closure syntax back to non-trailing closure syntax. The tricky part here is inferring the label for the function call, eg. the where to convert `firstIndex { … }` to `firstIndex(where: { … })`

While writing this, I think that I’m preferring option (2) or (3). What do you think, @woolsweater?

woolsweater:
> depends […] more on the context. For example, I think that in almost all cases a user is expecting Task.init to be followed by a trailing closure.

Thanks @ahoppen! Yes, absolutely; there are cases where it would be very unusual not to use a trailing closure. But I think there are **more** cases where the language server can't read the user's mind and can't know which it should choose. I agree that (2) and (3) are the best paths forward 🙌 It seems to me that they could mostly coexist, actually.

Having a code action to expand a call to trailing closures would be a great thing to have regardless of how the snippet expansion works. I would think the server could get around the "was this a snippet?" problem by simply offering it on any call expression where there is a closure literal present? That would make it useful even when revisiting the code long after the current LSP session is closed. But perhaps that is unworkable for some reason?

I saw this PR as a stepping stone to a state like your (3). I do genuinely think there's an element of user preference here: for example, I assume the author of <https://github.com/swiftlang/sourcekit-lsp/issues/854> likes the current implementation, and I didn't want to break their experience. So they and anyone else who prefers it can use the _full_ setting and get that behavior. The (new default) _medium_ or _basic_ level would be your (3).¹ And the curmudgeons can turn it off altogether with _never_. I think it's going to be impossible for the language server to please everyone in this regard. So having a good default with a little adjustability seems ideal.

I see that reworking the expansion requires changes to swift-syntax, and I have not gotten a chance to dig into that, but I was planning to do so this coming weekend. I would be happy to try to implement your item (3) as a resolution of <https://github.com/swiftlang/sourcekit-lsp/issues/1788> unless you would prefer to do it.

---

¹ And perhaps there's even a _high_ setting between that and _full_...but then someone's go to figure out what that means and then implement it, so maybe not 😅

ahoppen:
If you would be interested in implementing (3) to resolve #1788, that would be great. I’m happy to help with any concrete questions you might have. 

Regarding an option: I would prefer to try and avoid options as long as we don’t have two opinions that fundamentally oppose each other. Every option needs to be tested and with a growing number of configurations, the chance of missing some broken configuration rises. For now, I think that (2) or (3) offer a good middle ground and I would like to explore if they can please the majority before jumping to a configuration parameter.

woolsweater:
> Every option needs to be tested and with a growing number of configurations…

Yes, that's fair and understandable. Okay, I will put this idea aside and work on the swift-syntax changes. Is simply changing the behavior of `ExpandEditorPlaceholdersToTrailingClosures` the right approach? Are there any other clients of it besides sourcekit-lsp?

ahoppen:
> Yes, that's fair and understandable. Okay, I will put this idea aside and work on the swift-syntax changes. 
Thank you!

> Is simply changing the behavior of `ExpandEditorPlaceholdersToTrailingClosures` the right approach? Are there any other clients of it besides sourcekit-lsp?

SourceKit-LSP is the main client of `ExpandEditorPlaceholdersToTrailingClosures` and I assume that any changes to it will also benefit any other clients that might exist.

woolsweater:
Sorry for the delay, @ahoppen ; I did not get to this as soon as I'd hoped. I have the changes just about ready to go now. I will have to make a PR to swift-syntax, to change the expansion behavior, and then one to sourcekit-lsp to adopt the swift-syntax update.

Is there a good way for me to make a draft PR to sourcekit-lsp, while the swift-syntax PR is pending, that includes my swift-syntax changes? Should I just change the pin in sourcekit-lsp's Package.resolved to point to my branch, and then update it after the swift-syntax merge? Or do you have another preferred way to handle this?

ahoppen:
> Is there a good way for me to make a draft PR to sourcekit-lsp, while the swift-syntax PR is pending, that includes my swift-syntax changes? Should I just change the pin in sourcekit-lsp's Package.resolved to point to my branch, and then update it after the swift-syntax merge? Or do you have another preferred way to handle this?

Just open a PR to swift-syntax and reference it from this PR. We can do cross-PR testing in CI.

woolsweater:
Okay, I've opened https://github.com/swiftlang/swift-syntax/pull/2897 and https://github.com/swiftlang/sourcekit-lsp/pull/1831 for the change discussed here.

PR #1790, opened 2024-10-28
Fix race condition that caused `testMainFileChangesIfIncludeIsAdded` to fail

ahoppen:
In some situations, we could return the timeout error from the timeout task, but receive a notification from the `self.notifications` `AsyncStream`. That notification would then be dropped and never get delivered to the test case, which can cause test cases to fail.

### DIFF
--- a/Sources/SKTestSupport/TestSourceKitLSPClient.swift
+++ b/Sources/SKTestSupport/TestSourceKitLSPClient.swift
@@ -248,21 +248,26 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
   /// - Note: This also returns any notifications sent before the call to
   ///   `nextNotification`.
   package func nextNotification(timeout: Duration = .seconds(defaultTimeout)) async throws -> any NotificationType {
-    return try await withThrowingTaskGroup(of: (any NotificationType).self) { taskGroup in
-      taskGroup.addTask {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/TestSourceKitLSPClient.swift
```diff
@@ -248,21 +248,26 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
...
+    // After `timeout`, we tell `notificationYielder` that we are no longer interested in its result by cancelling it.
+    // We wait for `notificationYielder` to accept this cancellation instead of returning immediately to avoid a
+    // situation where `notificationYielder` continues running, eats the first notification but it then never gets
+    // delivered to the test because we already delivered a timeout.
+    let cancellationTask = Task {
+      try await Task.sleep(for: timeout)
+      notificationYielder.cancel()
+    }
+    defer {
+      // We have received a value and don't need teh cancellationTask anymore
```
hamishknight, line 267:
```suggestion
      // We have received a value and don't need the cancellationTask anymore
```

hamishknight (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1794, opened 2024-10-30
Fix nondeterministic test failure in `testDependenciesUpdatedSwift`

ahoppen:
With https://github.com/swiftlang/sourcekit-lsp/pull/1762 we use fallback build settings if the build system doesn’t provide real build systems quickly. This may cause us to initially return empty syntactic diagnostics in this test case, which causes it to fail. Change it to assert that we eventually get the expected semantic diagnostics.

rdar://138913004


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

hamishknight (approved):

PR #1795, opened 2024-10-30
Reduce the number of log messages to log the translation for token types and token modifiers from clangd to SourceKit-LSP

ahoppen:
This used to produce ~40 log entries, which was pretty spammy. Compact them into 1


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1798, opened 2024-10-31
Fix a crash when trying to apply in edit that has out of line positions

ahoppen:
`LineTable.replace` did not actually validate that the edit is in range. This could cause crashes of SourceKit-LSP if the editor is sending us bogus edits. Validate the position, like we do in all the other position conversions and log a fault if the edit is out-of-range.

While fixing this, I found a couple more places where line table accesses were not properly guarded. I migrated them to safe alternatives as well.

rdar://138962353

### DIFF
--- a/Sources/SKSupport/LineTable.swift
+++ b/Sources/SKSupport/LineTable.swift
@@ -36,25 +36,16 @@ package struct LineTable: Hashable, Sendable {
     }
   }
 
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

plemarquand:
@swift-ci Please test

PR #1804, opened 2024-11-04
Don’t log “Did not compute dependents for target”

ahoppen:
Leaf targets don’t have dependents and that’s expected. We shouldn’t log a fault for them.


ahoppen:
@swift-ci Please test

PR #1806, opened 2024-11-05
Handle `DocumentRangeFormattingRequest`

### LINKED ISSUE #1805: Handle "textDocument/rangeFormatting" and "textDocument/onTypeFormatting"
### Description

Now that swift-format has some support for range formatting, sourcekit-lsp can implement handling these requests as well.

This will elevate the VSCode formatting experience since the VSCode extension is calling through to sourcekit-lsp for formatting.

sourcekit-lsp seems to already have _some_ support for these. Just the logic isn't implemented.
For example there are some existing `DocumentRangeFormattingRequest` and `DocumentOnTypeFormattingRequest` types.

MahdiBM:
partially resolves #1805

### DIFF
--- a/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
+++ b/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
@@ -136,7 +136,26 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
 
 extension SwiftLanguageService {
   package func documentFormatting(_ req: DocumentFormattingRequest) async throws -> [TextEdit]? {
-    let snapshot = try documentManager.latestSnapshot(req.textDocument.uri)
+    return try await format(
+      textDocument: req.textDocument,
+      options: req.options
+    )
+  }
+
+  package func documentRangeFormatting(_ req: DocumentRangeFormattingRequest) async throws -> [TextEdit]? {
+    return try await format(
+      textDocument: req.textDocument,
+      options: req.options,
+      range: req.range
+    )
+  }
+
+  private func format(
+    textDocument: TextDocumentIdentifier,
+    options: FormattingOptions,
+    range: Range<Position>? = nil
+  ) async throws -> [TextEdit]? {
+    let snapshot = try documentManager.latestSnapshot(textDocument.uri)
 
     guard let swiftFormat else {
       throw ResponseError.unknown(
@@ -144,12 +163,19 @@ extension SwiftLanguageService {
       )
     }
 
-    let process = TSCBasic.Process(
-      args: swiftFormat.pathString,
+    var args = try [
+      swiftFormat.pathString,
       "format",
       "--configuration",
-      try swiftFormatConfiguration(for: req.textDocument.uri, options: req.options)
-    )
+      swiftFormatConfiguration(for: textDocument.uri, options: options),
+    ]
+    if let range {
+      args += [
+        "--offsets",
+        "\(snapshot.utf8Offset(of: range.lowerBound)):\(snapshot.utf8Offset(of: range.upperBound))",
+      ]
+    }
+    let process = TSCBasic.Process(arguments: args)
     let writeStream = try process.launch()
 
     // Send the file to format to swift-format's stdin. That way we don't have to write it to a file.
--- a/Tests/SourceKitLSPTests/RangeFormattingTests.swift
+++ b/Tests/SourceKitLSPTests/RangeFormattingTests.swift
@@ -1,0 +1,120 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKLogging
+import SKTestSupport
+import SourceKitLSP
+import XCTest
+
+final class RangeFormattingTests: XCTestCase {
+  func testOnlyFormatsSpecifiedLines() async throws {
+    try await SkipUnless.toolchainContainsSwiftFormat()
+    let testClient = try await TestSourceKitLSPClient()
+    let uri = DocumentURI(for: .swift)
+
+    let positions = testClient.openDocument(
+      """
+      func foo() {
+      if let SomeReallyLongVar = Some.More.Stuff(), let a = myfunc() {
+      1️⃣// do stuff2️⃣
+      }
+      }
+      """,
+      uri: uri
+    )
+
+    let response = try await testClient.send(
+      DocumentRangeFormattingRequest(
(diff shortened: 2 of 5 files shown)


Tests/SourceKitLSPTests/RangeFormattingTests.swift
```diff
@@ -0,0 +1,52 @@
...
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKLogging
+import SKTestSupport
+import SourceKitLSP
+import XCTest
+
+final class RangeFormattingTests: XCTestCase {
+  func testSelectComment() async throws {
```
MahdiBM, line 20: One test for now, will copy more of swift-format's own range-formatting tests over here.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
```diff
@@ -0,0 +1,52 @@
...
+      )
+    )
+
+    let edits = try XCTUnwrap(response)
+    XCTAssertEqual(
+      edits,
+      [
+        TextEdit(range: positions["1️⃣"]..<positions["2️⃣"], newText: "    // do stuff")
+      ]
+    )
```
MahdiBM, line 50: @ahoppen When I trigger this test, `edits` turns out to be empty no matter where I put the markers.
Any reasons? maybe I missed something obvious?
I don't notice any errors in the logs.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -136,20 +136,44 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+  }
+
+  package func documentRangeFormatting(_ req: DocumentRangeFormattingRequest) async throws -> [TextEdit]? {
+    return try await format(
+      textDocument: req.textDocument,
+      options: req.options,
+      extraArgs: [
+        "--offsets",
+        "\(req.range.lowerBound.line):\(req.range.upperBound.line)",
+      ]
```
ahoppen, line 152: I would prefer to keep the assembly of the swift-format command line invocation at one position and thus pass the range to format to `format` instead of an extra command line option.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -136,20 +136,44 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+    )
+  }
+
+  package func documentRangeFormatting(_ req: DocumentRangeFormattingRequest) async throws -> [TextEdit]? {
+    return try await format(
+      textDocument: req.textDocument,
+      options: req.options,
+      extraArgs: [
+        "--offsets",
+        "\(req.range.lowerBound.line):\(req.range.upperBound.line)",
```
ahoppen, line 151: swift-format expects UTF-8 byte offsets instead of line numbers for `--offsets`. You should be able to get the UTF-8 offset range using

https://github.com/swiftlang/sourcekit-lsp/blob/e73e8b798466aa8161ceb687ad2be092face8176/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift#L1174

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -136,20 +136,44 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
-      args: swiftFormat.pathString,
+    var args = try [
+      swiftFormat.pathString,
       "format",
       "--configuration",
-      try swiftFormatConfiguration(for: req.textDocument.uri, options: req.options)
-    )
+      swiftFormatConfiguration(for: textDocument.uri, options: options),
+    ]
+    args.append(contentsOf: extraArgs)
```
ahoppen, line 175: I generally prefer `+=` over `append(contentsOf:)`.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
ahoppen: I think this is because of the line vs. UTF-8 offset issue I mentioned above.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    let response = try await testClient.send(
+      DocumentRangeFormattingRequest(
+        textDocument: TextDocumentIdentifier(uri),
+        range: positions["1️⃣"]..<positions["2️⃣"],
+        options: FormattingOptions(tabSize: 2, insertSpaces: true)
+      )
+    )
+
+    let edits = try XCTUnwrap(response)
```
ahoppen, line 44: I don’t think you need to unwrap here. You should be able to compare the optional `response` to the array literal straight away.

ahoppen (commented):
Thanks for the PR, @MahdiBM. It’s looking great. I have left a few comments inline and I think they should also fix your issue that you don’t get any edits.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
MahdiBM: That's how it's done in `FormattingTests` so I think I should keep it that way for consistency?

Tests/SourceKitLSPTests/RangeFormattingTests.swift
MahdiBM: Thank you.

MahdiBM:
Ok I think the PR is looking good ... I'll mark as ready for review after I've added more tests.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
ahoppen: I don’t think we need to re-test swift-format’s range formatting in SourceKit-LSP. I would add another test that formats multiple lines and maybe one where only part of the line is selected. That should cover all possible position conversion issues that might occur in SourceKit-LSP’s code base.

Meant to post this yesterday but apparently forgot to hit send.

Tests/SourceKitLSPTests/RangeFormattingTests.swift
ahoppen: OK, fair enough. Let’s keep the `XCTUnwrap` then.

ahoppen (approved):
Thanks a lot for implementing this @MahdiBM. Changes look great to me.

ahoppen:
@swift-ci Please test

Sources/LanguageServerProtocol/Messages.swift
```diff
@@ -36,6 +36,7 @@ public let builtinRequests: [_RequestType.Type] = [
   DocumentColorRequest.self,
   DocumentDiagnosticsRequest.self,
   DocumentFormattingRequest.self,
+  DocumentRangeFormattingRequest.self,
```
ahoppen, line 39: `DocumentRangeFormattingRequest` is already in this array and that’s causing crashes during test execution and thus CI to fail

MahdiBM: 👌

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1807, opened 2024-11-06
Verify that all imports are declared as dependencies

ahoppen:
Pass `--explicit-target-dependency-import-check error` when building SourceKit-LSP in CI.


ahoppen:
@swift-ci Please test

hamishknight (approved):
TIL about this flag

PR #1815, opened 2024-11-10
Handle on-type formatting requests

### LINKED ISSUE #1805: Handle "textDocument/rangeFormatting" and "textDocument/onTypeFormatting"

MahdiBM:
To fully resolve #1805

### TODO
* [x] Hide behind an experimental flag
* [x] Make sure the experimental flag is enable-able in VSCode?

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -720,6 +720,8 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await self.handleRequest(for: request, requestHandler: self.documentFormatting)
     case let request as RequestAndReply<DocumentRangeFormattingRequest>:
       await self.handleRequest(for: request, requestHandler: self.documentRangeFormatting)
+    case let request as RequestAndReply<DocumentOnTypeFormattingRequest>:
+      await self.handleRequest(for: request, requestHandler: self.documentOnTypeFormatting)
     case let request as RequestAndReply<DocumentHighlightRequest>:
       await self.handleRequest(for: request, requestHandler: self.documentSymbolHighlight)
     case let request as RequestAndReply<DocumentSemanticTokensDeltaRequest>:
@@ -971,7 +973,8 @@ extension SourceKitLSPServer {
     let result = InitializeResult(
       capabilities: await self.serverCapabilities(
         for: req.capabilities,
-        registry: self.capabilityRegistry!
+        registry: self.capabilityRegistry!,
+        options: options
       )
     )
     logger.logFullObjectInMultipleLogMessages(header: "Initialize response", AnyRequestType(request: req))
@@ -980,7 +983,8 @@ extension SourceKitLSPServer {
 
   func serverCapabilities(
     for client: ClientCapabilities,
-    registry: CapabilityRegistry
+    registry: CapabilityRegistry,
+    options: SourceKitLSPOptions
   ) async -> ServerCapabilities {
     let completionOptions =
       await registry.clientHasDynamicCompletionRegistration
@@ -989,6 +993,11 @@ extension SourceKitLSPServer {
         resolveProvider: false,
         triggerCharacters: [".", "("]
       )
+
+    let onTypeFormattingOptions =
+      options.hasExperimentalFeature(.onTypeFormatting)
+      ? DocumentOnTypeFormattingOptions(triggerCharacters: ["\n", "\r\n", "\r", "{", "}", ";", ".", ":", "#"])
+      : nil
 
     let foldingRangeOptions =
       await registry.clientHasDynamicFoldingRangeRegistration
@@ -1039,6 +1048,7 @@ extension SourceKitLSPServer {
       codeLensProvider: CodeLensOptions(),
       documentFormattingProvider: .value(DocumentFormattingOptions(workDoneProgress: false)),
       documentRangeFormattingProvider: .value(DocumentRangeFormattingOptions(workDoneProgress: false)),
+      documentOnTypeFormattingProvider: onTypeFormattingOptions,
       renameProvider: .value(RenameOptions(prepareProvider: true)),
       colorProvider: .bool(true),
       foldingRangeProvider: foldingRangeOptions,
@@ -1535,6 +1545,14 @@ extension SourceKitLSPServer {
     languageService: LanguageService
   ) async throws -> [TextEdit]? {
     return try await languageService.documentRangeFormatting(req)
+  }
+
+  func documentOnTypeFormatting(
+    _ req: DocumentOnTypeFormattingRequest,
+    workspace: Workspace,
+    languageService: LanguageService
+  ) async throws -> [TextEdit]? {
+    return try await languageService.documentOnTypeFormatting(req)
   }
 
   func colorPresentation(
--- a/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
+++ b/Sources/SourceKitLSP/Swift/DocumentFormatting.swift
@@ -15,6 +15,7 @@ import Foundation
 package import LanguageServerProtocol
 import LanguageServerProtocolExtensions
 import SKLogging
+import SKUtilities
 import SwiftExtensions
 import SwiftParser
 import SwiftSyntax
@@ -28,6 +29,7 @@ import Foundation
 import LanguageServerProtocol
 import LanguageServerProtocolExtensions
 import SKLogging
+import SKUtilities
 import SwiftExtensions
 import SwiftParser
 import SwiftSyntax
@@ -143,6 +145,7 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
 extension SwiftLanguageService {
   package func documentFormatting(_ req: DocumentFormattingRequest) async throws -> [TextEdit]? {
     return try await format(
+      snapshot: documentManager.latestSnapshot(req.textDocument.uri),
       textDocument: req.textDocument,
       options: req.options
     )
@@ -150,19 +153,36 @@ extension SwiftLanguageService {
 
   package func documentRangeFormatting(_ req: DocumentRangeFormattingRequest) async throws -> [TextEdit]? {
     return try await format(
+      snapshot: documentManager.latestSnapshot(req.textDocument.uri),
       textDocument: req.textDocument,
       options: req.options,
       range: req.range
     )
   }
 
+  package func documentOnTypeFormatting(_ req: DocumentOnTypeFormattingRequest) async throws -> [TextEdit]? {
+    let snapshot = try documentManager.latestSnapshot(req.textDocument.uri)
+    guard let line = snapshot.lineTable.line(at: req.position.line) else {
+      return nil
+    }
+
+    let lineStartPosition = snapshot.position(of: line.startIndex, fromLine: req.position.line)
+    let lineEndPosition = snapshot.position(of: line.endIndex, fromLine: req.position.line)
+
+    return try await format(
+      snapshot: snapshot,
+      textDocument: req.textDocument,
+      options: req.options,
+      range: lineStartPosition..<lineEndPosition
+    )
+  }
+
   private func format(
+    snapshot: DocumentSnapshot,
     textDocument: TextDocumentIdentifier,
     options: FormattingOptions,
     range: Range<Position>? = nil
   ) async throws -> [TextEdit]? {
-    let snapshot = try documentManager.latestSnapshot(textDocument.uri)
-
     guard let swiftFormat else {
       throw ResponseError.unknown(
         "Formatting not supported because the toolchain is missing the swift-format executable"
@@ -176,9 +196,13 @@ extension SwiftLanguageService {
       swiftFormatConfiguration(for: textDocument.uri, options: options),
     ]
     if let range {
+      let utf8Range = snapshot.utf8OffsetRange(of: range)
+      // swift-format takes an inclusive range, but Swift's `Range.upperBound` is exclusive.
+      // Also make sure `upperBound` does not go less than `lowerBound`.
+      let utf8UpperBound = max(utf8Range.lowerBound, utf8Range.upperBound - 1)
       args += [
         "--offsets",
-        "\(snapshot.utf8Offset(of: range.lowerBound)):\(snapshot.utf8Offset(of: range.upperBound))",
+        "\(utf8Range.lowerBound):\(utf8UpperBound)",
       ]
     }
     let process = TSCBasic.Process(arguments: args)
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -1226,6 +1226,12 @@ extension DocumentSnapshot {
     return Position(line: zeroBasedLine, utf16index: utf16Column)
   }
 
+  /// Converts the given `String.Index` to a UTF-16-based line:column position.
+  func position(of index: String.Index, fromLine: Int = 0) -> Position {
+    let (line, utf16Column) = lineTable.lineAndUTF16ColumnOf(index, fromLine: fromLine)
+    return Position(line: line, utf16index: utf16Column)
+  }
+
   // MARK: Position <-> AbsolutePosition
 
   /// Converts the given UTF-8-offset-based `AbsolutePosition` to a UTF-16-based line:column.
--- a/Tests/SourceKitLSPTests/OnTypeFormattingTests.swift
+++ b/Tests/SourceKitLSPTests/OnTypeFormattingTests.swift
@@ -1,0 +1,127 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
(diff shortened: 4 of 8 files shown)


Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -170,9 +202,10 @@ extension SwiftLanguageService {
       swiftFormatConfiguration(for: textDocument.uri, options: options),
     ]
     if let range {
+      let utf8Range = snapshot.utf8OffsetRange(of: range)
       args += [
         "--offsets",
-        "\(snapshot.utf8Offset(of: range.lowerBound)):\(snapshot.utf8Offset(of: range.upperBound))",
+        "\(utf8Range.lowerBound):\(utf8Range.upperBound - 1)",
```
MahdiBM, line 208: swift-format takes an inclusive range, but Swift's `Range.upperBound` is exclusive.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -137,26 +137,58 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+    let capabilities = await server.serverCapabilities(
+      for: capabilityRegistry.clientCapabilities,
+      registry: capabilityRegistry
+    )
+    guard let documentOnTypeFormattingProvider = capabilities.documentOnTypeFormattingProvider,
+      documentOnTypeFormattingProvider.firstTriggerCharacter == req.ch
+        || documentOnTypeFormattingProvider.moreTriggerCharacter?.contains(req.ch) == true,
+      let line = snapshot.lineTable.line(at: req.position.line),
+      /// No need to go through whitespace checking if the trigger is not a newline
+      !req.ch.isNewline || !line.allSatisfy(\.isWhitespace)
```
MahdiBM, line 170: Don't trigger formatting if we're in an empty line.
Otherwise will mess up writing code. You'd write `{}` and try to go into the braces to write more code, only for on-type formatting to immediately close the braces again.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1041,6 +1043,9 @@ extension SourceKitLSPServer {
       codeLensProvider: CodeLensOptions(),
       documentFormattingProvider: .value(DocumentFormattingOptions(workDoneProgress: false)),
       documentRangeFormattingProvider: .value(DocumentRangeFormattingOptions(workDoneProgress: false)),
+      documentOnTypeFormattingProvider: DocumentOnTypeFormattingOptions(triggerCharacters: [
+        "\n", "\r\n", "\r", "{", "}", ";", ".", ":", "#",
+      ]),
```
MahdiBM, line 1048: Character list inspired by Xcode.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: I reckon this will go wrong if `lowerBound == upperBound`? do i just manually handle that with some kind of if statement, or is there a better way?

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Where did you get the list from in Xcode? Or did you just try different characters?

- `#` seems like an odd choice here. 
- `\r\n` isn’t a character, so not sure if that should be in the list.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -139,26 +139,58 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+    }
+    let snapshot = try documentManager.latestSnapshot(req.textDocument.uri)
+    let capabilities = await server.serverCapabilities(
+      for: capabilityRegistry.clientCapabilities,
+      registry: capabilityRegistry
+    )
+    
+    guard let documentOnTypeFormattingProvider = capabilities.documentOnTypeFormattingProvider,
+      documentOnTypeFormattingProvider.firstTriggerCharacter == req.ch
+        || documentOnTypeFormattingProvider.moreTriggerCharacter?.contains(req.ch) == true,
```
ahoppen, line 169: I generally find `?? false` easier to read because it directly answers the question: What happens if this is `nil`?

```suggestion
        || (documentOnTypeFormattingProvider.moreTriggerCharacter?.contains(req.ch) ?? false),
```

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
ahoppen: That’s a great explanation. Could you add it in the code?

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -139,26 +139,58 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+      let line = snapshot.lineTable.line(at: req.position.line),
+      /// No need to go through whitespace checking if the trigger is not a newline
```
ahoppen, line 171: This isn’t a doc comment, so it should  have only two slashes.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -139,26 +139,58 @@ private func edits(from original: DocumentSnapshot, to edited: String) -> [TextE
...
+      !req.ch.isNewline || !line.allSatisfy(\.isWhitespace)
+    else {
+      return nil
+    }
+
+    let lineStart = Position(line: req.position.line, utf16index: 0)
+    let nextLineStart = Position(line: req.position.line + 1, utf16index: 0)
```
ahoppen, line 178: What if this is the last line and there’s no trailing newline? Then we should use the end index of `req.position.line` to not go out of bounds.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
ahoppen: Nice catch. Can you add the explanation that swift-format takes an inclusive range to the code?

I would add a special case for `lowerBound == upperBound`.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -215,3 +248,9 @@ extension SwiftLanguageService {
     return edits(from: snapshot, to: formattedString)
   }
 }
+
+private extension String {
+  var isNewline: Bool {
+    self == "\n" || self == "\r\n" || self == "\r"
+  }
```
ahoppen, line 255: This property should already exist on `Character`. I would prefer to re-use that if possible instead of re-defining it.

Tests/SourceKitLSPTests/OnTypeFormattingTests.swift
```diff
@@ -0,0 +1,118 @@
...
+
+  func testFormatsFullLineAndDoesNotFormatNextLine() async throws {
+    try await SkipUnless.toolchainContainsSwiftFormat()
+    let testClient = try await TestSourceKitLSPClient()
+    let uri = DocumentURI(for: .swift)
+
+    let positions = testClient.openDocument(
+      """
+      func foo() {
+      1️⃣if let SomeReallyLongVar = 2️⃣    3️⃣Some.More.Stuff(), let a = 4️⃣    5️⃣myfunc() 6️⃣{
```
ahoppen, line 62: Did you check whether the location that VS Code sends is the position before or after the inserted character?

ahoppen (commented):
Thanks for working on the implementation. I’m still a little skeptical how well this will work in practice, but I’m happy to iterate on it behind an experimental flag. If we do get it working, that would be great.

> Make sure the experimental flag is enable-able in VSCode?
 
You should be able to enable the experimental feature in your [Configuration File](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Configuration%20File.md). There’s an option for that.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: I personally find `== true` more understandable, but I've seen some other people like `?? false` instead.

Sources/SourceKitLSP/SourceKitLSPServer.swift
MahdiBM: Screenshot 2024-11-12 at 3 33 27 AM

> `\r\n` isn’t a character

good catch, i just copy pasted right from Xcode.

MahdiBM: > # seems like an odd choice here.
Hmm right i did think that as well ... I _could_ remove that.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: I'm used to use doc comments everywhere since it comes with some markdown capabilities.
But I understand you're following the more "standardized" way of writing comments in this project.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: It appears this is not needed.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: It appears this is no needed considering swift-format only will strictly format that single line the formatting is triggered on. 

I've removed the related code but kept the test because this problem can render on-type formatting useless if it happens in the future, as users will need to disable on-type formatting to be able to code at all.

Sources/SourceKitLSP/SourceKitLSPServer.swift
MahdiBM: Actually `\r\n` _is_ one `Character`, but 2 unicode scalars.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: Does it matter considering the range is exclusive so the next line is never actually operated on?

Hmm i see how this could wrong though, I'll need to adjust the code.

If the next line doesn't exist, `snapshot` functions will just return the `endIndex` of the content.
That means `upperBound` could be the last `utf8Offset` in the same line. And we're doing `upperBound - 1` because Swift's `Range.upperBound` is exclusive, but in this case we'll be ignoring the last character of the last line which will be a bug.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: I originally did that because i suspected it'll be more performant.
There are some code in this project that also avoid calling into `CharacterSet`-related functions.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
ahoppen: The most important thing that I thought of was that the conversion functions would log a fault despite there not being an actual issue.

ahoppen (commented):
Changes look good to me once they are guarded behind an experimental feature.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
MahdiBM: I changed the code to find the positions more-strictly:
https://github.com/swiftlang/sourcekit-lsp/pull/1815/commits/51fab20c8c23adf18e54e219f3b052166f2ac3ab

Tests/SourceKitLSPTests/OnTypeFormattingTests.swift
MahdiBM: It appears the position [is not guaranteed](https://arc.net/l/quote/llyptxzs) to be exact. Doesn't matter for us either, we format the whole line in any case.

> /**
	 * The position around which the on type formatting should happen.
	 * This is not necessarily the exact position where the character denoted
	 * by the property `ch` got typed.
	 */
>	position: Position;

MahdiBM:
I'll handle the experimental-feature gating later 🙂.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -1228,6 +1228,16 @@ extension DocumentSnapshot {
     return Position(line: zeroBasedLine, utf16index: utf16Column)
   }
 
+  /// Converts the given `String.Index` to a UTF-16-based line:column position.
+  ///
+  /// If the index does not refer to a valid position within the snapshot, returns the closest valid
+  /// position and logs a fault containing the file and line of the caller (from `callerFile` and
+  /// `callerLine`).
+  func position(of index: String.Index, fromLine: Int = 0) -> Position {
```
ahoppen, line 1236: I don’t think the comment is correct because there’s no `callerFile` and `callerLine` comment. Also, no fault is logged by `LineTable.lineAndUTF16ColumnOf`.

ahoppen (commented):
Thank you. Just one comment on a comment 😉

ahoppen (approved):
Looks good. Thank you @MahdiBM

ahoppen:
@swift-ci Please test

MahdiBM:
Fixed this:
```
159 |   package func documentOnTypeFormatting(_ req: DocumentOnTypeFormattingRequest) async throws -> [TextEdit]? {
160 |     let snapshot = try documentManager.latestSnapshot(req.textDocument.uri)
161 |     guard let line = snapshot.lineTable.line(at: req.position.line) else {
    |                                         `- error: instance method 'line(at:callerFile:callerLine:)' is not available due to missing import of defining module 'SKUtilities'
162 |       return nil
163 |     }
```

ahoppen:
Thanks for the fix.

@swift-ci Please test

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -27,6 +27,7 @@ import Foundation
 import LanguageServerProtocol
 import LanguageServerProtocolExtensions
 import SKLogging
+import SKUtilities
```
ahoppen, line 30: You need to add the import to the `#if compiler(>=6)` branch as well.

Sources/SourceKitLSP/Swift/DocumentFormatting.swift
```diff
@@ -15,6 +15,7 @@ import Foundation
 package import LanguageServerProtocol
 import LanguageServerProtocolExtensions
 import SKLogging
+import SKUtilities
 import SwiftExtensions
 import SwiftParser
 import SwiftSyntax
```
MahdiBM, line 21: @ahoppen I noticed there has been 2 import changes to this file that i had to merge from main.
Just checking that this looks good?
First was addition of `LanguageServerProtocolExtensions` IIRC, the newer one was addition of `SwiftExtensions`.
I also added `SKUtilities`.

MahdiBM:
@ahoppen done

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

MahdiBM:
@ahoppen rebased with main again, this was the error which looks like unrelated to this PR?

```
sourcekit-lsp/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift:318:107: error: cannot convert value of type 'URL' to expected argument type 'TSCAbsolutePath' (aka 'AbsolutePath')
316 |       )
317 |     } else if let scratchDirectory = options.swiftPMOrDefault.scratchPath,
318 |       let scratchDirectoryPath = try? AbsolutePath(validating: scratchDirectory, relativeTo: AbsolutePath(projectRoot))
    |                                                                                                           `- error: cannot convert value of type 'URL' to expected argument type 'TSCAbsolutePath' (aka 'AbsolutePath')
319 |     {
320 |       location.scratchDirectory = scratchDirectoryPath

/home/build-user/sourcekit-lsp/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift:479:14: warning: When we drop support for Swift 5.10 we no longer need to adjust compiler arguments for the Modules move
477 | 
478 |     #if compiler(>=6.1)
479 |     #warning("When we drop support for Swift 5.10 we no longer need to adjust compiler arguments for the Modules move")
    |              `- warning: When we drop support for Swift 5.10 we no longer need to adjust compiler arguments for the Modules move
480 |     #endif
481 |     // Fix up compiler arguments that point to a `/Modules` subdirectory if the Swift version in the toolchain is less
```

ahoppen:
Ah, that was because of a merge conflict that I resolved in https://github.com/swiftlang/sourcekit-lsp/pull/1838

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1819, opened 2024-11-14
Allow waiting for the next notification from SourceKit-LSP when the previous wait timed out

ahoppen:
When `TestSourceKitLSPClient.nextNotification` timed out, it would cancel `AsyncSequence.Iterator.next()` and thus also cancel the underlying `AsyncStream` (which I only now found out). This means that calling `nextNotification` again would never return any notification. Hence, if we didn’t receive a `PublishDiagnosticsNotification` within the first timeout, we would never get one.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1820, opened 2024-11-14
Fix issue where test nondeterministically failed if SwiftPM doesn't return build settings quickly

ahoppen:
We might not have build settings at the time when we get the first diagnostics notification here and thus might get syntactic fallback diagnostics that don't contain 'Cannot find ... in scope'


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1821, opened 2024-11-14
Enable SwiftPM-baesd tests on Windows

ahoppen:
I ran Windows test 200 times locally without any failures, so I assume all the nondeterministic failures should be shaken out now.


bnbarham (approved):
🥳

PR #1824, opened 2024-11-15
Allow `scratchPath` to be a relative path

kateinoigakukun:
Interpret it as relative to the project root directory if it's a relative path.

### DIFF
--- a/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
+++ b/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
@@ -324,7 +324,7 @@ package actor SwiftPMBuildSystem: BuiltInBuildSystem {
     if options.backgroundIndexingOrDefault {
       location.scratchDirectory = AbsolutePath(projectRoot.appending(components: ".build", "index-build"))
     } else if let scratchDirectory = options.swiftPMOrDefault.scratchPath,
-      let scratchDirectoryPath = try? AbsolutePath(validating: scratchDirectory)
+      let scratchDirectoryPath = try? AbsolutePath(validating: scratchDirectory, relativeTo: AbsolutePath(projectRoot))
     {
       location.scratchDirectory = scratchDirectoryPath
     }
--- a/Tests/BuildSystemIntegrationTests/SwiftPMBuildSystemTests.swift
+++ b/Tests/BuildSystemIntegrationTests/SwiftPMBuildSystemTests.swift
@@ -90,6 +90,43 @@ final class SwiftPMBuildSystemTests: XCTestCase {
     }
   }
 
+  func testRelativeScratchPath() async throws {
+    try await withTestScratchDir { tempDir in
+      try localFileSystem.createFiles(
+        root: tempDir,
+        files: [
+          "pkg/Sources/lib/a.swift": "",
+          "pkg/Package.swift": """
+          // swift-tools-version:4.2
+          import PackageDescription
+          let package = Package(
+            name: "a",
+            targets: [.target(name: "lib")]
+          )
+          """,
+        ]
+      )
+      let packageRoot = tempDir.appending(component: "pkg")
+      let options = SourceKitLSPOptions(
+        swiftPM: .init(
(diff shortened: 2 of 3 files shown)


Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
```diff
@@ -303,7 +303,9 @@ package actor SwiftPMBuildSystem: BuiltInBuildSystem {
       swiftSDKSelector: options.swiftPMOrDefault.swiftSDK,
       store: SwiftSDKBundleStore(
         swiftSDKsDirectory: localFileSystem.getSharedSwiftSDKsDirectory(
-          explicitDirectory: options.swiftPMOrDefault.swiftSDKsDirectory.map { try AbsolutePath(validating: $0) }
+          explicitDirectory: options.swiftPMOrDefault.swiftSDKsDirectory.map {
+            try AbsolutePath(validating: $0, relativeTo: AbsolutePath(projectRoot))
+          }
```
ahoppen, line 308: Does it make sense to have a SwiftSDK installed relative to your project? Don’t you typically install Swift SDKs globally?

Tests/BuildSystemIntegrationTests/SwiftPMBuildSystemTests.swift
```diff
@@ -88,6 +88,42 @@ final class SwiftPMBuildSystemTests: XCTestCase {
...
+      let swiftpmBuildSystem = try await SwiftPMBuildSystem(
+        projectRoot: packageRoot,
+        toolchainRegistry: .forTesting,
+        options: options,
+        connectionToSourceKitLSP: LocalConnection(receiverName: "dummy"),
+        testHooks: SwiftPMTestHooks()
+      )
+
+      let dataPath = await swiftpmBuildSystem.destinationBuildParameters.dataPath
+      let expectedScratchPath = packageRoot.appending(component: options.swiftPMOrDefault.scratchPath!)
```
ahoppen, line 122:
```suggestion
      let expectedScratchPath = packageRoot.appending(component: try XCTUnwrap(options.swiftPMOrDefault.scratchPath))
```

ahoppen (commented):
Should `swiftSDKsDirectory` also be interpreted relative to the project root? And could you add that the paths are relative to the project root to [Configuration File.md](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Configuration%20File.md)?

Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
kateinoigakukun: Yeah, it would be typically an absolute path. But I think it's not harmful to interpret it as relative to project root only when it's a relative path. Note that [`AbsolutePath(validating: relativeTo:)`](https://github.com/swiftlang/swift-tools-support-core/blob/4074f4db0971328c441fc1621c673937b9ca3b08/Sources/TSCBasic/Path.swift#L62-L64) uses the first argument as is if it's absolute.

ahoppen: The thing I’m wondering is whether interpreting the path relative to the project root is the right way to go. You could have a global SourceKit-LSP configuration file at `~/.sourcekit-lsp/config.json` and it wouldn’t be unreasonable to expect that the Swift SDK path is interpreted relative to that configuration file. If we don’t allow relative paths, we don’t need to worry about any ambiguity here. 

I think for the scratch directory it’s natural to assume that it’s relative to the project root because having a global scratch directory for all project makes very little sense, so making that path relative to the project root makes perfect sense, I think.

kateinoigakukun: Ahh, that makes sense. Thank you for your explanation, I reverted swiftSDKsDirectory change.

ahoppen:
@swift-ci Please test

kateinoigakukun:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1827, opened 2024-11-16
Restrict directories that we search for `MyDependency.swift` in tests

ahoppen:
It appears that moving the index-build directory to be a subdirectory of `.build` caused some file in the index-build directory to exceed the `MAX_PATH` length on the SourceKit-LSP PR testing job (but not the swift PR job because that has a shorter job name). Because of https://github.com/swiftlang/swift-foundation/issues/1049, we would stop directory iteration at that file exceeding `MAX_PATH` and never find `MyDependency.swift`, causing these tests to fail. Only searching the `checkout` directory works around this issue because `FileManager.enumerator` never sees the file with the exceeding length that way.

### DIFF
--- a/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
+++ b/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
@@ -245,7 +245,11 @@ final class BackgroundIndexingTests: XCTestCase {
     )
 
     let dependencyUrl = try XCTUnwrap(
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1828, opened 2024-11-16
Add VS Code to `Editor Integration.md`

MahdiBM:
Noticed this is missing when going through the docs.


ahoppen (commented):
This document says at the start 

> https://www.swift.org/tools has a list of some popular editors and how to set them up. This page covers any editors not listed there.

And VS Code is include in that list of editors listed on swift.org/tools.

MahdiBM:
O well you're right. If It's specifically mentioned that this page is only for editors not mentioned in Swift.org, then there is no reason to add VS Code.

PR #1831, opened 2024-11-17
Handle new swift-syntax closure expansion behavior

woolsweater:
This resolves <https://github.com/swiftlang/sourcekit-lsp/issues/1788>, following the discussion of alternatives on <https://github.com/swiftlang/sourcekit-lsp/pulls/1789>. The bulk of the change updates the translation from SourceKit placeholders to LSP placeholders to handle nesting, which is implemented in https://github.com/swiftlang/swift-syntax/pull/2897

https://github.com/user-attachments/assets/0d5cca51-95f9-44e9-8a48-92a296e40678

### DIFF
--- a/Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
+++ b/Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
@@ -1,0 +1,73 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.0)
+public import SwiftBasicFormat
+public import SwiftSyntax
+#else
+import SwiftBasicFormat
+import SwiftSyntax
+#endif
+
+/// A specialization of `BasicFormat` for closure literals in a code completion
+/// context.
+///
+/// This is more conservative about newline insertion: unless the closure has
+/// multiple statements in its body it will not be reformatted to multiple
+/// lines.
+@_spi(Testing)
+public class ClosureCompletionFormat: BasicFormat {
+  @_spi(Testing)
+  public override func requiresNewline(
+    between first: TokenSyntax?,
+    and second: TokenSyntax?
+  ) -> Bool {
+    if let first, isEndOfSmallClosureSignature(first) {
+      return false
+    } else if let first, isSmallClosureDelimiter(first, kind: \.leftBrace) {
+      return false
+    } else if let second, isSmallClosureDelimiter(second, kind: \.rightBrace) {
+      return false
+    } else {
+      return super.requiresNewline(between: first, and: second)
+    }
+  }
+
+  /// Returns `true` if `token` is an opening or closing brace (according to
+  /// `kind`) of a closure, and that closure has no more than one statement in
+  /// its body.
+  private func isSmallClosureDelimiter(
+    _ token: TokenSyntax,
+    kind: KeyPath<ClosureExprSyntax, TokenSyntax>
+  ) -> Bool {
+    guard token.keyPathInParent == kind,
+      let closure = token.parent?.as(ClosureExprSyntax.self)
+    else {
+      return false
+    }
+
+    return closure.statements.count <= 1
+  }
+
+  /// Returns `true` if `token` is the last token in the signature of a closure,
+  /// and that closure has no more than one statement in its body.
+  private func isEndOfSmallClosureSignature(_ token: TokenSyntax) -> Bool {
+    guard
+      token.keyPathInParent == \ClosureSignatureSyntax.inKeyword,
+      let closure = token.ancestorOrSelf(mapping: { $0.as(ClosureExprSyntax.self) })
+    else {
+      return false
+    }
+
+    return closure.statements.count <= 1
+  }
+}
--- a/Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
+++ b/Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
@@ -14,34 +14,166 @@ import Foundation
 import SKLogging
 @_spi(RawSyntax) import SwiftSyntax
 
-func rewriteSourceKitPlaceholders(in string: String, clientSupportsSnippets: Bool) -> String {
-  var result = string
-  var index = 1
-  while let start = result.range(of: "<#") {
-    guard let end = result[start.upperBound...].range(of: "#>") else {
-      logger.fault("Invalid placeholder in \(string)")
-      return string
+/// Translate SourceKit placeholder syntax — `<#foo#>` — in `input` to LSP
+/// placeholder syntax: `${n:foo}`.
+///
+/// If `clientSupportsSnippets` is `false`, the placeholder is rendered as an
+/// empty string, to prevent the client from inserting special placeholder
+/// characters as if they were literal text.
+@_spi(Testing)
+public func rewriteSourceKitPlaceholders(in input: String, clientSupportsSnippets: Bool) -> String {
+  var result = ""
+  var nextPlaceholderNumber = 1
+  // Current stack of nested placeholders, most nested last. Each element needs
+  // to be rendered inside the element before it.
+  var placeholders: [(number: Int, contents: String)] = []
+  let tokens = tokenize(input)
+  for token in tokens {
+    switch token {
+    case let .text(text):
+      if placeholders.isEmpty {
+        result += text
+      } else {
+        placeholders.latest.contents += text
+      }
+
+    case let .curlyBrace(brace):
+      if placeholders.isEmpty {
+        result.append(brace)
+      } else {
+        // Braces are only escaped _inside_ a placeholder; otherwise the client
+        // would include the backslashes literally.
+        placeholders.latest.contents.append(contentsOf: ["\\", brace])
+      }
+
+    case .placeholderOpen:
+      placeholders.append((number: nextPlaceholderNumber, contents: ""))
+      nextPlaceholderNumber += 1
+
+    case .placeholderClose:
+      guard let (number, placeholderBody) = placeholders.popLast() else {
+        logger.fault("Invalid placeholder in \(input)")
+        return input
+      }
+      guard let displayName = nameForSnippet(placeholderBody) else {
+        logger.fault("Failed to decode placeholder \(placeholderBody) in \(input)")
+        return input
+      }
+      let placeholder =
+        clientSupportsSnippets
+        ? formatLSPPlaceholder(displayName, number: number)
+        : ""
+      if placeholders.isEmpty {
+        result += placeholder
+      } else {
+        placeholders.latest.contents += placeholder
+      }
     }
-    let rawPlaceholder = String(result[start.lowerBound..<end.upperBound])
-    guard let displayName = nameForSnippet(rawPlaceholder) else {
-      logger.fault("Failed to decode placeholder \(rawPlaceholder) in \(string)")
-      return string
-    }
-    let snippet = clientSupportsSnippets ? "${\(index):\(displayName)}" : ""
-    result.replaceSubrange(start.lowerBound..<end.upperBound, with: snippet)
-    index += 1
   }
+
   return result
 }
 
-/// Parse a SourceKit placeholder and extract the display name suitable for a
-/// LSP snippet.
-fileprivate func nameForSnippet(_ text: String) -> String? {
-  var text = text
+/// Scan `input` to identify special elements within: curly braces, which may
+/// need to be escaped; and SourceKit placeholder open/close delimiters.
+private func tokenize(_ input: String) -> [SnippetToken] {
+  var index = input.startIndex
+  var isAtEnd: Bool { index == input.endIndex }
+  func match(_ char: Character) -> Bool {
+    if isAtEnd || input[index] != char {
+      return false
+    } else {
+      input.formIndex(after: &index)
+      return true
+    }
+  }
+  func next() -> Character? {
+    guard !isAtEnd else { return nil }
+    defer { input.formIndex(after: &index) }
+    return input[index]
+  }
+
+  var tokens: [SnippetToken] = []
+  var text = ""
+  while let char = next() {
+    switch char {
+    case "<":
+      if match("#") {
+        tokens.append(.text(text))
+        text.removeAll()
+        tokens.append(.placeholderOpen)
+      } else {
+        text.append(char)
+      }
+
+    case "#":
+      if match(">") {
+        tokens.append(.text(text))
+        text.removeAll()
+        tokens.append(.placeholderClose)
+      } else {
+        text.append(char)
+      }
+
+    case "{", "}":
+      tokens.append(.text(text))
+      text.removeAll()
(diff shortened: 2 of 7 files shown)


Tests/SourceKitLSPTests/RewriteSourceKitPlaceholdersTests.swift
```diff
@@ -0,0 +1,54 @@
+import SKTestSupport
+@testable import SourceKitLSP
+import XCTest
+
+final class RewriteSourceKitPlaceholdersTests : XCTestCase {
```
woolsweater, line 5: ℹ️ These overlap somewhat with the SwiftCompletionTests but I thought including tests specifically directed at nesting would be worthwhile.

Tests/SourceKitLSPTests/RewriteSourceKitPlaceholdersTests.swift
```diff
@@ -0,0 +1,62 @@
+import SKTestSupport
```
ahoppen, line 1: Could you add the copyright header to this file?

Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
```diff
@@ -14,34 +14,163 @@ import Foundation
...
+  // to be rendered inside the element before it.
+  var placeholders: [(number: Int, contents: String)] = []
+  let tokens = tokenize(input)
+  for token in tokens {
+    switch token {
+    case let .text(text) where placeholders.isEmpty:
+      result += text
+
+    case let .text(text):
+      placeholders.latest.contents += text
```
ahoppen, line 36: Instead of having two cases that just different in a `where` clause, I would have one case with a nested `if`. I think that highlights the difference between whether we’re currently inside a placeholder better.

```suggestion
    case let .text(text):
      if placeholders.isEmpty {
        result += text
      } else {
        placeholders.latest.contents += text
      }
```

Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
```diff
@@ -14,34 +14,163 @@ import Foundation
...
+
+  tokens.append(.text(text))
+
+  return tokens
+}
+
+/// A syntactical element inside a SourceKit snippet.
+private enum SnippetToken {
+  /// A placeholder delimiter.
+  case open, close
```
ahoppen, line 133: I would name them `placeholderOpen` and `placeholderClose`. `open` and `close` sound pretty ambiguous.

Tests/SourceKitLSPTests/RewriteSourceKitPlaceholdersTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    let rewritten = rewriteSourceKitPlaceholders(in: input, clientSupportsSnippets: true)
+
+    XCTAssertEqual(rewritten, "foo(bar: ${1:Int}, baz: ${2:String}, quux: ${3:String})")
+  }
+
+  func testClosurePlaceholderReturnType() {
+    let input = "foo(bar: \(placeholder: "{ \(placeholder: "T##Int##Int") }"))"
+    let rewritten = rewriteSourceKitPlaceholders(in: input, clientSupportsSnippets: true)
+
+    XCTAssertEqual(rewritten, "foo(bar: ${1:\\{ ${2:Int} \\}})")
```
ahoppen, line 38: You can use raw string literals so you don’t need to write double backslashes here. I think that would improve readability.

```suggestion
    XCTAssertEqual(rewritten, #"foo(bar: ${1:\{ ${2:Int} \}})"#)
```

Tests/SourceKitLSPTests/RewriteSourceKitPlaceholdersTests.swift
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+private let placeholderStart = "<#"
+private let placeholderEnd = "#>"
+private extension DefaultStringInterpolation {
+  mutating func appendInterpolation(placeholder string: String) {
+    self.appendLiteral(placeholderStart + string + placeholderEnd)
+  }
+}
```
ahoppen, line 62: I don’t think I’m a fan of extending string interpolation in this way. I would prefer to spell out the placeholders in the tests as `<#…#>`. I know that it’s a little annoying when editing the file in Xcode but I think overall I’ve found that it’s easiest to reason about these sorts of tests if you can view the input text as it will returned from sourcekitd, even if that means opening the file in a non-Xcode editor at times.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -864,18 +864,14 @@ final class SwiftCompletionTests: XCTestCase {
           sortText: nil,
           filterText: "myMap(:)",
           insertText: """
-            myMap { ${1:Int} in
-                ${2:Bool}
-            }
+            myMap(${1:\\{ ${2:Int} in ${3:Bool} \\}})
```
ahoppen, line 867: Similar to my comment in the other test file, i would use raw string literals here so you don’t need to write `\\`.

ahoppen (commented):
Thank you @woolsweater, great changes!

I only have a few comments, otherwise looks great.

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
```diff
@@ -0,0 +1,63 @@
...
+      return super.requiresNewline(between: first, and: second)
+    }
+  }
+
+  /// Returns `true` if `token` is an opening or closing brace (according to
+  /// `kind`) of a closure, and that closure has no more than one statement in
+  /// its body.
+  private func isSmallClosureDelimiter(
+    _ token: TokenSyntax,
+    kind: KeyPath<ClosureExprSyntax, TokenSyntax>
```
woolsweater, line 40: Moving https://github.com/swiftlang/swift-syntax/pull/2897#discussion_r1847456663 here

> Do we need to pass in the key path here? It doesn’t really look very dynamic, so I’d just check for `\.leftBrace` and `\.rightBrace` in here.

It is defensive, but I think it's worthwhile. We don't want to override the `BasicFormat` for the token _before_ a `{` or _after_ a `}`. Passing the keypath guards against those.

Tests/SourceKitLSPTests/ClosureCompletionFormatTests.swift
```diff
@@ -0,0 +1,129 @@
...
+import XCTest
+
+fileprivate func assertFormatted<T: SyntaxProtocol>(
+  tree: T,
+  expected: String,
+  using format: ClosureCompletionFormat = ClosureCompletionFormat(indentationWidth: .spaces(4)),
+  file: StaticString = #filePath,
+  line: UInt = #line
+) {
+  XCTAssertEqual(tree.formatted(using: format).description, expected, file: file, line: line)
```
woolsweater, line 27: ℹ️ Unfortunately it doesn't seem like _SwiftSyntaxTestSupport can be added to the dependencies, so I had to use plain `XCTAssertEqual` instead of the nice `assertStringsEqualWithDiff`.

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
```diff
@@ -0,0 +1,63 @@
...
+import SwiftSyntax
+import SwiftBasicFormat
+
+/// A specialization of `BasicFormat` for closure literals in a code completion
+/// context.
+///
+/// This is more conservative about newline insertion: unless the closure has
+/// multiple statements in its body it will not be reformatted to multiple
+/// lines.
+class ClosureCompletionFormat: BasicFormat {
```
woolsweater, line 22: ℹ️ Moved into SourceKit from the prior revision of my swift-syntax PR, per https://github.com/swiftlang/swift-syntax/pull/2897#discussion_r1847462770

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
ahoppen: Sorry, looks like my comment was misleading. I meant that instead of taking a key path as a parameter to this function, we could hardcode it inside the body (ie. changing the body to `token.keyPathInParent == \.leftBrace || token.keyPathInParent == \.rightBrace `). But upon second thought, that would change semantics because it would also suppress the newline after the right brace. So, you’re implementation is good. Ignore my comment.

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
```diff
@@ -0,0 +1,63 @@
...
+  /// Returns `true` if `token` is the last token in the signature of a closure,
+  /// and that closure has no more than one statement in its body.
+  private func isEndOfSmallClosureSignature(_ token: TokenSyntax) -> Bool {
+    guard
+      let closure = token.ancestorOrSelf(mapping: { $0.as(ClosureExprSyntax.self) })
+    else {
+      return false
+    }
+
+    return token.keyPathInParent == \ClosureSignatureSyntax.inKeyword
```
ahoppen, line 60: I would move the check for `token.keyPathInParent == \ClosureSignatureSyntax.inKeyword` to the top of the function because it’s an O(1) check vs the `ancestorOrSelf` check, which needs to traverse the tree up to the root to determine that a token is not in a `ClosureExprSyntax`.

Tests/SourceKitLSPTests/ClosureCompletionFormatTests.swift
```diff
@@ -0,0 +1,129 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftBasicFormat
+import SwiftParser
+import SwiftSyntax
+import Swift
+@testable import SourceKitLSP
```
ahoppen, line 17: We build SourceKit-LSP with testability disabled in Swift CI because otherwise we would need to build everything twice: Once with testability enabled for testing and then with testability disabled to produce the binary we include in the toolchain. Could you change this to not use `@testable`? We use `@_spi(Testing)` to a similar effect where needed.

ahoppen (commented):
Thanks. I just have two minor comments, otherwise this looks great.

Tests/SourceKitLSPTests/ClosureCompletionFormatTests.swift
woolsweater: Got it, updated. 🤔 I am not very familiar with how `@_spi` works, but it seems to require everything including the imports to become public. Let me know if there is another way I should have done this.

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
woolsweater: Good point; done!

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
```diff
@@ -0,0 +1,64 @@
...
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public import SwiftSyntax
+public import SwiftBasicFormat
```
ahoppen, line 14: We need to support building with a Swift 5.10 compiler as well, which doesn’t support the access level imports.

```suggestion
#if compiler(>=6.0)
public import SwiftSyntax
public import SwiftBasicFormat
#else
import SwiftSyntax
import SwiftBasicFormat
#endif
```

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
```diff
@@ -0,0 +1,64 @@
...
+
+/// A specialization of `BasicFormat` for closure literals in a code completion
+/// context.
+///
+/// This is more conservative about newline insertion: unless the closure has
+/// multiple statements in its body it will not be reformatted to multiple
+/// lines.
+@_spi(Testing)
+public class ClosureCompletionFormat: BasicFormat {
+  public override func requiresNewline(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
```
ahoppen, line 24: I would add `@_spi(Testing)` here as well, just to be clear that it’s not a properly public declaration.

```suggestion
  @_spi(Testing) public override func requiresNewline(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
```

ahoppen (commented):
Thanks. A small comment that you couldn’t have been aware of regarding the imports.

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
woolsweater: Makes sense; done!

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
woolsweater: Done!

ahoppen (commented):
Looking good. Thanks for implementing this and following up on my review comments.

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test

Tests/SourceKitLSPTests/RewriteSourceKitPlaceholdersTests.swift
```diff
@@ -0,0 +1,66 @@
...
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SKTestSupport
+@testable import SourceKitLSP
```
woolsweater, line 14: Shoot, I missed this other `@testable` import last night. Does this need to be changed as well @ahoppen ?

ahoppen: Yes, we can’t have `@testable` imports.

woolsweater: Fixed!

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test Windows

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test

Sources/SourceKitLSP/Swift/ClosureCompletionFormat.swift
ahoppen, line 1: Could you add this file to [CMakeLists.txt](https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SourceKitLSP/CMakeLists.txt). We use CMake to build SourceKit-LSP on Windows.

woolsweater: Done!

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2897

@swift-ci Please test Windows

woolsweater:
Thank you for shepherding these changes, @ahoppen ! 🙌

PR #1832, opened 2024-11-18
Use `URL` in many cases where we used `AbsolutePath`

ahoppen:
We made quite a few fixes recently to make sure that path handling works correctly using `URL` on Windows. Use `URL` in most places to have a single type that represents file paths instead of sometimes using `AbsolutePath`.

While doing so, also remove usages of `TSCBasic.FileSystem` an `InMemoryFileSystem`. The pattern of using `InMemoryFileSystem` for tests was never consistently used and it was a little confusing that some types took a `FileSystem` parameter while other always assumed to work on the local file system.

### DIFF
--- a/Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
+++ b/Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
@@ -20,7 +20,6 @@ import SKOptions
 import SwiftExtensions
 import TSCExtensions
 
-import struct TSCBasic.AbsolutePath
 import func TSCBasic.getEnvSearchPaths
 import var TSCBasic.localFileSystem
 import func TSCBasic.lookupExecutablePath
@@ -34,7 +33,7 @@ private func executable(_ name: String) -> String {
   #endif
 }
 
-private let python3ExecutablePath: AbsolutePath? = {
+private let python3ExecutablePath: URL? = {
   let pathVariable: String
   #if os(Windows)
   pathVariable = "Path"
@@ -47,8 +46,8 @@ private let python3ExecutablePath: AbsolutePath? = {
       currentWorkingDirectory: localFileSystem.currentWorkingDirectory
     )
 
-  return lookupExecutablePath(filename: executable("python3"), searchPaths: searchPaths)
-    ?? lookupExecutablePath(filename: executable("python"), searchPaths: searchPaths)
+  return lookupExecutablePath(filename: executable("python3"), searchPaths: searchPaths)?.asURL
+    ?? lookupExecutablePath(filename: executable("python"), searchPaths: searchPaths)?.asURL
 }()
 
 struct ExecutableNotFoundError: Error {
@@ -75,16 +74,16 @@ private struct BuildServerConfig: Codable {
   /// Command arguments runnable via system processes to start a BSP server.
   let argv: [String]
 
-  static func load(from path: AbsolutePath) throws -> BuildServerConfig {
+  static func load(from path: URL) throws -> BuildServerConfig {
     let decoder = JSONDecoder()
-    let fileData = try localFileSystem.readFileContents(path).contents
-    return try decoder.decode(BuildServerConfig.self, from: Data(fileData))
+    let fileData = try Data(contentsOf: path)
+    return try decoder.decode(BuildServerConfig.self, from: fileData)
   }
 }
 
 /// Launches a subprocess that is a BSP server and manages the process's lifetime.
 actor ExternalBuildSystemAdapter {
-  private let projectRoot: AbsolutePath
+  private let projectRoot: URL
 
   /// The `BuildSystemManager` that handles messages from the BSP server to SourceKit-LSP.
   var messagesToSourceKitLSPHandler: MessageHandler
@@ -100,7 +99,7 @@ actor ExternalBuildSystemAdapter {
   /// Used to delay restarting in case of a crash loop.
   private var lastRestart: Date?
 
-  static package func projectRoot(for workspaceFolder: AbsolutePath, options: SourceKitLSPOptions) -> AbsolutePath? {
+  static package func projectRoot(for workspaceFolder: URL, options: SourceKitLSPOptions) -> URL? {
     guard getConfigPath(for: workspaceFolder) != nil else {
       return nil
     }
@@ -108,7 +107,7 @@ actor ExternalBuildSystemAdapter {
   }
 
   init(
-    projectRoot: AbsolutePath,
+    projectRoot: URL,
     messagesToSourceKitLSPHandler: MessageHandler
   ) async throws {
     self.projectRoot = projectRoot
@@ -154,11 +153,11 @@ actor ExternalBuildSystemAdapter {
     }
 
     let serverConfig = try BuildServerConfig.load(from: configPath)
-    var serverPath = try AbsolutePath(validating: serverConfig.argv[0], relativeTo: projectRoot)
+    var serverPath = URL(fileURLWithPath: serverConfig.argv[0], relativeTo: projectRoot.ensuringCorrectTrailingSlash)
     var serverArgs = Array(serverConfig.argv[1...])
 
-    if serverPath.suffix == ".py" {
-      serverArgs = [serverPath.pathString] + serverArgs
+    if serverPath.pathExtension == "py" {
+      serverArgs = [try serverPath.filePath] + serverArgs
       guard let interpreterPath = python3ExecutablePath else {
         throw ExecutableNotFoundError(executableName: "python3")
       }
@@ -167,7 +166,7 @@ actor ExternalBuildSystemAdapter {
     }
 
     return try JSONRPCConnection.start(
-      executable: serverPath.asURL,
+      executable: serverPath,
       arguments: serverArgs,
       name: "BSP-Server",
       protocol: bspRegistry,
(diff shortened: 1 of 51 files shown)


Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -154,11 +153,11 @@ actor ExternalBuildSystemAdapter {
...
 
     let serverConfig = try BuildServerConfig.load(from: configPath)
-    var serverPath = try AbsolutePath(validating: serverConfig.argv[0], relativeTo: projectRoot)
+    var serverPath = URL(fileURLWithPath: serverConfig.argv[0], relativeTo: projectRoot.ensuringCorrectTrailingSlash)
     var serverArgs = Array(serverConfig.argv[1...])
 
-    if serverPath.suffix == ".py" {
-      serverArgs = [serverPath.pathString] + serverArgs
+    if serverPath.pathExtension == "py" {
+      serverArgs = [try serverPath.filePath] + serverArgs
```
compnerd, line 160: Does it make sense to hoist out the `try`?

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -226,19 +225,17 @@ actor ExternalBuildSystemAdapter {
...
+      if let configFileURL = jsonFiles?.sorted(by: { $0.lastPathComponent < $1.lastPathComponent }).first {
+        return configFileURL
       }
     }
 
     // Pre Swift 6.1 SourceKit-LSP looked for `buildServer.json` in the project root. Maintain this search location for
     // compatibility even though it's not a standard BSP search location.
     if let workspaceFolder = workspaceFolder,
-      localFileSystem.isFile(workspaceFolder.appending(component: "buildServer.json"))
+      FileManager.default.fileExists(at: workspaceFolder.appendingPathComponent("buildServer.json"))
```
compnerd, line 236: There is a `FileManager.fileExists(atPath:isDirectory:)` which can help ensure that the path exists and is a file (`!bIsDirectory`).

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -226,19 +225,17 @@ actor ExternalBuildSystemAdapter {
...
     {
-      return workspaceFolder.appending(component: "buildServer.json")
+      return workspaceFolder.appendingPathComponent("buildServer.json")
```
compnerd, line 238: Perhaps create a local variable for this and re-use it?

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
```diff
@@ -281,3 +278,17 @@ actor ExternalBuildSystemAdapter {
...
+  /// If the path of this URL represents a directory, ensure that it has a trailing slash.
+  ///
+  /// This is important because if we form a file URL relative to eg. file:///tmp/a would assumes that `a` is a file
+  /// and use `/tmp` as the base, not `/tmp/a`.
+  var ensuringCorrectTrailingSlash: URL {
+    guard self.isFileURL else {
+      return self
+    }
+    // `URL(fileURLWithPath:)` checks the file system to decide whether a directory exists at the path.
+    return URL(fileURLWithPath: self.path)
```
compnerd, line 292: This will always `stat`. If we know a path is a directory, it is helpful to pass that as a hint to the initializer (`URL(filePath:isDirectoryHint:)`).

Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
```diff
@@ -291,11 +280,11 @@ package actor SwiftPMBuildSystem: BuiltInBuildSystem {
     self.testHooks = testHooks
     self.connectionToSourceKitLSP = connectionToSourceKitLSP
 
-    guard let destinationToolchainBinDir = toolchain.swiftc?.parentDirectory else {
+    guard let destinationToolchainBinDir = toolchain.swiftc?.deletingLastPathComponent() else {
       throw Error.cannotDetermineHostToolchain
     }
 
-    let hostSDK = try SwiftSDK.hostSwiftSDK(AbsolutePath(destinationToolchainBinDir))
+    let hostSDK = try SwiftSDK.hostSwiftSDK(try AbsolutePath(validating: destinationToolchainBinDir.filePath))
```
compnerd, line 287: Is the second `try` needed?

Sources/SKTestSupport/Utils.swift
```diff
@@ -14,7 +14,7 @@
 package import Foundation
 package import LanguageServerProtocol
 import SwiftExtensions
-package import struct TSCBasic.AbsolutePath
+import struct TSCBasic.AbsolutePath
```
compnerd, line 17: Why not scope this import?

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -150,19 +147,20 @@ package final actor ToolchainRegistry {
...
   package init(
-    installPath: AbsolutePath? = nil,
+    installPath: URL? = nil,
     environmentVariables: [ProcessEnvironmentKey] = ["SOURCEKIT_TOOLCHAIN_PATH"],
-    xcodes: [AbsolutePath] = [_currentXcodeDeveloperPath].compactMap({ $0 }),
-    darwinToolchainOverride: String? = ProcessEnv.block["TOOLCHAINS"],
-    _ fileSystem: FileSystem = localFileSystem
+    xcodes: [URL] = [_currentXcodeDeveloperPath].compactMap({ $0 }),
+    libraryDirectories: [URL] = FileManager.default.urls(for: .libraryDirectory, in: .allDomainsMask),
+    pathEnvironmentVariables: [ProcessEnvironmentKey] = ["SOURCEKIT_PATH", "PATH", "Path"],
```
compnerd, line 154: Was `ProcessEnvironmentKey` the modification that I had done to enable case insensitive environment variables on Windows? If so, aren't `Path` and `PATH` handled the same already?

compnerd (commented):
This is a huge amount of work, and I'm quite happy to see this done. I think that this helps make the codebase easier to understand and more robust.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: I think the `try` makes sense here because it makes it clear which expression can throw.

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: Good suggestion, I even wrote a convenience method for that in this PR `FileManager.isFile`. Though I think it’s unlikely that somebody would have a directory named `buildServer.json` 😉

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: Good suggestion 👍🏽

Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift
ahoppen: We only call this when launching a BSP server, ie. once per workspace, so the extra `stat` doesn’t matter and I’d like to avoid issues caused by making wrong assumptions about whether we have a directory here.

Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
ahoppen: No, probably an artifact of the refactoring. Thanks for spotting.

Sources/SKTestSupport/Utils.swift
ahoppen: It is scoped. We enable internal imports by default.

https://github.com/swiftlang/sourcekit-lsp/blob/bce09932ed87cb121b61bbbca1b3e970fa2c0d01/CMakeLists.txt#L15

https://github.com/swiftlang/sourcekit-lsp/blob/bce09932ed87cb121b61bbbca1b3e970fa2c0d01/Package.swift#L8

Sources/ToolchainRegistry/ToolchainRegistry.swift
ahoppen: Oh, I didn’t know that was the purpose of `ProcessEnvironmentKey`. Guess we don’t need `Path` then.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/ToolchainRegistryTests/ToolchainRegistryTests.swift
```diff
@@ -42,339 +42,401 @@ final class ToolchainRegistryTests: XCTestCase {
...
+      let tr = ToolchainRegistry(
+        installPath: nil,
+        environmentVariables: [],
+        xcodes: [xcodeDeveloper],
+        libraryDirectories: [],
+        pathEnvironmentVariables: [],
+        darwinToolchainOverride: nil
+      )
 
+      assertEqual(await tr.toolchains.map(\.path), [])
```
hamishknight, line 148: Isn't `isEmpty` equivalent? Or did you mean to use `compactMap`?

hamishknight (approved):

Tests/ToolchainRegistryTests/ToolchainRegistryTests.swift
ahoppen: It is. But this has the nice side effect of printing which toolchains paths it found when the assertion fails.

PR #1833, opened 2024-11-18
Skip tests that use background indexing when running tests with a Swift 5.10 toolchain

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1714 changed the background preparation mode to `enabled` but a Swift 5.10 toolchain does not support `--experimental-prepare-for-indexing`. Thus, these tests fail. Skip tests that rely on background indexing when testing SourceKit-LSP with a Swift 5.10 host toolchain.


ahoppen:
@swift-ci Please test

PR #1834, opened 2024-11-19
[cmake] Build with CMake on macOS

Steelskin:
* Add CXX as a language. This is required to link in the standard library on macOS, which is used by dependencies.
* Set up rpath for the macOS binary. This is necessary to produce a working binary post-installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.16.1)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
 
 project(SourceKit-LSP
-  LANGUAGES C Swift)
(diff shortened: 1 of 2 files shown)


ahoppen:
Thanks for the PR. Out of curiosity: What’s your motivation to build SourceKit-LSP using CMake on macOS?

Steelskin:
> Thanks for the PR. Out of curiosity: What’s your motivation to build SourceKit-LSP using CMake on macOS?

We are building our own Swift toolchain and would like to share most of the build infrastructure with our existing Windows build, which uses CMake. Specifically, we would like to lower total install size so we are building with shared libraries.

Steelskin:
I am temporarily moving this PR to draft as I am changing the installation directories in other repositories. See this [comment](https://github.com/swiftlang/swift-tools-support-core/pull/493#issuecomment-2489346062) for details.

PR #1840, opened 2024-11-22
Fix quadratic performance issue in `AsyncQueue<Serial>`

### LINKED ISSUE #1725: Quadratic performance issue in `AsyncQueue`
Adding an item to `AsyncQueue<Serial>` is linear in the number of pending queue items, thus adding n items to an `AsyncQueue` before any can execute is in O(n^2). This decision was made intentionally because the primary use case for `AsyncQueue` was to track pending LSP requests, of which we don’t expect to have too many pending requests at any given time.

But we should be able to resolve the quadratic issue of `AsyncQueue<Serial>` by making a new task only depend on the last item in the queue, which then transitively depends on all the previous items.

ahoppen:
Adding an item to `AsyncQueue` was linear in the number of pending queue items, thus adding n items to an `AsyncQueue` before any can execute is in O(n^2). This decision was made intentionally because the primary use case for `AsyncQueue` was to track pending LSP requests, of which we don’t expect to have too many pending requests at any given time.

While we can't fix the quadratic performance issue in general, we can resolve the quadratic issue of `AsyncQueue<Serial>` by making a new task only depend on the last item in the queue, which then transitively depends on all the previous items. `AsyncQueue<Serial>` are the queues that are most likely to contain many items.

Fixes #1725
rdar://137886469

### DIFF
--- a/Sources/SwiftExtensions/AsyncQueue.swift
+++ b/Sources/SwiftExtensions/AsyncQueue.swift
@@ -28,28 +28,31 @@ extension Task: AnyTask {
 
 /// A type that is able to track dependencies between tasks.
 package protocol DependencyTracker: Sendable {
-  /// Whether the task described by `self` needs to finish executing before
-  /// `other` can start executing.
-  func isDependency(of other: Self) -> Bool
+  /// Which tasks need to finish before a task described by `self` may start executing.
+  /// `pendingTasks` is sorted in the order in which the tasks were enqueued to `AsyncQueue`.
+  func dependencies(in pendingTasks: [PendingTask<Self>]) -> [PendingTask<Self>]
 }
 
 /// A dependency tracker where each task depends on every other, i.e. a serial
 /// queue.
 package struct Serial: DependencyTracker {
-  package func isDependency(of other: Serial) -> Bool {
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftExtensions/AsyncQueue.swift
```diff
@@ -28,28 +28,31 @@ extension Task: AnyTask {
 
 /// A type that is able to track dependencies between tasks.
 package protocol DependencyTracker: Sendable {
-  /// Whether the task described by `self` needs to finish executing before
-  /// `other` can start executing.
-  func isDependency(of other: Self) -> Bool
```
hamishknight, line 33: Could keep both requirements and add a default implementation of `dependencies(in:)` that does the filtering. Don't really mind though

hamishknight (approved):

Sources/SwiftExtensions/AsyncQueue.swift
ahoppen: I think it’s good to make adopters a little more aware of the potentially quadratic issue.

PR #1841, opened 2024-11-22
Log integers and booleans as public information by default in `NonDarwinLogger`

ahoppen:
`os_log` doesn’t consider integers and bools as private information and neither should `NonDarwinLogger`.

rdar://138659073


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1842, opened 2024-11-22
Reply with `null` to `shutdown` request

### LINKED ISSUE #1733: `shutdown` result is non-null
### Swift version

swift-6.0-DEVELOPMENT-SNAPSHOT-2024-09-17-a

### Platform

macOS 14.5

### Editor

https://github.com/ChimeHQ/LanguageClient

### Description

In response to a shutdown request, the server responds with:

`{"jsonrpc":"2.0","id":2,"result":{}}`

That result of `{}` doesn't match the specification, which indices that the result should be `null`.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown

This really isn't that big a deal, but we found it, and figured it was worth reporting.

### Steps to Reproduce

send a shutdown request

### Logging

_No response_

ahoppen:
The LSP spec says the result of `shutdown` is `null`, not an empty object.

Fixes #1733
rdar://137886488

### DIFF
--- a/Sources/LanguageServerProtocol/Requests/ShutdownRequest.swift
+++ b/Sources/LanguageServerProtocol/Requests/ShutdownRequest.swift
@@ -18,10 +18,17 @@
 /// - Returns: Void.
 public struct ShutdownRequest: RequestType, Hashable {
   public static let method: String = "shutdown"
-  public typealias Response = VoidResponse
+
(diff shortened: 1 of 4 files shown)


hamishknight (approved):

ahoppen:
@swift-ci Please test

PR #1851, opened 2024-12-04
Add `textDocument/doccDocumentation` request

matthewbastien:
Support documentation live preview in Swift files by handling requests to convert documentation using Swift DocC.

This PR adds a new request method `textDocument/doccDocumentation` that responds with a JSON encoded RenderNode created using Swift DocC's ConvertService. This request has two parameters:

1. A text document for which to convert documentation
2. An optional position within the text document that will be used to find the nearest documentable symbol

The request is only supported on macOS and Linux until Swift DocC can be made to support Windows.


matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+        let symbolUSR = cursorInfo.symbolInfo.usr
+      else {
+        throw ResponseError.internalError("Unable to retrieve symbol graph for the document")
+      }
+      guard let rawSymbolGraph = symbolGraph.data(using: .utf8) else {
+        throw ResponseError.internalError("Unable to encode symbol graph")
+      }
+      externalIDsToConvert = [symbolUSR]
+      symbolGraphs.append(rawSymbolGraph)
+      overridingDocumentationComments[symbolUSR] = nearestDocumentableSymbol.documentationComments
```
ahoppen, line 94: The general pattern we use for other requests is that we would create a requirement on `LanguageService` for the `convertDocumentation` request and implement it in `SwiftLanguageService`. `ClangLanguageService` could then reply with `.noDocumentation`. That way you don’t have to do any switching on the document’s language in here. This also means that we might not need `DocumentationManager` at all. `SyntaxTreeManager` and `MacroExpansionManager` are separate types because they are actors that cache data but `DocumentationManager` doesn’t store any data and thus doesn’t really manage anything.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  private func setResult(result symbol: Symbol) {
+    if result == nil {
+      result = symbol
+    }
+  }
+
+  private func visitNamedDeclWithMemberBlock(
+    node: some SyntaxProtocol,
```
ahoppen, line 189: Should this be the following? If so, we also don’t need to pass `name` because we can access it from `node.name`.

```suggestion
    node: some NamedDeclSyntax,
```

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+
+  /// Designated entry point for `DocumentableSymbolFinder`.
+  static func find(
+    in nodes: some Sequence<Syntax>,
+    at cursorPosition: AbsolutePosition
+  ) -> Symbol? {
+    let visitor = DocumentableSymbolFinder(cursorPosition)
+    for node in nodes {
+      visitor.walk(node)
+    }
```
ahoppen, line 157: Should we stop iteration once we’ve found a result? Similarly, should we stop visitation once we’ve found a result (eg. by making this a `SyntaxAnyVisitor` and implementing `visitAny`).

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+    return .skipChildren
+  }
+
+  override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+    // A variable declaration is only documentable if there is only one pattern binding
+    guard node.bindings.count == 1,
+      let identifier = node.bindings.first!.pattern.as(IdentifierPatternSyntax.self)
+    else {
```
ahoppen, line 264: You can simplify this slightly

```suggestion
    guard let identifier = node.bindings.only.pattern.as(IdentifierPatternSyntax.self) else {
```

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+    visitNamedDeclWithMemberBlock(node: node, name: node.name, memberBlock: node.memberBlock)
+  }
+
+  override func visit(_ node: ProtocolDeclSyntax) -> SyntaxVisitorContinueKind {
+    visitNamedDeclWithMemberBlock(node: node, name: node.name, memberBlock: node.memberBlock)
+  }
+
+  override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
+    let symbolPosition = node.name.positionAfterSkippingLeadingTrivia
+    if node.range.contains(cursorPosition) || cursorPosition < symbolPosition {
```
ahoppen, line 228: I find the `|| cursorPosition  < symbolPosition` checks a little odd. I assume there’s some motivating example for them. If so, what is it?

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+    name: TokenSyntax,
+    memberBlock: MemberBlockSyntax
+  ) -> SyntaxVisitorContinueKind {
+    if cursorPosition <= memberBlock.leftBrace.positionAfterSkippingLeadingTrivia {
+      setResult(node: node, position: name.positionAfterSkippingLeadingTrivia)
+    } else if let child = DocumentableSymbolFinder.find(
+      in: memberBlock.children(viewMode: .sourceAccurate),
+      at: cursorPosition
+    ) {
+      setResult(result: child)
```
ahoppen, line 199: Why do we need to create a new `DocumentableSymbolFinder` here? Couldn’t we just visit children in this visitor?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -772,6 +786,15 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await request.reply { try await rename(request.params) }
     case let request as RequestAndReply<ShutdownRequest>:
       await request.reply { try await shutdown(request.params) }
+    #if canImport(SwiftDocC)
+    case let request as RequestAndReply<ConvertDocumentationRequest>:
```
ahoppen, line 790: Could you order this alphabetically?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -772,6 +786,15 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
...
+      await request.reply {
+        try await documentationManager.convertDocumentation(
+          request.params.textDocument.uri,
+          at: request.params.position
+        )
```
ahoppen, line 795: Could you extract this to a method. I want to do as little logic as possible in here. This method should just dispatch the request.

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
```diff
@@ -0,0 +1,595 @@
...
+        func func1️⃣tion() {
+          // Some import2️⃣ant function contents.
+        }3️⃣
+        """,
+      expectedResponses: [
+        .renderNode(kind: .symbol, path: "test/function()"),
+        .renderNode(kind: .symbol, path: "test/function()"),
+        .renderNode(kind: .symbol, path: "test/function()"),
+        .error(.noDocumentation),
+      ]
```
ahoppen, line 45: I would prefer if the marker was a key of the `expectedResponses`. That way you don’t need to count elements in `expectedResponse` to match them up to the markers.

---

We generally start counting the position markers at 1 instead of 0. It would be nice if these tests could be consistent.

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
```diff
@@ -0,0 +1,595 @@
...
+      positions: positions,
+      expectedResponses: [.renderNode(kind: .symbol, containing: "A structure containing important information")]
+    )
+
+    // Change the content of the documentation comment
+    testClient.send(
+      DidChangeTextDocumentNotification(
+        textDocument: VersionedTextDocumentIdentifier(uri, version: 2),
+        contentChanges: [
+          TextDocumentContentChangeEvent(range: positions["0️⃣"]..<positions["0️⃣"], text: " very")
```
ahoppen, line 369:
```suggestion
          TextDocumentContentChangeEvent(range: Range(positions["0️⃣"]), text: " very")
```

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
```diff
@@ -0,0 +1,595 @@
...
+      switch expectedResponses[index] {
+      case .renderNode(let expectedKind, let expectedPath, let expectedContents):
+        XCTAssertEqual(
+          renderNode.kind,
+          expectedKind,
+          "render node kind did not match expected value at position \(marker)",
+          file: file,
+          line: line
+        )
+        if let expectedPath = expectedPath {
```
ahoppen, line 544:
```suggestion
        if let expectedPath {
```

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
```diff
@@ -0,0 +1,595 @@
...
+          XCTAssertEqual(
+            renderNode.identifier.path,
+            "/documentation/\(expectedPath)",
+            "render node path did not match expected value at position \(marker)",
+            file: file,
+            line: line
+          )
+        }
+        if let expectedContents = expectedContents {
```
ahoppen, line 553:
```suggestion
        if let expectedContents {
```

ahoppen (commented):
Thank you for implementing this @matthewbastien! Looking forward to seeing docc-support in VS Code.

Contributor Documentation/LSP Extensions.md
```diff
@@ -169,6 +169,94 @@ interface SKCompletionOptions {
...
+At the moment this request is only available on macOS and Linux. If SourceKit-LSP supports
+this request it will add `textDocument/convertDocumentation` to its experimental server
+capabilities.
+
+```ts
+export interface ConvertDocumentationParams {
+    /**
+     * The document to render documentation for.
+     */
+    textDocument: TextDocumentIdentifier;
```
ahoppen, line 191: Could you use 2 spaces for indentation here?

Contributor Documentation/LSP Extensions.md
```diff
@@ -169,6 +169,94 @@ interface SKCompletionOptions {
...
+
+    /**
+     * The document location at which to lookup symbol information.
+     * 
+     * This parameter is only used in Swift files to determine which symbol to render.
+     * The position is ignored for markdown and tutorial documents.
+     */
+    position: Position;
```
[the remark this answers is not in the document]
matthewbastien: Good point. Done.

Contributor Documentation/LSP Extensions.md
```diff
@@ -169,6 +169,94 @@ interface SKCompletionOptions {
 }
 ```
 
+## `textDocument/convertDocumentation`
```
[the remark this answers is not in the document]
matthewbastien: It is definitely docs-specific. However, it doesn't really fit in with the `textDocument/hover` request since its purpose is to drive a live preview editor rather than hover (though we could definitely have SwiftDocC drive our hovers in the future if we want).

I've opted to go with renaming this request to `textDocument/doccDocumentation`.

Sources/LanguageServerProtocol/Messages.swift
```diff
@@ -70,6 +70,7 @@ public let builtinRequests: [_RequestType.Type] = [
   ShowMessageRequest.self,
   ShutdownRequest.self,
   SignatureHelpRequest.self,
+  ConvertDocumentationRequest.self,
```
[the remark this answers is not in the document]
matthewbastien: I have reordered this alphabetically.

> Could you also add the request to MessageHanldingDependencyTracker.swift?

This case should be covered already by the fact that it implements the `TextDocumentRequest` protocol. However, I did notice a bug in `MessageHanldingDependencyTracker` whereby the case for `TextDocumentRequest` takes precedence over the `RenameRequest` case. I've fixed it as part of this PR, but I can break that out into its own PR if you want.

Sources/LanguageServerProtocol/Requests/ConvertDocumentationRequest.swift
```diff
@@ -0,0 +1,148 @@
...
+///
+/// - Returns: A ``ConvertDocumentationResponse`` for the given location, which may contain an error
+///   message if documentation could not be converted. This error message can be displayed to the user
+///   in the live preview editor.
+///
+/// ### LSP Extension
+///
+/// This request is an extension to LSP supported by SourceKit-LSP.
+/// The client is expected to display the documentation in an editor using swift-docc-render.
+public struct ConvertDocumentationRequest: TextDocumentRequest, Hashable {
```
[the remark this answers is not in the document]
matthewbastien: Done.

Sources/LanguageServerProtocol/Requests/ConvertDocumentationRequest.swift
```diff
@@ -0,0 +1,148 @@
...
+  var message: String {
+    switch self {
+    case .indexNotAvailable:
+      return "The index is not availble to complete the request"
+    case .noDocumentation:
+      return "No documentation could be rendered for the position in this document"
+    case .symbolNotFound(let symbolName):
+      return "Could not find symbol \(symbolName) in the project"
+    }
+  }
```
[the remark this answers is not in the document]
matthewbastien: This code is no longer present after switching to using `ResponseError` rather than a custom error type.

Sources/SourceKitLSP/Documentation/DocCServer.swift
```diff
@@ -0,0 +1,178 @@
...
+package struct DocCServer {
+  private let server: DocumentationServer
+  private let jsonEncoder = JSONEncoder()
+  private let jsonDecoder = JSONDecoder()
+
+  init(peer peerServer: DocumentationServer? = nil, qualityOfService: DispatchQoS) {
+    server = DocumentationServer.createDefaultServer(qualityOfService: qualityOfService, peer: peerServer)
+  }
+
+  func convert(
```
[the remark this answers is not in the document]
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocCServer.swift
```diff
@@ -0,0 +1,178 @@
...
+    documentationBundleLocation: URL?,
+    documentationBundleDisplayName: String,
+    documentationBundleIdentifier: String,
+    symbolGraphs: [Data],
+    overridingDocumentationComments: [String: [String]] = [:],
+    emitSymbolSourceFileURIs: Bool,
+    markupFiles: [Data],
+    tutorialFiles: [Data],
+    convertRequestIdentifier: String,
+    completion: @escaping (_: Result<ConvertResponse, DocCServerError>) -> Void
```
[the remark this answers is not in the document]
matthewbastien: Yeah that makes things a little easier to read. Done.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+import SwiftSyntax
+
+package final actor DocumentationManager {
+  private weak var sourceKitLSPServer: SourceKitLSPServer?
+
+  private let doccServer: DocCServer
+
+  init(sourceKitLSPServer: SourceKitLSPServer) {
+    self.sourceKitLSPServer = sourceKitLSPServer
+    self.doccServer = DocCServer(peer: nil, qualityOfService: .background)
```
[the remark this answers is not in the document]
matthewbastien: Changed to `default` since SwiftDocC does not have a means for changing the QoS on a per-request basis.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+    }
+    guard let workspace = await sourceKitLSPServer.workspaceForDocument(uri: documentURI) else {
+      throw ResponseError.workspaceNotOpen(documentURI)
+    }
+
+    let snapshot = try sourceKitLSPServer.documentManager.latestSnapshot(documentURI)
+    let targetId = await workspace.buildSystemManager.canonicalTarget(for: documentURI)
+    var moduleName: String? = nil
+    if let targetId = targetId {
+      moduleName = await workspace.buildSystemManager.buildTarget(named: targetId)?.displayName
```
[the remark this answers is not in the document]
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Eventually the `DocumentationManager` will have cached data to handle symbol lookups (coming in a subsequent PR after this gets merged). Additionally, documentation requests will need to be handled by cross-referencing multiple LanguageServices. E.g. markdown documentation requests will need to retrieve a symbol USR and symbol graph from Swift and vice versa. It made the request handling logic difficult to follow and had duplicate code in multiple LanguageServices.

This should hopefully make more sense when I've pushed up the subsequent PRs that rely on this change. We can have further discussions and refactoring then.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Good point! Done.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,274 @@
...
+  private let doccServer: DocCServer
+
+  init(sourceKitLSPServer: SourceKitLSPServer) {
+    self.sourceKitLSPServer = sourceKitLSPServer
+    self.doccServer = DocCServer(peer: nil, qualityOfService: .background)
+  }
+
+  func convertDocumentation(
+    _ documentURI: DocumentURI,
+    at position: Position = Position(line: 0, utf16index: 0)
```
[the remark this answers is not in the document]
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: It's a little more complicated than this since child nodes should take precedence over their parents when the cursor is within a `MemberBlock`. However, I was able to make this a little bit cleaner by adding a `depth` parameter that is incremented whenever a `MemberBlock` is entered. Visitation is now stopped when we already found a symbol and are at a depth less than said symbol.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Yeah, I have no idea why I added this check. I've removed it.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Good point. Done.

Sources/SourceKitLSP/SourceKitLSPServer.swift
matthewbastien: Done.

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
matthewbastien: Done.

Sources/SourceKitLSP/SourceKitLSPServer.swift
matthewbastien: Done.

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
matthewbastien: Done.

Tests/SourceKitLSPTests/ConvertDocumentationTests.swift
matthewbastien: Done.

Contributor Documentation/LSP Extensions.md
matthewbastien: Done.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

Contributor Documentation/LSP Extensions.md
```diff
@@ -169,6 +169,44 @@ interface SKCompletionOptions {
 }
 ```
 
+## `textDocument/doccDocumentation`
+
+New request that returns a RenderNode for a symbol at a given location that can then be
+rendered in an editor by `swiftlang/swift-docc-render`.
```
ahoppen, line 175: `swiftlang/swift-docc-render` looks a little bit like another LSP request to me, which I found confusing. 

```suggestion
Returns a RenderNode for a symbol at a given location that can then be rendered in an editor by https://github.com/swiftlang/swift-docc-render/.
```

Contributor Documentation/LSP Extensions.md
```diff
@@ -169,6 +169,44 @@ interface SKCompletionOptions {
...
+
+This request uses Swift DocC's convert service to convert documentation for a Swift symbol
+or Markup/Tutorial file. It responds with a string containing a JSON encoded `RenderNode`
+or an error if the documentation could not be converted. This error message can be displayed
+to the user in the live preview editor.
```
ahoppen, line 180: I think the part about the error no longer applies since we use standard LSP error responses now.

Sources/LanguageServerProtocol/Requests/DoccDocumentationRequest.swift
```diff
@@ -0,0 +1,53 @@
...
+///   - position: The document location at which to lookup symbol information. (optional)
+///
+/// - Returns: A ``DoccDocumentationResponse`` for the given location, which may contain an error
+///   message if documentation could not be converted. This error message can be displayed to the user
+///   in the live preview editor.
+///
+/// ### LSP Extension
+///
+/// This request is an extension to LSP supported by SourceKit-LSP.
+/// The client is expected to display the documentation in an editor using swift-docc-render.
```
ahoppen, line 30: This doc comment should (mostly) match what we have in `LSP Extensions.md` (at least that’s what we’re doing for the other request IIRC.

Sources/SKTestSupport/TestSourceKitLSPClient.swift
```diff
@@ -237,17 +237,21 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
 
   // MARK: - Sending messages
 
-  /// Send the request to `server` and return the request result.
-  package func send<R: RequestType>(_ request: R) async throws -> R.Response {
-    return try await withCheckedThrowingContinuation { continuation in
+  package func sendWithRawResponse<R: RequestType>(_ request: R) async -> Result<R.Response, ResponseError> {
```
ahoppen, line 240: Why do we need to return a `Result` here? Can’t callers catch any errors using `do { } catch { }` if needed?

Sources/SourceKitLSP/Documentation/DocCServer.swift
```diff
@@ -0,0 +1,202 @@
...
+      markupFiles: markupFiles,
+      tutorialFiles: tutorialFiles,
+      miscResourceURLs: [],
+      symbolIdentifiersWithExpandedDocumentation: nil
+    )
+    let response = try await makeRequest(
+      messageType: ConvertService.convertMessageType,
+      messageIdentifier: convertRequestIdentifier,
+      request: request
+    );
```
ahoppen, line 85: Superfluous semicolon

Sources/SourceKitLSP/Documentation/DocCServer.swift
```diff
@@ -0,0 +1,202 @@
...
+    }
+    return ConvertResponse(doccConvertResponse: doccConvertResponse)
+  }
+
+  private func makeRequest<Request: Encodable & Sendable>(
+    messageType: DocumentationServer.MessageType,
+    messageIdentifier: String,
+    request: Request
+  ) async throws(DocCServerError) -> DocumentationServer.Message {
+    let result: Result<DocumentationServer.Message, DocCServerError> = await withCheckedContinuation { continuation in
```
ahoppen, line 117: Could you use `withCheckedThrowingContinuation` instead of going through `Result`?

Sources/SourceKitLSP/Documentation/DocCServer.swift
```diff
@@ -0,0 +1,202 @@
...
+        encodedMessage = try JSONEncoder().encode(message)
+      } catch {
+        return continuation.resume(returning: .failure(.encodingFailure(error)))
+      }
+      // Send the request to the server and decode the response
+      server.process(encodedMessage) { response in
+        let decodeMessageResult: Result<DocumentationServer.Message, DocCServerError> = Result {
+          try JSONDecoder().decode(DocumentationServer.Message.self, from: response)
+        }
+        .flatMapError { .failure(.decodingFailure($0)) }
```
ahoppen, line 142: Could the error transformation be done using `do {} catch {}` instead of using `Result`?

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,294 @@
...
+      throw ResponseError.internalError("SourceKit-LSP is shutting down")
+    }
+    guard let workspace = await sourceKitLSPServer.workspaceForDocument(uri: documentURI) else {
+      throw ResponseError.workspaceNotOpen(documentURI)
+    }
+
+    let snapshot = try sourceKitLSPServer.documentManager.latestSnapshot(documentURI)
+    let targetId = await workspace.buildSystemManager.canonicalTarget(for: documentURI)
+    var moduleName: String? = nil
+    if let targetId = targetId {
```
ahoppen, line 48:
```suggestion
    if let targetId {
```

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,294 @@
...
+      symbolGraphs: symbolGraphs,
+      overridingDocumentationComments: overridingDocumentationComments,
+      emitSymbolSourceFileURIs: false,
+      markupFiles: [],
+      tutorialFiles: [],
+      convertRequestIdentifier: UUID().uuidString
+    )
+    guard let renderNodeData = convertResponse.renderNodes.first else {
+      throw ResponseError.internalError("SwiftDocC did not return any render nodes")
+    }
```
ahoppen, line 118: I don’t know much about docc but when would the response contain multiple render nodes? And would it be safe to ignore any but the first in that case?

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,294 @@
...
+    guard let renderNode = String(data: renderNodeData, encoding: .utf8) else {
+      throw ResponseError.internalError("Failed to encode render node from SwiftDocC")
+    }
+    return DoccDocumentationResponse(renderNode: renderNode)
+  }
+}
+
+package enum ConvertDocumentationError {
+  case indexNotAvailable
```
ahoppen, line 127: I think this error is never emitted.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
```diff
@@ -0,0 +1,294 @@
...
+      return "Could not find symbol \(symbolName) in the project"
+    }
+  }
+}
+
+fileprivate extension ResponseError {
+  static func requestFailed(_ error: ConvertDocumentationError) -> ResponseError {
+    return ResponseError(code: .requestFailed, message: error.message)
+  }
+}
```
ahoppen, line 147: Should be obsolete because you added the same extension in `Error.swift`, right?

Sources/LanguageServerProtocol/Messages.swift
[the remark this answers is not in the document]
ahoppen: Good catch. Thanks for noticing!

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: This is actually slightly different accepting a `ConvertDocumentationError` so that I don't have to duplicate error strings. I've renamed it to `requestFailed(convertError:)` to make this clearer.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Removed.

You got a sneak peak at the next PR 😜

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: DocC will only emit multiple render nodes if you request multiple USRs to convert.

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocCServer.swift
matthewbastien: Unfortunately, `withCheckedThrowingContinuation` does not allow throwing a specific error type. I had to use `Result` here to get around this.

Sources/SourceKitLSP/Documentation/DocCServer.swift
matthewbastien: Done.

Sources/SourceKitLSP/Documentation/DocCServer.swift
matthewbastien: Removed.

Sources/SKTestSupport/TestSourceKitLSPClient.swift
matthewbastien: Done.

Sources/LanguageServerProtocol/Requests/DoccDocumentationRequest.swift
matthewbastien: That makes more sense. I changed the comment to match what's in `LSP Extensions.md`.

Contributor Documentation/LSP Extensions.md
matthewbastien: Changed this to talk about failure messages that we send via the LSP request failed error.

Contributor Documentation/LSP Extensions.md
matthewbastien: Done.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

Sources/SourceKitLSP/Documentation/DocumentationManager.swift
ahoppen: Could you add that as a comment so we don’t have to wonder about it again?

ahoppen (approved):
Great work. Let’s get this in.

PR #1852, opened 2024-12-04
Include parameters in initializer document symbols

matthewbastien:
I noticed that initializer document symbols do not contain the initializer's parameters while working on the DocC support. This PR adds the parameters to the symbol name in the same way as functions.


ahoppen (approved):
Beautiful

matthewbastien:
@swift-ci please test

PR #1857, opened 2024-12-05
Don’t implicitly cancel code completion requests on document edits

ahoppen:
As the user types, we filter the code completion results. Cancelling the completion request on every keystroke means that we will never build the initial list of completion results for this code completion session if building that list takes longer than the user's typing cadence (eg. for global completions) and we will thus not show any completions.


ahoppen:
@swift-ci Please test

bnbarham (approved):

hamishknight (approved):

ahoppen:
@swift-ci Please test macOS

PR #1858, opened 2024-12-06
Fix background indexing behavior if a source file is included in two targets via a symlink

ahoppen:
Consider the following scenario: A project has target A containing A.swift an target B containing B.swift. B.swift is a symlink to A.swift. When A.swift is modified, both the dependencies of A and B need to be marked as having an out-of-date preparation status, not just A.

### DIFF
--- a/Sources/BuildSystemIntegration/BuildSystemManager.swift
+++ b/Sources/BuildSystemIntegration/BuildSystemManager.swift
@@ -1033,11 +1033,11 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
     return response.items
   }
 
-  /// Returns all source files in the project that can be built.
+  /// Returns all source files in the project.
   ///
   /// - SeeAlso: Comment in `sourceFilesAndDirectories` for a definition of what `buildable` means.
-  package func buildableSourceFiles() async throws -> [DocumentURI: SourceFileInfo] {
-    return try await sourceFilesAndDirectories(includeNonBuildableFiles: false).files
+  package func sourceFiles(includeNonBuildableFiles: Bool) async throws -> [DocumentURI: SourceFileInfo] {
+    return try await sourceFilesAndDirectories(includeNonBuildableFiles: includeNonBuildableFiles).files
   }
 
   /// Get all files and directories that are known to the build system, ie. that are returned by a `buildTarget/sources`
@@ -1093,7 +1093,7 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
   }
 
   package func testFiles() async throws -> [DocumentURI] {
-    return try await buildableSourceFiles().compactMap { (uri, info) -> DocumentURI? in
+    return try await sourceFiles(includeNonBuildableFiles: false).compactMap { (uri, info) -> DocumentURI? in
       guard info.isPartOfRootProject, info.mayContainTests else {
         return nil
       }
--- a/Sources/SourceKitLSP/Workspace.swift
+++ b/Sources/SourceKitLSP/Workspace.swift
@@ -62,6 +62,51 @@ fileprivate func firstNonNil<T>(
   return try await defaultValue()
 }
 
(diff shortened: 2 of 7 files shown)


ahoppen:
@swift-ci Please test

Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -671,6 +671,26 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
...
+  /// This is useful in the following scenario: A project has target A containing A.swift an target B containing B.swift
+  /// B.swift is a symlink to A.swift. When A.swift is modified, both the dependencies of A and B need to be marked as
+  /// having an out-of-date preparation status, not just A.
+  package func sourceFilesWithSameRealpath(as documents: [DocumentURI]) async -> [DocumentURI] {
+    let realPaths = Set(documents.map { $0.symlinkTarget ?? $0 })
+    return await orLog("") {
+      var result: [DocumentURI] = []
+      let filesAndDirectories = try await sourceFilesAndDirectories(includeNonBuildableFiles: true)
+      for file in filesAndDirectories.files.keys {
+        if realPaths.contains(file.symlinkTarget ?? file) && !realPaths.contains(file) {
```
hamishknight, line 686: Nit: Couldn't this be `if let symlinkTarget = file.symlinkTarget, realPaths.contains(symlinkTarget)`?

hamishknight (approved):

Sources/BuildSystemIntegration/BuildSystemManager.swift
ahoppen: No, because then we would not be reporting an edit in the following scenario:
A is a symlink to B. A and B are in the source files array. A gets modified. We should also consider B as modified but we wouldn’t with your suggested change because `B` doesn’t have a symlink target (since it’s not a symlink).

ahoppen:
Updated the PR to dod the inference of source files with equivalent realpaths to `Workspace.filesDidChange` and also added a realpath cache in there.

Sources/BuildSystemIntegration/BuildSystemManager.swift
hamishknight: To be clear, my suggestion was equivalent to the original code, since when `symlinkTarget` is `nil` it becomes `realPaths.contains(file) && !realPaths.contains(file)`. That being said, good catch!

ahoppen: Ah, and I realized that the second part of the check should have been `!documents.contains(file)` upon revisiting the changes.

hamishknight (approved):

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -62,6 +62,51 @@ fileprivate func firstNonNil<T>(
...
+      for file in filesAndDirectories.keys {
+        if realPaths.contains(realpath(of: file)) && !documents.contains(file) {
+          result.append(file)
+        }
+      }
+      return result
+    } ?? []
+  }
+
+  func filesDidChange(_ events: [FileEvent]) {
```
bnbarham, line 103: Did you check what event ends up being sent for changing a symlink? Is it a delete + add or just a change?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SourceKitLSP/Workspace.swift
bnbarham: Sure, I was wondering if we could limit it to eg. just delete/add rather than on change

ahoppen: Even if I checked how it behaves on macOS, I don’t want to make inferences about how it behaves on various Linux distributions, so I think erroring on the side of caution is better here. Plus, recomputing a single realpath when a file is changed shouldn’t be that bad.

PR #1860, opened 2024-12-06
Only load BSP servers that support one of the language we are interested in

ahoppen:
Instead of unconditionally loading any BSP server from one of the search locations, only load those that support, C, C++, Obj-C, Obj-C++ or Swift.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1862, opened 2024-12-06
Log which file caused the package to be reloaded

ahoppen:


ahoppen:
@swift-ci Please test

PR #1864, opened 2024-12-06
Don’t re-index file if we are waiting for its preparation

ahoppen:
This fixes an issue that caused the index progress to not make progress at the beginning when opening a new project because of the following:

When opening the project, we would cause package resolution and schedule indexing of all source files after the package has been resolved (and dependencies been downloaded). But dependency resolution will add the dependency files to be written to disk and the editor will inform us of these modified files, which will cause us to schedule a new index operation for these files. Thus, when the initial indexing of the files finishes, we won’t increase the number of tasks that we have already indexed.

To fix this, don’t schedule new indexing of these files if we are still waiting for their preparation to start.

This should not be a big performance win because even before this change, when we were trying to index the file the second time, we would have realized that the index store is up to date and thus do a no-op. The main benefit of this change is to have the index counter be more accurate.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -51,7 +51,8 @@ private struct OpaqueQueuedIndexTask: Equatable {
 }
 
 private enum InProgressIndexStore {
-  /// We are waiting for preparation of the file's target to finish before we can index it.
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Accidentally hit Merge without waiting for an approval. Will address any feedback in a follow-up PR.

PR #1865, opened 2024-12-06
Improve performance of `sourceFilesAndDirectories`

ahoppen:
`SourceFilesAndDirectoriesKey` contained all source files in the project and computing its hash value was pretty expensive. The key didn’t really provide any value here because the only way it changes is if the build targets change and if that’s the case, we already clear `cachedSourceFilesAndDirectories`, so we can just avoid the hash value computation.

### DIFF
--- a/Sources/BuildSystemIntegration/BuildSystemManager.swift
+++ b/Sources/BuildSystemIntegration/BuildSystemManager.swift
@@ -54,6 +54,12 @@ package struct SourceFileInfo: Sendable {
   /// from non-test targets or files that don't actually contain any tests.
   package var mayContainTests: Bool
 
+  /// Source files returned here fall into two categories:
+  ///  - Buildable source files are files that can be built by the build system and that make sense to background index
+  ///  - Non-buildable source files include eg. the SwiftPM package manifest or header files. We have sufficient
+  ///    compiler arguments for these files to provide semantic editor functionality but we can't build them.
+  package var isBuildable: Bool
+
   fileprivate func merging(_ other: SourceFileInfo?) -> SourceFileInfo {
     guard let other else {
       return self
@@ -61,7 +67,8 @@ package struct SourceFileInfo: Sendable {
     return SourceFileInfo(
       targets: targets.union(other.targets),
       isPartOfRootProject: other.isPartOfRootProject || isPartOfRootProject,
-      mayContainTests: other.mayContainTests || mayContainTests
+      mayContainTests: other.mayContainTests || mayContainTests,
+      isBuildable: other.isBuildable || isBuildable
     )
   }
 }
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -327,10 +327,8 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
 
   private var cachedTargetSources = RequestCache<BuildTargetSourcesRequest>()
 
-  /// The parameters with which `SourceFilesAndDirectories` can be cached in `cachedSourceFilesAndDirectories`.
   private struct SourceFilesAndDirectoriesKey: Hashable {
     let includeNonBuildableFiles: Bool
```
bnbarham, line 331: `cachedSourceFilesAndDirectories` is going to have two almost-identical `SourceFilesAndDirectories`. Would it make sense to just include whether the file is buildable as part of `SourceFileInfo?

ahoppen: Good idea 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -327,11 +334,9 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
 
   private var cachedTargetSources = RequestCache<BuildTargetSourcesRequest>()
 
-  /// The parameters with which `SourceFilesAndDirectories` can be cached in `cachedSourceFilesAndDirectories`.
-  private struct SourceFilesAndDirectoriesKey: Hashable {
-    let includeNonBuildableFiles: Bool
-    let sourcesItems: [SourcesItem]
-  }
+  /// `SourceFilesAndDirectories` is a global property that only gets reset when the build targets change and thus
+  /// has no real key.
```
bnbarham, line 338: Is it being a cache still useful in that case?

bnbarham (approved):

Sources/BuildSystemIntegration/BuildSystemManager.swift
ahoppen: Constructing `SourceFilesAndDirectories` is still non-trivial and caching it has significant performance benefits because we call it during `canonicalTarget(for:)`, which gets call pretty frequently.

bnbarham: Sorry, I didn't mean to not "cache" it. I just meant is there any point using `Cache` rather than just having a `SourceFilesAndDirectories` directly.

ahoppen: We don’t need to but I prefer to because it aligns this property with all the other cached properties and allow us to use the `get` function that computes on-demand, stores it in a `Task` etc. So, there’s still enough of a benefit from using `Cache`.

bnbarham: Okay 👍

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1867, opened 2024-12-06
Cached transformed results in `Cache`

ahoppen:
The transform to get the transformed result might be expensive, so we should cache its result.

### DIFF
--- a/Sources/SwiftExtensions/Cache.swift
+++ b/Sources/SwiftExtensions/Cache.swift
@@ -33,15 +33,28 @@ package class Cache<Key: Sendable & Hashable, Result: Sendable> {
     return try await task.value
   }
 
-  package func get(
+  /// Get the value cached for `key`. If no value exists for `key`, try deriving the result from an existing cache entry
+  /// that satisfies `canReuseKey` by applying `transform` to that result.
+  package func getDerived(
     isolation: isolated any Actor,
-    whereKey keyPredicate: (Key) -> Bool,
-    transform: @Sendable @escaping (Result) -> Result
+    _ key: Key,
+    canReuseKey: @Sendable @escaping (Key) -> Bool,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftExtensions/Cache.swift
```diff
@@ -40,7 +40,10 @@ package class Cache<Key: Sendable & Hashable, Result: Sendable> {
   ) async throws -> Result? {
     for (key, value) in storage {
       if keyPredicate(key) {
-        return try await transform(value.value)
+        let transformed = Task { try await transform(value.value) }
+        // Cache the transformed result.
+        storage[key] = transformed
+        return try await transformed.value
```
hamishknight, line 46: Won't this re-apply the transform for every subsequent access?

ahoppen: Indeed, this was wrong on so many different levels. Thanks for catching it in the review, this would have been a pain to debug.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftExtensions/Cache.swift
```diff
@@ -33,15 +33,28 @@ package class Cache<Key: Sendable & Hashable, Result: Sendable> {
...
   ) async throws -> Result? {
-    for (key, value) in storage {
-      if keyPredicate(key) {
-        return try await transform(value.value)
+    if let cached = storage[key] {
+      // If we have a value for the requested key, prefer that
+      return try await cached.value
+    }
+
+    // See if we ca
```
hamishknight, line 49: Incomplete comment?

hamishknight (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1868, opened 2024-12-06
Only show call-like occurrences in call hierarchy.

ahoppen:
`extension MyTask: AnyTask {}` includes an occurrence of `MyTask.cancel` to mark it as an override of `AnyTask.cancel` but we shouldn't show the extension in the call hierarchy.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1869, opened 2024-12-06
Batch updates to the syntactic test index on fileDidChange events

ahoppen:
This is more performant. In particular adding a new task to `indexingQueue` for each file to rescan can hit the quadratic issue in `AsyncQueue` if many files were changed.


ahoppen:
@swift-ci Please test

PR #1870, opened 2024-12-06
Don’t escape `{` inside placeholder snippets

ahoppen:
When `\{` is included inside an LSP placeholder, in VS Code will insert `\{` verbatim. At least in VS Code, we only need to escape the closing brace.

@woolsweater I’m not sure which editor you used in the video you included in https://github.com/swiftlang/sourcekit-lsp/pull/1831 but it looked like that did include the opening brace correctly. Could you check that this change doesn’t break your editor?

### DIFF
--- a/Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
+++ b/Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
@@ -37,13 +37,13 @@ public func rewriteSourceKitPlaceholders(in input: String, clientSupportsSnippet
         placeholders.latest.contents += text
       }
 
-    case let .curlyBrace(brace):
+    case .escapeInsidePlaceholder(let character):
       if placeholders.isEmpty {
-        result.append(brace)
+        result.append(character)
       } else {
-        // Braces are only escaped _inside_ a placeholder; otherwise the client
-        // would include the backslashes literally.
-        placeholders.latest.contents.append(contentsOf: ["\\", brace])
+        // A closing brace is only escaped _inside_ a placeholder; otherwise the client would include the backslashes
+        // literally.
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
```diff
@@ -134,8 +134,8 @@ private func tokenize(_ input: String) -> [SnippetToken] {
 private enum SnippetToken {
   /// A placeholder delimiter.
   case placeholderOpen, placeholderClose
-  /// One of '{' or '}', which may need to be escaped in the output.
-  case curlyBrace(Character)
+  /// '}', which may need to be escaped in the output.
```
woolsweater, line 137: Argh, yup, this was my mistake, @ahoppen; the LSP spec doesn't require escaping of open braces. This change works fine for me. (My editor is Emacs, using the eglot LSP client.)

Related note; `$` and `\` are also supposed to be escaped; I don't know if you think it's worth adding those too. I would be happy to do it myself as well.

woolsweater (approved):

Sources/SourceKitLSP/Swift/RewriteSourceKitPlaceholders.swift
ahoppen: Thanks for checking with emacs and noticing that we also need to escape `$` and `\`. I missed that in the LSP the first time I looked for a definition of the escaping rules.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1871, opened 2024-12-06
Treat `$/setTrace` as a freestanding message

ahoppen:
`$/setTrace` changes a global configuration setting but it doesn't affect the result of any other request. To avoid blocking other requests on a `$/setTrace` notification the client might send during launch, we treat it as a freestanding message. Also, we don't do anything with this notification at the moment, so it doesn't matter.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1872, opened 2024-12-06
Don’t show warning message when opening projects that don't support background indexing

ahoppen:
Since we enabled background indexing by default, the user is no longer explicitly opting into it. A user might be exclusively working with compilation database projects or BSP server without background indexing support and thus not care that we switched the background indexing default. We shouldn’t bother them with a warning message every time they launch sourcekit-lsp.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1873, opened 2024-12-06
Create child scopes for SwiftPM operations

ahoppen:
Some SwiftPM functions check whether their observability scope has errors. If we use the same observability scope for all SwiftPM operations during SourceKit-LSP’s lifetime, a single SwiftPM error will set the `hasError` bit in that observability scope for the entirety of SourceKit-LSP’s lifetime, impacting all upcoming SwiftPM operations.

Creating a separate child scope for every operation fixes


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1882, opened 2024-12-11
Fix an error that causes SourceKit-LSP to exit immediately on Windows

ahoppen:
For reasons that are completely oblivious to me, `DispatchIO.write`, which is used to write LSP responses to stdout fails with error code 5 on Windows unless we call `AbsolutePath(validating:)` on some URL first.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1885, opened 2024-12-13
Check that we get correct package version even if package manifest is invalid

ahoppen:
Addresses https://github.com/swiftlang/sourcekit-lsp/pull/1846/files#r1856529041

rdar://141311942


ahoppen:
@swift-ci Please test

hamishknight (approved):
Thanks!

PR #1886, opened 2024-12-13
When a sourcekitd diagnostics request fails, show the request error as a diagnostic on the source file

### LINKED ISSUE #1812: sourcekitd silently fails
### Swift version

swiftlang-6.0.3.1.4

### Platform

macOS 15

### Editor

VS Code

### Description

When there is an SDK mismatch, sourcekitd silently fails

### Steps to Reproduce

1. Have a tasks.json entry that points to an SDK from one Xcode installation
2. Build that task
3. Configure the SDK in settings to points to an SDK from a different Xcode installation
4. Reload the window
5. Next time sourcekit-lsp starts, no IntelliSense is available
6. Open the output channel and should see an error like this
```
2024-11-05 13:32:44.195827-0500 0x47bb1    Error       0x0                  29814  0    sourcekit-lsp: [org.swift.sourcekit-lsp:request-26] Received sourcekitd response:
error response (Request Failed): Loading the standard library failed
```

### Logging

_No response_

ahoppen:
The missing diagnostics might be due to an error that the user can fix. Report it to them.

Fixes #1812
rdar://139514623

### DIFF
--- a/Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
+++ b/Sources/SourceKitLSP/Swift/DiagnosticReportManager.swift
@@ -110,11 +110,27 @@ actor DiagnosticReportManager {
       keys.compilerArgs: compilerArgs as [SKDRequestValue],
     ])
 
-    let dict = try await self.sourcekitd.send(
-      skreq,
-      timeout: options.sourcekitdRequestTimeoutOrDefault,
-      fileContents: snapshot.text
-    )
+    let dict: SKDResponseDictionary
+    do {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):
Very interested to know whether you end up finding this annoying or not 😅. Is it worth having an option to turn it off, or will we just drop this if the general feedback is that it isn't helpful?

ahoppen:
I’d say let’s just enable this by default for now. If we discover that it’s more distracting than helpful, we would probably just revert this PR again.

PR #1887, opened 2024-12-13
Support semantic functionality in generated interfaces if the client supports `getReferenceDocument`

### LINKED ISSUE #1155: Navigation in generated .swiftinterface files

ahoppen:
This allows us to provide semantic functionality inside the generated interfaces, such as hover or jump-to-definition.

Fixes #1155
rdar://125663597

### DIFF
--- a/Sources/SourceKitLSP/Swift/OpenInterface.swift
+++ b/Sources/SourceKitLSP/Swift/OpenInterface.swift
@@ -10,23 +10,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(>=6)
 import Foundation
-package import LanguageServerProtocol
 import SKLogging
-import SKUtilities
-import SourceKitD
-#else
-import Foundation
-import LanguageServerProtocol
-import SKLogging
-import SKUtilities
-import SourceKitD
-#endif
 
-struct GeneratedInterfaceInfo {
-  var contents: String
-}
+#if compiler(>=6)
+package import LanguageServerProtocol
+#else
+import LanguageServerProtocol
+#endif
 
 extension SwiftLanguageService {
   package func openGeneratedInterface(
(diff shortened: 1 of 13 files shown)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -135,6 +189,80 @@ final class SwiftInterfaceTests: XCTestCase {
...
+    let response =
+      try await project.testClient.send(
+        DefinitionRequest(
+          textDocument: TextDocumentIdentifier(mainUri),
+          position: mainPositions["1️⃣"]
+        )
+      )
+    let referenceDocumentUri = try XCTUnwrap(response?.locations?.only).uri
+    let referenceDocument = try await project.testClient.send(GetReferenceDocumentRequest(uri: referenceDocumentUri))
+    let (referenceDocumentPositions, expectedReferenceDocumentContents) = DocumentPositions.extract(
```
bnbarham, line 228: I wouldn't be super surprised if we add/change imports and/or interface printing at some point. Could we just search for eg. Lib/foo/init + get the position of String from the result rather than assuming it's exactly this interface?

bnbarham (approved):
Nice!

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
ahoppen: I guess you’re right 😉

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -10,129 +10,51 @@
...
-      }
-    } catch {
-      return GeneratedInterfaceDetails(uri: uri, position: nil)
+    if case .dictionary(let experimentalCapabilities) = self.capabilityRegistry.clientCapabilities.experimental,
+      case .bool(true) = experimentalCapabilities["workspace/getReferenceDocument"]
+    {
+      return GeneratedInterfaceDetails(uri: try urlData.uri, position: position)
     }
+    let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent(urlData.displayName)
+    try await generatedInterfaceManager.snapshot(of: urlData).text.write(
```
Wilfred, line 50: @ahoppen this looks like we no longer cache generated .swiftinterface files unless the LSP client says it supports `workspace/getReferenceDocument`, i.e. only VS Code users using the official extension.

Is that intentional? I would have thought we'd still want to cache regardless. Switching from Swift 6.1 to 6.2 has regressed go-to-def time on AttributedString from 1.0 to 4.6 seconds in my testing, because Foundation.swiftinterface is no longer cached.

ahoppen: We should be caching the generated interfaces in the `generatedInterfaceManager`. Yes, we will always write it to disk but the previous implementation of assuming that we already had the correct generated interface if a file with that name exists on disk was incorrect because generated interfaces may look different based on the compiler arguments that are used to generate them (and which we infer from the file from which you jump-to-definition) – unfortunately I don’t have a nice example of this right now.

The regression is worrisome though, it probably means that the `generatedInterfaceManager` isn’t quite doing its job the way we intended it to. Would you be able to file an issue for this, ideally with [extended logging enabled](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Enable%20Extended%20Logging.md) so we can investigate this further?

Wilfred: I've filed an issue: https://github.com/swiftlang/sourcekit-lsp/issues/2369

PR #1889, opened 2024-12-14
Allow workspace options to affect build system search

bnbarham:
There were a few places that options only took place *after* determining
a build system, even though we have multiple that impact the search (eg.
`defaultBuildSystem` and `searchPaths`).

Do not search up the directory tree when searching for a build system
for a workspace. Only allow this for implicit workspaces (created when
no existing workspace can handle a file) and only up to any existing
workspace roots.


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Tests/BuildSystemIntegrationTests/SwiftPMBuildSystemTests.swift
```diff
@@ -976,9 +976,9 @@ final class SwiftPMBuildSystemTests: XCTestCase {
         .appendingPathComponent("pkg")
         .appendingPathComponent("Sources")
         .appendingPathComponent("lib")
-      let projectRoot = SwiftPMBuildSystem.projectRoot(for: workspaceRoot, options: .testDefault())
 
-      assertEqual(projectRoot, tempDir.appendingPathComponent("pkg", isDirectory: true))
+      let projectRoot = SwiftPMBuildSystem.projectRoot(for: workspaceRoot, options: .testDefault())
+      assertNil(projectRoot)
```
ahoppen, line 981: Because there’s no `await` in here, you can use `XCTAssertNil`

```suggestion
      XCTAssertNil(projectRoot)
```

Sources/BuildSystemIntegration/DetermineBuildSystem.swift
```diff
@@ -31,11 +31,13 @@ import struct TSCBasic.AbsolutePath
...
 
 /// Determine which build system should be started to handle the given workspace folder and at which folder that build
-/// system's project root is (see `BuiltInBuildSystem.projectRoot(for:options:)`).
+/// system's project root is (see `BuiltInBuildSystem.projectRoot(for:options:)`). `onlyConsiderRoot` controls whether
+/// paths outside the root should be considered (eg. configuration files in the user's home directory).
 ///
 /// Returns `nil` if no build system can handle this workspace folder.
 package func determineBuildSystem(
   forWorkspaceFolder workspaceFolder: DocumentURI,
+  onlyConsiderRoot: Bool,
```
ahoppen, line 40: Option unused?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -209,29 +209,40 @@ package actor SourceKitLSPServer {
...
-      // was added to it and thus currently doesn't know that it can handle that file. In that case, we shouldn't open
-      // a new workspace for the same root. Instead, the existing workspace's build system needs to be reloaded.
+
       let uri = DocumentURI(url)
-      guard let buildSystemSpec = determineBuildSystem(forWorkspaceFolder: uri, options: self.options) else {
+      let options = SourceKitLSPOptions.merging(base: self.options, workspaceFolder: uri)
+
+      // Some build systems consider paths outside of the folder (eg. BSP has settings in the home directory). If we
+      // allowed those paths, then the very first folder that the file is in would always be its own build system - so
+      // skip them in that case.
```
ahoppen, line 233: Comment out of date if `onlyConsiderRoot` is not used?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -818,26 +829,20 @@ extension SourceKitLSPServer {
 
   /// Creates a workspace at the given `uri`.
   ///
-  /// If the build system that was determined for the workspace does not satisfy `condition`, `nil` is returned.
+  /// A workspace does not necessarily have any buildsystem attached to it, in which case `buildSystemSpec` may be
```
ahoppen, line 832:
```suggestion
  /// A workspace does not necessarily have any build system attached to it, in which case `buildSystemSpec` may be
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -818,26 +829,20 @@ extension SourceKitLSPServer {
...
-      base: self.options,
-      override: SourceKitLSPOptions(
-        path: workspaceFolder.fileURL?
-          .appendingPathComponent(".sourcekit-lsp")
-          .appendingPathComponent("config.json")
-      )
-    )
-    logger.log("Creating workspace at \(workspaceFolder.forLogging) with options: \(options.forLogging)")
+
+    logger.log("Creating workspace at \(workspaceFolder.forLogging)")
```
ahoppen, line 845: Did you move the logging of `options` somewhere else? If not, I would prefer to continue logging them. I had a case once where logging them was crucial to figure out a bug.

bnbarham: It's logged in its entirety in the line below this

Sources/SourceKitLSP/SourceKitLSPServer.swift
bnbarham: Whoops, it's... not out of date. I just forgot to actually pass it through into searching for BSP. No test for it because it's all environment variables but given this, probably need a way to test that 🤔

bnbarham: Noticed some more fun when looking at this. Today we can have:
1. Explicit workspace + compilation db build system using build/compile_commands.json
2. File unhandled in build directory
3. Implicit workspace created as the roots don't match up (the explicit project root doesn't include build, but the implicit will)

IMO it's also a little confusing that we just ignore an invalid DB. Might change project root to just check for existence + return the directory that the DB is in rather than the directory it was searched from.

bnbarham:
@swift-ci please test

Tests/BuildSystemIntegrationTests/CompilationDatabaseTests.swift
```diff
@@ -190,45 +190,12 @@ final class CompilationDatabaseTests: XCTestCase {
...
         }
       ]
-      """.write(to: tempDir.appendingPathComponent("compile_commands.json"), atomically: true, encoding: .utf8)
+      """.write(to: tempDir.appendingPathComponent(JSONCompilationDatabase.dbName), atomically: true, encoding: .utf8)
 
       XCTAssertNotNil(tryLoadCompilationDatabase(directory: tempDir))
     }
   }
 
-  func testJSONCompilationDatabaseFromCustomDirectory() async throws {
```
bnbarham, line 199: There's a test for basically this in `WorkspaceTests` now (since `tryLoadCompilationDatabase` only loads from the directory itself). I'd be open to adding `projectRoot` tests if you wanted though.

bnbarham:

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
```diff
@@ -62,10 +62,29 @@ fileprivate enum Cachable<Value> {
...
+      } + [
+        // These default search paths match the behavior of `clangd`
+        try! RelativePath(validating: "."),
+        try! RelativePath(validating: "build"),
+      ]
+
+    return
+      searchPaths
+      .lazy
+      .compactMap({ searchPath in
```
ahoppen, line 77: Nit: Could be a trailing closure.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -967,4 +968,174 @@ final class WorkspaceTests: XCTestCase {
...
+        -sdk '\(defaultSDKPath)'
+        """
+      } else {
+        ""
+      }
+
+    let project = try await MultiFileTestProject(files: [
+      "projA/.sourcekit-lsp/config.json": """
+      {
+          "defaultWorkspaceType": "compilationDatabase"
```
ahoppen, line 1042: I’d use 2 spaces for indentation here. Same in `testWorkspaceOptionsCanAddSearchPaths`.

ahoppen (approved):
Two small comments, otherwise LGTM. Thanks!

bnbarham:
@swift-ci please test

bnbarham:

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
```diff
@@ -62,10 +64,29 @@ fileprivate enum Cachable<Value> {
...
+        let path = workspaceFolder.appending(searchPath)
+        if FileManager.default.isFile(at: path.appendingPathComponent(JSONCompilationDatabase.dbName))
+          || FileManager.default.isFile(at: path.appendingPathComponent(FixedCompilationDatabase.dbName))
+        {
+          return path
+        }
+
+        return nil
+      }
+      .first
```
ahoppen, line 89: If you place your `compile_commands.json` inside `build`, the project root should still be the directory containing `build` and not `build`. So, I think if we have any match in `searchPaths`, weh should return `workspaceFolder` and otherwise return `nil` here.

bnbarham: IIRC project root is really only used by the build system itself, where it (at least in the case of compile commands) needs to know the actual directory containing it so that it can reload. What I failed to mention in the commit is that doing this actually fixes an issue with the `build` case. Consider:
  * `build/compile_commands.json`
  * Request comes in for a file within `build` that isn't in compile commands
  * Isn't handled by the existing build system, so we try create an implicit
  * The workspace for the implicit isn't known, with the path being passed in just the directory that we are currently walking up
  * Thus we end up creating a new build system with the root of `build`
  * That doesn't match any of the existing roots and so we also create a new workspace

We also weren't checking for implicit workspaces here previously, which meant that we ended up spawning a clangd for basically every request 😅

Having said all that, we can track that separately if you think we do need project root == workspace folder for something.

ahoppen: Do we need to actually track the directory that `compile_commands.json` came from, though? Wouldn’t we arrive at the same location if we do the same computation when creating the workspace?

I noticed this because we use the last path component of the project root to identify the project for logging here and using the directory containing `compile_commands.json` might make this `build`, which is not as useful.

https://github.com/swiftlang/sourcekit-lsp/blob/233f2e6f371ca27a8ab19403db7137f1e56a9094/Sources/BuildSystemIntegration/BuildSystemManager.swift#L152

bnbarham: > Do we need to actually track the directory that compile_commands.json came from, though? Wouldn’t we arrive at the same location if we do the same computation when creating the workspace?

Taking my above example but with there being a top level `proj` folder (that contains `build`):
  * Opening `proj` as the workspace would give `proj` as the workspace root (as `build` is in the search paths)
  * Implicit workspaces start from the directory containing the file and work up
  * `build` contains the file + `compile_commands.json` and thus we return `proj/build` as the workspace + project root

We could pass both the workspace + project root in and then log workspace.lastPathComponent + relative path from workspace -> project?

ahoppen: I would prefer a solution where the project root continues being `proj`. I’m not sure if we do any path comparisons at the moment (checking a directory is a subdirectory of the project root) but I could also imagine us doing that in the future and having `build` as the project root can easily cause issues there because you wouldn’t expect it.

bnbarham:
@swift-ci please test

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
bnbarham: I've split this into `projectRoot` and `configPath`, also removing `projectRoot` where it's not actually used.

bnbarham:
@swift-ci please test

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
```diff
@@ -118,18 +140,20 @@ package actor CompilationDatabaseBuildSystem: BuiltInBuildSystem {
...
+    if let compdb = tryLoadCompilationDatabase(file: configPath) {
       self.compdb = compdb
     } else {
       return nil
     }
+    self.connectionToSourceKitLSP = connectionToSourceKitLSP
+
+    self.configPath = configPath
+    self.fileWatchers = [
+      FileSystemWatcher(globPattern: try configPath.filePath, kind: [.change])
```
ahoppen, line 155: There is one semi-common pattern that we’d be missing now: 

The user has a build that stores `compile_commands.json` in `myfancybuildfolder`. In order to pick it up, they create a symlink from `<project root>/compile_commands.json` to `myfancybuildfolder/compile_commands.json`. Now, we no longer reload the compile commands if `myfancybuildfolder/compile_commands.json` changes.

Though, maybe we should tell these users to use the custom search path config setting (which they can commit since we’ve got `config.json` since 6.0).

Sources/BuildSystemIntegration/DetermineBuildSystem.swift
```diff
@@ -49,24 +51,26 @@ package func determineBuildSystem(
...
-      }
-      #else
-      return nil
+      spec = SwiftPMBuildSystem.searchForConfig(in: workspaceFolderUrl, options: options)
       #endif
     }
+
+    if spec != nil {
+      return spec
+    }
```
ahoppen, line 73:
```suggestion
    if let spec {
      return spec
    }
```
 
Just looks a little more swifty.

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -507,6 +507,17 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     )
   }
 
+  public static func merging(base: SourceKitLSPOptions, workspaceFolder: DocumentURI) -> SourceKitLSPOptions {
```
ahoppen, line 510: Can we make this `package` instead of `public`?

Tests/BuildSystemIntegrationTests/CompilationDatabaseTests.swift
```diff
@@ -260,8 +231,14 @@ final class CompilationDatabaseTests: XCTestCase {
...
     try await withTestScratchDir { tempDir in
-      try "".write(to: tempDir.appendingPathComponent("compile_commands.json"), atomically: true, encoding: .utf8)
-      XCTAssertNil(tryLoadCompilationDatabase(directory: tempDir))
+      let dbFile = tempDir.appendingPathComponent(JSONCompilationDatabase.dbName)
+
+      try "".write(
+        to: dbFile,
+        atomically: true,
+        encoding: .utf8
+      )
```
ahoppen, line 240: Could probably fit on one line but doesn’t really matter (same below).

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -967,4 +968,171 @@ final class WorkspaceTests: XCTestCase {
...
+      """,
+      "Sources/MyLib/Bar.swift": "",
+      "build/compile_commands.json": """
+      [
+      {
+        "directory": "$TEST_DIR",
+        "command": "swiftc $TEST_DIR/src/Foo.swift -DHAVE_SETTINGS \(sdkArgs)",
+        "file": "src/Foo.swift",
+        "output": "$TEST_DIR/build/Foo.swift.o"
+      },
```
ahoppen, line 1001: I’m surprised this works. AFAIK JSON doesn’t support trailing commas…

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -967,4 +968,171 @@ final class WorkspaceTests: XCTestCase {
...
+  func testWorkspaceOptionsCanAddSearchPaths() async throws {
+    let sdkArgs =
+      if let defaultSDKPath {
+        """
+        -sdk '\(defaultSDKPath)'
+        """
+      } else {
+        ""
+      }
+
```
ahoppen, line 1091: Might be worth sharing this between test cases, similar to what we do in `BuildServerTestProject`. Also, we should replace `\` with `\\` in `defaultSDKPath`, so this doesn’t fail on Windows.

ahoppen (commented):
Looks good to me, just a few small comments.

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
```diff
@@ -182,7 +206,7 @@ package actor CompilationDatabaseBuildSystem: BuiltInBuildSystem {
 
   private func fileEventShouldTriggerCompilationDatabaseReload(event: FileEvent) -> Bool {
     switch event.uri.fileURL?.lastPathComponent {
-    case "compile_commands.json", "compile_flags.txt":
+    case JSONCompilationDatabase.dbName, FixedCompilationDatabase.dbName:
       return true
     default:
       return false
```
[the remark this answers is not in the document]
bnbarham: Oh nice catch, forgot all about this after I switched to configPath. Yes, definitely should be able to!

Sources/BuildSystemIntegration/CompilationDatabaseBuildSystem.swift
bnbarham: I slight preference for only watching the config file itself, but if you think that's a common enough pattern then I'm fine switching it back (and I'll add a comment to explain why we're doing that).

Tests/SourceKitLSPTests/WorkspaceTests.swift
bnbarham: TIL JSONDecoder allows trailing commas. I'll remove them in any case, doesn't really matter here.

bnbarham:
@swift-ci please test

bnbarham:

Sources/SKOptions/SourceKitLSPOptions.swift
ahoppen: I’ll switch it to `package` as well in a separate PR.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #1594, opened 2024-07-27
Return range from `textDocument/hover`


WeetHet:
### Description

In Zed the returned range from the `textDocument/hover` is used to not redraw hover every time the mouse cursor moves. As per the LSP specification:
```js
/**
 * The result of a hover request.
 */
export interface Hover {
    /**
     * The hover's content
     */
    contents: MarkedString | MarkedString[] | MarkupContent;

    /**
     * An optional range is a range inside a text document
     * that is used to visualize a hover, e.g. by changing the background color.
     */
    range?: Range;
}
```

Example of the rust-analyzer response:
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "contents": {
      "kind": "markdown",
      "value": "\n```rust\nrust_playground\n```\n\n```rust\nfn main()\n```"
    },
    "range": {
      "start": {
        "line": 0,
        "character": 3
      },
      "end": {
        "line": 0,
        "character": 7
      }
    }
  }
}
```

ahoppen:
Synced to Apple’s issue tracker as rdar://132623508

ahoppen:
It should be possible to get the token at the requested position from the syntax tree and return that as the range.

PR #1899, opened 2024-12-30
feat: Support range in `textDocument/hover`

iMostfa:
Hi, 

in this PR i add range when hovering.

this should solve #1594.

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -288,7 +288,12 @@ package actor SkipUnless {
         HoverRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
       )
       let hover = try XCTUnwrap(response, file: file, line: line)
-      XCTAssertNil(hover.range, file: file, line: line)
+      XCTAssertEqual(
+        hover.range,
+        .init(line: 1, utf16index: 5) ..< .init(line: 1, utf16index: 9),
+        file: file,
+        line: line
+      )
       guard case .markupContent(let content) = hover.contents else {
         throw ExpectedMarkdownContentsError()
       }
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -290,7 +290,7 @@ package actor SkipUnless {
       let hover = try XCTUnwrap(response, file: file, line: line)
       XCTAssertEqual(
         hover.range,
-        .init(line: 1, utf16index: 5) ..< .init(line: 1, utf16index: 9),
+        Position(line: 1, utf16index: 5)..<Position(line: 1, utf16index: 9),
         file: file,
         line: line
       )
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -746,8 +746,7 @@ extension SwiftLanguageService {
     if let snapshot = try? await latestSnapshot(for: uri) {
       let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
       if let token = tree.token(at: snapshot.absolutePosition(of: position)) {
-        let range: Range<Position> = snapshot.absolutePositionRange(of: token.position..<token.endPosition)
-        tokenRange = range
+        tokenRange = snapshot.absolutePositionRange(of: token.position..<token.endPosition)
       }
     }
 
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -741,9 +741,19 @@ extension SwiftLanguageService {
         """
     }
 
+    var tokenRange: Range<Position>?
+
+    if let snapshot = try? await latestSnapshot(for: uri) {
+      let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
+      if let token = tree.token(at: snapshot.absolutePosition(of: position)) {
+        let range: Range<Position> = snapshot.absolutePositionRange(of: token.position..<token.endPosition)
+        tokenRange = range
+      }
+    }
+
     return HoverResponse(
       contents: .markupContent(MarkupContent(kind: .markdown, value: joinedDocumentation)),
-      range: nil
+      range: tokenRange
     )
   }
 
--- a/Tests/SourceKitLSPTests/HoverTests.swift
+++ b/Tests/SourceKitLSPTests/HoverTests.swift
@@ -32,7 +32,9 @@ final class HoverTests: XCTestCase {
         This is a doc comment for S.
 
         Details.
-        """
+        """,
+      expectedRange:
+        .init(line: 3, utf16index: 7) ..< .init(line: 3, utf16index: 9)
     )
   }
 
@@ -81,7 +83,9 @@ final class HoverTests: XCTestCase {
         init()
         ```
 
-        """
+        """,
+      expectedRange:
+        .init(line: 3, utf16index: 4) ..< .init(line: 3, utf16index: 7)
     )
   }
 
@@ -113,7 +117,9 @@ final class HoverTests: XCTestCase {
         ```
 
         The initializer
-        """
+        """,
+      expectedRange:
+        .init(line: 5, utf16index: 4) ..< .init(line: 5, utf16index: 7)
     )
   }
 
@@ -130,7 +136,9 @@ final class HoverTests: XCTestCase {
         ```
 
         this is **bold** documentation
-        """##
+        """##,
+      expectedRange:
+        .init(line: 1, utf16index: 5) ..< .init(line: 1, utf16index: 9)
     )
   }
 
@@ -147,7 +155,9 @@ final class HoverTests: XCTestCase {
         ```
 
         this is *italic* documentation
-        """##
+        """##,
+      expectedRange:
+        .init(line: 1, utf16index: 5) ..< .init(line: 1, utf16index: 8)
     )
   }
 
@@ -168,7 +178,9 @@ final class HoverTests: XCTestCase {
         Eat an apple
 
         - Precondition: Must have an apple
-        """
+        """,
+      expectedRange:
+        .init(line: 3, utf16index: 5) ..< .init(line: 3, utf16index: 13)
     )
   }
 }
@@ -176,6 +188,7 @@ final class HoverTests: XCTestCase {
 private func assertHover(
   _ markedSource: String,
   expectedContent: String,
+  expectedRange: Range<Position>,
   file: StaticString = #filePath,
   line: UInt = #line
 ) async throws {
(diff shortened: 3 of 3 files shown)


Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -741,9 +741,19 @@ extension SwiftLanguageService {
         """
     }
 
+    var tokenRange: Range<Position>?
+
+    if let snapshot = try? await latestSnapshot(for: uri) {
+      let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
+      if let token = tree.token(at: snapshot.absolutePosition(of: position)) {
+        let range: Range<Position> = snapshot.absolutePositionRange(of: token.position..<token.endPosition)
+        tokenRange = range
```
ahoppen, line 750: Instead of using the token’s range from the syntax tree, we should be using the `key.offset` and `key.length` properties returned from sourcekitd’s cursorinfo request. This will ensure that e.g the parentheses are included in the range when doing a cursor info on `foo()`.

Tests/SourceKitLSPTests/HoverTests.swift
```diff
@@ -32,7 +32,9 @@ final class HoverTests: XCTestCase {
         This is a doc comment for S.
 
         Details.
-        """
+        """,
+      expectedRange:
+        .init(line: 3, utf16index: 7) ..< .init(line: 3, utf16index: 9)
```
ahoppen, line 37: Could you run swift-format on your changes? https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting

ahoppen (commented):
Thanks for tackling this, @iMostfa. I think that we should use the range returned from sourcekitd instead of the token’s range from the syntax tree. Could you take a look at that?

Sources/SourceKitLSP/Swift/CursorInfo.swift
```diff
@@ -103,7 +107,10 @@ struct CursorInfo {
         isDynamic: dict[keys.isDynamic] ?? false,
         isSystem: dict[keys.isSystem] ?? false,
         receiverUsrs: dict[keys.receivers]?.compactMap { $0[keys.usr] as String? } ?? [],
-        systemModule: module
+        systemModule: module,
+        line: line,
```
ahoppen, line 111: Instead of assigning to a variable, I would just use the value inline here, like we do for the other keys. Same for `column` and `length`.

```suggestion
        line: dict[keys.line] as Int?,
```

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -288,7 +288,12 @@ package actor SkipUnless {
         HoverRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
       )
       let hover = try XCTUnwrap(response, file: file, line: line)
-      XCTAssertNil(hover.range, file: file, line: line)
+      XCTAssertEqual(
+        hover.range,
+        .init(line: 2, utf16index: 6) ..< .init(line: 2, utf16index: 12),
```
ahoppen, line 293: I would just use `Position` here instead of `.init` because it’s more explicit about what it’s construing. It also means we don’t need spaces around `..<`. Same in the other tests.

```suggestion
        Position(line: 2, utf16index: 6)..<Position(line: 2, utf16index: 12),
```

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
iMostfa: i changed the logic here, 
https://github.com/swiftlang/sourcekit-lsp/pull/1899/commits/c551dd25b698e8852090f2c71f0097e44bfc7cab

i ended up using `key.column` instead of the `key.offset` because the offset gets me the offset relative to the file.

Tests/SourceKitLSPTests/HoverTests.swift
iMostfa: done

Sources/LanguageServerProtocol/Requests/SymbolInfoRequest.swift
```diff
@@ -135,6 +135,32 @@ public struct SymbolDetails: ResponseType, Hashable {
...
+  /// The column number where the symbol's declaration starts in the line.
+  ///
+  /// This value is 1-based, meaning the first character of the line is column 1.
+  /// For example, in the following line:
+  /// ```
+  /// func test() {}
+  /// ```
+  /// If `test` starts at the 6th character, the `column` would be `6`.
+  public var column: Int?
+
```
ahoppen, line 163: Since the `SymbolInfo` request is an LSP extension that  we expose publicly, would should use LSP’s `Position` type (which is UTF-16) instead of sourcekitd’s line, length and column (which is UTF-8). So, I would wrap all of these properties in a `Range<Position>` similar to how it’s returned in `HoverResponse`. And also add the addition to `LSP Extensions.md`.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
```diff
@@ -741,9 +741,19 @@ extension SwiftLanguageService {
...
 
+    var tokenRange: Range<Position>?
+
+    if let symbolUnderCursor = cursorInfoResults.first,
+      let line = symbolUnderCursor.symbolInfo.line,
+      let column = symbolUnderCursor.symbolInfo.column,
+      let length = symbolUnderCursor.symbolInfo.length
+    {
+      tokenRange = Position(line: line, utf16index: column)..<Position(line: line, utf16index: column + length)
+    }
```
ahoppen, line 752: Column and length returned by sourcekitd are UTF-8, while `Position` is UTF-16. There are methods on `DocumentSnapshot` to convert between the two representations.

ahoppen (commented):
Thanks for the update, @iMostfa. I have a few comments, mostly motivated by handling of UTF-8 to UTF-16 position conversions.

Sources/SourceKitLSP/Swift/CursorInfo.swift
```diff
@@ -71,12 +71,15 @@ struct CursorInfo {
...
       let uri = DocumentURI(filePath: filepath, isDirectory: false)
       if let snapshot = documentManager.latestSnapshotOrDisk(uri, language: .swift) {
-        let position = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
-        location = Location(uri: uri, range: Range(position))
+        let startPosition = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
+        let endPosition = snapshot.positionOf(
+          zeroBasedLine: line - 1,
+          utf8Column: column - 1 + (dict[keys.length] ?? 0)
+        )
+        location = Location(uri: uri, range: startPosition..<endPosition)
```
iMostfa, line 82: in https://github.com/swiftlang/sourcekit-lsp/pull/1899/commits/0e701fbae626f0d6e9f4a6876c0d69318d68f3bb

i found that i can use the `location` variable, which is used for `bestLocalDeclaration` and update the upperBound of it, because it was using the same lowerBound as upperBound.

Screenshot 2025-01-04 at 1 40 58 PM



Q: but is it fine for `bestLocalDeclaration` to return that upperBound ? it's not clear from the docs if it's wrong?

```swift
  /// Best known declaration or definition location without global knowledge.
  ///
  /// For a local or private variable, this is generally the canonical definition location -
  /// appropriate as a response to a `textDocument/definition` request. For global symbols this is
  /// the best known location within a single compilation unit. For example, in C++ this might be
  /// the declaration location from a header as opposed to the definition in some other
  /// translation unit.
  public var bestLocalDeclaration: Location? = nil

```

if it's not fine, i can just introduce `range` for the `SymbolDetails`

iMostfa: running tests, i see that some are failing, eg:

`DefinitionTests.testJumpToDefinitionAtEndOfIdentifier`
```swift
XCTAssertEqual failed: ("test.swift:1:5-1:8])") is not equal to ("test.swift:1:5-1:5])")
```

yet, i think the new results, are more helpful ? (because we are providing the upperBound ?

iMostfa (commented):
Thanks for the review @ahoppen.
i applied the changes, can you take a look and let me know if we shouldn't mess around with the `bestLocalDeclaration ` ?

Sources/SourceKitLSP/Swift/CursorInfo.swift
ahoppen: `bestLocalDeclaration` will refer to the location where the variable is defined. In your example above, if you do a hover on `location` on line 82, I would expect that `bestLocalDeclaration` is on line 71. But if I read the documentation of the hover request correctly, I would expect a range on line 82 in the response.

iMostfa: Ah, got it.

updated the logic here
https://github.com/swiftlang/sourcekit-lsp/pull/1899/commits/50988edaa5ab64bd0a45ddeec76041e39794f710

added  `range` in symbolDetails

ahoppen (commented):
@iMostfa I am so sorry, I only now realized that what I suggested before was bullshit and that the location in the cursor info response refers to the definition instead of the call underneath the cursor. 🤦🏽‍♂️  Having realized that, I think your initial implementation that used the syntax tree was a good one. Would you mind changing this PR back to that commit (deleting the ones you did afterwards and force-pushing)? I only have one comment left and that’s that I would use `Position` instead of `.init` in the test cases because I think it’s more explicit.

iMostfa:
no worries @ahoppen it's always nice to try multiple approaches. 
so just to verify, we should return to this one (?)  

https://github.com/swiftlang/sourcekit-lsp/pull/1899/commits/fb4ed6510e5e320065ee97f9a9f4f2f8ee2a2e45

ahoppen:
Yes, that’s what I meant 😊

iMostfa:
Hi @ahoppen 
i restored the impl. to the initial proposed solution. 

thinking about your comments now: 

>  only now realized that what I suggested before was bullshit and that the location in the cursor info response refers to the definition instead of the call underneath the cursor

as far as i understood, it refers to the definition when it's possible to refer to it, otherwise it will refer to  the symbol itself ? (well, under cursor) .. ? 

also, it means that the current solution won't hover/highlight the parentheses of a function, e.g 

`foo(param: name)`

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
ahoppen, line 750: Couldn’t this just be the following?

```suggestion
        tokenRange = snapshot.absolutePositionRange(of: token.position..<token.endPosition)
```

ahoppen (commented):
One small question, otherwise looks good to me. Thanks for your patience with me.

Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
iMostfa: of course ! 
i think i was doing something more with the variable and forgot to clean it. 
done here 
https://github.com/swiftlang/sourcekit-lsp/pull/1899/commits/b7ee7308d758ced6847f6750bbcae7308839ea18

ahoppen (approved):
🚀 Let’s land this

ahoppen:
@swift-ci Please test Windows

iMostfa:
@ahoppen 
there was a compilation error, i fixed it, let's run ci again to check

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1901, opened 2025-01-01
[Test] Link macro binaries to _SwiftSyntaxCShims objects

rintaro:
Update for https://github.com/swiftlang/swift-syntax/pull/2925 but this can be merged independently.

`_SwiftSyntaxCShims` objects weren't linked because it was a header-only library. But since actual code is being added, we need to link it.


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #1905, opened 2025-01-02
Add option to generate an output mirror file that contains the exact data sent from SourceKit-LSP to the client

ahoppen:
This could help us debug low-level issues of SourceKit-LSP to client communication such as https://github.com/swiftlang/sourcekit-lsp/issues/1890.

### DIFF
--- a/Sources/sourcekit-lsp/SourceKitLSP.swift
+++ b/Sources/sourcekit-lsp/SourceKitLSP.swift
@@ -226,6 +226,22 @@ struct SourceKitLSP: AsyncParsableCommand {
     return options
   }
 
+  /// Create a new file that can be used to use as an input or output mirror file and return a file handle that can be
+  /// used to write to that file.
+  private func createMirrorFile(in directory: URL) throws -> FileHandle {
+    let dateFormatter = ISO8601DateFormatter()
+    dateFormatter.timeZone = NSTimeZone.local
+    let date = dateFormatter.string(from: Date()).replacingOccurrences(of: ":", with: "-")
+
+    let inputMirrorURL = directory.appendingPathComponent("\(date).log")
+
+    logger.log("Mirroring input to \(inputMirrorURL)")
+    try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+    FileManager.default.createFile(atPath: try inputMirrorURL.filePath, contents: nil)
+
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -262,35 +278,35 @@ struct SourceKitLSP: AsyncParsableCommand {
     )
     cleanOldLogFiles(logFileDirectory: logFileDirectoryURL, maxAge: 60 * 60 /* 1h */)
 
+    let inputMirror = orLog("Setting up input mirror") {
+      if let inputMirrorDirectory = globalConfigurationOptions.loggingOrDefault.inputMirrorDirectory {
+        return try createMirrorFile(in: URL(fileURLWithPath: inputMirrorDirectory))
+      } else {
```
bnbarham, line 284: Usual nit with an else after a return (same just below)

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -262,35 +278,35 @@ struct SourceKitLSP: AsyncParsableCommand {
     )
     cleanOldLogFiles(logFileDirectory: logFileDirectoryURL, maxAge: 60 * 60 /* 1h */)
 
+    let inputMirror = orLog("Setting up input mirror") {
+      if let inputMirrorDirectory = globalConfigurationOptions.loggingOrDefault.inputMirrorDirectory {
```
bnbarham, line 282: Might be worth documenting that these options are global only, not per project

bnbarham (approved):

Sources/sourcekit-lsp/SourceKitLSP.swift
ahoppen: We have documentation on the `logging` configuration key that these options are global only already.

ahoppen:
@swift-ci Please test

Sources/sourcekit-lsp/SourceKitLSP.swift
bnbarham: So we do, thanks 👍

ahoppen:
@swift-ci Please test Windows

PR #1916, opened 2025-01-10
In macro tests, only link against files listed in `Objects.LinkFileList`

ahoppen:
Otherwise we can get a situation where a `.swift` file is removed from swift-syntax, its `.o` file is still in the build directory because the build folder wasn't cleaned and thus we would link against the stale `.o` file.

rdar://142666044

### DIFF
--- a/Sources/SKTestSupport/SwiftPMTestProject.swift
+++ b/Sources/SKTestSupport/SwiftPMTestProject.swift
@@ -30,8 +30,6 @@ import ToolchainRegistry
 import XCTest
 #endif
 
-private struct SwiftSyntaxCShimsModulemapNotFoundError: Error {}
-
 package class SwiftPMTestProject: MultiFileTestProject {
   enum Error: Swift.Error {
     /// The `swift` executable could not be found.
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/SwiftPMTestProject.swift
```diff
@@ -108,7 +133,7 @@ package class SwiftPMTestProject: MultiFileTestProject {
         let dir = productsDirectory.appendingPathComponent("\(moduleName).build")
         let enumerator = FileManager.default.enumerator(at: dir, includingPropertiesForKeys: nil)
         while let file = enumerator?.nextObject() as? URL {
-          if file.pathExtension == "o" {
+          if linkFileList.contains(try file.filePath) {
```
hamishknight, line 136: Should `linkFileList` be made into a Set? Although, out of curiosity, is there a reason not to just link everything in `linkFileList`?

hamishknight (approved):

Sources/SKTestSupport/SwiftPMTestProject.swift
ahoppen: > Should `linkFileList` be made into a Set? 

Good idea. I’ll do it in a follow-up PR.

> Although, out of curiosity, is there a reason not to just link everything in `linkFileList`?

I had issues with that in the past: https://github.com/swiftlang/sourcekit-lsp/pull/1367#discussion_r1623045741

PR #1920, opened 2025-01-15
Do not realpath the project root of a `SwiftPMBuildSystem`

ahoppen:
If you have a package located at `/pkg` and a symlink at `/symlink` and you open `/symlink` as a workspace, the SwiftPMBuildSystem’s project root would be `/pkg`. This would mean that it also only knew about build settings for files in `/pkg`, not in `/symlink`. Thus, whenever we were opening a file in `/symlink` we would create an implicit workspace to handle it (but which ended up having a project root at `/symlink` again) – or something close to this.

We shouldn’t need to realpath here. If you open `/symlink`, we should view `/symlink` as the project root of your workspace.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1923, opened 2025-01-16
Support background indexing when cross-compiling

kabiroberai:
Previously, if you selected a Swift SDK via the `swiftPM.swiftSDK` configuration option, background indexing would fail because the `swift build --experimental-prepare-for-indexing` invocation did not pass along the `--swift-sdk` argument. This PR fixes this by passing along that flag (+ other relevant cross-compilation flags) to `swift build`.

Best I can tell, the reason this was previously missing is simply that the PRs that added these two features ([background indexing](https://github.com/swiftlang/sourcekit-lsp/pull/1273) and [Swift SDK support](https://github.com/swiftlang/sourcekit-lsp/pull/1535)) were being worked on in parallel and didn't end up accounting for each other. Fortunately it's a trivial fix.

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -477,6 +477,34 @@ package actor SkipUnless {
     }
   }
 
+  package static func canSwiftPMCompileForIOS(
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws {
+    return try await shared.skipUnlessSupported(allowSkippingInCI: true, file: file, line: line) {
+      #if os(macOS)
+      let project = try await SwiftPMTestProject(files: [
+        "MyFile.swift": """
+        public func foo() {}
+        """
+      ])
+      do {
+        try await SwiftPMTestProject.build(
+          at: project.scratchDirectory,
+          extraArguments: [
+            "--swift-sdk", "arm64-apple-ios",
+          ]
+        )
+        return .featureSupported
+      } catch {
+        return .featureUnsupported(skipMessage: "Cannot build for iOS: \(error)")
+      }
+      #else
+      return .featureUnsupported(skipMessage: "Cannot build for iOS outside macOS by default")
+      #endif
+    }
+  }
+
   package static func canCompileForWasm(
     file: StaticString = #filePath,
     line: UInt = #line
--- a/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
+++ b/Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
@@ -626,6 +626,15 @@ package actor SwiftPMBuildSystem: BuiltInBuildSystem {
     if let configuration = options.swiftPMOrDefault.configuration {
       arguments += ["-c", configuration.rawValue]
     }
+    if let triple = options.swiftPMOrDefault.triple {
+      arguments += ["--triple", triple]
+    }
+    if let swiftSDKsDirectory = options.swiftPMOrDefault.swiftSDKsDirectory {
+      arguments += ["--swift-sdks-path", swiftSDKsDirectory]
+    }
+    if let swiftSDK = options.swiftPMOrDefault.swiftSDK {
+      arguments += ["--swift-sdk", swiftSDK]
+    }
     arguments += options.swiftPMOrDefault.cCompilerFlags?.flatMap { ["-Xcc", $0] } ?? []
     arguments += options.swiftPMOrDefault.cxxCompilerFlags?.flatMap { ["-Xcxx", $0] } ?? []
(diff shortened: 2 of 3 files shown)


kabiroberai:
Would be nice to add tests for this but as far as I can tell `BackgroundIndexingTests` are all integration-type tests so we'd need a real Swift SDK fixture to test with. Any suggestions on approaches I could take? A few options come to mind:

- Use the [implicit Darwin SDKs](https://github.com/swiftlang/swift-package-manager/pull/6828) that are available on macOS (`--swift-sdk arm64-apple-ios`), though this would require the tests to be gated by `#if os(macOS)`
- Create a complete Swift SDK fixture, but this feels like a lot of boilerplate
- Make this more of a unit test than an integration test — i.e. just check that we're passing along `--swift-sdk foo` to `swift build`. Unsure how best to fit this into the existing set of tests for the background indexing feature though.

ahoppen:
Thank you @kabiroberai. Looks like I indeed just missed passing those options through. Regarding a test case: If you can conjure one up with the iOS SDK, that would be great. To account for Xcode installations that might not include an iOS SDK, I think we should guard it behind a `SkipUnless` that check if a basic package can be compiled with`swift build --triple arm64-apple-ios`. That should give us sufficient coverage.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -477,6 +477,56 @@ package actor SkipUnless {
...
+        ],
+        manifest: """
+        let package = Package(
+          name: "MyLibrary",
+          targets: [
+           .target(name: "Lib"),
+          ]
+        )
+        """
+      )
```
ahoppen, line 506: I think you should be able to reduce this a little bit further to

```suggestion
      let project = try await SwiftPMTestProject(
        files: [
          "MyFile.swift": """
          public func foo() {}
          """
      )
```

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -477,6 +477,56 @@ package actor SkipUnless {
...
+          "--swift-sdk", "arm64-apple-ios",
+        ]
+        if let globalModuleCache = try globalModuleCache {
+          arguments += ["-Xswiftc", "-module-cache-path", "-Xswiftc", try globalModuleCache.filePath]
+        }
+        let status = try await Process.run(arguments: arguments, workingDirectory: nil)
+        guard case .terminated(code: 0) = status.exitStatus else {
+          let error = (try? String(decoding: status.stderrOutput.get(), as: UTF8.self)) ?? "unknown error"
+          return .featureUnsupported(skipMessage: "Cannot build for iOS: \(error)")
+        }
```
ahoppen, line 519: Could this be the following?

```suggestion
        try await SwiftPMTestProject.build(at: try project.scratchDirectory, extraArguments: ["--swift-sdk", "arm64-apple-ios"])
```

ahoppen (commented):
The test looks great. Thanks for adding it. I have two suggestions that should allow us to reduce the code in `SkipUnless` a little further.

kabiroberai:
Good callouts, ty!

ahoppen (approved):
Thank you. Looks good.

ahoppen:
@swift-ci Please test

ahoppen:
Could you format your changes? (https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting)

Also looks like the new test failed on Linux, though I’m not sure why the skipping wouldn’t work. Could you take a look?

kabiroberai:
Can format, my bad! As for the test not being skipped on Linux, it looks like `skipUnlessSupported` intentionally no-ops on CI by default:

https://github.com/swiftlang/sourcekit-lsp/blob/f1251c0154e9290372f716658a61aa51d4ecd76f/Sources/SKTestSupport/SkipUnless.swift#L98-L100

Looks like we can override this with `allowSkippingInCI: true` though; I'll go ahead and do that. There's an argument to be made that we should pass `allowSkippingInCI: false` on macOS, but I'm not sure whether the CI setup always includes the iOS SDK so will be conservative with this for now.

kabiroberai:
Formatted and fixed the CI issue. I _think_ we should be good now!

ahoppen:
Ah, completely forgot about `allowSkippingInCI`. Thanks for finding it. I think it’s good enough to skip it on macOS because the iOS SDK might not be installed in Xcode either (not that I would expect this on Swift CI nodes but who knows).

ahoppen:
@swift-ci Please test

kabiroberai:
@ahoppen Linux + macOS CI is green, do we need to request a Windows run?

ahoppen:
@swift-ci Please test Windows

kabiroberai:
@ahoppen ty, any chance we can cherry-pick this into 6.1?

kabiroberai:
Created #1939 if you're good with cherry-picking

PR #1927, opened 2025-01-16
Typedef enums in `CodeCompletionSwiftInterop.h`

ahoppen:
This means we can omit the `enum` prefix when using these types in the header.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #1928, opened 2025-01-16
Skip plugin tests if the host toolchain doesn’t support SourceKit plugins

ahoppen:


ahoppen:
@swift-ci Please test

PR #1931, opened 2025-01-17
build: remove SwiftSystem check

compnerd:
The swift-system package is not used in SourceKit LSP. It was required to fulfil the dependencies for swift-package-manager. With the changes in swiftlang/swift-package-manager#8231 which internalised the SystemPackage import, we no longer need this to be provided.


compnerd:
Please test with following PRs:
swiftlang/swift-package-manager#8231

@swift-ci please test

ahoppen (approved):

PR #1933, opened 2025-01-17
Split build systems for JSON compilation database and fixed compilation database

ahoppen:
I feel like the implementations are actually simpler if we split them. This will also allow us to add more advanced logic to the JSON compilation database build system in the future, such as inferring the toolchain from the compile command.


ahoppen:
@swift-ci Please test

bnbarham (approved):
Nice!

PR #1936, opened 2025-01-21
Flush the `NonDarwinLogger` before shutting down `SourceKitLSPServer`

ahoppen:
This ensures that we don't drop any log messages when stopping the LSP server, in particular while running tests.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1938, opened 2025-01-21
Support completionItem/resolve to compute documentation of a code completion item

ahoppen:
Fixes https://github.com/swiftlang/sourcekit-lsp/issues/1935

### DIFF
--- a/Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
+++ b/Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
@@ -23,28 +23,53 @@ import SwiftParser
 @_spi(SourceKitLSP) import SwiftRefactor
 import SwiftSyntax
 
-/// Data that is attached to a `CompletionItem`.
-private struct CompletionItemData: LSPAnyCodable {
-  let id: Int?
+/// Uniquely identifies a code completion session. We need this so that when resolving a code completion item, we can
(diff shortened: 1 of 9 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -134,7 +158,6 @@ class CodeCompletionSession {
           session.snapshot.uri == snapshot.uri
           && session.position == completionPosition
           && session.compileCommand == compileCommand
-          && session.clientSupportsSnippets == clientSupportsSnippets
```
hamishknight, line 137: Do we need to check whether the capabilities match, or does that not matter?

hamishknight (approved):

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
ahoppen: I don’t think that we need to. The only case where this would change is if the client initiated a code completion session and then dynamically registered the snippet support capability, which would be very unlikely since any of that dynamic registration would happen all up-front. And if should get it wrong, it’s not going to be a fatal error, the results might just look a little off (eg. no snippets even though client supports them).

ahoppen:
@swift-ci Please test

PR #1941, opened 2025-01-21
Log the request methods when logging a request's reply

ahoppen:
This should make it easier to find a request response when searching for the response method.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1943, opened 2025-01-22
Revert "Support background indexing when cross-compiling"

bnbarham:
Reverts swiftlang/sourcekit-lsp#1923


bnbarham:
No idea how this could possibly cause the error we're seeing but... let's try anyway

bnbarham:
Still have no idea what's going on here, new builds appear to be green 🤔

PR #1946, opened 2025-01-22
When the SourceKit plugins fail to load during testing, include the search base in the error

ahoppen:


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/PluginPaths.swift
```diff
@@ -112,23 +112,30 @@ private func pluginPaths(relativeTo base: URL) -> PluginPaths? {
 package var sourceKitPluginPaths: PluginPaths {
   get throws {
     struct PluginLoadingError: Error, CustomStringConvertible {
-      var description: String =
-        "Could not find SourceKit plugin. Ensure that you build the entire SourceKit-LSP package before running tests."
+      let searchBase: URL
+      var description: String {
+        """
+        Could not find SourceKit plugin. Ensure that you build the entire SourceKit-LSP package before running tests.
```
bnbarham, line 118: How would this not be the case, is there a missing dependency?

ahoppen: We can’t declare a dependency from the tests *target* on the SourceKit plugin *product*. That’s why you need to run a build, which build the product, which can then be picked up by the tests.

ahoppen: Filed https://github.com/swiftlang/swift-package-manager/issues/8245 to track that inability to declare a target -> product dependency.

ahoppen: Also added a comment to link the error message to that issue.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SKTestSupport/PluginPaths.swift
bnbarham: Ah, of course :(. Thanks for the explanation.

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1947, opened 2025-01-22
Use `realpath` when setting `SOURCEKIT_LSP_PLUGIN_SOURCEKITD_PATH_` environment variable

ahoppen:
Otherwise we run into path mismatch issues when the test products are in a path that includes symlinks, such as `/tmp`.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1950, opened 2025-01-22
Gardening

ahoppen:
- Fix build warnings, quite a few of these were reminders to clean things up once we no longer need to support testing using compilers and sourcekitd from older toolchains
- Remove `SkipUnless` checks that always evaluate to true, because we don’t support testing SourceKit-LSP using toolchains that don’t contain these features anymore


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1951, opened 2025-01-22
Load the SourceKit plugins when launching SourceKit-LSP

ahoppen:
Previously, we were only loading the SourceKit plugins during tests. Make sure we find them in the toolchain as well.

### DIFF
--- a/Sources/ToolchainRegistry/Toolchain.swift
+++ b/Sources/ToolchainRegistry/Toolchain.swift
@@ -99,6 +99,12 @@ public final class Toolchain: Sendable {
   /// The path to the Swift language server if available.
   package let sourcekitd: URL?
 
+  /// The path to the SourceKit client plugin if available.
+  package let sourceKitClientPlugin: URL?
+
+  /// The path to the SourceKit plugin if available.
+  package let sourceKitServicePlugin: URL?
+
   /// The path to the indexstore library if available.
   package let libIndexStore: URL?
 
@@ -153,6 +159,8 @@ public final class Toolchain: Sendable {
     swiftFormat: URL? = nil,
     clangd: URL? = nil,
     sourcekitd: URL? = nil,
+    sourceKitClientPlugin: URL? = nil,
+    sourceKitServicePlugin: URL? = nil,
     libIndexStore: URL? = nil
   ) {
     self.identifier = identifier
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -281,37 +293,53 @@ extension Toolchain {
...
-        dylibExt = ".so"
+      func findDylib(named name: String, searchFramework: Bool = false) -> URL? {
+        // If 'currentPlatform' is nil it's most likely an unknown linux flavor.
+        let dylibExt: String
+        if let dynamicLibraryExtension = Platform.current?.dynamicLibraryExtension {
+          dylibExt = dynamicLibraryExtension
+        } else {
+          logger.fault("Could not determine host OS. Falling back to using '.so' as dynamic library extension")
+          dylibExt = ".so"
+        }
```
bnbarham, line 304: You could pull this out so that we only run it once (and avoid the log spam if we couldn't determine the platform).

bnbarham (approved):

Sources/ToolchainRegistry/Toolchain.swift
ahoppen: We should only create `Toolchain`s when creating the `ToolchainRegistry`, so I’m not sure how we would pull this out further.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/ToolchainRegistry/Toolchain.swift
bnbarham: I suppose it's not actually clear what I highlighted here in the UI. By "pull out" I just meant moving the `dylibExt` logic out of this function (and the loop) so that we only do it once rather than up to twice * number of search paths.

ahoppen: Ah, I see. Pulled it out.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1956, opened 2025-01-28
Do not schedule a syntactic test rescan if we know that we already have an up-to-date syntactic test index for it

ahoppen:
If the build system is sending us with `buildTarget/didChange`, we would always create processor count * 4 new entries in the `indexingQueue` to rescan those files, if they are up-to-date. This can be quite wasteful if the underlying files didn’t actually change. Do a quick scan up-front to check if the files actually need to be re-scanned up-front.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

PR #1957, opened 2025-01-28
Allow non-file URLs when injecting a build system

ahoppen:
There is no reason why we should only support file URLs here.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #1960, opened 2025-01-29
Add target display name to run and debug code lenses

matthewbastien:
The `SwiftLanguageService` can retrieve the canonical build target and its display name for a given source file. Add this information (if available) to the run and debug code lenses.

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftCodeLensScanner.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftCodeLensScanner.swift
@@ -70,8 +70,10 @@ final class SwiftCodeLensScanner: SyntaxVisitor {
     if attribute.trimmedDescription == "@main" {
       let range = self.snapshot.absolutePositionRange(of: attribute.trimmedRange)
       var targetNameToAppend: String = ""
+      var arguments: [LSPAny] = []
       if let targetName {
         targetNameToAppend = " \(targetName)"
+        arguments.append(.string(targetName))
       }
 
       if let runCommand = supportedCommands[SupportedCodeLensCommand.run] {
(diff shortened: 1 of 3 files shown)


matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

matthewbastien:
@swift-ci please test windows platform

ahoppen (commented):
How does VS Code run a target if this code action is selected? Does it infer the target from the directory structure? Would it make sense to also pass the target to the editor to make sure it runs the target that the user just selected?

matthewbastien:
> How does VS Code run a target if this code action is selected?

It does it based on the active file (essentially performing the same query to Swift PM that I added here). I agree that it would be nicer to just pass the target information as command arguments. Would require a few changes on the VS Code side of things as well, though.

ahoppen:
How about we pass the target back in this PR and then the VS Code extension can start adopting it at its own pace.

matthewbastien:
> How about we pass the target back in this PR and then the VS Code extension can start adopting it at its own pace.

Added the target name to the arguments list for each command.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

ahoppen (approved):
Thanks. LGTM.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows platform

PR #1965, opened 2025-02-04
Split macro tests into multiple test cases

ahoppen:
The `testAttachedMacroExpansion` and `testFreestandingMacroExpansion` tests were  running an a multitude of configurations: They enabled/disabled the reference document and peek document capabilities and ran on multiple locations. Split the test runs up by which capabilities are enabled and log which position marker we are testing to make debugging easier.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1966, opened 2025-02-05
Retry writing file contents in tests on Windows

ahoppen:
Sometimes file writes fail on Windows because another process (like sourcekitd or clangd) still has exclusive access to the file but releases it soon after. Retry to save the file if this happens. This matches what a user would do.

This should fix some test flakiness on Windows.


ahoppen:
@swift-ci Please test

PR #1969, opened 2025-02-05
(chore): Use Native swift Atomics

### LINKED ISSUE #1949: Replace implementation of `AtomicBool` etc. by atomic from the standard library
We should be able to raise the deployment target of SourceKit-LSP to macOS 14.0 and thus use the atomics from the standard library.

mustiikhalil:
The following PR addresses the following:
- Replaces the usage of CAtomics with the native Atomics from the swift library
- Bumps the development platform to macOS 15
- Removes all references to CAtomics from the library since its out of commission

Closes #1949


Sources/SwiftExtensions/Atomics.swift
```diff
@@ -10,97 +10,80 @@
 //
 //===----------------------------------------------------------------------===//
 
-import CAtomics
+import Synchronization
 
-// TODO: Use atomic types from the standard library (https://github.com/swiftlang/sourcekit-lsp/issues/1949)
 package final class AtomicBool: Sendable {
```
ahoppen, line 15: I think we should remove the `AtomicBool` etc. types altogether and just use `Atomic<Bool>` instead. Their purpose was to wrap the C functions, which isn’t needed anymore.

ahoppen (commented):
Thank you @mustiikhalil 🙏🏽

Sources/SwiftExtensions/Atomics.swift
mustiikhalil: I actually tried that but I failed to implement it since Atomic is non-copyable and thus we can't pass it into tasks for example:

``` swift
 let didHitTimeout = Atomic<Bool>(false)

  let stream = AsyncThrowingStream<T?, Error> { continuation in
  // Passing closure as a 'sending' parameter risks causing data races between code in the current task and concurrent execution of the closure
    Task {
      didHitTimeout.store(true, ordering: ...)
    }
  }
```

ahoppen: Could we just store it in a class to get reference semantics in those cases?

```class
final class Wrapped<T: ~Copyable & Sendable>: Sendable {
  let value: T

  init(_ value: consuming T) {
    self.value = value
  }
}
```

in those cases? I would expect that in most cases we can still use `Atomic<Bool>` directly.

mustiikhalil: Should be better now

Sources/SourceKitLSP/SourceKitIndexDelegate.swift
```diff
@@ -24,27 +25,29 @@ actor SourceKitIndexDelegate: IndexDelegate {
...
   nonisolated package func processingAddedPending(_ count: Int) {
-    pendingUnitCount.value += Int32(count)
+    pendingUnitCount.add(Int32(count), ordering: .acquiringAndReleasing)
   }
 
   nonisolated package func processingCompleted(_ count: Int) {
-    pendingUnitCount.value -= Int32(count)
-    if pendingUnitCount.value == 0 {
+    pendingUnitCount.subtract(Int32(count), ordering: .acquiringAndReleasing)
+    if pendingUnitCount.load(ordering: .acquiring) == 0 {
```
ahoppen, line 38: Instead of having a `subtract` and `load`, we should probably use the value that is returned by `subtract` and then use it below as well. (having `load` be a method just surfaces this existing bug)

ahoppen (commented):
Thanks for the update. I have two high-level comments:
- Should we just use `sequentiallyConsistent` as the ordering? As far as I can tell, the current `releasing` and `acquiring` produce the same semantics and I find `sequentiallyConsistent` easier to reason about. Or am I missing something.
- I think some of the `Atomic<Int32>` and `Atomic<UInt32>` can now become `Atomic<Int>` if that simplified their usage. They were only `Int32` and `UInt32` because of C interop.

Sources/SKTestSupport/TestSourceKitLSPClient.swift
```diff
@@ -216,7 +217,7 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
     // It's really unfortunate that there are no async deinits. If we had async
     // deinits, we could await the sending of a ShutdownRequest.
     let sema = DispatchSemaphore(value: 0)
-    server.handle(ShutdownRequest(), id: .number(Int(nextRequestID.fetchAndIncrement()))) { result in
+    server.handle(ShutdownRequest(), id: .number(Int(nextRequestID.add(1, ordering: .sequentiallyConsistent).newValue))) { result in
```
ahoppen, line 220:
```suggestion
    server.handle(ShutdownRequest(), id: .number(Int(nextRequestID.add(1, ordering: .sequentiallyConsistent).oldValue))) { result in
```

Sources/SourceKitLSP/SourceKitIndexDelegate.swift
mustiikhalil: > Should we just use sequentiallyConsistent as the ordering? As far as I can tell, the current releasing and acquiring produce the same semantics and I find sequentiallyConsistent easier to reason about. Or am I missing something.

I had issues with the tests before, but when running the tests now. The issue seems to have resolved itself.

> I think some of the Atomic<Int32> and Atomic<UInt32> can now become Atomic<Int> if that simplified their usage. They were only Int32 and UInt32 because of C interop.

At the start I didn't really wanna add too many changes, so I thought of keeping it as it was

ahoppen (commented):
Very nice. I noticed one small mistake while looking through again, otherwise everything is great!

Sources/SKTestSupport/TestSourceKitLSPClient.swift
ahoppen: Yes, that’s correct. It’s just this case where we want to use the `oldValue` because `nextRequestID` is the next request ID that hasn’t been used yet. We use `nextRequestID` with `oldValue` in other places as well.

ahoppen (approved):

ahoppen:
@swift-ci Please test

mustiikhalil:
The Ci failed because of a newly merged commit that uses the AtomicUint32

ahoppen:
Thanks for the update @mustiikhalil 

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

mustiikhalil:
I'm assuming this will never pass on macOS since the Ci has is OS as 13.6 instead of 15.0

ahoppen:
Oh no, I completely forgot about that 😢. Thank you for the PR anyway. This will be great once we can update the deployment target.

mustiikhalil:
No worries! Always happy to help, I guess we can make it a draft until all of that is done

mustiikhalil:
Should we resume this?

ahoppen:
Yes, that would be greatly appreciated. Thanks for circling back.

mustiikhalil:
@ahoppen sorry for the late start, I just saw that Synchronization is only available on `macOS 15` and the package still supports `macOS 13`

ahoppen:
Ah, yes. We only recently raised the deployment target to 14.0. Let’s wait until we can raise it to 15.0 so we can take this change.

mustiikhalil:
@ahoppen just checking in, has the CI been bumped to macOS 15?

ahoppen:
Unfortunately, we can’t raise the deployment target to macOS 15 just yet.

jakepetroules:
I merged #2636 to update the deployment target and now this is _almost_ unblocked as we still need to update the toolchain itself.

mustiikhalil:
@jakepetroules perfect! Just ping me when everything is ready and I'll work on it

PR #1970, opened 2025-02-06
Use an active compilation condition to build SourceKit-LSP without SwiftPM support

ahoppen:
This is more explicit than using `#if canImport(PackageModel)`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,10 +4,16 @@ import Foundation
 import PackageDescription
 
 /// Swift settings that should be applied to every Swift target.
-let globalSwiftSettings: [SwiftSetting] = [
-  .enableUpcomingFeature("InternalImportsByDefault"),
-  .enableUpcomingFeature("MemberImportVisibility"),
-]
+var globalSwiftSettings: [SwiftSetting] {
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

Package.swift
```diff
@@ -4,10 +4,16 @@ import Foundation
...
-let globalSwiftSettings: [SwiftSetting] = [
-  .enableUpcomingFeature("InternalImportsByDefault"),
-  .enableUpcomingFeature("MemberImportVisibility"),
-]
+var globalSwiftSettings: [SwiftSetting] {
+  var result: [SwiftSetting] = [
+    .enableUpcomingFeature("InternalImportsByDefault"),
+    .enableUpcomingFeature("MemberImportVisibility"),
+  ]
+  if noSwiftPMDependency {
```
bnbarham, line 12: I was going to ask where this was set, but we apparently had it already 😅. The double negative is a bit weird, but I suppose having `SWIFTPM_DEPENDENCY` would also be a bit weird, so maybe the double negative is the lesser of two evils.

bnbarham (approved):

PR #1971, opened 2025-02-06
Update the compiler arguments used for background AST builds

bnbarham:
This fixes two issues:
1. The SwiftPM build system was setup without passing through whether it should prepare or not. This meant that we lost eg. the argument to allow compiler errors when building the AST (even though it was set when building the modules)
2. The compiler argument adjustment to remove harmful and unnecessary flags only applied to indexing arguments, not those passed to the AST builds

Resolves rdar://141508656.


Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -1239,3 +1248,154 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
...
+      .flag("no-color-diagnostics", [.singleDash]),
+      .flag("parseable-output", [.singleDash]),
+      .flag("save-temps", [.singleDash]),
+      .flag("serialize-diagnostics", [.singleDash]),
+      .flag("use-frontend-parseable-output", [.singleDash]),
+      .flag("validate-clang-modules-once", [.singleDash]),
+      .flag("whole-module-optimization", [.singleDash]),
+      .flag("experimental-skip-all-function-bodies", isFrontend: true, [.singleDash]),
+      .flag("experimental-skip-non-inlinable-function-bodies", isFrontend: true, [.singleDash]),
+      .flag("experimental-skip-non-exportable-decls", isFrontend: true, [.singleDash]),
```
bnbarham, line 1276: To save you comparing the two - these are new (sorry, should have done in a separate commit).

bnbarham:
@swift-ci please test

Sources/BuildSystemIntegration/BuildSystemManager.swift
```diff
@@ -1239,3 +1248,154 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
...
+    var iterator = compilerArguments.makeIterator()
+    while let argument = iterator.next() {
+      switch optionsToRemove.firstMatch(for: argument) {
+      case .removeOption:
+        continue
+      case .removeOptionAndNextArgument:
+        _ = iterator.next()
+        continue
+      case .removeOptionAndPreviousArgument:
+        _ = result.popLast()
```
ahoppen, line 1304: Should we validate that the previous argument is indeed `-Xfrontend`? If it is not, should we log a fault and not remove the argument? Probably best to give `removeOptionAndPreviousArgument` an associated value with the expected previous argument in that case.

ahoppen (approved):
Thank you!

Sources/BuildSystemIntegration/BuildSystemManager.swift
bnbarham: I skipped the log, seems noisy and if the flag is there then we should remove it regardless (but have skipped removing the previous).

bnbarham:
@swift-ci please test

Sources/BuildSystemIntegration/CompilerCommandLineOption.swift
```diff
@@ -39,16 +43,28 @@ package struct CompilerCommandLineOption {
   /// The name of the option, without any preceeding `-` or `--`.
   private let name: String
 
+  /// The name of the argument that prefixes this flag (eg. `-Xfrontend`/`-Xclang`).
+  private let frontendName: String?
```
ahoppen, line 47: The documentation doesn’t entirely match the behavior. The implementation expects `Xfrontend` while the documentation mentions `-Xfrontend`.

ahoppen (commented):
One more small comment. Otherwise, I think you need to update CMakeLists.txt since you moved files.

Sources/BuildSystemIntegration/CompilerCommandLineOption.swift
bnbarham: It was meant to be an example of a flag 😅 But I'll change given I forgot to update cmakelists :(

bnbarham:
@swift-ci please test

bnbarham:

bnbarham:
@swift-ci please test

bnbarham:

PR #1979, opened 2025-02-13
Add sendable annotation to `userInfo` closure

bnbarham:
swift-foundation recently landed a change (in
swiftlang/swift-foundation#764) which requires `any Sendable` values in `JSONEncoder.userInfo`. This causes a build failure:
```
JSONRPCConnection.swift:370:50: error: type '(RequestID) -> Optional<any ResponseType.Type>' does not conform to the 'Sendable' protocol
368 |
369 |     // Setup callback for response type.
370 |     decoder.userInfo[.responseTypeCallbackKey] = { (id: RequestID) -> ResponseType.Type? in
    |                                                  |- error: type '(RequestID) -> Optional<any ResponseType.Type>' does not conform to the 'Sendable' protocol
    |                                                  `- note: a function type must be marked '@Sendable' to conform to 'Sendable'
371 |       guard let outstanding = self.outstandingRequests[id] else {
372 |         logger.error("Unknown request for \(id, privacy: .public)")
```

Make the closure sendable, which is safe as we're only reading from `outstandingRequests` (where all our writes are guarded by the same queue that the decoding is on).

### DIFF
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -367,7 +367,9 @@ public final class JSONRPCConnection: Connection {
     decoder.userInfo[.messageRegistryKey] = messageRegistry
 
     // Setup callback for response type.
-    decoder.userInfo[.responseTypeCallbackKey] = { (id: RequestID) -> ResponseType.Type? in
+    decoder.userInfo[.responseTypeCallbackKey] = { @Sendable (id: RequestID) -> ResponseType.Type? in
+      // `outstandingRequests` should never be mutated in this closure. Reading is fine as all of our other writes are
+      // guarded by `queue`, but `JSONDecoder` could (since this is sendable) invoke this concurrently.
       guard let outstanding = self.outstandingRequests[id] else {
         logger.error("Unknown request for \(id, privacy: .public)")
         return nil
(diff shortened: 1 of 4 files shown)


bnbarham:

bnbarham:
(note: completely untested, just running Windows since Linux will fail with docc first)

ahoppen (approved):
Marking the closure as Sendable implies that JSONDecoder should be able to invoke it multiple times concurrently when deserializing the JSON at which point we would get concurrent accesses to `outstandingRequests`. I think that should be fine because they are all just reading, which should be fine (StackOverflow seems to imply so https://stackoverflow.com/questions/45666181/is-it-safe-to-access-not-modify-a-swift-dictionary-from-multiple-threads) because the lock on `queue` guarantees that there are no concurrent writes. 

I think it would be good to include that rationale in a comment so that we don't accidentally start modifying `outstandingRequests` in that closure in the future.

bnbarham:
https://github.com/swiftlang/swift/pull/79380

@swift-ci please test

xedin (approved):

ahoppen:
https://github.com/swiftlang/swift/pull/79380

@swift-ci please test Windows

bnbarham:

bnbarham:
@swift-ci please test

bnbarham:

bnbarham:

bnbarham:

PR #1984, opened 2025-02-17
Hide internal options from SourceKit-LSP configuration file documentation

ahoppen:
These options are not intended to be used by users.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1993, opened 2025-02-22
Add `CancelRequestNotification` to BSP notification types

ahoppen:
Otherwise a client that uses the `BuildServerProtocol` modules to implement a BSP server can’t deserialize the `$/cancelRequest` notification.


ahoppen:
@swift-ci Please test

PR #1997, opened 2025-02-25
Set `useNewAPI` when closing a code completion session

ahoppen:
Otherwise, the close does not get handled by the SourceKit plugin. In practice, it doesn’t matter too much because the SourceKit plugin will implicitly close the last completion session when a new one is opened but we should fix this.


ahoppen:
@swift-ci Please test

hamishknight (approved):

bnbarham (approved):

PR #2000, opened 2025-02-25
If a build server doesn’t specify `-index-store-path` in the SourceKit options, add it during background indexing

ahoppen:
Also, remove options to override the index store path and index database path. We should always infer these from the build system and I can’t think of a reason to override these settings.

### DIFF
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -299,6 +299,33 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
     await indexStoreUpToDateTracker.markUpToDate([file.sourceFile], updateOperationStartDate: startDate)
   }
 
+  /// If `args` does not contain an `-index-store-path` argument, add it, pointing to the build system's index store
+  /// path. If an `-index-store-path` already exists, validate that it matches the build system's index store path and
+  /// replace it by the build system's index store path if they don't match.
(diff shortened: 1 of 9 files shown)


Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -320,6 +320,25 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
         // Fake an output path so that we get a different unit file for every Swift file we background index
         "-index-unit-output-path", uri.pseudoPath + ".o",
       ]
+    guard let buildSystemIndexStorePath = await self.buildSystemManager.initializationData?.indexStorePath else {
+      struct NoIndexStorePathError: Error {}
+      throw NoIndexStorePathError()
+    }
+    if let indexStorePath = args.firstIndex(of: "-index-store-path"), indexStorePath + 1 < args.count {
```
bnbarham, line 327: `lastIndex` in case there's multiple? `lastIndexStorePathArgument` could also use `lastIndex` rather than the reverse loop it has now.

ahoppen: Good idea.

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2001, opened 2025-02-25
Add calls with no defaulted arguments to code completion results

ahoppen:
The completion items that don’t have any of the defaulted arguments disappeared when we switched to the SourceKit plugin.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2003, opened 2025-02-25
Don’t poll the index for unit changes on every `filesDidChange` call

ahoppen:
`pollForUnitChangesAndWait` is a costly operation since it iterates through all the unit files on the file system. We should only run it during initial indexing (where we might need to build the initial indexstore-db) but not for any subsequent indexing calls.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -54,7 +54,7 @@ private struct OpaqueQueuedIndexTask: Equatable {
 
 private enum InProgressIndexStore {
   /// We are waiting for preparation of the file's target to be scheduled. The next step is that we wait for
-  /// prepration to finish before we can update the index store for this file.
+  /// preparation to finish before we can update the index store for this file.
   ///
   /// `preparationTaskID` identifies the preparation task so that we can transition a file's index state to
   /// `updatingIndexStore` when its preparation task has finished.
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -273,18 +273,25 @@ package final actor SemanticIndexManager {
...
     let taskId = UUID()
     let generateBuildGraphTask = Task(priority: .low) {
       await withLoggingSubsystemAndScope(subsystem: indexLoggingSubsystem, scope: "build-graph-generation") {
         await hooks.buildGraphGenerationDidStart?()
         await self.buildSystemManager.waitForUpToDateBuildGraph()
-        // Ensure that we have an up-to-date indexstore-db. Waiting for the indexstore-db to be updated is cheaper than
-        // potentially not knowing about unit files, which causes the corresponding source files to be re-indexed.
-        index.pollForUnitChangesAndWait()
+        if ensureAllUnitsRegisteredInIndex {
+          index.pollForUnitChangesAndWait()
```
bnbarham, line 293: Should we add some additional documentation to `pollForUnitChangesAndWait` to describe what it actually does (and the fact it's costly) rather than just "for testing"?

bnbarham (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Yeah, I’ll add some to indexstore-db

PR #2013, opened 2025-02-27
When the build system sends an update for build targets, update the index for the files in those targets

ahoppen:
For each file in the changed targets, we still check whether it has an up-to-date unit based on timestamps. The important thing for this change is that we start indexing files for which we only receive build settings after an update from the build server.

While doing this, I found a couple ways in which we could clean up/improve `SemanticIndexManager`. These changes should be reviewable commit-by-commit.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -52,42 +52,53 @@ private struct OpaqueQueuedIndexTask: Equatable {
   }
 }
 
-private enum InProgressIndexStore {
-  /// We know that we need to index the file but and are currently gathering all information to create the `indexTask`
-  /// that will index it.
-  ///
-  /// This is needed to avoid the following race: We request indexing of file A. Getting the canonical target for A
-  /// takes a bit and before that finishes, we request another index of A. In this case, we don't want to kick off
-  /// two tasks to update the index store.
-  case creatingIndexTask
+private struct InProgressIndexStore {
+  enum State {
+    /// We know that we need to index the file but and are currently gathering all information to create the `indexTask`
+    /// that will index it.
+    ///
+    /// This is needed to avoid the following race: We request indexing of file A. Getting the canonical target for A
+    /// takes a bit and before that finishes, we request another index of A. In this case, we don't want to kick off
+    /// two tasks to update the index store.
+    case creatingIndexTask
 
-  /// We are waiting for preparation of the file's target to be scheduled. The next step is that we wait for
-  /// preparation to finish before we can update the index store for this file.
-  ///
-  /// `preparationTaskID` identifies the preparation task so that we can transition a file's index state to
-  /// `updatingIndexStore` when its preparation task has finished.
-  ///
-  /// `indexTask` is a task that finishes after both preparation and index store update are done. Whoever owns the index
-  /// task is still the sole owner of it and responsible for its cancellation.
-  case waitingForPreparation(preparationTaskID: UUID, indexTask: Task<Void, Never>)
+    /// We are waiting for preparation of the file's target to be scheduled. The next step is that we wait for
+    /// preparation to finish before we can update the index store for this file.
+    ///
+    /// `preparationTaskID` identifies the preparation task so that we can transition a file's index state to
+    /// `updatingIndexStore` when its preparation task has finished.
+    ///
+    /// `indexTask` is a task that finishes after both preparation and index store update are done. Whoever owns the index
+    /// task is still the sole owner of it and responsible for its cancellation.
+    case waitingForPreparation(preparationTaskID: UUID, indexTask: Task<Void, Never>)
 
-  /// We have started preparing this file and are waiting for preparation to finish before we can update the index
-  /// store for this file.
-  ///
-  /// `preparationTaskID` identifies the preparation task so that we can transition a file's index state to
-  /// `updatingIndexStore` when its preparation task has finished.
-  ///
-  /// `indexTask` is a task that finishes after both preparation and index store update are done. Whoever owns the index
-  /// task is still the sole owner of it and responsible for its cancellation.
-  case preparing(preparationTaskID: UUID, indexTask: Task<Void, Never>)
+    /// We have started preparing this file and are waiting for preparation to finish before we can update the index
+    /// store for this file.
+    ///
+    /// `preparationTaskID` identifies the preparation task so that we can transition a file's index state to
+    /// `updatingIndexStore` when its preparation task has finished.
+    ///
+    /// `indexTask` is a task that finishes after both preparation and index store update are done. Whoever owns the index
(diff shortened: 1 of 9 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -454,15 +466,7 @@ package final actor SemanticIndexManager {
...
-        return .headerFile(header: uri, mainFile: mainFile)
-      }
-      .filter {
-        switch inProgressIndexTasks[$0] {
-        case .waitingForPreparation, .creatingIndexTask:
-          return false
-        default:
-          return true
-        }
+        return (.headerFile(header: uri, mainFile: mainFile), modifiedFilesIndex.modificationDate(of: uri))
```
hamishknight, line 469: If the main file has changed, should that also cause a re-index?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -683,34 +687,62 @@ package final actor SemanticIndexManager {
...
+          state: .creatingIndexTask,
+          fileModificationDate: file.fileModificationDate
+        )
+      } else {
+        alreadyScheduledTasks.insert(file.file)
+
       }
-      inProgressIndexTasks[file] = .creatingIndexTask
     }
     indexTasksWereScheduled(newIndexTasks)
```
hamishknight, line 731: Worth avoiding calling this if `newIndexTasks` is 0?

hamishknight (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
hamishknight: Hmm seems like we don't currently consider that out of date, so I guess the answer is no

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -435,8 +435,24 @@ package final class Workspace: Sendable, BuildSystemManagerDelegate {
...
   package func buildTargetsChanged(_ changes: [BuildTargetEvent]?) async {
     await sourceKitLSPServer?.fileHandlingCapabilityChanged()
-    let testFiles = await orLog("Getting test files") { try await buildSystemManager.testFiles() } ?? []
-    await syntacticTestIndex.listOfTestFilesDidChange(testFiles)
+    await orLog("Scheduling re-indexing of changed targets") {
+      var sourceFiles = try await self.buildSystemManager.sourceFiles(includeNonBuildableFiles: false)
+      if let changes {
+        let changedTargets = changes.map(\.target)
+        sourceFiles = sourceFiles.filter {
+          !$0.value.targets.intersection(changedTargets).isEmpty
```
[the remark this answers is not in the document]
ahoppen: Nice. I didn’t know about `isDisjoint`

ahoppen:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -421,33 +397,60 @@ package final actor SemanticIndexManager {
...
+        // index.
+        // Deterministically pick a main file. This ensures that we always pick the same main file for a header. This way,
+        // if we request the same header to be indexed twice, we'll pick the same unit file the second time around,
+        // realize that its timestamp is later than the modification date of the header and we don't need to re-index.
+        let mainFile =
+          deletedFilesIndex
+          .mainFilesContainingFile(uri: uri, crossLanguage: false)
+          .sorted(by: { $0.stringValue < $1.stringValue }).first
+        guard let mainFile else {
+          return nil
```
bnbarham, line 464: Could we log in this case (ie. we have a file that isn't known to the project)? Or did you do that and it is that too spammy?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -421,33 +397,60 @@ package final actor SemanticIndexManager {
...
+  ///
+  /// The returned results will not include files that are known to be up-to-date based on `indexStoreUpToDateTracker`
+  /// or the unit timestamp will not be re-indexed unless `indexFilesWithUpToDateUnit` is `true`.
   private func filesToIndex(
-    toCover files: some Collection<DocumentURI> & Sendable
+    toCover files: some Collection<DocumentURI> & Sendable,
+    indexFilesWithUpToDateUnits: Bool
   ) async -> [FileToIndex] {
     let sourceFiles = await orLog("Getting source files in project") {
       Set(try await buildSystemManager.sourceFiles(includeNonBuildableFiles: false).keys)
```
bnbarham, line 429: I know this was here before, but this could be fairly large depending on the project - probably not great to reconstruct the set every time this is called.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -683,34 +687,64 @@ package final actor SemanticIndexManager {
...
+      switch inProgress?.state {
+      case nil:
         newIndexTasks += 1
+        shouldScheduleIndexing = true
+      case .creatingIndexTask, .waitingForPreparation:
+        // We already have a task that indexes the file but hasn't started preparation yet. Indexing the file again
+        // won't produce any new results.
+        alreadyScheduledTasks.insert(file.file)
+        shouldScheduleIndexing = false
+      case .preparing(_, _), .updatingIndexStore(_, _):
```
bnbarham, line 712: If it's currently preparing, we don't need to index it again do we? An update to the file itself shouldn't cause its own preparation to change. One possible exception to that could be a header file though.

Sources/SemanticIndex/SemanticIndexManager.swift
bnbarham: Is that true? IIRC we actually do index the main file, not the header. Whether we should or not is another question, but that's how it works today.

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: We could have triggered re-indexing because build settings (or built graph in general) have changed, in which case we might also need to re-prepare, eg. to pick up new dependencies.

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Good idea

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Good idea. Caching the set now.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2019, opened 2025-02-28
Small improvements to `workspace/_sourceKitOptions` request

ahoppen:
- Make response of `workspace/_sourceKitOptions` request non-optional
  - The response was not specified as being optional and shouldn’t be.
- Mark `workspace/_sourceKitOptions` as freestanding
  - The request doesn’t depend on any document state, so it should be freestanding instead of `workspaceRequest`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2020, opened 2025-02-28
Debounce `filesBuildSettingsChanged` calls

ahoppen:
This helps in the following situation: A build system takes 5s to return build settings for a file and we have 10 requests for those build settings coming in that time period. Once we get build settings, we get 10 calls to `resultReceivedAfterTimeout` in `buildSettings(for:in:language:fallbackAfterTimeout:)`, all for the same document. But calling `fileBuildSettingsChanged` once is totally sufficient.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2022, opened 2025-02-28
Add structure to the index logs

ahoppen:
When the client supports it, communicate the structure of tasks that were stared during background indexing or by the build server to the client. If there are multiple operations happening in parallel, this allows the client to display them in separate log tracks instead of interspersing them with the emoji prefixes like we do today.


Contributor Documentation/LSP Extensions.md
```diff
@@ -490,6 +490,60 @@ Added field:
...
+ * Indicates the beginning of a new task that may receive updates with `StructuredLogReport` or `StructuredLogEnd`
+ * payloads.
+ */
+export interface StructuredLogBegin {
+  kind: 'begin'
+
+  /**
+   * A succinct title that can be used to describe the task that started this structured.
+   */
+  title: string;
```
bnbarham, line 517: If we have title it feels a little odd to have a log message at all (though I realize that's currently being used in preparation). I think I'd prefer either having title and then mentioning that the message is ignored completely, or only having message and no title. For end, might make sense to add duration in there specifically (and also state that the message is ignored - the "finished" we have today could be its own log message)? Then completely ignore both in the unstructured case.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2025, opened 2025-03-01
Mark target preparation out-of-date when it has changed in the build server

ahoppen:
Depends on https://github.com/swiftlang/sourcekit-lsp/pull/2013.

---

When the target changes in the build server, it’s build settings might have changed, which  means that we can no longer assume that it’s up-to-date.


bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #2037, opened 2025-03-05
Fix nondeterministic failure in `testSyntacticOrIndexBasedXCTestsBasedOnWhetherFileIsIndexed`

ahoppen:
If we were indexing the modified file too fast, we would get semantic tests to the `workspace/tests` request while we were expecting syntactic results.

Add semaphores to ensure that indexing doesn’t finish until we have received the syntactic result.

rdar://146331921


ahoppen:
@swift-ci Please test

PR #2039, opened 2025-03-06
Poll index in `testBuildSystemUsesStandardizedFileUrlsInsteadOfRealpath`

ahoppen:
We need to make sure that all unit files are read into indexstore-db so that we know about the unit file that allows us to map from header to main file.

Fixes rdar://146392120


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2043, opened 2025-03-07
Add `-` to `swift-format` invocation to indicate that file contents should be read from stdin

ahoppen:
To match https://github.com/swiftlang/swift-format/pull/914


ahoppen:
@swift-ci Please test

hamishknight (approved):

Issue #2040, opened 2025-03-07
Provide public access to InProcessClient


Mordil:
### Description

I'm building tools on top of the LSP project, and rather than having to go round trip with the full LSP interface, I'd like to leverage the `InProcessSourceKitLSPClient` to issue commands directly without having to build my own client project.

In 6.0, I was able to copy what was in this repo, but in 6.1 I'm not able to do that because much of this package switched to `package` accessors rather than `public`.

ahoppen:
Synced to Apple’s issue tracker as rdar://146447619

ahoppen:
`InProcessSourceKitLSPClient` is a public type and it is the intended entry point to start an in-process SourceKit-LSP server. What are you missing from `InProcessSourceKitLSPClient`? Would you like to open a PR with the entry points that you’d like to see be public and then we can discuss whether they should be public on the PR. I think that’s easier than talking about abstract notions here.

Mordil:
It's not available to package consumers, though - right? Because the `InProcessClient` module is not publicly linkable - it's not vended as a product nor made available `_exported import`

I am able to import it using the quirk of transitive dependencies in SwiftPM, but I've seen it mentioned a few times is considered bad practice

Mordil:
I opened two PRs vending a new library product, one for the `main` branch and the other for a hopefully back ported version to ship with the upcoming 6.1 release

ahoppen:
Closing because this was implanted by https://github.com/swiftlang/sourcekit-lsp/pull/2045.

PR #2045, opened 2025-03-07
Vend InProcessClient as a package library product

Mordil:
This allows people writing tools on top of SourceKit-LSP to be able to directly use the `InProcessSourceKitLSPClient` to directly send requests within their applications.

This addresses #2040

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,6 +19,7 @@ var products: [Product] = [
   .executable(name: "sourcekit-lsp", targets: ["sourcekit-lsp"]),
   .library(name: "_SourceKitLSP", targets: ["SourceKitLSP"]),
   .library(name: "LSPBindings", targets: ["LanguageServerProtocol", "LanguageServerProtocolJSONRPC"]),
+  .library(name: "InProcessClient", targets: ["InProcessClient"]),
   .library(name: "SwiftSourceKitPlugin", type: .dynamic, targets: ["SwiftSourceKitPlugin"]),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -19,6 +19,7 @@ var products: [Product] = [
   .executable(name: "sourcekit-lsp", targets: ["sourcekit-lsp"]),
   .library(name: "_SourceKitLSP", targets: ["SourceKitLSP"]),
   .library(name: "LSPBindings", targets: ["LanguageServerProtocol", "LanguageServerProtocolJSONRPC"]),
+  .library(name: "InProcessLSPClient", targets: ["InProcessClient"]),
```
ahoppen, line 22: Should we name the product the same as the target?

```suggestion
  .library(name: "InProcessClient", targets: ["InProcessClient"]),
```

ahoppen (commented):
Thank you. One small comment, otherwise LGTM.

Package.swift
Mordil: I'm 50-50 on that, and it's up to you (and others) who maintain the project to ultimately tell me what makes more sense.

I went with `InProcessLSPClient` to closer match the rest of the product names and be more descriptive to package consumers what is in that product

ahoppen: Let’s go with `InProcessClient` to keep it consistent with the target name. Since you’re depending on the sourcekit-lsp package, the LSP is already implied.

Mordil: Updated!

ahoppen (approved):
Thank you for opening the PR @Mordil 🙏🏽

ahoppen:
@swift-ci Please test

PR #2050, opened 2025-03-10
Log sourcekitd requests inside `withCancellableCheckedThrowingContinuation`

ahoppen:
Otherwise, we would log sending the sourcekitd request even if the task was already cancelled and we thus took an early exit in `withCancellableCheckedThrowingContinuation`.

### DIFF
--- a/Sources/SourceKitD/SourceKitD.swift
+++ b/Sources/SourceKitD/SourceKitD.swift
@@ -139,10 +139,9 @@ extension SourceKitD {
     timeout: Duration,
     fileContents: String?
   ) async throws -> SKDResponseDictionary {
-    log(request: request)
-
     let sourcekitdResponse = try await withTimeout(timeout) {
       return try await withCancellableCheckedThrowingContinuation { (continuation) -> SourceKitDRequestHandle? in
+        self.log(request: request)
         var handle: sourcekitd_api_request_handle_t? = nil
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitD/SourceKitD.swift
```diff
@@ -139,6 +139,8 @@ extension SourceKitD {
     timeout: Duration,
     fileContents: String?
   ) async throws -> SKDResponseDictionary {
+    try Task.checkCancellation()
```
bnbarham, line 142: Worth checking this in `withCancellableCheckedThrowingContinuation` instead?

ahoppen: Oh, we already checked there. Guess we just needed to move the log statement into `withCancellableCheckedThrowingContinuation`.

ahoppen:
@swift-ci Please test

Sources/SourceKitD/SourceKitD.swift
bnbarham: Ah, we do too. Thanks 🙇‍♂️

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #2054, opened 2025-03-11
Handle file did change notifications as freestanding messages

ahoppen:
Technically, the watched files notification can change the response of any other request (eg. because a target needs to be re-prepared). But treating it as a `globalConfiguration` inserts a lot of barriers in request  handling and significantly prevents parallelism. Since many editors batch file change notifications already, they might have delayed the file change notification even more, which is equivalent to handling the  notification a little later inside SourceKit-LSP. Thus, treating it as `freestanding` should be acceptable.

Also, simplify the logic needed in tests to write modified files to disk because we now need to run a barrier request in tests to ensure that the watched file notification has been handled.

### DIFF
--- a/Sources/SKTestSupport/MultiFileTestProject.swift
+++ b/Sources/SKTestSupport/MultiFileTestProject.swift
@@ -12,6 +12,7 @@
 
 package import Foundation
 package import LanguageServerProtocol
+import SKLogging
 package import SKOptions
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2055, opened 2025-03-11
Don’t show “Scheduling tasks” indexing status when we are currently performing indexing

ahoppen:
Switching from eg. `Indexing 76 / 148` to `Scheduling tasks` produces noise and provides little value to the user – on the contrary, it hides the indexing progress. Thus, prefer showing `Indexing x / y` over `Scheduling tasks`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2061, opened 2025-03-13
Do not apply file path normalization in `mainFiles(containing:)` to the file itself

ahoppen:
Do not apply the standardized file normalization to the source file itself. Otherwise we would get the following behavior:
 - We have a build system that uses standardized file paths and index a file as /tmp/test.c
 - We are asking for the main files of /private/tmp/test.c
 - Since indexstore-db uses realpath for everything, we find the unit for /tmp/test.c as a unit containing /private/tmp/test.c, which has /private/tmp/test.c as the main file.
 - If we applied the path normalization, we would normalize /private/tmp/test.c to /tmp/test.c, thus reporting that /tmp/test.c is a main file containing /private/tmp/test.c, But that doesn't make sense (it would, in fact cause us to treat /private/tmp/test.c as a header file that we should index using /tmp/test.c as a main file.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2062, opened 2025-03-13
Skip formatting tests if the `swift-format` executable in the host toolchain doesn’t support `-` to indicate that it’s reading the source file from stdin

ahoppen:
Ie if SourceKit-LSP is being tested with a Swift 6.0 toolchain


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2066, opened 2025-03-14
Fix expected output in swift-format `SkipUnless` check

ahoppen:
We weren’t hitting this in Swift CI because we only test against Swift 6.2 toolchains, which take an earlier exit based on the Swift version.

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -372,8 +372,8 @@ package actor SkipUnless {
       let output = try result.utf8Output()
       switch output {
       case "": return false
-      case "let x = 1": return true
-      default: throw GenericError("Received unexpected formatting output: \(output)")
+      case "let x = 1\n": return true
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -372,8 +372,8 @@ package actor SkipUnless {
       let output = try result.utf8Output()
       switch output {
       case "": return false
-      case "let x = 1": return true
-      default: throw GenericError("Received unexpected formatting output: \(output)")
+      case "let x = 1\n": return true
```
bnbarham, line 375: Could just strip `output`?

bnbarham (approved):

Sources/SKTestSupport/SkipUnless.swift
ahoppen: 🤷🏽 This check will disappear pretty quickly anyway once we no longer support building and testing SourceKit-LSP using a Swift 6.0 compiler.

PR #2070, opened 2025-03-18
Change command to build SourceKit-LSP on Windows to use PowerShell instead of Command Prompt

ahoppen:


ahoppen:
@swift-ci Please test

PR #2073, opened 2025-03-18
If the compiler arguments only contain `-o`, add the output path as `-index-unit-output-path` to the adjusted options

ahoppen:
Otherwise, we would strip away the `-o`, leaving the command line without any output path option and thus not able to generate a unit file (which requires an output path).


ahoppen:
@swift-ci Please test

bnbarham (approved):
Nice catch.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2075, opened 2025-03-19
Introduce a timeout to test execution

ahoppen:
We have seen a few infinite spins in SourceKit-LSP testing recently. We should fix those spins but as a last stop-gap measure, introduce a timeout to all subprocess invocations, which will at least give us a failure instead of infinite waiting.


ahoppen:
@swift-ci Please test

PR #2076, opened 2025-03-19
SKTestSupport: remove hardcoded version for XCTest

compnerd:
The XCTest and Testing version information is encoded into the `PlatformInfo.plist` that is at the root of the platform. Use this to determine the path for XCTest. This allows us to migrate the XCTest location into an appropriate versioned directory.

### DIFF
--- a/Sources/SKTestSupport/IndexedSingleSwiftFileTestProject.swift
+++ b/Sources/SKTestSupport/IndexedSingleSwiftFileTestProject.swift
@@ -20,6 +20,41 @@ import SwiftExtensions
 import TSCBasic
 import ToolchainRegistry
 
+package struct WindowsPlatformInfo {
+  package struct DefaultProperties {
+    /// XCTEST_VERSION
+    /// specifies the version string of the bundled XCTest.
+    public let xctestVersion: String
+
+    /// SWIFT_TESTING_VERSION
+    /// specifies the version string of the bundled swift-testing.
+    public let swiftTestingVersion: String?
+
+    /// SWIFTC_FLAGS
+    /// Specifies extra flags to pass to swiftc from Swift Package Manager.
+    public let extraSwiftCFlags: [String]?
+  }
+
+  public let defaults: DefaultProperties
+}
+extension WindowsPlatformInfo.DefaultProperties: Decodable {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/SKTestSupport/IndexedSingleSwiftFileTestProject.swift
```diff
@@ -19,6 +19,8 @@ import SourceKitLSP
 import SwiftExtensions
 import TSCBasic
 import ToolchainRegistry
+import Basics
+import PackageModel
```
ahoppen, line 23: I would prefer not to add a dependency on SwiftPM here. We do have compilation conditions to build SourceKit-LSP without SwiftPM and they would need to be passed through here as well. Could we instead copy the relevant parts of `WindowsPlatformInfo`?

compnerd: Sure

compnerd:
@swift-ci please test

Sources/SKTestSupport/IndexedSingleSwiftFileTestProject.swift
```diff
@@ -20,6 +20,49 @@ import SwiftExtensions
...
+}
+
+extension WindowsPlatformInfo: Decodable {
+    enum CodingKeys: String, CodingKey {
+    case defaults = "DefaultProperties"
+    }
+}
+
+extension WindowsPlatformInfo {
+    package init?(reading path: URL) {
```
ahoppen, line 56: We could just throw here. That probably yields better error messages than silently not adding the XCTest search path.

ahoppen (commented):
One small comment. Also, could you run swift-format on the change? `swift format -ipr .`

Sources/SKTestSupport/IndexedSingleSwiftFileTestProject.swift
compnerd: Sure, I think that should be safe enough. Done with the formatting.

ahoppen (approved):

compnerd:
@swift-ci please test

compnerd:

PR #2079, opened 2025-03-20
Move `targets(dependingOn:)` call out of the changed files loop

ahoppen:
We can call this function once we have gathered all changed targets instead of once for each changed file, which should be slightly more performant.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2091, opened 2025-03-26
Add delay before writing file to disk in tests

ahoppen:
Depending on the system, mtime resolution might not be perfectly accurate. Particularly containers appear to have imprecise mtimes. Wait a short time period before writing the new file to avoid situations like the following:
 - We index a source file and the unit receives a time stamp and wait for indexing to finish
 - We modify the source file but so quickly after the unit has been modified that the updated source file receives the same mtime as the unit file
 - We now assume that the we have an up-to-date index for this source file even though we do not.

Waiting 10ms appears to be enough to avoid this situation on the systems we care about.

rdar://147811044


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2092, opened 2025-03-26
Make `workspace/synchronize` a non-experimental request

ahoppen:
This request is generally useful, not only for tests within SourceKit-LSP but also:
- In editor tests that want to test the integration with SourceKit-LSP
- In code analysis tools that want to gather project information using SourceKit-LSP and need an up-to-date index for that.

Remove the experimental feature guard from `workspace/_synchronize`, consequently rename it to `workspace/synchronize` and only guard the `buildServerUpdates` option on the synchronize request by an experimental feature because its long-term usefulness is still not fully understood yet.

### DIFF
--- a/Contributor Documentation/LSP Extensions.md
+++ b/Contributor Documentation/LSP Extensions.md
@@ -664,33 +664,6 @@ export interface SourceKitOptionsResult {
 }
 ```
 
-## `workspace/_synchronize`
-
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2093, opened 2025-03-26
Remove the split between `SourceKitD` and `DynamicallyLoadedSourceKitD`

ahoppen:
There is only one real class that implements the `SourceKitD` protocol, so there really isn’t any need for the protocol + class split at all. Unify them to make code simpler to reason about.

Removing the split caused an `Incorrect actor executor assumption` crash inside `SourceKitD.run(requestYaml:)`, which I fixed by marking all blocks in Csourcekitd as `@Sendable`.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test Windows

PR #2096, opened 2025-03-27
Ensure the `DLHandle` passed to `SourceKitD.init(dlhandle:path:pluginPaths:initialize:)` gets closed if the initializer fails

ahoppen:
Just something I noticed while looking through code. I am not aware of any issues caused by this.

Also relax the `precondition` in `DLHandle.deinit` to a fault. If we do not close a `DLHandle`, that’s a bug but it’s not so bad that we should crash sourcekit-lsp for it.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2097, opened 2025-03-27
Fix a race condition leading to out-of-order notifications in tests

ahoppen:
`TestSourceKitLSPClient.handle` created a new `Task`. This means that we could swap the order of notifications received from SourceKit-LSP. In most cases this doesn’t matter but `BackgroundIndexingTests.testProduceIndexLogWithTaskID` checks that the notification to start a structured log is received before the first report, which might not be the case because of the `Task` here.

Change `PendingNotifications` to a class with a `ThreadSafeBox` to remove the need for a `Task`.

rdar://147814254


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2098, opened 2025-03-27
Don’t crash SourceKit-LSP if swift-format crashes while we write the source file to stdin

ahoppen:
When swift-format crashes before we send all data to its stdin, we get a SIGPIPE when trying to write more data to its studio on Linux. This, in turn, takes SourceKit-LSP down.

Do the same trick that we do when launching `JSONRPCConnection` of disabling SIGPIPE globally if we can’t disable it for individual pipes.

rdar://147665695


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-tools-support-core/pull/502

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-tools-support-core/pull/502

@swift-ci Please test Windows

PR #2104, opened 2025-04-08
Fix a race condition that can cause `testDontReturnEmptyDiagnosticsIfDiagnosticRequestIsCancelled` to fail

ahoppen:
`testDontReturnEmptyDiagnosticsIfDiagnosticRequestIsCancelled` assumed that cancelling the request cancels the target preparation but that is not true. If we only handle the request cancellation after we started preparing the target, the test would time out.

The only reason why we had the target preparation wait in here was to ensure that we cannot finish the request handling before the request cancellation gets handled. We can achieve a similar result by using an example that is hard to typecheck.

rdar://148654514


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2105, opened 2025-04-08
Fix issue that caused request cancellation after a timeout in tests to not work

ahoppen:
We were calling `server.handle(notification:)`, which called directly into `SourceKitLSPServer` but request handling is implemented in `QueueBasedMessageHandler.handle(_:)`. Call that to ensure that we actually cancel requests after the test timeout.


ahoppen:
@swift-ci Please test

hamishknight (approved):

ahoppen:
@swift-ci Please test macOS

PR #2107, opened 2025-04-09
Remove test for default Swift SDK selection for WASI

kateinoigakukun:
After https://github.com/swiftlang/swift-package-manager/pull/8468, we no longer support wasi-sysroot embedded in the Swift toolchain, so the test no longer makes sense.


MaxDesiatov:
@swift-ci test

ahoppen (approved):

kateinoigakukun:
Thanks @ahoppen :pray:

Issue #1286, opened 2024-05-13
Log to syslog on Linux


ahoppen:
On Linux, we can log to syslog like we log to os_log on Darwin. That should allow us to also recapture logs in `sourcekit-lsp diagnose`.

rdar://127138318

PR #2116, opened 2025-04-18
Overview.md: Correct path to sourcekit-lsp logs

Wilfred:
This text was originally added in d8067fb4390e4569fa4941c51bf325179c4c359b, but as of 7d3b44079ac6cc89e065ddb8b21190d15f0e354d the log directory is now ~/.sourcekit-lsp.

### DIFF
--- a/Contributor Documentation/Overview.md
+++ b/Contributor Documentation/Overview.md
@@ -31,4 +31,4 @@ SourceKit-LSP has fairly extensive logging to help diagnose issues. The way logg
 
 On macOS, SourceKit-LSP logs to the system log. [CONTRIBUTING.md](../CONTRIBUTING.md#logging) contains some information about how to read the system logs. Since [OSLog](https://developer.apple.com/documentation/os/logging) cannot be wrapped, the decision to log to macOS’s system log is done at build time and cannot be modified at runtime.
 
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Good catch and thank you!

ahoppen:
@swift-ci Please test

Wilfred:
Incidentally, I also noticed that https://github.com/swiftlang/sourcekit-lsp/blob/main/Contributor%20Documentation/Environment%20Variables.md states that logs on Linux go to stderr, but it looks like that's no longer true as of #1286. It looks like logs go exclusively to ~/.sourcekit-lsp now.

@ahoppen how do you feel about logging to both stderr as well as a log file? Most LSP clients (e.g. VS Code, Emacs) make it easy for you to see stderr output as LSP servers usually write their logs there.

ahoppen:
> @ahoppen how do you feel about logging to both stderr as well as a log file? Most LSP clients (e.g. VS Code, Emacs) make it easy for you to see stderr output as LSP servers usually write their logs there.

I think that’s reasonable. Would you like to create a PR to change it?

ahoppen:
@swift-ci Please test Linux

PR #2123, opened 2025-04-28
Fix logging docs for non-Apple platforms

MahdiBM:
reference: https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/sourcekit-lsp/SourceKitLSP.swift#L255


ahoppen (approved):
Thanks

ahoppen:
@swift-ci Please test

PR #2142, opened 2025-05-08
Add SwiftASN1 as an explicit dependency on BuildSystemIntegration

dschaefer2:
The SwiftPM smoke test still fails to build. I see the SwiftASN1 modules aren't getting added to the include path. Hopefully an explicit dependency will fix that.


dschaefer2:
@swift-ci please test

dschaefer2:
I did get past the errors I was seeing so this works. I have another change up for build.ps1 to get the tests to build.

https://github.com/swiftlang/swift/pull/81391

ahoppen (approved):

PR #2144, opened 2025-05-09
Send `workspace/semanticTokens/refresh` to client when build settings have changed

### LINKED ISSUE #2141: InitializeResult is sent before Workspace is ready
### Swift version

swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)

### Platform

arm64-apple-macosx15.0

### Editor

_No response_

### Description

When opening a SwiftPM workspace (resolved and built, so it is ready to work), and when the sequence is

1. Receive `InitializeResult`
2. `textDocument/didOpen`
3. `textDocument/didChange` for document A
4. `textDocument/semanticTokens/full` for document A

the fourth request is going to fail (or partially return missing informations). The workaround is to add an artificial sleep, (like 1 second) before didOpen, but after receiving `InitializeResult`

1. Receive `InitializeResult`
2. **sleep 1s**
1. `textDocument/didOpen`
2. `textDocument/didChange` for document A
4. `textDocument/semanticTokens/full` for document A

As if something needs to happen between these two, but there is no notification to inform that the server has updated (is it?). It seems like the `InitializeResult` is sent before Workspace is fully setup.

> Until the server has responded to the initialize request with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult), the client must not send any additional requests or notifications to the server

I noticed that is the case for SwiftPM package that has a dependency (again, already fetched, built, and resolved)

ahoppen:
Similar to how we send a `workspace/diagnostic/refresh` request to the client when we get new build settings, we need to send a `workspace/semanticTokens/refresh` to the client to reload semantic tokens when builds settings change. This is particularly important so that we reload semantic tokens from real build setting after a SwiftPM package has finished loading (we previously used fallback settings).

Fixes #2141
rdar://150934682

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftLanguageService.swift
@@ -183,14 +183,14 @@ package actor SwiftLanguageService: LanguageService, Sendable {
   ///    `buildSettings(for:)` returns `nil` are not included.
   private var buildSettingsForOpenFiles: [DocumentURI: SwiftCompileCommand] = [:]
 
-  /// Calling `scheduleCall` on `refreshDiagnosticsDebouncer` schedules a `DiagnosticsRefreshRequest` to be sent to
-  /// to the client.
+  /// Calling `scheduleCall` on `refreshDiagnosticsAndSemanticTokensDebouncer` schedules a `DiagnosticsRefreshRequest`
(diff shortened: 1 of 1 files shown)


hamishknight (approved):

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #2151, opened 2025-05-13
Fix visibility on import triggering a compile warning

Wilfred:
Previously, building sourcekit-lsp would produce the following warning:

    $ swift build
    Building for debugging...
    /Users/wilfred/src/sourcekit-lsp/Sources/SKLogging/CustomLogStringConvertible.swift:13:9: warning: package import of 'Foundation' was not used in package declarations
     11 | //===----------------------------------------------------------------------===//
     12 |
     13 | package import Foundation
        |         `- warning: package import of 'Foundation' was not used in package declarations

Remove the visibility modifier, so no warning is produced.

### DIFF
--- a/Sources/SKLogging/CustomLogStringConvertible.swift
+++ b/Sources/SKLogging/CustomLogStringConvertible.swift
@@ -10,7 +10,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+// NSObject is defined in Foundation on non-darwin platforms. On
+// darwin, NSObject is package-visible.
+#if canImport(ObjectiveC)
+import Foundation
+#else
 package import Foundation
+#endif
 
 #if !NO_CRYPTO_DEPENDENCY
 import Crypto


ahoppen (approved):
Thanks for opening the PR to fix this. Looks good to me.

Wilfred:
Argh, the `package import` is needed when you're using the non-darwin logger. I've updated the PR.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Wilfred:
OK, I've tweaked the conditional to check whether we're on macOS, which seems to fix the build.

I don't understand why the warning only triggered on macOS though. I see `InternalImportsByDefault` is set in `Package.swift`.

```
/home/wilfred/src/sourcekit-lsp/Sources/SKLogging/CustomLogStringConvertible.swift:40:21: error: class cannot be declared package because its superclass is internal
 14 | import Foundation
 15 | #else
 16 | import Foundation
    | `- note: class 'NSObject' imported as 'internal' from 'Foundation' here
 17 | #endif
 18 | 
    :
 38 | /// thus pass this object to OSLog, which just forwards to `description` or
 39 | /// `redactedDescription` of an object that implements `CustomLogStringConvertible`.
 40 | package final class CustomLogStringConvertibleWrapper: NSObject, Sendable {
    |                     |- error: class cannot be declared package because its superclass is internal
    |                     `- note: class 'NSObject' is imported by this file as 'internal' from 'Foundation'
```

This is the build failure I see on my Linux box, which suggests that the Linux build is treating `import Foundation` as `private import Foundation`, whereas the macOS build isn't for some reason. `CustomLogStringConvertibleWrapper` is still defined on both platforms unless I'm missing something.

TLDR: This fixes the warning and I think should make CI green, although there's some detail I don't understand.

bnbarham:
I believe the difference here is where `NSObject` is defined (seems to be Foundation on non-Darwin), though I'm not positive how we end up finding a `package`-visible-NSObject on macOS 🤔 

We could maybe make this:
```
#if canImport(ObjectiveC)
import Foundation
package import ObjectiveC
#else
package import Foundation
#endif
```

To make it super obvious, or just:
```
#if canImport(ObjectiveC)
import Foundation
#else
package import Foundation
#endif
```
?

Wilfred:
OK, I've updated. I tried @bnbarham's first suggestion, but that also produces build warnings:

```
$ swift build
[1/1] Planning build
Building for debugging...
/Users/wilfred/src/sourcekit-lsp/Sources/SKLogging/CustomLogStringConvertible.swift:16:9: warning: package import of 'ObjectiveC' was not used in package declarations
 14 | #if canImport(ObjectiveC)
 15 | import Foundation
 16 | package import ObjectiveC
    |         `- warning: package import of 'ObjectiveC' was not used in package declarations
 17 | #else
 18 | package import Foundation
```

so I went with the second option.

Sources/SKLogging/CustomLogStringConvertible.swift
```diff
@@ -10,7 +10,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+// NSObject is defined in Foundation on non-darwin platforms. On
+// darwin, NSObject is package-visible.
+#if canImport(ObjectiveC)
```
ahoppen, line 15: One small nitpick: We usually use `#if canImport(Darwin)` to conditionally compile code for Darwin vs. non-Darwin platforms. I’d prefer to be consistent with that pattern here.

ahoppen (commented):
Thanks you so much for your continued effort on this PR 🙏🏽

Sources/SKLogging/CustomLogStringConvertible.swift
bnbarham: I'll merge this one and update to Darwin separately 👍

bnbarham:
@swift-ci please test

bnbarham:
```
error: class cannot be declared package because its superclass is internal
```

😅 I wonder if this was just a compiler bug and it's now fixed? I'm not seeing the warning on the 21 beta

ahoppen:
I think I found the underlying issue of this warning because really we should need to import `Foundation` at the package access level https://github.com/swiftlang/swift/issues/83893

I haven’t quite understood why we sometimes see the warning and sometimes don’t but my bet is that it’s got something to do with us using `@_exported import os` in a different file of `SKLogging` and that some order of type checker evaluation or the decision by the driver of which files should be considered primary vs. secondary in a given `swift-frontend` invocation plays a difference. 

I’m working around this in https://github.com/swiftlang/sourcekit-lsp/pull/2253 by marking the declaration as `public` instead of `package`. Let’s close this PR as I’ll be integrating the fix in #2253.

Wilfred:
FWIW there's some related discussion here about Foundation visibility across platforms: https://forums.swift.org/t/formalizing-the-unavailability-of-core-foundation/40216

PR #2152, opened 2025-05-13
Fix race condition that causes task cancellation to be missed in `TaskScheduler`

ahoppen:
A queued task might have been cancelled after the execution ask was started but before the task was yielded to `executionTaskCreatedContinuation`. In that case the result task will simply cancel the await on the `executionTaskCreatedStream` and hence not call `valuePropagatingCancellation` on the execution task. This means that the queued task cancellation wouldn't be propagated to the execution task. To address this, check if `resultTaskCancelled` was  set and, if so, explicitly cancel the execution task here.

Fixes an issue I saw in CI during PR testing.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2153, opened 2025-05-13
Do not schedule any new tasks in `TaskScheduler` if it has been shut down

ahoppen:
This is what `shutDown()` is documented to do. I also remember having this check before, it might have gotten lost during a rebase when I was working on https://github.com/swiftlang/sourcekit-lsp/pull/2081.

I noticed this while investigating https://github.com/swiftlang/sourcekit-lsp/pull/2152: In this case `buildTarget/prepare` was cancelled because the SourceKit-LSP server was shut down but indexing of a file was still started after the shutdown now that preparation had finished (because it was cancelled).


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2156, opened 2025-05-13
Add a `--enable-inlay-hints` setting

rbenegal:
The default behaviour for IDEs such as VS Code is to enable all inlay hints. It is possible to turn off inlay hints for Swift through VS Code through user settings but it may be tricky for new users to figure out how to do this. In order to not confuse new users and avoid annoying experienced users, it would be better to turn off the inlay hints by default. Most VS Code extensions such as  Go, Python etc. disable inlay hints (for variable type inference, at least) via the LSP probably to avoid touching user/IDE default settings.

This PR is an initial attempt at a new setting that will allow for inlay hint requests to be ignored if `--enable-inlay-hints` is set to `false` during SourceKit-LSP startup. The goal is to have Swift VS Code extension to disable inlay hints by default until there is a better way to do this via VS Code without having to edit user settings.

### DIFF
--- a/Sources/SKOptions/SourceKitLSPOptions.swift
+++ b/Sources/SKOptions/SourceKitLSPOptions.swift
@@ -433,6 +433,9 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     return .seconds(300)
   }
 
+  /// Enable or disable registering inlay hints for the inlayHintProvider
+  public var enableInlayHints: Bool? = nil
+
   public init(
     swiftPM: SwiftPMOptions? = .init(),
     fallbackBuildSystem: FallbackBuildSystemOptions? = .init(),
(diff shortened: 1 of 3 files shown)


rbenegal:
Suggestions for other ways to do this welcome. I was told I should explore doing this through capabilities directly, as it would be good to not have to restart the LSP to re-enable inlay hints, but putting up the initial idea here for comments.

rbenegal:
Good idea, thanks for the link to the vscode-swift repo @MahdiBM!

ahoppen:
We shouldn’t disable inlay hints in SourceKit-LSP by default. SourceKit-LSP is just reporting that it *can* provide inlay hints and it’s up to the editor to request them. With this PR, it’s very confusing how to enable inlay hints because you need to enable it both in your editor and in the SourceKit-LSP config file (remember that there are other editors that might have different default settings than VS Code). If anything, we should do something in the VS Code Swift extension to disable inlay hints by default – I’m not voicing an opinion on whether we should or should not do that.

rbenegal:
Thanks for your comments and concerns @ahoppen. This PR doesn't disable inlay hints in SourceKit-LSP. It only adds a setting where if you pass in `--enable-inlay-hints` _and_ `false` then it will not register the inlay hints handler. I'm open to other suggestions on how to do this, doesn't have to be a setting. Inlay hints will still be (and should continue to be) enabled by default in SourceKit itself.

The disabling will be done by the VS Code swift extension but it doesn't seem like there is good way to do this from an extension level without touching user settings (see thread here: https://github.com/swiftlang/vscode-swift/pull/1526). The other LSPs have the options to override the textDocument/inlayHints requests they receive whereas SourceKit-LSP does not. Other LSPs also have more inlay hint abilities as mentioned by Mahdi here https://github.com/swiftlang/vscode-swift/pull/1526#discussion_r2088352462 which we can control individually in the future as well when they are implemented. Suggestions welcome.

PR #2157, opened 2025-05-14
Pass `hostToolchainBinDir` to `SwiftSDKBundleStore`

MaxDesiatov:
This became a required parameter in https://github.com/swiftlang/swift-package-manager/pull/8668, which can be easily computed, since the host toolchain in practice is always available when `SwiftSDKBundleStore` is initialized.


MaxDesiatov:
https://github.com/swiftlang/swift-package-manager/pull/8668
@swift-ci test

ahoppen (approved):

PR #2167, opened 2025-05-30
FreeBSD: Fix build for FreeBSD

etcwilde:
This fixes two build failures that are part of SourceKit-LSP.
 - Missing sqlite3 header.
 - Incorrect call to `setpriority` API

FreeBSD `pkg` installs the sqlite3 header to `/usr/local/include`. Adding that as a search path for importing.
The `setpriority` API is imported in such a way that does not require any changes to the variables passed in. The conversion to `UInt32` and undefined `__priority_which_t` resulted in build failures.


etcwilde:
@swift-ci please test

Utilities/build-script-helper.py
```diff
@@ -137,6 +137,11 @@ def get_swiftpm_options(swift_exec: str, args: argparse.Namespace, suppress_verb
         swiftpm_args += [
             '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/android',
         ]
+    elif '-freebsd' in build_target:
+        # pkg installs packages to /usr/local/include on FreeBSD
+        # Required for SwiftPM to find sqlite
+        swiftpm_args += ['-Xcxx', '-I', '-Xcxx', '/usr/local/include',
+                         '-Xswiftc', '-I', '-Xswiftc', '/usr/local/include']
```
ahoppen, line 144: Should SwiftPM be fixed to add these search paths on FreeBSD? My expectation is that if we don’t need them on any other Unix distributions, we shouldn’t need them here either. 

To be clear, I’m fine with this change, just want to make sure we don’t forget to do the proper solution if there is one in the future. Maybe have an issue for it?

ahoppen (approved):

ahoppen:
@swift-ci Please test Windows

Utilities/build-script-helper.py
etcwilde: Created an issue for having SwiftPM figure out what search paths it should add: https://github.com/swiftlang/swift-package-manager/issues/8751

Adding random paths willy-nilly is a path to madness if that's not what is intended by the user, so I don't think adding it by default is the right move. Especially if there is no way to override it. If something is using the non-system sqlite for instance, it should use the search paths from the library instead of force-passing the path for the sqlite on the system (unless needed to find another library of course, though then we get into modularization issues... ¯\\\_(ツ)\_/¯).

ahoppen: Thanks for filing the issue. Let’s see if there is a proper solution / what that solution is.

PR #2170, opened 2025-06-04
Update for new `ExpandEditorPlaceholdersToLiteralClosures` API

hamishknight:
This refactoring now takes a `Syntax` parameter. We don't actually need to change anything else to handle placeholder expansion for macro expansion completions since they get parsed as function calls.


hamishknight:
https://github.com/swiftlang/swift-syntax/pull/3092

@swift-ci please test

ahoppen (approved):
Thanks

PR #2171, opened 2025-06-05
Compute the prefix of `orLog` only when the closure threw an error

ahoppen:
Probably not significant in practice but I noticed that we don’t need to compute the prefix of `orLog` unless the closure threw an error. Make `prefix` an autoclosure to avoid its eager computation.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2177, opened 2025-06-11
Fix the issue of indexPrefixMap being unordered, which causes the IndexStoreDB path to be unpredictable.

Kila2:


Sources/SourceKitLSP/Workspace.swift
```diff
@@ -306,6 +306,13 @@ package final class Workspace: Sendable, BuildSystemManagerDelegate {
         indexDelegate = SourceKitIndexDelegate()
         let prefixMappings =
           (indexOptions.indexPrefixMap ?? [:])
+          .sorted {
+              if $0.original.count != $1.original.count {
+                  return $0.original.count > $1.original.count
+              } else {
+                  return $0.original < $1.original
+              }
```
ahoppen: I assume that you want to sort by the string’s length to prefer more specific paths, right? Could you add a short comment explaining that?

Also, could you format your source code as described in https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting?

ahoppen (commented):
Thanks, good fix. I have a small comment inline but otherwise looks good to me.

Sources/SourceKitLSP/Workspace.swift
Kila2: Done. Thanks for the suggestion!

bnbarham:
@swift-ci please test

bnbarham (approved):
Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2179, opened 2025-06-11
Avoid crashing on invalid range in `editDocument`

hamishknight:
Rename `LineTable.replace(utf8Offset:length:with)` to `tryReplace` and bail if the provided range is out of bounds of the buffer. This ensures we match the behavior of SourceKit when handling an `editor.replacetext` request.

rdar://161268691

### DIFF
--- a/Sources/SKUtilities/LineTable.swift
+++ b/Sources/SKUtilities/LineTable.swift
@@ -115,21 +115,47 @@ extension LineTable {
     self.replace(fromLine: fromLine, utf16Offset: fromOff, toLine: toLine, utf16Offset: toOff, with: replacement)
   }
 
+  private struct OutOfBoundsError: Error, CustomLogStringConvertible {
+    // Note we use tuples here rather than Range since the latter would assert
+    // that upperBound >= lowerBound.
+    var utf8Range: (lower: Int, upper: Int)
+    var utf8Bounds: (lower: Int, upper: Int)
+
+    var description: String {
+      """
+      \(utf8Range.lower)..<\(utf8Range.upper) is out of bounds \
+      \(utf8Bounds.lower)..<\(utf8Bounds.upper)
+      """
+    }
+
+    var redactedDescription: String {
+      description
+    }
+  }
+
   /// Replace the line table's `content` in the given range and update the line data.
+  /// If the given range is out-of-bounds, throws an error.
   ///
-  /// - parameter fromLine: Starting line number (zero-based).
-  /// - parameter fromOff: Starting UTF-8 column offset (zero-based).
-  /// - parameter toLine: Ending line number (zero-based).
-  /// - parameter toOff: Ending UTF-8 column offset (zero-based).
+  /// - parameter utf8Offset: Starting UTF-8 offset (zero-based).
+  /// - parameter length: UTF-8 length.
   /// - parameter replacement: The new text for the given range.
   @inlinable
-  mutating package func replace(
+  mutating package func tryReplace(
     utf8Offset fromOff: Int,
     length: Int,
     with replacement: String
-  ) {
-    let start = content.utf8.index(content.startIndex, offsetBy: fromOff)
-    let end = content.utf8.index(content.startIndex, offsetBy: fromOff + length)
+  ) throws {
+    let utf8 = self.content.utf8
+    guard
+      fromOff >= 0, length >= 0,
+      let start = utf8.index(utf8.startIndex, offsetBy: fromOff, limitedBy: utf8.endIndex),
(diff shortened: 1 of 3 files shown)


hamishknight:
@swift-ci please test

Sources/SKTestSupport/Utils.swift
```diff
@@ -124,3 +125,15 @@ var globalModuleCache: URL? {
       .appendingPathComponent("shared-module-cache")
   }
 }
+
+/// Similar to XCTAssertThrowsError, but accepts an async function.
```
ahoppen: We should already have these functions here: https://github.com/swiftlang/sourcekit-lsp/blob/2a8592b93a1041ddb2d6f4a5d3717735ba225b07/Sources/SKTestSupport/Assertions.swift#L40-L76

Or don’t they work for some reason?

Sources/SKUtilities/LineTable.swift
```diff
@@ -116,44 +116,26 @@ extension LineTable {
...
-  /// Replace the line table's `content` in the given range and update the line data.
-  ///
-  /// - parameter fromLine: Starting line number (zero-based).
-  /// - parameter fromOff: Starting UTF-8 column offset (zero-based).
-  /// - parameter toLine: Ending line number (zero-based).
-  /// - parameter toOff: Ending UTF-8 column offset (zero-based).
-  /// - parameter replacement: The new text for the given range.
-  @inlinable
-  mutating package func replace(
+  mutating package func tryReplace(
```
ahoppen, line 143: I think it’s odd that the UTF-8-based replace methods now continue on error while the UTF-16-based methods trap (even though it is indicated by `tryReplace` vs `replace`). What do you think about changing `replace` to `throws` and then let the client handle an out-of-bounds error. For SourceKit-LSP I think crashing is still a reasonable resolution because we would definitely have an out-of-date document state and if we crash, the editor should re-launch us and thus send the new document state. For the plugin, I think wrapping it in `orLog` is reasonable to match the current behavior of SourceKitService. Although I find it quite concerning if we get out-of-bound replacements.

Sources/SKTestSupport/Utils.swift
hamishknight: Oh, I totally missed these!

Sources/SKUtilities/LineTable.swift
hamishknight: > For SourceKit-LSP I think crashing is still a reasonable resolution

Maybe I'm misunderstanding, but doesn't LSP currently recover by clipping the provided location to the start and end locations?

https://github.com/swiftlang/sourcekit-lsp/blob/69150c2d7214e86b57ddec6b3ad577e922f551db/Sources/SKUtilities/LineTable.swift#L241-L245

But yeah, I changed the name to `tryReplace` to indicate it has a different behavior to `replace`. Happy to make it `throw` + `orLog` instead.

hamishknight: Updated to throw an error

hamishknight:
@swift-ci please test

Sources/SKUtilities/LineTable.swift
ahoppen: Huh, I could have sworn that we crashed for out-of-bound edits but I changed that in https://github.com/swiftlang/sourcekit-lsp/pull/1798. Sorry for the back and forth but I think my main motivation is that the UTF-8 and UTF-16-based methods should behave consistently. With that new information, I think I’d prefer if we’d just match the UTF-16 behavior for UTF-8 and use 

```swift
let start = self.stringIndexOf(line: fromLine, utf8Column: fromOff)
let end = self.stringIndexOf(line: toLine, utf8Column: toOff)
```

That should fix the crash with the standard log behavior, right?

hamishknight:
@swift-ci please test

bnbarham:

hamishknight:
@swift-ci please test

Sources/SKUtilities/LineTable.swift
hamishknight: Sorry for taking so long to get back to this. We don't want the clipping behavior here because we need to keep the edits in sync with those in sourcekitd, when we last spoke about this you suggested to make the clipping behavior a configurable parameter. I took a look at doing that, but it ended up making things really awkward for all the other uses of `replace`. So for now I've just renamed this variant of `replace` back to `tryReplace`. Does that seem reasonable to you?

ahoppen: Thanks for investigating making this configurable. I’m fine with merging this as `tryReplace`.

ahoppen (approved):

PR #2187, opened 2025-06-23
Correctly handle raw identifiers for swift-testing

hamishknight:
Make sure we insert backticks if needed for the test identifier, and drop the parameter names from the display name if a raw identifier is used.

### DIFF
--- a/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
+++ b/Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
@@ -305,6 +305,20 @@ final class SyntacticSwiftTestingTestScanner: SyntaxVisitor {
     return visitTypeOrExtensionDecl(node, typeNames: [identifier.name])
   }
 
+  /// If the given identifier requires backticks to be a valid decl identifier,
+  /// applies backticks and returns `true` along with the new name. Otherwise
+  /// returns `false` with the original name.
+  func backtickIfNeeded(_ ident: Identifier) -> (backticked: Bool, name: String) {
+    let name = ident.name
+    if name == "``" {
+      // Special case: `` stays as ``. Any other backticked name will have
+      // been stripped by Identifier.
+      return (true, name)
+    }
+    let needsBackticks = !name.isValidSwiftIdentifier(for: .variableName)
+    return (needsBackticks, needsBackticks ? "`\(name)`" : name)
+  }
+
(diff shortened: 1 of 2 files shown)


hamishknight:
@swift-ci please test

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
```diff
@@ -318,18 +332,28 @@ final class SyntacticSwiftTestingTestScanner: SyntaxVisitor {
...
+          // Something like `_`, leave as-is.
+          param.firstName.text
+        }
+      return "\(result):"
     }.joined()
-    let name = "\(identifier.name)(\(parameters))"
+
+    let (hasBackticks, baseName) = backtickIfNeeded(identifier)
+    let fullName = "\(baseName)(\(parameters))"
+    let displayName = attributeData.displayName ?? (hasBackticks ? identifier.name : fullName)
```
ahoppen, line 348: Does this mean that we include the parameter labels only if the base name doesn’t require backticks? I don’t see the relation between omitting parameter labels and a backticked base name but maybe I’m missing something.

ahoppen (commented):
Thank you!

Sources/SourceKitLSP/Swift/SwiftTestingScanner.swift
hamishknight: Yes that's correct, the idea here is that a raw identifier is an alternative way of spelling a display name for a test, e.g these have equivalent display names:

```swift
@Test("foo bar") func foo() {}
@Test func `foo bar`() {}
```

and we don't include parameter labels for the former case, so shouldn't in the latter case either.

ahoppen: OK, interesting choice and I think I would have treated raw identifiers just like any other identifier (expecting that wrapping an identifier in backticks wouldn't cause a real change) but I don't have strong enough of an opinion to argue about it. 

Could you add a short comment with your explanation though since the choice isn't totally obvious.

hamishknight: Added a comment in https://github.com/swiftlang/sourcekit-lsp/pull/2191

ahoppen: Thank you 😄

PR #2191, opened 2025-06-25
Correctly handle raw identifier test suite names

hamishknight:
Missed this in my previous patch (#2187), make sure we also add backticks to test suite names when they use raw identifiers.


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test Windows

PR #2193, opened 2025-06-30
Allow overriding SwiftPM's build directory with background indexing

bnbarham:
If `scratchPath` is explicitly set, use it over the separate `index-build` arena even when background indexing is enabled.

### DIFF
--- a/Sources/SemanticIndex/PreparationTaskDescription.swift
+++ b/Sources/SemanticIndex/PreparationTaskDescription.swift
@@ -63,9 +63,10 @@ package struct PreparationTaskDescription: IndexTaskDescription {
     targetsToPrepare: [BuildTargetIdentifier],
     buildSystemManager: BuildSystemManager,
     preparationUpToDateTracker: UpToDateTracker<BuildTargetIdentifier, DummySecondaryKey>,
-    logMessageToIndexLog: @escaping @Sendable (
-      _ message: String, _ type: WindowMessageType, _ structure: StructuredLogKind
-    ) -> Void,
+    logMessageToIndexLog:
+      @escaping @Sendable (
+        _ message: String, _ type: WindowMessageType, _ structure: StructuredLogKind
+      ) -> Void,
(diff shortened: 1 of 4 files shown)


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
Also applied formatting changes caused by https://github.com/swiftlang/swift-format/pull/1039 (ideally we'd have github actions and use the latest release here but...)

hamishknight (approved):

bnbarham:

Sources/BuildSystemIntegration/SwiftPMBuildSystem.swift
```diff
@@ -237,10 +237,10 @@ package actor SwiftPMBuildSystem: BuiltInBuildSystem {
       fileSystem: localFileSystem
     )
 
-    if options.backgroundIndexingOrDefault {
-      location.scratchDirectory = absProjectRoot.appending(components: ".build", "index-build")
-    } else if let scratchDirectory = options.swiftPMOrDefault.scratchPath {
+    if let scratchDirectory = options.swiftPMOrDefault.scratchPath {
```
ahoppen, line 240: I wonder if we should use an `index-build` folder within the scratch directory for background indexing. Say a user stores their build directory in `.my-build` instead of `.build` for some reason. In that case they might specify `.my-build` as the scratch path but we‘d probably still want to use a separate build arena for background indexing.

bnbarham: I was torn on this. Technically there's no reason to match up their build + what's passed to sourcekit-lsp today, since we have a separate arena anyway. So if it is passed... seems weird to then tack on the arena?

ahoppen: I guess… My opinion isn’t strong enough to argue about this.

PR #2194, opened 2025-07-03
CMake: Fix SwiftCrypto/SwiftASN1 order

etcwilde:
SwiftCrypto requires SwiftASN1. SwiftCrypto should be declaring the dependency in its import files, but is not currently doing so resulting in `find_package(SwiftCrypto)` failing because the `SwiftASN1` target does not yet exist.
By placing `find_package(SwiftASN1)` before finding SwiftCrypto, we ensure that the `SwiftASN1` target is populated ahead of time and finding SwiftCrypto succeeds.

The fix to the missing dependency is posted for review: https://github.com/apple/swift-crypto/pull/376.
This change is to allow building SourceKit-LSP without needing a new tag on swift-crypto.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,8 @@ find_package(LLBuild QUIET)
 find_package(ArgumentParser CONFIG REQUIRED)
 find_package(SwiftCollections QUIET)
 find_package(SwiftSyntax CONFIG REQUIRED)
-find_package(SwiftCrypto CONFIG REQUIRED)
 find_package(SwiftASN1 CONFIG REQUIRED)
+find_package(SwiftCrypto CONFIG REQUIRED)
 
 include(SwiftSupport)
 


etcwilde:
CC @compnerd

etcwilde:
@swift-ci please test

etcwilde:
> This should be safe to go in before the swift-crypto change right?

Yes, this is so that the swift-crypto change isn't needed and I don't have to wait for a new crypto tag to bump the package versions. Once SourceKit-LSP depends on a version that is at least as new as that change, we should be able to drop the `find_package(SwiftASN1)` here since that will be handled by the `find_package(SwiftCrypto CONFIG)`, unless we want to do something special.

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2198, opened 2025-07-10
Always try to render some documentation for Swift files

matthewbastien:
The error message `No documentation could be rendered for the position in this document` can be rather confusing. It was also being displayed in situations where the cursor was placed at the beginning or end of a Swift file.

This PR changes the behaviour of finding the nearest symbol to the cursor in a`textDocument/doccDocumentation` request for Swift files. If the token at the cursor does not point at a valid documentable symbol we will now walk forwards through the list of tokens to see if any of them are a valid documentable symbol. If this still fails then we walk backwards from the cursor until we find a documentable symbol. The error message has also been changed to `No documentable symbols were found in this Swift file` in the event that no symbols are available at all for display.

Also, I've added a separate error message for unsupported languages with `textDocument/doccDocumentation`.

### DIFF
--- a/Sources/SourceKitLSP/Swift/DoccDocumentation.swift
+++ b/Sources/SourceKitLSP/Swift/DoccDocumentation.swift
@@ -157,15 +157,25 @@ fileprivate struct DocumentableSymbol {
   }
 
   static func findNearestSymbol(syntaxTree: SourceFileSyntax, position: AbsolutePosition) -> DocumentableSymbol? {
-    // token(at:) can return nil if the position is at the end of the document. Fall back to using the last token in this case.
-    let token = syntaxTree.token(at: position) ?? syntaxTree.lastToken(viewMode: .all)
+    let token: TokenSyntax
+    if let tokenAtPosition = syntaxTree.token(at: position) {
+      token = tokenAtPosition
+    } else if position >= syntaxTree.endPosition, let lastToken = syntaxTree.lastToken(viewMode: .sourceAccurate) {
+      // token(at:) returns nil if position is at the end of the document.
+      token = lastToken
+    } else if position < syntaxTree.position, let firstToken = syntaxTree.firstToken(viewMode: .sourceAccurate) {
+      // No case in practice where this happens but good to cover anyway
+      token = firstToken
+    } else {
+      return nil
+    }
     // Check if the current token is within a valid documentable symbol
-    if let token, let symbol = token.ancestorOrSelf(mapping: { DocumentableSymbol(node: $0) }) {
+    if let symbol = token.ancestorOrSelf(mapping: { DocumentableSymbol(node: $0) }) {
       return symbol
     }
     // Walk forward through the tokens until we find a documentable symbol
-    var previousToken = token
-    while let nextToken = previousToken?.nextToken(viewMode: .all) {
+    var previousToken: TokenSyntax? = token
+    while let nextToken = previousToken?.nextToken(viewMode: .sourceAccurate) {
       if let symbol = nextToken.ancestorOrSelf(mapping: { DocumentableSymbol(node: $0) }) {
         return symbol
       }
@@ -173,7 +183,7 @@ fileprivate struct DocumentableSymbol {
     }
     // Walk backwards through the tokens until we find a documentable symbol
     previousToken = token
-    while let nextToken = previousToken?.previousToken(viewMode: .all) {
+    while let nextToken = previousToken?.previousToken(viewMode: .sourceAccurate) {
       if let symbol = nextToken.ancestorOrSelf(mapping: { DocumentableSymbol(node: $0) }) {
         return symbol
       }


matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test Windows

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
```diff
@@ -130,30 +130,56 @@ fileprivate struct DocumentableSymbol {
...
+    visitor.walk(syntaxTree)
+    return visitor.topLevelSymbol
+  }
+
+  static func findNearestSymbol(syntaxTree: SourceFileSyntax, position: AbsolutePosition) -> DocumentableSymbol? {
+    guard let token = syntaxTree.token(at: position) else {
       return nil
     }
+    // Walk up the tree until we find a documentable symbol
+    return token.ancestorOrSelf { DocumentableSymbol(node: $0) }
```
bnbarham: Rather than giving up and returning the first symbol from `findTopLevelSymbol`, could we instead find the closest symbol? I'd probably define "closest" here as "next or previous if there is no next".

Sources/DocCDocumentation/DoccDocumentationError.swift
```diff
@@ -14,12 +14,15 @@ import Foundation
 package import LanguageServerProtocol
 
 package enum DocCDocumentationError: LocalizedError {
+  case unsupportedLanguage(String)
```
ahoppen: Should we keep the `Language` type as long as possible and only serialize the `Language` enum to a string when forming the description?

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
```diff
@@ -130,30 +130,56 @@ fileprivate struct DocumentableSymbol {
...
-  static func findNearestSymbol(syntaxTree: SourceFileSyntax, position: AbsolutePosition) -> DocumentableSymbol? {
-    guard let token = syntaxTree.token(at: position) else {
+  init?(node: any SyntaxProtocol) {
+    if let namedDecl = node.asProtocol(NamedDeclSyntax.self) {
+      self = DocumentableSymbol(node: namedDecl, position: namedDecl.name.positionAfterSkippingLeadingTrivia)
+    } else if let initDecl = node.as(InitializerDeclSyntax.self) {
+      self = DocumentableSymbol(node: initDecl, position: initDecl.initKeyword.positionAfterSkippingLeadingTrivia)
+    } else if let functionDecl = node.as(FunctionDeclSyntax.self) {
+      self = DocumentableSymbol(node: functionDecl, position: functionDecl.name.positionAfterSkippingLeadingTrivia)
+    } else if let variableDecl = node.as(VariableDeclSyntax.self) {
```
ahoppen, line 144: Should we also include `SubscriptDeclSyntax` and `DeinitializerDeclSyntax` here?

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
ahoppen: I agree here, returning the first top-level symbol is probably confusing if there are eg. two types defined in a file and the user has the cursor at the end of the document.

Sources/DocCDocumentation/DoccDocumentationError.swift
matthewbastien: Good point. Done.

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
matthewbastien: Forgot about those when I originally wrote this! Added them along with some tests.

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
matthewbastien: Changed from using the findTopLevelSymbol to first walking the tokens forward until we find a documentable symbol and, if we hit the end, walking backwards until we find one.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows

matthewbastien:
@swift-ci please test windows

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
```diff
@@ -130,30 +130,57 @@ fileprivate struct DocumentableSymbol {
...
-      } else if let enumCaseDecl = node.as(EnumCaseDeclSyntax.self) {
-        guard let name = enumCaseDecl.elements.only?.name else {
-          return nil
-        }
-        return DocumentableSymbol(node: enumCaseDecl, position: name.positionAfterSkippingLeadingTrivia)
+  }
+
+  static func findNearestSymbol(syntaxTree: SourceFileSyntax, position: AbsolutePosition) -> DocumentableSymbol? {
+    // token(at:) can return nil if the position is at the end of the document. Fall back to using the last token in this case.
+    let token = syntaxTree.token(at: position) ?? syntaxTree.lastToken(viewMode: .all)
```
ahoppen: Technically, `token(at:)` could also return `nil` if `position` is before the start position of `syntaxTree`. Not sure why this would happen but I think it would be better to be explicit about what we try to achieve here. Something like 

```swift
let token: TokenSyntax
if let tokenAtPosition = syntaxTree.token(at: position) {
  token = tokenAtPosition
} else if position >= syntaxTree.endPosition, let lastToken = syntaxTree.lastToken(viewMode: .sourceAccurate) {
  // token(at:) returns nil if position is at the end of the document.
  token = lastToken
} else if position < syntaxTree.position, let firstToken = syntaxTree.firstToken(viewMode: .sourceAccurate) {
  // No case in practice where this happens but good to cover anyway
  token = firstToken
} else {
  return nil
}
```

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
```diff
@@ -130,30 +130,57 @@ fileprivate struct DocumentableSymbol {
...
+    // Check if the current token is within a valid documentable symbol
+    if let token, let symbol = token.ancestorOrSelf(mapping: { DocumentableSymbol(node: $0) }) {
+      return symbol
+    }
+    // Walk forward through the tokens until we find a documentable symbol
+    var previousToken = token
+    while let nextToken = previousToken?.nextToken(viewMode: .all) {
```
ahoppen: We should only need to walk tokens that occur in the source code here, ie. change this from `.all` to `.sourceAccurate`, I think. Same below.

ahoppen (commented):
That looks straightforward to me. Two minor comments.

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
matthewbastien: Good point. Done.

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
matthewbastien: Done.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1091, opened 2024-02-27
Emit raw documentation comments instead of parsing XML

ahoppen:
While at it, clean up the hover tests

Fixes https://github.com/apple/sourcekit-lsp/issues/1014
rdar://120685874

Issue #1014, opened 2024-01-09
Bullets in doc comments rendered as "Discussion," labels are dropped.


dabrahams:
For example, in all clients I can get my hands on:

```
  /// Returns a sequence whose elements describe each rule being recognized at
  /// `position`, where the rule started, and how many of its RHS symbols have
  /// been recognized.
  ///
  /// - Precondition: no other progress report for `self` exists.
```

The precondition is rendered as 

> **Discussion**
>
> no other progress report for `self` exists.

Which is really misleading.  Even if “Discussion” is unavoidable, the bullet label should be preserved (perhaps unless it is exactly “`Discussion`”)

ahoppen:
Tracked in Apple’s issue tracker as rdar://120685874

PR #2207, opened 2025-07-15
Fetch full documentation in code completion

a7medev:
Depends on https://github.com/swiftlang/swift/pull/82464

---

When resolving documentation for code completion items, we fetch full documentation through the newly added `swiftide_completion_item_get_doc_full_copy` SourceKitD function, if not found we fallback to brief documentation as before using `swiftide_completion_item_get_doc_brief`.

> [!NOTE]
> Unlike brief documentation, SourceKitD doesn't cache full documentation for completion results to avoid bloating memory with a lot of large strings.
>
> As of now, SourceKit-LSP doesn't cache completion item documentation either, should we introduce a new full documentation cache (e.g. using `LRUCache`)?


Sources/SourceKitD/sourcekitd_functions.swift
```diff
@@ -146,6 +146,7 @@ extension sourcekitd_ide_api_functions_t {
       completion_item_get_source_text: try loadRequired("swiftide_completion_item_get_source_text"),
       completion_item_get_type_name: try loadRequired("swiftide_completion_item_get_type_name"),
       completion_item_get_doc_brief: try loadRequired("swiftide_completion_item_get_doc_brief"),
+      completion_item_get_doc_full_copy: try loadRequired("swiftide_completion_item_get_doc_full_copy"),
```
ahoppen: Can we make this a `loadOptional` so that SourceKitD can work with Swift toolchains that don’t include `swiftide_completion_item_get_doc_full_copy` yet?

Sources/Csourcekitd/include/CodeCompletionSwiftInterop.h
```diff
@@ -324,6 +324,12 @@ typedef struct {
     void (^_Null_unspecified handler)(const char *_Null_unspecified)
   );
 
+  void (*_Nonnull completion_item_get_doc_full_copy)(
+    _Null_unspecified swiftide_api_completion_response_t,
+    _Null_unspecified swiftide_api_completion_item_t,
+    void (^_Null_unspecified handler)(char *_Null_unspecified)
+  );
```
ahoppen: Since you’re adding this as a new function, could you use `_Nullable` or `_Nonnull` instead of `_Null_unspecified`? I was just too lazy to figure out the nullability of all the other functions when I added the annotations a while back.

Sources/SwiftSourceKitPlugin/ASTCompletion/CompletionSession.swift
```diff
@@ -241,6 +241,17 @@ struct ExtendedCompletionInfo {
     return result
   }
 
+  var fullDocumentation: String? {
+    var result: String? = nil
+    session.sourcekitd.ideApi.completion_item_get_doc_full_copy(session.response, rawItem) {
+      if let cstr = $0 {
+        result = String(cString: cstr)
+        free(cstr)
```
ahoppen: Do we need to free the pointer here? I would assume that the memory is still owned by sourcedkitd since it only yields the pointer to us in a closure and is thus still responsible for its lifetime.

Sources/SwiftSourceKitPlugin/CompletionProvider.swift
```diff
@@ -262,10 +262,17 @@ actor CompletionProvider {
...
-      request.sourcekitd.keys.docBrief: info.briefDocumentation,
-      request.sourcekitd.keys.associatedUSRs: info.associatedUSRs as [SKDResponseValue]?,
+    var response = request.sourcekitd.responseDictionary([
+      request.sourcekitd.keys.associatedUSRs: info.associatedUSRs as [SKDResponseValue]?
     ])
+
+    if let fullDocumentation = info.fullDocumentation {
+      response.set(request.sourcekitd.keys.docFullAsXML, to: fullDocumentation)
+    } else {
+      response.set(request.sourcekitd.keys.docBrief, to: info.briefDocumentation)
```
ahoppen: Open question: Does it make sense to also return the brief documentation in addition to the full documentation? Eg. an editor might want to display a brief documentation of the symbol by default and then expand that to full documentation once the user performs an action.

Tests/SwiftSourceKitPluginTests/SwiftSourceKitPluginTests.swift
```diff
@@ -436,15 +436,34 @@ final class SwiftSourceKitPluginTests: XCTestCase {
...
+                   + "<Name>foo1()</Name>"
+                   + "<USR>s:1a1PP4foo1yyF</USR>"
+                   + "<Declaration>func foo1()</Declaration>"
+                   + "<CommentParts>"
+                   + "<Abstract><Para>Protocol P foo1</Para></Abstract>"
+                   + "<Discussion><Note>"
+                   + "<Para>This documentation comment was inherited from <codeVoice>P</codeVoice>.</Para>"
+                   + "</Note></Discussion>"
+                   + "</CommentParts>"
+                   + "</Function>")
```
ahoppen: Would it make sense to make this a multi-line string literal and escape the lines using `\`. Ie something like


```suggestion
                   """
                   <Function file="\(path)" line="3" column="8">\
                   <Name>foo1()</Name>\
                   ...
```

Sources/SwiftSourceKitPlugin/ASTCompletion/CompletionSession.swift
a7medev: sourcekitd doesn't keep full documentation comments in-memory, it prints them in a temporary buffer and passes a copy through `strdup` that clients should free (hence the name is suffixed with `_copy`).

Sources/SwiftSourceKitPlugin/CompletionProvider.swift
a7medev: Do you have a specific flow in mind? AFAIK this doesn't happen in code completion since documentation is only shown when the completion item is selected in the completions list.

Tests/SwiftSourceKitPluginTests/SwiftSourceKitPluginTests.swift
a7medev: That looks nicer indeed, done. ✅

Sources/Csourcekitd/include/CodeCompletionSwiftInterop.h
a7medev: Done ✅

Sources/SourceKitD/sourcekitd_functions.swift
a7medev: Makes sense, done. ✅

Sources/SwiftSourceKitPlugin/ASTCompletion/CompletionSession.swift
ahoppen: If we pass ownership to the client, I think `completion_item_get_doc_full_copy` should return the pointer. E.g. like `swiftide_completion_result_description_copy`.

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
```diff
@@ -574,13 +574,23 @@ class CodeCompletionSession {
...
+         let docString = documentationString(from: response, sourcekitd: sourcekitd) {
         item.documentation = .markupContent(MarkupContent(kind: .markdown, value: docString))
       }
     }
     return item
   }
 
+  private static func documentationString(from response: SKDResponseDictionary, sourcekitd: SourceKitD) -> String? {
+    if let docFullAsXML: String = response[sourcekitd.keys.docFullAsXML] {
+      return try? xmlDocumentationToMarkdown(docFullAsXML)
```
ahoppen: Sorry, I should have noticed this before: We generally prefer to use `orLog` over `try?` because it means that the error will get logged instead of silently swallowing it, which helps debugging.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -1253,6 +1310,39 @@ private func countFs(_ response: CompletionList) -> Int {
...
+  line: UInt = #line
+) async throws {
+  let supportsFullDocumentation = try await sourcekitdSupportsFullDocumentation()
+  let expected = supportsFullDocumentation ? expectedFull : expectedBrief
+
+  assertMarkdown(documentation: documentation, expected: expected, file: file, line: line)
+}
+
+private func sourcekitdSupportsFullDocumentation() async throws -> Bool {
+  let sourcekitdPath = await ToolchainRegistry.forTesting.default!.sourcekitd!
```
ahoppen: Could you use `XCTUnwrap` instead of `!` to avoid crashing the test execution if for some reason one of these values is `nil`.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -1253,6 +1310,39 @@ private func countFs(_ response: CompletionList) -> Int {
...
+/// Asserts that documentation matches the expected values based on whether full documentation is supported in sourcekitd or not.
+private func assertDocumentation(
+  documentation: StringOrMarkupContent?,
+  expectedBrief: String,
+  expectedFull: String,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) async throws {
+  let supportsFullDocumentation = try await sourcekitdSupportsFullDocumentation()
+  let expected = supportsFullDocumentation ? expectedFull : expectedBrief
```
ahoppen: I’m a little concerned with executing a different test path depending on whether sourcekitd supports full documentation because in Swift CI I imagine the sourcekitd will always support full documentation which means that the `expectedBrief` case will not get tested in Swift CI.

Instead, I would prefer to add a function to `SkipUnless` and skip the tests if sourcekitd doesn’t support full documentation.

Same comment for `assertDocumentation` for the sourcekitd plugin.

a7medev: My only concern is that tests like `SwiftCompletionTests.testCompletionBasic` will get skipped in environments without full documentation support.
This might be accepted since it's still exercised in Swift CI anyway.

What do you think? Should we just assume full documentation exists with a `SkipUnless` or do you prefer duplicating the tests each skipped depending on full documentation support to ensure brief documentation is still tested if full documentation isn't supported?

Sources/SwiftSourceKitPlugin/ASTCompletion/CompletionSession.swift
a7medev: I just modified `completion_item_get_doc_full` to not `strdup` the string and pass it as is to the handler, and SourceKit-LSP will copy it into a Swift `String` anyway and it doesn't have to manually call `free`.

Can you please recheck?

Sources/SourceKitLSP/Swift/CodeCompletionSession.swift
a7medev: Done ✅

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
a7medev: Removed now that we're using `SkipUnless` ✅

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
a7medev: I assumed we'd just skip them if full documentation is not supported with no tests for brief documentation alone (avoiding duplication & relying on the fact that Swift CI is going to have full documentation support anyway) to speedup review.

If you'd prefer another approach please let me know.

ahoppen: CI should be able to execute all tests. Otherwise tests can break and you don’t notice until you run them locally. So, since CI will only able to test full documentation and the brief documentation part is being phased out, we should just test full documentation support.

Sources/SwiftSourceKitPlugin/CompletionProvider.swift
ahoppen: For example, you could think about a UI where the brief documentation is displayed in the completion list (similar to how Xcode shows documentation) and only showing the full documentation when the user requests it.

a7medev: I assume the plugin is used in SourceKit-LSP only, no? If so, then we don't have to return the brief documentation now as we don't need it. We can always add it later if we found a use case for it.

If not, it'd probably make sense to return both yes, especially if it's used by Xcode which already has this use case.

hamishknight: The plugin can potentially be used by any SourceKit client yes, so I think we should return both by default. We can always add options to the request for cases where the client only wants the brief or full doc

a7medev: Now it makes sense, thanks for the clarification @ahoppen and @hamishknight 🙏🏼 

I've changed the request to return both now.

a7medev:
I have added one more functionality change.
In the initial implementation, I converted XML documentation to markdown using the same existing flow which includes the declaration as part of the documentation.

I found that existing LSP implementations (mainly Clang and TypeScript) don't show the declaration itself as part of the full documentation in completion (contrary to the hover request!) and I think the reason for this is that the completion item already describes the declaration and typically the popup showing the documentation is small so it makes sense to show the documentation comment right away.

@ahoppen @hamishknight Can you please recheck? 🙏🏼

ahoppen (approved):
Looks great to me. Thank you 🚀

ahoppen:
@swift-ci Please test

ahoppen:
CI failed because Swift CI unconditionally enables all tests that are guarded by `skipUnlessSupportedByToolchain` (we don’t want to accidentally leave tests disabled in CI due to configuration issues). I suggest that we wait until https://github.com/swiftlang/swift/pull/82464 is merged before merging this. Is that alright with you?

a7medev:
@ahoppen Yes, I think that makes sense.

I wonder where in the console output did you find this though 😅

**Edit:** I found it by downloading the plain text output and using `grep` to filter it out. Let me know if there's a more efficient way to do that though 😅

ahoppen:
No, I always download the log files as well.

bnbarham (approved):
Woo! This is awesome, thanks @a7medev!

> As of now, SourceKit-LSP doesn't cache completion item documentation either, should we introduce a new full documentation cache (e.g. using LRUCache)?

It's probably worth merging this, seeing how it performs, and only adding the cache if we need it. If we do then it'll be a little more than just caching them as they're retrieved, since if that's too slow then getting them in the first place is as well 😅 (in which case we might need to pre-fetch a bunch).

a7medev:
We've found that the XML to markdown implementation has some problems.

Consider this example:
```swift
/// Adds two integers together
/// A really interesting function, no?
///
/// Usage:
/// ```swift
/// add(x: 14, y: 22)
/// ```
///
/// - Parameters:
///   - x: First arg
///   - y: Second arg
/// - Returns: Sum
func add(x: Int, y: Int) -> Int {
    return x + y
}
```

The full documentation for add is shown like this in VS Code:
Full documentation using XML to markdown conversion

this has some issues:
1. Code blocks have line numbers and extra spacing before each line (prepended as part of the code to each line which can also cause issues with syntax highlighting) and there’s always an empty numbered line at the end which makes it even weirder. These line numbers are being added since the full XML output adds code between `<zCodeLineNumbered>` tags. When I checked Xcode’s documentation output, it didn’t actually include any line numbers at all, which I believe looks a lot nicer.
2. Section headings are a bit large considering the area reserved for documentation output.
3. The current XML to markdown implementation is incomplete (see #1014) which led to switching to the raw comment for the `hover` LSP request in #1091.

If we just output the raw comment instead, it looks like this. Which looks a lot nicer.
image

The downside of course is losing the structured format that the XML-based comment has so we should ultimately switch back to using the XML to markdown approach after improving it.

I agreed with @hamishknight on sticking to the raw comment for now to follow what `hover` is doing and to not block this on improving XML to markdown output.

@hamishknight @ahoppen @bnbarham Can you please re-review the PR after these changes? 🙏🏼

ahoppen:
I agree that we should render the raw Markdown instead of converting Markdown to XML only to convert it back to Markdown for the same reason as I made the change in https://github.com/swiftlang/sourcekit-lsp/pull/1091.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -386,6 +393,18 @@ package actor SkipUnless {
...
+  private static func getSourceKitD() async throws -> SourceKitD {
+    guard let sourcekitdPath = await ToolchainRegistry.forTesting.default?.sourcekitd else {
+      throw GenericError("Could not find SourceKitD")
+    }
+    let sourcekitd = try await SourceKitD.getOrCreate(
+      dylibPath: sourcekitdPath,
+      pluginPaths: try sourceKitPluginPaths
+    )
+
+    return sourcekitd
```
ahoppen, line 444: Can be simplified to

```suggestion
    return try await SourceKitD.getOrCreate(
      dylibPath: sourcekitdPath,
      pluginPaths: try sourceKitPluginPaths
    )
```

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -1154,6 +1186,7 @@ final class SwiftCompletionTests: XCTestCase {
 
   func testCompletionItemResolve() async throws {
     try await SkipUnless.sourcekitdSupportsPlugin()
+    try await SkipUnless.sourcekitdSupportsFullDocumentationInCompletion()
```
ahoppen, line 1189: Shouldn’t this case pass even if sourcekitd doesn’t support full documentation in completion? You didn’t actually change the behavior or am I missing something.

a7medev: It would pass since the brief documentation is equal to the full documentation in this test case but I don't think this is something we need to maintain since we can change the formatting for full/brief documentation or update the doc comment in the test for some reason which would in turn break the test if we didn't have full documentation support in sourcekitd.

ahoppen: Oh, I see. Makes sense. Thanks for explaining. 👍🏽

a7medev:
@ahoppen @hamishknight I've rebased the PR on `main` since it had a conflict. Can you please check and run CI since https://github.com/swiftlang/swift/pull/82464/ has been merged? 🙏🏼

hamishknight:
@swift-ci please test

ahoppen (approved):
Very cool to see this landing.

PR #2216, opened 2025-07-22
[DocC Live Preview] Support parameter and return type disambiguations

matthewbastien:
Adds support for parameter and return type disambiguations to `textDocument/doccDocumentation` requests. See [the docc documentation for more info](https://www.swift.org/documentation/docc/linking-to-symbols-and-other-content#Ambiguous-Symbol-Links).

The information for parameter and return types of functions does not exist in the index. So, we have to fetch the symbol graph from sourcekitd in order to get all of the info we need. This adds some overhead, but not enough to cause any significant performance impacts in my testing.

### DIFF
--- a/Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
+++ b/Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
@@ -10,10 +10,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
 package import IndexStoreDB
 import SKLogging
 import SemanticIndex
-@_spi(LinkCompletion) import SwiftDocC
+@preconcurrency @_spi(LinkCompletion) import SwiftDocC
+import SwiftExtensions
+import SymbolKit
 
 extension CheckedIndex {
   /// Find a `SymbolOccurrence` that is considered the primary definition of the symbol with the given `DocCSymbolLink`.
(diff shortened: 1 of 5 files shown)


matthewbastien:
Depends on PR from swift-docc: https://github.com/swiftlang/swift-docc/pull/1253

matthewbastien:
https://github.com/swiftlang/swift-docc/pull/1253

@swift-ci please test

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
```diff
@@ -10,65 +10,82 @@
...
   /// the same result every time.
   package func primaryDefinitionOrDeclarationOccurrence(
-    ofDocCSymbolLink symbolLink: DocCSymbolLink
-  ) -> SymbolOccurrence? {
-    var components = symbolLink.components
-    guard components.count > 0 else {
+    ofDocCSymbolLink symbolLink: DocCSymbolLink,
+    fetchSymbolGraph: @Sendable (_: SymbolLocation) async throws -> String?
+  ) async throws -> SymbolOccurrence? {
+    guard !symbolLink.components.isEmpty else {
```
ahoppen: Would it make sense to avoid the force unwrap below by making this

```suggestion
guard let topLevelSymbolName = symbolLink.components.last.name else {
```

Sources/SourceKitLSP/Documentation/DoccDocumentationHandler.swift
```diff
@@ -59,7 +59,30 @@ extension DocumentationLanguageService {
...
+          let symbolOccurrence = try await index.primaryDefinitionOrDeclarationOccurrence(
+            ofDocCSymbolLink: symbolLink,
+            fetchSymbolGraph: { location in
+              guard let symbolWorkspace = try await workspaceForDocument(uri: location.documentUri),
+                let languageService = try await languageService(for: location.documentUri, .swift, in: symbolWorkspace)
+                  as? SwiftLanguageService
+              else {
+                throw ResponseError.internalError("Unable to find Swift language service for \(location.documentUri)")
+              }
+              return try await languageService.withSnapshotFromDiskOpenedInSourcekitd(
```
ahoppen, line 71: If I understand correctly, we run multiple cursor info requests for a single docc request, right? Wit this implementation we need to open a new sourcekitd document for every request, which means that we also need to build as many ASTs, which is really the expensive part for the cursor info request. Would it be possible to only open each document once?

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
```diff
@@ -10,65 +10,82 @@
...
+  }
+
+  /// Find the DocCSymbolLink for a given symbol USR.
+  ///
+  /// - Parameters:
+  ///   - usr: The symbol USR to find in the index.
+  ///   - fetchSymbolGraph: Callback that returns a SymbolGraph for a given SymbolLocation
+  package func doccSymbolInformation(
+    ofUSR usr: String,
+    fetchSymbolGraph: (_: SymbolLocation) async throws -> String?
```
ahoppen: I don’t think that specifying `_: ` provides any value. It’s valuable if you want to name the parameter (eg. `_ location: SymbolLocation`) but otherwise, you can just make this `(SymbolLocation)`. Same in at least one more spot.

Sources/SourceKitLSP/Documentation/DoccDocumentationHandler.swift
```diff
@@ -59,7 +59,30 @@ extension DocumentationLanguageService {
...
+                uri: location.documentUri,
+                fallbackSettingsAfterTimeout: false
+              ) {
+                (snapshot, compileCommand) in
```
ahoppen:
```suggestion
              ) { (snapshot, compileCommand) in
```

Sources/SourceKitLSP/Swift/DoccDocumentation.swift
```diff
@@ -73,7 +74,22 @@ extension SwiftLanguageService {
...
         documentationManager: documentationManager,
         catalogURL: catalogURL,
-        for: symbolUSR
+        for: symbolUSR,
+        fetchSymbolGraph: { symbolLocation in
+          try await withSnapshotFromDiskOpenedInSourcekitd(
+            uri: symbolLocation.documentUri,
+            fallbackSettingsAfterTimeout: false
+          ) {
+            (snapshot, compileCommand) in
```
ahoppen:
```suggestion
          ) { (snapshot, compileCommand) in
```

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
```diff
@@ -10,65 +10,82 @@
...
+    var symbolOccurrence: SymbolOccurrence = topLevelSymbolOccurrence
+    while let parentSymbolOccurrence = symbolOccurrence.parent(self) {
+      symbols.insert(parentSymbolOccurrence, at: 0)
+      symbolOccurrence = parentSymbolOccurrence
+    }
+    // Fetch symbol information from the symbol graph
+    var components = [DocCSymbolInformation.Component(fromModuleName: moduleName)]
+    for symbolOccurence in symbols {
+      guard let rawSymbolGraph = try await fetchSymbolGraph(symbolOccurence.location) else {
+        return nil
```
ahoppen: Would it be worthwhile throwing errors here explaining what went wrong instead of returning `nil` in this method? Just thinking that it might make debugging easier / make it possible to see what’s going on in the logs.

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
```diff
@@ -10,65 +10,82 @@
...
       }
-      // If we have exactly one component left, check to see if it's the module name
-      if components.count == 1 {
-        let lastComponent = components.removeLast()
-        guard lastComponent.name == topLevelSymbolOccurrence.location.moduleName else {
-          return false
-        }
+    }
+    // Ensure that this is deterministic by sorting the results
+    result.sort()
```
ahoppen, line 48: Could this be just a `filter`?

```suggestion
    let result = try await topLevelSymbolOccurrences.asyncFilter { occurrence in
      let info = try await doccSymbolInformation(ofUSR: occurrence.symbol.usr, fetchSymbolGraph: fetchSymbolGraph)
      return info?.matches(symbolLink) ?? false
    }.sorted()
```

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
matthewbastien: Good point. Added errors for these cases.

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
matthewbastien: I wanted to use `asyncFilter(_:)`, but got the following error when I tried it:

```
Capture of 'self' with non-Sendable type 'CheckedIndex' in a '@Sendable' closure
```

and it doesn't look like we can easily make `CheckedIndex` conform to `Sendable`.

Sources/SourceKitLSP/Documentation/DoccDocumentationHandler.swift
matthewbastien: This is correct. I can cache the snapshots instead of throwing them away, but that will require a rework of `withSnapshotFromDiskOpenedInSourcekitd(uri:fallbackSettingsAfterTimeout:body:)`. I'll see how feasible that is.

FWIW I don't expect there to be too many lookups, but it'll only get worse with the size of the project and the number of colliding symbol names...

Sources/DocCDocumentation/IndexStoreDB+Extensions.swift
ahoppen: Looks like none of the functions in `Sequence+AsyncMap.swift` need `@_inheritActorContext` and `@Sendable`. Removing those also allows you to use `asyncFilter` here. Not sure why I added those annotations. But we can also make that a follow-up change.

Sources/SourceKitLSP/Documentation/DoccDocumentationHandler.swift
ahoppen: I’m happy to take this change as-is as well and address it in a follow-up PR (or file an issue if you don’t think that you’ll get to it soon).

ahoppen:
@swift-ci Please test

bnbarham:
https://github.com/swiftlang/swift-docc/pull/1253

@swift-ci please test

matthewbastien:
Thanks for the review! I'll submit a follow-up PR to address the two remaining comments.

PR #2218, opened 2025-07-24
Fix algorithmic bug causing exponential blowup in dependency analysis

josh-arnold-1:
### Problem

The `targetDepthsAndDependents` function in `BuildSystemManager` was taking ~24 seconds to process 636 build targets, causing significant delays during build system initialization and target change notifications.

This should complete in well under a second for this number of targets.

### Root Cause

The issue was an algorithmic bug in the dependency traversal logic. The function was checking the wrong variable when deciding whether to add dependencies to the work queue:

```swift
// ❌ BEFORE: Checking current target instead of dependency
if depths[target, default: 0] < depth + 1 {
  worksList.append((dependency, depth + 1))
}
```

This caused dependencies to be added to the work list even when they had already been processed with adequate depth, leading to exponential time complexity as the same nodes were visited repeatedly.

### Solution

Fixed the condition to check the dependency's depth instead of the current target's depth:

```swift
// ✅ AFTER: Correctly checking dependency depth
if depths[dependency, default: 0] < depth + 1 {
  worksList.append((dependency, depth + 1))
}
```

This ensures proper memoization and restores the intended O(V+E) linear time complexity.

### Impact

- **Performance**: Reduces target depth computation from ~24 seconds to well under 1 second
- **User Experience**: Significantly faster build system initialization and target change handling
- **Algorithmic Complexity**: Fixes exponential blowup, restoring linear time complexity

### Testing

The fix maintains the same algorithmic correctness while dramatically improving performance. The function still correctly computes:
- Target depths (distance from root targets)
- Dependent relationships (reverse dependency mapping)

But now does so efficiently without redundant work.

---

**Files Changed:**
- `Sources/BuildSystemIntegration/BuildSystemManager.swift` - Fixed variable name in depth checking condition

### DIFF
--- a/Sources/BuildSystemIntegration/BuildSystemManager.swift
+++ b/Sources/BuildSystemIntegration/BuildSystemManager.swift
@@ -1071,7 +1071,7 @@ package actor BuildSystemManager: QueueBasedMessageHandler {
         dependents[dependency, default: []].insert(target)
         // Check if we have already recorded this target with a greater depth, in which case visiting it again will
         // not increase its depth or any of its children.
-        if depths[target, default: 0] < depth + 1 {
+        if depths[dependency, default: 0] < depth + 1 {
           worksList.append((dependency, depth + 1))
         }
       }


bnbarham:
@swift-ci please test

bnbarham (approved):
Oof, nice find. Thanks @josh-arnold-1!

ahoppen:
Great catch, thanks for debugging this and finding the fix! 🤩

PR #2223, opened 2025-07-29
Fix issue that cause semantic tokens to be out-of-sync in neovim after a document changes on disk

ahoppen:
When neovim detects a change of the document on-disk (eg. caused by git operations). It closes the document and re-opens it with the same document version but different contents. We didn’t clear the swift-syntax tree of a document when it was closed, so we re-used the old syntax tree for the re-opened document. Ensure we clear the syntax tree when the document is closed so we build a new tree when it is re-opened.


ahoppen:
@swift-ci Please test

MaxDesiatov (approved):

hamishknight (approved):

PR #2226, opened 2025-07-31
[DocC Live Preview] Cache on-disk snapshots opened in sourcekitd

matthewbastien:
Removed `SwiftLanguageService.withSnapshotFromDiskOpenedInSourcekitd()` in favor of a new `OnDiskDocumentManager` actor that will cache all snapshots opened for the duration of a `textDocument/doccDocumentation` request. This will prevent the syntax tree from being parsed multiple times for the same document in a single request.

Follow up to #2216


matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows

Sources/SourceKitLSP/Swift/SymbolGraphCache.swift
```diff
@@ -0,0 +1,134 @@
...
+    return swiftLanguageService
+  }
+
+  deinit {
+    guard let sourceKitLSPServer else {
+      return
+    }
+
+    let documentsToClose = openSnapshots.values
+    Task {
```
ahoppen: Launching a task here means that the documents could be closed in parallel while another request is handled, which would be very hard to debug if this ends up creating some kind of data races in sourcekitd. I think I’d prefer a pattern where you have something like the following and then explicitly close the documents at end of the cache’s lifetime.

```swift
func withSymbolGraphCache<T>(_ body: (SymbolGraphCache) async throws -> T) async throws -> T
```

I recently wrote the following for another project, which might be helpful for such a `with*` implementation (or generally something that I think might be useful to have in `SwiftExtensions`).

```swift
/// Run `body` and always ensure that `cleanup` gets run, independently of whether `body` threw an error or returned a
/// value.
package func run<T>(
  _ body: () async throws -> T,
  cleanup: () async -> ()
) async throws -> T {
  do {
    let result = try await body()
    await cleanup()
    return result
  } catch {
    await cleanup()
    throw error
  }
}
```

Sources/SourceKitLSP/Swift/SymbolGraphCache.swift
```diff
@@ -0,0 +1,134 @@
...
+  /// Open a unique dummy document in sourcekitd that has the contents of the file on disk for uri, but an arbitrary
+  /// URI which doesn't exist on disk. Return the symbol graph from sourcekitd.
+  ///
+  /// The document will be retained until ``DocCSymbolGraphCache`` is de-initialized. This will avoid parsing the same
+  /// document multiple times if more than one symbol needs to be looked up.
+  ///
+  /// - Parameter symbolLocation: The location of a symbol to find the symbol graph for.
+  /// - Returns: The symbol graph for this location, if any.
+  func fetchSymbolGraph(at symbolLocation: SymbolLocation) async throws -> String? {
+    let swiftLanguageService = try await swiftLanguageService(for: symbolLocation.documentUri)
```
ahoppen: I think we’re never re-using an open document, are we? I don’t even think that we ever register open documents.

Sources/SourceKitLSP/Swift/SymbolGraphCache.swift
```diff
@@ -0,0 +1,134 @@
...
+import SKUtilities
+import SwiftExtensions
+
+/// A cache of symbol graphs and their associated snapshots opened in sourcekitd. Any opened documents will be
+/// closed when the cache is de-initialized.
+///
+/// Used by `textDocument/doccDocumentation` requests to retrieve symbol graphs for files that are not currently
+/// open in the editor. This allows for retrieving multiple symbol graphs from the same file without having
+/// to re-open and parse the syntax tree every time.
+actor SymbolGraphCache: Sendable {
```
ahoppen: Would something like `OnDiskDocumentManager` be a more appropriate name because this type neither really caches values (it manages document lifetimes), nor does it really cache symbol graphs but it keeps track of sourcekitd documents.

matthewbastien: Renamed to `OnDiskDocumentManager`

Sources/SourceKitLSP/Swift/SymbolGraphCache.swift
matthewbastien: I must have slipped into a coma or something while writing this... Good catch!

Sources/SourceKitLSP/Swift/SymbolGraphCache.swift
matthewbastien: Removed the de-initializer in favor of a `closeAllDocuments()` function for cleanup. I also extended `SourceKitLSPServer` to add a `withOnDiskDocumentManager()` which will perform the cleanup automatically.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test Windows

Sources/SourceKitLSP/Swift/OnDiskDocumentManager.swift
```diff
@@ -0,0 +1,169 @@
...
+  ///
+  /// The document will be retained until ``closeAllDocuments()`` is called. This will avoid parsing the same
+  /// document multiple times if more than one symbol needs to be looked up.
+  ///
+  /// - Parameter symbolLocation: The location of the document that will be opened.
+  func openDocumentSnapshot(
+    at symbolLocation: SymbolLocation
+  ) async throws -> (snapshot: DocumentSnapshot, patchedCompileCommand: SwiftCompileCommand?) {
+    try await openDocumentSnapshot(for: symbolLocation.documentUri)
+  }
```
ahoppen: Is it really worth to have this method if we can just add `.documentUri` wherever it is called?

Sources/SourceKitLSP/Swift/OnDiskDocumentManager.swift
```diff
@@ -0,0 +1,169 @@
...
+    guard let cachedSnapshot = openSnapshots[uri] else {
+      let languageService = try await swiftLanguageService(for: uri)
+      let snapshot = try await languageService.openSnapshotFromDiskOpenedInSourcekitd(
+        uri: uri,
+        fallbackSettingsAfterTimeout: false
+      )
+      openSnapshots[uri] = snapshot
+      return snapshot
+    }
+    return cachedSnapshot
```
ahoppen: I would flip this to an `if` so that the majority of the code is not indented.

```suggestion
    if let cachedSnapshot = openSnapshots[uri] else {
      return cachedSnapshot
    }
    let languageService = try await swiftLanguageService(for: uri)
    let snapshot = try await languageService.openSnapshotFromDiskOpenedInSourcekitd(
      uri: uri,
      fallbackSettingsAfterTimeout: false
    )
    openSnapshots[uri] = snapshot
    return snapshot
```

Sources/SourceKitLSP/Swift/OnDiskDocumentManager.swift
```diff
@@ -0,0 +1,169 @@
...
+/// Caches document snapshots from disk opened in sourcekitd.
+///
+/// Used by `textDocument/doccDocumentation` requests to retrieve symbol graphs for files that are not currently
+/// open in the editor. This allows for retrieving multiple symbol graphs from the same file without having
+/// to re-open and parse the syntax tree every time.
+actor OnDiskDocumentManager: Sendable {
+  private weak var sourceKitLSPServer: SourceKitLSPServer?
+  private var openSnapshots: [DocumentURI: (snapshot: DocumentSnapshot, patchedCompileCommand: SwiftCompileCommand?)]
+
+  init(sourceKitLSPServer: SourceKitLSPServer) {
```
ahoppen: Should this be `fileprivate` to ensure that the `OnDiskDocumentManager` is always created using `withOnDiskDocumentManager`?

matthewbastien: Good point. Done.

Sources/SourceKitLSP/Swift/OnDiskDocumentManager.swift
matthewbastien: Done.

Sources/SourceKitLSP/Swift/OnDiskDocumentManager.swift
matthewbastien: Done.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test Windows

matthewbastien:
Rebasing on top of main meant that I needed to rearrange a few things:

- Moved `withOnDiskDocumentManager(_:)` and `OnDiskDocumentManager` to `SourceKitLSPServer`
- Added `openDocumentOnDisk(snapshot:originalFile:)` and `closeDocumentOnDisk(snapshot:)` to `LanguageService` so that `SwiftLanguageService` can manage opening and closing the dummy snapshots in sourcekitd.
- Renamed `symbolGraph(forOnDiskContentsOf:at:)` to `symbolGraphForDocumentOnDisk(at:manager:)` which accepts an `OnDiskDocumentManager` that it will use to retrieve snapshots.

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

triceviola43-hue:
Graphic Web Design Consultant
383706149
triceviola43-hue
IMG_0016
IMG_0020
IMG_0019

Issue #2210, opened 2025-07-17
Do not `fatalError` when syncKind is different than `.incremental`


ahoppen:
Instead of crashing here, we should be reporting an error. I don’t know of any cases hitting this but we generally shouldn’t crash in SourceKit-LSP.

https://github.com/swiftlang/sourcekit-lsp/blob/d18aba71ce3c539ec1474470228ab7fea9b66105/Sources/SourceKitLSP/SourceKitLSPServer.swift#L522

PR #2227, opened 2025-08-03
Replace fatalError with proper error handling for non-incremental

Clemo97:
Fixes issue #2210

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -519,7 +519,6 @@ package actor SourceKitLSPServer {
         syncKind = .incremental
       }
       guard syncKind == .incremental else {
-        logger.error("Received non-incremental update request, which is not implemented")
         throw ResponseError.internalError("non-incremental update not implemented")
       }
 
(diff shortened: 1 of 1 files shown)


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -519,7 +519,8 @@ package actor SourceKitLSPServer {
         syncKind = .incremental
       }
       guard syncKind == .incremental else {
-        fatalError("non-incremental update not implemented")
+        logger.error("Received non-incremental update request, which is not implemented")
```
ahoppen: We don’t need a `logger.error` call here. Throwing the error will log an error from the `orLog` call in line 480.

ahoppen (commented):
Thanks for taking this! One small comment.

Sources/SourceKitLSP/SourceKitLSPServer.swift
Clemo97: Hello, thanks for the feedback.
I've made the necessary changes in the new commit.

ahoppen:
@swift-ci Please test

PR #2228, opened 2025-08-04
Move ClangLanguageService to its own module

ahoppen:
I’m working on moving each of the three language services to its own module, ie. having a separate module for `ClangLanguageService`, `SwiftLanguageService`, and `DocumentationLanguageService`. This should separate concerns for the three languages a little better instead of having them all in the single `SourceKitLSP` module.

The overall design is that the modules for the language services depend on the `SourceKitLSP` module, so the language services need to be registered dynamically in a `LanguageServiceRegistry`. Other than that, a bit of code needed to be moved around but the changes should all be NFCs.

https://github.com/ahoppen/sourcekit-lsp/tree/split-language-services contains the result of the final split. Getting there will probably take another couple of PRs.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/sourcekit-lsp/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ add_executable(sourcekit-lsp
 target_link_libraries(sourcekit-lsp PRIVATE
   BuildServerIntegration
   Diagnose
+  InProcessClient
   LanguageServerProtocol
   LanguageServerProtocolExtensions
   LanguageServerProtocolJSONRPC
```
bnbarham, line 10: Not sure how this worked before, but we're apparently missing a DocCDocumentation here (we have a `package import DocCDocumentation` in `Sources/SourceKitLSP/Workspace.swift`)

ahoppen: That's because we can't build DocC on Windows. All usages of DocC should be guarded by `#if`. Maybe I missed a `#if` now.

bnbarham: Ah, I forgot about that :(

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2229, opened 2025-08-06
Make `BuildSettingsLogger` an instance member of `BuildServerManager`

ahoppen:
Having `BuildSettingsLogger` can be problematic if we launch multple `SourceKitLSPServer` instances in the same process (such as during testing). In that case it could happen that a previous `SourceKitLSPServer` instance logs build settings and then the second instance doesn’t log its build settings (if they are the same) because the shared logger has already logged them.

This issue is resolved by scoping `BuildSettingsLogger` to the lifetime of `BuildServerManager` (ie. one per workspace).


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2231, opened 2025-08-11
Apply exhaustive swift-format configuration from swift-syntax

ahoppen:
Apply the exhaustive swift-format configuration from https://github.com/swiftlang/swift-syntax/pull/3117 to sourcekit-lsp. Also apply all automatic formattings.

From the linter rules that needed manual fixing (addressed those in a separate commit), I have two findings, would like to hear other opinions here as well:
- I think there are valid calls with mixing inline and trailing closures (notably `Debouncer`), so I’m considering whether we should disable that rule.
- The `forEach` rule is a little annoying because we have `forEach` on `SKDResponseArray`. But it caught two cases of using `forEach` on arrays, so I think it’s worth keeping.

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -460,18 +460,19 @@ package actor BuildServerManager: QueueBasedMessageHandler {
     // The debounce duration of 500ms was chosen arbitrarily without any measurements.
     self.filesDependenciesUpdatedDebouncer = Debouncer(
       debounceDuration: .milliseconds(500),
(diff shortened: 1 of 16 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):
> I think there are valid calls with mixing inline and trailing closures (notably Debouncer), so I’m considering whether we should disable that rule.

Either seems fine to me

> The forEach rule is a little annoying because we have forEach on SKDResponseArray. But it caught two cases of using forEach on arrays, so I think it’s worth keeping.

Yeah, maybe we should rename the SKD case? `forEachSKD` :)?

PR #2233, opened 2025-08-11
Fix module definition jumping in Swift interface files

aelam:
# Fix module definition jumping for submodules and Swift interface files

### 1. Submodule Import Navigation Issues
- **Foundation.NSArray**: Clicking on `NSArray` in `import Foundation.NSArray` fails to jump to definition
- **Darwin Submodules**: Clicking on `sysdir` in `import Darwin.sysdir` fails to jump to definition  
- **Root Cause**: SourceKit-LSP cannot properly handle module names containing dots (module.submodule)

### 2. Swift Interface File Navigation Problems
- **Generated Interface Navigation**: Command+Click navigation functionality is broken within generated `.swiftinterface` files
- **Multiple Interface Caching**: Repeatedly opening the same module interface generates multiple duplicate files
- **Language Service Mapping**: Interface files cannot be properly mapped to corresponding language services

### 1. Enhanced Module Symbol Processing

**File**: `Sources/SourceKitLSP/SourceKitLSPServer.swift`

#### Module Name Parsing Logic
```swift
// Enhanced module symbol processing
if symbol.kind == .module {
  let moduleName: String
  let groupName: String?
  
  if let systemModule = symbol.systemModule {
    // Prefer using systemModule information
    moduleName = systemModule.moduleName
    groupName = systemModule.groupName
  } else if let name = symbol.name {
    // Handle module names containing dots like "Darwin.sysdir"
    if name.contains(".") {
      let components = name.split(separator: ".", maxSplits: 1)
      moduleName = String(components[0])
      groupName = components.count == 2 ? String(components[1]) : nil
    } else {
      moduleName = name
      groupName = nil
    }
  }
}
```

#### System Module Navigation
```swift
// System module navigation handling
if symbol.isSystem ?? false, let systemModule = symbol.systemModule {
  return try await self.definitionInInterface(
    moduleName: systemModule.moduleName,
    groupName: systemModule.groupName,
    symbolUSR: symbol.usr,
    originatorUri: uri,
    languageService: languageService
  )
}
```

#### Generated Interface Manager Enhancement
**File**: `Sources/SourceKitLSP/Swift/GeneratedInterfaceManager.swift`

```swift
// Improved interface file caching
private var interfaceDocuments: [InterfaceKey: GeneratedInterfaceDocument] = [:]

// Avoid duplicate generation of interface files for the same module
struct InterfaceKey: Hashable {
  let moduleName: String
  let groupName: String?
  let buildSettings: String
}
```

#### Language Service Mapping Fix  
**File**: `Sources/SourceKitLSP/Workspace.swift`

```swift
// Fix language service mapping for interface files
func setDocumentService(_ service: DocumentService, for uri: DocumentURI) {
  let key = buildSettingsFile ?? uri
  documentService[key] = service
}
```

### 3. Interface File Recognition
**File**: `Sources/SourceKitLSP/Swift/OpenInterface.swift`

```swift
// Recognize .swiftinterface file extensions
private static let swiftInterfaceExtensions: Set<String> = ["swiftinterface"]

var isSwiftInterface: Bool {
  swiftInterfaceExtensions.contains(pathExtension.lowercased())
}
```

### 1. Foundation Submodules
✅ `import Foundation.NSArray` → Correctly jumps to NSArray interface  
✅ `import Foundation.NSURL` → Correctly jumps to NSURL interface

### 2. Darwin Submodules  
✅ `import Darwin.sysdir` → Correctly jumps to sysdir interface  
✅ `import Darwin.uuid` → Correctly jumps to uuid interface  
✅ `import Darwin.POSIX` → Correctly jumps to POSIX interface

### 3. Swift Interface Navigation
✅ Command+Click navigation works properly within generated `.swiftinterface` files  
✅ Avoids duplicate generation of interface files for the same module  
✅ Symbols in interface files can correctly jump to definitions

### Key Improvements

1. **Module Name Parsing**: Properly handles `module.submodule` format import statements
2. **System Module Support**: Uses `systemModule` information for accurate module navigation  
3. **Interface File Caching**: Avoids duplicate generation, improves performance
4. **Language Service Mapping**: Ensures interface files can be properly mapped to language services
5. **File Type Recognition**: Correctly identifies `.swiftinterface` file types

### Backward Compatibility

- ✅ Maintains normal functionality of existing import statements
- ✅ Does not affect navigation behavior of non-system modules
- ✅ Compatible with existing interface generation mechanisms

### Feature Improvements
1. **Submodule Navigation**: Full support for Foundation and Darwin submodule navigation
2. **Interface File Navigation**: Swift interface file navigation functionality restored
3. **Performance**: Reduced duplicate interface file generation, improved cache efficiency
4. **Stability**: Fixed interface file language service mapping issues

## Related Issues

- Fix Foundation.NSArray and other submodule import jumping issues
- Fix Darwin.sysdir and other system submodule navigation
- Resolve navigation failures within generated Swift interface files
- Optimize interface file caching mechanism to avoid duplicate generation

## Testing Recommendations

It is recommended to test in the following scenarios:
1. Various Foundation submodule import jumping
2. Darwin system submodule import jumping  
3. Navigation within generated Swift interface files
4. Caching behavior when opening the same module interface multiple times
5. Ensure normal module imports are not affected

symbols

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1857,10 +1857,24 @@ extension SourceKitLSPServer {
     languageService: LanguageService
   ) async throws -> [Location] {
     // If this symbol is a module then generate a textual interface
-    if symbol.kind == .module, let name = symbol.name {
+    if symbol.kind == .module {
+      // For module symbols, prefer using systemModule information if available
+      let moduleName: String
+      let groupName: String?
+
+      if let systemModule = symbol.systemModule {
+        moduleName = systemModule.moduleName
+        groupName = systemModule.groupName
+      } else if let name = symbol.name {
+        moduleName = name
+        groupName = nil
+      } else {
+        return []
+      }
+
       let interfaceLocation = try await self.definitionInInterface(
-        moduleName: name,
-        groupName: nil,
+        moduleName: moduleName,
+        groupName: groupName,
         symbolUSR: nil,
         originatorUri: uri,
         languageService: languageService
@@ -2058,9 +2072,12 @@ extension SourceKitLSPServer {
     originatorUri: DocumentURI,
     languageService: LanguageService
   ) async throws -> Location {
+    // Let openGeneratedInterface handle all the logic, including checking if we're already in the right interface
+    let documentForBuildSettings = originatorUri.buildSettingsFile
+
     guard
       let interfaceDetails = try await languageService.openGeneratedInterface(
-        document: originatorUri,
+        document: documentForBuildSettings,
         moduleName: moduleName,
         groupName: groupName,
         symbolUSR: symbolUSR
--- a/Sources/SwiftLanguageService/OpenInterface.swift
+++ b/Sources/SwiftLanguageService/OpenInterface.swift
@@ -24,9 +24,8 @@ extension SwiftLanguageService {
   ) async throws -> GeneratedInterfaceDetails? {
     // Include build settings context to distinguish different versions/configurations
     let buildSettingsFileHash = "\(abs(document.buildSettingsFile.stringValue.hashValue))"
-    let sourcekitdDocumentName = [moduleName, groupName, buildSettingsFileHash].compactMap(\.self).joined(
-      separator: "."
-    )
+    let sourcekitdDocumentName = [moduleName, groupName, buildSettingsFileHash].compactMap(\.self)
+      .joined(separator: ".")
 
     let urlData = GeneratedInterfaceDocumentURLData(
       moduleName: moduleName,
@@ -46,7 +45,13 @@ extension SwiftLanguageService {
     if self.capabilityRegistry.clientHasExperimentalCapability(GetReferenceDocumentRequest.method) {
       return GeneratedInterfaceDetails(uri: try urlData.uri, position: position)
     }
-    let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent(urlData.displayName)
+    let interfaceFilePath = self.generatedInterfacesPath
+      .appendingPathComponent(buildSettingsFileHash)
+      .appendingPathComponent(urlData.displayName)
+    try FileManager.default.createDirectory(
+      at: interfaceFilePath.deletingLastPathComponent(),
+      withIntermediateDirectories: true
+    )
     try await generatedInterfaceManager.snapshot(of: urlData).text.write(
       to: interfaceFilePath,
       atomically: true,
--- a/Sources/SwiftLanguageService/OpenInterface.swift
+++ b/Sources/SwiftLanguageService/OpenInterface.swift
@@ -22,10 +22,16 @@ extension SwiftLanguageService {
     groupName: String?,
     symbolUSR symbol: String?
   ) async throws -> GeneratedInterfaceDetails? {
+    // Include build settings context to distinguish different versions/configurations
+    let buildSettingsFileHash = "\(abs(document.buildSettingsFile.stringValue.hashValue))"
+    let sourcekitdDocumentName = [moduleName, groupName, buildSettingsFileHash].compactMap(\.self).joined(
+      separator: "."
+    )
+
     let urlData = GeneratedInterfaceDocumentURLData(
       moduleName: moduleName,
       groupName: groupName,
-      sourcekitdDocumentName: "\(moduleName)-\(UUID())",
+      sourcekitdDocumentName: sourcekitdDocumentName,
       primaryFile: document
     )
     let position: Position? =
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
     )
     XCTAssertEqual(diagnostics.fullReport?.items, [])
   }
+
+  func testFoundationImportNavigation() async throws {
+    let testClient = try await TestSourceKitLSPClient(
+      capabilities: ClientCapabilities(experimental: [
+        GetReferenceDocumentRequest.method: .dictionary(["supported": .bool(true)])
+      ])
+    )
+    let uri = DocumentURI(for: .swift)
+
+    let positions = testClient.openDocument(
+      """
+      import 1️⃣Foundation
+      """,
+      uri: uri,
+      language: .swift
+    )
+
+    // Test navigation to Foundation module
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
+    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
+  }
+
+  func testFoundationSubmoduleNavigation() async throws {
+    let testClient = try await TestSourceKitLSPClient(
+      capabilities: ClientCapabilities(experimental: [
+        GetReferenceDocumentRequest.method: .dictionary(["supported": .bool(true)])
+      ])
+    )
+    let uri = DocumentURI(for: .swift)
+
+    let positions = testClient.openDocument(
+      """
+      import 1️⃣Foundation.2️⃣NSAffineTransform
+      """,
+      uri: uri
+    )
+
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
+    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
+
+    // Test navigation to NSAffineTransform
+    let transformDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["2️⃣"])
+    )
+    let transformLocation = try XCTUnwrap(transformDefinition?.locations?.only)
+    // Verify we can identify this as a swiftinterface file
+    XCTAssertTrue(transformLocation.uri.pseudoPath.contains("Foundation.NSAffineTransform.swiftinterface"))
+    XCTAssertTrue(transformLocation.uri.scheme == "sourcekit-lsp")
+  }
 }
 
 private func assertSystemSwiftInterface(
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -341,8 +341,8 @@ final class SwiftInterfaceTests: XCTestCase {
       DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
     )
     let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
-    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
-    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
+    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")  
+    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")  
   }
 
   func testFoundationSubmoduleNavigation() async throws {
@@ -364,8 +364,8 @@ final class SwiftInterfaceTests: XCTestCase {
       DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
     )
     let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
-    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
-    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
+        XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")  
+    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")  
 
     // Test navigation to NSAffineTransform
     let transformDefinition = try await testClient.send(
@@ -373,8 +373,8 @@ final class SwiftInterfaceTests: XCTestCase {
     )
     let transformLocation = try XCTUnwrap(transformDefinition?.locations?.only)
     // Verify we can identify this as a swiftinterface file
-    XCTAssertTrue(transformLocation.uri.pseudoPath.contains("Foundation.NSAffineTransform.swiftinterface"))
-    XCTAssertTrue(transformLocation.uri.scheme == "sourcekit-lsp")
+    XCTAssertEqual(transformLocation.uri.scheme, "sourcekit-lsp")
+    assertContains(transformLocation.uri.pseudoPath, "Foundation.NSAffineTransform.swiftinterface")
   }
 }
 
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -341,8 +341,8 @@ final class SwiftInterfaceTests: XCTestCase {
       DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
     )
     let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
-    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")  
-    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")  
+    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
+    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")
   }
 
   func testFoundationSubmoduleNavigation() async throws {
@@ -364,8 +364,8 @@ final class SwiftInterfaceTests: XCTestCase {
       DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
     )
     let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
-        XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")  
-    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")  
+    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
+    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")
 
     // Test navigation to NSAffineTransform
     let transformDefinition = try await testClient.send(
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -346,6 +346,8 @@ final class SwiftInterfaceTests: XCTestCase {
   }
 
   func testFoundationSubmoduleNavigation() async throws {
+    try SkipUnless.platformIsDarwin("Non-Darwin platforms don't have Foundation submodules")
+
     let testClient = try await TestSourceKitLSPClient(
       capabilities: ClientCapabilities(experimental: [
         GetReferenceDocumentRequest.method: .dictionary(["supported": .bool(true)])
--- a/Sources/SourceKitLSP/GeneratedInterfaceDocumentURLData.swift
+++ b/Sources/SourceKitLSP/GeneratedInterfaceDocumentURLData.swift
@@ -67,7 +67,7 @@ package struct GeneratedInterfaceDocumentURLData: Hashable, ReferenceURLData {
     self.moduleName = moduleName
     self.groupName = groupName
     self.sourcekitdDocumentName = sourcekitdDocumentName
-    self.buildSettingsFrom = primaryFile
+    self.buildSettingsFrom = primaryFile.buildSettingsFile
   }
 
   init(queryItems: [URLQueryItem]) throws {
--- a/Sources/SourceKitLSP/ReferenceDocumentURL.swift
+++ b/Sources/SourceKitLSP/ReferenceDocumentURL.swift
@@ -170,8 +170,4 @@ extension DocumentURI {
 
 package struct ReferenceDocumentURLError: Error, CustomStringConvertible {
   package var description: String
-
-  init(description: String) {
-    self.description = description
-  }
 }


aelam:
@swift-ci Please test

Sources/SourceKitLSP/Swift/GeneratedInterfaceManager.swift
```diff
@@ -92,7 +92,13 @@ actor GeneratedInterfaceManager {
     incrementingRefCount: Bool
   ) async throws -> OpenGeneratedInterfaceDocumentDetails {
     func loadFromCache() -> OpenGeneratedInterfaceDocumentDetails? {
-      guard let cachedIndex = openInterfaces.firstIndex(where: { $0.url == document }) else {
+      // Cache by module name and group name, not the full document data
+      // This allows reuse across different buildSettingsFrom URIs
+      guard let cachedIndex = openInterfaces.firstIndex(where: {
```
ahoppen: I don’t think this is correct. Depending on the build settings used to retrieve the generated interface, the interface might look differently. I can’t come up with a great real-world example right now where the contents differ but a canonical example would be that we can’t load the generated interface if we are missing an include path to one of the module’s dependencies but we can generate the interface if all build settings are present.

But if we continue passing along the same `buildSettingsFrom` file, I don’t think we need this change. Or am I missing something?

Sources/LanguageServerProtocolExtensions/Language+Inference.swift
```diff
@@ -39,7 +39,7 @@ extension Language {
     case "cpp", "cc", "cxx", "hpp": self = .cpp
     case "m": self = .objective_c
     case "mm", "h": self = .objective_cpp
-    case "swift": self = .swift
+    case "swift", "swiftinterface": self = .swift
```
ahoppen: Out of curiosity, why is this needed?

Sources/SourceKitLSP/Swift/OpenInterface.swift
```diff
@@ -21,10 +21,18 @@ extension SwiftLanguageService {
     groupName: String?,
     symbolUSR symbol: String?
   ) async throws -> GeneratedInterfaceDetails? {
+    // Generate a deterministic document name based on module name and group name
+    // This ensures we reuse the same interface for the same module/group combination
+    let documentName = if let groupName {
```
ahoppen: Similar to my comment above, generated interfaces might look different depending on the build settings that were used to generate them, so module name and group name don’t uniquely identify its contents. That’s why we used the UUID-based filename below.

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -165,13 +165,14 @@ extension DocumentURI {
       return referenceDocument.buildSettingsFile
     }
     return self
+  }  /// Convert sourcekit-lsp:// URIs to actual file system paths when possible.
```
ahoppen: Missing newlines before `///`?

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -165,13 +165,14 @@ extension DocumentURI {
...
+  /// For generated Swift interfaces, this returns the path where the interface file
+  /// would be stored in the GeneratedInterfaces directory.
+  var actualFileSystemPath: DocumentURI {
```
ahoppen: I don’t think this is used anywhere, or am I missing something?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -553,7 +553,7 @@ package actor SourceKitLSPServer {
     }
 
     let toolchain = await workspace.buildServerManager.toolchain(
-      for: uri,
+      for: uri.buildSettingsFile,
```
ahoppen: Is this change necessary? I just checked and it looks like `toolchain(for:in:language:)` doesn’t even use `uri` right now.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2051,9 +2065,36 @@ extension SourceKitLSPServer {
     originatorUri: DocumentURI,
     languageService: LanguageService
   ) async throws -> Location {
+    // Check if we're already in the target interface with the same module/group/symbol
+    if let referenceDoc = try? ReferenceDocumentURL(from: originatorUri),
+       case .generatedInterface(let interfaceData) = referenceDoc,
```
ahoppen: I think you can slightly shorten this

```suggestion
    if case .generatedInterface(let interfaceData) = try? ReferenceDocumentURL(from: originatorUri),
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2051,9 +2065,36 @@ extension SourceKitLSPServer {
...
+       interfaceData.moduleName == moduleName && interfaceData.groupName == groupName {
```
ahoppen: Similarly to my other comments about the build settings, we need to check here that we are using build settings from the same original file. But I think as long as we pass along the same build settings file, this shouldn’t cause any issues.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -432,11 +432,12 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
   /// document is returned.
   func setDocumentService(for uri: DocumentURI, _ newLanguageService: any LanguageService) -> LanguageService {
     return documentService.withLock { service in
-      if let languageService = service[uri] {
+      let key = uri.buildSettingsFile
```
ahoppen: Why is this change necessary? The document service should be associated with the URL of the document, not the file that provides its build settings. For example, you could imagine opening the generated interface of a C header (not currently supported in SourceKit-LSP but would be nice to have). In that case the build settings are probably from a C file but the generated interface is handled by the `SwiftLanguageService`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2051,9 +2065,36 @@ extension SourceKitLSPServer {
...
+      
+      // If we have a specific symbol USR, try to find its position in the current interface
+      if let symbolUSR = symbolUSR,
+         let swiftLanguageService = languageService as? SwiftLanguageService {
```
ahoppen: I’m currently trying to unravel the dependency of the `SourceKitLSP` module on `SwiftLanguageService` in https://github.com/swiftlang/sourcekit-lsp/pull/2234, so getting the `SwiftLanguageService` won’t work in the future.

What are the challenges of always going through `openGeneratedInterface`? I can see the following but I might be missing some:
- Every time `openGeneratedInterface` is called, we get a new `sourcekitdDocumentName`. We could fix that by instead of appending the group name and a hash of `document` (or something of that sort).
- If the issue is that we always write a new file if the client doesn’t support reference documents, we could just not write out the file if the contents haven’t changed

Sources/LanguageServerProtocolExtensions/Language+Inference.swift
aelam: If a swiftinterface is opened, itself won't be recoginzed as a swift file mean sourcekit-lsp woundn't work on it

```swift
  package func defaultLanguage(for document: DocumentURI, in target: BuildTargetIdentifier) async -> Language? {
    let languageFromBuildServer = await orLog("Getting source files to determine default language") {
      try await languageInferredFromBuildServer(for: document, in: target)
    }
    return languageFromBuildServer ?? Language(inferredFromFileExtension: document)
  }
```

ahoppen: The function you referenced should only be used for background tasks such as indexing. Since we don’t index generated Swift interfaces, this shouldn’t be necessary. The editor will pass the language that the document should be opened as in the `textDocument/didOpen` notification.

Sources/SourceKitLSP/SourceKitLSPServer.swift
aelam: @ahoppen Thanks for the comment


without this change, here would happen 


if I open defination of `Foundation` in `Foundation.swiftInterace`, it will open another one, repeat, it will open another new one 
even if it's acceptable but it's quite slow (please check the video)

Cause:
```
originalUri is sourcekit-lsp://xxx?buildSettingsFrom=A
A is sourcekit-lsp://xxx?buildSettingsFrom=B
B is sourcekit-lsp://xxx?buildSettingsFrom=C
C is file://
```

this makes the logic quite complicated
1. we have to recursively find the `buildSettingsFrom`
2. originalUri are different for same `swiftinterface`


I think we could assume the interface is the same if the interface was originally opened from a same `buildSettingsFrom` 
because interface opened from A, B  should always be consistent? 

then we could simplify the originalUri as
```
originalUri is sourcekit-lsp://xxx?buildSettingsFrom=C
```

the benefits are
1. we don't need to resursively search
2. caching/creating of `swiftinterface` could be simplified and easy to hit the cache
3. much faster
4. the code change here is not needed any more.




https://github.com/user-attachments/assets/7a2d573e-ef80-466f-b939-6726fd9d9b70

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -122,14 +122,22 @@ package enum ReferenceDocumentURL {
...
 
   var primaryFile: DocumentURI? {
     switch self {
     case .macroExpansion(let data): return data.primaryFile
-    case .generatedInterface(let data): return data.buildSettingsFrom.primaryFile
+    case .generatedInterface(let data):
+      if let referenceDocumentURL = try? ReferenceDocumentURL(from: data.buildSettingsFrom) {
+        return referenceDocumentURL.primaryFile
+      }
+    return data.buildSettingsFrom.primaryFile
```
aelam: check https://github.com/swiftlang/sourcekit-lsp/pull/2233/files#r2269006407

Sources/SourceKitLSP/SourceKitLSPServer.swift
aelam: adding a group might be a good idea 

i imagine there could be different versions of swiftinterface which are used by different modules/files

i think the challenge is how to determine it in the  beginning instead of a uuid
then the cache logic could be simplified

Sources/SourceKitLSP/Swift/GeneratedInterfaceDocumentURLData.swift
```diff
@@ -67,7 +67,14 @@ package struct GeneratedInterfaceDocumentURLData: Hashable, ReferenceURLData {
...
     self.sourcekitdDocumentName = sourcekitdDocumentName
-    self.buildSettingsFrom = primaryFile
+    if let referenceDocumentURL = try? ReferenceDocumentURL(from: primaryFile),
+       let fileURL = referenceDocumentURL.buildSettingsFile.fileURL {
+      self.buildSettingsFrom = DocumentURI(fileURL)
+    } else if let fileURL = primaryFile.fileURL {
+      self.buildSettingsFrom = DocumentURI(fileURL)
+    } else {
+      self.buildSettingsFrom = primaryFile
+    }
```
aelam: @ahoppen 

make sure buildSettingsFrom can be only an actual file otherwise

it could be 
sourcekit-lsp://xxx?buildSettingsFrom=A
A=sourcekit-lsp://xxx?buildSettingsFrom=B

NOTE: this happens if we enable navigation inside swiftinterface in this PR

Sources/SourceKitLSP/Swift/GeneratedInterfaceManager.swift
aelam: you are right

but about same `buildSettingsFrom` file  please check here https://github.com/swiftlang/sourcekit-lsp/pull/2233#discussion_r2269006407

Sources/SourceKitLSP/SourceKitLSPServer.swift
bnbarham: Yeah, IMO when doing a jump to definition from within an interface, the `buildSettingsFrom` should be the original source file rather than the interface you're currently in. IIUC this is what you're saying in:
> then we could simplify the originalUri as
> originalUri is sourcekit-lsp://xxx?buildSettingsFrom=C
? If so, then 👍

Sources/SourceKitLSP/Swift/GeneratedInterfaceDocumentURLData.swift
ahoppen: We have an accessor on `DocumentURI` that should do exactly the same, so you should be able to simplify this to

```suggestion
    self.buildSettingsFrom = primaryFile.buildSettingsFile
```

Sources/SKTestSupport/Utils.swift
```diff
@@ -32,7 +32,7 @@ extension Language {
     case "cpp": self = .cpp
     case "m": self = .objective_c
     case "mm": self = .objective_cpp
-    case "swift": self = .swift
+    case "swift", "swiftinterface": self = .swift
```
ahoppen: Similar to my other comment about the `.swiftinterface` extension, I don’t think we need this here, do we? This only applies to tests and I don’t think we open any `.swiftinterface` files in tests.

Sources/SourceKitLSP/Swift/ReferenceDocumentURL.swift
```diff
@@ -122,14 +122,25 @@ package enum ReferenceDocumentURL {
   var buildSettingsFile: DocumentURI {
     switch self {
     case .macroExpansion(let data): return data.primaryFile
-    case .generatedInterface(let data): return data.buildSettingsFrom
+    case .generatedInterface(let data):
+      if let referenceDocumentURL = try? ReferenceDocumentURL(from: data.buildSettingsFrom),
+         let fileURL = referenceDocumentURL.buildSettingsFile.fileURL {
+        return DocumentURI(fileURL)
+      }
+      return data.buildSettingsFrom
```
ahoppen: If we always set the `buildSettingsFrom` to the original source file, I don’t think we need to transitively walk up the `buildSettingsFile` chain here, right? Same for `primaryFile` below.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -149,7 +153,7 @@ final class SwiftInterfaceTests: XCTestCase {
           public init() {}
         }
         """,
-        "Exec/main.swift": "import 1️⃣MyLibrary",
+        "Exec/main.swift": "import 1️⃣MyLibrary"
```
ahoppen: Our format configuration expects trailing commas for collections with more than one element. Could you run swift-format as described in https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting?

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -192,21 +196,13 @@ final class SwiftInterfaceTests: XCTestCase {
...
       ],
       manifest: """
         let package = Package(
           name: "MyLibrary",
-          targets: [
-            .target(name: "MyLibrary"),
-            .executableTarget(name: "Exec", dependencies: ["MyLibrary"])
-          ]
+          targets: [.target(name: "MyLibrary"), .executableTarget(name: "Exec", dependencies: ["MyLibrary"])]
```
ahoppen: Why are you modifying these existing test cases? They shouldn’t be affected by your changes (applies to all existing tests in this file).

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
     XCTAssertEqual(diagnostics.fullReport?.items, [])
   }
+
+  func testInternalNavigationInSwiftInterface() async throws {
+    let testClient = try await TestSourceKitLSPClient(
+      capabilities: ClientCapabilities(experimental: [
+        GetReferenceDocumentRequest.method: .dictionary(["supported": .bool(true)])
+      ])
+    )
+    let uri = DocumentURI(for: .swiftinterface)
```
ahoppen: The original source file is a plain `.swift` file, not a Swift interface, so this should be `for: .swift`. Also, I don’t think you need the `Language` definition for `.swiftinterface` at the top of this file.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+
+    let positions = testClient.openDocument(
+      """
+      func test(x: 1️⃣String) {
+        let a: 2️⃣Int = 5
```
ahoppen: The location marker for `Int` is never used. You should be able to remove it.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+    testClient.send(
+      DidOpenTextDocumentNotification(
+        textDocument: TextDocumentItem(
+          uri: interfaceUri,
+          language: .swift,
+          version: 0,
+          text: interfaceContents.content
+        )
+      )
+    )
```
ahoppen: I think you can simplify this slightly to 


```suggestion
    testClient.openDocument(interfaceContents.content, interfaceUri, language: .swift)
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+
+    // Find a symbol within the interface (e.g., "init" method in String)
+    guard let initRange = interfaceContents.content.range(of: "public init()") else {
+      XCTFail("Could not find 'public init()' in String interface")
+      return
+    }
```
ahoppen: Can be simplified to the following, I think.


```suggestion
    let initRange = try XCTUnwrap(interfaceContents.content.range(of: "public init()"))
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+          uri: interfaceUri,
+          language: .swift,
+          version: 0,
+          text: interfaceContents.content
+        )
+      )
+    )
+
+    // Find a symbol within the interface (e.g., "init" method in String)
+    guard let initRange = interfaceContents.content.range(of: "public init()") else {
```
ahoppen: Could we pick a function that’s unique within the Swift interface (eg. `public func withUTF8Buffer`) to make this test less likely to change if new initializers get introduced.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      DefinitionRequest(
+        textDocument: TextDocumentIdentifier(interfaceUri),
+        position: initPosition
+      )
+    )
+
+    // The definition should either:
+    // 1. Return a position within the same interface (internal navigation)
+    // 2. Return the same position if it's already at the definition
+    // 3. Return nil if no definition is available
```
ahoppen: The test should deterministically return the same results, in which case only one of these checks should be true. Could we remove the other options?

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      // If we get a location, it should be in the same interface or a related one
+      XCTAssertTrue(
+        location.uri.pseudoPath.hasSuffix(".swiftinterface") ||
+        location.uri == interfaceUri,
+        "Internal navigation should stay within interface files, got: \(location.uri.pseudoPath)"
+      )
+    }
+  }
+
+  func testFoundationImportNavigation() async throws {
```
ahoppen, line 323: Is there anything in this test that isn’t already covered by `testSwiftInterfaceAcrossModules`?

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      """
+      import 1️⃣Foundation.2️⃣NSAffineTransform
+      """,
+      uri: uri
+    )
+
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    if let foundationLocation = foundationDefinition?.locations?.only {
```
ahoppen: Should we check that we actually get a location and make this an `XCTUnwrap` (same below).


```suggestion
    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +306,137 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      XCTAssertTrue(
+        foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface") ||
+        foundationLocation.uri.scheme == "sourcekit-lsp"
+      )
```
ahoppen: Both of these conditions should be satisfied, right? In which case we can also split this into two `XCAssert` calls

```suggestion
      XCTAssert(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
      XCTAssert(foundationLocation.uri.scheme == "sourcekit-lsp")
```

Same below

ahoppen (commented):
Thanks for your continued work on this PR and adding the tests, @aelam. I left some comments, mostly on the tests. Also, addressing the comments in SourceKitLSPServer.swift would be important to make the next review round meaningful.

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
aelam: ~~I assumed user is already in interface file~~
~~since it makes no difference I can remove it~~

```swift
 XCTAssertTrue(
        location.uri.pseudoPath.hasSuffix(".swiftinterface") ||
```

I added test case for file suffix, is it necessary?

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
aelam: sorry about the formatting issue

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
aelam: deleted

Sources/SKTestSupport/Utils.swift
aelam: I deleted but please check the tests about `swiftinterface`

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
aelam: I deleted this tests, maybe it's duplicated

aelam:
>  Also, addressing the comments in SourceKitLSPServer.swift 

I went through all but maybe I missed any you may want to highlight? 

Sorry about too many unrelated changes, I just reverted a lot of unneccessary change and `resovle conversation`  to make it clean

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2051,9 +2065,36 @@ extension SourceKitLSPServer {
     originatorUri: DocumentURI,
     languageService: LanguageService
   ) async throws -> Location {
+    // Check if we're already in the target interface with the same module/group/symbol
+    if case .generatedInterface(let interfaceData) = try? ReferenceDocumentURL(from: originatorUri),
```
ahoppen: This comment still applies here: https://github.com/swiftlang/sourcekit-lsp/pull/2233/#discussion_r2267852544

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +323,64 @@ final class SwiftInterfaceTests: XCTestCase {
...
+    )
+    let uri = DocumentURI(for: .swiftinterface)
+
+    let positions = testClient.openDocument(
+      """
+      import 1️⃣Foundation
+      """,
+      uri: uri,
+      language: .swift
+    )
```
ahoppen: This comment still applies here: https://github.com/swiftlang/sourcekit-lsp/pull/2233/#discussion_r2272261588. This should be able to become


```suggestion
    let uri = DocumentURI(for: .swift)

    let positions = testClient.openDocument(
      """
      import 1️⃣Foundation
      """,
      uri: uri
    )
```

It doesn’t make sense of the user to start in a `.swiftinterface` file they found somewhere on disk because we don’t have a build system that would be able to provide build settings for that `.swiftinterface` file – we only have build settings for swift interface files that were generated by SourceKit-LSP.

Sources/SourceKitLSP/Swift/OpenInterface.swift
aelam: please check the latest change. 
it's like `module.group.buildSettingFileHash`
but I think it's still not perfect. 

In my understanding A.swift and B.swift would generate different hash. then it would be different Foundation.hash1.swiftinterface and Foundation.hash2.swiftinterface

I tried to use the workspace.rootUri hash, it can fix the above issue
but the workspace is the same for sourcekit-lsp and the `.build/checkout/IndexStoreDB/IndexStoreDB.swift`. is it the expected behavior?

```swift
// Get workspace to use its rootUri for build context identification
let workspace = await self.workspaceForDocument(uri: originatorUri)
let buildContextId = workspace?.rootUri?.stringValue ?? originatorUri.buildSettingsFile.stringValue
let buildContextHash = "\(abs(buildContextId.hashValue))"
```

Sources/SwiftLanguageService/OpenInterface.swift
```diff
@@ -22,10 +22,16 @@ extension SwiftLanguageService {
     groupName: String?,
     symbolUSR symbol: String?
   ) async throws -> GeneratedInterfaceDetails? {
+    // Include build settings context to distinguish different versions/configurations
+    let buildSettingsFileHash = "\(abs(document.buildSettingsFile.stringValue.hashValue))"
+    let sourcekitdDocumentName = [moduleName, groupName, buildSettingsFileHash].compactMap(\.self).joined(
+      separator: "."
+    )
```
ahoppen: Nitpick. I think this looks nicer if we break in front of `.joined`. Something like

```suggestion
    let sourcekitdDocumentName = [moduleName, groupName, buildSettingsFileHash].compactMap(\.self)
      .joined(separator: ".")
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
...
+
+    // Test navigation to Foundation module
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
+    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
```
ahoppen: We can make this return slightly nicer error messages if these assertions should fail.

```suggestion
    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      """,
+      uri: uri
+    )
+
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+    XCTAssertTrue(foundationLocation.uri.pseudoPath.contains("Foundation.swiftinterface"))
+    XCTAssertTrue(foundationLocation.uri.scheme == "sourcekit-lsp")
```
ahoppen:
```suggestion
    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
...
+
+    // Test navigation to NSAffineTransform
+    let transformDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["2️⃣"])
+    )
+    let transformLocation = try XCTUnwrap(transformDefinition?.locations?.only)
+    // Verify we can identify this as a swiftinterface file
+    XCTAssertTrue(transformLocation.uri.pseudoPath.contains("Foundation.NSAffineTransform.swiftinterface"))
+    XCTAssertTrue(transformLocation.uri.scheme == "sourcekit-lsp")
```
ahoppen:
```suggestion
    XCTAssertEqual(transformLocation.uri.scheme, "sourcekit-lsp")
    assertContains(transformLocation.uri.pseudoPath, "Foundation.NSAffineTransform.swiftinterface")
    
```

Sources/SwiftLanguageService/OpenInterface.swift
```diff
@@ -22,10 +22,16 @@ extension SwiftLanguageService {
...
+
```
ahoppen, line 29: Let’s also include the `buildSettingsFileHash` below so we don’t override generated interfaces if they have a different build settings file.

Ie. below, make it

```swift
let interfaceFilePath = self.generatedInterfacesPath.appendingPathComponent(buildSettingsFileHash).appendingPathComponent(urlData.displayName)
try FileManager.default.createDirectory(at: interfaceFilePath.deletingLastPathComponent(), withIntermediateDirectories: true)
```

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
...
+      import 1️⃣Foundation.2️⃣NSAffineTransform
+      """,
+      uri: uri
+    )
+
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+        XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
```
ahoppen: Could you run swift format to format your changes?

ahoppen (commented):
Just one formatting issue left, otherwise LGTM.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -319,6 +319,63 @@ final class SwiftInterfaceTests: XCTestCase {
...
+    let foundationDefinition = try await testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+    let foundationLocation = try XCTUnwrap(foundationDefinition?.locations?.only)
+    XCTAssertEqual(foundationLocation.uri.scheme, "sourcekit-lsp")
+    assertContains(foundationLocation.uri.pseudoPath, "Foundation.swiftinterface")
+  }
+
+  func testFoundationSubmoduleNavigation() async throws {
+    let testClient = try await TestSourceKitLSPClient(
```
ahoppen, line 351: Looks like Foundation only has submodules on Darwin platforms (macOS, iOS, etc.). Let’s skip this test on Linux and Windows by adding the following at the start of the test.

```swift
try SkipUnless.platformIsDarwin("Non-Darwin platforms don't have Foundation submodules")
```

aelam: Thanks for very detailed code every time  😄

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Congrats on your first contribution to SourceKit-LSP, @aelam.

PR #2235, opened 2025-08-12
[CodeActions] Switch to use swift-syntax package manifest abstraction…

xedin:
…s and refactoring actions

The refactoring actions and supporting types have been recently ported from swift-package-manager to swift-syntax, to support this change let's move code actions to use types and refactoring actions from SwiftRefactor module.


xedin:
@swift-ci please test

xedin:

PR #2237, opened 2025-08-13
Ensure that we end the `Reloading package` interval if a call in `reloadPackageAssumingOnPackageLoadingQueue` throws

ahoppen:
If a call in `reloadPackageAssumingOnPackageLoadingQueue` throws, we weren’t getting to the end of it and would thus never end the signpost started within. Move the `endInterval` call up ensure it is always ended.

Found this while investigating https://ci-external.swift.org/job/swift-PR-windows/44247/console, which failed SourceKit-LSP testing because PackageA could not be loaded due to

```
Initial package loading: invalid access to C:\Users\swift-ci\jenkins\workspace\swift-PR-windows\build\tmp\lsp-test\924F7085\PackageA\.build\index-build\x86_64-unknown-windows-msvc\debug\MyLibrary.build\output-file-map.json
```

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -389,6 +389,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     )
     await testHooks.reloadPackageDidStart?()
     defer {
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

PR #2238, opened 2025-08-13
Allow non-SwiftPM build systems to have larger indexing batch sizes

rockbruno:
I noticed that sourcekit-lsp forces the batch size to be 1 for SwiftPM reasons, but nowadays there are other ways to run it. On the [Bazel BSP](https://github.com/spotify/sourcekit-bazel-bsp) for example we have no issues building tons of targets at the same time, so it seemed feasible to keep the existing limitation for SwiftPM but allow other build systems to parallelize more index tasks.

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -316,13 +316,17 @@ package actor BuildServerManager: QueueBasedMessageHandler {
   /// `nil` if the `BuildServerManager` does not have an underlying build server.
   package let configPath: URL?
 
+  /// The kind of underlying build server adapter that is being managed.
+  /// `nil` if the `BuildServerManager` does not have an underlying build server.
+  package var kind: BuildServerSpec.Kind?
+
   /// The files for which the delegate has requested change notifications, ie. the files for which the delegate wants to
   /// get `fileBuildSettingsChanged` and `filesDependenciesUpdated` callbacks.
   private var watchedFiles: [DocumentURI: (mainFile: DocumentURI, language: Language)] = [:]
 
   private var connectionToClient: BuildServerManagerConnectionToClient
 
-  /// The build serer adapter that is used to answer build server queries.
+  /// The build server adapter that is used to answer build server queries.
   private var buildServerAdapter: BuildServerAdapter?
 
   /// The build server adapter after initialization finishes. When sending messages to the BSP server, this should be
@@ -451,6 +455,7 @@ package actor BuildServerManager: QueueBasedMessageHandler {
     self.toolchainRegistry = toolchainRegistry
     self.options = options
     self.connectionToClient = connectionToClient
+    self.kind = buildServerSpec?.kind
     self.configPath = buildServerSpec?.configPath
     self.buildServerAdapter = await buildServerSpec?.createBuildServerAdapter(
       toolchainRegistry: toolchainRegistry,
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -222,6 +222,9 @@ package final actor SemanticIndexManager {
   /// The parameter is the number of files that were scheduled to be indexed.
   private let indexTasksWereScheduled: @Sendable (_ numberOfFileScheduled: Int) -> Void
 
+  /// The number of tasks to prepare concurrently, whenever a index request is scheduled.
+  private let indexTaskBatchSize: Int
+
   /// Callback that is called when `progressStatus` might have changed.
   private let indexProgressStatusDidChange: @Sendable () -> Void
 
@@ -261,6 +264,7 @@ package final actor SemanticIndexManager {
     updateIndexStoreTimeout: Duration,
     hooks: IndexHooks,
     indexTaskScheduler: TaskScheduler<AnyIndexTaskDescription>,
+    indexTaskBatchSize: Int,
     logMessageToIndexLog:
       @escaping @Sendable (
         _ message: String, _ type: WindowMessageType, _ structure: StructuredLogKind
@@ -273,6 +277,7 @@ package final actor SemanticIndexManager {
     self.updateIndexStoreTimeout = updateIndexStoreTimeout
     self.hooks = hooks
     self.indexTaskScheduler = indexTaskScheduler
+    self.indexTaskBatchSize = indexTaskBatchSize
     self.logMessageToIndexLog = logMessageToIndexLog
     self.indexTasksWereScheduled = indexTasksWereScheduled
     self.indexProgressStatusDidChange = indexProgressStatusDidChange
@@ -877,10 +882,7 @@ package final actor SemanticIndexManager {
 
     var indexTasks: [Task<Void, Never>] = []
 
-    // TODO: When we can index multiple targets concurrently in SwiftPM, increase the batch size to half the
-    // processor count, so we can get parallelism during preparation.
-    // (https://github.com/swiftlang/sourcekit-lsp/issues/1262)
-    for targetsBatch in sortedTargets.partition(intoBatchesOfSize: 1) {
+    for targetsBatch in sortedTargets.partition(intoBatchesOfSize: indexTaskBatchSize) {
       let preparationTaskID = UUID()
       let filesToIndex = targetsBatch.flatMap({ filesByTarget[$0]! })
 
--- a/Sources/SourceKitLSP/Workspace.swift
+++ b/Sources/SourceKitLSP/Workspace.swift
@@ -162,12 +162,23 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
     if options.backgroundIndexingOrDefault, let uncheckedIndex,
       await buildServerManager.initializationData?.prepareProvider ?? false
     {
+      let batchSize: Int
+      switch await buildServerManager.kind {
+      case .swiftPM:
+        // TODO: When we can index multiple targets concurrently in SwiftPM, increase the batch size to half the
+        // processor count, so we can get parallelism during preparation.
+        // (https://github.com/swiftlang/sourcekit-lsp/issues/1262)
+        batchSize = 1
+      default:
+        batchSize = max(1, ProcessInfo.processInfo.activeProcessorCount / 2)
+      }
       self.semanticIndexManager = SemanticIndexManager(
         index: uncheckedIndex,
         buildServerManager: buildServerManager,
         updateIndexStoreTimeout: options.indexOrDefault.updateIndexStoreTimeoutOrDefault,
         hooks: hooks.indexHooks,
         indexTaskScheduler: indexTaskScheduler,
+        indexTaskBatchSize: batchSize,
         logMessageToIndexLog: { [weak sourceKitLSPServer] in
           sourceKitLSPServer?.logMessageToIndexLog(message: $0, type: $1, structure: $2)
         },


Sources/SourceKitLSP/Workspace.swift
```diff
@@ -162,12 +162,23 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
...
+      let batchSize: Int
+      switch await buildServerManager.kind {
+      case .swiftPM:
+        // TODO: When we can index multiple targets concurrently in SwiftPM, increase the batch size to half the
+        // processor count, so we can get parallelism during preparation.
+        // (https://github.com/swiftlang/sourcekit-lsp/issues/1262)
+        batchSize = 1
+      default:
+        batchSize = max(1, ProcessInfo.processInfo.activeProcessorCount / 2)
+      }
```
rockbruno: If increasing this for everyone else (not just me) is undesirable, I think one other possible alternative could be to pass this as an argument in the initialize response

bnbarham (commented):
Oh interesting, thanks for looking into this @rockbruno! Have you been able to run any experiments with this to see if it helps much? I'm mostly wondering if half the number of processor count actually makes sense here - it would seem to imply that the buildsystem itself isn't very parallel.

Another question is whether we hook this up to `maxCoresPercentageToUseForBackgroundIndexing`. It's currently marked as internal and right now that actually doesn't do much - it only controls the number of background tasks, which we limit to a single task anyway (or zero if `maxCoresPercentageToUseForBackgroundIndexing` is 0). Seems like a nice way to easily control this though. It could default to 0.5 instead of the 1 it does today.

tinder-maxwellelliott (approved):

rockbruno:
@bnbarham Having it customizable sounds better yeah. For Bazel specifically the bottleneck is that starting Bazel takes some time, so individual requests are crazy slow. So having one request that builds tons of targets is much better.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,81 @@ final class BackgroundIndexingTests: XCTestCase {
...
+
+    preparationTaskSemaphore.waitOrXCTFail()
+
+    // The test project has 10 targets, and we should have received them in batches of 3.
+    XCTAssertEqual(preparationTasks.value.count, 4)
+    let preparedTargetBatches = preparationTasks.value.map(\.targetsToPrepare).sorted { $0.count > $1.count }
+    XCTAssertEqual(preparedTargetBatches[0].count, 3)
+    XCTAssertEqual(preparedTargetBatches[1].count, 3)
+    XCTAssertEqual(preparedTargetBatches[2].count, 3)
+    XCTAssertEqual(preparedTargetBatches[3].count, 1)
```
rockbruno: @bnbarham I noticed that the order and contents of the batches were unpredictable (like the same batch being triggered twice in a row), which I assumed to be intentional, so I went for this approach of just checking the values instead. Let me know if this is wrong!

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -322,9 +322,13 @@ package actor BuildServerManager: QueueBasedMessageHandler {
 
   private var connectionToClient: BuildServerManagerConnectionToClient
 
-  /// The build serer adapter that is used to answer build server queries.
+  /// The build server adapter that is used to answer build server queries.
   private var buildServerAdapter: BuildServerAdapter?
 
+  /// The kind of underlying build server adapter that is being managed.
+  /// `nil` if the `BuildServerManager` does not have an underlying build system.
+  private var kind: BuildServerSpec.Kind?
```
ahoppen: Is the kind needed anywhere?

Sources/BuildServerProtocol/Messages/InitializeBuildRequest.swift
```diff
@@ -278,6 +278,9 @@ public struct SourceKitInitializeBuildResponseData: LSPAnyCodable, Codable, Send
   /// Whether the server implements the `textDocument/sourceKitOptions` request.
   public var sourceKitOptionsProvider: Bool?
 
+  /// The number of targets to prepare concurrently, when an index request is scheduled.
+  public var indexTaskBatchSize: Int?
```
ahoppen: I don’t think that the build server should control how many targets SourceKit-LSP prepares in parallel. The build server should communicate its capabilities and SourceKit-LSP should decide the optimal partition size – starting from a fixed value over a user-configured to parameter to some more sophisticated logic that decides on the number of targets to prepare depending on how many files are waiting to be indexed.

My proposal would be to change this to the following and using `{"supported": true}` as the default value since the `buildTarget/prepare` request has been designed to support multiple targets from the start. The SwiftPM build server would be the one that returns `"supported": false` here.
```swift
public struct MultiTargetPreparationSupport {
  /// Whether the build server can prepare multiple targets in parallel. If this value is omitted, it is assumed to be `true`.
  var supported: Bool?
}

/// Options to control how many targets should be prepared simultaneously by SourceKit-LSP.
public var multiTargetPreparation: MultiTargetPreparationSupport?
```

Using a dictionary here allows us to add more fine-tuning parameters in the future if we decide that they are helpful for SourceKit-LSP to make optimal decisions like the expected startup time to build time ratio (which would influence how large the batches are).

Also, could you add documentation for this in https://github.com/swiftlang/sourcekit-lsp/blob/main/Contributor%20Documentation/BSP%20Extensions.md?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -877,10 +882,7 @@ package final actor SemanticIndexManager {
 
     var indexTasks: [Task<Void, Never>] = []
 
-    // TODO: When we can index multiple targets concurrently in SwiftPM, increase the batch size to half the
-    // processor count, so we can get parallelism during preparation.
-    // (https://github.com/swiftlang/sourcekit-lsp/issues/1262)
-    for targetsBatch in sortedTargets.partition(intoBatchesOfSize: 1) {
+    for targetsBatch in sortedTargets.partition(intoBatchesOfSize: indexTaskBatchSize) {
```
ahoppen, line 877: Now, if we don’t let the build system specify the batch size as I suggested above, we need to pick a batch size here. An idea that I’d like to toss in there would be to batch the targets so that each target batch has roughly `<some factor> * processor count` source files (eg. keep adding targets until the batch has more than that limit of source files). This would have the following advantages:
- We batch up many small targets that are quick to prepare to avoid the startup cost but don’t batch up multiple large targets, which would stall background indexing when prepared in a batch and only have marginal benefits from sharing the build system setup cost.
- Each of the files in the target will cause one compiler process to be launched to index the source file (at least until https://github.com/swiftlang/sourcekit-lsp/issues/1268 is fixed). So we can keep each processor core busy indexing a file for `<some factor>` times the compiler’s execution duration after one preparation batch is finished.

Of course, this assumes that all source files are roughly the same size but I think that’s a reasonable assumption in most projects in practice.

What do you think of this idea? If it makes sense to you, I’d like to hear your opinions on what a good value for `<some factor>` would be. My gut feeling tells me that it should be something around 5 - 10 but I’d be interested to see some benchmark results (doesn’t have to be too sophisticated). Attaching Instruments to SourceKit-LSP during the background indexing with the _Regions of Interest_ instrument should give you an idea on which tasks are happening in parallel if that helps.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
ahoppen: The same batch being triggered twice in a row sounds concerning to me. I would prefer to have some assertion here that actually checks that all targets are getting prepared.

ahoppen: This test will crash if `preparationTasks.value.count < 4`. We should early exit the test if `preparationTasks.value.count != 4`.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,81 @@ final class BackgroundIndexingTests: XCTestCase {
...
+        "test.swift": """
+        func testFunction() {}
+        """
+      ],
+      buildServer: BuildServer.self,
+      hooks: Hooks(
+        indexHooks: IndexHooks(
+          preparationTaskDidStart: { task in
+            preparationTasks.withLock { preparationTasks in
+              // Ignore everything after the 4th batch to avoid flakiness.
```
ahoppen: What’s the flakiness here?

Sources/BuildServerIntegration/BuildServerManager.swift
rockbruno: Ah, the previous code was using this to determine if it was a SwiftPM server, but I force pushed that out and forgot to remove this one

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
rockbruno: Similar to what I mentioned above, I noticed that the tasks kept coming for tasks that were supposedly already processed, so I had some flakiness on my end. I assumed it was some particularity of how the test build system was implemented but I'll double check it

rockbruno:
@ahoppen I like it, sounds like a mix between the two approaches (I force pushed the previous code, but it was using processors / 2 as the number). I'm gonna run some tests and see what seems to work better here for Bazel to get an idea how high this can be.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
rockbruno: I found out why that happened. Sometimes the first batch will be marked as `cancelToBeRescheduled` mid execution, meaning the prepare/ hook gets called twice for it.

rockbruno: Most specifically this bit here: https://github.com/swiftlang/sourcekit-lsp/blob/1ab9e2e9364c492f00701ed29f579c783c0ace22/Sources/SemanticIndex/PreparationTaskDescription.swift#L118-L130

Is this something we'd need to change?

ahoppen: Oh, yes, that would make sense. I did write a comment referencing this piece of code earlier today but GitHub must have somehow lost it 😞 Though I didn’t think about this case in particular.

What I did mention in that comment is that target preparation for editor functionality will cancel preparation for background indexing if the user keeps switching between module before initial indexing finishes or if they are editing a file in one module and switching to another module while preparation for background indexing is happening. That’s why if in doubt, I would prefer smaller batch sizes, simply because they are more likely to finish making progress.

Also, if the build system can support running multiple prepare requests in parallel (not sure if Bazel can do that), it would be worth reporting that as an option in the initialize response because in that case we don’t need to cancel any preparation tasks like we do here. One interesting case to consider is that if this option is set to true, SourceKit-LSP could request preparation of targets A and B and then also request preparation of A, which it expect to somehow handle concurrently.

Back to your comment: I didn’t think about the case that batches can be of different sizes for background indexing when I wrote the comment, so it doesn’t really make sense. Instead, what we need to do is keep track of the `TargetPreparationPurpose` in the `PreparationTaskDescription` and only let `forEditorFunctionality` cancel `forIndexing`.

bnbarham:
> For Bazel specifically the bottleneck is that starting Bazel takes some time, so individual requests are crazy slow

Out of interest, is it possible to run Bazel as eg. a daemon in order to avoid that startup time? How are requests for compiler arguments handled, is Bazel run once to get all the commands, and then those are cached?

rockbruno:
@bnbarham @ahoppen Gave this another run. I ran some quick benchmarks with our example project and this is what I got:

```
Batch size 1: 1:30 minutes
Batch size 4: 30 seconds
Batch size 8: 20 seconds
```

Important to note that the times vary heavily based on the order SK builds internally (which is not always the same), so there's some luck involved for the size 1 case. But for my test project at least it seemed to be the case that the more parallelism the better (since Bazel is good at it).

Based on this however I thought it would be better to make a couple of changes:
- I would suggest defaulting to false instead of the other way around, mostly to avoid accidentally breaking something on the other build systems that we didn't test.
- While we can have SK have a default batch size, I think it will be good to still allow this to be customized for two reasons:
    - I think the "sweet spot" depends on the build system capabilities and project complexity. For my test project for example it seems that building as much as possible is the best choice, but I'm not sure that would be the best choice for the main Spotify project (larger targets, so longer downtime / more cancellations issued from the LSP side). So I think that each project will have its own sweet spot that is not necessarily something like `cpu / 2`.
    - (less relevant) I wasn't sure how to properly unit test this without allowing it to be changed :)

@bnbarham For your Bazel questions:

- Bazel uses a daemon already, but we have some overhead to 1) establish a connection to the server, and 2) determine whether the invocation will result in Bazel needing to re-calculate some info internally, which can take a while depending on the size of the repo. So doing something like `bazel build targetA targetB targetC` will always be faster than individual calls.

ahoppen:
> While we can have SK have a default batch size, I think it will be good to still allow this to be customized for two reasons:

I am wondering whether the customization option should come from the build system or the user. My feeling is that the ideal value is more dependent on the project that the user is opening, which means that the optimal preparation batch size should also be customizable on a per-project level using SourceKit-LSP.

> I would suggest defaulting to false instead of the other way around, mostly to avoid accidentally breaking something on the other build systems that we didn't test.

I think we should default continue to default to `true` because:
- The BSP extension was always explicitly shaped to allow parallel preparation of multiple tasks, so if a BSP server can’t handle multiple targets, that’s definitely a bug on their side.
- I am not aware of any BSP servers besides the built-in SwiftPM server and your Bazel BSP server now that support background preparation.
- I don’t want to be stuck with a historically motivated default for the future. If a new BSP server gets implemented, it should support parallel preparation of targets.

> I think the "sweet spot" depends on the build system capabilities and project complexity. For my test project for example it seems that building as much as possible is the best choice, but I'm not sure that would be the best choice for the main Spotify project (larger targets, so longer downtime / more cancellations issued from the LSP side). So I think that each project will have its own sweet spot that is not necessarily something like cpu / 2.

Have you had a chance to try the file count-based approach I mentioned in https://github.com/swiftlang/sourcekit-lsp/pull/2238#discussion_r2276329676? The more I think about it, the more I believe that file count is a better measure here than target count because targets can vary drastically in size. It would also make it easier to interpret the results of your measurements since I don’t know how big the targets are that you are testing but I do have a rough idea how big a typical source file is.

rockbruno:
I think the file count has the same issue of having a potentially different sweet spot per project (protobuf generated specs are _huge_ yet are "just" one file, for example), but like you said you could probably assume that in the majority of projects this would be relatively predictable. Personally I find the num of targets one easier to reason about as an user, but using file counts also feels like a valid choice. But for any choice, I think it would be valuable to let those with a more unusual project structure be able to fine-tune it if the default causes it to be too bottleneck-y for some structural reason. Let me know what you prefer for this PR!

Will make the change to make `true` the default shortly

ahoppen:
Sorry for the delayed response, I needed to think about how best to specify the target batching strategy. Here are my thoughts:

I want to allow ourselves to evolve the target batching strategy in the future and thus don’t want to include any batch size related options in our BSP extensions, since we want to keep those stable. Instead, the target batching strategy should be configured in the [Configuraiton File](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Configuration%20File.md), which we can evolve more easily. This also has the advantage that users can adjust the strategy to find the optimal values for their project insted of the BSP server deciding for them.

I would propose a configurations schema similar to the following inside the [`index` key](https://github.com/swiftlang/sourcekit-lsp/blob/c1a8301b1ffd5dae3e887f8ee8a0689329ffaf93/config.schema.json#L125) of the configuration file.

```json
{
  "preparationBatchingStrategy": {
    "description": "If the BSP server supports preparation of multiple targets in a batch, controls the size of the batches. Note that SwiftPM currently does not support batched target preparation, so this has no affect on Swift packages. Adjusting these options might improve performance of background preparation based on the used BSP server. The available batching strategies may change in the future.",
    "oneOf": [
      {
        "type": "object",
        "description": "Prepare a fixed number of targets in a single batch",
        "properties": {
          "strategy": {
            "const": "target"
          },
          "batchSize": {
            "type": "integer",
            "description": "Defines how many targets should be prepared in a single batch"
          }
        },
        "required": [
          "strategy",
          "batchSize"
        ]
      },
      {
        "type": "object",
        "description": "Prepare as many targets in a single batch so that these targets contain more than the specified number of source files",
        "properties": {
          "strategy": {
            "const": "files"
          },
          "files": {
            "type": "integer",
            "description": "Accumulate targets in a target batch until they contain more files than specified by this property."
          }
        },
        "required": [
          "strategy",
          "batchSize"
        ]
      }
    ]
  }
}
```

As for finding a default strategy, you have the most experience and should pick a reasonable value. If you can, I would appreciate if you could back the choice of that value up with some kind of measurement in a doc comment so it’s less of a magic number, nothing too sophisticated but something we can read in a year and follow the steps that lead to this value.

We can make the file-based batching strategy a follow-up PR but I’d really like to see how that behaves. As you noted, files may also differ in size but in my experience the file count is a more accurate representation of compile time then the number of targets. As a side note, allowing us to generate the `oneOf` in the schema above like this will likely need quite a bit new functionality in [ConfigSchemaGen](https://github.com/swiftlang/sourcekit-lsp/tree/main/SourceKitLSPDevUtils/Sources/ConfigSchemaGen). If we only stick to the target-based strategy, we should only need support for the `const` key in the JSON schema, which should be a lot easier to accomplish.

@bnbarham, does this make sense to you as well? Do you have anything to add?

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,84 @@ final class BackgroundIndexingTests: XCTestCase {
...
     XCTAssertEqual(symbols?.count, 1)
   }
+
+  func testBuildServerUsesCustomTaskBatchSize() async throws {
+    final class BuildServer: CustomBuildServer {
+      let inProgressRequestsTracker = CustomBuildServerInProgressRequestTracker()
+      private let projectRoot: URL
+      private var testFileURL: URL { projectRoot.appendingPathComponent("test.swift").standardized }
+
+      nonisolated(unsafe) var preparedTargetBatches = [[BuildTargetIdentifier]]()
```
ahoppen, line 2634: If you make `BuildServer` an `actor`, you don’t need `nonisolated(unsafe)`.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,84 @@ final class BackgroundIndexingTests: XCTestCase {
     let symbols = try await project.testClient.send(WorkspaceSymbolsRequest(query: "myTestFu"))
     XCTAssertEqual(symbols?.count, 1)
   }
+
+  func testBuildServerUsesCustomTaskBatchSize() async throws {
+    final class BuildServer: CustomBuildServer {
+      let inProgressRequestsTracker = CustomBuildServerInProgressRequestTracker()
+      private let projectRoot: URL
+      private var testFileURL: URL { projectRoot.appendingPathComponent("test.swift").standardized }
```
ahoppen, line 2632: Do we need `.standardized` here? I suppose you copied this from `testBuildServerUsesStandardizedFileUrlsInsteadOfRealpath`, which needed to use `.standardized` because it specifically tested behavior around `standardized` vs `realpath`.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,84 @@ final class BackgroundIndexingTests: XCTestCase {
...
+          projectRoot: projectRoot,
+          outputPathsProvider: false,
+          multiTargetPreparation: MultiTargetPreparationSupport(supported: true, batchSize: 3)
+        )
+      }
+
+      func buildTargetSourcesRequest(_ request: BuildTargetSourcesRequest) async throws -> BuildTargetSourcesResponse {
+        var dummyTargets = [BuildTargetIdentifier]()
+        for i in 0..<10 {
+          dummyTargets.append(BuildTargetIdentifier(uri: try! URI(string: "dummy://dummy-\(i)")))
```
ahoppen, line 2651: I think this `try!` could just be a `try`. Just removes one possibility that might lead to the test crashing and resolves my allergic reaction to `try!` 😉 

Same below

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,84 @@ final class BackgroundIndexingTests: XCTestCase {
...
+        }
+        return BuildTargetSourcesResponse(items: dummyTargets.map {
```
ahoppen, line 2653: Oh, interesting that this even works. The BSP server only reports a single target [here](https://github.com/swiftlang/sourcekit-lsp/blob/c1a8301b1ffd5dae3e887f8ee8a0689329ffaf93/Sources/SKTestSupport/CustomBuildServerTestProject.swift#L231-L242) and we return more targets than included in the `BuildTargetSourcesRequest` request here, which is actually a bug in SourceKit-LSP. We should implement `workspaceBuildTargetsRequest ` and return all 10 targets from it and ideally filter to only return the requested targets here (or add an assertion that the request contains all 10 targets at least).

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2624,6 +2624,84 @@ final class BackgroundIndexingTests: XCTestCase {
...
+        """
+      ],
+      buildServer: BuildServer.self,
+      enableBackgroundIndexing: true,
+    )
+
+    try await project.testClient.send(SynchronizeRequest(index: true))
+
+    let buildServer = try project.buildServer()
+    let preparedBatches = buildServer.preparedTargetBatches.sorted { $0[0].uri.stringValue < $1[0].uri.stringValue }
```
ahoppen, line 2683: Not that I would expect to receive this but this will crash if a preparation batch is empty. It might be easier to just make this `let preparedBatches = Set(buildServer.preparedTargetBatches)` to make it order-invariant. Or just make `preparedTargetBatches` inside the `BuildServer` a `Set<Set<BuildTargetIdentifier>>` to start with.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -222,6 +213,9 @@ package final actor SemanticIndexManager {
   /// The parameter is the number of files that were scheduled to be indexed.
   private let indexTasksWereScheduled: @Sendable (_ numberOfFileScheduled: Int) -> Void
 
+  /// The size of the batches in which the `SemanticIndexManager` should dispatch preparation tasks.
+  private let indexTaskBatchSize: Int
```
ahoppen, line 217: Let’s name this `preparationBatchSize`.

bnbarham:
> @bnbarham, does this make sense to you as well? Do you have anything to add?

Makes sense to me, it being in the LSP configuration is what I was thinking of in my:
> Maybe the better option here is to just add an option to control it so that users can tweak it as they desire?

Limiting it to just target in this PR also sounds good to me.

> @bnbarham For your Bazel questions:
> Bazel uses a daemon already, but we have some overhead to 1) establish a connection to the server, and 2) determine whether the invocation will result in Bazel needing to re-calculate some info internally, which can take a while depending on the size of the repo. So doing something like bazel build targetA targetB targetC will always be faster than individual calls.

Interesting, good to know, thanks.

rockbruno:
All sounds good to me :) I'm quite swamped by now so will not have time to re-write this right now, so feel free to take over and open a new PR if you want to. Otherwise I'll circle back to this at some point in the future

josh-arnold-1:
I was chatting with Bruno — happy to take over this PR if you don’t already have plans to, @bnbarham 🙏

One thought: would it make sense to default batching to false for now, and let the build server specify its own batch size if it wants to enable batching?

I’m wondering if trying to “guess” a reasonable default batch size (independent of build system type) might end up being tricky.

Thanks!

bnbarham:
Feel free to pick it up if you have time :)

> One thought: would it make sense to default batching to false for now, and let the build server specify its own batch size if it wants to enable batching?
> 
> I’m wondering if trying to “guess” a reasonable default batch size (independent of build system type) might end up being tricky.

"false" as in set "target" batch size to 1? If so, I'm okay with that - it is no different to what it is today. I'd prefer to keep it in the project configuration rather than build server for now though, we can always extend it later to add a "bsp" strategy + extend the BSP interface.

josh-arnold-1:
I put up a PR here! https://github.com/swiftlang/sourcekit-lsp/pull/2294

ahoppen:
Closing this in favor of https://github.com/swiftlang/sourcekit-lsp/pull/2294.

PR #2243, opened 2025-08-14
Add module map for CSKTestSupport

maxwellE:
Create a module map for the CSKTestSupport module, exporting the header "CSKTestSupport.h". This allows for better modularization and encapsulation of the support functionalities.

### DIFF
--- a/Sources/CSKTestSupport/include/CSKTestSupport.h
+++ b/Sources/CSKTestSupport/include/CSKTestSupport.h
@@ -1,0 +1,10 @@
+#ifndef CSKTestSupport_h
+#define CSKTestSupport_h
+
+#ifdef __linux__
+// For testing, override __cxa_atexit to prevent registration of static
+// destructors due to https://github.com/swiftlang/swift/issues/55112.
+int __cxa_atexit(void (*f) (void *), void *arg, void *dso_handle);
+#endif
+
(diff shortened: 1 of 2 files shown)


Sources/CSKTestSupport/include/CSKTestSupport.h
ahoppen, line 1: Could you add a copyright header to this file?

bnbarham:
My guess is bazel (?), but I'm still wondering why the modulemap is needed - nothing is importing CSKTestSupport, it's just being linked into SKTestSupport to not call static destructors. I assume the modulemap is there at all purely to appease SwiftPM.

And if we *do* need to specify something in the module, my preference would be for it to still be empty rather than adding this header. There's no intention that anything would ever import and call this function.

maxwellE:
This was done to support using this library via https://github.com/cgrindel/rules_swift_package_manager building with Bazel

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2244, opened 2025-08-14
[cmake] Remove dependency on PackageModelSyntax library

xedin:
This is a follow-up to https://github.com/swiftlang/sourcekit-lsp/pull/2235.

The change removed uses of `PackageModelSyntax` but left is in `target_link_libraries` of `SourceKitLSP`.

### DIFF
--- a/Sources/SwiftLanguageService/CMakeLists.txt
+++ b/Sources/SwiftLanguageService/CMakeLists.txt
@@ -74,7 +74,6 @@ target_link_libraries(SourceKitLSP PRIVATE
   Crypto
   TSCBasic
   PackageModel
-  PackageModelSyntax
   SwiftSyntax::SwiftDiagnostics
   SwiftSyntax::SwiftIDEUtils
   SwiftSyntax::SwiftParser


xedin:
@swift-ci please test

Sources/SwiftLanguageService/CMakeLists.txt
```diff
@@ -74,7 +74,6 @@ target_link_libraries(SourceKitLSP PRIVATE
   Crypto
   TSCBasic
   PackageModel
```
bnbarham, line 76: Do we still need PackageModel?

bnbarham (approved):

Sources/SwiftLanguageService/CMakeLists.txt
xedin: Yeah, it's used in a few places. I can look into removing that as well separately!

xedin: This is going to be addressed by https://github.com/swiftlang/sourcekit-lsp/pull/2245

PR #2248, opened 2025-08-19
Make `LanguageService` initializer non-failable

ahoppen:
There’s no reason why we need to allow returning `nil` here, we can just throw an error in the cases where the language service can’t be created.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2250, opened 2025-08-22
Add signature help LSP request support

a7medev:
Depends on https://github.com/swiftlang/swift/pull/83378

---

Adds support for the LSP signature help request.

> [!NOTE]
> As of https://github.com/swiftlang/swift/pull/83378, SourceKitD still doesn't separate parameter documentation from the signature documentation and thus parameters don't have their own separate documentation. This should just work once SourceKitD implements this functionality and we'll only need to modify the tests.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1189,6 +1189,9 @@ extension SourceKitLSPServer {
     if let completionOptions = server.completionProvider {
       await registry.registerCompletionIfNeeded(options: completionOptions, for: languages, server: self)
     }
+    if let signatureHelpOptions = server.signatureHelpProvider {
+      await registry.registerSignatureHelpIfNeeded(options: signatureHelpOptions, for: languages, server: self)
+    }
     if server.foldingRangeProvider?.isSupported == true {
       await registry.registerFoldingRangeIfNeeded(options: FoldingRangeOptions(), for: languages, server: self)
     }
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1083,6 +1083,7 @@ extension SourceKitLSPServer {
       ? nil
       : LanguageServerProtocol.SignatureHelpOptions(
         triggerCharacters: ["(", "["],
+        // We retrigger on `:` as it's potentially after an argument label which can change the active parameter or signature.
         retriggerCharacters: [",", ":"]
       )
 
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -781,6 +781,8 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await self.handleRequest(for: request, requestHandler: self.completion)
     case let request as RequestAndReply<CompletionItemResolveRequest>:
       await request.reply { try await completionItemResolve(request: request.params) }
+    case let request as RequestAndReply<SignatureHelpRequest>:
+      await self.handleRequest(for: request, requestHandler: self.signatureHelp)
     case let request as RequestAndReply<DeclarationRequest>:
       await self.handleRequest(for: request, requestHandler: self.declaration)
     case let request as RequestAndReply<DefinitionRequest>:
@@ -1076,6 +1078,14 @@ extension SourceKitLSPServer {
         triggerCharacters: [".", "("]
       )
 
+    let signatureHelpOptions =
+      await registry.clientHasDynamicSignatureHelpRegistration
+      ? nil
+      : LanguageServerProtocol.SignatureHelpOptions(
+        triggerCharacters: ["(", "["],
+        retriggerCharacters: [",", ":"]
+      )
+
     let onTypeFormattingOptions =
       options.hasExperimentalFeature(.onTypeFormatting)
       ? DocumentOnTypeFormattingOptions(triggerCharacters: ["\n", "\r\n", "\r", "{", "}", ";", ".", ":", "#"])
@@ -1130,6 +1140,7 @@ extension SourceKitLSPServer {
       ),
       hoverProvider: .bool(true),
       completionProvider: completionOptions,
+      signatureHelpProvider: signatureHelpOptions,
       definitionProvider: .bool(true),
       implementationProvider: .bool(true),
       referencesProvider: .bool(true),
@@ -1637,6 +1648,14 @@ extension SourceKitLSPServer {
     languageService: LanguageService
   ) async throws -> HoverResponse? {
     return try await languageService.hover(req)
+  }
+
+  func signatureHelp(
+    _ req: SignatureHelpRequest,
+    workspace: Workspace,
+    languageService: LanguageService
+  ) async throws -> SignatureHelp? {
+    return try await languageService.signatureHelp(req)
   }
 
   /// Handle a workspace/symbol request, returning the SymbolInformation.
--- a/Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
+++ b/Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
@@ -1,0 +1,82 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKLogging
+import SourceKitLSP
+import SwiftSyntax
+
+private class StartOfArgumentFinder: SyntaxAnyVisitor {
+  let requestedPosition: AbsolutePosition
+  var resolvedPosition: AbsolutePosition?
+
+  init(position: AbsolutePosition) {
+    self.requestedPosition = position
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
+    if (node.position...node.endPosition).contains(requestedPosition) {
+      return .visitChildren
+    }
+    return .skipChildren
+  }
+
+  override func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+    return visit(arguments: node.arguments)
+  }
+
+  override func visit(_ node: SubscriptCallExprSyntax) -> SyntaxVisitorContinueKind {
+    return visit(arguments: node.arguments)
+  }
+
+  private func visit(arguments: LabeledExprListSyntax) -> SyntaxVisitorContinueKind {
+    guard (arguments.position...arguments.endPosition).contains(requestedPosition) else {
+      return .skipChildren
+    }
+
+    guard !arguments.isEmpty else {
+      self.resolvedPosition = arguments.position
+      return .skipChildren
+    }
+
+    for argument in arguments {
+      if (argument.position...argument.endPosition).contains(requestedPosition) {
+        if let trailingComma = argument.trailingComma,
+          requestedPosition >= trailingComma.endPositionBeforeTrailingTrivia
+        {
+          self.resolvedPosition = trailingComma.endPositionBeforeTrailingTrivia
+        } else {
+          self.resolvedPosition = argument.expression.positionAfterSkippingLeadingTrivia
+        }
+        return .visitChildren
+      }
+    }
+
+    return .skipChildren
+  }
+}
+
+extension SwiftLanguageService {
+  func adjustPositionToStartOfArgument(
+    _ position: Position,
+    in snapshot: DocumentSnapshot
+  ) async -> Position {
+    let tree = await self.syntaxTreeManager.syntaxTree(for: snapshot)
+    let visitor = StartOfArgumentFinder(position: snapshot.absolutePosition(of: position))
+    visitor.walk(tree)
+    if let resolvedPosition = visitor.resolvedPosition {
+      return snapshot.position(of: resolvedPosition)
+    }
+    return position
+  }
+}
--- a/Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
+++ b/Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
@@ -25,7 +25,7 @@ private class StartOfArgumentFinder: SyntaxAnyVisitor {
   }
 
   override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
-    if (node.position...node.endPosition).contains(requestedPosition) {
+    if node.range.contains(requestedPosition) {
       return .visitChildren
     }
     return .skipChildren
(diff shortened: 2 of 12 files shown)


Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -136,6 +136,10 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     return req.item
   }
 
+  package func signatureHelp(_ req: SignatureHelpRequest) async throws -> SignatureHelp? {
+    nil
```
ahoppen: Let’s make this throw an error instead because the request hasn’t been implemented


```suggestion
    throw ResponseError.requestNotImplemented(SignatureHelpRequest.self)
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1048,6 +1050,14 @@ extension SourceKitLSPServer {
         triggerCharacters: [".", "("]
       )
 
+    let signatureHelpOptions =
+      await registry.clientHasDynamicSignatureHelpRegistration
+      ? nil
+      : LanguageServerProtocol.SignatureHelpOptions(
+        triggerCharacters: ["(", "["],
+        retriggerCharacters: [",", ":"]
```
ahoppen, line 1087: Do we need to re-trigger on `:`? You would still be in the same parameter after typing a `:`, right?

Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
```diff
@@ -0,0 +1,82 @@
...
+  let requestedPosition: AbsolutePosition
+  var resolvedPosition: AbsolutePosition?
+
+  init(position: AbsolutePosition) {
+    self.requestedPosition = position
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
+    if (node.position...node.endPosition).contains(requestedPosition) {
```
ahoppen: This can be simplified to 

```suggestion
    if node.range.contains(requestedPosition) {
```

Same below

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -0,0 +1,97 @@
...
+    else {
+      return nil
+    }
+
+    let activeParameter = signature[keys.activeParameter] as Int?
+    let parameters = skParameters.compactMap { ParameterInformation($0, keys) }
+
+    let documentation: StringOrMarkupContent? = signature[keys.docComment].map {
+      .markupContent(MarkupContent(kind: .markdown, value: $0))
+    }
```
ahoppen: I’m generally not a fan of using `map` on `Optional` because I always think of arrays when seeing `map`. I think this would read easier as

```suggestion
    let documentation: StringOrMarkupContent? = if let docComment: String = signature[keys.docComment] {
      .markupContent(MarkupContent(kind: .markdown, value: docComment))
    } else {
      nil
    }
```

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -0,0 +1,97 @@
...
+
+fileprivate extension ParameterInformation {
+  init?(_ parameter: SKDResponseDictionary, _ keys: sourcekitd_api_keys) {
+    guard let nameOffset = parameter[keys.nameOffset] as Int?,
+      let nameLength = parameter[keys.nameLength] as Int?
+    else {
+      return nil
+    }
+
+    self.init(label: .offsets(start: nameOffset, end: nameOffset + nameLength))
```
ahoppen: I suppose that sourcekitd will return the offsets in terms of UTF-8 byte offsets while LSP requires UTF-16 byte offsets, so you’ll need to do a UTF-8 to UTF-16 conversion here, similar to what we do in all the position conversion functions on `DocumentSnapshot`. I don’t remember a canonical example where this matters but emojis or other non-ASCII characters are usually a good way to trigger differences between UTF-8 and UTF-16.

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
```diff
@@ -0,0 +1,434 @@
...
+          position: positions[marker]
+        )
+      )
+
+      let signatureHelp = try XCTUnwrap(result)
+      let signature = try XCTUnwrap(signatureHelp.signatures.only)
+
+      XCTAssertEqual(signature.label, "add(first: Double!, second: Float, third: Int) -> Double", line: line)
+      XCTAssertNil(signature.documentation)
+      XCTAssertEqual(
```
ahoppen: I suggest passing `line` here as well so we know which test case caused the error when these assertions fail.

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
```diff
@@ -0,0 +1,434 @@
...
+        position: positions["1️⃣"]
+      )
+    )
+
+    let signatureHelp = try XCTUnwrap(result)
+    XCTAssertEqual(signatureHelp.activeSignature, 0)
+    XCTAssertEqual(signatureHelp.activeParameter, 0)
+    XCTAssertEqual(signatureHelp.signatures.count, 2)
+
+    let firstSignature = try XCTUnwrap(signatureHelp.signatures[0])
```
ahoppen: This will crash the test case if there are no signatures in the signature help. I suggest to change `XCTAssertEqual(signatureHelp.signatures.count, 2)` to

```swift
guard signatureHelp.signatures.count == 2 else {
  XCTFail("expected 2 signatures, got \(signatureHelp.signatures)")
}
```

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
```diff
@@ -0,0 +1,434 @@
...
+    let project = try await SwiftPMTestProject(
+      files: [
+        "utils.swift": """
+        /// A utility function that combines values
+        /// - Parameters:
+        ///   - first: The first value
+        ///   - second: The second value
+        /// - Returns: The combined result
+        func combine(first: String, second: Int) -> String {
+          return "\\(first)-\\(second)"
```
ahoppen: You can replace the `\\` with `\` if you make this a raw string literal starting with `#"""`

Sources/SourceKitLSP/SourceKitLSPServer.swift
a7medev: Not necessarily. Imagine a function `add(x: Int = 0, y: Int)`.
If the user types `add(y` that's the first parameter, but once a colon is added `add(y: |` we should switch to the second parameter

Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
a7medev: `range` doesn't contain `node.endPosition` which is fine in `visitAny` as we're only interested in `FunctionCallExprSyntax` and `SubscriptCallExprSyntax` and the cursor position is always within the call expression for a valid signature help invocation. I've updated that check.

For checks within `visit(arguments:)` though, we need to consider arguments even if the cursor is at the end of the argument (e.g. `test(x|)`) so we need the inclusive check.

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
a7medev: Done. ✅

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
a7medev: Makes sense. Fixed. ✅

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
a7medev: Yeah nice catch, I missed these 😅 Fixed. ✅

Sources/SwiftLanguageService/SignatureHelp.swift
a7medev: Thanks for pointing that out, I've fixed it and added a test case for it. Please recheck. 🙏🏼

Sources/SwiftLanguageService/SignatureHelp.swift
a7medev: Updated that. Please recheck. 🙏🏼

a7medev:
Here's a quick demo of signature help in VS Code.

https://github.com/user-attachments/assets/2208ff9f-c0c8-4668-9472-b6f2bc4cd274

There are some things to improve though:
1. Automatically triggering signature help after accepting a function or subscript completion item. We should ideally do that in vscode-swift by injecting the `editor.action.triggerParameterHints` command in completion items using a middleware. I plan on opening a PR for it once we get the basic implementation done.
2. Persisting the active signature chosen by the user across re-triggers.

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -0,0 +1,130 @@
...
+        """
+        UTF-8 offset is past the end of the string while getting UTF-16 offset of \(utf8Offset) \
+        (\(callerFile, privacy: .public):\(callerLine, privacy: .public))
+        """
+      )
+      return self.utf16.count
+    }
+    return self.utf16.distance(from: self.startIndex, to: stringIndex)
+  }
+}
```
a7medev, line 36: An alternative would've been to create a `LineTable` with the signature label and use its `utf16ColumnAt`.

I'm also not sure if it's worth moving this to `SwiftExtensions`.

a7medev:
I've opened a PR on vscode-swift (https://github.com/swiftlang/vscode-swift/pull/1802) to automatically trigger signature help on accepting a function-like completion item in VS Code.

Sources/SwiftLanguageService/SignatureHelp.swift
ahoppen: I think creating a `LineTable` is overkill and having to deal with lines might actually get in the way if for some reason the description of the completion item spans multiple lines. Also, keeping it here is good for now since we’re not using it anywhere else and I don’t expect us to need this anywhere anytime soon. If we do, we can always move it when we need to.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Oh, I see. That makes sense to me. Would you mind adding that as a comment as it’s not super obvious and I could imagine somebody stumbling across this in the future again.

Sources/SwiftLanguageService/AdjustPositionToStartOfArgument.swift
ahoppen: Good catch, I didn’t think about the cursor being at the end of the range. 👍🏽

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -0,0 +1,145 @@
...
+        activeParam
+      } else if !parameters.isEmpty {
+        // If we have parameters and no active parameter is present, we return
+        // an out-of-range index that way editors don't show an active parameter.
+        // As of LSP 3.17, not returning an active parameter defaults to choosing
+        // the first parameter which isn't the desired behavior.
+        // LSP 3.18 defines a `noActiveParameterSupport` option which allows the
+        // active parameter to be `null` causing editors not to show an active
+        // parameter which would be the best solution here.
+        parameters.count
```
ahoppen, line 86: Oh, that behavior is interesting. The LSP specification says the following. Does this mean that VS Code does not comply with the LSP spec in this regard?

> If omitted or the value lies outside the range of `signatures[activeSignature].parameters` defaults to 0 if the active signature has parameters

I’m fine keeping this in but we should mention the intended behavior from the LSP spec in the comment and mention that we are deliberately relying on a VS Code LSP-non-compliance if that’s what we’re doing.

ahoppen:
The screen recording looks great 🤩 

One thought that just crossed my mind to extract the documentation for the parameters: SourceKit-LSP does link against docc, so we should be able to use docc to parse the doc comment into a symbol graph and then extract the parameters from there. In general, getting docc into the loop here might be open up new possibilities such as https://github.com/swiftlang/sourcekit-lsp/issues/2256. If you want to tackle this, I would make that a separate PR though.

Sources/SwiftLanguageService/SignatureHelp.swift
a7medev: Yes, that's VS Code-specific behavior which we use as a workaround here. For editors that are fully LSP-compliant this will have the same behavior as if we didn't return a value and we actually don't have a way of expressing "no active parameter" in that case so it's fine. I actually found rust-analyzer using this trick too.

There are also other VS Code-specific workarounds like setting the active parameter to `-1` (https://github.com/microsoft/vscode/issues/145851#issuecomment-1078689732).

I'll update the comment to explain this part too.

a7medev:
@ahoppen Thanks for the suggestion, I'll look into it more as I'm not familiar with DocC.

I've updated the comments you mentioned as well, can you please recheck? 🙏🏼

a7medev:
@ahoppen I rebased the PR on main as it had some conflicts (due to moving default implementations from `DocumentationLanguageService` to `LanguageService` in #2261).

ahoppen:
Thanks for rebasing and sorry for causing the conflict.

ahoppen:
https://github.com/swiftlang/swift/pull/83378

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift/pull/8337

@swift-ci Please test Windows

a7medev:
@ahoppen I've implemented preserving the user's selected active signature in 02caee71c36b92695b7456d9a264f31cef24b1cb, can you please review this part as well? 🙏🏼 

Sorry for modifying this after the review.

a7medev:
@ahoppen Updated. Can you please recheck? 🙏🏼 
FYI, https://github.com/swiftlang/swift/pull/83378 has been merged.

ahoppen (approved):
LGTM. Let’s ship this 🚀

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

a7medev:
@ahoppen @hamishknight Tests are failing on Windows with errors similar to the following:

```
<EXPR>:0: error: SwiftSignatureHelpTests.testSignatureHelpEnumCase : threw error "ResponseError(code: LanguageServerProtocol.ErrorCode(rawValue: -32001), message: "No language service implements textDocument/signatureHelp", data: nil)"
````

Any idea why that is happening? Shouldn't these tests run fine on Windows?

bnbarham:
> Any idea why that is happening? Shouldn't these tests run fine on Windows?

I haven't checked the logs, but I see new files with no update to `CMakeLists.txt`, so my guess would be that is the cause. You'll want to update `Sources/SwiftLanguageService/CMakeLists.txt` with the new files

a7medev:
@bnbarham Interesting. I didn't notice that I needed to add that since it was working fine on my machine. Do we primarily use this `CMakeLists.txt` file on Windows? 👀 

I've added them now. Can you please rerun the tests? 🙏🏼

bnbarham:
> Do we primarily use this CMakeLists.txt file on Windows?

Yep :(

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

hamishknight:
🚢

bnbarham:
Could we add a `SkipUnless` similar to your completion PR, to make sure that the tests pass when we have an older sourcekitd 🙏?

a7medev:
@bnbarham Makes sense. I've opened #2284 to add that change.

PR #2254, opened 2025-08-24
Replace a couple of `#if` conditions by runtime checks

ahoppen:
Depends on #2249.

---

This ensures that we build all branches on all platforms, reducing the likelihood of build failures on other platforms.


ahoppen:
@swift-ci Please test

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -393,6 +393,7 @@ package actor SkipUnless {
     file: StaticString = #filePath,
     line: UInt = #line
   ) async throws {
+    #if os(Windows)  // docc is supported on Linux and Windows
```
bnbarham: Why is this needed? And presumably the comment is incorrect 🤔 (ie. should be macOS and Linux)?

ahoppen: Clarified the comment

> docc is supported on macOS and Linux. Do not perform a check on those platforms to avoid accidentally skipping the tests in CI on those platforms.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #2255, opened 2025-08-24
Support `trace-from-signpost` for logs captured in time zones ahead of UTC

ahoppen:
We need to match `+` in the date component of the log so we can capture times such as `2025-08-20 09:19:44.156241+0200`. Previously this only worked for times that were behind UTC.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2261, opened 2025-08-25
Add default implementations to `LanguageService` for methods that are not expected to be implemented by all language services

ahoppen:
With the introduction of secondary language services, we don’t expect every language service to implement every request anymore. To simplify the addition of language services like `DocumentationLanguageService` add default implementations for methods that satisfy the following criteria:
 - `SourceKitLSPServer` does not expect side effects to happen when they are called
 - The method can throw or there is a reasonable default value
 - It is reasonable to expect that not all language services need to implement it


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/DocumentationLanguageServiceTests.swift
ahoppen, line 1: These tests no longer worked since the DocumentationLanguageService now returns a `requestNotImplemented` for the completion request instead of returning empty results, which really makes more sense. These tests didn’t really provide much value, so I decided to remove them.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2264, opened 2025-08-28
Fix issue that causes type hierarchy to not work for clang files

ahoppen:
Clang’s `textDocument/symbolInfo` request does not return a `kind` field. We were thus filtering all symbols from clang, which left us with no USR that could be used for the type hierarchy.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2464,20 +2464,31 @@ extension SourceKitLSPServer {
     guard let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(checkedFor: .deletedFiles) else {
       return nil
     }
-    let usrs =
-      symbols
-      .filter {
-        // Only include references to type. For example, we don't want to find the type hierarchy of a constructor when
-        // starting the type hierarchy on `Foo()``.
-        switch $0.kind {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2464,21 +2464,32 @@ extension SourceKitLSPServer {
...
-        // starting the type hierarchy on `Foo()``.
-        switch $0.kind {
-        case .class, .enum, .interface, .struct: return true
-        default: return false
-        }
+    let usrs = symbols.filter {
+      // Only include references to type. For example, we don't want to find the type hierarchy of a constructor when
+      // starting the type hierarchy on `Foo()`.
+      // Consider a symbol a class if its kind is `nil`, eg. for a symbol returned by clang's SymbolInfo, which
+      // doesn't support the `kind` field.
```
bnbarham, line 2471: Should it 🤔? This is our extension isn't it? I'm fine with this change regardless, since we still filter from the index.

bnbarham (approved):

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Yeah, would probably be nice if we added the `kind` field to the clangd request.

PR #2265, opened 2025-08-28
When `PipeAsStringHandler` is destroyed, ensure that we flush any pending data

ahoppen:
Otherwise, we would not handle the last line if it didn’t end with a newline character.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2266, opened 2025-08-28
If a BSP server fails to initialize, show the error message to the user

ahoppen:
When the BSP server fails to initialize, it can’t give us any compiler arguments and we won’t be able to provide semantic functionality. Show the message to the user so they can do something about fixing it instead having to dig into SourceKit-LSP logs.


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2268, opened 2025-08-30
Log each read config

bnbarham:
There's been a couple issue where it is clear from the log of the global configuration that options are being set *somewhere*, but users then have to go searching in all the paths we lookup (which is quite a few). Log each config we read so it's easy to see where the options are coming from.


bnbarham:
@swift-ci please test

PR #2269, opened 2025-08-30
Pass objects to linker flags as is in macro test projects

bnbarham:
These are absolute paths, no "-l" is necessary here and is actually an error with eg. `ld.bfd`.


bnbarham:
@swift-ci please test

PR #2271, opened 2025-08-30
Add support for copied header files to SourceKit-LSP

ahoppen:
If a build server copies files (eg. header) to the build directory during preparation and those copied files are referenced for semantic functionality, we would currently jump to the file in the build directory. Teach SourceKit-LSP about files that are copied during preparation and if we detect that we are jumping to such a file, jump to the original file instead.

So far only the definition request checks the copied file paths. Adding support for copied file paths in the other requests will be a follow-up change.


ahoppen:
@swift-ci Please test

PR #2272, opened 2025-08-31
Add a ShowFocusedDiagnosticsRequest LSP extension and wire it up to inferred types remarks

owenv:
The goal here is to provide a general way for the compiler to provide on-demand visualizations/markup of source focused on a particular region of code the user is working in. Initially, I've wired this up to the remark added in https://github.com/swiftlang/swift/pull/84032 which emits remarks for each type inferred in an expression at the given position.

### DIFF
--- a/Contributor Documentation/LSP Extensions.md
+++ b/Contributor Documentation/LSP Extensions.md
@@ -770,6 +770,35 @@ export interface PeekDocumentsResult {
 }
 ```
 
+## `workspace/showFocusedDiagnostics`
+
+Request from the server to the client to display focused diagnostics for a specific subset of the source.
+
+It requires the experimental client capability `"workspace/showFocusedDiagnostics"` to use.
+
+- params: `ShowFocusedDiagnosticsParams`
+- result: `ShowFocusedDiagnosticsResult`
+
+```ts
+export interface ShowFocusedDiagnosticsParams {
+  /**
+   * Array of diagnostics to display
+   */
+  diagnostics: Diagnostic[];
+  /**
+   * The `DocumentUri` of the text document in which to present the diagnostics.
+   */
+  uri: DocumentUri;
+}
+
+/**
+ * Response to indicate the `success` of the `ShowFocusedDiagnosticsRequest`
+ */
+export interface ShowFocusedDiagnosticsResult {
+  success: boolean;
+}
+```
+
 ## `workspace/synchronize`
 
 Request from the client to the server to wait for SourceKit-LSP to handle all ongoing requests and, optionally, wait for background activity to finish.
--- a/Sources/SwiftLanguageService/FocusedRemarksCommand.swift
+++ b/Sources/SwiftLanguageService/FocusedRemarksCommand.swift
@@ -1,0 +1,101 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+package import LanguageServerProtocol
+import SourceKitD
+
+/// Describes a kind of focused remarks supported by the compiler. Remarks should be exposed via
+/// a flag which accepts a position in the document as a <line:column> pair and only emits
+/// diagnostics relavant to that position (e.g. expressions or function bodies with source ranges that
+/// contain the position).
+package enum FocusedRemarksKind: String, CaseIterable, Codable {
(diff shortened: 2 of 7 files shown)


Sources/SwiftLanguageService/FocusedRemarksCommand.swift
```diff
@@ -0,0 +1,101 @@
...
+  case showInferredTypes
+
+  package var defaultTitle: String {
+    switch self {
+    case .showInferredTypes:
+      return "Show Inferred Types"
+    }
+  }
+
+  package func additionalCompilerArgs(line: Int, column: Int) -> [String] {
```
ahoppen, line 30: I would prefer if we used `SourceKitDPosition` instead of `line` and `column` here to indicate whether we are talking about 1 or 0-based line/column and whether we use UTF-8 or UTF-16 columns. If you convert between `Position` and `SourceKitDPosition` using `DocumentSnapshot.sourcekitdPosition`, that should also take care of the UTF-16 to UTF-8 conversion that is currently not handled.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1119,13 +1135,51 @@ extension SwiftLanguageService {
...
     }
 
     return nil
   }
 
+  package func executeFocusedRemarksCommand(_ command: FocusedRemarksCommand) async throws {
+    let snapshot = try self.documentManager.latestSnapshot(command.textDocument.uri)
+    let buildSettings = await self.buildSettings(
+      for: command.textDocument.uri,
+      fallbackAfterTimeout: true
```
ahoppen, line 1151: I don’t think we should use fallback build settings after a timeout. The focused remarks are not going to be useful with fallback build settings and we don’t have a way to refresh them when we would get real build settings.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -189,4 +189,41 @@ final class ExecuteCommandTests: XCTestCase {
     req.arguments = [1, 2, "", metadata.encodeToLSPAny()]
     XCTAssertEqual([1, 2, ""], req.argumentsWithoutSourceKitMetadata)
   }
+
+  func testShowInferredTypesCommand() async throws {
```
ahoppen, line 193: Could you write a `SkipUnless` function that skips this test if the compiler doesn’t support `-Rinferred-types-at`? This is important because people might be developing SourceKit-LSP with toolchains that are older.

Tests/SourceKitLSPTests/ExecuteCommandTests.swift
```diff
@@ -189,4 +189,41 @@ final class ExecuteCommandTests: XCTestCase {
...
+    let metadata = SourceKitLSPCommandMetadata(textDocument: textDocument)
+    let request = ExecuteCommandRequest(
+      command: FocusedRemarksCommand.identifier,
+      arguments: [command.encodeToLSPAny(), metadata.encodeToLSPAny()]
+    )
+    let expectation = self.expectation(description: "Handle ShowFocusedDiagnosticsRequest")
+
+    testClient.handleSingleRequest { (req: ShowFocusedDiagnosticsRequest) in
+      expectation.fulfill()
+      XCTAssertEqual(req.diagnostics.map(\.message), ["integer literal was inferred to be of type 'Int'"])
```
ahoppen, line 223: Could you also check the diagnostic’s kind (ie. that it’s not an `error`). I think that’s kind of interesting as well.

Contributor Documentation/LSP Extensions.md
```diff
@@ -770,6 +770,35 @@ export interface PeekDocumentsResult {
...
+export interface ShowFocusedDiagnosticsParams {
+  /**
+   * Array of diagnostics to display
+   */
+  diagnostics: Diagnostic[];
+  /**
+   * The `DocumentUri` of the text document in which to present the diagnostics.
+   */
+  uri: DocumentUri;
+}
```
ahoppen, line 792: I’m not sure if diagnostics are the best currency here. If I understand you correctly, you want to show additional information for source ranges, so what do you think about changing this request to `textDocument/showAnnotations` with the following parameters

```ts
interface TextDocumentAnnotation {
  /**
   * The range of the document to which the annotation applies.
   */
  range: Range;

  /**
   * The annotation to show at the given range.
   */
  message: string;
}

export interface ShowTextDocumentAnnotationsParams {
  /**
   * The `DocumentUri` of the text document in which to present the annotations.
   */
  uri: DocumentUri;

  /**
   * Array of annotations to display
   */
  diagnostics: TextDocumentAnnotation[];
}
```

Description for the request could be something like

> Request from the server to the client to display supplementary annotations for a source file. 
> 
> The annotations should be displayed in a transient style that can be dismissed by the user when they are no longer of interest.

owenv:
thanks - I'll take a closer look at the feedback in the next week or so

owenv:
Not pursuing this in the near term

PR #2273, opened 2025-08-31
Improve `Indexing: Scheduling tasks` message

ahoppen:
The majority of the work here is getting the list of files to index from the build server. Let’s rename the message to reflect that.


ahoppen:
@swift-ci Please test

PR #2274, opened 2025-08-31
Fix issue that caused the index progress indicator to get stuck if there are no sources in a target

ahoppen:


ahoppen:
@swift-ci Please test

PR #2277, opened 2025-09-01
Add a timeout for `workspace/buildTargets` and `buildTarget/sources` requests

### LINKED ISSUE #2252: Help investigating root cause of SourceKit-LSP requests hanging when my Build Server processes requests
Hello,

I [started a discussion here](https://forums.swift.org/t/any-advice-for-a-sourcekit-lsp-bottleneck-using-the-build-server-protocol/81673) and was encouraged to file an issue for additional help.

I’m using [this Bazel build server BSP implementation](https://github.com/spotify/sourcekit-bazel-bsp) with SourceKit-LSP.

The issue I’m seeing is that during buildTarget/prepare requests, language features such as jump-to-definition, autocomplete, and syntax highlighting stop working until the request completes. I’m trying to understand why this happens and whether it’s expected.

Ideally, I’d like language features to remain responsive during buildTarget/prepare—even if the results are temporarily stale.

I’ve attached logs from a reproduction of the issue, along with an AI-generated summary (which may or may not be accurate).

Any guidance or pointers on what direction to investigate would be greatly appreciated. 

Thank you so much!

- [issue-summary.md](https://github.com/user-attachments/files/21957247/issue-summary.md)
- [log.txt](https://github.com/user-attachments/files/21957248/log.txt)

ahoppen:
This allows us to provide functionality based on fallback settings for unresponsive BSP servers.

Fixes #2252


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -670,31 +668,76 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+      //   know about. Even if we did, the build server now telling us about the target should not change the options of
+      //   the file within the target
+      // - cachedTargetSources: Similar to cachedAdjustedSourceKitOptions, we would not have requested sources for
+      //   targets that we didn't know about and if we did, they wouldn't be affected
+      self.cachedBuildTargets.set(request, to: newResult)
+    case .sourceFilesReceivedResultAfterTimeout(let request, let newResult):
+      changedTargets = Set(request.targets)
+
+      // Caches not invalidated:
+      // - cachedAdjustedSourceKitOptions: Same as for buildTargetsReceivedResultAfterTimeout.
```
bnbarham, line 731: Assuming that the files are different, we probably do need to clear options here? ie. there may now be added/removed files and thus we need to re-request options.

I also wonder if this is true for build targets - that they have changed could imply that new search paths are needed (or old removed). For SwiftPM this isn't relevant today (as it has the one folder that everything is built into), but for Bazel I imagine this is much more relevant?

And lastly, is it now possible that we have multiple target/source requests in flight in different orders?

ahoppen: > Assuming that the files are different, we probably do need to clear options here? ie. there may now be added/removed files and thus we need to re-request options.

I don’t think so. If we got built settings for a source files then that’s expected to be up-to-date until the build server sends us a `buildTarget/didChange` notification. In practice I don’t know how we would have gotten here since we won’t request the build settings for a file in a target unless we know that it’s part of the target but even if we did, and the build server was able to return build settings for the file while the `buildTarget/sources` was still in progress, we would expect that information to be correct until we receive a `buildTarget/didChange` notification. Does this make sense? It could be that I missed something here. Race conditions and cache invalidation are hard.

> I also wonder if this is true for build targets - that they have changed could imply that new search paths are needed (or old removed). For SwiftPM this isn't relevant today (as it has the one folder that everything is built into), but for Bazel I imagine this is much more relevant?

Similar argument here. If we get any build settings for a file, we expect that to be correct until the build server notifies us that they have changed. Consider the `workspace/buildTargets` as a pure getter request that shouldn’t modify the build server’s state in an observable way. Receiving information about the build targets won’t change the build settings of the files within them.

> And lastly, is it now possible that we have multiple target/source requests in flight in different orders?

No more than before. All the timeout handling happens within the closures of the `Cache.get` calls, so if there are multiple calls to the same target set, they will be served by the same `Task`, just like before this change. We could have concurrent request for two target sets that aren’t overlapping, eg. sources of targets A, B and another request for targets B, C but then we rely on the build server to give us consistent answers for target B in both requests.

bnbarham: > until the build server notifies us that they have changed

If the assumption is that we expect a `buildTarget/didChange`, then sounds good to me 👍

PR #2278, opened 2025-09-01
Explicitly implement `syntacticTestItems` in `DocumentationLanguageService`

ahoppen:
`SyntacticTestIndex` queries all language services for a document for its syntactic test items. This caused `syntacticDocumentTests not implemented in DocumentationLanguageService` to be logged for every Swift file in the project. Remove the default implementation of `syntacticDocumentTests` in `LanguageService` and instead provide an implementation that returns an empty array in `DocumentationLanguageService`.

While at it, also log the files for which unimplemented methods are called on `LanguageService`. This makes it easier to dermine why they are called.


ahoppen:
@swift-ci Please test

PR #2283, opened 2025-09-05
Compare against expected output for formatting tests

natecook1000:
The formatting request tests currently test for the specific list of changes returned by the server, but that list is tied to the differences algorithm used by the standard library, which isn't guaranteed to be stable. This change compares the expected result with a new string generated by applying the list of changes, which is stable.

_Note:_ This includes a function for applying the list of `TextEdit`s received from the server to a string, since I couldn't find an existing method with that purpose.


natecook1000:
This change should enable swiftlang/swift#83212 to land.

Sources/SKTestSupport/String+applyingTextEdits.swift
```diff
@@ -0,0 +1,65 @@
...
+
+package import LanguageServerProtocol
+
+extension String {
+  /// Returns a new string obtained by applying the given text edits to this
+  /// string.
+  package func applying(_ edits: [TextEdit]) -> String {
+    let sortedEdits = edits.sorted { lhs, rhs in
+      lhs.range.lowerBound < rhs.range.lowerBound
+    }
```
bnbarham: I would generally expect this to already be the case, is it not with the new algorithm? We should also just be able to go through this list back to front, applying the changes along the way. `RenameAssertions.swift` has a private function that does that already (`apply(edits:to:)`) - we could move that into `SKTestSupport` instead and use it here?

Tests/SourceKitLSPTests/FormattingTests.swift
```diff
@@ -25,15 +25,13 @@ final class FormattingTests: XCTestCase {
     let testClient = try await TestSourceKitLSPClient()
     let uri = DocumentURI(for: .swift)
 
-    let positions = testClient.openDocument(
-      """
+    let source = """
       struct S {
       1️⃣var foo: 2️⃣ 3️⃣Int
```
bnbarham: This applies for most of the changes - you can remove the markers if they aren't used any more

natecook1000: 🤦🏼‍♂️ Of course!

natecook1000:
@bnbarham Thanks for the notes! I removed my new function for applying changes and switched to the one in `RenameAssertions.swift`, and took out the markers where they are no longer needed.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

bnbarham (approved):

PR #2289, opened 2025-09-11
Standardize file paths when attempting to find toolchains

### LINKED ISSUE #2174: Language support not available on Windows
**Describe the bug**
Basically, no completion, intellisense, errors, etc. Nothing kinda works.

**To Reproduce**
Be on windows
Install swift with winget (6.1.0)
Install vscode
Install swift extension for vscode
Create a project (either with swift vscode commands or the CLI)
Open a .swift file
done, nothing works

**Expected behavior**
Having basic code completion, type inference and error reporting

**Environment**
 - OS: Windows 11 23H2
 - Swift version: Swift version 6.1 (swift-6.1-RELEASE) \
                          Target: x86_64-unknown-windows-msvc
 - Visual Studio Code version: 1.100.3
 - vscode-swift version: 2.3.20250521

**Additional context**
Image
Image
Image

and code is working with cli:
Image

till now, I was only able to "reliably" use swift with a proper lsp support was in an wsl ubuntu, and it was getting heavier and heavier (too much) with each line of code
maybe this all is related to the constant CPU hogging problem Image, but either way

bnbarham:
`containingXCToolchain` loops infinitely when given eg. `C:\foo\..\bar`. The underlying cause is that `deletingLastPathComponent` does not remove the `..` on Windows. On macOS it's potentially worse - it *adds* `..`. We don't see the infinite loop on macOS/Linux though because `AbsolutePath` already removes them (which is not the case on Windows).

Resolves #2174.

### DIFF
--- a/Sources/ToolchainRegistry/Toolchain.swift
+++ b/Sources/ToolchainRegistry/Toolchain.swift
@@ -366,10 +366,23 @@ public final class Toolchain: Sendable {
 }
 
 /// Find a containing xctoolchain with plist, if available.
-func containingXCToolchain(
+private func containingXCToolchain(
   _ path: URL
 ) -> (XCToolchainPlist, URL)? {
-  var path = path
+  // `deletingLastPathComponent` only makes sense on resolved paths (ie. those without symlinks or `..`). Any given
+  // toolchain path should have already been realpathed, but since this can turn into an infinite loop otherwise, it's
+  // better to be safe than sorry.
+  let resolvedPath = orLog("Toolchain realpath") {
+    try path.realpath
+  }
+  guard let resolvedPath else {
+    return nil
+  }
+  if path != resolvedPath {
+    logger.fault("\(path) was not realpathed")
+  }
+
+  var path = resolvedPath
   while !((try? path.isRoot) ?? true) {
     if path.pathExtension == "xctoolchain" {
       if let infoPlist = orLog("Loading information from xctoolchain", { try XCToolchainPlist(fromDirectory: path) }) {
--- a/Sources/ToolchainRegistry/ToolchainRegistry.swift
+++ b/Sources/ToolchainRegistry/ToolchainRegistry.swift
@@ -60,13 +60,13 @@ package final actor ToolchainRegistry {
   private let toolchainsByIdentifier: [String: [Toolchain]]
 
   /// The toolchains indexed by their path.
-  private let toolchainsByPath: [URL: Toolchain]
+  private var toolchainsByPath: [URL: Toolchain]
 
   /// Map from compiler paths (`clang`, `swift`, `swiftc`) mapping to the toolchain that contained them.
   ///
   /// This allows us to find the toolchain that should be used for semantic functionality based on which compiler it is
   /// built with in the `compile_commands.json`.
-  private let toolchainsByCompiler: [URL: Toolchain]
+  private var toolchainsByCompiler: [URL: Toolchain]
 
   /// The currently selected toolchain identifier on Darwin.
   package let darwinToolchainOverride: String?
@@ -96,17 +96,18 @@ package final actor ToolchainRegistry {
     var toolchainsByPath: [URL: Toolchain] = [:]
     var toolchainsByCompiler: [URL: Toolchain] = [:]
     for (toolchain, reason) in toolchainsAndReasonsParam {
+      // Toolchain should always be unique by path. It isn't particularly useful to log if we already have a toolchain
+      // though, as we could have just found toolchains through symlinks (this is actually quite normal - eg. OSS
+      // toolchains add a `swift-latest.xctoolchain` symlink on macOS).
+      if toolchainsByPath[toolchain.path] != nil {
+        continue
+      }
+
       // Non-XcodeDefault toolchain: disallow all duplicates.
       if toolchainsByIdentifier[toolchain.identifier] != nil,
         toolchain.identifier != ToolchainRegistry.darwinDefaultToolchainIdentifier
       {
         logger.error("Found two toolchains with the same identifier: \(toolchain.identifier)")
-        continue
-      }
-
-      // Toolchain should always be unique by path.
-      if toolchainsByPath[toolchain.path] != nil {
-        logger.fault("Found two toolchains with the same path: \(toolchain.path)")
         continue
       }
 
@@ -218,9 +219,14 @@ package final actor ToolchainRegistry {
       }
     }
 
-    let toolchainsAndReasons = toolchainPaths.compactMap {
-      if let toolchain = Toolchain($0.path) {
-        return (toolchain, $0.reason)
+    let toolchainsAndReasons = toolchainPaths.compactMap { toolchainAndReason in
+      let resolvedPath = orLog("Toolchain realpath") {
+        try toolchainAndReason.path.realpath
+      }
+      if let resolvedPath,
+        let toolchain = Toolchain(resolvedPath)
+      {
+        return (toolchain, toolchainAndReason.reason)
       }
       return nil
     }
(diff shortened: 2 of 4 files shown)


bnbarham:
@swift-ci please test

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -369,7 +369,7 @@ public final class Toolchain: Sendable {
 func containingXCToolchain(
   _ path: URL
 ) -> (XCToolchainPlist, URL)? {
-  var path = path
+  var path = path.standardizedFileURL
```
ahoppen: I would suggest using `.realpath` instead of `standardizedFileURL` to avoid the `/tmp` vs `/private/tmp` issue on macOS.

bnbarham:
> Also, we should check how this behaves if we discover two toolchains that have different paths (due to symlinks) but point to the same realpath. This is usually the case on macOS when you have open source toolchains installed and swift-latest.xctoolchain is a symlink to the latest installed toolchain. Ideally, we would only keep one copy of these toolchains in ToolchainRegistry now because they are identical.

FWIW this already happens today - they both have the same identifier, so we only use one and the other is skipped (with an error logged).

> I think this will cause issues if you have a toolchain with a symlink during the discovery phase and the BSP server returns a path that contains the symlink

Good point, I'll fix up the lookups.

> I would suggest using .realpath instead of standardizedFileURL to avoid the /tmp vs /private/tmp issue on macOS.

Yeah, this is the cause of the test failures at the moment (well, sort of, in that the test doesn't do a standardize and thus we get different /tmp vs /private/tmp).

bnbarham:
@swift-ci please test

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -366,10 +366,20 @@ public final class Toolchain: Sendable {
...
 /// Find a containing xctoolchain with plist, if available.
-func containingXCToolchain(
+private func containingXCToolchain(
   _ path: URL
 ) -> (XCToolchainPlist, URL)? {
-  var path = path
+  // `deletingLastPathComponent` only makes sense on resolved paths (ie. those without symlinks or `..`). Any given
+  // toolchain path should have already been realpathed, but since this can turn into an infinite loop otherwise, it's
+  // better to be safe than sorry.
+  guard let resolvedPath = try? path.realpath else {
```
ahoppen: Use `orLog` here so that we get some indication in case `realpath` fails, in which case we would only return `nil` here without logging the fault below?

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -219,7 +220,9 @@ package final actor ToolchainRegistry {
     }
 
     let toolchainsAndReasons = toolchainPaths.compactMap {
-      if let toolchain = Toolchain($0.path) {
+      if let resolvedPath = try? $0.path.realpath,
```
ahoppen: Same here, should we use `orLog`. I imagine this could help if the user has broken symlinks in their path. Or generally for all uses of `realpath`.

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -283,7 +286,18 @@ package final actor ToolchainRegistry {
   /// If we have a toolchain in the toolchain registry that contains the compiler with the given URL, return it.
   /// Otherwise, return `nil`.
   package func toolchain(withCompiler compiler: URL) -> Toolchain? {
-    return toolchainsByCompiler[compiler]
+    if let resolvedPath = try? compiler.realpath {
+      return toolchainsByCompiler[resolvedPath]
```
ahoppen: We should run `realpath` when populating `toolchainsByCompiler` as well to cover symlinks within the toolchain, eg. `swift` and `swiftc` to `swift-driver`.

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -283,7 +286,18 @@ package final actor ToolchainRegistry {
   /// If we have a toolchain in the toolchain registry that contains the compiler with the given URL, return it.
   /// Otherwise, return `nil`.
   package func toolchain(withCompiler compiler: URL) -> Toolchain? {
-    return toolchainsByCompiler[compiler]
+    if let resolvedPath = try? compiler.realpath {
```
ahoppen: This means that we need to run `realpath` and thus hit the filesystem for every entry in a JSON compilation database. Should we try and look up the non-realpath version first? And if we have a hit using the `realpath`, we can add a mapping for the original path to the toolchain as well as a cache. That way we only need to realpath every given path in a JSON compilation database once.

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -283,7 +286,18 @@ package final actor ToolchainRegistry {
...
+    }
+    return nil
+  }
+
+  /// If we have a toolchain in the toolchain registry with the given URL, return it. Otherwise, return `nil`.
+  package func toolchain(withPath path: URL) -> Toolchain? {
+    if let resolvedPath = try? path.realpath {
+      return toolchainsByPath[resolvedPath]
+    }
+    return nil
```
ahoppen: Similar to my comment on `toolchain(withCompiler:)`, should we try a non-realpath lookup first and then cache if necessary to avoid hitting the file system unnecessarily? E.g. something like

```swift
if let toolchain = toolchainsByPath[path] {
  return toolchain
}
guard let realpath = orLog("Realpath toolchain", { try path.realpath }) else {
  return nil
}
guard let toolchain = lchainsByPath[realpath] else {
  return nil
}
// Cache mapping of non-realpath to the realpath toolchain for faster subsequent lookups
toolchainsByPath[path] = toolchain
return toolchain
```

Sources/ToolchainRegistry/Toolchain.swift
bnbarham: Sure, I didn't want to be too spammy here but... 🤷

Sources/ToolchainRegistry/ToolchainRegistry.swift
bnbarham: Ah, good point. I didn't do this since we already realpathed the toolchain path coming into ToolchainRegistry, but forgot about the case the binaries themselves could be. We could just realpath the parent folder here?

EDIT: Hmmm, but we don't know the "toolchain" path in order to do that. I don't love the idea of realpathing to swift-driver though.

bnbarham: So I ended up going with just realpathing the parent dir, which while not "correct" ("bin" might itself be a symlink), it seems sufficiently unlikely to me for it not to matter. Happy to change if you disagree, though ideally we'd have a third solution as I'd really like to avoid realpathing the binaries themselves.

Sources/ToolchainRegistry/ToolchainRegistry.swift
bnbarham: Little sad to make all these mutable, but we'll likely need to do that even more in the future anyway (to allow registering toolchains as they're added) 👍

bnbarham:
@swift-ci please test

Sources/ToolchainRegistry/Toolchain.swift
ahoppen: I wouldn’t expect this to fails, so if it does, I want to know.

Sources/ToolchainRegistry/ToolchainRegistry.swift
ahoppen: I think it’s fair enough to assume that `bin` is not a symlink. If somebody starts symlinking `bin`, I really want to know what they are trying to achieve.

ahoppen (approved):
Looks good. Thank you!

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

PR #2291, opened 2025-09-12
Remove errant test

natecook1000:
Because of a difference in algorithms, the linear-space `CollectionDifference` algorithm includes some non-whitespace changes for this test example.

The current algorithm only identifies whitespace as differences, but the updated algorithm includes both a removal and insertion of the "A" character here: https://github.com/natecook1000/sourcekit-lsp/blob/46c99dabc4acf116a572fd58ae162a83b61e6cec/Tests/SourceKitLSPTests/FormattingTests.swift#L266

### DIFF
--- a/Tests/SourceKitLSPTests/FormattingTests.swift
+++ b/Tests/SourceKitLSPTests/FormattingTests.swift
@@ -284,7 +284,6 @@ final class FormattingTests: XCTestCase {
     let edits = try XCTUnwrap(response)
     let formattedSource = apply(edits: edits, to: source)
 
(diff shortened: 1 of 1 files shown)


natecook1000:
@swift-ci Please test

ahoppen (approved):
Hmm, unfortunate that we no longer report whitespace-only changes but fair enough. Maybe if we changed the test to have a little more real text than `A` would we get back to the whitespace-only changes. I’ll investigate once your collection diffing algorithm changes are in.

PR #2292, opened 2025-09-13
Separate signature and parameter documentation using swift-markdown

a7medev:
## Description

Separate signature and parameter documentation by parsing the documentation returned by sourcekitd using swift-markdown and extracting the parameter documentation.

## Alternatives considered

The parameter extraction implementation is almost ported from the implementation in the Swift compiler codebase. The problem with doing that in the Swift compiler codebase is that once you parse a the comment as markdown into a `Document` you cannot easily convert it back into markdown (we'd need to write our own markdown formatter). Besides, cmark doesn't handle Doxygen commands.

We considered using swift-docc but we faced some problems with it:

1. It's not easy to use. You need to retrieve a symbol graph (typically through a cursor info request) and use that along with other info to interact with DocC.
2. The result returned by DocC can't be directly converted to markdown, we'd need to provide our own DocC markdown renderer.

Implementing this using swift-markdown allows us to easily parse the comment, process it, convert it back to markdown and it also provides minimal parsing for Doxygen commands (we're only interested in `\param`) allowing us to use the same implementation for Clang-based declarations too.

Although this approach involves code duplication, it's works for our simple use case and we can probably improve it a bit in the future.

## Dependencies

1. https://github.com/swiftlang/swift-markdown/pull/237: Fixes a bug in swift-markdown where Doxygen commands are formatted without preceding newlines.
2. https://github.com/swiftlang/swift/pull/84272: Adds the parameter name to the signature's parameters allowing matching parameters with their documentation using their name.

### DIFF
--- a/Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
+++ b/Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
@@ -18,11 +18,15 @@ private struct Parameter {
   let documentation: String
 }
 
+/// Default markdown formatter options that output all Doxygen commands as separate paragraphs
+/// so they get rendered properly in markdown renderers that don't understand Doxygen.
+private let formatterOptions = MarkupFormatter.Options(adjacentDoxygenCommandsSpacing: .separateParagraphs)
+
 /// Extracts parameter documentation from a Doxygen parameter command.
 private func extractParameter(from doxygenParameter: DoxygenParameter) -> Parameter {
   return Parameter(
     name: doxygenParameter.name,
-    documentation: Document(doxygenParameter.blockChildren).format(),
+    documentation: Document(doxygenParameter.blockChildren).format(formatterOptions),
   )
 }
 
@@ -160,7 +164,7 @@ private func extractParameter(listItem: ListItem) -> Parameter? {
   let remainingFirstTextContent = String(components[1]).trimmingCharacters(in: .whitespaces)
   let remainingParagraphChildren = [Text(remainingFirstTextContent)] + paragraph.inlineChildren.dropFirst()
   let remainingChildren = [Paragraph(remainingParagraphChildren)] + listItem.blockChildren.dropFirst()
-  let documentation = Document(remainingChildren).format()
+  let documentation = Document(remainingChildren).format(formatterOptions)
 
   return Parameter(name: name, documentation: documentation)
 }
@@ -194,7 +198,7 @@ private func extractParameterWithRawIdentifier(from listItem: ListItem) -> Param
   let remainingParagraphChildren =
     [Text(remainingTextContent)] + paragraph.inlineChildren.dropFirst(2)
   let remainingChildren = [Paragraph(remainingParagraphChildren)] + listItem.blockChildren.dropFirst(1)
-  let documentation = Document(remainingChildren).format()
+  let documentation = Document(remainingChildren).format(formatterOptions)
 
   return Parameter(name: rawIdentifier.code, documentation: documentation)
 }
@@ -253,7 +257,7 @@ package func extractParametersDocumentation(
     }
   }
 
-  let remaining = Document(remainingBlocks).format()
+  let remaining = Document(remainingBlocks).format(formatterOptions)
 
   return (parameters: parameters, remaining: remaining)
 }


Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
...
+    guard let firstChild = listItem.child(at: 0) as? Paragraph,
+      let headingText = firstChild.child(at: 0) as? Text
+    else {
+      return false
+    }
+
+    let parametersPrefix = "parameters:"
+    let headingContent = headingText.string.trimmingCharacters(in: .whitespaces)
+
+    guard headingContent.lowercased().hasPrefix(parametersPrefix) else {
```
a7medev: We check for a prefix instead of `headingContent == parametersPrefix` to match the behavior in Xcode.

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test Windows

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test Windows

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test Windows

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen, line 1: Could you add a license header?

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
+import Foundation
+import Markdown
+
+private struct ParametersDocumentationExtractor {
```
ahoppen: Could you add the discussion in your PR about why we can’t (easily) use swift-docc here? I think that’s going to be a useful reference in the future as well.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
...
+  private var parameters = [String: String]()
```
ahoppen: Nitpick, we tend to prefer this spelling.

```suggestion
  private var parameters: [String: String] = [:]
```

(same in a couple more locations)

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -36,16 +36,23 @@ fileprivate extension String {
 }
 
 fileprivate extension ParameterInformation {
-  init?(_ parameter: SKDResponseDictionary, _ signatureLabel: String, _ keys: sourcekitd_api_keys) {
+  init?(
```
ahoppen, line 40: Could you add a short comment what the key and value of `parameterDocumentation` is?

Sources/SwiftLanguageService/SignatureHelp.swift
```diff
@@ -36,16 +36,23 @@ fileprivate extension String {
...
       return nil
     }
 
     let documentation: StringOrMarkupContent? =
-      if let docComment: String = parameter[keys.docComment] {
-        .markupContent(MarkupContent(kind: .markdown, value: docComment))
+      if let name = parameter[keys.name] as String?,
+        let documentation = parameterDocumentation[name]
+      {
+        .markupContent(MarkupContent(kind: .markdown, value: documentation))
```
ahoppen, line 56: Should we fall back to the entire doc comment if we can’t extract the parameter comment for some reason?

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
...
+  ///
+  /// Example:
+  /// ```markdown
+  /// - Parameters:
+  ///   - param: description
+  /// ```
+  ///
+  /// - Returns: True if the list item has parameter outline documentation, false otherwise.
+  private mutating func extractOutline(from listItem: ListItem) -> Bool {
+    guard let firstChild = listItem.child(at: 0) as? Paragraph,
```
ahoppen: Just checking: How does `child(at:)` behave if there is no element in the list? (Not sure why we would have parsed a list in that case but still). If it returns `nil`, then fine, otherwise, it would be nice to check the number of children in `listItem` and `firstChild`.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen: Couldn’t you just simplify this to the following? Seems easier for me to understand if all transformations happen in the same statement.

```suggestion
    guard headingText.string.trimmingCharacters(in: .whitespaces).lowercased().hasPrefix("parameters:") else {
```

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
...
+
+    for child in listItem.children {
+      guard let nestedList = child as? UnorderedList else {
+        continue
+      }
+
+      for nestedItem in nestedList.listItems {
+        if let parameter = extractOutlineItem(from: nestedItem) {
+          parameters[parameter.name] = parameter.documentation
+        }
```
ahoppen: A couple cases that I can think of for which it would be good to codify the behavior in a test case + add a short explanation. It would be good to cross-check the expected behavior with what docc generates:

---

IIUC we drop any elements in `Parameters:` that don’t look like paramters

```swift
/// - Parameters: 
///    - number: The number to do stuff with
///    - TODO Improve this documentation
func compute(number: Int) {}
```

--- 

Similarly, I think we’ll drop parameter documentation for all parameters that don’t actually exist in the signature. Similarly here, I don’t have super strong opinions on how we handle this but we should make our choice explicit with a test case + a short explanation.

```swift
/// - Parameters: 
///    - numberWithTypo: The number to do stuff with
func compute(number: Int) {}
```

---

If a parameter has two documentation comments.

```swift
/// - Parameters: 
///    - number: The number to do stuff with
///    - number: The number to do amazing stuff with
func compute(number: Int) {}
```

Tests/SourceKitLSPTests/SwiftSignatureHelpTests.swift
```diff
@@ -650,8 +657,14 @@ final class SwiftSignatureHelpTests: XCTestCase {
...
-        ParameterInformation(label: .offsets(start: 8, end: 21)),
-        ParameterInformation(label: .offsets(start: 23, end: 34)),
+        ParameterInformation(
+          label: .offsets(start: 8, end: 21),
+          documentation: .markupContent(MarkupContent(kind: .markdown, value: "The first value"))
+        ),
+        ParameterInformation(
+          label: .offsets(start: 23, end: 34),
+          documentation: .markupContent(MarkupContent(kind: .markdown, value: "The second value"))
+        ),
```
ahoppen, line 668: Maybe we have this already somewhere but if not, could you add a test case that uses external + internal parameter names, eg.

```swift
/// - number: The number to add 1 to
func addOne(to number: Int) -> Int { number + 1 }
```

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen: Instead of modifying a (global-ish, at least shared within this struct) member inside the struct, I wonder if it would be easier to read and debug if each of the `extract` returned the parameters that it extracted from the markdown it was given. That would follow the paradigm of immutable data structures more closely.

ahoppen (commented):
Very cool. I would have preferred to use docc here but maybe this is the simpler approach. 

Do you know how hard it would be to get the symbol graph for a completion item, similar to how we get the documentation comment of a completion item. I’m just wondering that maybe it wouldn’t be all to hard to do so. But I haven’t tried.

a7medev:
> Do you know how hard it would be to get the symbol graph for a completion item, similar to how we get the documentation comment of a completion item. I’m just wondering that maybe it wouldn’t be all to hard to do so. But I haven’t tried.

@ahoppen That should be doable using a cursor info request provided the USR I think. There are other things that DocC needs (and I'm not entirely aware of what each of them is used for/represents). Besides, conversion from DocC response to markdown would require implementing a renderer within SourceKit-LSP (maybe that should be done in the long-term anyway but I'm not sure if it fits in the timeline here).

I can attempt to use DocC again and see how it goes though. Last time I tried it wasn't a very pleasant experience. 😅

---

QQ, tests are failing on Windows due to issues with `CMakeLists.txt`, is there any documentation for how the CMake configuration with Swift is structured and how that maps to the information in `Package.swift`? It'd also be great if I can test the build locally on macOS before running the Windows tests (which take too long), is that doable?

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: Should I add the PR link or the full discussion? 👀

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: It does return `nil` when the index passed is ≥ the children count.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,192 @@
...
+    let components = firstTextContent.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+
+    guard components.count == 2 else {
+      return nil
+    }
+
+    let name = String(components[0]).trimmingCharacters(in: .whitespaces)
+    guard !name.isEmpty else {
+      return nil
+    }
```
hamishknight: What about raw identifiers e.g ``- `foo bar`: hello``?

Sources/SwiftLanguageService/SignatureHelp.swift
a7medev: Do you mean `parameter[keys.docComment]` or do you mean actually returning the entire signature documentation (without separating parameters) when we don't find the signature documentation one of the parameters?

For `parameter[keys.docComment]`, sourcekitd doesn't currently return anything there so it shouldn't make a difference at the moment and if we ever change sourcekitd to actually return a doc comment for parameters we should probably also update SourceKit-LSP to use that instead of having its own parameter extraction logic.

For option 2, I don't think that would be consistent with existing implementation of signature help, some parameters may not need specific explanation and thus we shouldn't show other parameters' documentation instead here.

WDYT?

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: Added, can you please recheck? 🙏🏼

a7medev:
@ahoppen @hamishknight I added the discussion for why we didn't use swift-docc, eliminated mutable state in `ParametersDocumentationExtractor`, added support for raw identifiers, and added more tests. Can you please recheck? 🙏🏼

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test Windows

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen: The way you added the full discussion is great. It means one less click for somebody to retrieve the information.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen: Great. Thanks for checking.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,267 @@
...
+///    providing the required arguments.
+/// 2. The result returned by DocC can't be directly converted to markdown, we'd need to provide our own DocC markdown renderer.
+///
+/// Implementing this using `swift-markdown` allows us to easily parse the comment, process it, convert it back to markdown.
+/// It also provides minimal parsing for Doxygen commands (we're only interested in `\param`) allowing us to use the same
+/// implementation for Clang-based declarations.
+///
+/// Although this approach involves code duplication, it's simple enough for the initial implementation. We should consider
+/// `swift-docc` in the future.
+private struct ParametersDocumentationExtractor {
```
ahoppen: If the `ParametersDocumentationExtractor` struct no longer has any members, we should be able to make all the functions top level, right? I think that would read a little nicer and also removes one level of indentation.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    let parameterPrefix = "parameter "
+    let paragraphContent = paragraphText.string
+
+    guard paragraphContent.count >= parameterPrefix.count else {
+      return nil
+    }
+
+    let prefixEnd = paragraphContent.index(paragraphContent.startIndex, offsetBy: parameterPrefix.count)
```
ahoppen: I think this can be combined into

```suggestion
    guard let prefixEnd = paragraphContent.index(paragraphContent.startIndex, offsetBy: parameterPrefix.count, limitedBy: paragraphContent.endIndex) else {
      return nil
    }
```

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
```diff
@@ -0,0 +1,267 @@
...
+
+  /// Extracts a parameter field from a list item provided the relevant first text content allowing reuse in ``extractOutlineItem`` and ``extractSingle``
+  ///
+  /// - Parameters:
+  ///   - firstTextContent: The content of the first text child of the list item's first paragraph
+  ///   - listItem: The list item to extract the parameter from
+  ///   - single: Whether the parameter is a single parameter or part of a parameter outline
+  ///
+  /// - Returns: A tuple containing the parameter name and documentation if a parameter was found, nil otherwise.
+  private func extractParam(
```
ahoppen: There are a couple minor-ish things that I find confusing with this implementation:
- `firstTextContent` kind of repeats some information that is already present in `listItem`, just in a slightly modified form. It’s not trivial to reason about when the first text in `listItem` should be used and when `firstTextContent`. If it’s possible, I would prefer if we could modify `listItem` in the `extractSingle` case to strip away the `parameter: ` prefix. Alternatively, I would either re-implement the `parameter: ` stripping here (I guess it shouldn’t be too difficult) or add documentation that explains exactly how `firstTextContent` and `listItem` relate to each other.
- `extractRawIdentifier` takes a completely different code path than the non-raw identifier path, which makes it easy for bugs to creep in in one of the paths that aren’t present in any of the others. Would it be possible to unify the two implementations so that both use the technique using `split(separator:)` or the technique of eating elements from the markdown structure until they hit a `:`?

Sources/SwiftLanguageService/SignatureHelp.swift
ahoppen: Looking at your reply and the code again, my comment didn’t really make sense. Forget what I said.

Tests/SourceKitLSPTests/ParametersDocumentationExtractorTests.swift
```diff
@@ -0,0 +1,350 @@
...
+        - TODO Improve this documentation
+      """
+
+    let (parameters, remaining) = extractParametersDocumentation(from: comment)
+
+    XCTAssertEqual(parameters, ["number": "The number to do stuff with"])
+    XCTAssertTrue(remaining.isEmpty)
+  }
+
+  /// Tests that we drop duplicate parameter documentation and keep thefirst one. Aligns with swift-docc.
```
ahoppen:
```suggestion
  /// Tests that we drop duplicate parameter documentation and keep the first one. Aligns with swift-docc.
```

ahoppen:
> That should be doable using a cursor info request provided the USR I think. There are other things that DocC needs (and I'm not entirely aware of what each of them is used for/represents). Besides, conversion from DocC response to markdown would require implementing a renderer within SourceKit-LSP (maybe that should be done in the long-term anyway but I'm not sure if it fits in the timeline here).

Definitely out-of-scope for this PR, I would just like to explore future options so we can decide whether it is something we want to investigate in the future and file an actionable issue. Do you know if it would be possible to get the symbol graph of a completion item without having to go through a cursor info (@hamishknight, maybe you know)? Because if it was, I think the main thing to do would be to implement the symbol graph to Markdown renderer, which while being a bit of work might come in handy in the future anyway.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: Makes sense. I made all members top-level and renamed the file to `ExtractParametersDocumentation.swift` since `ParametersDocumentationExtractor` doesn't exist anymore.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: Updated. Please recheck.

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
a7medev: > * `firstTextContent` kind of repeats some information that is already present in `listItem`, just in a slightly modified form. It’s not trivial to reason about when the first text in `listItem` should be used and when `firstTextContent`. If it’s possible, I would prefer if we could modify `listItem` in the `extractSingle` case to strip away the `parameter: ` prefix. Alternatively, I would either re-implement the `parameter: ` stripping here (I guess it shouldn’t be too difficult) or add documentation that explains exactly how `firstTextContent` and `listItem` relate to each other.

I removed the `firstTextContent` parameter and passed a modified `ListItem` instead in `extractSingleParameter` as you suggested.

> * `extractRawIdentifier` takes a completely different code path than the non-raw identifier path, which makes it easy for bugs to creep in in one of the paths that aren’t present in any of the others. Would it be possible to unify the two implementations so that both use the technique using `split(separator:)` or the technique of eating elements from the markdown structure until they hit a `:`?

The problem is the markdown structure is different for raw vs. normal parameter name:
1. Normal parameter names exist within a `Text` node as the first child of `Paragraph` and you can find the name by simply splitting `Text` around `:`.
2. Raw parameter names split parameter documentation into multiple inline markup nodes:
    1. In a parameter outline, the structure is: `InlineCode` (name), `Text` (": " + part of the parameter documentation).
    2. In a single parameter, the structure is: `Text` ("parameter "), `InlineCode` (name), `Text` (": " + part of parameter documentation).

That makes merging these into a single flow a bit difficult. Parsing until a `:` is found can also cause issues depending on the implementation:
1. If we just format the list item and look for the first `:` anywhere in the formatted string it'd cause problems with raw identifiers containing a colon (e.g. ``- `interesting: param`: abc``).
2. If we specifically look for the first `Text` item containing a colon, we'd still need to enforce the valid parameter structures. (e.g. ``- `foo` **bar**: baz`` isn't a valid parameter). It's not impossible, but it'd still have some complexity.

I'm curious to know what you think.

Tests/SourceKitLSPTests/ParametersDocumentationExtractorTests.swift
a7medev: Updated. Please recheck.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/// Extracts parameter documentation from a single parameter.
+///
+/// Example:
+/// ```markdown
+/// - Parameter param: description
+/// ```
+///
+/// - Returns: True if the list item has single parameter documentation, false otherwise.
```
ahoppen: Sorry for not noticing this before but I think the doc comment is out-of-date. Same for some other `Returns` clauses as well.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
```diff
@@ -0,0 +1,257 @@
...
+  var remainingParagraph = paragraph
+  remainingParagraph.replaceChildrenInRange(0..<1, with: [Text(remainingContent)])
+
+  var remainingListItem = listItem
+  remainingListItem.replaceChildrenInRange(0..<1, with: [remainingParagraph])
+
+  return extractParameter(listItem: remainingListItem, single: true)
+}
+
+/// Extracts a parameter field from a list item provided the relevant first text content allowing reuse in ``extractParametersOutline`` and ``extractSingleParameter``
```
ahoppen: Looks like this comment is out-of-date now.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
```diff
@@ -0,0 +1,257 @@
...
+
+  let potentialMatch = paragraphContent[..<prefixEnd].lowercased()
+
+  guard potentialMatch == parameterPrefix else {
+    return nil
+  }
+
+  let remainingContent = String(paragraphContent[prefixEnd...]).trimmingCharacters(in: .whitespaces)
+
+  var remainingParagraph = paragraph
```
ahoppen, line 123: I think a short comment saying something like the following would be helpful here.

> Remove the `Parameter: ` prefix from the list item from the list item so we can extract the parameter’s documentation using `extractParameter`

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
```diff
@@ -0,0 +1,257 @@
...
+/// - Parameters:
+///   - `foo bar`: documentation
+/// ```
+///
+/// - Parameters:
+///   - listItem: The list item to extract the parameter from
+///   - single: Whether the parameter is a single parameter or part of a parameter outline
+private func extractParameterWithRawIdentifier(from listItem: ListItem, single: Bool) -> Parameter? {
+  /// The index of ``InlineCode`` for the raw identifier parameter name in the first paragraph of ``listItem``
+  let inlineCodeIndex = single ? 1 : 0
```
ahoppen: Could we avoid needing to passing through the `single` parameter if you change `extractSingleParameter` to do something like the following? I think that would be nicer because the markdown structure of single and multiple parameters would be more similar this way.

```swift
if remainingContent.isEmpty {
  remainingParagraph.replaceChildrenInRange(0..<1, with: [])
} else {
  remainingParagraph.replaceChildrenInRange(0..<1, with: [Text(remainingContent)])
}
```

Sources/SwiftLanguageService/ParametersDocumentationExtractor.swift
ahoppen: Yeah, I guess with your explanation I agree that having two entirely different code paths for the documentation extraction is the best way to do it. A little unfortunate but that’s the way it is.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
a7medev: Nice suggestion, I've updated the code, please recheck. 🙏🏼

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
a7medev: Updated.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
a7medev: Yeah, sorry I missed that. Updated.

Sources/SwiftLanguageService/ExtractParametersDocumentation.swift
a7medev: Sorry I missed this one. Updated. 🙏🏼

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test Windows

a7medev:
https://github.com/swiftlang/swift-markdown/pull/237
https://github.com/swiftlang/swift/pull/84272

@swift-ci please test Windows

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test Windows

a7medev:
@ahoppen @hamishknight @bnbarham The Windows build is still failing even after linking `SwiftMarkdown` in SwiftLanguageService's `CMakeLists.txt` and the modifications in https://github.com/swiftlang/swift/pull/84302 producing the following error.

```
Building for production...
[0/28] Write sources
[18/28] Copying INPUTS
[19/28] Compiling CSKTestSupport CSKTestSupport.c
[20/28] Write swift-version-773C61ADED4095D8.txt
[21/28] Copying INPUTS
[23/30] Compiling CompletionScoringTestSupport RepeatableRandomNumberGenerator.swift
[24/32] Compiling CompletionScoringPerfTests CandidateBatchPerfTests.swift
[25/32] Compiling CompletionScoringTests BinaryCodingTests.swift
[26/32] Compiling SKTestSupport Assertions.swift
<unknown>:0: error: missing required module 'CAtomic'

error: fatalError
Error: Error: swift exited with code 1.
Invocation:
  swift test --scratch-path T:\x86_64-unknown-windows-msvc\SourceKitLSPTests --package-path C:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\sourcekit-lsp -c release -debug-info-format none --parallel -Xcc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\swift-corelibs-libdispatch -Xcc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\swift-corelibs-libdispatch\src\BlocksRuntime -Xswiftc -IT:\5\lib\swift\host -Xswiftc -LT:\5\lib\swift\host -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\cmark\src\include -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\cmark\extensions\include -Xlinker -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\cmark\extensions\include -Xlinker T:\x86_64-unknown-windows-msvc\cmark-gfm-0.29.0.gfm.13\src\cmark-gfm.lib -Xlinker T:\x86_64-unknown-windows-msvc\cmark-gfm-0.29.0.gfm.13\extensions\cmark-gfm-extensions.lib -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\swift-system\Sources\CSystem\include -Xswiftc -IT:\x86_64-unknown-windows-msvc\System\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\System\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\ToolsSupportCore\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\ToolsSupportCore\lib -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\llbuild\products\libllbuild\include -Xswiftc -IT:\x86_64-unknown-windows-msvc\LLBuild\products\llbuildSwift -Xlinker -LT:\x86_64-unknown-windows-msvc\LLBuild\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\ArgumentParser\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\ArgumentParser\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\Crypto\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\Crypto\lib -Xlinker T:\x86_64-unknown-windows-msvc\Crypto\lib\libCCryptoBoringSSL.lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\ASN1\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\ASN1\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\PackageManager\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\PackageManager\lib -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\swift-markdown\Sources\CAtomic\inclde -Xlinker T:\x86_64-unknown-windows-msvc\Markdown\lib\CAtomic.lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\Markdown\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\Markdown\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\Format\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\Format\lib -Xswiftc -IT:\x86_64-unknown-windows-msvc\IndexStoreDB\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_CIndexStoreDB\CIndexStoreDB.lib -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_Core\Core.lib -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_Database\Database.lib -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_Index\Index.lib -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_LLVMSupport\LLVMSupport.lib -Xlinker T:\x86_64-unknown-windows-msvc\IndexStoreDB\Sources\IndexStoreDB_Support\Support.lib -Xlinker T:\x86_64-unknown-windows-msvc\LMDB\lib\CLMDB.lib -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\sourcekit-lsp\Sources\CAtomics\include -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\sourcekit-lsp\Sources\CSourcekitd\include -Xlinker T:\x86_64-unknown-windows-msvc\SourceKitLSP\lib\CSourcekitd.lib -Xswiftc -IC:\Users\swift-ci\jenkins\workspace\sourcekit-lsp-PR-windows\sourcekit-lsp\Sources\CCompletionScoring\include -Xswiftc -IT:\x86_64-unknown-windows-msvc\SourceKitLSP\swift -Xlinker -LT:\x86_64-unknown-windows-msvc\SourceKitLSP\lib
```

After some investigation, it can't find `CAtomic` (singular) which is a target defined in swift-markdown (not to be confused with SourceKit-LSP's `CAtomics` target). I checked swift-format's `CMakeLists.txt` files configuration for swift-markdown and it doesn't add any special handling for `CAtomic` so I'm not sure what should be done here.

Do you have any specific suggestions here? Thanks in advance.

hamishknight:
Ah looks like there's a typo in build.ps1:

https://github.com/swiftlang/swift/blob/aa1f0009a1fc3ca9796807997fb2b92f066ebd49/utils/build.ps1#L3742

Should be `include` not `inclde`. Note also the error here is related to the SwiftPM build, not CMake build

hamishknight (approved):

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test Windows

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test macOS
@swift-ci please test Linux

a7medev:
https://github.com/swiftlang/swift/pull/84302
https://github.com/swiftlang/swift-markdown/pull/237

@swift-ci please test macOS

a7medev:
https://github.com/swiftlang/swift/pull/84302

@swift-ci please test

PR #2293, opened 2025-09-14
Support indexing of mulitple Swift files within the same compiler invocation

ahoppen:
Since the indexing performance of most projects is bound by target preparation instead of actual indexing right now, I wasn’t able to measure any performance benefits of this. But it seems like a good thing to add regardless, because it might save CPU cycles.

Fixes https://github.com/swiftlang/sourcekit-lsp/issues/1268

### DIFF
--- a/Contributor Documentation/BSP Extensions.md
+++ b/Contributor Documentation/BSP Extensions.md
@@ -130,8 +130,12 @@ export interface SourceKitSourceItemData {
   kind?: "source" | "header" | "doccCatalog";
 
   /**
-   * The output path that is during indexing for this file, ie. the `-index-unit-output-path`, if it is specified
-   * in the compiler arguments or the file that is passed as `-o`, if `-index-unit-output-path` is not specified.
+   * The output path that is a string that uniquely identifies the index output of this file in this target. If an index
+   * store should be re-used between build and background indexing, it must match the `-o` path or
+   * `-index-unit-output-path` used during the build.
+   *
+   * The index unit output path historically matched the path used for `-o` during compilation but has since evolved to
+   * be an opaque string. In particular, it does not have to match to any file on disk.
    *
    * This allows SourceKit-LSP to remove index entries for source files that are removed from a target but remain
    * present on disk and to index a file that is part of multiple targets in the context of each target.
--- a/Sources/BuildServerIntegration/BuildSettingsLogger.swift
+++ b/Sources/BuildServerIntegration/BuildSettingsLogger.swift
@@ -12,6 +12,7 @@
 
 package import LanguageServerProtocol
 package import SKLogging
+import SwiftExtensions
 
 // MARK: - Build settings logger
 
@@ -27,11 +28,31 @@ package actor BuildSettingsLogger {
     Self.log(level: level, settings: settings, for: uri)
   }
 
-  /// Log the given build settings.
+  /// Log the given build settings for a single file
   ///
   /// In contrast to the instance method `log`, this will always log the build settings. The instance method only logs
   /// the build settings if they have changed.
   package static func log(level: LogLevel = .default, settings: FileBuildSettings, for uri: DocumentURI) {
+    log(level: level, settings: settings, for: [uri])
+  }
+
+  /// Log the given build settings for a list of source files that all share the same build settings.
+  ///
+  /// In contrast to the instance method `log`, this will always log the build settings. The instance method only logs
+  /// the build settings if they have changed.
+  package static func log(level: LogLevel = .default, settings: FileBuildSettings, for uris: [DocumentURI]) {
+    let header: String
+    if let uri = uris.only {
+      header = "Build settings for \(uri.forLogging)"
+    } else if let firstUri = uris.first {
+      header = "Build settings for \(firstUri.forLogging) and \(firstUri) and \(uris.count - 1) others"
+    } else {
+      header = "Build settings for empty list"
+    }
+    log(level: level, settings: settings, header: header)
+  }
+
+  private static func log(level: LogLevel = .default, settings: FileBuildSettings, header: String) {
     let log = """
       Compiler Arguments:
       \(settings.compilerArguments.joined(separator: "\n"))
@@ -47,7 +68,7 @@ package actor BuildSettingsLogger {
       logger.log(
         level: level,
         """
-        Build settings for \(uri.forLogging) (\(index + 1)/\(chunks.count))
+        \(header) (\(index + 1)/\(chunks.count))
         \(chunk)
         """
       )
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -242,15 +242,7 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
         // Disjoint sets of files can be indexed concurrently.
         return nil
       }
-      if self.filesToIndex.count < other.filesToIndex.count {
-        // If there is an index operation with more files already running, suspend it.
-        // The most common use case for this is if we schedule an entire target to be indexed in the background and then
-        // need a single file indexed for use interaction. We should suspend the target-wide indexing and just index
-        // the current file to get index data for it ASAP.
-        return .cancelAndRescheduleDependency(other)
-      } else {
-        return .waitAndElevatePriorityOfDependency(other)
-      }
+      return .waitAndElevatePriorityOfDependency(other)
     }
   }
 
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -88,6 +88,26 @@ package struct FileAndOutputPath: Sendable, Hashable {
   fileprivate var sourceFile: DocumentURI { file.sourceFile }
 }
 
+/// A single file or a list of files that should be indexed in a single compiler invocation.
+private enum UpdateIndexStorePartition {
+  case multipleFiles(filesAndOutputPaths: [(file: FileToIndex, outputPath: String)], buildSettings: FileBuildSettings)
+  case singleFile(file: FileAndOutputPath, buildSettings: FileBuildSettings)
+
+  var buildSettings: FileBuildSettings {
+    switch self {
+    case .multipleFiles(_, let buildSettings): return buildSettings
+    case .singleFile(_, let buildSettings): return buildSettings
+    }
+  }
+
+  var files: [FileToIndex] {
+    switch self {
+    case .multipleFiles(let filesAndOutputPaths, _): return filesAndOutputPaths.map(\.file)
+    case .singleFile(let file, _): return [file.file]
+    }
+  }
+}
+
 /// Describes a task to index a set of source files.
 ///
 /// This task description can be scheduled in a `TaskScheduler`.
@@ -264,6 +284,16 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
       )
     }
 
+    // Compute the partitions within which we can perform multi-file indexing. For this we gather all the files that may
+    // only be indexed by themselves in `partitions` an collect all other files in `fileInfosByBuildSettings` so that we
+    // can create a partition for all files that share the same build settings.
+    // In most cases, we will only end up with a single partition for each `UpdateIndexStoreTaskDescription` since
+    // `UpdateIndexStoreTaskDescription.batches(toIndex:)` tries to batch the files in a way such that all files within
+    // the batch can be indexed by a single compiler invocation. However, we might discover that two Swift files within
+    // the same target have different build settings in the build server. In that case, the best thing we can do is
+    // trigger two compiler invocations.
+    var swiftFileInfosByBuildSettings: [FileBuildSettings: [(file: FileToIndex, outputPath: String)]] = [:]
+    var partitions: [UpdateIndexStorePartition] = []
     for fileInfo in fileInfos {
       let buildSettings = await buildServerManager.buildSettings(
         for: fileInfo.mainFile,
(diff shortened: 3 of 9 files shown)


ahoppen:
@swift-ci Please test

Contributor Documentation/BSP Extensions.md
```diff
@@ -130,8 +130,12 @@ export interface SourceKitSourceItemData {
   kind?: "source" | "header" | "doccCatalog";
 
   /**
-   * The output path that is during indexing for this file, ie. the `-index-unit-output-path`, if it is specified
-   * in the compiler arguments or the file that is passed as `-o`, if `-index-unit-output-path` is not specified.
+   * The output path that is a string that uniquely identifies the index output of this file in this target. If an index
+   * store should be re-used between build and background indexing, it must match the `-o` path or
+   * `-index-unit-output-path` used during the build.
+   *
+   * The index unit output path historically matched th path used for `-o` during compilation but has since evolved to
```
bnbarham:
```suggestion
   * The index unit output path historically matched the path used for `-o` during compilation but has since evolved to
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -264,72 +276,150 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
       )
     }
 
+    // Compute the partitions within which we can perform multi-file indexing. For this we gather all files tha may only
+    // be indexed by itself in `partitions` an collect all other files in `fileInfosByBuildSettings` so that we can
+    // create a partition for all files that share the same build settings.
```
bnbarham: "For this we gather all the files that may only be indexed by themselves...". Also no need to fix the comment nit picks in this PR.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -264,72 +276,150 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+    // In most cases, we will only end up with a single partition for each `UpdateIndexStoreTaskDescription` since
+    // `UpdateIndexStoreTaskDescription.batches(toIndex:)` tries to batch the files in a way such that all files within
+    // the batch can be indexed by a single compiler invocation. However, we might discover that two Swift files within
+    // the same target have different build settings in the build server. In that case, the best thing we can do is
+    // trigger two compiler invocations.
```
bnbarham, line 286: How would this happen 🤔? For C-family, sure. But I would expect all files within a target to have the same build settings for Swift.

It's super unfortunate we have to look through the arguments to find `-index-unit-output-path` for every settings per file, even though they should all be the same for every batch (except for the output path).

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -264,72 +276,150 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
         logger.error("Not indexing \(fileInfo.file.forLogging) because it has no compiler arguments")
         continue
       }
       if buildSettings.isFallback {
-        // Fallback build settings don’t even have an indexstore path set, so they can't generate index data that we would
-        // pick up. Also, indexing with fallback args has some other problems:∂
-        // - If it did generate a unit file, we would consider the file’s index up-to-date even if the compiler arguments
-        //   change, which means that we wouldn't get any up-to-date-index even when we have build settings for the file.
+        // Fallback build settings don’t even have an indexstore path set, so they can't generate index data that we
+        // would pick up. Also, indexing with fallback args has some other problems:∂
```
bnbarham:
```suggestion
        // would pick up. Also, indexing with fallback args has some other problems:
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -264,72 +276,150 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+            forSwiftFilesInPartition: partition,
+            buildSettings: partition.buildSettings,
             toolchain: toolchain
           )
         } catch {
-          logger.error("Updating index store for \(fileInfo.mainFile) failed: \(error.forLogging)")
-          BuildSettingsLogger.log(settings: buildSettings, for: fileInfo.mainFile)
+          logger.error(
+            """
+            Updating index store for failed: \(error.forLogging).
```
bnbarham:
```suggestion
            Updating index store failed: \(error.forLogging).
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -264,72 +276,150 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
       let startDate = Date()
-      switch buildSettings.language.semanticKind {
+      switch partition.buildSettings.language.semanticKind {
       case .swift:
         do {
           try await updateIndexStore(
-            forSwiftFile: fileInfo.mainFile,
-            buildSettings: buildSettings,
+            forSwiftFilesInPartition: partition,
+            buildSettings: partition.buildSettings,
```
bnbarham: Could just pass the partition rather than both the partition + build settings?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -340,9 +430,9 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
       if indexStorePath != buildServerIndexStorePath {
         logger.error(
           """
-          Compiler arguments for \(uri) specify index store path \(indexStorePath) but build server specified an \
+          Compiler arguments for specify index store path \(indexStorePath) but build server specified an \
```
bnbarham:
```suggestion
          Compiler arguments specify index store path \(indexStorePath) but build server specified an \
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -353,33 +443,100 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+      for (fileInfo, outputPath) in filesAndOutputPaths {
+        guard let filePath = try? fileInfo.mainFile.fileURL?.filePath else {
+          logger.error("Failed to determine file path of file to index \(fileInfo.mainFile.forLogging)")
+          continue
+        }
+        outputFileMap[filePath] = .init(indexUnitOutputPath: outputPath)
+      }
+      let tempFileUri = FileManager.default.temporaryDirectory
+        .appending(component: "sourcekit-lsp-output-file-map-\(UUID().uuidString).json")
+      try JSONEncoder().encode(outputFileMap).write(to: tempFileUri)
```
bnbarham, line 500: Do we ever delete this?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -475,29 +634,89 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
-        Updating index store for \(indexFile.forLogging) terminated with non-zero exit code \(code)
+        Updating index store for terminated with non-zero exit code \(code) for \(indexFiles)
         Stderr:
         \(stderr)
         Stdout:
         \(stdout)
         """
       )
-      BuildSettingsLogger.log(level: .debug, settings: buildSettings, for: indexFile)
+    // BuildSettingsLogger.log(level: .debug, settings: buildSettings, for: indexFile)
```
bnbarham: `.first` like above (in which case, maybe add another overloads that takes a list and does that and then use https://github.com/swiftlang/sourcekit-lsp/pull/2293/files#diff-ea45e11d345f3369fa911535dbd114c644d1757f43bbc4e726f644757da2bd83R388 as well)

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -475,29 +634,89 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+        // languages.
+        partitions.append((fileIndexInfo.target, fileIndexInfo.language, [fileIndexInfo]))
+        continue
+      }
+      // Even for Swift files, we can only index files in a single compiler invocation if they have the same build
+      // settings (modulo some normalization in `updateIndexStore(forFiles:)`). We can't know that all the files do
+      // indeed have the same compiler arguments but loading build settings during scheduling from the build server
+      // is not desirable because it slows down a bottleneck.
+      // Since Swift files within the same target should build a single module, it is reasonable to assume that they all
+      // share the same build settings.
```
bnbarham, line 700: I'm not sure I'd even call this an assumption TBH - if it wasn't the case, things would be very broken.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -475,29 +634,89 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+    }
+    let batchedPartitions =
+      fileIndexInfosToBatch
+      .sorted { $0.key < $1.key }  // Ensure we get a deterministic partition order
+      .flatMap { targetAndLanguage, files in
+        // The batch size of 5 was chosen without too many significant performance measurements because most projects
+        // currently indexed by SourceKit-LSP are limited by preparation time instead of indexing time and it's thus
+        // hard to quanify the performance characteristics of different batch sizes. 5 seems like a good trade-off to
+        // share work between files within the same target without overloading a single job with too many files and
+        // thus loosing parallelism.
```
bnbarham:
```suggestion
        // thus losing parallelism.
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -475,29 +634,89 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
+        files.partition(intoBatchesOfSize: 5).map { (targetAndLanguage.target, targetAndLanguage.language, $0) }
```
bnbarham, line 713: Could make configurable? What's the driver choose for its batch sizes?

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -144,6 +145,56 @@ public final class Toolchain: Sendable {
...
+  package var canIndexMultipleSwiftFilesInSingleInvocation: Bool {
+    get async {
+      let task = canIndexMultipleSwiftFilesInSingleInvocationTask.withLock { task in
+        if let task {
+          return task
+        }
+        let newTask = Task<Bool, Never> { () -> Bool in
+          #if compiler(>=6.4)
+          #warning(
+            "Once we no longer Swift 6.2 toolchains, we can assume that the compiler has https://github.com/swiftlang/swift-driver/pull/1979"
```
bnbarham, line 163:
```suggestion
            "Once we no longer support Swift 6.3 toolchains, we can assume that the compiler has https://github.com/swiftlang/swift-driver/pull/1979"
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: We don’t control what the BSP server gives us for the build settings, so a BSP server may return different build settings for different Swift files within the same target. It wouldn’t make much sense from a build server’s perspective in practice but I’ve learned that it’s better to design SourceKit-LSP in a way that makes as few implicit assumptions about the BSP server responses as possible and assuming that all Swift files within the same target have the same build settings would be one of them.

> It's super unfortunate we have to look through the arguments to find -index-unit-output-path for every settings per file, even though they should all be the same for every batch (except for the output path).

Do you think from a performance perspective. If we launch a compiler to index the file, I doubt that scanning for a command line argument makes much of a difference. And really, we should be getting the output path from the `buildTarget/sources` response (`fileInfo.outputPath`), in which case the extraction of `-index-unit-output-path` is just a validity check to ensure the two match.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Good catch. Was an artifact from a previous iteration.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Ooops, must have lost the `defer` block to delete this at some point.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Good catch. Added an overload that takes a list

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Same reply as https://github.com/swiftlang/sourcekit-lsp/pull/2293#discussion_r2365638192. I guess it’s just nice to see that SourceKit-LSP continues to function even if the build server gives us unreasonable responses.

Sources/ToolchainRegistry/Toolchain.swift
ahoppen: By the time that we stop supporting 6.2 we don’t need to worry about the early Swift 6.3 toolchain snapshots that didn’t have multi-file indexing support anymore.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
ahoppen: Good idea to check with the driver. It schedules as many concurrent batches as the machine has CPU cores but limits the batch size to 25 (https://github.com/swiftlang/swift-driver/blob/main/Sources/SwiftDriver/Jobs/Planning.swift#L917-L1022). I think copying that behavior is as reasonable an ideas as I have. But I’d prefer to do that in a follow-up PR to keep this one focused on the actual implementation of batch indexing instead of introducing complications due to configuration options.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/ToolchainRegistry/Toolchain.swift
bnbarham: I was more on auto-pilot and just comparing versions in the #if. But fair enough 👍

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
bnbarham: I've seen extremely large compiler arguments, it all adds up 🤷 But I'm fine leaving in handling different arguments, can always re-visit if we notice it being a real issue.

bnbarham (approved):

bnbarham:
Thanks! Exciting to see this in :)

Sources/BuildServerIntegration/BuildSettingsLogger.swift
```diff
@@ -27,11 +28,31 @@ package actor BuildSettingsLogger {
...
+  /// Log the given build settings for a list of source files that all share the same build settings.
+  ///
+  /// In contrast to the instance method `log`, this will always log the build settings. The instance method only logs
+  /// the build settings if they have changed.
+  package static func log(level: LogLevel = .default, settings: FileBuildSettings, for uris: [DocumentURI]) {
+    let header: String
+    if let uri = uris.only {
+      header = "Build settings for \(uri.forLogging)"
+    } else if let firstUri = uris.first {
+      header = "Build settings for \(firstUri.forLogging) and \(firstUri) and \(uris.count - 1) others"
```
bnbarham, line 48:
```suggestion
      header = "Build settings for \(firstUri.forLogging) and \(uris.count - 1) others"
```

bnbarham: (tests have already passed, another PR is fine)

PR #2294, opened 2025-09-14
Support custom preparationBatchSize defined via SourceKit's options

josh-arnold-1:


ahoppen:
Thanks for picking this up @josh-arnold-1 🙏. One high-level comment: I would really like us to design the configuration option that allows us to customize the batching strategy as I described in https://github.com/swiftlang/sourcekit-lsp/pull/2238#issuecomment-3207531527. do you think you could look into that?

josh-arnold-1:
> Thanks for picking this up @josh-arnold-1 🙏. One high-level comment: I would really like us to design the configuration option that allows us to customize the batching strategy as I described in [#2238 (comment)](https://github.com/swiftlang/sourcekit-lsp/pull/2238#issuecomment-3207531527). do you think you could look into that?

Thanks for the review! 

What if we update the schema to be an enum of configuration options like you specified, but we just supply a single option for now, which we default to a target size of 1 to maintain SourceKit-LSPs default behavior?

That way, we can easily extend additional strategies in future PRs, whilst maintaining the current configuration API?

What are your thoughts? Thanks!

```
      {
        "type": "object",
        "description": "Prepare a fixed number of targets in a single batch",
        "properties": {
          "strategy": {
            "const": "target"
          },
          "batchSize": {
            "type": "integer",
            "description": "Defines how many targets should be prepared in a single batch"
          }
        },
        "required": [
          "strategy",
          "batchSize"
        ]
      },
```

ahoppen:
Your proposal for the JSON schema sounds great to me!

brentleyjones:
Is there a way to set `batchSize` to `inf`/"batch everything at once"?

bnbarham:
> Is there a way to set batchSize to inf/"batch everything at once"?

Setting to a high value seems reasonable to me for this rather than handling it specifically

josh-arnold-1:
Sorry for the delayed response — I was OOO recently. I’m looking into this now and realized that config.schema.json generation might not support enums with associated values (unless I’m missing something?).

If that’s the case, what would be the best way to represent the different batching strategies?

@ahoppen, any guidance here would be super helpful. Thanks!

ahoppen:
Yeah, the JSON schema generation would need to be expanded to support this. That’s what I meant in https://github.com/swiftlang/sourcekit-lsp/pull/2238#issuecomment-3207531527.

> As a side note, allowing us to generate the oneOf in the schema above like this will likely need quite a bit new functionality in [ConfigSchemaGen](https://github.com/swiftlang/sourcekit-lsp/tree/main/SourceKitLSPDevUtils/Sources/ConfigSchemaGen). If we only stick to the target-based strategy, we should only need support for the const key in the JSON schema, which should be a lot easier to accomplish.

josh-arnold-1:
Thanks @ahoppen, I just updated the code with what we discussed!

SourceKitLSPDevUtils/Sources/ConfigSchemaGen/JSONSchema.swift
```diff
@@ -82,6 +87,10 @@ struct JSONSchema: Encodable {
     if let markdownEnumDescriptions {
       try container.encode(markdownEnumDescriptions, forKey: .markdownEnumDescriptions)
     }
+    if let oneOf = oneOf, !oneOf.isEmpty {
```
ahoppen:
```suggestion
    if let oneOf, !oneOf.isEmpty {
```

SourceKitLSPDevUtils/Sources/ConfigSchemaGen/JSONSchema.swift
```diff
@@ -126,13 +135,53 @@ struct JSONSchemaBuilder {
...
       schema.required = required
     case .enum(let enumInfo):
-      schema.type = "string"
-      schema.enum = enumInfo.cases.map(\.name)
-      // Set `markdownEnumDescriptions` for better rendering in VSCode rich hover
-      // Unlike `description`, `enumDescriptions` field is not a part of JSON Schema spec,
-      // so we only set `markdownEnumDescriptions` here.
-      if enumInfo.cases.contains(where: { $0.description != nil }) {
-        schema.markdownEnumDescriptions = enumInfo.cases.map { $0.description ?? "" }
+      let hasAssociatedTypes = enumInfo.cases.contains { $0.associatedProperties != nil && !$0.associatedProperties!.isEmpty }
```
ahoppen: I personally think it reads nicer like this because you can just infer that `nil` is considered empty. But I know that some people disagree.
 
```suggestion
      let hasAssociatedTypes = enumInfo.cases.contains { !($0.associatedProperties?.isEmpty ?? true) }
```

SourceKitLSPDevUtils/Sources/ConfigSchemaGen/OptionDocument.swift
```diff
@@ -66,14 +66,34 @@ struct OptionDocumentBuilder {
...
-          guard let description = caseInfo.description else {
-            continue
+        let hasAssociatedTypes = schema.cases.contains { $0.associatedProperties != nil && !$0.associatedProperties!.isEmpty }
+        
+        if hasAssociatedTypes {
+          let discriminatorFieldName = schema.discriminatorFieldName ?? "type"
+          doc += "\(indent)  - This is a tagged union discriminated by the `\(discriminatorFieldName)` field. Each case has the following structure:\n"
+          
+          for caseInfo in schema.cases {
+            doc += "\(indent)  - `\(discriminatorFieldName): \"\(caseInfo.name)\"`"
```
ahoppen: Very nitpicky: I generally find that string literals that contain quotes read nicer when used in multi-line string literals because then you don’t need to escape the quotes.

```suggestion
            doc += """
              \(indent)  - `\(discriminatorFieldName): "\(caseInfo.name)"`
              """
```

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -441,14 +441,13 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     if let buildServerWorkspaceRequestsTimeout {
       return .seconds(buildServerWorkspaceRequestsTimeout)
     }
-    // The default value needs to strike a balance: If the build server is slow to respond, we don't want to constantly
-    // run into this timeout, which causes somewhat expensive computations because we trigger the `buildTargetsChanged`
-    // chain.
-    // At the same time, we do want to provide functionality based on fallback settings after some time.
-    // 15s seems like it should strike a balance here but there is no data backing this value up.
```
ahoppen, line 448: Accidentally removed this comment?

Sources/SKOptions/PreparationBatchingStrategy.swift
```diff
@@ -0,0 +1,54 @@
...
+  
+  public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    switch self {
+    case .target(let batchSize):
+      try container.encode(StrategyValue.target, forKey: .strategy)
+      try container.encode(batchSize, forKey: .batchSize)
+    }
+  }
+}
```
ahoppen: Could you [run swift-format](https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting) to format the source files, which should also add a trailing newline here.

Sources/SKOptions/PreparationBatchingStrategy.swift
```diff
@@ -0,0 +1,54 @@
...
+/// Defines the batch size for target preparation.
+///
+/// If nil, SourceKit-LSP will default to preparing 1 target at a time.
+///
+/// - discriminator: strategy
+public enum PreparationBatchingStrategy: Sendable, Equatable {
+  /// Prepare a fixed number of targets in a single batch.
+  ///
+  /// `batchSize`: The number of targets to prepare in each batch.
+  case target(batchSize: Int)
```
ahoppen: `target` is pretty ambiguous here, should we make this more specific and call it `fixedTargetBatchSize`. That should really clarify what this strategy does.

Sources/SKOptions/PreparationBatchingStrategy.swift
josh-arnold-1: I ran `swift format -ipr .`, thanks!

Sources/SKOptions/SourceKitLSPOptions.swift
josh-arnold-1: Oops, it should be resolved now. Thanks!

josh-arnold-1:
Sorry for the month-long hiatus, I was quite strapped for time. 

I tried to respond to all the comments, please let me know if I missed anything! Thanks @ahoppen for your time reviewing this 🙏

SourceKitLSPDevUtils/Sources/ConfigSchemaGen/OptionSchema.swift
```diff
@@ -234,4 +269,52 @@ struct OptionSchemaContext {
...
+      default:
+        return []
+      }
+    }
+
+    for line in docLines {
+      var trimmed = line
+      while trimmed.first?.isWhitespace == true {
+        trimmed = trimmed.dropFirst()
+      }
```
ahoppen: I think this could just be 

```suggestion
      var trimmed = line.drop(while: \.isWhitespace)
```

config.schema.json
```diff
@@ -186,6 +186,31 @@
       },
       "type" : "object"
     },
+    "preparationBatchingStrategy" : {
+      "description" : "Defines the batch size for target preparation. If nil, defaults to preparing 1 target at a time.",
+      "markdownDescription" : "Defines the batch size for target preparation. If nil, defaults to preparing 1 target at a time.",
+      "oneOf" : [
+        {
+          "description" : "Prepare a fixed number of targets in a single batch. `batchSize`: The number of targets to prepare in each batch.",
```
ahoppen, line 194: Shouldn’t we get `markdownEnumDescriptions` in here as well? I suspect that it’s just something stupid that’s stopping it from appearing.

ahoppen (commented):
No worries with the delay and thanks again for your work on this. I have two nitpicky comments, but I’d be happy to merge this as-is as well and address those in follow-up PRs.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

SourceKitLSPDevUtils/Sources/ConfigSchemaGen/OptionSchema.swift
josh-arnold-1: Thanks! Updated!

config.schema.json
josh-arnold-1: Ah, I believe the current implementation uses `markdownEnumDescriptions` for simple enum's without associated values, whilst it nests the "markdownDescription"  inside of the "oneOf" when the enum has associated values. 

I think if I'm not mistaken, using `markdownEnumDescriptions` could allow for better VSCode support (e.g intellisense over hover), and is easy to parse for simple cases.

We could always default to "oneOf" approach instead of using `markdownEnumDescriptions`, but I think we might lose out on the better integration with vscode's `markdownEnumDescriptions`. 

Please correct me if I'm making any wrong assumptions here! Thanks!

josh-arnold-1:
Thank you for the review! I responded to the comments. Happy thanksgiving 🦃

config.schema.json
ahoppen: OK, I just had a brain fart. This no longer is an enum in the JSON spec sense but a `oneOf`, so `markdownEnumDescriptions` doesn’t make sense. Must have missed the distinction when reviewing.

ahoppen (approved):
Thanks. LGTM. Very excited to see this get in 🤩

ahoppen:
@swift-ci Please test

Sources/SKOptions/PreparationBatchingStrategy.swift
ahoppen, line 1: Could you add this file to https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SKOptions/CMakeLists.txt#L6 so it gets picked up by the Windows build, which uses CMake instead of SwiftPM.

ahoppen (approved):

Sources/SKOptions/PreparationBatchingStrategy.swift
josh-arnold-1: Just updated! Thanks!

josh-arnold-1:
@ahoppen just updated! Thank you!

ahoppen:
@swift-ci Please test

josh-arnold-1:
I see the testing failed — where is the right place to interpret the test failure logs? I can look into this and try to fix.  Thanks!

ahoppen:
Searching for the last occurrences of `error:` yields the following. I think you need to rebase your PR on top of main to pick up https://github.com/swiftlang/sourcekit-lsp/pull/2366 and then resolve the build failure.

```
[2025-12-03T03:55:36.906Z] /Users/ec2-user/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/sourcekit-lsp/Sources/SKOptions/PreparationBatchingStrategy.swift:46:34: error: use of protocol 'Encoder' as a type must be written 'any Encoder'; this will be an error in a future Swift language mode [#ExistentialAny]
[2025-12-03T03:55:36.906Z] 44 |   }
[2025-12-03T03:55:36.906Z] 45 | 
[2025-12-03T03:55:36.906Z] 46 |   public func encode(to encoder: Encoder) throws {
[2025-12-03T03:55:36.906Z]    |                                  `- error: use of protocol 'Encoder' as a type must be written 'any Encoder'; this will be an error in a future Swift language mode [#ExistentialAny]
[2025-12-03T03:55:36.906Z] 47 |     var container = encoder.container(keyedBy: CodingKeys.self)
[2025-12-03T03:55:36.906Z] 48 |     switch self {

```

josh-arnold-1:
Updated! Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2297, opened 2025-09-19
Fix `sourcekitdSupportsFullDocumentationInCompletion` to not rely on `ideApi`

hamishknight:
`ideApi` is only available within the plugin itself, it's not available to sourcekitd. Switch to sending a completion request and checking that we get full documentation back.


hamishknight:
@swift-ci please test

bnbarham (approved):

Issue #2122, opened 2025-04-25
Code completion does not show cross-file new identifiers (unless more characters are typed)


DimaKozhinov:
### Swift version

Swift version 6.1 (swift-6.1-RELEASE)

### Platform

Windows 11

### Editor

Visual Studio Code 1.99.3 with Swift plugin 2.2.0, other editors

### Description

When a new identifier is added (a let constant, a func, etc.) in a different source file within a module, it is not listed in results of `textDocument/completion` request, unless a significant portion of the new identifier is typed by user.

### Steps to Reproduce

1. Create a Swift package:
```
mkdir test
cd test
swift package init --type executable
```
2. Add a new Swift file in the Sources folder of the package, e.g. `helpers.swift` in addition to the default `main.swift` .
3. Open the `test` package folder in Visual Studio Code.
4. Edit the `helpers.swift` in VS Code: add for example `let zzzConst = 42` .
5. Switch to `main.swift` file in VS Code and type something that would use the new constant from `helpers.swift`, e.g. "let aaa = z"
6. Code completion does not show `zzzConst` (shows only two suggestions starting with "zip")
7. Add more text, now "let aaa = zz"
8. Code completion does not show any suggestions at all.
9. Add more text, now "let aaa = zzz"
10. At last `zzzConst` appears in code suggestions.

DimaKozhinov:
After saving `helpers.swift` the `zzzConst` does appear in code completion.

ahoppen:
Thanks for clarifying.

DimaKozhinov:
Update for Swift 6.2:
With  Swift 6.2 the bug is not reproduced anymore in VS Code.

It is partially reproduced in my custom editor though:
 - When I switch from `main. swift` and modify `helpers.swift` _for the first time_ (add a `let` constant or rename it), when I return to `main.swift` I see the correct code suggestion for the new/renamed identifier;
 - But when I repeat this (e.g. switch to `helpers.swift` and rename or delete the constant), in the `main.swift` I see previous version in code suggestion (old name of deleted or renamed identifier).
 - Saving the `helpers.swift` does not help (helped in Swift 6.1).

I suspect that integration of sourcekit-lsp in my custom editor lacks something (I use minimum required parameters in every request). Of course I would like to improve it, but I don't know how.

As the bug is not reproduced anymore in VS Code, I think we can close this issue @ahoppen ?

ahoppen:
[...]

PR #2299, opened 2025-09-20
Add test case checking that we show completions from in-memory contents of files within same module

ahoppen:
This functionality is implemented in sourcekitd, we just weren’t testing it anywhere. I wasn’t sure that it actually worked in #2122, so let’s document this behavior in a test case.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2302, opened 2025-09-22
Match the batch size for multi-file indexing to the driver's batch size

ahoppen:
Until we have better measurements that would motivate a different batching strategy, copying the driver’s batch size seems like the most reasonable thing to do.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2303, opened 2025-09-22
Use `activeProcessorCount` instead of `processorCount` in short-lived use-cases

ahoppen:
According to https://developer.apple.com/documentation/foundation/processinfo/activeprocessorcount

> Whereas the processorCount property reports the number of advertised processing cores, the activeProcessorCount property reflects the actual number of active processing cores on the system. There are a number of different factors that may cause a core to not be active, including boot arguments, thermal throttling, or a manufacturing defect.

For short-lived workloads like `concurrentMap` we want to parallelize across the number of cores that are currently active, so use `activeProcessorCount` instead. The only case where we want to continue using `processorCount` is the computation of concurrent tasks for `TaskScheduler` because the value is stored for the lifetime of the SourceKit-LSP process and we don’t want to limit parallelism if SourceKit-LSP was launched during a time of thermal throttling.

I stumbled across this while working on #2302

### DIFF
--- a/Sources/CompletionScoring/Text/CandidateBatch.swift
+++ b/Sources/CompletionScoring/Text/CandidateBatch.swift
@@ -440,7 +440,7 @@ extension Pattern {
     compactScratchArea(capacity: Self.totalCandidates(batches: batches)) { matchesScratchArea in
       let scoringWorkloads = ScoringWorkload.workloads(
         for: batches,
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2305, opened 2025-09-22
Add advice that the `workspace/buildTargets` etc. requests should return ASAP and not wait for expensive computations

ahoppen:
This has come up in https://github.com/swiftlang/sourcekit-lsp/issues/2252#issuecomment-3242099324 and likely https://github.com/swiftlang/sourcekit-lsp/issues/2304. Let’s make it general advice.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2308, opened 2025-09-26
Avoid crashing when loading client plugins from the same path

bnbarham:
When `DYLD_(FRAMEWORK|LIBRARY)_PATH` is set, `dlopen` will first check if the basename of the provided path is within any of its search paths. Thus it's possible that only a single library is loaded for each toolchain, rather than a separate like we expect. The paths should be equal in this case, since the client plugin is loaded based on the path of `sourcekitd.framework` (and we should only have one for the same reason). Allow this case and just avoid re-initializing.


bnbarham:
@swift-ci please test

rintaro (approved):

bnbarham:
@swift-ci please test

bnbarham:

ahoppen (approved):

PR #2309, opened 2025-09-26
Log when absolute compilation DB search paths are skipped

bnbarham:
Absolute search paths were being ignored without logging, which makes it somewhat difficult to diagnose. Log when they're skipped.

Also remove a duplicate options merging block - both `createWorkspaceWithInferredBuildServer` and `findImplicitWorkspace` (the only callers of `createWorkspace`) already merge in the workspace options.


bnbarham:
@swift-ci please test

PR #2310, opened 2025-09-29
Flush logs at the end of test execution

ahoppen:
Otherwise it sometimes happens that shutdown and other messages are logged during the next test’s execution, which makes it harder to diagnose issues.


ahoppen:
@swift-ci Please test

PR #2311, opened 2025-09-29
Do not block SourceKit-LSP functionality when a build server takes long to initialize

### LINKED ISSUE #2276: Support copied file mappings in all requests
https://github.com/swiftlang/sourcekit-lsp/pull/2271 added the ability to map copied header files back to their original for jump-to-definition. We will need to do the same mapping for all other requests that return source files or locations. Depending on how many these are, this can be done in multiple PRs.

### LINKED ISSUE #2304: BSP build/taskStart notifications seems to not be processed until sometime after `buildTargets/` returns
### Swift version

6.1.2, Xcode 16.4

### Platform

macOS Sequoia

### Editor

_No response_

### Description

I noticed this behavior and couldn't figure out why it was happening, so thought about posting it here in case it raises some eyebrows. Essentially, any `build/taskStart` (and related) notifications sent by a BSP will not be shown by the IDE until sometime after the `buildTargets` request returns, which can take a while.

Looking at [BuildServerMessageDependencyTracker.swift](https://github.com/swiftlang/sourcekit-lsp/tree/main/Sources/BuildServerIntegration/BuildServerMessageDependencyTracker.swift) and the logic in AsyncQueue.swift indicates that this notification should be able to run in parallel with everything else, but something in the initialization flow is somehow blocking it. Once `buildTargets` returns, I can see that all such notifications that were queued suddenly get processed at once and any new future tasks show instantly as one would expect.

I thought it might have something to do with the calls to `buildServerAdapterAfterInitialized`, but since this is Swift Concurrency it should work fine in theory. Any ideas? The BSP is in this case is fully asyc, so I can see on my side that the notifications are indeed being dispatched when I intende them to.

### Steps to Reproduce

_No response_

### Logging

_No response_

ahoppen:
We previously waited for the initialization response from the build server during the creation of a `Workspace` so that we could create a `SemanticIndexManager` with the index store path etc. that was returned by the `build/initialize` response. This caused all functionality (including syntactic) of SourceKit-LSP to be blocked until the build server was initialized.

Change the computation of the `SemanticIndexManager` and related types to happen in the background so that we can provide functionality that doesn’t rely on the build server immediately.

Fixes #2304

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -343,7 +343,26 @@ package actor BuildServerManager: QueueBasedMessageHandler {
   }
 
   /// Provider of file to main file mappings.
-  private var mainFilesProvider: MainFilesProvider?
+  ///
+  /// Force-unwrapped optional because initializing it requires access to `self`.
+  private var mainFilesProvider: Task<MainFilesProvider?, Never>! {
+    didSet {
+      // Must only be set once
+      precondition(oldValue == nil)
+      precondition(mainFilesProvider != nil)
+    }
+  }
+
+  package func mainFilesProvider<T: MainFilesProvider>(as: T.Type) async -> T? {
+    guard let mainFilesProvider = mainFilesProvider else {
+      return nil
+    }
+    guard let index = await mainFilesProvider.value as? T else {
+      logger.fault("Expected the main files provider of the build server manager to be an `\(T.self)`")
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #2315, opened 2025-10-01
Prototype pipe handling without dispatch IO

roman-bcny:


Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -58,7 +58,7 @@ public final class JSONRPCConnection: Connection {
   /// The queue on which we send data.
   private let sendQueue: DispatchQueue = DispatchQueue(label: "jsonrpc-send-queue", qos: .userInitiated)
 
-  private let receiveIO: DispatchIO
+  private let inFD: FileHandle
   private let sendIO: DispatchIO
```
compnerd: I think that we should replace this to be symmetric with the read side.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -293,27 +277,17 @@ public final class JSONRPCConnection: Connection {
...
-        }
-
-        if done {
-          self.closeAssumingOnQueue()
-          return
-        }
-
-        guard let data = data, !data.isEmpty else {
+    self.inFD.readabilityHandler = { fileHandle in
+      let data = (try? fileHandle.read(upToCount: Int.max)) ?? Data()
```
ahoppen: Instead of `try?`, I would prefer to use `orLog` so that we log the error thrown by `read(upToCount:)`.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -293,27 +277,17 @@ public final class JSONRPCConnection: Connection {
...
+      if data.isEmpty {
+          fileHandle.readabilityHandler = nil
+          self.close()
           return
-        }
+      }
 
+      self.queue.sync {
```
ahoppen, line 268: `self.queue.async` should be sufficient here, it still guarantees that all incoming data is handled in-order and frees up the thread that the readability handler is running on. Or do you think I’m missing something?

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -86,7 +86,6 @@ public final class JSONRPCConnection: Connection {
   /// Buffer of received bytes that haven't been parsed.
   ///
   /// Access to this must be be guaranteed to be sequential to avoid data races. Currently, all access are
-  ///  - The `receiveIO` handler: This is synchronized on `queue`.
```
ahoppen, line 89: We now have a corresponding use in `self.inFD.readabilityHandler`, so we should include that here:

```swift
  /// - The `inFD.readabilityHandler`: This is synchronized on `queue`.
```

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -660,7 +634,6 @@ public final class JSONRPCConnection: Connection {
 
       logger.log("Closing JSONRPCConnection...")
       // Attempt to close the reader immediately; we do not need to accept remaining inputs.
-      receiveIO.close(flags: .stop)
```
ahoppen, line 663: Should we set the `readabilityHandler` of `inFD` to `nil` here to ensure that we don’t receive any more messages? If we do so, I think we also no longer need to set `fileHandle.readabilityHandler = nil` if `data.isEmpty` in the `readabilityHandler`.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -58,7 +58,7 @@ public final class JSONRPCConnection: Connection {
   /// The queue on which we send data.
   private let sendQueue: DispatchQueue = DispatchQueue(label: "jsonrpc-send-queue", qos: .userInitiated)
 
-  private let receiveIO: DispatchIO
+  private let inFD: FileHandle
```
ahoppen, line 61: Should we match the previous naming and call this `receiveFD`? `inFD` can get confusing if you eg. set up a `JSONRPCConnection` from SourceKit-LSP to a BSP server, in which case the `receiveFD` will be stdout of the BSP server process and thus `receiveFD` wouldn’t be stdin.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -140,23 +139,12 @@ public final class JSONRPCConnection: Connection {
 
     #if os(Windows)
     let rawInFD = dispatch_fd_t(bitPattern: inFD._handle)
```
ahoppen: Do we still need `rawInFD`?

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
roman-bcny: yup I'll need to update the docs once this settles

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
roman-bcny: I think I'm doing this now but might be doing it wrong :)

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
roman-bcny: I'm now using async in the last iteration, but I'm not sure I'm doing it right in each particular case

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
roman-bcny: Naming follows `JSONRPCConnection.init()` args names, are you suggesting we change these as well?

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -293,27 +253,19 @@ public final class JSONRPCConnection: Connection {
...
       self.closeHandler = closeHandler
+    }
 
-      receiveIO.read(offset: 0, length: Int.max, queue: queue) { done, data, errorCode in
-        guard errorCode == 0 else {
-          #if !os(Windows)
-          if errorCode != POSIXError.ECANCELED.rawValue {
-            logger.fault("IO error reading \(errorCode)")
+    self.inFD.readabilityHandler = { fileHandle in
+      let data = fileHandle.availableData
```
ahoppen, line 259: I unfortunately can’t remember the example when `fileHandle.availableData` has issues but I’ve been bitten by it more than once. I just remember that it’s a very delicate scenario where it misbehaves and it’s a pain to debug when it happens. So, I really want to use `read(upToCount: Int.max))` here.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -293,27 +253,19 @@ public final class JSONRPCConnection: Connection {
...
+      if data.isEmpty {
+          fileHandle.readabilityHandler = nil
+          self.queue.async {
+            self.closeAssumingOnQueue()
           }
```
ahoppen, line 264: I think we need to `return` here. If we have set the `readabilityHandler` to `nil` and scheduled queue closing, we don’t have any further processing to do, do we?

Also, `if data.isEmpty` could then become a `guard` statement.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
ahoppen: Why do you think we shouldn’t be using `async` here? My thought is that with `sync`, we are unnecessarily blocking the thread that handles the `readabilityHandler` while we are performing work on `queue`.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -554,16 +506,16 @@ public final class JSONRPCConnection: Connection {
...
-    sendIO.write(offset: 0, data: dispatchData, queue: sendQueue) { [weak self] done, _, errorCode in
-      if errorCode != 0 {
-        logger.fault("IO error sending message \(errorCode)")
-        if done, let self {
-          // An unrecoverable error occurs on the channel’s file descriptor.
-          // Close the connection.
-          self.queue.async {
-            self.closeAssumingOnQueue()
-          }
+    sendQueue.sync {
```
ahoppen, line 509: Similar to my other `sync` vs `async` comment, I think we can make this `async` as well. The transmission of the data via stdout is asynchronous anyway, so as long as data is kept in-order (which it is using the queue), we should be able to handle sending in the background.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -660,9 +615,12 @@ public final class JSONRPCConnection: Connection {
...
       logger.log("Closing JSONRPCConnection...")
       // Attempt to close the reader immediately; we do not need to accept remaining inputs.
-      receiveIO.close(flags: .stop)
       // Close the writer after it finishes outstanding work.
-      sendIO.close()
+      do {
+        try outFD.close()
+      } catch {
+        logger.error("Failed to close outFD: \(error.forLogging)")
+      }
```
ahoppen, line 623: Can be simplified slightly.

```suggestion
      orLog("Closing outFD") {
        try outFD.close()
      }
```

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
ahoppen: Yeah, I think changing the argument names `JSONRPCConnection.init()` would be my preference. Happy to change the argument names in a follow-up PR though, this PR is delicate enough that we don’t need to litter it with other renames. But I would prefer to name the members in here `receiveFD` and `sendFD` now already.

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -58,8 +58,9 @@ public final class JSONRPCConnection: Connection {
   /// The queue on which we send data.
   private let sendQueue: DispatchQueue = DispatchQueue(label: "jsonrpc-send-queue", qos: .userInitiated)
 
-  private let receiveIO: DispatchIO
-  private let sendIO: DispatchIO
+  private let inFD: FileHandle
+  private let outFD: FileHandle
+  let ioGroup: DispatchGroup
```
ahoppen, line 63: Should this be `private`?

Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
```diff
@@ -136,45 +136,12 @@ public final class JSONRPCConnection: Connection {
...
-        }
-        ioGroup.leave()
-      }
-    )
+    self.inFD = inFD
+    self.outFD = outFD
+
+    self.ioGroup.enter()
 
     ioGroup.notify(queue: queue) { [weak self] in
```
ahoppen, line 146: I think the entire `ioGroup` is now just a convoluted way of saying: This should be executed when the connection is closed: This will get executed when `ioGroup.leave()` is called, which happens after calling `closeAssumingOnQueue` in `JSONRPCConnection.close()` and the error handler of `JSONRPCConnectin.send`.

So, if we just executed this code inside `closeAssumingOnQueue`, I think we could get rid of the entire `ioGroup` logic and greatly simplify this.

Incidentally, now that I look at this, we might have been missing a few `ioGroup.leave()` before as well.

ahoppen (commented):
Reviewed this half way through again and then decided that it would be more efficient to modify this PR locally, attaching the diff. The key changes apart from my PR comments are:
- Ensuring we set `readabilityHandler` to `nil` whenever we close the connection
- Removing `sendQueue`. Now that we no longer have `DispatchIO` that’s backed by queues, I don’t think we need it anymore.

Now, the BIGGGGGG question is how to actually read data in `readabilityHandler`. We kind of have three options:
1. `fileHandle.availableData`: As noted in https://github.com/swiftlang/swift-package-manager/pull/8047, this has the potential of hanging when a significant amount of data is sent but also [various issues](https://github.com/swiftlang/swift-stress-tester/blob/3f5f86655e32c204ebeb38c0a3a8d1085957c859/SourceKitStressTester/Sources/Common/ProcessRunner.swift#L91) on other platforms (because swift-stress-tester doesn’t even run on Windows and I’m pretty sure this line was written without Windows in mind). 
2. `fileHandle.read(upToLength: Int.max)` only returns data when the end of the file handle is reached. While this is fine for use cases like the [ProcessRunner in swift-stress-tester](https://github.com/swiftlang/swift-stress-tester/blame/main/SourceKitStressTester/Sources/Common/ProcessRunner.swift#L91) that run short-lived processes, it is completely unusable for our case where we need to stream data
3. `fileHandle.read(upToLength: 1)`: Probably quite inefficient to read one byte at a time, especially since we try and will try and parse the header on every byte that’s read. If we want to consider this, we should do some performance measurements, eg. by recording an editor session in an input mirror file, replaying it and seeing how much time is spent parsing.

What do I think about these options?
- (2) does not work for our use case so we can rule that out. 
- (1) would be the best solution but I am very hesitant to switch to this unless we at least identified and fixed the Windows issue, which we have the best grasp on. @bnbarham I suppose you don’t remember what kind of issues you were facing 5 years ago when deciding against `availableData` in `ProcessRunner.swift`?
- (3) seems like the only option that’s left to us, though I definitely don’t like it.

Overall, this analysis gives me pretty bad vibes and it feels like we are quite likely to run into new problems on all platforms and it’s really hard to test these edge cases. At the moment we at least have a situation that we understand, does not hang and generally works without the quirk of spinning one thread on Windows. Maybe it would be worth waiting for [swift-subprocess](https://github.com/swiftlang/swift-subprocess) to be available in the toolchain and then directly switch to it (I don’t have any information about when this might be)? I’d like to hear more thoughts on this as well.

<details><summary>My changes</summary>
<p>

```diff
diff --git a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
index c2c1c696..8fb2e491 100644
--- a/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
+++ b/Sources/LanguageServerProtocolJSONRPC/JSONRPCConnection.swift
@@ -45,21 +45,17 @@ public final class JSONRPCConnection: Connection {
 
   /// The message handler that handles requests and notifications sent through this connection.
   ///
-  /// Access to this must be be guaranteed to be sequential to avoid data races. Currently, all access are
-  ///  - `init`: Reference to `JSONRPCConnection` trivially can't have escaped to other isolation domains yet.
-  ///  - `start`: Is required to be call in the same serial code region as the initializer, so
-  ///    `JSONRPCConnection` can't have escaped to other isolation domains yet.
-  ///  - `deinit`: Can also only trivially be called once.
+  /// - Important: Must only be accessed from `queue`, `init` and `deinit` to avoid data races.
   nonisolated(unsafe) private var receiveHandler: MessageHandler?
 
-  /// The queue on which we read the data
+  /// The queue on which we read and send the data
   private let queue: DispatchQueue = DispatchQueue(label: "jsonrpc-queue", qos: .userInitiated)
 
-  /// The queue on which we send data.
-  private let sendQueue: DispatchQueue = DispatchQueue(label: "jsonrpc-send-queue", qos: .userInitiated)
+  /// The file handle with which the other process sends data to us.
+  private let receiveFD: FileHandle
+  /// The file handle with which we can send data to the other process
+  private let sendFD: FileHandle
 
-  private let receiveIO: DispatchIO
-  private let sendIO: DispatchIO
   private let messageRegistry: MessageRegistry
 
   /// If non-nil, all input received by this `JSONRPCConnection` will be written to the file handle
@@ -86,7 +82,6 @@ public final class JSONRPCConnection: Connection {
   /// Buffer of received bytes that haven't been parsed.
   ///
   /// Access to this must be be guaranteed to be sequential to avoid data races. Currently, all access are
-  ///  - The `receiveIO` handler: This is synchronized on `queue`.
   ///  - `requestBufferIsEmpty`: Also synchronized on `queue`.
   private nonisolated(unsafe) var requestBuffer: [UInt8] = []
 
@@ -136,64 +131,8 @@ public final class JSONRPCConnection: Connection {
     state = .created
     self.messageRegistry = messageRegistry
 
-    let ioGroup = DispatchGroup()
-
-    #if os(Windows)
-    let rawInFD = dispatch_fd_t(bitPattern: inFD._handle)
-    #else
-    let rawInFD = inFD.fileDescriptor
-    #endif
-
-    ioGroup.enter()
-    receiveIO = DispatchIO(
-      type: .stream,
-      fileDescriptor: rawInFD,
-      queue: queue,
-      cleanupHandler: { (error: Int32) in
-        if error != 0 {
-          logger.fault("IO error \(error)")
-        }
-        ioGroup.leave()
-      }
-    )
-
-    #if os(Windows)
-    let rawOutFD = dispatch_fd_t(bitPattern: outFD._handle)
-    #else
-    let rawOutFD = outFD.fileDescriptor
-    #endif
-
-    ioGroup.enter()
-    sendIO = DispatchIO(
-      type: .stream,
-      fileDescriptor: rawOutFD,
-      queue: sendQueue,
-      cleanupHandler: { (error: Int32) in
-        if error != 0 {
-          logger.fault("IO error \(error)")
-        }
-        ioGroup.leave()
-      }
-    )
-
-    ioGroup.notify(queue: queue) { [weak self] in
-      guard let self else { return }
-      for outstandingRequest in self.outstandingRequests.values {
-        outstandingRequest.replyHandler(LSPResult.failure(ResponseError.internalError("JSON-RPC Connection closed")))
-      }
-      self.outstandingRequests = [:]
-      self.receiveHandler = nil  // break retain cycle
-      Task {
-        await self.closeHandler?()
-      }
-    }
-
-    // We cannot assume the client will send us bytes in packets of any particular size, so set the lower limit to 1.
-    receiveIO.setLimit(lowWater: 1)
-    receiveIO.setLimit(highWater: Int.max)
-
-    sendIO.setLimit(lowWater: 1)
-    sendIO.setLimit(highWater: Int.max)
+    self.receiveFD = inFD
+    self.sendFD = outFD
   }
 
   /// Creates and starts a `JSONRPCConnection` that connects to a subprocess launched with the specified arguments.
@@ -293,27 +232,20 @@ public final class JSONRPCConnection: Connection {
       state = .running
       self.receiveHandler = receiveHandler
       self.closeHandler = closeHandler
+    }
 
-      receiveIO.read(offset: 0, length: Int.max, queue: queue) { done, data, errorCode in
-        guard errorCode == 0 else {
-          #if !os(Windows)
-          if errorCode != POSIXError.ECANCELED.rawValue {
-            logger.fault("IO error reading \(errorCode)")
-          }
-          #endif
-          if done { self.closeAssumingOnQueue() }
-          return
-        }
-
-        if done {
-          self.closeAssumingOnQueue()
-          return
-        }
-
-        guard let data = data, !data.isEmpty else {
-          return
-        }
+    self.receiveFD.readabilityHandler = { fileHandle in
+      let data = orLog("Reading from \(self.name)") { try fileHandle.read(upToCount: 1) } ?? Data()
+      guard !data.isEmpty else {
+        // We have reached the end of `receiveFD`, close the connection. This will also set the `readabilityHandler` of
+        // `receiveFD` to `nil`, breaking the retain cycle.
+        self.close()
+        return
+      }
 
+      // We can run the actual data handling asynchronously in the background, all we need to ensure is that the data is
+      // handled in-order, which the queue does.
+      self.queue.async {
         orLog("Writing input mirror file") {
           try self.inputMirrorFile?.write(contentsOf: data)
         }
@@ -554,17 +486,17 @@ public final class JSONRPCConnection: Connection {
     orLog("Writing output mirror file") {
       try outputMirrorFile?.write(contentsOf: dispatchData)
     }
-    sendIO.write(offset: 0, data: dispatchData, queue: sendQueue) { [weak self] done, _, errorCode in
-      if errorCode != 0 {
-        logger.fault("IO error sending message \(errorCode)")
-        if done, let self {
-          // An unrecoverable error occurs on the channel’s file descriptor.
-          // Close the connection.
-          self.queue.async {
-            self.closeAssumingOnQueue()
-          }
-        }
+    do {
+      try sendFD.write(contentsOf: dispatchData)
+    } catch {
+      logger.fault("IO error sending message \(error.forLogging)")
+      // Match the pattern of `close()` but call `closeAssumingOnQueue` asynchronously to make sure that
+      // `closeAssumingOnQueue` is executed after all data from `receiveFD` has been read.
+      self.receiveFD.readabilityHandler = nil
+      self.queue.async {
+        self.closeAssumingOnQueue()
       }
+      return
     }
   }
 
@@ -584,7 +516,7 @@ public final class JSONRPCConnection: Connection {
   /// If an unrecoverable error occurred on the channel's file descriptor, the connection gets closed.
   ///
   /// - Important: Must be called on `queue`
-  func send(_ message: JSONRPCMessage) {
+  private func send(_ message: JSONRPCMessage) {
     dispatchPrecondition(condition: .onQueue(queue))
 
     let encoder = JSONEncoder()
@@ -646,7 +578,12 @@ public final class JSONRPCConnection: Connection {
   /// The user-provided close handler will be called *asynchronously* when all outstanding I/O
   /// operations have completed. No new I/O will be accepted after `close` returns.
   public func close() {
-    queue.sync { closeAssumingOnQueue() }
+    // Stop reading any more data from `receiveFD`. Scheduling `closeAssumingOnQueue` on `queue` after closing
+    // `receiveFD` ensures that we won't read any more data after `closeAssumingOnQueue`.
+    self.receiveFD.readabilityHandler = nil
+    queue.sync {
+      closeAssumingOnQueue()
+    }
   }
 
   /// Close the connection, assuming that the code is already executing on `queue`.
@@ -654,15 +591,25 @@ public final class JSONRPCConnection: Connection {
   /// - Important: Must be called on `queue`.
   private func closeAssumingOnQueue() {
     dispatchPrecondition(condition: .onQueue(queue))
-    sendQueue.sync {
-      guard state == .running else { return }
-      state = .closed
-
-      logger.log("Closing JSONRPCConnection...")
-      // Attempt to close the reader immediately; we do not need to accept remaining inputs.
-      receiveIO.close(flags: .stop)
-      // Close the writer after it finishes outstanding work.
-      sendIO.close()
+    // Stop receiving any more
+    guard state == .running else { return }
+    state = .closed
+
+    logger.log("Closing JSONRPCConnection...")
+    // Attempt to close the reader immediately; we do not need to accept remaining inputs.
+    // Close the writer after it finishes outstanding work.
+    do {
+      try sendFD.close()
+    } catch {
+      logger.error("Failed to close outFD: \(error.forLogging)")
+    }
+    for outstandingRequest in self.outstandingRequests.values {
+      outstandingRequest.replyHandler(LSPResult.failure(ResponseError.internalError("JSON-RPC Connection closed")))
+    }
+    self.outstandingRequests = [:]
+    self.receiveHandler = nil  // break retain cycle
+    Task {
+      await self.closeHandler?()
     }
   }
 
@@ -694,13 +641,13 @@ public final class JSONRPCConnection: Connection {
     id: RequestID,
     reply: @escaping @Sendable (LSPResult<Request.Response>) -> Void
   ) {
-    self.queue.sync {
-      guard readyToSend() else {
+    self.queue.async {
+      guard self.readyToSend() else {
         reply(.failure(.serverCancelled))
         return
       }
 
-      outstandingRequests[id] = OutstandingRequest(
+      self.outstandingRequests[id] = OutstandingRequest(
         responseType: Request.Response.self,
         replyHandler: { anyResult in
           let result = anyResult.map { $0 as! Request.Response }
@@ -732,7 +679,7 @@ public final class JSONRPCConnection: Connection {
         """
       )
 
-      send(.request(request, id: id))
+      self.send(.request(request, id: id))
       return
     }
   }

```

</p>
</details>

roman-bcny:
> Reviewed this half way through again and then decided that it would be more efficient to modify this PR locally, attaching the diff.

Thanks Alex, that's a very valid approach - this is still very much a prototype from someone who has little idea what they're doing in Swift!

> Now, the BIGGGGGG question is how to actually read data in `readabilityHandler`.

I think we should broaden this a bit to include options whether we should be using `readabilityHandler` at all. Note that my initial draft was just using an equivalent of thread-per-pipe, which may be a bit wasteful but as I already mentioned on the discussion thread this is what happens with DispatchIO on Windows (due to the monitor thread). There is the question whether this should be special-cased for Windows though - like maybe using blocking threads is fine on Windows but wouldn't be fine elsewhere, in which case we might want to choose a different approach to avoid having per-platform code (which may turn out to be difficult in the end anyways as there are clearly many complications)

> * (2) does not work for our use case so we can rule that out.

Yes, I believe this just doesn't work. I switched to it when it was suggested without realizing it would block until EOF, just to discover later on that everything would hang during testing, and it made sense after looking into that. I don't think this works at all.

> * (3) seems like the only option that’s left to us, though I definitely don’t like it.

Even a function call per each byte seems likely to run into performance issues, so I think it's fair to rule this out.

> * (1) would be the best solution but I am very hesitant to switch to this unless we at least identified and fixed the Windows issue, which we have the best grasp on. @bnbarham I suppose you don’t remember what kind of issues you were facing 5 years ago when deciding against `availableData` in `ProcessRunner.swift`?

If there is value in doing that, I'm happy to investigate issues caused by `availableData` on Windows if someone can provide pointers to a repro. If this is broader than just Windows I am not sure how to approach this though.

> Overall, this analysis gives me pretty bad vibes and it feels like we are quite likely to run into new problems on all platforms and it’s really hard to test these edge cases. At the moment we at least have a situation that we understand, does not hang and generally works without the quirk of spinning one thread on Windows. Maybe it would be worth waiting for [swift-subprocess](https://github.com/swiftlang/swift-subprocess) to be available in the toolchain and then directly switch to it (I don’t have any information about when this might be)? I’d like to hear more thoughts on this as well.

I think this is very fair as long as swift-subprocess doesn't use DispatchIO on Windows. I wasn't aware of this effort and it definitely looks like the better option if available. I suppose the best route might depend on how long it's going to take until this can be integrated - if this is still like a year away then maybe we could try to do something in the mean time - at least it could be platform-specific code for Windows to avoid creating hazards on other platforms.

bnbarham:
> @bnbarham I suppose you don’t remember what kind of issues you were facing 5 years ago when deciding against availableData in ProcessRunner.swift

I do not :(. I'm fairly sure it was just based on online discussions about availableData/readData

ahoppen:
> If there is value in doing that, I'm happy to investigate issues caused by `availableData` on Windows if someone can provide pointers to a repro. If this is broader than just Windows I am not sure how to approach this though.

I vaguely remember reproducing (or at least trying to) something like this or this issue (again, memory is very faulty) by having two processes (I implemented both in Swift): The parent process launches a child process, which listens to new data on stdin using the `readabilityHandler` + `availableData` pattern. The parent process then writes large amounts of data (think multiple kilobytes) to the pipe that the child process is listening to in single `write` call. If that doesn’t work, I would also try varying the size of the data written to the pipe to some interesting values like powers of 2 or (2^n + 1). Also I would try if this issue only occurs on the first package of data being sent over the pipe or maybe only on follow-up packages. 

I really regret now not filing an issue with some kind of reproducer when I opened https://github.com/swiftlang/swift-package-manager/pull/8047 😢 

If you find a way to reproduce and fix this issue, I would be willing to give `availableData` another shot because I can’t find any clues what the issues on other platforms are. And when we do find that `availableData` has issues like deadlocks on other platforms as well, we will need to either figure out how to fix those or, unfortunately, revet the change.

---

Two other approaches, I just thought of now:
1. In https://github.com/swiftlang/swift-package-manager/pull/9191, I noticed that they are using `read(fromFileDescriptor:maxLength:runningHandlerOn:handler:)` instead of `read(offset:length:queue:)`. I don’t have particularly high hopes that the former doesn’t have the hanging thread issue but it might be worth a try because it should be very low-effort.
2. If that doesn’t help, I would propose that we shouldn’t try and re-invent reading from pipes across platforms since (as we keep on learning), it’s very much non-trivial to get right. swift-subprocess seems to have an implementation that’s tailored for every platform in its [AsyncIO](https://github.com/swiftlang/swift-subprocess/tree/main/Sources/Subprocess/IO) implementation. All of it is internal right now but I think it’s worth a try copying those files into the SourceKit-LSP repo, adding a little bit of logic to allow reading from stdin instead of output pipes of subprocesses and seeing if that fixes all our issues. If it does, I think we could pitch a proposal to add a public  API to stream data from stdin, because this will be necessary to implement subprocesses and we can prove that it’s really nothing that anyone should try and implement themselves (again, as we keep on learning very painfully). If there are any minor issues left, we should report them to swift-subprocess, which should be interested in fixing them either way. For a transition period, I think it would be acceptable to vendor a copy of AsyncIO in SourceKit-LSP, if that’s the best way of fixing our issues.

(1) would be a very welcome and easy fix but (2) is really what I put my hopes on.

Would you be willing to give these ideas a shot? Also, just want to say again how much we appreciate your efforts, @roman-bcny  🙏🏽

bnbarham:
There's a third option here - switch to `read` (macOS/Linux) and `_read` on Windows. We'd have to read single bytes until we've read the message header, but could then have a single read for the content length. That's fairly simple - swift-syntax's `SwiftCompilerPluginMessageHandling` basically does that today (though that protocol is a little simpler, the length is always a single UInt64). So probably easier than (2), though making sure Subprocess ends up with what we'd need would also be useful.

ahoppen:
> There's a third option here - switch to `read` (macOS/Linux) and `_read` on Windows. We'd have to read single bytes until we've read the message header, but could then have a single read for the content length. That's fairly simple - swift-syntax's `SwiftCompilerPluginMessageHandling` basically does that today (though that protocol is a little simpler, the length is always a single UInt64). So probably easier than (2), though making sure Subprocess ends up with what we'd need would also be useful.

I think this is a great idea, better than what I suggested so far. A couple ideas to improve this even further:
- We don’t need to read one byte at a time in the header. We know the header starts with `Content-Length: `, so we can start by reading 16 bytes.
- To read how many bytes are in the message, we can read 4 bytes at a time until we hit a non-numeric character because  we know that the content length is followed by at least four bytes, namely the `\r\n\r\n` separating the content length header from the actual message. Though maybe this is an over-optimization, if reading one byte at a time here is easier, we can also do that.
- We don’t need to drop down to `read` at all, we could still use `readabilityHandler` and explicitly pass in the number of bytes we need in `fileHandle.read(upToLength:)`. This would fix the issues we have with `availableData`. Essentially this is a more optimized version of option (3) in https://github.com/swiftlang/sourcekit-lsp/pull/2315#pullrequestreview-3349115617.

bnbarham:
@roman-bcny do you mind if I take over this change? This has now moved to swift-tools-protocols and I've implemented the `fileHandle.read(upToLength:)` logic today (still need to test it against sourcekit-lsp though).

roman-bcny:
@bnbarham absolutely, please! This was meant to be a "demo" PR really :)

Note that I didn't get to merging @ahoppen's improvements posted as a diff here (as it wasn't clear if this was going anywhere):
https://github.com/swiftlang/sourcekit-lsp/pull/2315#pullrequestreview-3349115617

roman-bcny:
@ahoppen apologies, I actually wasn't aware of your comments above Oct 19 and after - notifications getting lost in the noise :\

bnbarham:
> @bnbarham absolutely, please! This was meant to be a "demo" PR really :)

Thanks so much for getting it started. Really appreciate your effort here!

> Note that I didn't get to merging @ahoppen's improvements posted as a diff here (as it wasn't clear if this was going anywhere):

Ah I forgot about that one, thanks - I'll make sure it's in too!

bnbarham:
Put that up https://github.com/swiftlang/swift-tools-protocols/pull/26

ahoppen:
Just want to join @bnbarham in giving a huge thank you to get this change started, @roman-bcny 😍 It’s going to be great for users of SourceKit-LSP on Windows.

Let’s close this PR as the changes are incorporated into https://github.com/swiftlang/swift-tools-protocols/pull/26.

roman-bcny:
Woohoo awesome - thanks so much @bnbarham for driving this and your PR is looking so much better than what I could've possibly made at this point with so little experience with Swift and a rather involved threading/queue interaction.

@ahoppen sorry again this got lost in my inbox, will do my best to not let this happen again. I'm catching up now and from what I understand we're getting around the `readabilityHandler` + `availableData` issue in this specific case. However, this seems like something that can bite us or users again in the future. I'll see if I can find a way to repro any hiccups following the info in https://github.com/swiftlang/sourcekit-lsp/pull/2315#issuecomment-3419305142 and post an update here either way. If you happen to have stumbled upon any additional clues since, please lmk!

ahoppen:
No worries at all for not replying, @roman-bcny. These things happen and thank you so much for all your initial investigations .

> However, this seems like something that can bite us or users again in the future.

I think we should be good if we aren’t using `availableData`.

PR #2321, opened 2025-10-08
Update formatting test so we can check that only whitespace edits are produced

ahoppen:
Change the test case to something a little more realistic than lines containing A, B, and C. Then add the check again that all changes reported are whitespaces, which was remvoed by https://github.com/swiftlang/sourcekit-lsp/pull/2291.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2324, opened 2025-10-15
Adopt swift-tools-protocols

owenv:
Adopt the new package and remove duplicated code


owenv:
I expect I'll need to iterate on this a bit. https://github.com/swiftlang/swift/pull/84882 needs to land before it's testable in CI

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -1085,7 +1087,7 @@ package actor BuildServerManager: QueueBasedMessageHandler {
   ) async -> FileBuildSettings? {
     let buildSettingsFromBuildServer = await orLog("Getting build settings") {
       if fallbackAfterTimeout {
-        try await withTimeout(options.buildSettingsTimeoutOrDefault) {
+        try await SwiftExtensions.withTimeout(options.buildSettingsTimeoutOrDefault) {
```
ahoppen: Why do we need to module-qualify this? Shouldn’t this be at a below-SPI access level in swift-tools-protocols and thus there wouldn’t be a name conflict?

Sources/BuildServerIntegration/BuildTargetIdentifierExtensions.swift
```diff
@@ -146,7 +146,7 @@ extension BuildTargetIdentifier {
   }
 }
 
-extension BuildTargetIdentifier: CustomLogStringConvertible {
+@_spi(SourceKitLSP) extension BuildTargetIdentifier: @retroactive CustomLogStringConvertible {
```
ahoppen, line 149: Hmm, should this conformance be in swift-tools-protocols? Seems like most people would need to add it. Though I’m not quite sure which module that conformance would live in.

Sources/sourcekit-lsp/SourceKitLSP.swift
```diff
@@ -234,18 +234,20 @@ struct SourceKitLSP: AsyncParsableCommand {
       fatalError("failed to redirect stdout -> stderr: \(strerror(errno)!)")
     }
 
+    LoggingScope.configureDefaultLoggingSubsystem("org.swift.sourcekit-lsp")
```
ahoppen, line 237: Hmm, this is a little tricky. We’d need to do the same in at least one other places (part from test execution), namely when launching a SourceKitLSP server in-processes using `InProcessClient`

Tests/BuildServerIntegrationTests/BuildServerManagerTests.swift
```diff
@@ -82,6 +83,10 @@ private func createBuildServerManager(
 }
 
 final class BuildServerManagerTests: XCTestCase {
+  override func setUp() async throws {
```
ahoppen: I really dislike having this `setUp` function in every single test case. Should we instead have a custom class that inherits from `XCTestCase` and that all our test cases inherit from?

Also, could we make `LoggingScope.subsystem` not `fatalError` if no default logging subsystem was set? Instead, maybe just use `org.swift` or `default` and log a fault that no logging subsystem has been set? There’s no need to crash the process just because no logging subsystem has been configured.

ahoppen (commented):
Sorry, hit submit to early

Tests/BuildServerIntegrationTests/BuildServerManagerTests.swift
owenv: I can get behind not crashing, https://github.com/swiftlang/swift-tools-protocols/pull/16

Sources/BuildServerIntegration/BuildServerManager.swift
owenv: https://github.com/swiftlang/swift-tools-protocols/pull/17

Sources/BuildServerIntegration/BuildTargetIdentifierExtensions.swift
owenv: We could put it in LanguageServerProtocolTransport although it's a bit of a stretch

Sources/sourcekit-lsp/SourceKitLSP.swift
owenv: added the missing configure call to InProcessClient

Tests/BuildServerIntegrationTests/BuildServerManagerTests.swift
owenv: Also introduced an XCTestCase subclass for XCTests, and a suite trait for Swift Testing Tests

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test windows

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test

Sources/BuildServerIntegration/BuildTargetIdentifierExtensions.swift
ahoppen: In SourceKit-LSP, I’d put this in `SKUtilities` but I don’t think it’s worth adding a module in `swift-tools-protocols` just for this conformance. Let’s leave it for now and try to remember that if we should introduce such a module in swift-tools-protocols, we could move this conformance.

Tests/SourceKitDTests/SourceKitDRegistryTests.swift
```diff
@@ -11,11 +11,12 @@
...
-import LanguageServerProtocolExtensions
+@_spi(SourceKitLSP) import LanguageServerProtocolExtensions
 import SKTestSupport
 import SourceKitD
 import SwiftExtensions
 import TSCBasic
+@_spi(SourceKitLSP) import ToolsProtocolsSwiftExtensions
 import XCTest
 
 final class SourceKitDRegistryTests: XCTestCase {
```
ahoppen: Shouldn’t this also be `SourceKitLSPTestCase`? Are there any other lingering `XCTestCase` subclasses?

Package.swift
```diff
@@ -135,14 +106,15 @@ var targets: [Target] = [
...
       "LanguageServerProtocolExtensions",
-      "LanguageServerProtocolJSONRPC",
-      "SKLogging",
+      .product(name: "LanguageServerProtocolTransport", package: "swift-tools-protocols"),
+      .product(name: "SKLogging", package: "swift-tools-protocols"),
       "SKOptions",
       "SourceKitLSP",
       "SwiftExtensions",
       "ToolchainRegistry",
+      .product(name: "ToolsProtocolsSwiftExtensions", package: "swift-tools-protocols"),
```
ahoppen: We generally sort the dependencies by targets in the `sourcekit-lsp` package (alphabetically), followed by external dependencies (also alphabetically), can you sort the new `swift-tools-protocols` dependencies accordingly as well?

Package.swift
```diff
@@ -871,6 +791,7 @@ var dependencies: [Package.Dependency] {
       .package(url: "https://github.com/swiftlang/swift-docc.git", branch: relatedDependenciesBranch),
       .package(url: "https://github.com/swiftlang/swift-docc-symbolkit.git", branch: relatedDependenciesBranch),
       .package(url: "https://github.com/swiftlang/swift-markdown.git", branch: relatedDependenciesBranch),
+      .package(url: "https://github.com/swiftlang/swift-tools-protocols.git", from: "0.0.7"),
```
ahoppen: Since update-checkout uses an exact version, should this also use `exact: "0.0.7"`? Might even be necessary because the `0.` releases aren’t guaranteed to be API-stable and we don’t want old commits of sourcekit-lsp to be unable to build because a new version of `swift-tools-protocols` has been released.

Sources/SKTestSupport/SourceKitLSPTestCase.swift
```diff
@@ -0,0 +1,40 @@
...
+
+import SKLogging
+package import Testing
+public import XCTest
+
+/// Base class for a unit tests in SourceKit-LSP.
+/// Handles configuring the logging system before running tests.
+open class SourceKitLSPTestCase: XCTestCase {
+  override public func setUp() async throws {
+    LoggingScope.configureDefaultLoggingSubsystem("org.swift.sourcekit-lsp-tests")
```
ahoppen: Nitpick opinion, feel free to ignore (same adjustment for the test trait)

```suggestion
    LoggingScope.configureDefaultLoggingSubsystem("org.swift.sourcekit-lsp.tests")
```

Sources/SKTestSupport/SourceKitLSPTestCase.swift
```diff
@@ -0,0 +1,40 @@
...
+//===----------------------------------------------------------------------===//
+
+import SKLogging
+package import Testing
+public import XCTest
+
+/// Base class for a unit tests in SourceKit-LSP.
+/// Handles configuring the logging system before running tests.
+open class SourceKitLSPTestCase: XCTestCase {
+  override public func setUp() async throws {
```
ahoppen: Since configuring the logging scope is set globally, I think `override static func` should also work. Also, there’s no need for this function to be `async throws` at the moment.

ahoppen (commented):
A couple small comments, otherwise LGTM.

Tests/SourceKitDTests/SourceKitDRegistryTests.swift
owenv: I fixed this instance. There are a few unit tests of data structures and things which I left subclasses of SKTestCase to avoid adding an SKTestSupport dependency where it wasn't needed

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test

owenv:
https://github.com/swiftlang/swift-installer-scripts/pull/474
https://github.com/swiftlang/swift/pull/85078
@swift-ci test windows

Sources/BuildServerProtocol/Messages/CancelRequestNotification.swift
ahoppen, line 1: Accidentally removed after Rintaro added this?

ahoppen (approved):
One small comment, otherwise LGTM.

Sources/BuildServerProtocol/Messages/CancelRequestNotification.swift
owenv: This moved to swift-tools-protocols, so it's ok to remove here

ahoppen: Sorry, must have forgotten which repo I was reviewing 🤦🏽‍♂️

PR #2326, opened 2025-10-19
Add a code action to remove unused imports in a source file

ahoppen:
The idea is pretty simple: When `MemberImportVisibility` is enabled, we know that imports can only affect the current source file. So, we can just try and remove every single `import` declaration in the file, check if a new error occurred and if not, we can safely remove it.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
+++ b/Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
@@ -1,0 +1,304 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import BuildServerIntegration
+import Csourcekitd
+import Foundation
+package import LanguageServerProtocol
+import SKLogging
+import SourceKitD
+import SourceKitLSP
+import SwiftExtensions
+import SwiftSyntax
+
+/// The remove unused imports command tries to remove unnecessary imports in a file on a best-effort basis by deleting
+/// imports in reverse source order and seeing if the file still builds. Note that while this works in most cases, there
+/// are a few edge cases, in which this isn't correct. We decided that those are rare enough that the benefits of the
+/// refactoring action outweigh these potential issues.
+///
+/// ### 1. Overload resolution changing
+///
+/// LibA.swift
+/// ```swift
+/// func foo(_ x: Int) -> Int { "Wrong" }
+/// ```
+///
+/// LibB.swift
+/// ```swift
+/// func foo(_ x: Double) -> Int { "Correct" }
+/// ```
+///
+/// Test.swift
+/// ```swift
+/// import LibA
+/// import LibB
+///
+/// print(foo(1.2))
+/// ```
+///
+/// The action will remove the import to LibB because the code still compiles fine without it (we now pick the
+/// `foo(_:Int)` overload instead of `foo(_:Double)`). This seems pretty unlikely though.
+///
+/// ### 2. Loaded extension used by other source file
+///
+/// Importing a module in this file might make members and conformances available to other source files as well, so just
+/// checking the current source file for issues is not technically enough. The former of those issues is fixed by the
+/// upcoming `MemberImportVisibility` language feature and importing a module and only using a conformance from it in a
+/// different file seems pretty unlikely.
+package struct RemoveUnusedImportsCommand: SwiftCommand {
+  package static let identifier: String = "remove.unused.imports.command"
+  package var title: String = "Remove Unused Imports"
+
+  /// The text document related to the refactoring action.
+  package var textDocument: TextDocumentIdentifier
+
+  internal init(textDocument: TextDocumentIdentifier) {
+    self.textDocument = textDocument
+  }
+
+  package init?(fromLSPDictionary dictionary: [String: LanguageServerProtocol.LSPAny]) {
+    guard case .dictionary(let documentDict)? = dictionary[CodingKeys.textDocument.stringValue] else {
+      return nil
+    }
+    guard let textDocument = TextDocumentIdentifier(fromLSPDictionary: documentDict) else {
+      return nil
+    }
+
+    self.init(
+      textDocument: textDocument
+    )
+  }
+
+  package func encodeToLSPAny() -> LSPAny {
+    return .dictionary([
+      CodingKeys.textDocument.stringValue: textDocument.encodeToLSPAny()
+    ])
+  }
+}
+
+extension SwiftLanguageService {
+  func retrieveRemoveUnusedImportsCodeAction(_ request: CodeActionRequest) async throws -> [CodeAction] {
+    let snapshot = try await self.latestSnapshot(for: request.textDocument.uri)
+
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    guard
+      let node = SyntaxCodeActionScope(snapshot: snapshot, syntaxTree: syntaxTree, request: request)?
+        .innermostNodeContainingRange,
+      node.findParentOfSelf(ofType: ImportDeclSyntax.self, stoppingIf: { _ in false }) != nil
+    else {
+      // Only offer the remove unused imports code action on an import statement.
+      return []
+    }
+
+    guard
+      let buildSettings = await self.compileCommand(for: request.textDocument.uri, fallbackAfterTimeout: true),
+      !buildSettings.isFallback,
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Sleeping over this, I think there is at least one case where this action could change behavior.

LibA.swift
```swift
func foo(_ x: Int) -> Int { "Wrong" }
```

LibB.swift
```swift
func foo(_ x: Double) -> Int { "Correct" }
```

Test.swift
```swift
import LibA
import LibB

print(foo(1.2))
```

The action will remove the import to LibB because the code still compiles fine without it (we now pick the `foo(_:Int)` overload instead of `foo(_:Double)`). But I think these cases are extremely rare, so I wouldn’t worry about them for now.

tshortli:
It's probably not too impactful to your goal, but `MemberImportVisibility` does not really guarantee that a source file only depends on its direct imports. Conformances are one example of something that can be used from a source file without a direct import. Re-exports also make the analysis hard in the general case since the same declarations may be reachable through multiple imports in the same file.

tshortli:
As a concrete example of the difficulty with conformances, an "unused" import in one source file could be the only import in the entire module that makes a necessary conformance visible in another source file. So removing that unused import would cause an error in another file.

ahoppen:
Oh, thanks for mentioning the conformance issue. I only thoughts about members in extensions, not about the actual conformance that might be coming from a different file in the same module.

I’m on the edge now whether we should take this, knowing that it’s technically incorrect or if we say it’s good enough in most cases (which I suspect it is).

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    guard let buildSettings = await self.compileCommand(for: request.textDocument.uri, fallbackAfterTimeout: true),
+      !buildSettings.isFallback,
+      buildSettings.compilerArgs.contains(["-enable-upcoming-feature", "MemberImportVisibility"])
+        || buildSettings.compilerArgs.contains(["-swift-version", "7"])
+    else {
+      // We can only offer the remove unused imports code action if `MemberImportVisibility` is enabled because
+      // otherwise imports of this source file might be necessary for other files to compile.
+      // We assume that `MemberImportVisibility` will be enabled in Swift version 7.
```
bnbarham: Comment is technically incorrect given the discussion. If we're happy taking this change, IMO we don't need this check - it's basically "best effort".

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+          keys.sourceFile: temporaryDocUri.pseudoPath,
+          keys.compilerArgs: patchedCompileCommand.compilerArgs as [SKDRequestValue],
+        ]),
+        snapshot: nil
+      )
+      guard let diagnostics = (response[sourcekitd.keys.diagnostics] as SKDResponseArray?) else {
+        return true
+      }
+      // swift-format-ignore: ReplaceForEachWithForLoop
+      // Reference is to `SKDResponseArray.forEach`, not `Array.forEach`.
```
bnbarham, line 170: We could just rename ours to avoid this 😅 (not this PR and not necessarily at all, just an idea).

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+      guard try await !temporaryDocumentHasErrorDiagnostic() else {
+        // If the source file has errors to start with, we can't check if removing an import declaration would introduce
+        // a new error, give up. This really shouldn't happen anyway because the remove unused imports code action is
+        // only offered if the source file is free of error.
+        throw ResponseError.unknown("Failed to remove unused imports because the document currently contains errors")
+      }
+
+      // Only consider import declarations at the top level and ignore ones eg. inside `#if` clauses since those might
+      // be inactive in the current build configuration and thus we can't reliably check if they are needed.
+      let importDecls = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
```
bnbarham, line 202: We could evaluate them and still run it for active clauses?

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+
+      var declsToRemove: [ImportDeclSyntax] = []
+
+      // Try removing the import decls and see if the file still compiles without syntax errors. Do this in revers order
```
bnbarham:
```suggestion
      // Try removing the import decls and see if the file still compiles without syntax errors. Do this in reverse order
```

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+      if !applyResponse.applied {
+        let reason: String
+        if let failureReason = applyResponse.failureReason {
+          reason = " reason: \(failureReason)"
+        } else {
+          reason = ""
+        }
+        logger.error("client refused to apply edit for removing unused imports: \(reason)")
+      }
+    } cleanup: {
```
bnbarham, line 297: Poor man's defer because of the async call I assume?

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -0,0 +1,278 @@
...
+        )
+      )
+      if !applyResponse.applied {
+        let reason: String
+        if let failureReason = applyResponse.failureReason {
+          reason = " reason: \(failureReason)"
+        } else {
+          reason = ""
+        }
+        logger.error("client refused to apply edit for removing unused imports: \(reason)")
```
bnbarham, line 295: Do you know when this happens? Is it actually an error?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1093,6 +1057,153 @@ final class CodeActionTests: XCTestCase {
...
+      files: [
+        "LibA/LibA.swift": "",
+        "LibB/LibB.swift": "",
+        "Test/Test.swift": """
+        // Some file header
+        // over multiple lines
+
+        1️⃣import LibA
+        2️⃣import Foundation3️⃣
+        import LibB4️⃣
```
bnbarham: Could include some trailing comments to make sure that's handled as well. Maybe leading as well, though that seems far less common.

tshortli:
How will the refactoring handle code like this?

```
import Foundation
import Dispatch

func test(_ queue: DispatchQueue) {
  queue.sync { }
}
```

Either import can be removed since `Foundation` is not used directly anywhere but re-exports `Dispatch`. This is what I mean about re-exports making the analysis difficult. I think ideally only the Dispatch import would be kept as the most specific import that is required by the code, but that can't really be determined outside of the compiler.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
ahoppen: Something for a follow-up PR 😉

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
ahoppen: Correct. Good reminder to add a `#if swift(>=6.4)` warning to remove this once we have https://github.com/swiftlang/swift-evolution/blob/main/proposals/0493-defer-async.md

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
ahoppen: I don’t know when the client might refuse to apply the edits. But I think it’s worth logging it so that if we get an issue that refactoring edits aren’t applied, we know what’s going on.

Tests/SourceKitLSPTests/CodeActionTests.swift
ahoppen: Added two comments to the test case.

ahoppen:
The refactoring action just tries to remove imports in the reverse order, so in this case it would remove `import Dispatch`. If the imports would have been spelled the other way round, `import Foundation` would have been removed. Which seems good enough for me. Otherwise you could argue that this action would try and minimize imports and should thus also replace `import Foundation` with `import Dispatch` if only re-exported declarations are used. And I think that’s out-of-scope (at least for now).

ahoppen:
Also added a few comment lines on the known drawbacks of this refactoring action and why we decided to include it despite these.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2329, opened 2025-10-22
Add missing parentheses around logging scope computation

ahoppen:
The modulo operator associated `0` and `100`, so the computation here was essentially `handle?.numericValue ?? (0 % 100)`, equivalent to `handle?.numericValue ?? 0`, which means that we didn’t acutally perform modulo operations on the numeric value, which means that we would exceed the maximum number of `os_log_t` objects after some time.

rdar://162891887


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #798, opened 2023-08-11
Subscripts do not appear in the completion list


tristanlabelle:
Function completion lists do not include subscripts. This would be especially useful to discover more exotic subscripts as `dict[_:default:]`

![image](https://github.com/apple/sourcekit-lsp/assets/2314287/7aef8089-e646-4dea-b7d4-8e3492e67707)

ahoppen:
Tracked in Apple’s issue tracker as rdar://113768792

ahoppen:
Looking at this again, I just want to note that we do provide completions for key paths when typing `[`, so this issue is just for providing operator completions when in function completion mode after `.`. The slight difference is that `[` is not a completion trigger character because it’s mostly used for array and dictionary literals and types in which it’s usually not useful to invoke code completion straight away. Thus you don’t get into that completion mode as often.

PR #2332, opened 2025-10-25
Test that we get subscript completions when completing after `[`

ahoppen:
Noticed that we don’t have any test for this when investigating #798.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2333, opened 2025-10-25
Remove TODO that `filterText` of override completions should not contain `func` in filterText

ahoppen:
The current behavior is the correct one. Otherwise you don’t get any override completions in VS Code after typing `func <something>`.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2338, opened 2025-10-28
Don’t set empty array values in the `SourceKitLSPOptions` from command-line arguments

ahoppen:


ahoppen:
@swift-ci Please test

PR #2340, opened 2025-10-29
Parse #Playground macro expansions and include them in CodeLens request

award999:
Support swift.play in textDocument/codelens request
- New `swift.play` CodeLens support that is an experimental feature while [swift play](https://github.com/apple/swift-play-experimental/) is still experimental
- Add #Playground macro visitor to parse the macro expansions
- File must `import Playgrounds` to record the macro expansion
- The `swift-play` binary must exist in the toolchain to
- TextDocumentPlayground will record the id and optionally label to match detail you get from
```
$ swift play --list
Building for debugging...
Found 1 Playground
* Fibonacci/Fibonacci.swift:23 "Fibonacci"
```
- Add LSP extension documentation for designing pending `workspace/playground` request
- Add new parsing test cases
- Update CMake files

Issue: https://github.com/swiftlang/sourcekit-lsp/issues/2339 https://github.com/swiftlang/sourcekit-lsp/issues/2343

### DIFF
--- a/Sources/SwiftLanguageService/CMakeLists.txt
+++ b/Sources/SwiftLanguageService/CMakeLists.txt
@@ -25,7 +25,7 @@ add_library(SwiftLanguageService STATIC
   InlayHints.swift
   MacroExpansion.swift
   OpenInterface.swift
-  SwiftPlaygroundScanner.swift
+  SwiftPlaygroundsScanner.swift
   RefactoringEdit.swift
   RefactoringResponse.swift
   RelatedIdentifiers.swift


ahoppen (commented):
I haven’t looked at this in full detail yet but is there a reason why this needs to be an LSP extension and can’t be code lenses returned from the `textDocument/codeLens` request?

award999:
@ahoppen that's the next step, here is commit https://github.com/award999/sourcekit-lsp/commit/061075fde6a607dfebcf3f10dc69f63a842373e2 was just waiting for this piece first before staging that PR

award999:
> is there a reason why this needs to be an LSP extension and can’t be code lenses returned from the `textDocument/codeLens` request?

intent is to be listing the Playgrounds in the project panel. We'll do a workspace/playgrounds request on startup for initial list, and on document changes, fetch the list of textDocument/playgrounds to update the list, without heavy weight operation of fetching whole list and updating the whole tree

Sources/LanguageServerProtocol/Requests/DocumentPlaygroundsRequest.swift
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A request that returns #Playground macro expansion locations within a file.
+///
+/// **(LSP Extension)**
```
ahoppen: Could you also add an entry in `LSP Extensions.md`?

Sources/LanguageServerProtocol/Requests/DocumentPlaygroundsRequest.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct DocumentPlaygroundsRequest: TextDocumentRequest, Hashable {
```
ahoppen: Could you also add this request to [swift-tools-protocols](http://github.com/swiftlang/swift-tools-protocols)

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
```diff
@@ -0,0 +1,48 @@
...
+  /// Identifier for the `PlaygroundItem`.
+  ///
+  /// This identifier uniquely identifies the playground. It can be used to run an individual playground with `swift play`.
+  public var id: String
+
+  /// Display name describing the test.
+  public var label: String?
+
+  /// The range of the playground item in the source code.
+  public var location: Location
```
ahoppen: The name of the document is already implied by the `textDocument` parameter of the request, so this could just be `Range<Position>`.

Sources/LanguageServerProtocol/SupportTypes/PlaygroundItem.swift
```diff
@@ -0,0 +1,48 @@
...
+//===----------------------------------------------------------------------===//
+
+/// A playground item that can be used to identify playgrounds alongside a source file.
+public struct PlaygroundItem: ResponseType, Equatable {
+  /// Identifier for the `PlaygroundItem`.
+  ///
+  /// This identifier uniquely identifies the playground. It can be used to run an individual playground with `swift play`.
+  public var id: String
+
+  /// Display name describing the test.
```
ahoppen: Typo?

```suggestion
  /// Display name describing the playground.
```

Sources/LanguageServerProtocol/SupportTypes/PlaygroundItem.swift
```diff
@@ -0,0 +1,48 @@
...
+    id: String,
+    label: String?,
+    location: Location,
+  ) {
+    self.id = id
+    self.label = label
+    self.location = location
+  }
+
+  public func encodeToLSPAny() -> LSPAny {
```
ahoppen: Why do we need the ability to encode this as `LSPAny`? I don’t see where that is used but I might be missing something.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -527,6 +532,10 @@ package extension LanguageService {
     throw ResponseError.internalError("syntacticDocumentTests not implemented in \(Self.self) for \(uri)")
   }
 
+  func syntacticDocumentPlaygrounds(for uri: DocumentURI, in workspace: Workspace) async throws -> [PlaygroundItem] {
+    throw ResponseError.requestNotImplemented(DocumentOnTypeFormattingRequest.self)
```
ahoppen:
```suggestion
    throw ResponseError.requestNotImplemented(DocumentPlaygroundsRequest.self)
```

Sources/LanguageServerProtocol/SupportTypes/PlaygroundItem.swift
bnbarham: Not a test 😅

bnbarham (commented):
> > is there a reason why this needs to be an LSP extension and can’t be code lenses returned from the textDocument/codeLens request?
> intent is to be listing the Playgrounds in the project panel. We'll do a workspace/playgrounds request on startup for initial list, and on document changes, fetch the list of textDocument/playgrounds to update the list, without heavy weight operation of fetching whole list and updating the whole tree

I suppose it is a reasonable point that `textDocument/playgrounds` is unlikely to ever need a hierarchy (which is presumably the main reason `textDocument/tests` exists). So I could see arguments for either and regardless `codeLens` would have custom data for this case. I don't particularly mind either way, how strongly do you feel about having in `textDocument/codeLens` @ahoppen?

> We'll do a workspace/playgrounds request on startup for initial list, and on document changes, fetch the list of textDocument/playgrounds to update the list

Worth noting that I was looking at test discovery with @rintaro recently and I believe we're not handling external changes today (I'm assuming test discovery following a similar process).

Sources/SwiftLanguageService/PlaygroundsMacro.swift
```diff
@@ -0,0 +1,130 @@
...
+
+final class PlaygroundMacroFinder: SyntaxAnyVisitor {
+  /// The base ID to use to generate IDs for any playgrounds found in this file.
+  private let baseID: String
+
+  /// The snapshot of the document for which we are getting playgrounds.
+  private let snapshot: DocumentSnapshot
+
+  /// Only playgrounds that intersect with this range get reported.
+  private let range: Range<AbsolutePosition>
```
ahoppen: Are we using this anywhere? If not, should the `textDocument/playgrounds` request take a range?

Sources/SwiftLanguageService/PlaygroundsMacro.swift
```diff
@@ -0,0 +1,130 @@
...
+  private init(baseID: String, snapshot: DocumentSnapshot, range: Range<AbsolutePosition>) {
+    self.baseID = baseID
+    self.snapshot = snapshot
+    self.range = range
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  /// Designated entry point for `PlaygroundMacroFinder`.
+  static func find(
+    in nodes: some Sequence<Syntax>,
```
ahoppen: Why do we need to take a sequence of syntax nodes here? I think `some SyntaxProtocol` would be sufficient.

Sources/SwiftLanguageService/PlaygroundsMacro.swift
```diff
@@ -0,0 +1,130 @@
...
+    snapshot: DocumentSnapshot,
+    range: Range<AbsolutePosition>
+  ) async -> [PlaygroundItem] {
+    guard let canonicalTarget = await workspace.buildServerManager.canonicalTarget(for: snapshot.uri),
+      let moduleName = await workspace.buildServerManager.moduleName(for: snapshot.uri, in: canonicalTarget),
+      let baseName = snapshot.uri.fileURL?.lastPathComponent 
+    else {
+      return []
+    }
+    let visitor = PlaygroundMacroFinder(baseID: "\(moduleName)/\(baseName)", snapshot: snapshot, range: range)
```
ahoppen: Doesn’t it cause issues if you have two files with the same name in a module, just in different directories?

Sources/SwiftLanguageService/PlaygroundsMacro.swift
```diff
@@ -0,0 +1,130 @@
...
+
+  override func visit(_ node: ImportPathComponentSyntax) -> SyntaxVisitorContinueKind {
+    if node.name.text == "Playgrounds" {
+      isPlaygroundImported = true
+    }
+    return .skipChildren
+  }
+
+  override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+    guard isPlaygroundImported, node.macroName.text == "Playground" else {
```
ahoppen: While unconventional, the `import Playground` could be at the bottom of the source file, so we can’t skip the macro expansion expressions just because `Playgrounds` hasn’t been imported yet.

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,130 @@
...
+
+  override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+    guard isPlaygroundImported, node.macroName.text == "Playground" else {
+      return .skipChildren
+    }
+
+    let startPosition = snapshot.position(of: node.positionAfterSkippingLeadingTrivia)
+    let stringLiteral = node.arguments.first?.expression.as(StringLiteralExprSyntax.self)
+    let playgroundLabel = stringLiteral?.representedLiteralValue
+    let playgroundID = "\(baseID):\(startPosition.line + 1)"
```
ahoppen: What happens if you have two playgrounds on the same line, eg.

```swift
#Playground { print("first") }; #Playground { print("second") }
```

I agree that this looks stupid but you never know 🤷🏽

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,130 @@
...
+  /// Keep track of if "Playgrounds" has been imported
+  private var isPlaygroundImported: Bool = false
+
+  private init(baseID: String, snapshot: DocumentSnapshot, range: Range<AbsolutePosition>) {
+    self.baseID = baseID
+    self.snapshot = snapshot
+    self.range = range
+    super.init(viewMode: .sourceAccurate)
+  }
+
```
ahoppen, line 41: In order to speed this up in case a `range` is specified, I’d implement `visitAny` and then return `.skipChildren` if the node’s range doesn’t intersect with `range`.

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        }6️⃣
+        """
+      ],
+      manifest: """
+        import PackageDescription
+        let package = Package(
+          name: "MyLibrary",
+          targets: [.target(name: "MyLibrary")]
+        )
+        """,
```
ahoppen: You can omit the manifest if it just declares a library named `MyLibrary`, that’s the default.

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
```diff
@@ -0,0 +1,160 @@
...
+import LanguageServerProtocol
+import SKTestSupport
+import SourceKitLSP
+import XCTest
+
+final class DocumentPlaygroundDiscoveryTests: XCTestCase {
+  func testParsePlaygroundsTests() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Sources/MyLibrary/MyLib.swift": """
```
ahoppen: The default is to place source files in `Sources/MyLibrary`, so this can just be


```suggestion
        "MyLib.swift": """
```

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        """
+      ],
+      manifest: """
+        import PackageDescription
+        let package = Package(
+          name: "MyLibrary",
+          targets: [.target(name: "MyLibrary")]
+        )
+        """,
+      enableBackgroundIndexing: false
```
ahoppen: Background indexing is disabled by default, so this isn’t needed

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        public func foo() -> String {
+            "bar"
+        }
+
+        1️⃣#Playground("foo") {
+            print(foo())
+        }2️⃣
+
+        3️⃣#Playground {
+            print(foo())
```
ahoppen: Could you use 2 spaces of indentation in the test cases to match the rest of the codebase?

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,130 @@
...
+  override func visit(_ node: ImportPathComponentSyntax) -> SyntaxVisitorContinueKind {
+    if node.name.text == "Playgrounds" {
+      isPlaygroundImported = true
+    }
+    return .skipChildren
+  }
+
+  override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
```
ahoppen, line 83: As far as I can tel `#Playground` also supports the declaration macro role, so you need to handle `MacroExpandionDeclSyntax` as well. This is important in eg.

```swift
struct Foo {
  #Playground { … }
}
```

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
```diff
@@ -0,0 +1,160 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKTestSupport
+import SourceKitLSP
+import XCTest
+
+final class DocumentPlaygroundDiscoveryTests: XCTestCase {
+  func testParsePlaygroundsTests() async throws {
```
ahoppen: `Tests` suffixes in the test cases don’t really make sense.

ahoppen (commented):
I know that we talked about extending this to a `workspace/playgrounds` request and that the `textDocument/playgrounds` request thus makes sense. I would like to see and discuss the structure of that declaration in this PR because I think that the two requests can heavily influence each other and decide if it would be a better fit to use `textDocument/codeLens` to get the playgrounds in a single document.

Sources/LanguageServerProtocol/SupportTypes/PlaygroundItem.swift
award999: Nope :p

Sources/LanguageServerProtocol/Requests/DocumentPlaygroundsRequest.swift
award999: Dnne

Sources/SwiftLanguageService/PlaygroundsMacro.swift
rintaro: FWIW Swift doesn't support multiple files with the same filename in a module
https://github.com/swiftlang/swift-driver/blob/3269bb37fc6d436cbeeeeaffe2bd29a47dd1d320/Sources/SwiftDriver/Driver/Driver.swift#L2555-L2561
Since the compilation would fail, the playground entry points are not runnable anyway. But still avoiding possible ID duplication would be nice for the clients.

Sources/LanguageServerProtocol/SupportTypes/PlaygroundItem.swift
award999: No seems to work without it, I probably just added it after adding to Location

Sources/SwiftLanguageService/PlaygroundsMacro.swift
award999: Nah I guess I don't see a need...

Sources/SwiftLanguageService/PlaygroundsMacro.swift
award999: Done

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
award999: Fair point

Sources/SwiftLanguageService/PlaygroundsMacro.swift
award999: This is the pattern "swift play" follows when generating these unique IDs

Sources/SwiftLanguageService/PlaygroundsMacro.swift
award999: Thanks

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
award999: My intent is to return PlaygroundItem in a future workspace/playgrounds request where document will be needed. Agree?

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Good catch :)

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
award999: No not needed

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

Tests/SourceKitLSPTests/DocumentPlaygroundDiscoveryTests.swift
award999: Done

award999:
@ahoppen sorry if some redundant comments were spamming you, GH is acting really weird for me, not showing some conversation comments as posted. IDK what's up

Sources/SwiftLanguageService/PlaygroundDiscovery.swift
```diff
@@ -0,0 +1,119 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+package import LanguageServerProtocol
+import SKLogging
+package import SourceKitLSP
+import SwiftSyntax
+import Playgrounds
```
rintaro: Is anything from `Playgrounds` used?

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
rintaro, line 1: Not sure about this file name, `Macro` prefix might be confusing because it's often used for macro implementations. Also, `#Playground` being a "macro" is just an implementation detail. Following test discovery, maybe `PlaygroundDiscovery.swift`?

rintaro (commented):
Could you update `CMakeLists.txt` for the new files?

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
award999: Sure, done

Sources/SwiftLanguageService/PlaygroundDiscovery.swift
award999: Nope good catch, I had been experimenting with the APIs in that library but nothing useful

award999:
> Could you update `CMakeLists.txt` for the new files?

@rintaro done

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
ahoppen: I would prefer to have two separate item types in that case. Even if it means a little bit of duplication, I think on the client side that’s clearer as well because you don’t need to wonder about the case that the `textDocument/playgrounds` request returns playgrounds for a different source file and how you would handle that case.

Sources/LanguageServerProtocol/SupportTypes/Location.swift
```diff
@@ -37,4 +37,11 @@ public struct Location: ResponseType, Hashable, Codable, CustomDebugStringConver
   public var debugDescription: String {
     return "\(uri):\(range.lowerBound)-\(range.upperBound)"
   }
+
+  public func encodeToLSPAny() -> LSPAny {
```
ahoppen: Why does `Location` now need to be encodable as `LSPAny`?

Sources/SwiftLanguageService/PlaygroundDiscovery.swift
```diff
@@ -0,0 +1,118 @@
...
+extension SwiftLanguageService {
+  package func syntacticDocumentPlaygrounds(for uri: DocumentURI, in workspace: Workspace) async throws -> [PlaygroundItem] {
+    let snapshot = try self.documentManager.latestSnapshot(uri)
+
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+
+    try Task.checkCancellation()
+    return
+      await PlaygroundFinder.find(
+        in: Syntax(syntaxTree),
```
ahoppen: Can be slightly simpler now that you take `some SyntaxProtocol`.

```suggestion
        in: syntaxTree,
```

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
ahoppen: Ah, looks like this isn’t really supported by `swift play`, so I guess we don’t need to care about it either 🤷🏽

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
award999: Done

award999: Done

award999:
@ahoppen I did make the document playground response type just a range instead of whole location like you've asked for. I also updated the LSP Extensions doc to add the workspace/playgrounds request to make it easier to design what's next at the same time: https://github.com/swiftlang/sourcekit-lsp/pull/2340/files#diff-005ce06479af2f7c1f5ad69b549cb89b9c26b16b23417f54ec96a81b05c883edR745

My vote still is for a separate `textDocument/playgrounds` request. There are editors that don't support CodeLens, and some who can display the text but not the command execution. So I think it is not very intuitive that you need to know to issue the CodeLens request indicating that you support swift.play as a command and filter out the swift.run and swift.debug commands. Both requests are going to use the same `PlaygroundFinder`, so if you look at the CodeLens request that is the next PR I'll open: https://github.com/award999/sourcekit-lsp/commit/b10cfecdd6a9105b3edfb2cc86d2de0a6695269b#diff-f8a9947dfe74d3bd634971372a2af1b1dc338e18ece6ab591e1f6d2ebed3c73aR74, there is not a lot of extra code to support both these requests and the documentation is a lot simpler to explain, i.e. use a combination of workspace/playgrounds and textDocument/playgrounds...

To answer your question about VSCode, no this does not cause an issue in VS Code but it is more code to make this happen, we'll need a CodeLens request middleware, filter out the CodeLens we care about, and then pass on those playground items for merging.

Sources/LanguageServerProtocol/SupportTypes/Location.swift
award999: In next PR the `CodeLens` will have a list of `Command#arguments` that need to be `LSPAny`

award999:
After further discussion me and @ahoppen have agree we will not include a textDocument/playgrounds request right now, and VSCode can determine updates through the CodeLens request. Will wait and see if other editors have a compelling case where they want to get list of playgrounds but don't support CodeLens

Sources/LanguageServerProtocol/SupportTypes/Location.swift
award999: OK CodeLens request added which is what this is needed for

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
award999: I dug into and it actually is supported. Although `--list` will print:
```
* MyPlaygroundsLibrary/MyPlaygroundsLibrary.swift:14 (unnamed)
* MyPlaygroundsLibrary/MyPlaygroundsLibrary.swift:14 (unnamed)
```

you need to specify the column, so `MyPlaygroundsLibrary/MyPlaygroundsLibrary.swift:14:1`. I'm raising an issue against swift play to include that column when listing. I've added the `:column` to the unique ID returned by CodeLens and verified my local build of `swift play` works as expected. Great catch @ahoppen

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
award999: `Playground macro nil: Must be at file scope (from macro 'Playground')` no looks like it always needs to be at file scope

Contributor Documentation/LSP Extensions.md
```diff
@@ -690,6 +690,52 @@ export interface PeekDocumentsResult {
 }
 ```
 
+## `workspace/playgrounds`
+
+New request for return the list of all #Playground macro expansions in the workspace.
```
ahoppen: We only return the macros, not their expansions (which would be the expanded text)

```suggestion
New request for return the list of all #Playground macros in the workspace.
```

Same in a couple more places below.

Contributor Documentation/LSP Extensions.md
```diff
@@ -690,6 +690,52 @@ export interface PeekDocumentsResult {
...
+
+Primarily designed to allow editors to provide a list of available playgrounds in the project workspace and allow
+jumping to the locations where the #Playground macro was expanded.
+
+The request fetches the list of all macro expansions found in the workspace, returning the location, identifier, and optional label
+when available for each #Playground macro expansion. If you want to keep the list of playgrounds up to date without needing to
+call `workspace/playgrounds` each time a document is changed, you can filter for `swift.play` CodeLens returned by the `textDocument/codelens` request.
```
ahoppen, line 702:
```suggestion
call `workspace/playgrounds` each time a document is changed, you can filter for `swift.play` CodeLens returned by the `textDocument/codeLens` request.
```

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A playground item that can be used to identify playgrounds alongside a source file.
```
ahoppen: Can you make this match the doc comments in `LSP Extensions.md`? Same for the properties.

Also could you indicate this as an LSP extension like we do for other types?

Sources/LanguageServerProtocol/SupportTypes/TextDocumentPlayground.swift
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A playground item that can be used to identify playground. Differs from `TextDocumentPlayground`
+/// by not including location which is given for `textDocument/playgrounds` request
+public struct TextDocumentPlayground: ResponseType, Equatable {
```
ahoppen: Should this conform to `LSPAnyCodable` so that it can also be easily decoded from `LSPAny` if anyone writes an LSP client using these types? Same for `Playground`.

Sources/SwiftLanguageService/SwiftCodeLensScanner.swift
```diff
@@ -42,19 +45,38 @@ final class SwiftCodeLensScanner: SyntaxVisitor {
...
+    if snapshot.text.contains("@main") {
+      let visitor = SwiftCodeLensScanner(snapshot: snapshot, targetName: targetDisplayName, supportedCommands: supportedCommands)
+      visitor.walk(syntaxTree)
+      codeLenses += visitor.result
+    }
+
+    // "swift.play" CodeLens should be ignored if "swift-play" is not in the toolchain as the client has no way of running
+    if toolchain.swiftPlay != nil, let workspace, let playCommand = supportedCommands[SupportedCodeLensCommand.play], snapshot.text.contains("#Playground") {
+      let playgrounds = await SwiftPlaygroundsScanner.findDocumentPlaygrounds(in: syntaxTree, workspace: workspace, snapshot: snapshot)
+      codeLenses += playgrounds.map({ p in CodeLens(range: p.range, command: Command(title: "Play \"\(p.label ?? p.id)\"", command: playCommand, arguments: [p.encodeToLSPAny()])) })
```
ahoppen:
```suggestion
      codeLenses += playgrounds.map({ CodeLens(range: $0.range, command: Command(title: "Play \"\($0.label ?? $0.id)\"", command: playCommand, arguments: [$0.encodeToLSPAny()])) })
```

Also seems like this is a very long line, could you [run swift-format](https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting)?

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
ahoppen: Oh, interesting.

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,101 @@
...
+  private let baseID: String
+
+  /// The snapshot of the document for which we are getting playgrounds.
+  private let snapshot: DocumentSnapshot
+
+  /// Accumulating the result in here.
+  private var result: [TextDocumentPlayground] = []
+
+  /// Keep track of if "Playgrounds" has been imported
+  fileprivate var isPlaygroundImported: Bool = false
```
ahoppen:
```suggestion
  private var isPlaygroundImported: Bool = false
```

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    let startPosition = snapshot.position(of: node.positionAfterSkippingLeadingTrivia)
+    let stringLiteral = node.arguments.first?.expression.as(StringLiteralExprSyntax.self)
+    let playgroundLabel = stringLiteral?.representedLiteralValue
+    let playgroundID = "\(baseID):\(startPosition.line + 1):\(startPosition.utf16index + 1)"
+
+    record(
+      id: playgroundID, 
+      label: playgroundLabel,
+      range: node.positionAfterSkippingLeadingTrivia..<node.endPositionBeforeTrailingTrivia
```
ahoppen:
```suggestion
      range: node.trimmedRange
```

Or something like that

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
```diff
@@ -0,0 +1,101 @@
...
+
+  override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+    guard node.macroName.text == "Playground" else {
+      return .skipChildren
+    }
+
+    let startPosition = snapshot.position(of: node.positionAfterSkippingLeadingTrivia)
+    let stringLiteral = node.arguments.first?.expression.as(StringLiteralExprSyntax.self)
+    let playgroundLabel = stringLiteral?.representedLiteralValue
+    let playgroundID = "\(baseID):\(startPosition.line + 1):\(startPosition.utf16index + 1)"
```
ahoppen: Do we know if `swift play` uses UTF-8 or UTF-16 based indices? You should be able to try this by putting a complicated emoji in front of `#Playground` such as

```swift
/* 🧑‍🧑‍🧒‍🧒 */ #Playground { print("blub") }
```

There are functions on `DocumentSnapshot` to convert to UTF-8 based columns as well if that’s what we need.

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -337,6 +343,12 @@ public final class Toolchain: Sendable {
         foundAny = true
       }
 
+      let swiftPlayPath = binPath.appending(component: "swift-play\(execExt)")
```
ahoppen, line 348: Could you also add this to `isSuperset(of:)`? And looks like I forgot to add the checks for `swift-format` and the SourceKit plugins there as well, so if you could add those as well, that would be appreciated.

Tests/SourceKitLSPTests/CodeLensTests.swift
```diff
@@ -12,9 +12,32 @@
...
 import SKTestSupport
+import ToolchainRegistry
 import XCTest
 
 final class CodeLensTests: XCTestCase {
+
+  var toolchain: Toolchain!
+  var toolchainWithSwiftPlay: Toolchain!
+  
+  override func setUp() async throws {
```
ahoppen: I’m generally not a fan of `setUp` functions and their need of force-unwrapped optionals in the class because errors become less diagnosable if things fail in `setUp` in my opinion and because it tends to do work that isn’t necessary for all tests (such as setting up `toolchainWithSwiftPlay` here). I would just add a `fileprivate` extension of `Toolchain` for this

```swift
extension Toolchain {
#if compiler(>6.4)
#warning("Once we require swift-play in the toolchain that’s used to test SourceKit-LPS, we can just use `forTesting`")
#endif
  static var forTestingWithSwiftPlay: Toolchain { 
    get throws {
      ...
    }
 }
}
```

Tests/SourceKitLSPTests/CodeLensTests.swift
```diff
@@ -12,9 +12,32 @@
...
+    toolchain = try await unwrap(ToolchainRegistry.forTesting.default)
+    toolchainWithSwiftPlay = Toolchain(
+      identifier: "\(toolchain.identifier)-swift-swift",
+      displayName: "\(toolchain.identifier) with swift-play",
+      path: toolchain.path,
+      clang: toolchain.clang,
+      swift: toolchain.swift,
+      swiftc: toolchain.swiftc,
+      swiftPlay: URL(string: "/path/to/swift-play"),
```
ahoppen: This seems like it will cause issues on Windows 🫠

Tests/SourceKitLSPTests/CodeLensTests.swift
```diff
@@ -110,7 +156,246 @@ final class CodeLensTests: XCTestCase {
...
+    let toolchainRegistry = ToolchainRegistry(toolchains: [toolchainWithSwiftPlay])
+
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Sources/MyApp/Test.swift": """
+        import Playgrounds
+        1️⃣#Playground { print("Hello Playground!") }2️⃣;  3️⃣#Playground { print("Hello Again!") }4️⃣
+        """
+      ],
+      manifest: """
```
ahoppen: Do we need a custom package manifest here or can we just use the default and then name the file `"Test.swift"` instead of `"Sources/MyApp/Test.swift"`?

Tests/SourceKitLSPTests/CodeLensTests.swift
```diff
@@ -110,7 +156,246 @@ final class CodeLensTests: XCTestCase {
...
+      CodeLensRequest(textDocument: TextDocumentIdentifier(uri))
+    )
+
+    XCTAssertEqual(
+      response,
+      [
+        CodeLens(
+          range: positions["1️⃣"]..<positions["2️⃣"],
+          command: Command(
+            title: "Play \"MyApp/Test.swift:2:1\"",
```
ahoppen: Let’s use raw strings and eliminate the need for escaped quotes.

```suggestion
            title: #"Play "MyApp/Test.swift:2:1""#,
```

Tests/SourceKitLSPTests/CodeLensTests.swift
```diff
@@ -110,7 +156,246 @@ final class CodeLensTests: XCTestCase {
...
+                range: positions["3️⃣"]..<positions["4️⃣"],
+              ).encodeToLSPAny()
+            ]
+          )
+        ),
       ]
     )
   }
+
+  func testCodeLensRequestSwiftPlayMissing() async throws {
```
ahoppen, line 278: This will start failing as soon as `swift-play` is installed in the toolchain. Could you create a toolchain registry that explicitly removes `swiftPlay` from it?

ahoppen (commented):
Overall, this looks good to me, just a few nitpicky comments. Also, all the LSP type definitions are now in `swift-tools-protocols`, so you’ll need to 
- Add the new types there
- Ask @owenv to tag a new release of swift-tools-protocols
- Update the version of swift-tools-protocols that’s used [here](https://github.com/swiftlang/swift/blob/main/utils/update_checkout/update-checkout-config.json#L148) and [here](https://github.com/swiftlang/sourcekit-lsp/blob/0876b7d40397523190d13f1a938004e30aba4546/Package.swift#L803)
- Then you can test this PR again

I hope I didn’t forget any step and sorry for the extra complication with this new process…

Sources/LanguageServerProtocol/SupportTypes/TextDocumentPlayground.swift
award999: Sure

Sources/LanguageServerProtocol/SupportTypes/Playground.swift
award999: Done

Tests/SourceKitLSPTests/CodeLensTests.swift
award999: huh had that, idk why I reverted that, fixed anyway

Tests/SourceKitLSPTests/CodeLensTests.swift
award999: only used for a nil vs. non-nil check

award999: ok I see you use a format like `URL(fileURLWithPath: "/dummy/usr/bin/clang")` elsewhere so I'll add that "/dummy" prefix so no confusion

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
chrismiles: Good catch. Multiple `#Playground`s on the same line is pretty unlikely, but possible as you saw.  `--list` should certainly show the column numbers in those cases. I will make `--list` only show column numbers when needed for ambiguity, as 99.9% of the columns will be 1.

Tests/SourceKitLSPTests/CodeLensTests.swift
ahoppen: OK, we’re in luck then.

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
award999: @chrismiles doing something like in the example above compiles, and if it was named i.e.
`/* 🧑‍🧑‍🧒‍🧒 */ #Playground("bar") { print("blub") }`

then `swift play bar` runs fine, but using the column in the ID, ex. `swift play MyPlaygroundsLibrary/MyPlaygroundsLibrary.swift:15:19`, does not find the playground. So how would the column be computed for the ID with something more complicated like an emoji in the line?

Sources/ToolchainRegistry/Toolchain.swift
award999: Done

Tests/SourceKitLSPTests/CodeLensTests.swift
award999: Simplified where I could

Tests/SourceKitLSPTests/CodeLensTests.swift
award999: Done

Sources/LanguageServerProtocol/Requests/DocumentPlaygroundsRequest.swift
award999: Done https://github.com/swiftlang/swift-tools-protocols/pull/21

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
ahoppen: My bet would be that `swift play` is using UTF-8 based column offsets, so this would be position `15:33`, I think, measured using the following. If that’s the case, you can use `snapshot.sourcekitdPosition(of:)` instead of `snapshot.position(of:)` to get the position with a UTF-8 based column.


```swift
print("/* 🧑‍🧑‍🧒‍🧒 */ ".utf8.count)
print("/* 🧑‍🧑‍🧒‍🧒 */ ".utf16.count)
```

award999: You're right @ahoppen , done

award999:
@ahoppen seems to be utf8 columns for swift play so made that change and added some tests.

Sources/SwiftLanguageService/SwiftPlaygroundsScanner.swift
chrismiles: The line and column locations come from SwiftSyntax parsing, during expansion. It appears to be UTF-8 based, as you've found.

ahoppen:
https://github.com/swiftlang/swift/pull/85420

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift/pull/85420

@swift-ci Please test Windows

Sources/SwiftLanguageService/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ add_library(SwiftLanguageService STATIC
   InlayHints.swift
   MacroExpansion.swift
   OpenInterface.swift
+  PlaygroundDiscovery.swift
```
ahoppen:
```suggestion
  SwiftPlaygroundScanner.swift
```

award999: Great catch! Obviously didn't think of that when renaming the file :/

award999:
https://github.com/swiftlang/swift/pull/85420

@swift-ci Please test

award999:
https://github.com/swiftlang/swift/pull/85420

@swift-ci Please test Windows

award999:
@swift-ci Please test

award999:
https://github.com/swiftlang/swift/pull/85420

@swift-ci Please test Windows

award999:
@swift-ci Please test Windows

owenv:
```
Failed Tests (1):
  Swift(windows-x86_64) :: ModuleInterface/swift_build_sdk_interfaces/find-modules.test-sh
```

@swift-ci test Windows

PR #2342, opened 2025-10-30
[LSP/BSP] Separate base message protocol for LSP and BSP message types

rintaro:
Introduce `LSPRequestType`, `LSPNotificationType`, `BSPRequestType`, and `BSPNotificationType`. And change the base protocol of the each concrete type. This is beneficial for implementing LSP/BSP only message handlers.

### DIFF
--- a/Sources/BuildServerProtocol/CMakeLists.txt
+++ b/Sources/BuildServerProtocol/CMakeLists.txt
@@ -4,6 +4,7 @@ add_library(BuildServerProtocol STATIC
   Messages/BuildShutdownRequest.swift
   Messages/BuildTargetPrepareRequest.swift
   Messages/BuildTargetSourcesRequest.swift
+  Messages/CancelRequestNotification.swift
   Messages/InitializeBuildRequest.swift
   Messages/OnBuildExitNotification.swift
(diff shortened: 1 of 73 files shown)


rintaro:
@swift-ci Please test

xedin:
Out of curiosity, why Type suffix is used here instead of just `{BSP, LSP}Request`?

ahoppen:
Before this PR, it was so we could write functions like `func foo<Request: RequestType>(x: Request) -> Request.Response` where `Request` is just the natural name for the generic parameter.

xedin:
I see, so this is no longer the case then since we can have `<Request: BSPRequest>(...)` right?

rintaro:
:+1: I removed `Type` suffixes from the new protocol names.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #2344, opened 2025-10-31
Update launch configurations

award999:
With 2.12 of the Swift extension, moved away from "program" and instead using "target" and "configuration" so determine the path dynamically


ahoppen:
@swift-ci Please test

PR #2349, opened 2025-11-05
Add comment that SourceKit-LSP needs to be restarted when configuration file is modified

ahoppen:
Motivated by https://forums.swift.org/t/is-a-restart-needed-after-a-change-to-config-json/83022


ahoppen:
@swift-ci Please test

PR #2351, opened 2025-11-05
Temporarily skip macro tests on Amazon Linux

bnbarham:
These are failing on the 2023 bootstrap job, skip for now to allow getting a toolchain out.

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -129,6 +129,16 @@ package actor SkipUnless {
     try XCTSkipUnless(Platform.current == .windows, message)
   }
 
+  package static func platformIsNotAmazonLinux(_ message: String) throws {
+    guard Platform.current == .linux,
+      let release = try? String(contentsOf: URL(filePath: "/etc/system-release"), encoding: .utf8)
+    else {
+      return
+    }
+
+    try XCTSkipUnless(!release.hasPrefix("Amazon"), message)
+  }
+
   package static func platformSupportsTaskPriorityElevation() throws {
     #if os(macOS)
     guard #available(macOS 14.0, *) else {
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -129,6 +130,21 @@ package actor SkipUnless {
...
+  package static func platformIsNotAmazonLinux(_ message: String) throws {
+    if Platform.current != .linux {
+      return
+    }
+
+    let release: String
+    do {
+      release = try localFileSystem.readFileContents(AbsolutePath(validating: "/etc/system-release")).description
+    } catch {
+        return
```
ahoppen: swift-format will fail because of the 4 space indent here.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -129,6 +130,21 @@ package actor SkipUnless {
...
+    }
```
ahoppen: Since we don’t do anything with the error, this can be simplified. Also, no need to drop down to TSC, `Foundation` is perfectly sufficient here (and then remove the `localFileSystem` import)

```suggestion
    guard let release = try? String(contentsOf: URL(filePath: "/etc/system-release"), encoding: .utf8) else {
      return
    }
```

ahoppen (commented):
Don’t like the double negative with `SkipUnless.isNot` but I don’t have a better solution.

bnbarham:
> Don’t like the double negative with SkipUnless.isNot but I don’t have a better solution.

Yeah me neither, I don't care too much though - I'll fix this properly next week

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

bnbarham:
Merging to unblock AL2023 (windows failure is unrelated)

PR #2362, opened 2025-11-20
Resolve `/usr/bin/*` shims on macOS

bnbarham:
CMake was previously doing this itself before 4.0, but seems to be inserting `/usr/bin/*` now. Resolve the `/usr/bin` trampoline ourselves in a similar fashion to swiftly (but with xcrun).

Resolves rdar://163462990.

### DIFF
--- a/Sources/BuildServerIntegration/SwiftToolchainResolver.swift
+++ b/Sources/BuildServerIntegration/SwiftToolchainResolver.swift
@@ -18,10 +18,10 @@ import TSCExtensions
 import struct TSCBasic.AbsolutePath
 import class TSCBasic.Process
 
-/// Given a path to a compiler, which might be a symlink to `swiftly`, this type determines the compiler executable in
-/// an actual toolchain. It also caches the results. The client needs to invalidate the cache if the path that swiftly
-/// might resolve to has changed, eg. because `.swift-version` has been updated.
-actor SwiftlyResolver {
+/// Given a path to a compiler, which might be a symlink to `swiftly` or `/usr/bin` on macOS, this type determines the
+/// compiler executable in an actual toolchain and caches the result. The client needs to invalidate the cache if the
+/// path that this may resolve to has changed, eg. because `.swift-version` or `SDKROOT` has been updated.
+actor SwiftToolchainResolver {
   private struct CacheKey: Hashable {
     let compiler: URL
     let workingDirectory: URL?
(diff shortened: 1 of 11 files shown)


bnbarham:
@swift-ci please test

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
```diff
@@ -18,56 +18,86 @@ import TSCExtensions
...
 
-  /// Check if `compiler` is a symlink to `swiftly`. If so, find the executable in the toolchain that swiftly resolves
-  /// to within the given working directory and return the URL of the corresponding compiler in that toolchain.
-  /// If `compiler` does not resolve to `swiftly`, return `nil`.
+  /// Check if `compiler` is a symlink to `swiftly` or in `/usr/bin` on macOS. If so, find the executable in the
+  /// toolchain that would be resolved to within the given working directory and return the URL of the corresponding
+  /// compiler in that toolchain. If `compiler` does not resolve to `swiftly` or `/usr/bin` on macOS, return `nil`.
   func resolve(compiler: URL, workingDirectory: URL?) async throws -> URL? {
-    let cacheKey = CacheKey(compiler: compiler, workingDirectory: workingDirectory)
+    let cacheKey: CacheKey = CacheKey(compiler: compiler, workingDirectory: workingDirectory)
```
ahoppen: Why is the type annotation necessary

```suggestion
    let cacheKey = CacheKey(compiler: compiler, workingDirectory: workingDirectory)
```

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
```diff
@@ -18,56 +18,86 @@ import TSCExtensions
...
     }
     return nil
   }
 
+  private func resolveXcrunTrampoline(compiler: URL, workingDirectory: URL?) async throws -> URL? {
+    #if !os(macOS)
+    return nil
+    #endif
+
+    guard compiler.deletingLastPathComponent() == URL(fileURLWithPath: "/usr/bin") else {
```
ahoppen: We can avoid a trip to the file system here to check if `/usr/bin` is a directory
```suggestion
    guard compiler.deletingLastPathComponent() == URL(filePath: "/usr/bin/") else {
```

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
```diff
@@ -18,56 +18,86 @@ import TSCExtensions
...
     if FileManager.default.fileExists(at: resolvedCompiler) {
       return resolvedCompiler
     }
     return nil
   }
 
+  private func resolveXcrunTrampoline(compiler: URL, workingDirectory: URL?) async throws -> URL? {
+    #if !os(macOS)
+    return nil
+    #endif
```
ahoppen: This will cause a build warning on platforms that aren’t macOS because the code after `return` will never be executed. I’d suggest


```suggestion
    guard Platform.current == .macOS else {
      return nil
    }
```

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
```diff
@@ -18,56 +18,86 @@ import TSCExtensions
...
+      workingDirectory: try AbsolutePath(validatingOrNil: workingDirectory?.filePath)
+    )
+
+    let resolvedCompiler = URL(
+      fileURLWithPath: try xcrunResult.utf8Output().trimmingCharacters(in: .whitespacesAndNewlines)
+    )
+    if FileManager.default.fileExists(at: resolvedCompiler) {
+      return resolvedCompiler
+    }
+    return nil
```
ahoppen, line 94: Should we return `compiler` here? Just in case the `xcrun` dance fails, it might be a good fallback to just use whatever the user happens to have in `/usr/bin`

Tests/SourceKitLSPTests/CompilationDatabaseTests.swift
```diff
@@ -389,6 +389,43 @@ final class CompilationDatabaseTests: SourceKitLSPTestCase {
     )
     XCTAssertEqual(definition?.locations, [try project.location(from: "1️⃣", to: "2️⃣", in: "header.h")])
   }
+
+  func testLookThroughXcrun() async throws {
```
ahoppen, line 393: How did this test fail before? Did we fail to find sourcekitd?

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
bnbarham: I was initially leaving off workingDirectory in the usr/bin case, but decided to leave it in and then didn't update it back 😅

Tests/SourceKitLSPTests/CompilationDatabaseTests.swift
bnbarham: Yeah, which is also what actually happens locally

Sources/BuildServerIntegration/SwiftToolchainResolver.swift
bnbarham: I did do this initially but after looking at the one caller ended up changing it back to `nil`. `CompilationDatabaseCompileCommand` already does this, ie.
```
    let resolved = await orLog("Resolving compiler") {
      try await toolchainResolver.resolve(
        compiler: URL(fileURLWithPath: compiler),
        workingDirectory: directoryURL(compileCommandsDirectory: compileCommandsDirectory)
      )?.filePath
    }
    if let resolved {
      return resolved
    }
    return compiler
```

Having said that, I'd be fine switching things here so that `resolve` never returns `Optional` and instead gives back the original if the trampoline checks fail, if you'd prefer.

ahoppen: Ah, I missed that, thought I looked at the caller but must have looked at the wrong place or something. I think returning `nil` makes sense then 👍🏽

bnbarham:
@swift-ci please test

PR #2364, opened 2025-11-20
Don’t use `Optional.map` or `Optional.flatMap`

ahoppen:
Replace usages of `Optional.map` and `Optional.flatMap` by if expressions or other expressions.

I personally find `Optional.map` to be hard to read because `map` implies mapping a collection to me. Usually the alternative constructs seem clearer to me.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2366, opened 2025-11-21
Enable some Swift 7 mode features

AnthonyLatsis:


Package.swift
```diff
@@ -688,10 +643,41 @@ var targets: [Target] = [
...
+    // Warning escalation.
+    .unsafeFlags(["-Werror", "ExistentialAny"]),
+  ]
+  if noSwiftPMDependency {
+    globalSwiftSettings.append(
+      .define("NO_SWIFTPM_DEPENDENCY")
+    )
+  }
+
+  for target in targets where target.type != .plugin {
```
ahoppen, line 669: Why not apply this to plugins? We shouldn’t even have any plugin targets.

Package.swift
```diff
@@ -688,10 +643,41 @@ var targets: [Target] = [
...
+    .enableUpcomingFeature("ExistentialAny"),
+    .enableUpcomingFeature("ImmutableWeakCaptures"),
+
+    // Warning escalation.
+    .unsafeFlags(["-Werror", "ExistentialAny"]),
+  ]
+  if noSwiftPMDependency {
+    globalSwiftSettings.append(
+      .define("NO_SWIFTPM_DEPENDENCY")
+    )
```
ahoppen:
```suggestion
    globalSwiftSettings.append(.define("NO_SWIFTPM_DEPENDENCY"))
```

ahoppen (approved):
Thanks for enabling `ExistentialAny`. I was too lazy to do this when I enabled the other upcoming language features. Just two minor comments in the package manifest.

AnthonyLatsis:
> I was too lazy to do this when I enabled the other upcoming language features

And now we have `swift package migrate` 🙂

Package.swift
AnthonyLatsis: Plugin targets do not support Swift settings (you get a package resolution error). I copied this over from https://github.com/swiftlang/swift-format/pull/1096, where we do have plugin targets.

ahoppen: Ah, fair. I don’t have a strong opinion either way.

AnthonyLatsis:
@swift-ci please test

AnthonyLatsis:
@swift-ci please test

PR #2367, opened 2025-11-23
Plugins: install DLL to `bin`, import library to `lib`

compnerd:
Adjust the install rules to account for the different installation expectations on Windows. DLL files are installed into the `bin` and import libraries are to be installed in `lib`.

### DIFF
--- a/Sources/SwiftSourceKitClientPlugin/CMakeLists.txt
+++ b/Sources/SwiftSourceKitClientPlugin/CMakeLists.txt
@@ -17,4 +17,7 @@ target_link_libraries(SwiftSourceKitClientPlugin PRIVATE
   SwiftSourceKitPluginCommon
   $<$<NOT:$<PLATFORM_ID:Darwin>>:FoundationXML>)
 
-install(TARGETS SwiftSourceKitClientPlugin DESTINATION lib)
+install(TARGETS SwiftSourceKitClientPlugin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

ahoppen:
We only use these libraries using `LoadLibraryW`, should it still be installed to `bin`? Also, in case you didn’t, could you check that we pick up the plugin from the new installation location by checking if we still receive code completion results with the SourceKit plugin in the new location?

compnerd:
> We only use these libraries using `LoadLibraryW`, should it still be installed to `bin`? Also, in case you didn’t, could you check that we pick up the plugin from the new installation location by checking if we still receive code completion results with the SourceKit plugin in the new location?

Well, all the DLLs are meant to be in `bin`. It really doesn't matter how you use them. That said, it seems that the installer still put them into `bin`, just did so in the installer.

This PR is paired with https://github.com/swiftlang/swift-installer-scripts/pull/485 and you will see that only the source is changed, not the destination.

compnerd:

compnerd:

compnerd:

rintaro:
https://github.com/swiftlang/swift-installer-scripts/pull/485
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift-installer-scripts/pull/485
@swift-ci Please test Windows

compnerd:
I wonder if CI is setting `SOURCEKIT_LSP_TEST_PLUGIN_PATHS`

PR #2374, opened 2025-12-01
[6.2] Pass objects to linker flags as is in macro test projects

bnbarham:
- **Explanation**: Amazon Linux 2023 tests are failing because it uses `ld.bfd` and errors since these are absolute paths and thus `-l` is not required.
- **Scope**: Test only change
- **Issues**: rdar://165188592
- **Original PRs**: https://github.com/swiftlang/sourcekit-lsp/pull/2269
- **Risk**: None, just fixes tests


bnbarham:
@swift-ci please test

PR #2377, opened 2025-12-03
Do not error if the directory to be created already exists

bnbarham:
Resolves rdar://165785860.


bnbarham:
Merging to unblock CI

PR #2378, opened 2025-12-03
Allow opting in to using the new SwiftPM BSP server when opening a package

owenv:
Depends on https://github.com/swiftlang/swift-package-manager/pull/9129

Add a SourceKit-LSP option to allow opting in to use of the new SwiftPM build server, using the existing ExternalBuildServerAdapter with a pre-generated configuration. This allows us to integrate The Swift Build build system backend without having SourceKitLSP need to continue linking parts of SwiftPM directly

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -280,22 +280,43 @@ private extension BuildServerSpec {
         )
       }
     case .swiftPM:
-      #if !NO_SWIFTPM_DEPENDENCY
-      return await createBuiltInBuildServerAdapter(
-        messagesToSourceKitLSPHandler: messagesToSourceKitLSPHandler,
-        buildServerHooks: buildServerHooks
-      ) { connectionToSourceKitLSP in
-        try await SwiftPMBuildServer(
-          projectRoot: projectRoot,
-          toolchainRegistry: toolchainRegistry,
-          options: options,
-          connectionToSourceKitLSP: connectionToSourceKitLSP,
-          testHooks: buildServerHooks.swiftPMTestHooks
-        )
+      switch options.swiftPMOrDefault.buildSystem {
+      case .swiftbuild:
+        let buildServer = await orLog("Creating external SwiftPM build server") {
+          try await ExternalBuildServerAdapter(
+            projectRoot: projectRoot,
+            config: BuildServerConfig.forSwiftPMBuildServer(
+              projectRoot: projectRoot,
+              swiftPMOptions: options.swiftPMOrDefault,
+              toolchainRegistry: toolchainRegistry
+            ),
+            messagesToSourceKitLSPHandler: messagesToSourceKitLSPHandler
+          )
+        }
+        guard let buildServer else {
+          logger.log("Failed to create external SwiftPM build server at \(projectRoot)")
+          return nil
+        }
+        logger.log("Created external SwiftPM build server at \(projectRoot)")
+        return .external(buildServer)
+      case .native, nil:
+        #if !NO_SWIFTPM_DEPENDENCY
+        return await createBuiltInBuildServerAdapter(
+          messagesToSourceKitLSPHandler: messagesToSourceKitLSPHandler,
+          buildServerHooks: buildServerHooks
+        ) { connectionToSourceKitLSP in
+          try await SwiftPMBuildServer(
+            projectRoot: projectRoot,
+            toolchainRegistry: toolchainRegistry,
+            options: options,
+            connectionToSourceKitLSP: connectionToSourceKitLSP,
+            testHooks: buildServerHooks.swiftPMTestHooks
+          )
+        }
+        #else
+        return nil
+        #endif
       }
-      #else
-      return nil
-      #endif
     case .injected(let injector):
       let connectionToSourceKitLSP = LocalConnection(
         receiverName: "BuildServerManager for \(projectRoot.lastPathComponent)",
--- a/Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
+++ b/Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
@@ -19,6 +19,7 @@ import Foundation
 import SKOptions
 import SwiftExtensions
 import TSCExtensions
+import ToolchainRegistry
 
 import func TSCBasic.getEnvSearchPaths
 import var TSCBasic.localFileSystem
@@ -61,7 +62,7 @@ enum BuildServerNotFoundError: Error {
 /// BSP configuration
 ///
 /// See https://build-server-protocol.github.io/docs/overview/server-discovery#the-bsp-connection-details
-private struct BuildServerConfig: Codable {
+struct BuildServerConfig: Codable {
   /// The name of the build tool.
   let name: String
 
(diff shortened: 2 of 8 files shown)


owenv:
I haven't done much porting of tests yet, because I want to take a look at which ones should stay in sourcekit-lsp vs sink into SwiftPM.

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -181,18 +187,34 @@ struct SwiftPMBuildServerTests {
       expectArgumentsContain("-target", arguments: arguments)  // Only one!
       #if os(macOS)
       let versionString = PackageModel.Platform.macOS.oldestSupportedVersion.versionString
+      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer == true {
+        expectArgumentsContain(
+          "-target",
+          // Account for differences in macOS naming canonicalization
```
bnbarham, line 196: I assume this is a built-in vs swift-build difference rather than a built-in SwiftPM support vs build server? Given this depends on PackageModel anyway, this will probably be one of the tests that will end up moving. For in sourcekit-lsp, checking we have `-target`/`-sdk`/`-I`/`-F` (and not their values) is probably enough. And to be clear, no need to block the PR on this/moving tests in general.

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
```diff
@@ -82,15 +83,90 @@ private struct BuildServerConfig: Codable {
...
+      for flag in buildToolsSwiftCompilerFlags {
+        args.append(contentsOf: ["-Xbuild-tools-swiftc", flag])
+      }
+    }
+    if swiftPMOptions.disableSandbox == true {
+      args.append("--disable-sandbox")
+    }
+    return BuildServerConfig(
+      name: "SwiftPM Build Server",
+      version: "",
```
jakepetroules, line 157: use Swift version tag?

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -76,6 +76,9 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     ///   background indexing.
     public var skipPlugins: Bool?
 
+    /// Whether to use SwiftPM's experimental build server in place of SourceKit-LSP's builtin package support..
```
bnbarham:
```suggestion
    /// Whether to use SwiftPM's experimental build server in place of SourceKit-LSP's builtin package support.
```

(nit-pick - feel free to include with some later change)

I was also considering `useExperimentalBuildServer` here since it's already under SwiftPM, though "SwiftPMBuildServer" is technically more accurate. Don't have a strong opinion either way.

bnbarham (approved):
Woooo 🥳 Thanks for working on this @owenv!

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
bnbarham: We don't have it in sourcekit-lsp right now (we have no --version) and until we have a common version that all the projects can use, I'd prefer not to add it.

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -280,22 +280,42 @@ private extension BuildServerSpec {
...
-        buildServerHooks: buildServerHooks
-      ) { connectionToSourceKitLSP in
-        try await SwiftPMBuildServer(
-          projectRoot: projectRoot,
-          toolchainRegistry: toolchainRegistry,
-          options: options,
-          connectionToSourceKitLSP: connectionToSourceKitLSP,
-          testHooks: buildServerHooks.swiftPMTestHooks
-        )
+      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer == true {
```
ahoppen:
```suggestion
      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer ?? false {
```

I just find this more readable because it specifies which value `nil` should default to and then you’re back to a normal boolean check.

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
```diff
@@ -82,15 +83,90 @@ private struct BuildServerConfig: Codable {
...
+      }
+    }
+    if let buildToolsSwiftCompilerFlags = swiftPMOptions.buildToolsSwiftCompilerFlags {
+      for flag in buildToolsSwiftCompilerFlags {
+        args.append(contentsOf: ["-Xbuild-tools-swiftc", flag])
+      }
+    }
+    if swiftPMOptions.disableSandbox == true {
+      args.append("--disable-sandbox")
+    }
```
ahoppen, line 152: Do we need to forward `skipPlugins` as well or are we just dropping that option (which would be OK with me because I don’t think we’ve needed it so far.

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -76,6 +76,9 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
...
+    public var useExperimentalSwiftPMBuildServer: Bool?
```
ahoppen: Please correct me if I’m wrong but IIUC the `swiftbuild` build system is only supported by the new external BSP server and the old native SwiftPM build system is only supported by the built-in BSP server, right? If that’s the case, should we re-phrase this option to something that essentially picks the build system, eg. `buildSystem?: "swiftbuild"|"native"`?

ahoppen: Could you run `./sourcekit-lsp-dev-utils generate-config-schema` to re-generate the configuration file’s documentation?

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -181,18 +187,34 @@ struct SwiftPMBuildServerTests {
       expectArgumentsContain("-target", arguments: arguments)  // Only one!
       #if os(macOS)
       let versionString = PackageModel.Platform.macOS.oldestSupportedVersion.versionString
+      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer == true {
```
ahoppen:
```suggestion
      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer ?? false {
```

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -181,18 +187,34 @@ struct SwiftPMBuildServerTests {
...
+          arguments: arguments
+        )
+      }
       expectArgumentsContain(
-        "-target",
-        try await hostTriple.tripleString(forPlatformVersion: versionString),
-        arguments: arguments
+        "-sdk",
+        arguments: arguments,
+        allowMultiple: options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer == true
```
ahoppen:
```suggestion
        allowMultiple: options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer ?? false
```

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -181,18 +187,34 @@ struct SwiftPMBuildServerTests {
...
       )
-      expectArgumentsContain("-sdk", arguments: arguments)
       expectArgumentsContain("-F", arguments: arguments, allowMultiple: true)
       #else
       expectArgumentsContain("-target", try await hostTriple.tripleString, arguments: arguments)
       #endif
 
-      expectArgumentsContain("-I", try build.appending(component: "Modules").filePath, arguments: arguments)
+      if options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer != true {
```
ahoppen:
```suggestion
      if !(options.swiftPMOrDefault.useExperimentalSwiftPMBuildServer ?? false) {
```

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -132,8 +138,8 @@ struct SwiftPMBuildServerTests {
     }
   }
 
-  @Test
-  func testBasicSwiftArgs() async throws {
+  @Test(arguments: [SourceKitLSPOptions(), .forTestingExperimentalSwiftPMBuildServer])
```
ahoppen: Did you try running all tests that use `SwiftPMTestProject` with the new build server?

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
owenv: Right now there's nothing to wire it up to in the new build server. We could consider bringing it back before flipping the default if we decide it would be useful though

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
owenv: I haven't audited everything yet but I've gone through a decent number. Some failures due to tests making assumptions that don't hold under the new build system, and a few which look like bugs or bits of missing functionality that'll need to be fixed up

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
owenv: yep, one uses macosx, the other macos. It's pretty annoying but hard to easily align them.

Sources/SKOptions/SourceKitLSPOptions.swift
owenv: right. That makes sense to me, I renamed the option

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
ahoppen: Could you add a comment to it, saying that the option doesn’t have an effect when using the experimental build server. Then I’m fine with ignoring it.

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
ahoppen: Nice. I’d like to find some way to run tests using both configurations at some point (maybe an environment variable and just run `swift test` twice 🤔) but that’s something for the future.

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -122,7 +127,9 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
         linkerFlags: override?.linkerFlags ?? base.linkerFlags,
         buildToolsSwiftCompilerFlags: override?.buildToolsSwiftCompilerFlags ?? base.buildToolsSwiftCompilerFlags,
         disableSandbox: override?.disableSandbox ?? base.disableSandbox,
-        skipPlugins: override?.skipPlugins ?? base.skipPlugins
+        skipPlugins: override?.skipPlugins ?? base.skipPlugins,
+        buildSystem: override?.buildSystem
+          ?? base.buildSystem
```
ahoppen: Fits in one line now

```suggestion
        buildSystem: override?.buildSystem ?? base.buildSystem
```

Sources/SKOptions/SwiftPMBuildSystem.swift
ahoppen, line 1: Add to CMakeLists.txt?

ahoppen (approved):
Just https://github.com/swiftlang/sourcekit-lsp/pull/2378#discussion_r2590931032 and two nits, otherwise LGTM.

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
Fixed up cmake

@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test windows

owenv:
@swift-ci test Windows

owenv:
lldb test timeout

@swift-ci test Windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows

owenv:
lldb test timeout again

@swift-ci test Windows

owenv:
DocC build is broken on Windows

owenv:
@swift-ci test Windows

owenv:
test timeout

@swift-ci test Windows

PR #2379, opened 2025-12-04
Log an error when ignoring an unknown notification sent by the client

ahoppen:
Just noticed that we don’t emit a log message here while reviewing the new SwiftPM BSP server.


ahoppen:
@swift-ci Please test

PR #2385, opened 2025-12-09
Extend copied file mapping to all LSP requests returning locations

loveucifer:
This addresses issue #2276 by ensuring that all LSP requests that return source file locations map copied header files back to their original locations, not just jump-to-definition.

Previously, only the definition request applied this mapping. Now, the following requests also adjust locations for copied files:
- textDocument/references
- textDocument/implementation
- workspace/symbol
- callHierarchy/prepare
- callHierarchy/incomingCalls
- callHierarchy/outgoingCalls
- typeHierarchy/prepare
- typeHierarchy/supertypes
- typeHierarchy/subtypes

This provides consistent navigation behavior, ensuring users are always taken to the original source files instead of build artifacts when possible.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2281,15 +2281,17 @@ extension SourceKitLSPServer {
 
     // Only return a single call hierarchy item. Returning multiple doesn't make sense because they will all have the
     // same USR (because we query them by USR) and will thus expand to the exact same call hierarchy.
-    let callHierarchyItems = await usrs.asyncCompactMap { (usr) -> CallHierarchyItem? in
+    var callHierarchyItems: [CallHierarchyItem] = []
+    for usr in usrs {
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: usr) else {
-        return nil
+        continue
       }
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
-    }.sorted(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
+      callHierarchyItems.append(await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item))
+    }
+    callHierarchyItems.sort(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
 
     // Ideally, we should show multiple symbols. But VS Code fails to display call hierarchies with multiple root items,
     // failing with `Cannot read properties of undefined (reading 'map')`. Pick the first one.
@@ -2358,24 +2360,25 @@ extension SourceKitLSPServer {
       return self.indexToLSPCallHierarchyItem(definition: definition, index: index)
     }
 
-    let calls = await callersToCalls.asyncCompactMap { (caller, callsList) -> CallHierarchyIncomingCall? in
+    var calls: [CallHierarchyIncomingCall] = []
+    for (caller, callsList) in callersToCalls {
       // Resolve the caller's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: caller.usr) else {
-        return nil
+        continue
       }
 
       let locations = callsList.compactMap { indexToLSPLocation2($0.location) }.sorted()
       let remappedLocations = await workspace.buildServerManager.locationsAdjustedForCopiedFiles(locations)
       guard !remappedLocations.isEmpty else {
-        return nil
+        continue
       }
 
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
       let remappedItem = await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
 
-      return CallHierarchyIncomingCall(from: remappedItem, fromRanges: remappedLocations.map(\.range))
+      calls.append(CallHierarchyIncomingCall(from: remappedItem, fromRanges: remappedLocations.map(\.range)))
     }
     return calls.sorted(by: { $0.from.name < $1.from.name })
   }
@@ -2404,26 +2407,27 @@ extension SourceKitLSPServer {
     let callableUsrs = [data.usr] + index.occurrences(relatedToUSR: data.usr, roles: .accessorOf).map(\.symbol.usr)
     let callOccurrences = callableUsrs.flatMap { index.occurrences(relatedToUSR: $0, roles: .containedBy) }
       .filter(\.shouldShowInCallHierarchy)
-    let calls = await callOccurrences.asyncCompactMap { occurrence -> CallHierarchyOutgoingCall? in
+    var calls: [CallHierarchyOutgoingCall] = []
+    for occurrence in callOccurrences {
       guard occurrence.symbol.kind.isCallable else {
-        return nil
+        continue
       }
       guard let location = indexToLSPLocation2(occurrence.location) else {
-        return nil
+        continue
       }
       let remappedLocation = await workspace.buildServerManager.locationAdjustedForCopiedFiles(location)
 
       // Resolve the callee's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: occurrence.symbol.usr) else {
-        return nil
+        continue
       }
 
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
       let remappedItem = await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
 
-      return CallHierarchyOutgoingCall(to: remappedItem, fromRanges: [remappedLocation.range])
+      calls.append(CallHierarchyOutgoingCall(to: remappedItem, fromRanges: [remappedLocation.range]))
     }
     return calls.sorted(by: { $0.to.name < $1.to.name })
   }
@@ -2624,17 +2628,18 @@ extension SourceKitLSPServer {
 
     // Convert occurrences to type hierarchy items
     let occurs = baseOccurs + retroactiveConformanceOccurs
-    let types = await occurs.asyncCompactMap { occurrence -> TypeHierarchyItem? in
+    var types: [TypeHierarchyItem] = []
+    for occurrence in occurs {
       // Resolve the supertype's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: occurrence.symbol.usr) else {
-        return nil
+        continue
       }
 
       let moduleName = definition.location.moduleName
       guard let item = indexToLSPTypeHierarchyItem2(definition: definition, moduleName: moduleName, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
+      types.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
     }
     return types.sorted(by: { $0.name < $1.name })
   }
@@ -2664,7 +2669,8 @@ extension SourceKitLSPServer {
     }
 
     // Convert occurrences to type hierarchy items
-    let types = await occurs.asyncCompactMap { occurrence -> TypeHierarchyItem? in
+    var types: [TypeHierarchyItem] = []
+    for occurrence in occurs {
       if occurrence.relations.count > 1 {
         // An occurrence with a `baseOf` or `extendedBy` relation is an occurrence inside an inheritance clause.
         // Such an occurrence can only be the source of a single type, namely the one that the inheritance clause belongs
@@ -2672,19 +2678,19 @@ extension SourceKitLSPServer {
         logger.fault("Expected at most extendedBy or baseOf relation but got \(occurrence.relations.count)")
       }
       guard let related = occurrence.relations.sorted().first else {
-        return nil
+        continue
       }
 
       // Resolve the subtype's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: related.symbol.usr) else {
-        return nil
+        continue
       }
 
       let moduleName = definition.location.moduleName
       guard let item = indexToLSPTypeHierarchyItem2(definition: definition, moduleName: moduleName, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
+      types.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
     }
     return types.sorted { $0.name < $1.name }
   }


Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,96 @@ package actor BuildServerManager: QueueBasedMessageHandler {
     return locations.map { locationAdjustedForCopiedFiles($0) }
   }
 
+  package func workspaceEditAdjustedForCopiedFiles(_ workspaceEdit: WorkspaceEdit?) async -> WorkspaceEdit? {
```
ahoppen: I don’t think any of these functions need to be `async`.

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,96 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+    guard var edit = workspaceEdit else {
+      return nil
+    }
+    if let changes = edit.changes {
+      var newChanges: [DocumentURI: [TextEdit]] = [:]
+      for (uri, edits) in changes {
+        let newUri = self.locationAdjustedForCopiedFiles(
+          Location(uri: uri, range: Position(line: 0, utf16index: 0)..<Position(line: 0, utf16index: 0))
```
ahoppen: Instead of synthesizing a `Location` with an irrelevant position, I think we should just have a function that adjusts a `DocumentURI` for copied files.

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,96 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+        ).uri
+        newChanges[newUri, default: []].append(contentsOf: edits)
```
ahoppen: Just personal preference 😉 

```suggestion
        newChanges[newUri, default: []] += edits
```

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,96 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+      let remappedLocations = await self.locationsAdjustedForCopiedFiles(locations)
+      return .locations(remappedLocations)
+    case .locationLinks(let locationLinks):
+      var remappedLinks: [LocationLink] = []
+      for link in locationLinks {
+        let newUri = self.locationAdjustedForCopiedFiles(Location(uri: link.targetUri, range: link.targetRange)).uri
+        remappedLinks.append(LocationLink(
+          originSelectionRange: link.originSelectionRange,
+          targetUri: newUri,
+          targetRange: link.targetRange,
```
ahoppen: I think we should use the target range of the adjusted location here. Just in case the location adjustment in the future also modifies the range. Same for the selection range.

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,96 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+  package func typeHierarchyItemAdjustedForCopiedFiles(_ item: TypeHierarchyItem) async -> TypeHierarchyItem {
+    let adjustedLocation = await self.locationAdjustedForCopiedFiles(Location(uri: item.uri, range: item.range))
+    return TypeHierarchyItem(
+      name: item.name,
+      kind: item.kind,
+      tags: item.tags,
+      detail: item.detail,
+      uri: adjustedLocation.uri,
+      range: adjustedLocation.range,
+      selectionRange: item.selectionRange,
```
ahoppen: It might seem redundant now but I’d like to run the selection range through `locationAdjustedForCopiedFiles` as well, just so we have a centralized place that might also adjust the range eg. if it discovers that copying the file removed the first line and thus all line number need to be shifted by 1 or something like that (not proposing that we do this now but it might be something we could consider doing in the future).

Sources/ClangLanguageService/ClangLanguageService.swift
```diff
@@ -634,7 +642,12 @@ extension ClangLanguageService {
       position: renameRequest.position
     )
     let symbolDetail = try await forwardRequestToClangd(symbolInfoRequest).only
-    return (try await edits ?? WorkspaceEdit(), symbolDetail?.usr)
+    let workspaceEdit = try await edits
```
ahoppen:
```suggestion
    let workspaceEdit = try await edits ?? WorkspaceEdit()
```

And then you don’t need the `??` below.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1690,10 +1690,10 @@ extension SourceKitLSPServer {
       }
       try Task.checkCancellation()
       symbolsAndIndex += symbolOccurrences.map {
-        return ($0, index)
+        return ($0, index, workspace)
       }
     }
-    return symbolsAndIndex.sorted(by: { $0.symbol < $1.symbol }).map { symbolOccurrence, index in
+    return await symbolsAndIndex.sorted(by: { $0.symbol < $1.symbol }).asyncMap { symbolOccurrence, index, workspace in
```
ahoppen: I think this can continue to be a `map` if the location adjustment isn’t `async`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2275,12 +2279,38 @@ extension SourceKitLSPServer {
...
       }
-      return self.indexToLSPCallHierarchyItem(definition: definition, index: index)
-    }.sorted(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
+      let location = indexToLSPLocation(definition.location)
+      guard let originalLocation = location else {
+        continue
+      }
+      let remappedLocation = await workspace.buildServerManager.locationAdjustedForCopiedFiles(originalLocation)
+
+      // Create a new CallHierarchyItem with the remapped location (similar to how we handle TypeHierarchyItem)
```
ahoppen: Why can’t we use `indexToLSPCallHierarchyItem` here anymore?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2327,7 +2358,7 @@ extension SourceKitLSPServer {
     var callersToCalls: [Symbol: [SymbolOccurrence]] = [:]
 
     for call in callOccurrences {
-      // Callers are all `calledBy` relations of a call to a USR in `callableUsrs`, ie. all the functions that contain a
+      // Callers are all `calledBy` relations of a call to a USR in `callableUSrs`, ie. all the functions that contain a
```
ahoppen: Unintentional edit?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2305,7 +2335,8 @@ extension SourceKitLSPServer {
 
   func incomingCalls(_ req: CallHierarchyIncomingCallsRequest) async throws -> [CallHierarchyIncomingCall]? {
     guard let data = extractCallHierarchyItemData(req.item.data),
-      let index = await self.workspaceForDocument(uri: data.uri)?.index(checkedFor: .deletedFiles)
+      let workspace = await self.workspaceForDocument(uri: data.uri),
+      let index = await workspace.index(checkedFor: .deletedFiles)
```
ahoppen, line 2320: Superfluous edit?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2275,12 +2279,38 @@ extension SourceKitLSPServer {
...
+      guard let originalLocation = location else {
+        continue
+      }
+      let remappedLocation = await workspace.buildServerManager.locationAdjustedForCopiedFiles(originalLocation)
+
+      // Create a new CallHierarchyItem with the remapped location (similar to how we handle TypeHierarchyItem)
+      let name = index.fullyQualifiedName(of: definition)
+      let symbol = definition.symbol
+
+      let item = CallHierarchyItem(
```
ahoppen: Should we also have an `adjust*` method for `CallHierarchyItem`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2425,15 +2506,25 @@ extension SourceKitLSPServer {
...
-    switch symbol.kind {
-    case .extension:
-      // Query the conformance added by this extension
-      let conformances = index.occurrences(relatedToUSR: symbol.usr, roles: .baseOf)
-      if conformances.isEmpty {
-        name = symbol.name
-      } else {
-        name = "\(symbol.name): \(conformances.map(\.symbol.name).sorted().joined(separator: ", "))"
-      }
+      switch symbol.kind {
```
ahoppen: This looks like incorrect indentation to me. Could you [run swift-format](https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting) on your change?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2425,15 +2506,25 @@ extension SourceKitLSPServer {
...
+        // Add the file name and line to the detail string
+        if let url = remappedLocation.uri.fileURL,
+          let basename = (try? AbsolutePath(validating: url.filePath))?.basename
+        {
+          detail = "Extension at \(basename):\(remappedLocation.range.lowerBound.line + 1)"
+        } else if !definition.location.moduleName.isEmpty {
+          detail = "Extension in \(definition.location.moduleName)"
+        } else {
+          detail = "Extension"
+        }
```
ahoppen: Did you unintentionally duplicate this code?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2493,28 +2584,75 @@ extension SourceKitLSPServer {
...
+          detail = "Extension in \(info.location.moduleName)"
+        } else {
+          detail = "Extension"
+        }
+      default:
+        name = index.fullyQualifiedName(of: info)
+        detail = info.location.moduleName
+      }
+
+       let item = TypeHierarchyItem(
```
ahoppen: Shouldn’t we be able to just all into the `adjust` method for `TypeHierarchyItem` here and not have these 40-ish lines of adjustment code? Same for the other `TypeHierarchyItem` returning functions.

Tests/SourceKitLSPTests/CopiedHeaderTests.swift
```diff
@@ -0,0 +1,234 @@
...
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+    response.sort()
+    var expected = [
+      try project.location(from: "1️⃣", to: "1️⃣", in: "Test.h"),
+      try project.location(from: "2️⃣", to: "2️⃣", in: "Test.c"),
+    ]
+    expected.sort()
+    XCTAssertEqual(response, expected)
```
ahoppen: Ordering should be deterministic, so I don’t think that `.sort()` should be needed.

Tests/SourceKitLSPTests/CopiedHeaderTests.swift
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  func DISABLED_testFindImplementationInCopiedHeader() async throws {
```
ahoppen: Why is this disabled? Artifact of local development?

loveucifer:
hey @ahoppen can i get another review :)

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,88 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+          delete.uri = self.uriAdjustedForCopiedFiles(delete.uri)
+          newDocumentChanges.append(.deleteFile(delete))
+        }
+      }
+      edit.documentChanges = newDocumentChanges
+    }
+    return edit
+  }
+
+  package func locationsOrLocationLinksAdjustedForCopiedFiles(_ response: LocationsOrLocationLinksResponse?) -> LocationsOrLocationLinksResponse? {
```
ahoppen: Could your run `swift-format` on your changes? https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2256,6 +2276,36 @@ extension SourceKitLSPServer {
     )
   }
 
+  private func callHierarchyItemAdjustedForCopiedFiles(
```
ahoppen: Shouldn’t this be a method on `BuildServerManager` like the other methods that adjust for copied files?

loveucifer:
> Thanks, I left two more comments inline but I think a few of my last comments are still outstanding (see the ones that I didn’t mark as resolved). In particular, I don’t think that `SourceKitLSPServer.swift` should contain any large-scale edits for this change, it should just call to adjust methods in `BuildServerManager` to take copied files into account.

got it thanks !

Sources/SourceKitLSP/SourceKitLSPServer.swift
loveucifer: hey i believe `workspace` is used further down in the function for `workspace.buildServerManager.locationsAdjustedForCopiedFiles()` and `workspace.buildServerManager.locationAdjustedForCopiedFiles()` to remap the locations. but i mayy be missing something - could you clarify what you had in mind , sorry if this is extra work on your end

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1689,23 +1689,24 @@ extension SourceKitLSPServer {
...
+      let locations = symbolOccurrences.map { symbolOccurrence -> Location in
+        let symbolPosition = Position(
+          line: symbolOccurrence.location.line - 1,  // 1-based -> 0-based
+          // Technically we would need to convert the UTF-8 column to a UTF-16 column. This would require reading the
+          // file. In practice they almost always coincide, so we accept the incorrectness here to avoid the file read.
+          utf16index: symbolOccurrence.location.utf8Column - 1
+        )
+        return Location(uri: symbolOccurrence.location.documentUri, range: Range(symbolPosition))
+      }
+      let remappedLocations = await workspace.buildServerManager.locationsAdjustedForCopiedFiles(locations)
```
ahoppen: Do we need to store the mapped location here? Wouldn’t it be easier if we just call `await workspace.buildServerManager.locationsAdjustedForCopiedFiles` inside `symbolsAndIndex.sorted(by: { $0.symbol < $1.symbol }).map` where we currently have `let symbolLocation = Location`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2199,10 +2201,27 @@ extension SourceKitLSPServer {
...
+        // check if there are declarations that don't have a definition at the same location
+        let hasDeclarationOnlyLocations = declarations.contains { decl in
+          !definitions.contains { def in
+            def.location.path == decl.location.path && def.location.line == decl.location.line
+          }
+        }
+        if hasDeclarationOnlyLocations {
+          occurrences = definitions
+        }
+      }
```
ahoppen: Is this change unrelated to the copied file mapping? If so, could you open a separate PR for it?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2275,12 +2295,18 @@ extension SourceKitLSPServer {
 
     // Only return a single call hierarchy item. Returning multiple doesn't make sense because they will all have the
     // same USR (because we query them by USR) and will thus expand to the exact same call hierarchy.
-    let callHierarchyItems = usrs.compactMap { (usr) -> CallHierarchyItem? in
+    var callHierarchyItems: [CallHierarchyItem] = []
+    for usr in usrs {
```
ahoppen: Is there any particular reason why you switched from the `compactMap` based implementation to `for … in`? I’d prefer if we just sticked with the old `compactMap`-based approach. It it was to allow `await` calls in the body, we have the `asyncCompactMap` helper for that.

Same for the other locations where you switched from `compactMap` to `for … in`

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I see. Missed that. Thanks for the explanation.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2439,7 +2474,7 @@ extension SourceKitLSPServer {
         let basename = (try? AbsolutePath(validating: url.filePath))?.basename
       {
         detail = "Extension at \(basename):\(location.range.lowerBound.line + 1)"
-      } else if let moduleName = moduleName {
+      } else if let moduleName = moduleName, !moduleName.isEmpty {
```
ahoppen: Is this change unrelated to the PR? Could you open a separate PR for it, so we can give this change its own dedicated discussion?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2568,91 +2611,60 @@ extension SourceKitLSPServer {
...
-      definition: SymbolOccurrence,
-      moduleName: String?,
-      index: CheckedIndex
-    ) -> TypeHierarchyItem? {
-      return self.indexToLSPTypeHierarchyItem(
-        definition: definition,
-        moduleName: moduleName,
-        index: index
-      )
-    }
```
ahoppen, line 2587: It seems that these workarounds might indeed no longer needed when using Swift 6.3 snapshots but the issue does still persist in Swift 6.2. Since SourceKit-LSP still needs to compile using the released Swift 6.2 compiler, we still need them. 

Even if the changes were no longer needed, I’d prefer to remove the workarounds in a separate PR.

Sources/SourceKitLSP/SourceKitLSPServer.swift
loveucifer: I think this is needed because the `testFindImplementationInCopiedHeader` - without it the test fails because C functions need this fallback to find the definition from a declaration , the tests kept failing without it i think the error was something like "Go to Implementation" on a C function returning empty because C functions don't have the inheritance relationships i might be wrong but now that i tried without using this the tests failed again : /

Sources/SourceKitLSP/SourceKitLSPServer.swift
loveucifer: The original change to `for...in` was because we needed `await` for the location remappingand stuff . I didn't quite think about `asyncCompactMap` at the time idk might've missed it my bad. I've now switched them all back to use `asyncCompactMap` changing all of them soon :D

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1685,26 +1684,22 @@ extension SourceKitLSPServer {
...
       }
       try Task.checkCancellation()
-      symbolsAndIndex += symbolOccurrences.map {
-        return ($0, index)
-      }
     }
-    return symbolsAndIndex.sorted(by: { $0.symbol < $1.symbol }).map { symbolOccurrence, index in
+
+    var result: [WorkspaceSymbolItem] = []
+    for (symbolOccurrence, index, workspace) in symbolsIndexAndWorkspaces.sorted(by: { $0.symbol < $1.symbol }) {
```
ahoppen: Couldn’t this also just continue to be an `asyncMap`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2588,36 +2641,32 @@ extension SourceKitLSPServer {
...
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: occurrence.symbol.usr) else {
         return nil
       }
 
-      return indexToLSPTypeHierarchyItem2(
-        definition: definition,
-        moduleName: definition.location.moduleName,
-        index: index
-      )
+      let moduleName = definition.location.moduleName.isEmpty ? nil : definition.location.moduleName
```
ahoppen: Why is the change for `moduleName.isEmpty` necessary here? I agree that it might be good to add but if so, it should be part of a separate PR. Just in case issues arise, it makes bisecting easier if each PR only addresses one issue. Same for the other checks for empty modules.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Oh, I see (please correct me if I’m wrong): Find implementation has a bug already, that’s not related to to copied files, and you just noticed this now when writing the test case. 

What I’d prefer is that we remove the test case and this check in this PR and have a separate follow-up PR or an issue to fix the find implementation bug, just to keep things focussed. Would that work for you?

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -962,6 +962,123 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+      var newChanges: [DocumentURI: [TextEdit]] = [:]
+      for (uri, edits) in changes {
+        let newUri = self.uriAdjustedForCopiedFiles(uri)
+        newChanges[newUri, default: []] += edits
+      }
+      edit.changes = newChanges
+    }
+    if let documentChanges = edit.documentChanges {
+      var newDocumentChanges: [WorkspaceEditDocumentChange] = []
+      for change in documentChanges {
```
ahoppen: Nitpick: Could this use `map` or `reserverCapacity` since we already know how many elements we’ll need in `newDocumentChanges`? Same for `newChanges` and in `locationsOrLocationLinksAdjustedForCopiedFiles`.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

loveucifer:
> Everything looks great to me. Thanks for addressing this issue 😍

Thanks! This was my first contribution to Swift, and I loved the whole PR process

rintaro:
(I'm asking around about the compiler crashes)

rintaro:
I preserved the current state in https://github.com/rintaro/sourcekit-lsp/tree/PR-2385-crasher so compiler folks can look into it.
Feel free to modify the PR and try workaround if there's any.

ahoppen:
I filed https://github.com/swiftlang/swift/issues/86170 for the compiler crash. A workaround seems to be to refer to `for … in` instead of `asyncCompactMap` in `prepareTypeHierarchy`, eg. using the following diff. @loveucifer could you apply that and I’ll trigger CI again?

```diff
diff --git a/Sources/SourceKitLSP/SourceKitLSPServer.swift b/Sources/SourceKitLSP/SourceKitLSPServer.swift
index 80055d84..c92b7c68 100644
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2546,7 +2546,8 @@ extension SourceKitLSPServer {
       )
     }
 
-    let typeHierarchyItems = await usrs.asyncCompactMap { (usr) -> TypeHierarchyItem? in
+    var typeHierarchyItems: [TypeHierarchyItem] = []
+    for usr in usrs {
       guard let info = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: usr) else {
         return nil
       }
@@ -2570,8 +2571,9 @@ extension SourceKitLSPServer {
       guard let item = indexToLSPTypeHierarchyItem2(definition: info, moduleName: moduleName, index: index) else {
         return nil
       }
-      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
-    }.sorted(by: { $0.name < $1.name })
+      typeHierarchyItems.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
+    }
+    typeHierarchyItems.sort(by: { $0.name < $1.name })
 
     if typeHierarchyItems.isEmpty {
       // When returning an empty array, VS Code fails with the following two errors. Returning `nil` works around those
```

loveucifer:
> I filed [swiftlang/swift#86170](https://github.com/swiftlang/swift/issues/86170) for the compiler crash. A workaround seems to be to refer to `for … in` instead of `asyncCompactMap` in `prepareTypeHierarchy`, eg. using the following diff. @loveucifer could you apply that and I’ll trigger CI again?
> 
> ```diff
> diff --git a/Sources/SourceKitLSP/SourceKitLSPServer.swift b/Sources/SourceKitLSP/SourceKitLSPServer.swift
> index 80055d84..c92b7c68 100644
> --- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
> +++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
> @@ -2546,7 +2546,8 @@ extension SourceKitLSPServer {
>        )
>      }
>  
> -    let typeHierarchyItems = await usrs.asyncCompactMap { (usr) -> TypeHierarchyItem? in
> +    var typeHierarchyItems: [TypeHierarchyItem] = []
> +    for usr in usrs {
>        guard let info = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: usr) else {
>          return nil
>        }
> @@ -2570,8 +2571,9 @@ extension SourceKitLSPServer {
>        guard let item = indexToLSPTypeHierarchyItem2(definition: info, moduleName: moduleName, index: index) else {
>          return nil
>        }
> -      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
> -    }.sorted(by: { $0.name < $1.name })
> +      typeHierarchyItems.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
> +    }
> +    typeHierarchyItems.sort(by: { $0.name < $1.name })
>  
>      if typeHierarchyItems.isEmpty {
>        // When returning an empty array, VS Code fails with the following two errors. Returning `nil` works around those
> ```

got it doing it rn

loveucifer:
hey @ahoppen sorry if this is the wrong place to ask is there a server or communication channel for this repo

ahoppen:
@swift-ci Please test

ahoppen:
> sorry if this is the wrong place to ask is there a server or communication channel for this repo

For anything that’s related to a specific issue or a code change, GitHub is really the best communication channel. For anything more broad, you can also post in the [SourceKit-LSP category on the Swift Forums](https://forums.swift.org/c/development/sourcekit-lsp/47).

loveucifer:
seems like its still failing :(

ahoppen:
Looks like we need to apply the same trick in a couple new places. It builds with the following diff applied using `swift-DEVELOPMENT-SNAPSHOT-2025-11-26-a`. Could you apply it to your PR. And sorry for this churn.

<details><summary>Diff</summary>

```diff
diff --git a/Sources/SourceKitLSP/SourceKitLSPServer.swift b/Sources/SourceKitLSP/SourceKitLSPServer.swift
index dbf4896f..3542baf1 100644
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2281,15 +2281,17 @@ extension SourceKitLSPServer {
 
     // Only return a single call hierarchy item. Returning multiple doesn't make sense because they will all have the
     // same USR (because we query them by USR) and will thus expand to the exact same call hierarchy.
-    let callHierarchyItems = await usrs.asyncCompactMap { (usr) -> CallHierarchyItem? in
+    var callHierarchyItems: [CallHierarchyItem] = []
+    for usr in usrs {
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: usr) else {
-        return nil
+        continue
       }
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
-    }.sorted(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
+      callHierarchyItems.append(await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item))
+    }
+    callHierarchyItems.sort(by: { Location(uri: $0.uri, range: $0.range) < Location(uri: $1.uri, range: $1.range) })
 
     // Ideally, we should show multiple symbols. But VS Code fails to display call hierarchies with multiple root items,
     // failing with `Cannot read properties of undefined (reading 'map')`. Pick the first one.
@@ -2358,24 +2360,25 @@ extension SourceKitLSPServer {
       return self.indexToLSPCallHierarchyItem(definition: definition, index: index)
     }
 
-    let calls = await callersToCalls.asyncCompactMap { (caller, callsList) -> CallHierarchyIncomingCall? in
+    var calls: [CallHierarchyIncomingCall] = []
+    for (caller, callsList) in callersToCalls {
       // Resolve the caller's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: caller.usr) else {
-        return nil
+        continue
       }
 
       let locations = callsList.compactMap { indexToLSPLocation2($0.location) }.sorted()
       let remappedLocations = await workspace.buildServerManager.locationsAdjustedForCopiedFiles(locations)
       guard !remappedLocations.isEmpty else {
-        return nil
+        continue
       }
 
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
       let remappedItem = await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
 
-      return CallHierarchyIncomingCall(from: remappedItem, fromRanges: remappedLocations.map(\.range))
+      calls.append(CallHierarchyIncomingCall(from: remappedItem, fromRanges: remappedLocations.map(\.range)))
     }
     return calls.sorted(by: { $0.from.name < $1.from.name })
   }
@@ -2404,26 +2407,27 @@ extension SourceKitLSPServer {
     let callableUsrs = [data.usr] + index.occurrences(relatedToUSR: data.usr, roles: .accessorOf).map(\.symbol.usr)
     let callOccurrences = callableUsrs.flatMap { index.occurrences(relatedToUSR: $0, roles: .containedBy) }
       .filter(\.shouldShowInCallHierarchy)
-    let calls = await callOccurrences.asyncCompactMap { occurrence -> CallHierarchyOutgoingCall? in
+    var calls: [CallHierarchyOutgoingCall] = []
+    for occurrence in callOccurrences {
       guard occurrence.symbol.kind.isCallable else {
-        return nil
+        continue
       }
       guard let location = indexToLSPLocation2(occurrence.location) else {
-        return nil
+        continue
       }
       let remappedLocation = await workspace.buildServerManager.locationAdjustedForCopiedFiles(location)
 
       // Resolve the callee's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: occurrence.symbol.usr) else {
-        return nil
+        continue
       }
 
       guard let item = indexToLSPCallHierarchyItem2(definition: definition, index: index) else {
-        return nil
+        continue
       }
       let remappedItem = await workspace.buildServerManager.callHierarchyItemAdjustedForCopiedFiles(item)
 
-      return CallHierarchyOutgoingCall(to: remappedItem, fromRanges: [remappedLocation.range])
+      calls.append(CallHierarchyOutgoingCall(to: remappedItem, fromRanges: [remappedLocation.range]))
     }
     return calls.sorted(by: { $0.to.name < $1.to.name })
   }
@@ -2624,17 +2628,18 @@ extension SourceKitLSPServer {
 
     // Convert occurrences to type hierarchy items
     let occurs = baseOccurs + retroactiveConformanceOccurs
-    let types = await occurs.asyncCompactMap { occurrence -> TypeHierarchyItem? in
+    var types: [TypeHierarchyItem] = []
+    for occurrence in occurs {
       // Resolve the supertype's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: occurrence.symbol.usr) else {
-        return nil
+        continue
       }
 
       let moduleName = definition.location.moduleName
       guard let item = indexToLSPTypeHierarchyItem2(definition: definition, moduleName: moduleName, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
+      types.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
     }
     return types.sorted(by: { $0.name < $1.name })
   }
@@ -2664,7 +2669,8 @@ extension SourceKitLSPServer {
     }
 
     // Convert occurrences to type hierarchy items
-    let types = await occurs.asyncCompactMap { occurrence -> TypeHierarchyItem? in
+    var types: [TypeHierarchyItem] = []
+    for occurrence in occurs {
       if occurrence.relations.count > 1 {
         // An occurrence with a `baseOf` or `extendedBy` relation is an occurrence inside an inheritance clause.
         // Such an occurrence can only be the source of a single type, namely the one that the inheritance clause belongs
@@ -2672,19 +2678,19 @@ extension SourceKitLSPServer {
         logger.fault("Expected at most extendedBy or baseOf relation but got \(occurrence.relations.count)")
       }
       guard let related = occurrence.relations.sorted().first else {
-        return nil
+        continue
       }
 
       // Resolve the subtype's definition to find its location
       guard let definition = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: related.symbol.usr) else {
-        return nil
+        continue
       }
 
       let moduleName = definition.location.moduleName
       guard let item = indexToLSPTypeHierarchyItem2(definition: definition, moduleName: moduleName, index: index) else {
-        return nil
+        continue
       }
-      return await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item)
+      types.append(await workspace.buildServerManager.typeHierarchyItemAdjustedForCopiedFiles(item))
     }
     return types.sorted { $0.name < $1.name }
   }
```

</details>

loveucifer:
@ahoppen hope this fixes it :3

rintaro:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Some LSP test seems to be spinning even though we should have timeouts everywhere but unfortunately we can’t tell which one it is… Let’s re-try it and if that still has issues, the fun begins…

@swift-ci Please test Windows

loveucifer:
still failing it seems :(

rintaro:
Windows timeout was possibly https://github.com/swiftlang/swift-package-manager/pull/9535

rintaro:
@swift-ci Please test Windows

loveucifer:
okei worked yaay @ahoppen @rintaro

loveucifer:
@ahoppen is there any additional stuff that needs to be addressed ? noticed that its still not merged lol

ahoppen:
The branches were [locked](https://forums.swift.org/t/holiday-schedule-2025/83058) during the holidays and open again now. Merging now. Thanks a lot for implementing this, @loveucifer ❤️

PR #2387, opened 2025-12-10
[#2118] Implement keyword snippets for control-flow keywords (if/for/while/etc.)

### LINKED ISSUE #2118: Provide `if`, `for`, `while`, etc. as snippets instead of plain keywords
### Description

Other language servers, e.g. `clangd`, provide snippets for control structures such as `if`, `for`, `while`, etc. Accepting a suggestion for one of those keywords then will insert the following text:
```swift
if expression {
    |
}
```
The cursor will first be on `expression`, allowing the user to enter the expression that controls the if statement. Pressing the tab key will move the cursor to the next location, i.e. `|` - where the user can write the code that shall be executed when the expression is true.

DPrakashh:
This PR implements snippet completions for control-flow keywords (if, for, while, guard, switch, repeat) when they are used in code completion.

The solution involved updating "CodeCompletionSession.swift" to identify keyword completions (.keyword) and overwrite their "insertText" and "textEdit" with the correct LSP snippet strings, setting the insertTextFormat to .snippet.

This ensures that keywords now contain placeholders (e.g., `${1:condition}` and `$0`) required by LSP clients.

Verification
New unit tests were added to "SwiftCompletionSnippetTests.swift" to explicitly verify:
1. The "insertTextFormat" is correctly set to ".snippet".
2. The presence of the required snippet placeholders (e.g., `$1`, `$0`).

All local tests, including the focused test (`swift test --filter SwiftCompletionSnippetTests/testKeywordIfProvidesSnippet`), passed successfully.

Checklist (Based on Contributing Guide)
1-Code has been formatted using `swift format -ipr 
2-New tests were added to cover the feature.

Fixes #2118

### DIFF
--- a/Sources/SwiftLanguageService/CodeCompletionSession.swift
+++ b/Sources/SwiftLanguageService/CodeCompletionSession.swift
@@ -720,40 +720,48 @@ class CodeCompletionSession {
   private func keywordSnippet(for keyword: String) -> String? {
     guard clientSupportsSnippets else { return nil }
 
-    func indentationUnitString() -> String {
-      // Convert the inferred Trivia (spaces/tabs) to a string we can embed into snippets.
-      if let trivia = indentationWidth {
-        for piece in trivia {
-          switch piece {
-          case .spaces(let n):
-            return String(repeating: " ", count: n)
-          case .tabs(let n):
-            return String(repeating: "\t", count: n)
-          default:
-            continue
-          }
-        }
-      }
-      // Fallback to a single tab to preserve previous behaviour when indentation cannot be inferred.
-      return "\t"
-    }
-
-    let indent = indentationUnitString()
-    let doubleIndent = indent + indent
+    // Use the `description` of inferred `Trivia` (e.g. "    " or "\t").
+    // Fall back to four spaces to match `BasicFormat`.
+    let indent = indentationWidth?.description ?? "    "
 
     switch keyword {
     case "if":
-      return "if ${1:condition} {\n\(indent)${0:}\n}"
+      return """
+        if ${1:condition} {
+        \(indent)${0:body}
+        }
+        """
     case "for":
-      return "for ${1:item} in ${2:sequence} {\n\(indent)${0:}\n}"
+      return """
+        for ${1:item} in ${2:sequence} {
+        \(indent)${0:body}
+        }
+        """
     case "while":
-      return "while ${1:condition} {\n\(indent)${0:}\n}"
+      return """
+        while ${1:condition} {
+        \(indent)${0:body}
+        }
+        """
     case "guard":
-      return "guard ${1:condition} else {\n\(indent)${0:}\n}"
+      return """
+        guard ${1:condition} else {
+        \(indent)${0:body}
+        }
+        """
     case "switch":
-      return "switch ${1:value} {\n\(indent)case ${2:pattern}:\n\(doubleIndent)${0:}\n}"
+      return """
+        switch ${1:value} {
+        case ${2:pattern}:
+        \(indent)${0:body}
+        }
+        """
     case "repeat":
-      return "repeat {\n\(indent)${0:}\n} while ${1:condition}"
+      return """
+        repeat {
+        \(indent)${0:body}
+        } while ${1:condition}
+        """
     default:
       return nil
     }
--- a/Tests/SourceKitLSPTests/SwiftCompletetionSnippetTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftCompletetionSnippetTests.swift
@@ -51,7 +51,7 @@ final class SwiftCompletionSnippetTests: SourceKitLSPTestCase {
 
     let insertText = try XCTUnwrap(ifItem.insertText)
     XCTAssertTrue(insertText.contains("${1:condition}"))
-    XCTAssertTrue(insertText.contains("${0:}"))
+    XCTAssertTrue(insertText.contains("${0:body}"))
   }
 
   func testKeywordForProvidesSnippet() async throws {
@@ -163,6 +163,9 @@ final class SwiftCompletionSnippetTests: SourceKitLSPTestCase {
     let insertText = try XCTUnwrap(switchItem.insertText)
     XCTAssertTrue(insertText.contains("${1:value}"))
     XCTAssertTrue(insertText.contains("case"))
+    // Expect `case` to be not indented and the body placeholder to exist (indentation tested elsewhere).
+    XCTAssertTrue(insertText.contains("case ${2:pattern}:"))
+    XCTAssertTrue(insertText.contains("${0:body}"))
   }
 
   func testKeywordRepeatProvidesSnippet() async throws {
@@ -271,7 +274,10 @@ final class SwiftCompletionSnippetTests: SourceKitLSPTestCase {
     let insertText = try XCTUnwrap(ifItem.insertText)
 
     // Expect newline + two spaces before the final placeholder
-    XCTAssertTrue(insertText.contains("\n  ${0:}"), "expected two-space indentation in snippet, got: '\(insertText)'")
+    XCTAssertTrue(
+      insertText.contains("\n  ${0:body}"),
+      "expected two-space indentation in snippet, got: '\(insertText)'"
+    )
   }
 
   func testKeywordSnippetUsesInferredTabsIndentation() async throws {
@@ -297,6 +303,7 @@ final class SwiftCompletionSnippetTests: SourceKitLSPTestCase {
 
     let ifItem = try XCTUnwrap(completions.items.first(where: { $0.label == "if" }))
     let insertText = try XCTUnwrap(ifItem.insertText)
-    XCTAssertTrue(insertText.contains("\n\t"), "expected tab indentation in snippet, got: '\(insertText)'")
+    // Expect newline + tab then the body placeholder
+    XCTAssertTrue(insertText.contains("\n\t${0:body}"), "expected tab indentation in snippet, got: '\(insertText)'")
   }
 }


DPrakashh:
Hi @ahoppen , @bnbarham , @hamishknight , @rintaro 

Hope you're having a good week.

I wanted to check on the status of my PR from four days ago. I know these core completion changes take time to review carefully.
Just confirming that all local tests passed on my end. Would it be possible to get the CI tests triggered when a maintainer is able to look at it?

Thanks for your time!

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +719,29 @@ class CodeCompletionSession {
...
   }
+
+  /// Generate a snippet for control flow keywords like if, for, while, etc.
+  /// Returns the snippet text if the keyword is a control flow keyword and snippets are supported, otherwise nil.
+  private func keywordSnippet(for keyword: String) -> String? {
+    guard clientSupportsSnippets else { return nil }
+
+    switch keyword {
+    case "if":
+      return "if ${1:condition} {\n\t${0:}\n}"
```
ahoppen: We should not assume that the source file is indented with tabs. We have `BasicFormat.inferIndenatation` to infer a source files’s indentation and should rely and that instead.

Tests/SourceKitLSPTests/SwiftCompletetionSnippetTests.swift
```diff
@@ -0,0 +1,247 @@
+@_spi(SourceKitLSP) import LanguageServerProtocol
```
ahoppen, line 13: Could you add the copyright header to this file?

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -501,6 +501,21 @@ class CodeCompletionSession {
       let kind: sourcekitd_api_uid_t? = value[sourcekitd.keys.kind]
       let completionKind = kind?.asCompletionItemKind(sourcekitd.values) ?? .value
 
+      // Check if this is a keyword that should be converted to a snippet
+      var isKeywordSnippet = false
+      if completionKind == .keyword, let snippetText = keywordSnippet(for: name) {
```
ahoppen, line 499: If this evaluates to `true`, we can avoid computing the previous `textEdit`. Since almost all parameters are the same, it should even be sufficient to just use `snippetText` for `text` and just have a single call to `computeCompletionTextEdit`.

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -577,8 +592,8 @@ class CodeCompletionSession {
         deprecated: notRecommended,
         sortText: sortText,
         filterText: filterName,
-        insertText: text,
-        insertTextFormat: isInsertTextSnippet ? .snippet : .plain,
+        insertText: insertText,
```
ahoppen, line 590: I don’t think this is correct. `text` will have SourceKit placeholders replaced by LSP placeholders while `insertText` still contains SourceKit placeholders. I expect that we have tests that fail with this change. Did you run all tests locally?

Tests/SourceKitLSPTests/SwiftCompletetionSnippetTests.swift
```diff
@@ -0,0 +1,247 @@
...
+    )
+
+    let completions = try await testClient.send(
+      CompletionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["1️⃣"])
+    )
+
+    guard let ifItem = completions.items.first(where: { $0.label == "if" }) else {
+      XCTFail("No completion item with label 'if'")
+      return
+    }
```
ahoppen:
```suggestion
    let ifItem = try XCTUnwrap(completions.items.first(where: { $0.label == "if" }))
```

Same in a couple more places below

ahoppen (commented):
Thank you. The overall structure looks good, I have left a few comments inline. 

Also, do you have a screen recording or something like that how this looks like in VS Code? Ie. point VS Code to a locally built version of SourceKit-LSP as described in https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#using-a-locally-built-sourcekit-lsp-in-an-editor.

DPrakashh:
@ahoppen on it sir

DPrakashh:
@ahoppen 

**I have updated the PR to address all previous feedback. Here is a summary of the changes:**

1-Dynamic Indentation: Implemented indentationUnitString() to use inferred indentation (spaces or tabs) instead of hardcoded tab characters.
2-Refactored Completion Logic: Simplified CodeCompletionSession to calculate textEdit once, using either the standard text or the keyword snippet as appropriate.

**Test Improvements:**
1-Added the required copyright header to SwiftCompletionSnippetTests.swift.
2-Updated test assertions to use try XCTUnwrap for better error handling and readability.
3-Added new test cases to verify snippet behavior with both inferred spaces and inferred tabs.

**Verification: The attached video demonstrates:**
1-Running swift test --filter SwiftCompletionSnippetTests with all tests passing.
2-The if, while, and for snippets expanding correctly in VS Code.
3-The snippets dynamically adjusting their indentation when switching the editor from Spaces to Tabs.

**Thank you for the guidance on this! As this is my first contribution to a project of this scale, I appreciate your patience and would be happy to make further adjustments if needed.**

https://github.com/user-attachments/assets/fddfeaed-9dbb-4e4e-8936-7eb16dbaba4b

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +714,50 @@ class CodeCompletionSession {
...
   }
+
+  /// Generate a snippet for control flow keywords like if, for, while, etc.
+  /// Returns the snippet text if the keyword is a control flow keyword and snippets are supported, otherwise nil.
+  private func keywordSnippet(for keyword: String) -> String? {
+    guard clientSupportsSnippets else { return nil }
+
+    func indentationUnitString() -> String {
+      // Convert the inferred Trivia (spaces/tabs) to a string we can embed into snippets.
+      if let trivia = indentationWidth {
```
ahoppen: Nice, I forgot that we already have the inferred indentation in `CodeCompletionSession` 🤩

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +714,50 @@ class CodeCompletionSession {
 
     return Position(line: completionPos.line, utf16index: deletionStartUtf16Offset)
   }
+
+  /// Generate a snippet for control flow keywords like if, for, while, etc.
+  /// Returns the snippet text if the keyword is a control flow keyword and snippets are supported, otherwise nil.
+  private func keywordSnippet(for keyword: String) -> String? {
+    guard clientSupportsSnippets else { return nil }
+
+    func indentationUnitString() -> String {
```
ahoppen: I don’t think this function is necessary. You can just call `indentationWidth.description`.

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +714,50 @@ class CodeCompletionSession {
...
+            return String(repeating: " ", count: n)
+          case .tabs(let n):
+            return String(repeating: "\t", count: n)
+          default:
+            continue
+          }
+        }
+      }
+      // Fallback to a single tab to preserve previous behaviour when indentation cannot be inferred.
+      return "\t"
```
ahoppen: I think we should fall back to four spaces to match [BasicFormat](https://github.com/swiftlang/swift-syntax/blob/606e0e7d06ba371a3f8fabe3ec40ea4f6d8687ea/Sources/SwiftBasicFormat/BasicFormat.swift#L71).

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +714,50 @@ class CodeCompletionSession {
...
+    }
+
+    let indent = indentationUnitString()
+    let doubleIndent = indent + indent
+
+    switch keyword {
+    case "if":
+      return "if ${1:condition} {\n\(indent)${0:}\n}"
```
ahoppen: Just something that just crossed my mind. Would this read nicer as a multiline string literal? Same for the others.


```suggestion
      return """
        if ${1:condition} {
        \(indent)${0:}
        }
        """
```

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -704,6 +714,50 @@ class CodeCompletionSession {
...
+    case "for":
+      return "for ${1:item} in ${2:sequence} {\n\(indent)${0:}\n}"
+    case "while":
+      return "while ${1:condition} {\n\(indent)${0:}\n}"
+    case "guard":
+      return "guard ${1:condition} else {\n\(indent)${0:}\n}"
+    case "switch":
+      return "switch ${1:value} {\n\(indent)case ${2:pattern}:\n\(doubleIndent)${0:}\n}"
```
ahoppen: Let’s not indent the `case` label to match the default behavior of [swift-format](https://github.com/swiftlang/swift-format/blob/main/Documentation/Configuration.md#indentswitchcaselabels)

Ie. make this

```suggestion
      return """
        switch ${1:value} {
        case ${2:pattern}:
        \(indent)${0:}
        }
        """
```

Sources/SwiftLanguageService/CodeCompletionSession.swift
ahoppen: Let’s call the 0 placeholder `body`, ie. `${0:body}`.

ahoppen: Did you have any particular reason to make the body the first tab stop that the cursor will jump to first? Because I would usually expect to write the condition before the body and thus be annoyed if my cursor jumped to the body first. 

FWIW, I have the documentation of how snippets work from [the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/), which contains the following

```ts
		/**
		 * Client supports snippets as insert text.
		 *
		 * A snippet can define tab stops and placeholders with `$1`, `$2`
		 * and `${3:foo}`. `$0` defines the final tab stop, it defaults to
		 * the end of the snippet. Placeholders with equal identifiers are
		 * linked, that is typing in one will update others too.
		 */
```

DPrakashh: Thats a great point, Alex. My initial thought was to get the user into the 'meat' of the code (the body) quickly, but I realize now that logically, you always define the 'what' (condition) before the 'how' (body). Jumping to the body first actually breaks the natural flow of writing a statement. I have updated the tab stops to $1 for the condition and $0 for the body to match that expected workflow.

DPrakashh:
@ahoppen 

I have finalised the refinements based on your feedback. I am attaching a video demo (recorded with Screencast Mode) which demonstrates the comparison between the default VS Code snippets and the updated SourceKit-LSP snippets.

**Changes implemented based on your suggestions:**
1-Logical Tab Order: Reordered the tab stops so the cursor lands on the condition first ($1) and moves to the body last ($0). This ensures the user fills out the logic before jumping into the braces.
2-Switch Statement Alignment: Corrected the switch snippet so that case labels align vertically with the switch keyword, fixing the previous double-indentation issue.
3-Code Cleanup: Removed the unused doubleIndent variable and the legacy indentation helper functions, opting for a cleaner implementation using indentationWidth.
4-Refined Test Suite: Updated CodeCompletionSnippetTests to be more resilient, verifying the presence of specific keywords and placeholders rather than exact indentation strings.

**Usability Improvement:**
1-**Descriptive "Body" Placeholder:** I added a small improvement for the user by changing the final stop from an empty ${0:} to a descriptive ${0:body}. This provides a clear visual hint of where the code should go while remaining the final exit point for the snippet.

I have verified these changes using a locally built SourceKit-LSP server in VS Code as shown in the video. Ready for your final review!

**Note**: I’ve done my best to address all the points from the review. If there are any further refinements or "nits" you'd like me to fix to get this fully aligned with the project's standards, please let me know—I’m happy to make additional changes!
https://github.com/user-attachments/assets/f14026c9-c66b-4c39-b836-8aa5dcbdd270

ahoppen (approved):
Looks great to me. Thank you @DPrakashh ❤️

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

DPrakashh:
Thank you so much for the guidance and the review, @ahoppen  I am  glad the refinements and the added placeholders look good. I will keep an eye on the CI tests for macOS and Windows. Excited to see this merged!!!

DPrakashh:
Hi, @ahoppen I noticed that the Windows CI is failing with 'No test results found' across multiple PRs today. Since Linux and macOS have already passed for my changes, I’ll wait for the infrastructure to stabilize. Let me know if you’d like me to try anything else once the Windows workers are back up!

ahoppen:
@swift-ci Please test Windows

ahoppen:
Congrats to your first contribution to Swift, @DPrakashh 🎉

DPrakashh:
Thank you so much, @ahoppen . I am beyond happy to have my first contribution merged, and I am already looking forward to solving the next set of challenges in SourceKit-LSP! 🚀

PR #2390, opened 2025-12-12
OpenBSD requires casting id_t for setpriority.

3405691582:
Unlike FreeBSD.

(As a heads up, I will create a pr for the release branch after this.)


3405691582:
@swift-ci please test.

PR #2394, opened 2025-12-18
Implement literal detection for jump-to-definition functionality

### LINKED ISSUE #2368: Do not offer jump-to-definition for literals
https://github.com/swiftlang/swift/pull/67613 added cursor info support to be able to inspect the type of a literal using Hover. This also implicitly added jump-to-definition support for those literals, which probably isn’t useful in most cases and causes issues. We should probably suppress jump-to-definition for literals either by:
- Looking at the syntax tree of Swift files and not returning jump-to-definition if the requested position is a literal
- Adding a field to the sourcekitd response for literals that can be picked up by SourceKit-LSP to suppress jump-to-definition for it.

Clemo97:
Fixes #2368

This PR suppresses jump-to-definition for literal values (strings, integers, floats, booleans, arrays, dictionaries, and nil).

**Implementation:**
- Added `isPositionOnLiteral()` method to `LanguageService` protocol with default implementation
- Implemented literal detection in `SwiftLanguageService` using SwiftSyntax tree traversal
- Added early return check in `indexBasedDefinition()` to skip literals

**Testing:**
- Added `testDefinitionOnLiteralsShouldReturnEmpty()` covering string, integer, boolean, and array literals

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2044,11 +2044,6 @@ extension SourceKitLSPServer {
     workspace: Workspace,
     languageService: any LanguageService
   ) async throws -> [Location] {
-    // Check if the position is on a literal value (string, integer, etc.)
-    if await languageService.isPositionOnLiteral(req.position, in: req.textDocument.uri) {
-      return []
-    }
-
     let symbols = try await languageService.symbolInfo(
       SymbolInfoRequest(
         textDocument: req.textDocument,
--- a/Sources/SwiftLanguageService/SwiftLanguageService.swift
+++ b/Sources/SwiftLanguageService/SwiftLanguageService.swift
@@ -1297,12 +1297,25 @@ extension sourcekitd_api_uid_t {
 // MARK: - Literal Detection
 
 extension SwiftLanguageService {
-  /// Check if the given position in the document is on a literal expression.
+  /// Determines whether the given document position is on (or immediately after) a literal value token.
   ///
-  /// Literal expressions (string, integer, boolean, float, array, dictionary) should not support
-  /// jump-to-definition since they don't have a "definition" location - they are values typed
-  /// directly in the code.
+  /// This check is used to disable jump-to-definition on values that do not have a meaningful definition location
+  /// in source code. By using `tokenKind` to check for literal tokens directly, this avoids incorrectly blocking
+  /// jump-to-definition for identifiers inside literals (e.g., variables in string interpolation like `"\(foo)"`
+  /// or array literals like `[myVar]`).
   ///
+  /// The function returns `true` when the position resolves to:
+  /// - A literal token at the exact position, or
+  /// - A literal token immediately preceding the position (to handle cursor placements just after the literal).
+  ///
+  /// Supported literal kinds:
+  /// - String segments (`.stringSegment`)
+  /// - Integer literals (`.integerLiteral`)
+  /// - Floating-point literals (`.floatLiteral`)
+  /// - Boolean keywords (`.keyword(.true)`, `.keyword(.false)`)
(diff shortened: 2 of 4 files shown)


Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1293,3 +1293,63 @@ extension sourcekitd_api_uid_t {
...
+          || node.is(BooleanLiteralExprSyntax.self)
+          || node.is(ArrayExprSyntax.self)
+          || node.is(DictionaryExprSyntax.self)
+          || node.is(NilLiteralExprSyntax.self)
+        {
+          return true
+        }
+        currentNode = node.parent
+      }
+      return false
```
ahoppen: This will also prevent all jump-to-definition functionality for code inside literals, eg. jump-to-definition will no longer work inside literals, eg. you couldn’t jump to the definition of `foo` anymore in `"Value: \(foo)"` or `[foo]`. Instead, what I think we should do, is to check if the `tokenKind` is `stringSegment`, `integerLiteral` or `floatLiteral`. Maybe you can find other token kinds as well that we should suppress as well, I didn’t walk through all of them.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1293,3 +1293,63 @@ extension sourcekitd_api_uid_t {
...
+    }
+
+    // Also check the token before the cursor position, as users might click right after a literal
+    if absolutePosition.utf8Offset > 0,
+      let tokenBefore = tree.token(at: AbsolutePosition(utf8Offset: absolutePosition.utf8Offset - 1))
+    {
+      if isTokenInLiteral(tokenBefore) {
+        return true
+      }
+    }
```
ahoppen, line 1354: Hmm, that’s quite unfortunate. I would prefer if we could find this previous token using `token.previousToken(viewMode: .sourceAccurate)` because it will not need to do a tree traversal to find the token again and should thus be more efficient. It likely won’t matter much in practice but still.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2044,6 +2044,11 @@ extension SourceKitLSPServer {
     workspace: Workspace,
     languageService: any LanguageService
   ) async throws -> [Location] {
+    // Check if the position is on a literal value (string, integer, etc.)
+    if await languageService.isPositionOnLiteral(req.position, in: req.textDocument.uri) {
+      return []
+    }
```
ahoppen: I think we should do this check in `SwiftLanguageService.symbolInfo`. I think we want to suppress the jump-to-definition kind of behavior as well for all the callers of `symbolInfo` (implementation, definition, call hierarchy, type hierarchy, find references).

Clemo97:
Thank you for the feedback @ahoppen.
Let me make the necessary changes 👍.

ahoppen:
I just realized that we have two open PRs open for this from @a7maad-ayman, (https://github.com/swiftlang/sourcekit-lsp/pull/2404) and you, @Clemo97, (https://github.com/swiftlang/sourcekit-lsp/pull/2394). I posted a suggested consolidation of these two PRs at https://github.com/swiftlang/sourcekit-lsp/pull/2404#issuecomment-3706992499. I’m sorry that we got into this race but would what I posted there work for you, @Clemo97?

Clemo97:
> I just realized that we have two open PRs open for this from @a7maad-ayman, (#2404) and you, @Clemo97, (#2394). I posted a suggested consolidation of these two PRs at [#2404 (comment)](https://github.com/swiftlang/sourcekit-lsp/pull/2404#issuecomment-3706992499). I’m sorry that we got into this race but would what I posted there work for you, @Clemo97?

Yeah, sounds good to me 👍🏿

PR #2399, opened 2025-12-21
Add Show Objective-C Selector code action

### LINKED ISSUE #2145: Add Action to Copy Objective-C Selector Name of @objc Swift Method
### Description

When a Swift class is marked @objcMembers or it has Objective-C/@objc ancestry and @objc methods without visible selector names, the Swift compiler applies a pretty complex set of heuristics to produce a readable Objective-C selector name. We should provide an action to run the compiler’s Objective-C selector mapping logic and copy the selector name for a Swift method to the pasteboard.

iMostfa:
Resolves https://github.com/swiftlang/sourcekit-lsp/issues/2145

Adds a *Show Objective-C Selector* code action for declarations exposed to Objective-C.

The action is only offered when the declaration under the cursor actually has a selector, and the selector is fetched while computing the code action, so invoking it needs no second request.

Requires https://github.com/swiftlang/swift/pull/86173.

https://github.com/user-attachments/assets/467cec3f-d0cb-4e64-91f1-adfc7879d33f

### DIFF
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -478,7 +478,7 @@ package actor SkipUnless {
         )
 
         let testURL = scratchDirectory.appending(component: "test.swift")
-        try source.write(to: testURL, atomically: false, encoding: .utf8)
+        try await source.writeWithRetry(to: testURL)
 
         let sourceFile = try testURL.filePath
 
--- a/Sources/SKTestSupport/SkipUnless.swift
+++ b/Sources/SKTestSupport/SkipUnless.swift
@@ -459,6 +459,47 @@ package actor SkipUnless {
     }
   }
 
+  package static func sourcekitdSupportsObjCSelector(
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws {
+    return try await shared.skipUnlessSupportedByToolchain(swiftVersion: SwiftVersion(6, 5), file: file, line: line) {
+      let sourcekitd = try await getSourceKitD()
+
+      return try await withTestScratchDir { scratchDirectory in
+        let (positions, source) = extractMarkers(
+          """
+          import Foundation
+
+          class Foo: NSObject {
+            @objc func 1️⃣bar() {}
+          }
+          """
+        )
+
+        let testURL = scratchDirectory.appending(component: "test.swift")
+        try source.write(to: testURL, atomically: false, encoding: .utf8)
+
+        let sourceFile = try testURL.filePath
+
+        let skreq = sourcekitd.dictionary([
+          sourcekitd.keys.offset: positions["1️⃣"]!.utf8Offset,
+          sourcekitd.keys.sourceFile: sourceFile,
+          sourcekitd.keys.compilerArgs: [sourceFile],
+        ])
+
+        do {
+          let response = try await sourcekitd.send(\.objcSelector, skreq)
+
+          let selector: String? = response[sourcekitd.keys.text]
+          return selector != nil
+        } catch {
+          return false
+        }
+      }
+    }
+  }
+
   package static func swiftPMBuildServerSupportedWithoutBackgroundIndexing(
     file: StaticString = #filePath,
     line: UInt = #line
--- a/Sources/SwiftLanguageService/ShowObjCSelector.swift
+++ b/Sources/SwiftLanguageService/ShowObjCSelector.swift
@@ -19,7 +19,7 @@ extension SwiftLanguageService {
   /// The Objective-C selector of the declaration at `position`, or `nil` if it isn't exposed to Objective-C or the
   /// toolchain doesn't know the request.
   func objcSelector(_ uri: DocumentURI, at position: Position) async -> String? {
-    return await orLog("Getting Objective-C selector", level: .debug) {
+    return await orLog("Getting Objective-C selector") {
       let snapshot = try await self.latestSnapshot(for: uri)
       let skreq = sourcekitd.dictionary([
         keys.offset: snapshot.utf8Offset(of: position),
--- a/Sources/SwiftLanguageService/ShowObjCSelector.swift
+++ b/Sources/SwiftLanguageService/ShowObjCSelector.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+import SourceKitD
+import SourceKitLSP
+
+extension SwiftLanguageService {
+  /// The Objective-C selector of the declaration at `position`, or `nil` if it isn't exposed to Objective-C or the
+  /// toolchain doesn't know the request.
+  func objcSelector(_ uri: DocumentURI, at position: Position) async -> String? {
+    return await orLog("Getting Objective-C selector", level: .debug) {
+      let snapshot = try await self.latestSnapshot(for: uri)
+      let skreq = sourcekitd.dictionary([
+        keys.offset: snapshot.utf8Offset(of: position),
+        keys.sourceFile: snapshot.uri.sourcekitdSourceFile,
+        keys.primaryFile: snapshot.uri.primaryFile?.pseudoPath,
+        keys.compilerArgs: await self.compileCommand(for: uri, fallbackAfterTimeout: true)?.compilerArgs
+          as [any SKDRequestValue]?,
+      ])
+      let dict = try await self.send(sourcekitdRequest: \.objcSelector, skreq, snapshot: snapshot)
+      return dict[keys.text] as String?
+    }
+  }
+
+  func showObjCSelector(_ command: ShowObjCSelectorCommand) -> LSPAny {
+    sourceKitLSPServer?.sendNotificationToClient(ShowMessageNotification(type: .info, message: command.selector))
+    return .string(command.selector)
+  }
+}
--- a/Sources/SwiftLanguageService/ShowObjCSelectorCommand.swift
+++ b/Sources/SwiftLanguageService/ShowObjCSelectorCommand.swift
@@ -1,0 +1,27 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+
+/// Command to show the Objective-C selector of a method that is exposed to Objective-C.
+package struct ShowObjCSelectorCommand: SwiftCommand {
+  package static let identifier: String = "show.objc.selector.command"
+
+  package var title = "Show Objective-C Selector"
+
+  /// The selector to show, computed when the code action is created.
+  package var selector: String
+
+  init(selector: String) {
+    self.selector = selector
+  }
+}
--- a/Sources/SwiftLanguageService/SwiftLanguageService.swift
+++ b/Sources/SwiftLanguageService/SwiftLanguageService.swift
@@ -959,7 +959,7 @@ extension SwiftLanguageService {
     }
     let providersAndKinds: [(provider: CodeActionProvider, kind: CodeActionKind?)] = [
       (retrieveSyntaxCodeActions, nil),
-      (retrieveRefactorCodeActions, .refactor),
+      (retrieveCursorInfoBasedCodeActions, nil),
       (retrieveQuickFixCodeActions, .quickFix),
       (retrieveRemoveUnusedImportsCodeAction, .sourceOrganizeImports),
     ]
@@ -1054,7 +1054,19 @@ extension SwiftLanguageService {
     )
   }
 
-  func retrieveRefactorCodeActions(_ params: CodeActionRequest) async throws -> [CodeAction] {
+  private func codeActionKind(_ kind: CodeActionKind, matches requested: CodeActionKind) -> Bool {
+    return kind == requested || kind.rawValue.hasPrefix("\(requested.rawValue).")
+  }
+
+  func retrieveCursorInfoBasedCodeActions(_ params: CodeActionRequest) async throws -> [CodeAction] {
+    if let requestedKinds = params.context.only,
+      !requestedKinds.contains(where: {
+        codeActionKind($0, matches: .refactor) || codeActionKind(.refactor, matches: $0)
+      })
+    {
+      return []
+    }
+
     let additionalCursorInfoParameters: ((SKDRequestDictionary) -> Void) = { skreq in
       skreq.set(self.keys.retrieveRefactorActions, to: 1)
     }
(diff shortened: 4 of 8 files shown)


Sources/SwiftLanguageService/CodeActions/SyntaxCodeActions.swift
```diff
@@ -15,21 +15,21 @@ import SwiftRefactor
 /// List of all of the syntactic code action providers, which can be used
 /// to produce code actions using only the swift-syntax tree of a file.
 let allSyntaxCodeActions: [any SyntaxCodeActionProvider.Type] = {
-  var result: [any SyntaxCodeActionProvider.Type] = [
```
ahoppen, line 18: Did you accidentally change indentation to 4 spaces here? Could you ensure that you run [swift-format](https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting) on your PR?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -978,9 +983,52 @@ extension SwiftLanguageService {
...
+    let keys = self.keys
+    let skreq = sourcekitd.dictionary([:])
+    skreq.set(keys.offset, to: offset)
+    skreq.set(keys.sourceFile, to: snapshot.uri.sourcekitdSourceFile)
+    if let primaryFile = snapshot.uri.primaryFile?.pseudoPath {
+      skreq.set(keys.primaryFile, to: primaryFile)
+    }
+    if let compilerArgs = compileCommand?.compilerArgs {
+      skreq.set(keys.compilerArgs, to: compilerArgs as [any SKDRequestValue])
+    }
```
ahoppen: Slightly more succinct.

```suggestion
    let skreq = sourcekitd.dictionary([
      keys.offset: offset,
      keys.sourceFile: snapshot.uri.sourcekitdSourceFile,
      keys.primaryFile: primaryFile,
      keys.compilerArgs: compilerArgs as [any SKDRequestValue]
    ])
```

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -978,9 +983,52 @@ extension SwiftLanguageService {
...
+      let dict = try await send(sourcekitdRequest: \.objcSelector, skreq, snapshot: snapshot)
+      if dict[keys.text] as String? != nil {
+        let copyCommand = CopyObjCSelectorCommand(
+          positionRange: range,
+          textDocument: textDocument
+        ).asCommand()
+        return CodeAction(title: "Copy Objective-C Selector", kind: .refactor, command: copyCommand)
+      }
+    } catch {
+      logger.debug("CopyObjCSelector: applicability check failed: \(error.localizedDescription)")
```
ahoppen:
```suggestion
      logger.debug("CopyObjCSelector applicability check failed: \(error.forLogging)")
```

This will make sure that the error correctly gets redacted based on the user’s [logging sensitivity settings](https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Enable%20Extended%20Logging.md).

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -978,9 +983,52 @@ extension SwiftLanguageService {
...
+      skreq.set(keys.primaryFile, to: primaryFile)
+    }
+    if let compilerArgs = compileCommand?.compilerArgs {
+      skreq.set(keys.compilerArgs, to: compilerArgs as [any SKDRequestValue])
+    }
+
+    do {
+      let dict = try await send(sourcekitdRequest: \.objcSelector, skreq, snapshot: snapshot)
+      if dict[keys.text] as String? != nil {
+        let copyCommand = CopyObjCSelectorCommand(
```
ahoppen: Am I missing something or where is `CopyObjCSelectorCommand` defined?

ahoppen (commented):
Thanks for opening the PR. One thing I’m not entirely convinced about is whether this should be a refactoring actions that modifies the current source file. I would probably have implemented this as a code action that shows a message with the Objective-C selector name to the user via the `window/showMessageRequest` request if the client supports that request. What do you think about that approach?

Sources/SwiftLanguageService/SwiftLanguageService.swift
iMostfa: my bad, still familiarizing myself with jujutsu, now you can find it.

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActions.swift
iMostfa: done, i need to use swift-format in my editor instead of SwiftFormtter 🙏🏻

iMostfa:
@ahoppen 
Thanks for the quick review !
yup, i think it's more convenient, and won't need to depend on `workspace/executeCommand` in the client. 
i will work on modifying.

iMostfa:
https://github.com/user-attachments/assets/652f2408-ada2-4b61-9ae9-ed7ee8967a18

@ahoppen I tried to use `window/showMessageRequest` but i found that we won't provide any actions, So instead I used `window/showMessage` which works by just showing a message

 [e37ec29](https://github.com/swiftlang/sourcekit-lsp/pull/2399/commits/e37ec29e3086035ff9f477b2443bb2e49e9db6da)

After giving it more thought, I think it’s more useful to add the Objective-C selector as a comment rather than only showing it in a dialog.

The most common use case for this action is likely copying the selector to use it in an Objective-C file. While it’s possible to move the mouse and copy it from the dialog, it’s more convenient to stay on the keyboard and simply navigate to and select the comment.

Do you think it makes sense to prefer adding a comment in this case?

ahoppen:
> After giving it more thought, I think it’s more useful to add the Objective-C selector as a comment rather than only showing it in a dialog.
> 
> The most common use case for this action is likely copying the selector to use it in an Objective-C file. While it’s possible to move the mouse and copy it from the dialog, it’s more convenient to stay on the keyboard and simply navigate to and select the comment.
> 
> Do you think it makes sense to prefer adding a comment in this case?

My thoughts are that the Objective-C selector name is a transient bit of information that you are very unlikely to want to keep around in your source file. While I agree that copying may be easier if the selector gets added as a comment, I see two downsides (both not big, but still): We give a small indication that it may be a good pattern to keep the Objective-C selector of these functions around in your source file, which I don’t want to give. Furthermore, you need to remember to remove the comment again after copying the selector name. And doing so may counteract any workflow advantages from putting the selector in the source file.

iMostfa:
fair points, so, if we decide to continue with just showMessage request, maybe we can rename the code action to be "Show Objective-c selector" instead of "Copy Objective-c selector" ?

ahoppen:
> can rename the code action to be "Show Objective-c selector" instead of "Copy Objective-c selector" ?

Absolutely!

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -962,13 +965,23 @@ extension SwiftLanguageService {
 
     var canInlineMacro = false
 
-    var refactorActions = cursorInfoResponse.refactorActions.compactMap {
-      let lspCommand = $0.asCommand()
+    var refactorActions: [CodeAction] = cursorInfoResponse.refactorActions.compactMap { action in
+      if action.actionString == "source.refactoring.kind.copy.objc.selector" {
```
ahoppen: Does this PR still match https://github.com/swiftlang/swift/pull/86173? I’m a little confused because as far as I can tell https://github.com/swiftlang/swift/pull/86173 only adds a sourcekitd request to get the Objective-C selector at a given position and I can’t find a `source.refactoring.kind.copy.objc.selector` refactoring action. Am I missing something?

And to be clear: I think getting the Objective-C selector without modeling it as a refactoring in sourcekitd is the right path.

iMostfa: Sorry for taking some time to check it, i checked both PRs, and currently this the content of each PR:

  #### Changes in swiftlang/swift#86173:
  - Fixed a bug in the request handler (inverted condition check)
  - Added `objc-selector` support to `sourcekitd-test` for testing
  - Moved test from `test/refactoring/` to `test/SourceKit/ObjCSelector/` since it's not a refactoring

  #### Changes here:
  - Removed dependency on the non-existent `source.refactoring.kind.copy.objc.selector` action
  - calling `source.request.objc.selector` directly and reads `key.text` from response

Sources/SwiftLanguageService/ShowObjCSelectorCommand.swift
```diff
@@ -0,0 +1,67 @@
...
+      case .dictionary(let rangeDict)? = dictionary[CodingKeys.positionRange.stringValue]
+    else {
+      return nil
+    }
+
+    guard let positionRange = Range<Position>(fromLSPDictionary: rangeDict),
+      let textDocument = TextDocumentIdentifier(fromLSPDictionary: documentDict)
+    else {
+      return nil
+    }
```
ahoppen: I think this can be simplified slightly by using `init(fromLSPAny:)`

```suggestion
    guard let textDocument = TextDocumentIdentifier(fromLSPAny: dictionary[CodingKeys.textDocument.stringValue]),
      case .string(let title)? = dictionary[CodingKeys.title.stringValue],
      let positionRange = Range<Position>(fromLSPAny: dictionary[CodingKeys.positionRange.stringValue])
    else {
      return nil
    }
```

Sources/SwiftLanguageService/ShowObjCSelector.swift
ahoppen, line 1: Could you add the new files to [CMakeLists.txt](https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SwiftLanguageService/CMakeLists.txt) for the Windows build?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -978,6 +979,22 @@ extension SwiftLanguageService {
       refactorActions.append(CodeAction(title: expandMacroCommand.title, kind: .refactor, command: expandMacroCommand))
     }
 
+    let methodKinds: [SymbolKind] = [.method, .function, .constructor]
+    let isOnMethod = cursorInfoResponse.cursorInfo.contains { cursorInfo in
+      if let kind = cursorInfo.symbolInfo.kind {
+        return methodKinds.contains(kind)
+      }
+      return false
+    }
```
ahoppen, line 1108: Does this mean that we show the refactoring action on every Swift method, even if it isn’t backed by an Objective-C method and thus the *Show Objective-C Selector* would just fail? Do we have some indication in the cursor info response already to see whether this method is defined in Objective-C / that we have an Objective-C selector? One ugly idea I have is checking whether the USR starts with `s:` but there might be nicer implementations as well.

ahoppen (commented):
Nice, this looks good. Could you also add a test case for the new refactoring action? Both for the case where it shows and also a case checking that we don’t show it if the function doesn’t have an Objective-C selector.

Sources/SwiftLanguageService/SwiftLanguageService.swift
iMostfa: Good catch, yes it was offered on every function.

Your USR idea was the only option i found as well.

and is what the code does now: cursor info reports the Objective-C USR (`c:`) for declarations exposed to Objective-C and the Swift USR (`s:`) otherwise, so the check is free. We only ask `sourcekitd` for a selector when the cursor is on a function with a `c:` USR, and since that gives us the selector it's carried in the command, so invoking the action won't need a second request

wdyt ?

Sources/SwiftLanguageService/ShowObjCSelectorCommand.swift
iMostfa: done

iMostfa:
Hi, @ahoppen 
Sorry for going quiet on this one for so long, life got busy.
i rebased both PRs to latest main, and applied comments.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -459,6 +459,47 @@ package actor SkipUnless {
     }
   }
 
+  package static func sourcekitdSupportsObjCSelector(
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws {
+    return try await shared.skipUnlessSupportedByToolchain(swiftVersion: SwiftVersion(6, 5), file: file, line: line) {
```
iMostfa, line 466: i'm not sure if there's a better way to do this, i'm assuming that swift's PR will be in Swift 6.5, but i could be wrong here.

Sources/SKTestSupport/SkipUnless.swift
```diff
@@ -459,6 +459,47 @@ package actor SkipUnless {
...
+          import Foundation
+
+          class Foo: NSObject {
+            @objc func 1️⃣bar() {}
+          }
+          """
+        )
+
+        let testURL = scratchDirectory.appending(component: "test.swift")
+        try source.write(to: testURL, atomically: false, encoding: .utf8)
```
ahoppen: Could you use `writeWithRetry` here?

Sources/SwiftLanguageService/ShowObjCSelector.swift
```diff
@@ -0,0 +1,40 @@
...
+@_spi(SourceKitLSP) import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+import SourceKitD
+import SourceKitLSP
+
+extension SwiftLanguageService {
+  /// The Objective-C selector of the declaration at `position`, or `nil` if it isn't exposed to Objective-C or the
+  /// toolchain doesn't know the request.
+  func objcSelector(_ uri: DocumentURI, at position: Position) async -> String? {
+    return await orLog("Getting Objective-C selector", level: .debug) {
```
ahoppen: Why are you logging at the debug level here?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1084,6 +1084,23 @@ extension SwiftLanguageService {
...
+        return false
+      }
+    }
+    if isOnObjCFunction, let selector = await objcSelector(params.textDocument.uri, at: params.range.lowerBound) {
+      let showObjCSelectorCommand = ShowObjCSelectorCommand(selector: selector).asCommand()
+      refactorActions.append(
+        CodeAction(title: showObjCSelectorCommand.title, kind: .refactor, command: showObjCSelectorCommand)
+      )
+    }
+
```
ahoppen, line 1115: This isn’t really a refactoring action so the code action kind probably shouldn’t be `.refactor`. Could you check if we could either:
1. Define our own code action kind and announce it in the initialization options. This should be supported according to the LSP spec. Check that VS Code displays the code action.
> The set of kinds is open and client needs to announce the kinds it supports
> to the server during initialization.
2. If that doesn’t work, could we use `.empty` instead?

Also, the assumption that `retrieveRefactorCodeActions` only returns `.refactor` actions no longer holds, so the filter after `providersAndKinds` no longer works. I would suggest renaming this method from `retrieveRefactorCodeActions` to `retrieveCursorInfoBasedCodeActions` and make it take a list of requested code action kinds to then filter by them.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1084,6 +1084,23 @@ extension SwiftLanguageService {
       refactorActions.append(CodeAction(title: expandMacroCommand.title, kind: .refactor, command: expandMacroCommand))
     }
 
+    // Declarations exposed to Objective-C have an Objective-C USR (`c:`), everything else has a Swift USR (`s:`).
+    // Checking it avoids a semantic request per code action on Swift-only declarations.
+    let isOnObjCFunction = cursorInfoResponse.cursorInfo.contains { cursorInfo in
+      switch cursorInfo.symbolInfo.kind {
+      case .method, .function, .constructor:
+        return cursorInfo.symbolInfo.usr?.hasPrefix("c:") ?? false
```
ahoppen, line 1104: Instead of interpreting the mangling prefix, could you inspect `key.decl_lang` from the cursor info response instead?

ahoppen (commented):
Thanks for picking this back up.

Sources/SKTestSupport/SkipUnless.swift
iMostfa: Done in 1197fb47.

Sources/SwiftLanguageService/ShowObjCSelector.swift
iMostfa: It made sense when the request was sent for every
function and was expected to fail on most of them. Now that it only runs for
declarations cursor info already reports as exposed to Objective-C, a failure
is unexpected, so it uses the default level in 5da7c73c.

Sources/SwiftLanguageService/SwiftLanguageService.swift
iMostfa: as far as i can see , but it returns from [`getUIDForDeclLanguage`](https://github.com/swiftlang/swift/blob/d37faf0b09c2b819926f3a72549cf7bd5cf215a9/tools/SourceKit/lib/SwiftLang/SwiftLangSupport.cpp#L326-L330), which returns `source.lang.objc` only when the declaration has a Clang node:

```cpp
UIdentSwiftLangSupport::getUIDForDeclLanguage(const swift::Decl *D) {
  if (D->hasClangNode())
    return KindObjC;
  return KindSwift;
}
```

for `@objc func annotated()` i got `source.lang.swift`, only for `NSString.substring(from:)` it will work.

one thing we can do is try to check it first, otherwise we fallback for usr? but i'm not sure if makes any different. what do you think ?

Sources/SwiftLanguageService/SwiftLanguageService.swift
iMostfa: thanks, i will need to dip up more to see which option is better

Sources/SwiftLanguageService/SwiftLanguageService.swift
ahoppen: @rintaro @hamishknight Do you have a good suggestion of how to decide whether we should offer a code action to copy a Swift function’s Objective-C selector? Can you think of some property of an `AbstractFunctionDecl` that we could add to a cursor info response to indicate whether it’s callable from Objective-C?

PR #2403, opened 2025-12-22
🚥 Implement Document Highlight for all return / throws (function/closure/accessor/init/deinit) and break / continue (for/while/repeat...while)

lokesh-tr:
⚠️ Under Progress ⚠️

- **Explanation**:
Upgrades `textDocument/documentHighlight` to highlight various return / throw statements corresponding to a function / closure / accessor / init / deinit, and break / continue statements corresponding to a loop, thereby allowing us to not wonder about which statements could cause an exit from a function or loop.

This also makes use of a new API introduced in SwiftLexicalLookup Library in Swift Syntax.
- **Scope**:
This doesn't break existing code. This is Additive.
- **Issues**:
Fixes: https://github.com/swiftlang/sourcekit-lsp/issues/2400
- **Original PRs**:
This is the original PR. New API for Swift Syntax introduced in https://github.com/swiftlang/swift-syntax/pull/3216
- **Risk**:
I don't think there's any risk associated with this new feature.
- **Testing**:
Tests and Docs are yet to be written. The code & This PR description will also be polished.
- **Reviewers**:
Suggested by @ahoppen and Open to be reviewed by anyone. Note: At the current state, the code is messy since I was focused on getting it to work, will polish it.

### DIFF
--- a/Sources/SwiftLanguageService/SwiftLanguageService.swift
+++ b/Sources/SwiftLanguageService/SwiftLanguageService.swift
@@ -926,7 +926,7 @@ extension SwiftLanguageService {
         if node.lookupControlStructure() == targetStructure {
           highlights.append(
             DocumentHighlight(
-              range: snapshot.absolutePositionRange(of: node.positionAfterSkippingLeadingTrivia..<node.endPosition),
+                range: snapshot.absolutePositionRange(of: node.firstToken(viewMode: .sourceAccurate).map { $0.positionAfterSkippingLeadingTrivia..<$0.endPositionBeforeTrailingTrivia } ?? node.positionAfterSkippingLeadingTrivia..<node.endPositionBeforeTrailingTrivia),
               kind: .read
             )
           )


lokesh-tr:
This is how it looks:

Inner function return statements are highlighted
Outer function return statements are highlighted

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
...
+        self.targetStructure = targetStructure
+        self.snapshot = snapshot
+        super.init(viewMode: .sourceAccurate)
+      }
+      
+      private func addHighlightIfMatches(_ node: some SyntaxProtocol) {
+        if node.lookupControlStructure() == targetStructure {
+          highlights.append(
+            DocumentHighlight(
+              range: snapshot.absolutePositionRange(of: node.positionAfterSkippingLeadingTrivia..<node.endPosition),
```
ahoppen: Should we just highlight the control flow keyword (eg. `return`). I think that might look nicer, especially if the returned expression spans multiple lines.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
...
+    at position: Position,
+    in snapshot: DocumentSnapshot
+  ) async throws -> [DocumentHighlight]? {
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    
+    guard let tokenSyntax = syntaxTree.token(at: snapshot.absolutePosition(of: position)) else {
+      return nil
+    }
+
+    guard let targetStructure = Syntax(tokenSyntax).parent!.lookupControlStructure() else {
```
ahoppen: We shouldn’t force unwrap the parent here.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
     }
   }
 
+  private func documentSymbolHighlightHelper(
```
ahoppen: I would give this a more descriptive name which highlights that this is about control flow keyword highlighting.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
...
+          highlights.append(
+            DocumentHighlight(
+              range: snapshot.absolutePositionRange(of: node.positionAfterSkippingLeadingTrivia..<node.endPosition),
+              kind: .read
+            )
+          )
+        }
+      }
+      
+      override func visit(_ node: BreakStmtSyntax) -> SyntaxVisitorContinueKind {
```
ahoppen, line 936: You shouldn’t need to repeat the list of control flow nodes if this is a `SyntaxAnyVisitor` and you add a `visitAny` override.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
...
+      let snapshot: DocumentSnapshot
+      
+      init(targetStructure: Syntax, snapshot: DocumentSnapshot) {
+        self.targetStructure = targetStructure
+        self.snapshot = snapshot
+        super.init(viewMode: .sourceAccurate)
+      }
+      
+      private func addHighlightIfMatches(_ node: some SyntaxProtocol) {
+        if node.lookupControlStructure() == targetStructure {
```
ahoppen, line 926: This could be a `guard` to avoid the nesting level.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -889,6 +898,67 @@ extension SwiftLanguageService {
...
+          highlights.append(
+            DocumentHighlight(
+                range: snapshot.absolutePositionRange(of: node.firstToken(viewMode: .sourceAccurate).map { $0.positionAfterSkippingLeadingTrivia..<$0.endPositionBeforeTrailingTrivia } ?? node.positionAfterSkippingLeadingTrivia..<node.endPositionBeforeTrailingTrivia),
```
ahoppen, line 929: Could you run swift-format on your changes?

ahoppen, line 929: While currently the `return`/`break` etc keyword is always the first token in the syntax node, this may change in the future. I’d prefer if we explicitly pass in the token to highlight.

ahoppen, line 929: Instead of `$0.positionAfterSkippingLeadingTrivia..<$0.endPositionBeforeTrailingTrivia` you should be able to do `$0.trimmedRange` or something like that.

DPrakashhh:
@lokesh-tr  are you working on this problem?

lokesh-tr:
@DPrakashhh yeah, i got some pushes to be made

DPrakashhh:
@lokesh-tr Okay!

PR #2404, opened 2025-12-25
enhancement: Refine jump-to-definition for literals

### LINKED ISSUE #2368: Do not offer jump-to-definition for literals

a7maad-ayman:
This PR refines jump-to-definition behavior by excluding literal values from definition lookups.

### Problem

Currently, invoking jump-to-definition on literal values (strings, integers, floats, booleans, `nil`) triggers a definition lookup that returns no meaningful result, since literals are concrete values written directly in source code with no declaration to navigate to.

### Solution

Detect when the cursor is positioned on a literal token and return an empty response early, avoiding unnecessary SourceKit queries. This aligns with how SourceKit handles non-symbol tokens.

### Changes

- Add literal token detection in `SwiftLanguageService` using SwiftSyntax
- Handle string segments, integer/float literals, and boolean/nil keywords
- Return empty definition response for literal positions in `symbolInfo`
- Add tests covering various literal types

Resolves #2368

Co-Authored-By: clemo97 <lumumbaclement@gmail.com>

### DIFF
--- a/Tests/SourceKitLSPTests/DefinitionTests.swift
+++ b/Tests/SourceKitLSPTests/DefinitionTests.swift
@@ -725,7 +725,7 @@ class DefinitionTests: SourceKitLSPTestCase {
     )
     XCTAssertTrue(definitions?.locations?.first?.uri.pseudoPath.hasSuffix("Lib.swiftinterface") ?? false)
   }
-  
+
   func testDefinitionOnLiteralsShouldReturnEmpty() async throws {
     let testClient = try await TestSourceKitLSPClient()
     let uri = DocumentURI(for: .swift)


Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -201,6 +201,13 @@ package protocol LanguageService: AnyObject, Sendable {
   func doccDocumentation(_ req: DoccDocumentationRequest) async throws -> DoccDocumentationResponse
   func symbolInfo(_ request: SymbolInfoRequest) async throws -> [SymbolDetails]
 
+  /// Checks whether the given position is on (or immediately after) a literal value token.
+  ///
+  /// Literal value tokens do not support jump-to-definition because they represent
+  /// concrete values written directly in source code and therefore have no
+  /// meaningful definition location.
+  func isPositionOnLiteralToken(_ position: Position, in uri: DocumentURI) async -> Bool
```
ahoppen: I don’t think we need this as a requirement in `LanguageService` at all if we are only calling it from `SwiftLanguageService.symbolInfo`, right?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1293,3 +1293,65 @@ extension sourcekitd_api_uid_t {
...
+    let absolutePosition = snapshot.absolutePosition(of: position)
+
+    func isLiteralTokenKind(_ token: TokenSyntax) -> Bool {
+      switch token.tokenKind {
+      case .stringSegment,
+           .integerLiteral,
+           .floatLiteral:
+        return true
+      case .keyword(let keyword) where keyword == .true || keyword == .false || keyword == .nil:
+        return true
```
ahoppen, line 1337: I think this can be simplified slightly

```suggestion
      case .stringSegment,
           .integerLiteral,
           .floatLiteral,
           .keyword(.true), .keyword(.false), .keyword(.nil)::
        return true
```

Tests/SourceKitLSPTests/DefinitionTests.swift
```diff
@@ -689,4 +689,82 @@ class DefinitionTests: SourceKitLSPTestCase {
     )
     XCTAssertEqual(response?.locations?.map(\.uri), [try project.uri(for: "Test.h")])
   }
+
+  func testDefinitionOnLiteralsShouldReturnEmpty() async throws {
```
ahoppen, line 729: Could you run swift-format on your changes? https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting

Tests/SourceKitLSPTests/DefinitionTests.swift
```diff
@@ -689,4 +689,82 @@ class DefinitionTests: SourceKitLSPTestCase {
...
+  }
+
+  func testDefinitionInsideLiteralsShouldStillWork() async throws {
+      let testClient = try await TestSourceKitLSPClient()
+      let uri = DocumentURI(for: .swift)
+
+      let positions = testClient.openDocument(
+          """
+          let name = "Ayman"
+          let greeting = "Hello \\(1️⃣name)"
```
ahoppen: If you start the string literal with `#"""` you don’t need the double backslash here.

ahoppen (commented):
This looks great. Just a few comments to clean up the code slightly.

ahoppen:
I just realized that we have two open PRs open for this from @Clemo97 (https://github.com/swiftlang/sourcekit-lsp/pull/2394) and you, @a7maad-ayman, (https://github.com/swiftlang/sourcekit-lsp/pull/2404). First of all, thanks to both of you for working on this issue 🙏🏽. Given that both PRs contain (nearly) the same contents, you have already converged on a solution. To consolidate the two PRs, I would suggest that: 
- We use this PR as the main one since @a7maad-ayman expressed their interest in working on https://github.com/swiftlang/sourcekit-lsp/issues/2368 first. 
- @a7maad-ayman since you based this PR on @Clemo97’s could you add `Co-Authored-By: clemo97 <lumumbaclement@gmail.com>` to your commit message. That way you both get credited for collaborating on this issue and get listed as authors of this fix?

Would that work for both of you?

a7maad-ayman:
I’ve squashed the commits and added the Co-Authored-By line for @Clemo97 so we’re both properly credited.

Thanks again for the review and for pointing that out — much appreciated.

Screenshot 2026-01-03 at 10 05 02 PM

ahoppen:
Looks like the `Co-Authored-By:` line needs to be the last in your commit message. Also, the history in your branch now contains a merge commit, which I personally dislike because it creates a more convoluted git history. The best way to remedy this is probably if you just create a fresh commit with all changes in your PR and then force push it

```sh
git reset 7495f5532fdb17184d69518f46a207e596b26c64
# Commit your changes, including Co-Authored-By in the last line of the commit
git push -f
```

ahoppen:
Sorry for not triggering a test run when you pushed your changes. There’s a merge conflict now, could you rebase your changes on top of the latest `main` to resolve it?

a7maad-ayman:
Hi Alex, I solved the conflict could U run the CI now 🙏🏼 .

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #2406, opened 2025-12-30
feat(swift-language-service): implement De Morgan's Law code action

PhantomInTheWire:
fixes: https://github.com/swiftlang/sourcekit-lsp/issues/1243
previous/reference pr: https://github.com/swiftlang/sourcekit-lsp/pull/1552

#### Summary

- Adds a new "Apply De Morgan's Law" code action for refactoring complement expressions
- Supports both boolean operators (!, &&, ||) and bitwise operators (~, &, |)
- Automatically handles operator precedence by adding parentheses where needed
- Simplifies double negations (e.g., !!a → a) when possible
- Includes comprehensive tests covering nested expressions, ternary operators, and edge cases

https://github.com/user-attachments/assets/7ffeadbf-1873-4151-96a9-949e5c059df8

### DIFF
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1260,9 +1260,9 @@ final class CodeActionTests: SourceKitLSPTestCase {
   func testApplyDeMorganLawNestedActionAvailability() async throws {
     try await assertCodeActions(
       """
-      let x = 2️⃣!3️⃣(4️⃣!(1️⃣a && b) 5️⃣|| c6️⃣)7️⃣
+      let x = 1️⃣!2️⃣(3️⃣!(4️⃣a && b) 5️⃣|| c6️⃣)7️⃣
       """,
-      markers: ["1️⃣"],
+      markers: ["4️⃣"],
       exhaustive: false
     ) { uri, positions in
       [
@@ -1273,7 +1273,7 @@ final class CodeActionTests: SourceKitLSPTestCase {
             changes: [
               uri: [
                 TextEdit(
-                  range: positions["4️⃣"]..<positions["5️⃣"],
+                  range: positions["3️⃣"]..<positions["5️⃣"],
                   newText: "(!a || !b) "
                 )
               ]
@@ -1287,7 +1287,7 @@ final class CodeActionTests: SourceKitLSPTestCase {
             changes: [
               uri: [
                 TextEdit(
-                  range: positions["4️⃣"]..<positions["6️⃣"],
+                  range: positions["3️⃣"]..<positions["6️⃣"],
                   newText: "!((a && b) && !c)"
                 )
               ]
@@ -1301,7 +1301,7 @@ final class CodeActionTests: SourceKitLSPTestCase {
             changes: [
               uri: [
                 TextEdit(
-                  range: positions["3️⃣"]..<positions["7️⃣"],
+                  range: positions["2️⃣"]..<positions["7️⃣"],
                   newText: "!((a && b) && !c)"
                 )
               ]
@@ -1315,7 +1315,7 @@ final class CodeActionTests: SourceKitLSPTestCase {
             changes: [
               uri: [
                 TextEdit(
-                  range: positions["2️⃣"]..<positions["7️⃣"],
+                  range: positions["1️⃣"]..<positions["7️⃣"],
                   newText: "((a && b) && !c)"
                 )
               ]


Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1058,431 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testApplyDeMorganLawNegatedOr() async throws {
+    // !(a || b) -> !a && !b
```
ahoppen: To help keep the tests concise, I would prefer to have one test that verifies that a code action is generated and have all other tests call straight into an entry point to `DeMorganCandidateSequence` without any LSP request setup. That way there’s less overhead / boilerplate involved, which also makes it easier to debug these tests if you need to step into the actual transformation.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1058,431 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+      [
+        CodeAction(
+          title: "Convert !(a && b) to (!a || !b)",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["1️⃣"]..<positions["2️⃣"],
+                  newText: "(!a || !b)"
```
ahoppen, line 1081: Could we infer that the surrounding brackets are not necessary because there isn’t an operator expression around it that has a higher precedence than `||`? If it’s not trivial, this could also be a great follow-up PR.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+      return []
+    }
+
+    for candidate in DeMorganCandidateSequence(node: node, snapshot: scope.snapshot) {
+      let transformer = DeMorganTransformer()
+      guard let complement = transformer.computeComplement(of: candidate.expr) else {
+        continue
+      }
+
+      let complementText = "\(complement)"
```
ahoppen: Nitpick, I’d prefer 

```suggestion
      let complementText = complement.description
```

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+  private var candidates: [ExprSyntax]
+
+  init(node: Syntax, snapshot: DocumentSnapshot) {
+    self.snapshot = snapshot
+    self.candidates = []
+
+    var current = node.findDeMorganExprParent()
+    while let expr = current {
+      candidates.append(expr)
+      current = expr.parent?.findDeMorganExprParent()
```
ahoppen, line 130: Could you add a test case where there are multiple candidate expressions to apply De Morgan’s law to?

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+
+  mutating func next() -> (expr: ExprSyntax, range: Range<Position>)? {
+    guard let expr = candidates.popLast() else {
+      return nil
+    }
+    return preprocess(expr)
+  }
+
+  /// Preprocess the expression, extracting the right-hand side of assignments.
+  private func preprocess(_ expr: ExprSyntax) -> (expr: ExprSyntax, range: Range<Position>) {
```
ahoppen: I don’t quite understand what `preprocess` is for. Could you maybe add an example to the doc comment to illustrate it?

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+      return (mapThroughParentheses(expr, replacement: result.expr), result.change)
+    }
+
+    if let result = negateTernary(inner, exprType: exprType) {
+      if result.change == .negation {
+        return result
+      }
+      return (mapThroughParentheses(expr, replacement: result.expr), result.change)
+    }
+
```
ahoppen, line 308: Nitpick: Should we also try to replace `true` by `false` and vise versa here?

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+    guard let leftNegated = negateExpression(infixExpr.leftOperand, exprType: exprType),
+      let rightNegated = negateExpression(infixExpr.rightOperand, exprType: exprType)
+    else {
+      return nil
+    }
+
+    guard leftNegated.change != .negation || rightNegated.change != .negation else {
+      return nil
+    }
```
ahoppen, line 247: Why is this check necessary? Could you maybe add a short example as a comment that illustrates it?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1058,431 @@ final class CodeActionTests: SourceKitLSPTestCase {
     }
   }
 
+  func testApplyDeMorganLawNegatedAnd() async throws {
```
ahoppen, line 1064: Could you add one test case that checks that we maintain trivia, eg. `/*a*/ ! /*b*/ ( /*c*/ a /*d*/ && /*e*/ b /*f*/ ) /*g*/`.

I would deem it acceptable if we drop trivia that doesn’t really have a good place to fit anymore but we should make these choices deliberately.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+      \.operator,
+      ExprSyntax(binOp.with(\.operator.tokenKind, .binaryOperator(flipped)))
+    )
+
+    return (ExprSyntax(newInfix), .comparison)
+  }
+
+  /// Propagates negation through ternary: `a ? b : c` → `a ? !b : !c`
+  ///
+  /// If both branches would just add negation prefixes, wraps the whole ternary instead: `!(a ? b : c)`
```
ahoppen, line 400: Do we have a test case that adds the negation outside of the ternary? I might just have missed it.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+  }
+  return replacement
+}
+
+fileprivate extension ExprSyntax {
+  /// Returns the expression looking through parentheses.
+  var lookingThroughParentheses: ExprSyntax {
+    if let tuple = self.as(TupleExprSyntax.self),
+      let inner = tuple.singleUnlabeledExpression
+    {
```
ahoppen: I think this can be simplified slightly.

```suggestion
    if let inner = self.as(TupleExprSyntax.self)?.singleUnlabeledExpression {
```

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+
+  /// Returns true if this expression is wrapped in parentheses.
+  var isParenthesized: Bool {
+    self.as(TupleExprSyntax.self)?.isParentheses ?? false
+  }
+
+  /// Returns true if this is a composite expression that needs parentheses when negated.
+  var isComposite: Bool {
+    switch self.kind {
+    case .asExpr, .awaitExpr, .infixOperatorExpr, .isExpr, .ternaryExpr, .tryExpr:
```
ahoppen: Since new expression types are likely going to be composites and adding parentheses is always the safer option, I would prefer to explicitly state the kinds of expressions that are not composite instead of listing those that aren’t.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+    for candidate in DeMorganCandidateSequence(node: node, snapshot: scope.snapshot) {
+      let transformer = DeMorganTransformer()
+      guard let complement = transformer.computeComplement(of: candidate.expr) else {
+        continue
+      }
+
+      let complementText = "\(complement)"
+      return [
+        CodeAction(
+          title: "Convert \(candidate.expr) to \(complementText)",
```
ahoppen: Should we explicitly say that this is applying De Morgan’s law? Otherwise users might be surprised why we are suggesting this transform.


```suggestion
          title: "Apply De Morgan’s law, converting '\(candidate.expr)' to '\(complementText)'",
```

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+  /// Returns true if this is a parenthesized expression (single unlabeled element).
+  var isParentheses: Bool {
+    singleUnlabeledExpression != nil
+  }
+
+  /// Returns the single unlabeled expression if this is a parenthesized expression.
+  var singleUnlabeledExpression: ExprSyntax? {
+    guard elements.count == 1,
+      let only = elements.first,
+      only.label == nil
```
ahoppen: We should have `.only` defined on `Sequence` to simplify this.

```suggestion
    guard let only = elements.only, only.label == nil
```

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,549 @@
...
+  /// Propagated negation to ternary branches
+  case ternary
+  /// Changed `&&` to `||`
+  case andToOr
+  /// Changed `||` to `&&`
+  case orToAnd
+}
+
+/// Result of negating an expression.
+private typealias NegatedResult = (expr: ExprSyntax, change: NegationChange)
```
ahoppen: I think it would be slightly cleaner to define this as a struct instead of a type alias.

ahoppen (commented):
Thanks a lot! The algorithm looks great! 🤩 I left some smaller comments inline.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: awesome, tests went from 6s to 0.5s. done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: sure i'll make a follow up pr for this, should i also create an issue or a todo in the code to refer to for this?

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: i think i forgot to remove this. removed

Tests/SourceKitLSPTests/CodeActionTests.swift
ahoppen: If you think that you’ll get to it in the next couple of weeks, there’s no need to create an issue for it.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: makes sense. done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: added a more explicit one.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: makes sense. done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: sure that makes sense

PhantomInTheWire:
@ahoppen should be good to review

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1060,259 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+  }
+
+  /// Tests De Morgan transformation directly without LSP overhead.
+  private func assertDeMorganTransform(
+    input: String,
+    expected: String,
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) throws {
+    let sourceFile = Parser.parse(source: "let x = \(input)")
```
ahoppen: I don’t think we need the `let x = `. We should be able to parse top-level expressions directly. The easiest way might be to go through the `ExprSyntax` initializer that takes a string literal, defined in `SwiftSyntaxBuilder`

```swift
import SwiftSyntax
import SwiftSyntaxBuilder

let expr = ExprSyntax("\(input)")
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1060,259 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+
+    // Comments attached to inner operators
+    try assertDeMorganTransform(
+      input: "!(a /*op*/ && b)",
+      expected: "(!a /*op*/ || !b)"
+    )
+  }
+
+  /// Tests De Morgan transformation directly without LSP overhead.
+  private func assertDeMorganTransform(
```
ahoppen: Could you make this a top-level function (because it doesn’t need any members from the test file) and keep it at the end of this file instead of mixing it with the test cases?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1060,259 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+  func testApplyDeMorganLawNegatedOr() throws {
+    // !(a || b) -> !a && !b
+    try assertDeMorganTransform(
+      input: "!(a || b)",
+      expected: "(!a && !b)"
+    )
+  }
+
+  func testApplyDeMorganLawDoubleNegation() throws {
+    // !(!a && !b) -> a || b
```
ahoppen: I don’t think these comments are necessary anymore because the `assertDeMorganTransform` now exactly states what we’re testing without any noise. 😉

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1060,259 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+          !a
+          ||
+          !b
+        )
+        """
+    )
+
+    // Line comments
+    try assertDeMorganTransform(
+      input: "!(a && b // check\n)",
```
ahoppen: I think this would also read a little nicer as a multi-line string literal

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
ahoppen: Correct me if I’m wrong but in this code you are collecting a list of candidates, which leads me to believe that there are cases where we offer multiple code actions to convert using De Morgan’s law at the same location. Is that correct? If so, could we have a test case that checks that two code actions are offered?

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
ahoppen: Nice explanation and good job thinking about the case where we don’t want to increase complexity 😍

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,568 @@
...
+    return result.expr
+  }
+}
+
+/// Maps a replacement through parentheses, preserving the outer structure.
+private func mapThroughParentheses(_ expr: ExprSyntax, replacement: ExprSyntax) -> ExprSyntax {
+  if let inner = expr.as(TupleExprSyntax.self)?.singleUnlabeledExpression {
+    let mappedInner = mapThroughParentheses(inner, replacement: replacement)
+    return ExprSyntax(
+      expr.cast(TupleExprSyntax.self).with(\.elements, [LabeledExprSyntax(expression: mappedInner)])
```
ahoppen: Oh, sorry, totally missed that you used the `TupleExprSyntax` here when I suggested the simplification. Let’s stick with your original implementation to avoid the `cast` call.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,568 @@
...
+  mutating func next() -> (expr: ExprSyntax, range: Range<Position>)? {
+    guard let expr = candidates.popLast() else {
+      return nil
+    }
+    return (expr, snapshot.range(of: expr))
+  }
+}
+
+/// Encapsulates all De Morgan transformation logic.
+@_spi(Testing) public struct DeMorganTransformer {
```
ahoppen: Instead of `@_spi(Testing)` you could make this a `package` access level, to reduce accessibility a little further. Same for the `public` functions inside of it.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
PhantomInTheWire: I initially restricted this to the outermost eligible expression because I wasn’t fully sure whether multiple actions should be offered, or at which positions. but I agree it’s reasonable to offer all valid candidates on any position on the line. i have updated the code and tests.

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -0,0 +1,569 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
```
ahoppen:
```suggestion
// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
```

😉

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1058,6 +1061,273 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    ) { uri, positions in
+      [
+        CodeAction(
+          title: "Apply De Morgan’s law, converting '!(a && b) ' to '(!a || !b) '",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: Position(line: 0, utf16index: 10)..<Position(line: 0, utf16index: 20),
```
ahoppen: One final minor nitpick: Line + column-based positions are always finicky to work with because nobody likes counting characters in the test input. I would suggest adding markers for all positions we need. If you pass `markers: ["1️⃣"]` (or whichever number you end up choosing, these markers won’t be used as test locations.

PhantomInTheWire: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1260,64 +1260,62 @@ final class CodeActionTests: SourceKitLSPTestCase {
   func testApplyDeMorganLawNestedActionAvailability() async throws {
     try await assertCodeActions(
       """
-      let x = !(!(1️⃣a && b) || c)
+      let x = 2️⃣!3️⃣(4️⃣!(1️⃣a && b) 5️⃣|| c6️⃣)7️⃣
```
ahoppen: I know I’m picky but could you order the markers ascendingly? And then probably use `4` as the marker at which we invoke the code action.

PhantomInTheWire: naah its no biggie

PhantomInTheWire: done.

ahoppen (approved):
Let’s 🚢 it

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2409, opened 2026-01-03
Fix language services lifecycle management in Workspace

### LINKED ISSUE #2209: `Workspace.documentServices` is never cleared up
Reading through code, I noticed that `Workspace.documentServices` only ever has elements added to it and we never remove elements from it, even when documents are closed.

What’s more is that `setDocumentServices` is called from places other than `openDocument`, so `documentServices` might contain more documents than the ones that have been opened by the client.

I think the correct thing to do here is to make `SourceKitLSPServer.languageService` just return an existing or newly started language service and only call `setLanguageServices` from `openDocument`. That way we don’t need to deal with race conditions for multiple calls to `setLanguageServices` anymore (`openDocument` is a blocker request for requests affecting that document URI) and we can remove documents from `documentServices` when a document is closed.

loveucifer:
Closes #2209

- Only call `setLanguageServices` from `openDocument` to avoid race conditions
- Remove language services when documents are closed via `removeLanguageServices`
- `SourceKitLSPServer.languageServices` now just returns services without storing them
- Fixed a small typo (serveer -> server) while I was in there

This way the languageServices dictionary only tracks documents that are actually open, and we avoid race conditions since openDocument is a blocking request for that document.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -592,7 +592,7 @@ package actor SourceKitLSPServer {
       """
     )
 
-    return workspace.setLanguageServices(for: uri, languageServices)
+    return languageServices
   }
 
   /// The language service with the highest precedence that can handle the given document.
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

PR #2410, opened 2026-01-03
Use for loops to accumulate SourceItem elements in `SwiftPMBuildServer`

ahoppen:
This seems cleaner than having multiple map calls that then get concatenated.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #2412, opened 2026-01-03
Provide semantic functionality in copy destinations

ahoppen:
If we have a source item `include/Test.h` that gets copied to `build/Test.h`, we currently don’t provide any semantic functionality for `build/Test.h`. Improve the build settings fallback logic to construct fallback build settings for `build/Test.h` based on `include/Test.h`.

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -503,17 +503,15 @@ package actor BuildServerManager: QueueBasedMessageHandler {
 
   private let cachedSourceFilesAndDirectories = Cache<SourceFilesAndDirectoriesKey, SourceFilesAndDirectories>()
 
-  /// The latest map of copied file URIs to their original source locations.
-  ///
-  /// We don't use a `Cache` for this because we can provide reasonable functionality even without or with an
-  /// out-of-date copied file map - in the worst case we jump to a file in the build directory instead of the source
-  /// directory.
-  /// We don't want to block requests like definition on receiving up-to-date index information from the build server.
-  private var cachedCopiedFileMap: [DocumentURI: DocumentURI] = [:]
+  /// Task that computes the latest map of copied file URIs to their original source locations.
+  private var copiedFileMap: Task<[DocumentURI: DocumentURI], Never>?
 
-  /// The latest task to update the `cachedCopiedFileMap`. This allows us to cancel previous tasks to update the copied
-  /// file map when a new update is requested.
-  private var copiedFileMapUpdateTask: Task<Void, Never>?
+  /// The last computed copied file map, which may be out-of-date.
+  ///
+  /// Even with out-of-date information for the copied file map, we can provide reasonable functionality - in the worst
+  /// case we jump to a file in the build directory instead of the source directory. We don't want to block requests
+  /// like definition on receiving up-to-date build target information from the build server.
+  private var cachedCopiedFileMap: [DocumentURI: DocumentURI] = [:]
 
   /// The `SourceKitInitializeBuildResponseData` received from the `build/initialize` request, if any.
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -1258,22 +1294,25 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
-      // build server, but this is a good first check to avoid requesting build settings for too many documents. And
-      // since all of this is fallback-logic, skipping over possibly valid files is not a correctness issue.
-      guard let siblingLanguage = Language(inferredFromFileExtension: uri), siblingLanguage == language else {
-        return nil
-      }
-      return uri
-    }.sorted(by: { $0.pseudoPath < $1.pseudoPath }).first
+    var siblingFile: DocumentURI? = cachedCopiedFileMap[document]
+    if siblingFile == nil {
+      siblingFile = try await self.sourceFilesAndDirectories().files.compactMap { (uri, info) -> DocumentURI? in
```
hamishknight, line 1319: Not sure I understand the logic here, shouldn't this be handled by `fallbackBuildSettingsInferredFromCopySource`?

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -486,18 +486,16 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+  /// Task that computes the latest map of copied file URIs to their original source locations.
+  private var copiedFileMap: Task<[DocumentURI: DocumentURI], Never>?
+
+  /// The last computed copied file map, which may be out-of-date.
   ///
-  /// We don't use a `Cache` for this because we can provide reasonable functionality even without or with an
-  /// out-of-date copied file map - in the worst case we jump to a file in the build directory instead of the source
-  /// directory.
-  /// We don't want to block requests like definition on receiving up-to-date index information from the build server.
+  /// Even with out-of-date information for the copied file map,we can provide reasonable functionality - in the worst
```
hamishknight:
```suggestion
  /// Even with out-of-date information for the copied file map, we can provide reasonable functionality - in the worst
```

Sources/BuildServerIntegration/BuildServerManager.swift
ahoppen: I’m not sure if I fully understood your comment now but the goal of `fallbackBuildSettingsInferredFromSiblingFile` is to find a file in the same directory as the document that the user has opened. This is useful if eg. the build system hasn’t been reloaded yet to include this new file. `fallbackBuildSettingsInferredFromCopySource` only looks at the file itself.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/BuildServerIntegration/BuildServerManager.swift
hamishknight: Right, but the new logic checks `cachedCopiedFileMap` and uses that if present, otherwise does the sibling file logic, which seems like it's no different to what `fallbackBuildSettingsInferredFromSiblingFile` is doing?

ahoppen: You’re right. I think I tried to implement the entire copied file fallback logic with the sibling fallback logic and then realized that that‘s not possible because the sibling fallback logic adds the source file while the copied file fallback logic replaces the source file. 

Reverting this implementation back to the original one in https://github.com/swiftlang/sourcekit-lsp/pull/2593

Thanks for noticing.

PR #2413, opened 2026-01-03
Only adjust location for copied files if the original file exists

ahoppen:
If the original file does not exist, it is preferable to jump to a file generated by the build system instead of trying to jump to a file that the editor can’t open. This will become important when using the copied file logic to jump to files in `.build/checkouts` instead of `.build/index-build/checkouts` in a follow-up PR.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

PR #2415, opened 2026-01-04
Do not log fault if `BuildServermanager.mainFilesProvider` is `nil`

ahoppen:
It’s perfectly fine for `mainFilesProvider` to be `nil` if we don’t have an index. We shouldn’t log a fault for that.

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -380,11 +380,11 @@ package actor BuildServerManager: QueueBasedMessageHandler {
   }
 
   package func mainFilesProvider<T: MainFilesProvider>(as: T.Type) async -> T? {
-    guard let mainFilesProvider = mainFilesProvider else {
+    guard let mainFilesProvider = await mainFilesProvider.value else {
       return nil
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

rintaro (approved):

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -380,11 +380,11 @@ package actor BuildServerManager: QueueBasedMessageHandler {
   }
 
   package func mainFilesProvider<T: MainFilesProvider>(as: T.Type) async -> T? {
-    guard let mainFilesProvider = mainFilesProvider else {
+    guard let mainFilesProvider = await mainFilesProvider.value else {
```
rintaro, line 383: This will trap if `mainFilesProvider` was `nil` as it's IUO, but I assume it's fine because it's always initialized in `init`.

ahoppen: Yes exactly. All the IUO members in SourceKit-LSP should be guaranteed to be initialized in `init`.

PR #2417, opened 2026-01-04
Don’t implicitly cancel requests when an edit is detected

### LINKED ISSUE #1650: Inspect `ClientCapabilities.general.staleRequestSupport` to decide whether to cancel requests on edit
I just realized that there is a `staleRequestSupport` section in the client capabilities in LSP:

```ts
/**
 * Client capability that signals how the client
 * handles stale requests (e.g. a request
 * for which the client will not process the response
 * anymore since the information is outdated).
 *
 * @since 3.17.0
 */
staleRequestSupport?: {
  /**
   * The client will actively cancel the request.
   */
  cancel: boolean;

  /**
   * The list of requests for which the client
   * will retry the request if it receives a
   * response with error code `ContentModified``
   */
   retryOnContentModified: string[];
}
```

VS Code sends the following options

```json
"staleRequestSupport": {
    "cancel": true,
    "retryOnContentModified": [
        "textDocument/semanticTokens/full",
        "textDocument/semanticTokens/range",
        "textDocument/semanticTokens/full/delta"
    ]
}
```

I am still unsure how to interpret the `retryOnContentModified` section because the `ContentModified` error code is described as follows

```ts
/**
 * The server detected that the content of a document got
 * modified outside normal conditions. A server should
 * NOT send this error code if it detects a content change
 * in it unprocessed messages. The result even computed
 * on an older state might still be useful for the client.
 *
 * If a client decides that a result is not of any use anymore
 * the client should cancel the request.
 */
export const ContentModified: integer = -32801;
```

which indicates that LSP servers should not cancel requests when they detect an edit to the document, contradicting the comment in `staleRequestSupport` IMO. The commit that introduced `staleRequestSupport` also isn’t very insightful: https://github.com/microsoft/language-server-protocol/commit/8aec48a51caf88e5c58cb082828abe028eb08396

Based on VS Code’s runtime behavior, I guess that it’s intended as follows:
- `staleRequestSupport.cancel: true` indicates that the LSP server can generally rely on the client cancelling stale requests and doesn’t need to do any implicit cancellation
- `staleRequestSupport.retryOnContentModified` means that there is a list of requests that the client does not cancel. For these, it relies on the server replying with `ContentModified` if it detects a content change, even though that contradicts the definition of the `ContentChange` error code.

So, maybe the correct way of implementing implicit cancellation is:
- If the client does not set `staleRequestSupport` in the client capabilities or if `staleRequsetSupport.cancel` is `false`, we should cancel requests on document edits (the behavior we currently get with `cancelTextDocumentRequestsOnEditAndClose: true`). I am not sure if the error code that we return here matters.
- If `staleRequestSupport.cancel` is `true` then only document requests in `staleRequestSupport.retryOnContentModified` should be implicitly cancelled on document edits, returning a `ContentModified` error code.
- I think it’s still valuable to have a `cancelTextDocumentRequestsOnEditAndClose` configuration setting to override the behavior to always cancel requests on edits, forcing the same behavior as if `staleRequestSupport` was not set in the client capabilities. This configuration setting should default to `false`.

### LINKED ISSUE #2136: Different LSP errors for neovim / helix
This is branched from https://github.com/swiftlang/sourcekit-lsp/issues/2117 as suggested by @ahoppen 

Setup:
- extended logging is enabled
- `swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)`
- logging is captured using `sourcekit-lsp diagnose` from `Apple Swift version 6.2-dev (LLVM 81ab6d9f7e4810f, Swift 9cc1947527bacea)`
- the issue can be reproduced on `neovim` or `helix editor` but oddly **NOT** on VS code

Notes:
- all the issues seem influenced by the activation of the inlay hints

What happens:
- On `neovim` I get `sourcekit: -32001: sourcekitd request timed out`
- On `helix` I get `protocol error: ServerError(-32800): request cancelled by client`

[log.txt](https://github.com/user-attachments/files/20068621/log.txt)

ahoppen:
Instead, use `staleRequestSupport.retryOnContentModified` as an indicator which requests should return a `ContentModified` error when the document is modified. This fixes https://github.com/swiftlang/sourcekit-lsp/issues/2136 by no longer implicitly cancelling inlayHints in Helix while still implicitly cancelling semantic tokens request, which was the motivation for https://github.com/swiftlang/sourcekit-lsp/pull/1629.

Fixes #1650

### DIFF
--- a/Sources/SKOptions/SourceKitLSPOptions.swift
+++ b/Sources/SKOptions/SourceKitLSPOptions.swift
@@ -388,7 +388,7 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable, LSPAnyCodable {
   public var cancelTextDocumentRequestsOnEditAndClose: Bool? = nil
 
   public var cancelTextDocumentRequestsOnEditAndCloseOrDefault: Bool {
-    return cancelTextDocumentRequestsOnEditAndClose ?? true
+    return cancelTextDocumentRequestsOnEditAndClose ?? false
   }
 
   /// Whether the response for `textDocument/semanticTokens` should include semantic tokens for syntactic and semantic
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -663,10 +663,25 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
   /// - Important: Should be invoked on `textDocumentTrackingQueue` to ensure that new text document requests are
   ///   registered before a notification that triggers cancellation might come in.
   private func cancelTextDocumentRequests(for uri: DocumentURI, reason: ImplicitTextDocumentRequestCancellationReason) {
-    guard self.options.cancelTextDocumentRequestsOnEditAndCloseOrDefault else {
-      return
-    }
+    let staleRequestSupport = self.capabilityRegistry?.clientCapabilities.general?.staleRequestSupport
     for (requestID, requestMethod) in self.inProgressTextDocumentRequests[uri, default: []] {
+      // Implicitly cancel text document requests if:
+      //  - We have enabled implicit text document request cancellation in the SourceKit options
+      //  - The client has indicated that it does not cancel stale requests. Defaults to `true` because this is an
+      //    option introduced in LSP 3.17 and most clients cancel requests diligently without setting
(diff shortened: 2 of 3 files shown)


ahoppen:
@swift-ci Please test

rintaro (approved):

hamishknight (approved):

PR #2418, opened 2026-01-04
Remove `DocCReferenceResolutionService`

ahoppen:
I wanted to fix build warnings  due to the deprecation of `OutOfProcessReferenceResolver.Request` in favor of `RequestV2` but realized that `DocCReferenceResolutionService.resolve` was never called in any of the test cases, so it seems that we can completely remove it, unless I missed something.


ahoppen:
@swift-ci Please test

matthewbastien:
The `DocCReferenceResolutionService` is needed to show the `Topics` at the bottom of the documentation preview (as well as for highlighting links in the documentation). We could omit it for simplicity's sake, but it would mean that the live preview will not match the actual generated documentation.

Here's what the live preview looks like in VS Code before this change. I've highlighted the differences in red:
Live preview currently at the tip of main

Here's what it looks like after this change:
Live preview with this PR

matthewbastien:
You bring up a good point about tests. We should add some if we intend on keeping this.

ahoppen:
OK, in that case we shouldn’t remove it. Closing this PR.

PR #2421, opened 2026-01-04
map semantic refactoring kinds to specific LSP code action kinds

### LINKED ISSUE #2382: Support 'refactor.extract.variable' refactoring
**Describe the solution you'd like**

A very common refactoring task is to "extract variable" currently the swift plugin cannot handle it though:

Image

**Additional context**
Extracting a variable from a selection is a common task and we should provide support for it

loveucifer:
## Summary

fixes #2382 

Maps SourceKit semantic refactoring action strings to specific LSP `CodeActionKind` values instead of using the generic `refactor` kind. This enables standard refactoring keybindings and code action filtering to work properly for Swift.

## Images After The Change
test1
test2

## Changes

- Added `lspKind` computed property to `SemanticRefactorCommand` that maps:
  - Actions containing `.extract.` → `refactor.extract`
  - Actions containing `.inline.` → `refactor.inline`  
  - Others → `refactor` (fallback)
- Updated `retrieveRefactorCodeActions` to use the new `lspKind` property


ahoppen:
@swift-ci Please test

PR #2422, opened 2026-01-05
Shut down language services when a workspace is closed

### LINKED ISSUE #2211: Shut down language services when a workspace is closed
We currently only ever add items to `SourceKitLSPServer.languageServices`. But when a workspace folder is closed (`workspace/didChangeWorkspaceFolders`), we could shut down language service that can only handle that workspace, eg. terminating a `clangd` instance that we spawned for that workspace folder.

loveucifer:
Shut down language services when a workspace is closed

closes #2211

- added allLanguageServices property to Workspace to get all services it references
- added shutdownOrphanedLanguageServices to clean up services no longer in use
- when workspace folders are removed, we now shut down their associated language services
- this properly terminates clangd and other language server processes when workspaces close

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1561,7 +1561,7 @@ extension SourceKitLSPServer {
     }
 
     // Shut down orphaned services in a background task to avoid blocking other requests.
-    
+
     if !orphanedServices.isEmpty {
       Task {
         for service in orphanedServices {


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1529,6 +1526,45 @@ extension SourceKitLSPServer {
...
+      if remainingServices.count != services.count {
+        languageServices[serviceType] = remainingServices.isEmpty ? nil : remainingServices
+      }
+    }
+
+    // Shut down orphaned services
+    for service in orphanedServices {
+      logger.info("Shutting down orphaned language service: \(type(of: service))")
+      await service.shutdown()
+    }
```
ahoppen: We should be able to shut down these language services on a background tasks. Because `DidChangeWorkspaceFolders` is a global configuration change request (see `MessageHandlingDependencyTracker`), we won’t be able to handle any other requests until the language service has shut down at the moment. I’m not sure how long clangd takes to shut down but being conservative here would be good.

loveucifer: got it

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,160 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+    let workspaceBUri = DocumentURI(project.scratchDirectory.appending(component: "WorkspaceB"))
+    project.testClient.send(
+      DidChangeWorkspaceFoldersNotification(
+        event: WorkspaceFoldersChangeEvent(removed: [WorkspaceFolder(uri: workspaceBUri)])
+      )
+    )
+    _ = try await project.testClient.send(SynchronizeRequest())
+
+  
+    try await Task.sleep(for: .seconds(3))
```
ahoppen: Tests that rely on `sleep` calls tend to be brittle or slow down test execution unnecessarily. Is there anything in particular you need to sleep for. There’s `repeatUntilExpectedResult` if that’s what you need. Alternatively you might want to use `self.expectation`.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,160 @@ final class WorkspaceTests: SourceKitLSPTestCase {
     )
     XCTAssertEqual(outputPaths.outputPaths.map { $0.suffix(13) }.sorted(), ["FileA.swift.o", "FileB.swift.o"])
   }
+
+  func testOrphanedClangLanguageServiceShutdown() async throws {
+    // test that when we remove a workspace, the ClangLanguageService for that workspace is shut down.
+    //  verify this by checking that the language service is removed from the server's languageServices.
```
ahoppen:
```suggestion
    // verify this by checking that the language service is removed from the server's languageServices.
```

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,160 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+        event: WorkspaceFoldersChangeEvent(removed: [WorkspaceFolder(uri: workspaceBUri)])
+      )
+    )
+    _ = try await project.testClient.send(SynchronizeRequest())
+
+  
+    try await Task.sleep(for: .seconds(3))
+
+
+    let workspaceAfterRemoval = await project.testClient.server.workspaceForDocument(uri: mainUri)
```
ahoppen, line 1448: Could you check that we indeed shut down `clangd` by sending it a `ShutdownRequest`? You should be able to do so using `Hooks.preForwardRequestToClangd`.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,160 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+    if let workspaceUri = workspaceAfterRemoval?.rootUri?.fileURL?.lastPathComponent {
+      XCTAssertFalse(workspaceUri == "WorkspaceB", "WorkspaceB should have been removed")
+    }
```
ahoppen:
```suggestion
    XCTAssertNotEqual(try XCTUnwrap(workspaceAfterRemoval?.rootUri?.fileURL?.lastPathComponent) == "WorkspaceB", "WorkspaceB should have been removed")
```

So we don’t skip this assertion if `workspaceUri` is `nil`. Also, I would prefer to have `XCTAssertEqual` for the value it now has so this check doesn’t accidentally pass because of a typo.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,160 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+
+  
+    let initialHover = try await project.testClient.send(
+      HoverRequest(textDocument: TextDocumentIdentifier(libBUri), position: Position(line: 1, utf16index: 14))
+    )
+    XCTAssertNotNil(initialHover, "Should get hover response for LibB.swift")
+
+    // close the document in WorkspaceB
+    project.testClient.send(DidCloseTextDocumentNotification(textDocument: TextDocumentIdentifier(libBUri)))
+    _ = try await project.testClient.send(SynchronizeRequest())
```
ahoppen: I don’t think we need this synchronize request. `DidChangeWorkspaceFoldersNotification` is a global configuration change and will thus wait for `textDocument/didClose` to have been processed.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+    // open a .c file in WorkspaceB to launch clangd
+    let (mainUri, _) = try project.openDocument("main.c")
+
+    // send a request to ensure clangd is up and running
+    _ = try await project.testClient.send(
+      DocumentSymbolRequest(textDocument: TextDocumentIdentifier(mainUri))
+    )
+
+    // get the language service for WorkspaceB before closing
+    let clangdServerBeforeClose = try await project.testClient.server.primaryLanguageService(
```
ahoppen: Technically this is the `ClangLanguageService`, which holds the reference to `clangd`. This confused me for a short bit, so I’d just call it `clangLanguageServiceBeforeClose`. Same for all other `clangd*` variables.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+    project.testClient.send(DidCloseTextDocumentNotification(textDocument: TextDocumentIdentifier(mainUri)))
+
+    // remove WorkspaceB
+    let workspaceBUri = DocumentURI(project.scratchDirectory.appending(component: "WorkspaceB"))
+    project.testClient.send(
+      DidChangeWorkspaceFoldersNotification(
+        event: WorkspaceFoldersChangeEvent(removed: [WorkspaceFolder(uri: workspaceBUri)])
+      )
+    )
+    _ = try await project.testClient.send(SynchronizeRequest())
```
ahoppen, line 1444: I don’t think we need to synchronize here if we wait for clangd’s shutdown next.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+      .c,
+      in: unwrap(project.testClient.server.workspaceForDocument(uri: dummyUri))
+    )
+
+
+    XCTAssertFalse(clangdServerBeforeClose === clangdServerForWorkspaceA, "WorkspaceB's clangd should have been shut down and a new one created for WorkspaceA")
+  }
+
+  func testOrphanedSwiftLanguageServiceShutdownAndRelaunch() async throws {
+
```
ahoppen: Superfluous newline

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+        return [
+          WorkspaceFolder(uri: DocumentURI(scratchDir.appending(component: "WorkspaceA"))),
+          WorkspaceFolder(uri: DocumentURI(scratchDir.appending(component: "WorkspaceB"))),
+        ]
+      }
+    )
+
+
+    let (libBUri, _) = try project.openDocument("LibB.swift")
+
```
ahoppen: Superfluous double newline here (twice)

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+  
+    let initialHover = try await project.testClient.send(
+      HoverRequest(textDocument: TextDocumentIdentifier(libBUri), position: Position(line: 1, utf16index: 14))
```
ahoppen: Instead of hard-coding line + column, you should be able to use `project.position(of:in:)` with a position marker.

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+    let workspaceBUri = DocumentURI(project.scratchDirectory.appending(component: "WorkspaceB"))
+    project.testClient.send(
+      DidChangeWorkspaceFoldersNotification(
+        event: WorkspaceFoldersChangeEvent(removed: [WorkspaceFolder(uri: workspaceBUri)])
+      )
+    )
+    _ = try await project.testClient.send(SynchronizeRequest())
+
+    //  orphaned service to be shut down in the background
+    try await Task.sleep(for: .milliseconds(500))
```
ahoppen: Similar to my comment from clangd here: Can we implement this without a sleep call?

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,166 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+
+    // open a file in WorkspaceA
+    let (libAUri, positions) = try project.openDocument("LibA.swift")
+
+    // verify that the language service in WorkspaceA still works correctly
+    let hover = try await project.testClient.send(
+      HoverRequest(textDocument: TextDocumentIdentifier(libAUri), position: positions["1️⃣"])
+    )
+    XCTAssertNotNil(hover, "Should still get hover response after removing WorkspaceB")
```
ahoppen, line 1548: Do we get a new `SwiftLanguageService` here? Could we have the same check as for `ClangdLanguageService` here? Because if we don’t, we would have removed the notification handlers from sourcekitd without creating new ones by creating a new `SwiftLanguageService`.

loveucifer: It looks like SwiftLanguageService instances are shared across workspaces (unlike ClangLanguageService which is workspace-specific). Should I loook into this further or address this in some other pr ?

loveucifer: oh i forgot to add I implemented the check you suggested, and the test failed as suspected :/

ahoppen: > It looks like SwiftLanguageService instances are shared across workspaces

Yes, that’s correct. sourcekitd is global and thus all instances are shared. It doesn’t really make sense to shut down the `SwiftLanguageService` for that reason. It will just re-launch when you open the next Swift file. Thinking about this a bit more, it might be easiest to add a static `isImmortal` property to the language services and don’t shut down language services that are marked as immortal (and add a few lines of documentation why `SwiftLanguageService` is immortal).

Tests/SourceKitLSPTests/WorkspaceTests.swift
```diff
@@ -1386,6 +1386,181 @@ final class WorkspaceTests: SourceKitLSPTestCase {
...
+          WorkspaceFolder(uri: DocumentURI(scratchDir.appending(component: "WorkspaceA"))),
+          WorkspaceFolder(uri: DocumentURI(scratchDir.appending(component: "WorkspaceB"))),
+        ]
+      },
+      hooks: Hooks(preForwardRequestToClangd: { request in
+        if request is ShutdownRequest {
+          clangdReceivedShutdown.fulfill()
+        }
+      }),
+      usePullDiagnostics: false
```
ahoppen: Is it important to use push diagnostics here? I don’t see any diagnostic related assertions here, so pull diagnostics should be just fine as well, right?

ahoppen (commented):
Very nice 🤩 

Just one small nitpick left

ahoppen:
@swift-ci Please test

loveucifer:
Hmm tests failed ig

ahoppen:
Ah, they failed because the source files aren’t formatted properly. 
Could you [run swift-format](https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting) on your changes?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Thanks for all the changes you’ve made over the last weeks @loveucifer ❤️ I think with your record, you qualify for the [Member](https://www.swift.org/contributing/#member) role in the swiftlang organization, which means that you’d be able to also trigger CI yourself. If you want to, I’d encourage you to apply for that role.

loveucifer:
> Thanks for all the changes you’ve made over the last weeks @loveucifer ❤️ I think with your record, 
>
Oh no way !! thanksss soo muchh , thankss for reviewing all my commits , hope i can help more in the future :D

PR #2423, opened 2026-01-05
Index targets top-down instead of bottom up

ahoppen:
- **Explanation**: We originally started preparing and indexing the low-level targets in the hope that we can unblock the files of those targets as early as possible for indexing and thus get best indexing performance. Revisiting this design decision, we can actually get a ~2x speed up (SourceKit-LSP gets indexed in 2:45 instead of 5:49 minutes on my machine) in initial indexing performance by preparing indexing top-level targets first. This speed-up is for two reasons:
  - When preparing a top-level target, all of its dependencies are implicitly also prepared. That means that we don’t need to send any prepare requests to the build server for those targets, which is very advantages for SwiftPM projects, in which the constant cost in the order of 500ms (based on my memory) even for a null build.
  - The underlying build server is able to prepare dependencies of the top-level target in parallel instead of implicitly serializing them by preparing all targets in the package in dependency order.
- **Scope**: Background indexing
- **Issue**: n/a
- **Risk**: Targets may not get prepared properly but I don’t see this happening and we should catch any issues like this in living on
- **Testing**: Added/updates tests


ahoppen:
@swift-ci Please test

ahoppen:
Closing in favor of https://github.com/swiftlang/sourcekit-lsp/pull/2684.

PR #2426, opened 2026-01-05
Report diagnostics for script files with shebangs based on fallback build settings

### LINKED ISSUE #2383: Swift scripts update errors and warnings only after running
**Describe the bug**

Starting with version 2.4.0, warnings and errors in Swift scripts (standalone files) only update after running the `Swift: Run Swift Script` command but not after the file is modified or saved. In version 2.2.0, they update immediately when the file is modified.

**To Reproduce**

1. Create a new file `Test.swift`
2. Write some Swift code to the file that produces an error or warning, then save the file
```swift
let x: Int = "" // expected error: `Cannot convert value of type 'String' to specified type 'Int'`
```
3. The error only appears after running `Swift: Run Swift Script`
4. Replace `""` with `0`, then save the file again
5. The error only disappears after running `Swift: Run Swift Script` again

**Environment**
 - OS: macOS Tahoe 26.0.1 (25A362)
 - Swift version: `swift-driver version: 1.127.14.1 Apple Swift version 6.2.1 (swiftlang-6.2.1.4.8 clang-1700.4.4.1)`
 - Visual Studio Code version: 1.105.1
 - vscode-swift version: 2.12.0

ahoppen:
If a file has a shebang, we know that it can be executed on its own and thus we can produce semantic diagnostics even with fallback build settings.

Fixes #2383

### DIFF
--- a/Sources/SwiftLanguageService/DiagnosticReportManager.swift
+++ b/Sources/SwiftLanguageService/DiagnosticReportManager.swift
@@ -78,7 +78,12 @@ actor DiagnosticReportManager {
       }
     }
     let reportTask: ReportTask
-    if let buildSettings, !buildSettings.isFallback {
+    // If we don't have build settings or we only have fallback build settings, sourcekitd won't be able to give us
+    // accurate semantic diagnostics.
+    // Fall back to providing syntactic diagnostics from the built-in swift-syntax. That's the best we can do for now.
+    // The only exception is if the file starts with a shebang. In that case we know that the file can be executed on
+    // its own without additional compiler arguments.
+    if let buildSettings, !buildSettings.isFallback || snapshot.text.starts(with: "#!") {
       reportTask = ReportTask {
         return try await self.requestReport(with: snapshot, compilerArgs: buildSettings.compilerArgs)
       }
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -78,18 +78,19 @@ actor DiagnosticReportManager {
       }
     }
     let reportTask: ReportTask
-    if let buildSettings, !buildSettings.isFallback {
+    // If we don't have build settings or we only have fallback build settings, sourcekitd won't be able to give us
+    // accurate semantic diagnostics.
+    // Fall back to providing syntactic diagnostics from the built-in swift-syntax. That's the best we can do for now.
+    // The only exception is if the file starts with a shebang. In that case we know that the file can be executed on
+    // its own without additional compiler arguments.
```
hamishknight, line 85: Strictly speaking you could have something like `#!/usr/bin/swift -language-mode 6`, would it be worth trying to detect that or not?

hamishknight (approved):

Sources/SwiftLanguageService/DiagnosticReportManager.swift
ahoppen: Hmm, we probably should. The correct implementation would be:
- `BuildServerManager` should look at file contents when it synthesizes fallback settings and if it discovers a shebang (a) include compiler arguments from there in the build settings it returns and (b) not mark the returned build settings as fallback
- Now `BuildServerManager` needs to watch this file for changes so that we can reload build settings when they are modified. It would likely be sufficient to only do this when the file has been saved to disk but we’d need to start keeping track which files we have returned fallback build settings for and watch all of them in `BuildServerManager`, which we currently don’t do. This will likely be the tricky part. Let’s see if I have an idea in a few days.

rintaro: Random thoughts:
* It's possible the file has a weird shebang like `#!/opt/bin/myRunner` where `myRunner` is a wrapper of `swift` with extra compiler arguments. So having shebang might not be enough for getting the build arguments for the file.
* `BuildServerManager` might want to delegate the primary `languageService` to translate the shebang line to the build settings. Because it's language/compiler specific.

PR #2432, opened 2026-01-06
Don’t show sourcekitd refactoring actions that have been superseded by swift-syntax refactoring actions

ahoppen:
For now, this is only the action to add thousands separators to integer literals but hopefully this list will grow as we translate more refactoring actions to swift-syntax.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #2433, opened 2026-01-06
Fix #2430: Hide 'Convert to computed property' for Void functions

### LINKED ISSUE #2430: “Convert to computed property” should not be offered when function does not have a return value
When I have the following, we shouldn’t offer the *Convert to computed property* refactoring action because the function returns `Void` / has no return value and thus a computed property does not make sense.

```swift
struct Foo {
  func test() {
  }
}

```

DPrakashhh:
**Summary**
This PR resolves issue #2430 by preventing the "Convert to computed property" code action from being suggested for functions that do not return a meaningful value. Converting a function with no return value to a computed property is semantically incorrect and results in nonsensical refactoring.

Changes
1-**Logic Enhancement**: Updated SyntaxRefactoringCodeActionProvider.swift to include a guard statement that checks for void return types before offering the refactoring.
2-**Void Detection Helper**: Added a private "isVoidReturnType" helper that robustly identifies:
2.1-**Implicit Void**: Functions with no return clause.
2.2-**Explicit Void**: Functions returning Void.
2.3-**Empty Tuples**: Functions returning ().
3-**Normalization:** Used whitespace filtering to ensure the check is resistant to formatting variations like -> Void or -> ( ).
4-**Regression Tests**: Added three comprehensive test cases to CodeActionTests.swift to ensure these scenarios are permanently covered.

**Verification Results**
1-**Automated Tests**: Ran swift test --filter CodeActionTests. All 37 tests passed, including the new negative test cases.
2-**Manual Verification**: Confirmed in VS Code that the refactoring lightbulb is correctly hidden for func test() {} but remains available for functions with actual return types like func test() -> String.

Screenshot 2026-01-06 at 1 53 05 PM

Screenshot 2026-01-06 at 1 53 36 PM

Open for any changes :)

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
+++ b/Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
@@ -121,7 +121,7 @@ extension ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringCodeA
       ofType: FunctionDeclSyntax.self,
       stoppingIf: { $0.is(CodeBlockSyntax.self) || $0.is(MemberBlockSyntax.self) }
     )
-    guard let functionDecl, !isVoidReturnType(functionDecl.signature.returnClause?.type) else {
+    guard let functionDecl, !(functionDecl.signature.returnClause?.type.isVoid ?? true) else {
       return nil
     }
     return functionDecl
@@ -183,19 +183,17 @@ extension [SourceEdit] {
   }
 }
 
-// MARK: - Helper Functions
+// MARK: - Helper Extensions
 
-/// Checks if a return type is effectively Void (no return type, explicit Void, or empty tuple)
-func isVoidReturnType(_ returnType: TypeSyntax?) -> Bool {
-  guard let returnType else {
-    return true
-  }
-  switch returnType.as(TypeSyntaxEnum.self) {
-  case .identifierType(let identifierType) where identifierType.name.text == "Void":
-    return true
-  case .tupleType(let tupleType) where tupleType.elements.isEmpty:
-    return true
-  default:
-    return false
+private extension TypeSyntax {
+  var isVoid: Bool {
+    switch self.as(TypeSyntaxEnum.self) {
+    case .identifierType(let identifierType) where identifierType.name.text == "Void":
+      return true
+    case .tupleType(let tupleType) where tupleType.elements.isEmpty:
+      return true
+    default:
+      return false
+    }
   }
 }


Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -178,3 +190,18 @@ extension [SourceEdit] {
     )
   }
 }
+
+// MARK: - Helper Functions
+
+/// Checks if a return type is effectively Void (no return type, explicit Void, or empty tuple)
+private func isVoidReturnType(_ returnType: TypeSyntax?) -> Bool {
```
ahoppen: You should be able to implement this the same way as follows. This will also correctly handle types such as `( /* comment */ )`. Not that they make much sense but still.

https://github.com/swiftlang/swift-syntax/blob/6c4ac31505a494531ddc885aea14186a50426a0c/Sources/SwiftRefactor/SyntaxUtils.swift#L35-L43

Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -117,10 +118,21 @@ extension ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringCodeA
...
-      stoppingIf: { $0.is(CodeBlockSyntax.self) || $0.is(MemberBlockSyntax.self) }
-    )
+    guard
+      let functionDecl = scope.innermostNodeContainingRange?.findParentOfSelf(
+        ofType: FunctionDeclSyntax.self,
+        stoppingIf: { $0.is(CodeBlockSyntax.self) || $0.is(MemberBlockSyntax.self) }
+      )
+    else {
+      return nil
+    }
```
ahoppen: I dislike guard bindings that span multiple lines. Let’s write this as

```swift
      let functionDecl = scope.innermostNodeContainingRange?.findParentOfSelf(
        ofType: FunctionDeclSyntax.self,
      )
      guard let functionDecl ...

File: Tests/SourceKitLSPTests/CodeActionTests.swift [hunk 3]
```diff
@@ -1238,6 +1238,96 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    guard case .codeActions(let codeActions) = result else {
+      XCTFail("Expected code actions")
+      return
+    }
+
+    let convertToComputedPropertyAction = codeActions.first { $0.title == "Convert to computed property" }
+    XCTAssertNil(
+      convertToComputedPropertyAction,
+      "Convert to computed property should not be offered for implicit Void"
+    )
```
ahoppen: Let’s explicitly match the list of returned code actions. This could also pass if there was a typo in the title of the test or if we renamed the refactoring action. Ie. something like


```swift
XCTAssertEqual(codeActions.map(\.title), [])
```

Also, I think you might be able to use `assertCodeActions` to shorten the test slightly.

Same for the other test.

DPrakashhh:
On it :)

DPrakashhh:
Thanks for the feedback, I have updated the PR with a new commit that switches to the `TypeSyntax` extension logic you suggested. I also cleaned up the guard bindings and refactored the tests to use `assertCodeActions` to keep things concise. Ready for another look!

Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
```diff
@@ -178,3 +182,35 @@ extension [SourceEdit] {
     )
   }
 }
+
+// MARK: - Helper Functions
+
+/// Checks if a return type is effectively Void (no return type, explicit Void, or empty tuple)
+func isVoidReturnType(_ returnType: TypeSyntax?) -> Bool {
```
ahoppen: Leftover implementation?

ahoppen (commented):
Just some leftover code, otherwise looks good to me 👍🏽

Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
DPrakashhh: Sorry about that! I had actually removed that extension on my device, but it looks like it didn't make it into the last push. My mistake—I have pushed the cleanup now so only the active helper function remains. Thanks for double-checking!

DPrakashhh:
Done!

Sources/SwiftLanguageService/CodeActions/SyntaxRefactoringCodeActionProvider.swift
ahoppen: I personally preferred the version of this function that’s an extension on `TypeSyntax` to have parity with swift-syntax in case we want to move refactoring actions between the two repos in the future.

DPrakashhh: Done !

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

DPrakashhh:
Thank you! for merging it :D

PR #2435, opened 2026-01-06
Fix JSON base indentation in Create Codable structs

PhantomInTheWire:
fixes: https://github.com/swiftlang/sourcekit-lsp/issues/2429

### DIFF
--- a/Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
+++ b/Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
@@ -252,11 +252,21 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
   func testJSONIndentedWith4SpacesButFileWith2Spaces() throws {
     try assertRefactor(
       """
+      func dummy1() {
+        print("a")
+        print("b")
+      }
+
+      func dummy2() {
+        print("c")
+        print("d")
+      }
+
       func test() {
         if true {
           1️⃣{
-                 "a": 1
-             }
+              "a": 1
+          }
         }
       }
       """,
@@ -264,10 +274,10 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
       provider: ConvertJSONToCodableStruct.self,
       expected: [
         SourceEdit(
-          range: AbsolutePosition(utf8Offset: 30)..<AbsolutePosition(utf8Offset: 58),
+          range: AbsolutePosition(utf8Offset: 120)..<AbsolutePosition(utf8Offset: 142),
           replacement: """
             struct JSONValue: Codable {
-                    var a: Double
+                  var a: Double
                 }
             """
         )


Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -99,26 +99,44 @@ package struct ConvertJSONToCodableStruct: EditRefactoringProvider {
...
 
     // Render the top-level object as a struct.
-    let indentation = BasicFormat.inferIndentation(of: syntax)
+    let inferenceNode: Syntax
+    switch preflight {
+    case .closure(let closure): inferenceNode = Syntax(closure)
+    case .endingClosure(let closure, _): inferenceNode = Syntax(closure)
+    case .stringLiteral(let literal, _): inferenceNode = Syntax(literal)
+    }
+    let indentation = BasicFormat.inferIndentation(of: inferenceNode)
```
ahoppen: I think we can use the entire file to infer the indentation. The JSON may be indented by 4 spaces but if the rest of the source file is indented by 2 spaced, we should pick that up.

Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -99,26 +99,44 @@ package struct ConvertJSONToCodableStruct: EditRefactoringProvider {
...
+    // Subsequent lines need to be explicitly indented.
+    let baseIndentation = getBaseIndentation(from: inferenceNode)
+    let indentedDecls = decls.description.split(separator: "\n", omittingEmptySubsequences: false)
+      .enumerated()
+      .map { (index, line) in
+        if index == 0 { return String(line) }
+        return baseIndentation.description + line
+      }
+      .joined(separator: "\n")
+
```
ahoppen: I think you should be able to just set `initialIndentation` in `BasicFormat` above to and then you don’t need to do any string-post processing. Or am I missing something?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1295,6 +1295,44 @@ final class CodeActionTests: SourceKitLSPTestCase {
       []
     }
   }
+  /// Test Create Codable structs JSON base indentation
+  func testJSONCodableCodeActionResultIndentation() async throws {
```
ahoppen: You should be able to use `assertRefactor` in `SyntaxRefactorTests.swift`. The advantage of that is that it doesn’t need to launch a SourceKit-LSP server, which reduces test overhead.

ahoppen (commented):
Thanks for looking at this @PhantomInTheWire!

Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
PhantomInTheWire: no that makes sense i was over complicating things

Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
PhantomInTheWire: that makes sense.

Tests/SourceKitLSPTests/CodeActionTests.swift
PhantomInTheWire: yeah, good catch.

PhantomInTheWire:
it now infers indentation from file context. also I updated the existing tests to expect 4-space indentation (since BasicFormat defaults to 4 spaces when no context implies otherwise)

Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
```diff
@@ -127,12 +138,23 @@ package struct ConvertJSONToCodableStruct: EditRefactoringProvider {
...
         SourceEdit(
           range: literal.endPosition..<literal.endPosition,
-          replacement: "\n" + decls.description
+          replacement: "\n" + indentedDecls
         )
       ]
     }
   }
 
+  /// Get the indentation of the given node from the source file.
```
ahoppen: Would `node.indentationOfFirstLine` also work? I think it does conceptually the same thing (just first line instead of last) and then we don’t need to re-invent logic.

ahoppen (commented):
Just one last small comment (sorry for not noticing this earlier), then it looks good 👍🏽

Sources/SwiftLanguageService/CodeActions/ConvertJSONToCodableStruct.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -247,32 +247,32 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
...
-                var a: Double
-              }
-              """
-          )
-        ]
-      )
-    }
+  }
 
+  func testJSONIndentedWith4SpacesButFileWith2Spaces() throws {
```
ahoppen, line 252: Interesting, why are we now inferring the indentation of the file to be 4 spaces where we previously inferred it to 2 spaces? If there’s a good reason to get the base indentation based on the last line, I’m happy to stick with that, I’d just like to understand why it’s necessary.

PhantomInTheWire: When inferring from just the JSON closure { "a": 1 }, there's no clear indentation pattern, so it defaults to 4 spaces. the number 4 comes from this line in swift-syntax
```
self.indentationWidth = indentationWidth ?? .spaces(4)
```
im not sure if changing it over there is okay since it might break a lot of things downstream, i could ignore BasicFormat and override this but is that a good idea?

ahoppen: So, we should keep that at a default to 4 spaces. But shouldn’t we be inferring the indentation from the entire file? Maybe do we need to add a little more code to the test case to help it infer indentation to 2 spaces from the surroundings, eg. add something like the following at the bottom (not sure how many lines are needed)

```swift
func dumm() {
  print("a")
  print("b")
  print("c")
  print("d")
  print("e")
}
```

PhantomInTheWire: The inferIndentation function needs at least 3 indented lines to reliably infer. Adding more code with consistent 2-space indentation should help it correctly detect the pattern.

PhantomInTheWire: the 1️⃣ is kind of confusing me on the test(i think it should be correct though it looks a little off) but the implementation should work


https://github.com/user-attachments/assets/3047a0ac-dbf6-4596-8119-2c8a6ca489be

ahoppen (approved):
Nice. Let’s get this in 🚀

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
ahoppen: > the 1️⃣ is kind of confusing me on the test

The 1️⃣ is just a position marker, it gets removed before the actual test execution. Having these position markers is usually more convenient than having to specify line-column positions, which you then need to count again from the input source file. Or did I misunderstand your question?

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: no as in when a emoji is rendered this way it is kind of annoying to verify the utf8 offset visually and i just have to assume the range the failing test gives me is correct(or sometimes have a small manual script verify it for me) 
image

ahoppen: Ah, I see what you mean. Do you want to open a PR to refactor `assertRefactor` to take an `expected` parameter of type `(_ positions: [String: AbsolutePosition]) -> [SourceEdit]`? Then we can use position markers in the expected ranges.

ahoppen:
@swift-ci Please test Windows

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: sure i think that would be a good follow up pr

ahoppen: Thank you ❤️

Issue #2318, opened 2025-10-06
Support going to definition on an InlayHint


award999:
### Description

The InlayHints the LSP sends are very useful in VSCode, especially in a language like Swift where you don't need to be explicit with you types. What would make things even better would be being able to go to definition with these inlay hints, possibly using the `inlayHint/resolve` request

PR #2436, opened 2026-01-06
add go-to-definition for inlay hints (#2318)

loveucifer:
Implements resolveProvider for inlay hints to enable navigating to type definitions. When an inlay hint showing a type is resolved, the server looks up the type's definition location using cursorInfo and the index.

- store variable position in InlayHint.data for resolution
- add inlayHintResolve to LanguageService protocol
- implement resolve handler using cursorInfo and index lookup
- enable resolveProvider: true in capabilities
- add test for resolve functionality

Addresses #2318

### DIFF
--- a/Sources/SwiftLanguageService/CMakeLists.txt
+++ b/Sources/SwiftLanguageService/CMakeLists.txt
@@ -23,6 +23,7 @@ add_library(SwiftLanguageService STATIC
   FoldingRange.swift
   GeneratedInterfaceManager.swift
   InlayHints.swift
+  InlayHintResolve.swift
   MacroExpansion.swift
   OpenInterface.swift
   PlaygroundDiscovery.swift


Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,36 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+    XCTAssertNotNil(typeHint.data, "Expected type hint to have data for resolution")
+
+    // resolve the hint to get type location
+    let resolvedHint = try await project.testClient.send(InlayHintResolveRequest(inlayHint: typeHint))
+
+    if case .parts(let parts) = resolvedHint.label {
+      XCTAssertEqual(parts.count, 1)
+      XCTAssertNotNil(parts.first?.location, "Expected label part to have location for go-to-definition")
+    } else if case .string = resolvedHint.label {
+
```
ahoppen: I think the test only passes because we get into this case because there are no parts. AFAIK the problem is that the typeUSR field in the cursor info request isn’t actually a valid USR (looks like it hasn’t been used in a while). Furthermore, a purely USR-based implementation won’t be able to through the fallback locations in `indexBasedDefinition` and `definitionLocations`. In particular, it won’t be able to jump to a generated interface or jump to a definition if background indexing is disabled.

I think the correct implementation would be to extend the cursorinfo request to return line, column, filepath, module name and USR of the declaration’s type (maybe I forgot something else as well). The place to start debugging in order to do this would likely be https://github.com/swiftlang/swift/blob/f4e78a61dfbbaae7f1f03861d7ea8077546a4146/tools/SourceKit/tools/sourcekitd/lib/Service/Requests.cpp#L2736

The good news is that https://github.com/swiftlang/sourcekit-lsp/issues/548 would fall out for (almost) free and you would be able to fix the oldest open issue in this repo 😉

loveucifer: hmm seems like im having a hard time figuring out how to go forward with this ( does this require me to work within the compiler i mean swift/swiftlang and handle edgecases and tests for those as well because thats gotta be long ig )

loveucifer: if not ,i spent yesterday looking up fixes and found one ( idk if it works or has any more edgecases ) we could look up the type by its name string (e.g., "MyType") instead of the broken USR; it handles most casesig  and would get this feature working immediately without me having to go through the compiler repo , that might take this pr a while , i mean i can do it but it will just take time maybe 2-4 days ig atleast to get fmailiar with the repo

ahoppen: > does this require me to work within the compiler i mean swift/swiftlang

Yes, the entry barrier is definitely a little higher there but if you’re in for a challenge, I think you’d be up for it. If you get stuck somewhere, also feel free to just open a PR with how far you got and we can help you go on. I’d prefer to proper compiler-backed solution over something hacky in SourceKit-LSP because it will likely be a lot more error-resistant and easier to maintain.

loveucifer: Okey ,  i will try to learn the repo for a few days and try to fix the issue , should i convert this to a draft PR in the meantime

ahoppen: Haha, yes at some point you will need to know about sourcekitd when contributing to SourceKit-LSP. It’s where most of the actual semantic understanding meat is implemented. And because it’s a little harder to set up, those are the issues that don’t get tackled as quickly 😉

loveucifer: will try and do my best :)

loveucifer:
new commit requires swiftlang/swift#86381  for compiler changes

ahoppen:
Very nice, I was expecting you to take longer to create a sourcekitd PR, good job 🌟  

Let’s wait for https://github.com/swiftlang/swift/pull/86381 to be approved and then I’ll review this PR in detail. Looked good from a first glance.

loveucifer:
> Very nice, I was expecting you to take longer to create a sourcekitd PR, good job 🌟
>
yess i thought so too but i did have some previous experience with c++ while wriitng my game engine and I spent some time yesterday and today reading through the repo and thought might as well do it and learn more as i go on :D , thanks for your kind wordsss ❤️

loveucifer:
following the discussion in https://github.com/swiftlang/swift/pull/86381. with hamishknight , ultimately reached the conclusion that we dont want any compiler changes , i am sending a pr that oughtta fix this but if not back to the compiler repo again we go :D

loveucifer:
as mentinoed above, i discovered that no changes are needed in the Swift repo. The existing cursorInfo already supports looking up types by USR via key.usr.
one of the key insight was that key.typeusr returned by cursorInfo is a mangled type name (e.g., $sSiD for Int), not a declaration USR. to use it for index lookup or cursorInfo queries
so :
replacing $s prefix with s: (e.g., $sSi → s:Si)
Strip trailing [D]suffix (type descriptor marker)
This converts the mangled type to a proper declaration USR that works with:

Index lookups via primaryDefinitionOrDeclarationOccurrence(ofUSR:)
cursorInfo queries via key.usr parameter
Implementation:

added convertMangledTypeToUSR() helper for the conversion
lookupTypeDefinitionLocation() tries index lookup first (fast path), falls back to cursorInfo for types not in index
should ideally work for both local project types and SDK/external types

also i closed the Swift repo PR #86381 as it's no longer needed.

Sources/SourceKitD/sourcekitd_uids.swift.gyb
```diff
@@ -111,9 +111,7 @@
     # Maintained from initializeService in Requests.cpp
     KIND('SemaEnabledNotification', 'source.notification.sema_enabled'),
     KIND('DocumentUpdateNotification', 'source.notification.editor.documentupdate'),
-
-    # Used exclusively within the SourceKit Plugin
-    KIND('DiagRemark', 'source.diagnostic.severity.remark'),
+    # Note: DiagRemark was moved to UID_KINDS in UIDs.py
```
ahoppen: I don’t think this comment is providing any value. Let’s just remove it.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -234,6 +234,7 @@ package protocol LanguageService: AnyObject, Sendable {
   func definition(_ request: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse?
 
   func declaration(_ request: DeclarationRequest) async throws -> LocationsOrLocationLinksResponse?
+  func typeDefinition(_ request: TypeDefinitionRequest) async throws -> LocationsOrLocationLinksResponse?
```
ahoppen: I think `typeDefinition` doesn’t have any tests at the moment yet. I would suggest we move its implementation to a follow-up PR and focus this one on the inlay hints.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -471,6 +477,11 @@ package extension LanguageService {
     throw ResponseError.requestNotImplemented(InlayHintRequest.self)
   }
 
+  func inlayHintResolve(_ req: InlayHintResolveRequest) async throws -> InlayHint {
+    // default: return hint unchanged
```
ahoppen: I don’t think this comment provides much value, so I’d just remove it.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1098,7 +1102,7 @@ extension SourceKitLSPServer {
     let inlayHintOptions =
       await registry.clientHasDynamicInlayHintRegistration
       ? nil
-      : ValueOrBool.value(InlayHintOptions(resolveProvider: false))
+      : ValueOrBool.value(InlayHintOptions(resolveProvider: true))
```
ahoppen: Could you also set the resolve provider to `true` in https://github.com/swiftlang/sourcekit-lsp/blob/29a0bdd15f360539f1f456fec943acd44654f860/Sources/SwiftLanguageService/SwiftLanguageService.swift#L416?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1904,6 +1906,24 @@ extension SourceKitLSPServer {
     return try await languageService.inlayHint(req)
   }
 
+  func inlayHintResolve(
+    request: InlayHintResolveRequest
+  ) async throws -> InlayHint {
+    // inlay hints store the uri in data for resolution
+    // extract uri from the lspany dictionary
```
ahoppen: Similar comment on the comment here. This exactly explains what’s being done below, so I don’t think it provides much value. In general, I’m a fan of comments that describe _why_ something is done, not _what_ is being done.

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -207,4 +207,48 @@ extension SwiftLanguageService {
...
+  }
+
+  /// get cursor info for a type by looking up its USR.
+  /// this takes a mangled type (from `key.typeusr`) and converts it to a proper USR
+  /// (by replacing `$s` prefix with `s:`), then queries cursorInfo with that USR.
+  ///
+  /// - parameters:
+  ///   - mangledType: the mangled type string (e.g., `$sSS` for String)
+  ///   - uri: document URI for context (used to get compile command)
+  /// - returns: cursorInfo for the type declaration, or nil if not found
```
ahoppen: Let’s just assume that this function takes a proper USR (that’s what sourcekit promises to give us based on the key in the response) and consider the need for `convertMangledTypeToUSR` a hack that we don’t need to explain too much in the doc comment. Otherwise, we generally start doc comments with uppercase letters, just like English sentences.

```suggestion
  /// Get cursor info for a type by looking up its USR.
  ///
  /// - parameters:
  ///   - mangledType: The mangled name of the type 
  ///   - uri: Document URI for context (used to get compile command)
  /// - returns: Cursor Info for the type declaration, or `nil` if not found
```

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -207,4 +207,48 @@ extension SwiftLanguageService {
       additionalParameters: appendAdditionalParameters
     )
   }
+
+  /// converts a mangled type string to a USR format.
+  /// mangled types start with `$s` while USRs start with `s:`.
+  /// for instance `$sSS` becomes `s:SS` (for String type).
```
ahoppen: Could we file an issue in github.com/swiftlang/swift that the cursor info request should return a proper USR here, so we can reference that issue and hopefully get rid of this hack when it’s no longer needed?

```suggestion
  /// Because of <insert issue in swiftlang/swift here> sourcekitd returns a mangled name instead of a USR as the type 
  /// USR. Work around this by replacing mangled names (starting with `$s`) to a USR, starting with `s:`.
  // TODO: Remove once <swiftlang issue> is fixed
```

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -207,4 +207,48 @@ extension SwiftLanguageService {
...
+  /// this takes a mangled type (from `key.typeusr`) and converts it to a proper USR
+  /// (by replacing `$s` prefix with `s:`), then queries cursorInfo with that USR.
+  ///
+  /// - parameters:
+  ///   - mangledType: the mangled type string (e.g., `$sSS` for String)
+  ///   - uri: document URI for context (used to get compile command)
+  /// - returns: cursorInfo for the type declaration, or nil if not found
+  func cursorInfoFromTypeUSR(
+    _ mangledType: String,
+    in uri: DocumentURI
```
ahoppen: I think this should take a `DocumentSnapshot` instead of a `DocumentURI`. That way the caller can make sure that the snapshot matches the version of the document they are interested in. 

In the inlay type resolve request, we should include the document version in the `data`. If the document version of a resolve request isn’t the latest document version, we should fail the request because we would now be resolving the inlay hint against a different version of the document than the one that originally created it and types might have changed since.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    // step 2: Convert mangled type to proper USR
+    // The typeUsr is a mangled type like "$s4test6MyTypeVD" for struct MyType
+    // To get the declaration USR, we need to:
+    // 1. Replace "$s" prefix with "s:"
+    // 2. Strip the trailing "D" which is a mangling suffix (type descriptor)
+    var mangledName = typeUsr
+    if mangledName.hasPrefix("$s") {
+      mangledName = "s:" + mangledName.dropFirst(2)
+    }
```
ahoppen: Isn’t this duplicated with `convertMangledTypeToUSR`. FWIW, I would prefer to keep this transformation closer to where we extract the `typeUSR` field from the cursor info response, if possible, because that’s where the bug is that we need to work around.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -0,0 +1,133 @@
...
+    // step 3: Try index lookup first (works well for local and external types)
+    if let workspace = await sourceKitLSPServer?.workspaceForDocument(uri: uri),
+      let index = await workspace.index(checkedFor: .deletedFiles),
+      let occurrence = index.primaryDefinitionOrDeclarationOccurrence(ofUSR: usr)
+    {
+      let definitionUri = DocumentURI(filePath: occurrence.location.path, isDirectory: false)
+      let definitionPosition = Position(
+        line: occurrence.location.line - 1,
+        utf16index: occurrence.location.utf8Column - 1
+      )
```
ahoppen: Interpreting UTF-8 as UTF-16 columns isn’t correct if there are eg. emojis in the line. There are functions on `DocumentSnapshot` to convert positions.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -0,0 +1,133 @@
...
+    // 1. Replace "$s" prefix with "s:"
+    // 2. Strip the trailing "D" which is a mangling suffix (type descriptor)
+    var mangledName = typeUsr
+    if mangledName.hasPrefix("$s") {
+      mangledName = "s:" + mangledName.dropFirst(2)
+    }
+    // Strip trailing 'D' (type descriptor suffix in mangling)
+    if mangledName.hasSuffix("D") {
+      mangledName = String(mangledName.dropLast())
+    }
```
ahoppen: Uh, this seems iffy to me. I really would prefer if we didn’t start interpreting mangled names, I just feel like there lie dragons that way (I’m somewhat fine with the `$s` to `s:` replacement because the difference between USR and mangled names is pretty much just defined by the prefix). 

Would you be able to handle USRs with a `D` suffix in sourcekitd as well. And if you figure out what exactly the `D` suffix is for along the way, I’d be interested to hear that.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -0,0 +1,133 @@
...
+    let usr = mangledName
+
+    // step 3: Try index lookup first (works well for local and external types)
+    if let workspace = await sourceKitLSPServer?.workspaceForDocument(uri: uri),
```
ahoppen: This won’t handle jumping to the declaration of a type in a generated interface. We handle this for normal definition requests in `definitionLocations`, called from `indexBasedDefinition` (which honestly, is a bit of mis-namer).

I think it would be great if we could use the same infrastructure to resolve generated interfaces in the same way for the inlay hint resolve as well, not sure how much `indexBasedDefinition` would need to be restructured for that.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -11,10 +11,19 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(SourceKitLSP) package import LanguageServerProtocol
+import Foundation
 import SourceKitLSP
 import SwiftExtensions
 import SwiftSyntax
 
+
+package struct InlayHintResolveData: Codable {
```
ahoppen: If you make this conform to `LSPAnyCodable`, you don’t need to do any of the ad-hoc coding where you use this type in the inlay hints data.

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,44 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
   }
+
+  func testInlayHintResolve() async throws {
+    // test that resolving an inlay hint returns label parts with type location
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      struct 1️⃣MyType {}
+      let x2️⃣ = MyType()
+      """
+    )
```
ahoppen: I think this should also work without having an index for the file, ie. by just creating a `TestSourceKitLSPClient` an opening a source file in it. It makes the test slightly faster.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -0,0 +1,133 @@
...
+    {
+      let definitionUri = DocumentURI(filePath: occurrence.location.path, isDirectory: false)
+      let definitionPosition = Position(
+        line: occurrence.location.line - 1,
+        utf16index: occurrence.location.utf8Column - 1
+      )
+      return Location(uri: definitionUri, range: Range(definitionPosition))
+    }
+
+    // Fallback: Try cursorInfo with USR (for types not in index)
```
ahoppen: Even if we have index data, we should prefer data from a type USR cursor info request. The index will always only reflect the state of the document as it was last saved and even then it may be out-of-date. Sourcekitd knows the current in-memory state of the document and is thus more accurate.

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,44 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+    XCTAssertEqual(parts.count, 1, "Expected exactly one label part")
+
+    guard let location = parts.first?.location else {
+      XCTFail("Expected label part to have location for go-to-definition")
+      return
+    }
+
+    XCTAssertEqual(location.uri, project.fileURI)
+    XCTAssertEqual(location.range.lowerBound, project.positions["1️⃣"])
+  }
```
ahoppen, line 325: Could you also add test cases for the following cases:
- The type we want to jump to is defined in a different modules (use a `SwiftPMTestProject` with two targets for that)
- The type we want to jump to is defined in the SDK, eg. a variable of type `String`

Sources/SwiftLanguageService/InlayHintResolve.swift
hamishknight: "D" in this case just means "type mangling", dropping it isn't guaranteed to give you the equivalent decl USR though, e.g you could have generic arguments present. The USR cursor info request should give back the right decl USR to use, I think it would make more sense to use that (though I don't know if an index lookup is still necessary in that case?)

loveucifer:
>  I hope you still learned something worthwhile setting up you compiler development environment 😉
>
Yess totally and i bet it will come of use while fixing other issues ,i have noticed that some of the issues needs me to work within it so its good .

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -207,4 +207,51 @@ extension SwiftLanguageService {
...
+  // TODO: Remove once https://github.com/swiftlang/swift/issues/86432 is fixed
+  private func convertMangledTypeToUSR(_ mangledType: String) -> String {
+    var result = mangledType
+    if result.hasPrefix("$s") {
+      result = "s:" + result.dropFirst(2)
+    }
+    // Strip trailing 'D' (type mangling suffix) to get declaration USR
+    if result.hasSuffix("D") {
+      result = String(result.dropLast())
+    }
```
hamishknight, line 224: You shouldn't need to do this, cursor info should be able to handle a type USR with it

loveucifer: Hmm but the tests fail without the D stripping.

hamishknight: Oh interesting, I saw handling of `Demangle::Node::Kind::Type` in the code and assumed that was referring to `D` but it's actually a different kind. I guess this is okay for now but we really ought to fix this on the sourcekitd side, ideally at the same time as the `$s` -> `s:` change.

loveucifer: okei

ahoppen: Just to add my 2cts, while I’m not a huge fan, I’m fine with having the stripping of `D` as a temporary workaround for now as long as we try to adjust sourcekitd to work without this hack in the future.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -10,11 +10,50 @@
...
+    self.version = version
+  }
+
+  package init?(fromLSPDictionary dictionary: [String: LSPAny]) {
+    guard case .string(let uriString) = dictionary["uri"],
+      let uri = try? DocumentURI(string: uriString),
+      case .dictionary(let posDict) = dictionary["position"],
+      case .int(let line) = posDict["line"],
+      case .int(let character) = posDict["character"],
+      case .int(let version) = dictionary["version"]
```
ahoppen: We can use the fact that `Position` conforms to `LSPAnyCodable`

```suggestion
      case .int(let version) = dictionary["version"],
      let position = Position(fromLSPAny: dictionary["position")
```

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -10,11 +10,50 @@
...
+    self.version = version
+  }
+
+  package func encodeToLSPAny() -> LSPAny {
+    return .dictionary([
+      "uri": .string(uri.stringValue),
+      "position": .dictionary([
+        "line": .int(position.line),
+        "character": .int(position.utf16index),
+      ]),
```
ahoppen:
```suggestion
      "position": position.encodeToLSPAny(),
```

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,118 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+      XCTFail("Expected resolved hint to have label parts, got: \(resolvedHint.label)")
+      return
+    }
+
+    XCTAssertEqual(parts.count, 1, "Expected exactly one label part")
+
+    guard let location = parts.first?.location else {
+      XCTFail("Expected label part to have location for go-to-definition")
+      return
+    }
```
ahoppen: Can be simplified slightly by using `.only`

```suggestion
    guard let location = parts.only?.location else {
      XCTFail("Expected label part to have location for go-to-definition")
      return
    }
```

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,118 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+    let hints = try await testClient.send(request)
+
+    guard let typeHint = hints.first(where: { $0.kind == .type }) else {
+      XCTFail("Expected type hint for String")
+      return
+    }
+
+    // Resolve should not crash, and returns the hint (possibly without location for SDK types in test env)
+    let resolvedHint = try await testClient.send(InlayHintResolveRequest(inlayHint: typeHint))
+    XCTAssertEqual(resolvedHint.kind, .type)
```
ahoppen: We should take the user to the generated interface of the SDK, similar to what we do [here](https://github.com/swiftlang/sourcekit-lsp/blob/29a0bdd15f360539f1f456fec943acd44654f860/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift#L73). This is what I meant by https://github.com/swiftlang/sourcekit-lsp/pull/2436#discussion_r2677776419.

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -267,4 +284,118 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+      let part = parts.first,
+      let location = part.location
+    else {
+      XCTFail("Expected label part to have location for go-to-definition")
+      return
+    }
+
+    // The location should point to LibA/MyType.swift where MyType is defined
+    XCTAssertEqual(location.uri, try project.uri(for: "MyType.swift"))
+    XCTAssertEqual(location.range.lowerBound, try project.position(of: "1️⃣", in: "MyType.swift"))
```
ahoppen: For completeness sake, let’s also check the upper bound

```suggestion
    XCTAssertEqual(location.range, try Range(project.position(of: "1️⃣", in: "MyType.swift")))
```

ahoppen (commented):
Left a few more comments inline. I think the best way forward might be:
- Remove all remainders of the type definition request from this PR
- Get this merged
- Support jumping to a generated interface from inlay hints. I suspect that this might require some careful re-shuffling of code which might be easier to review if done on its own without having to worry the changes that are currently part of this PR. And this PR already provides great value on its own

Also, just want to express how extremely excited I am to use the type definition request once it gets in. I keep wanting that multiple times a week 🙂

loveucifer:
there is already a follow up PR in https://github.com/swiftlang/sourcekit-lsp/pull/2445 , should i include things that i left out here on that pr or work on an entirely new one ?

ahoppen:
> there is already a follow up PR in https://github.com/swiftlang/sourcekit-lsp/pull/2445 , should i include things that i left out here on that pr or work on an entirely new one ?

I just noticed that we eg. still have `typeDefinition` in `LangaugeService`. I’d pull that out of this PR and only add it in #2445, same for `typeDefinitionProvider: .bool(true)`, not sure if there’s anything else.

ahoppen (approved):
Good to go, let’s just remove all references to `typeDefinition` from the diff.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

loveucifer:
hmm failed ig 😵‍💫

ahoppen:
swift-format validation failed with the following errors because the `swift-format-ignore` lines were removed

```
+ /Users/ec2-user/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/build/buildbot_incremental/unified-swiftpm-build-macosx-x86_64/release/swift-format lint --parallel --strict --recursive /Users/ec2-user/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/sourcekit-lsp
branch-main/sourcekit-lsp/Sources/SourceKitD/sourcekitd_uids.swift:17:16: error: [TypeNamesShouldBeCapitalized] rename the struct 'sourcekitd_api_keys' using UpperCamelCase; for example, 'Sourcekitd_api_keys'
branch-main/sourcekit-lsp/Sources/SourceKitD/sourcekitd_uids.swift:794:16: error: [TypeNamesShouldBeCapitalized] rename the struct 'sourcekitd_api_requests' using UpperCamelCase; for example, 'Sourcekitd_api_requests'
branch-main/sourcekit-lsp/Sources/SourceKitD/sourcekitd_uids.swift:968:16: error: [TypeNamesShouldBeCapitalized] rename the struct 'sourcekitd_api_values' using UpperCamelCase; for example, 'Sourcekitd_api_values'
ERROR: command `['/Users/ec2-user/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/build/buildbot_incremental/unified-swiftpm-build-macosx-x86_64/release/swift-format', 'lint', '--parallel', '--strict', '--recursive', '/Users/ec2-user/jenkins/workspace/swift-sourcekit-lsp-PR-macOS/branch-main/sourcekit-lsp']` terminated with a non-zero exit status 1, aborting
```

Could you add them back in, ideally in a way that they don’t get removed when re-generating the file 😉

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftLanguageService/InlayHintResolve.swift
ahoppen, line 1: Could you add this file to [CMakeLists.txt](https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SwiftLanguageService/CMakeLists.txt) to make the tests pass on Windows?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

loveucifer:
hmm another macos fail :/

ahoppen:
Failures is unrelated. Let’s try again.

@swift-ci Please test macOS

loveucifer:
hmm failed again ig

ahoppen:
Oh, interesting. Almost all the tests failed with `error: Exited with unexpected signal code <random signal here>`. No idea how that happens. Let’s try again to see if it’s infrastructure or your PR.

@swift-ci Please test macOS

loveucifer:
Oh lol , lets see also i did apply for the member program , been 6 days didnt hear back seems like not upto it :/

PhantomInTheWire:
seems like a github issue again @swift-ci please test macos

ahoppen:
Looking through the logs after the `--- Running tests in parallel failed. Re-running tests serially to capture more actionable output.` marker, I noticed the following:

We are seeing a sourcekitd crash in `testCodeCompletionShowsUpdatedResultsAfterDependencyUpdated`, which we really shouldn’t because this is not a test that tests crash recovery of sourcekitd. It’s also the first time that we invoke code completion in the tests, so this is likely an issue with the [SourceKit plugin](https://github.com/swiftlang/sourcekit-lsp/tree/main/Sources/SwiftSourceKitPlugin). But even though sourcekitd crashes (which is an XPC service), it shouldn’t take down the SourceKit-LSP test runner itself.

I tested if this might have been introduced by a recent change in SwiftPM but have been unable to reproduce the issue by building SwiftPM from source so far. I’m now testing if this issue is caused by changes in this PR (which I can’t really believe) or if it’s caused by something else by re-triggering CI on https://github.com/swiftlang/sourcekit-lsp/pull/2450. I fear this will boil down to having to build an entire toolchain locally and trying to reproduce it that way. 

All that is to say, I’ll take care of it and come back once I know more.

ahoppen:
Update: Also seeing the same failure in https://github.com/swiftlang/indexstore-db/pull/276, so it’s not related to this PR. Now the debugging fun begins.

ahoppen:
Crash should be fixed by: https://github.com/swiftlang/swift/pull/86583. Let's try again. 

@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

ahoppen:
Finally 🎉🎉🎉

When you have time, do you mind rebasing the type definition PR and I'll review it next.

loveucifer:
> Finally 🎉🎉🎉
> 
> When you have time, do you mind rebasing the type definition PR and I'll review it next.

>
doing it rn hold on :D

PR #2439, opened 2026-01-07
Fix workspace/tests returning disambiguated XCTest IDs

### LINKED ISSUE #2015: `workspace/tests` returns unnecessarily disambiguated test items
### Swift version

swift-driver version: 1.120.4 Apple Swift version 6.1 (swiftlang-6.1.0.108.3 clang-1700.0.11.1

### Platform

macOS

### Editor

Visual Studio Code

### Description

When you have more than one XCTest test in an extension of an XCTestCase, the `workspace/tests` request will incorrectly return XCTests with the filename and line appended to the end of the ID for all tests in the extension.  This only happens with background indexing enabled.

This disambiguation is really only supposed to be for swift-testing tests that share the same ID signature (i.e. `@Test foo(_ x:Int)` and `@Test foo(_ x:String)`, which both have an id of `foo(_:)`).

### Steps to Reproduce

Add the following test to WorkspaceTestDiscoveryTests.swift and run it:

```swift
func testXCTestExtensionWithMultipleMethodsAndBackgroundIndexing() async throws {
    files: [
      "Tests/MyLibraryTests/MyTests.swift": """
      import XCTest
      1️⃣final class MyTests: XCTestCase {}6️⃣

      extension MyTests {
        2️⃣func testOne() {}3️⃣
        4️⃣func testOneTwo() {}5️⃣
      }
      """,
    ],
    manifest: packageManifestWithTestTarget,
    enableBackgroundIndexing: true
  )

  let tests = try await project.testClient.send(WorkspaceTestsRequest())
  XCTAssertEqual(
    tests,
    [
      TestItem(
        id: "MyLibraryTests.MyTests",
        label: "MyTests",
        location: try project.location(from: "1️⃣", to: "6️⃣", in: "MyTests.swift"),
        children: [
          TestItem(
            id: "MyLibraryTests.MyTests/testOne()",
            label: "testOne()",
            location: try project.location(from: "2️⃣", to: "3️⃣", in: "MyTests.swift")
          ),
          TestItem(
            id: "MyLibraryTests.MyTests/testOneTwo()",
            label: "testOneTwo()",
            location: try project.location(from: "4️⃣", to: "5️⃣", in: "MyTests.swift")
          )
        ]
      )
    ]
  )
}
```

This produces a TestItem with two children that have IDs `MyLibraryTests.MyTests/testOne():MyTests.swift:4:4` `MyLibraryTests.MyTests/testOneTwo():MyTests.swift:5:4`

Expected:
`MyLibraryTests.MyTests/testOne()` `MyLibraryTests.MyTests/testOneTwo()`

### Logging

_No response_

loveucifer:
fixes #2015 
When XCTest methods are defined in an extension of XCTestCase,
the semantic index returns them with childOf relations pointing
to the extension symbol rather than the class. This caused
duplicate test items and unnecessary ID disambiguation like
testOne():MyTests.swift:4:4 instead of testOne().

Now when processing test symbols, if an extension's extended
class is already in the test symbols for this request, we map
the extension USR to the class USR so tests parent correctly.

### DIFF
--- a/Sources/SourceKitD/SourceKitD.swift
+++ b/Sources/SourceKitD/SourceKitD.swift
@@ -23,8 +23,8 @@ extension sourcekitd_api_values: @unchecked Sendable {}
 fileprivate extension ThreadSafeBox {
   /// If the wrapped value is `nil`, run `compute` and store the computed value. If it is not `nil`, return the stored
   /// value.
-  func computeIfNil<WrappedValue>(compute: () -> WrappedValue) -> WrappedValue where T == WrappedValue? {
-    return withLock { value in
+  func computeIfNil<WrappedValue: Sendable>(compute: () -> WrappedValue) -> WrappedValue where T == WrappedValue? {
+    return withLock { (value: inout WrappedValue?) -> WrappedValue in
       if let value {
         return value
       }
--- a/Sources/SourceKitLSP/DocumentManager.swift
+++ b/Sources/SourceKitLSP/DocumentManager.swift
@@ -10,12 +10,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Dispatch
 import Foundation
 @_spi(SourceKitLSP) package import LanguageServerProtocol
 @_spi(SourceKitLSP) import SKLogging
 package import SKUtilities
 import SemanticIndex
+import SwiftExtensions
 package import SwiftSyntax
 
 /// An immutable snapshot of a document at a given time.
@@ -92,18 +92,15 @@ package final class DocumentManager: InMemoryDocumentManager, Sendable {
     case missingDocument(DocumentURI)
   }
 
-  // TODO: Migrate this to be an AsyncQueue (https://github.com/swiftlang/sourcekit-lsp/issues/1597)
-  private let queue: DispatchQueue = DispatchQueue(label: "document-manager-queue")
-
-  // `nonisolated(unsafe)` is fine because `documents` is guarded by queue.
-  private nonisolated(unsafe) var documents: [DocumentURI: Document] = [:]
+  // Documents storage, protected by a `ThreadSafeBox` to ensure thread safety without making APIs async.
+  private let documents: ThreadSafeBox<[DocumentURI: Document]> = ThreadSafeBox(initialValue: [:])
 
   package init() {}
 
   /// All currently opened documents.
   package var openDocuments: Set<DocumentURI> {
-    return queue.sync {
-      return Set(documents.keys)
+    return documents.withLock { docs in
+      return Set(docs.keys)
     }
   }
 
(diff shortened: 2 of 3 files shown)


Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -106,13 +106,26 @@ extension SourceKitLSPServer {
...
+
     // Gather any extension declarations that contains tests and add them to `occurrencesByParent` so we can properly
     // arrange their test items as the extension's children.
     for testSymbolOccurrence in testSymbolOccurrences {
       for parentSymbol in testSymbolOccurrence.relations.filter({ $0.roles.contains(.childOf) }).map(\.symbol) {
         guard parentSymbol.kind == .extension else {
           continue
         }
+        // find the class this extension extends
+        if let extendedSymbol = index?.occurrences(relatedToUSR: parentSymbol.usr, roles: .extendedBy).first?.symbol {
```
ahoppen: We should be able to skip this index lookup if we already have an entry for `parentSymbol.usr` in `extensionToClassUsr`, right?

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -123,8 +136,11 @@ extension SourceKitLSPServer {
     }
 
     for testSymbolOccurrence in testSymbolOccurrences {
-      let childOfUsrs = testSymbolOccurrence.relations
-        .filter { $0.roles.contains(.childOf) }.map(\.symbol.usr).filter { testSymbolUsrs.contains($0) }
+      let childOfRelations = testSymbolOccurrence.relations.filter { $0.roles.contains(.childOf) }
+      // resolve extension usrs to class usrs if class is in scope
+      let childOfUsrs = childOfRelations.map {
+        extensionToClassUsr[$0.symbol.usr] ?? $0.symbol.usr
```
ahoppen: I don’t think that’s the right approach. We want to report the test methods as children of the extension, not the class itself.

In your PR comment you say

> This caused duplicate test items

Why does this cause duplicate test items?

loveucifer: so the duplicates happen because both the class and the extension end up as roots with the same name "MyTests". The class comes from testSymbolOccurrences (since it inherits from XCTestCase), and the extension gets added separately at lines 128-134 because the test methods have a childOf relation pointing to it.

when both are roots , mergingTestsInExtensions() sees 2 items with the same ID. Since neither item's children are empty , it tries to merge them ig, but something goes wrong and the child tests get disambiguated IDs.

But tbh now I see your point - maybe the fix shouldn't be something like  remapping parent relationships. ig the right approach might be to not add the extension as a root when the class is already present? That way the tests would still parent to the extension USR, but the extension wouldn't be a separate root competing with the class.

loveucifer: Hmmm closing this for now since i want to concentrate on the compiler fix  we talked about and having multiple open prs is making it hard :p

ahoppen: For sure, if you (or somebody else) wants to pick this up in the future again, please do.

PR #2442, opened 2026-01-09
Re-use module cache for background preparation

### LINKED ISSUE #2428: Re-use module cache from `.build` or have a global module cache for background preparation
Preparing the `BuildServerIntegrationTests` target in the SourceKit-LSP target takes 42 seconds on my machine. When setting the module cache path to a module cache generated during a build, this time is reduced to 22 seconds. This means that we can significantly reduce background indexing time by either
- Sharing the module cache with the build (this won’t have a direct indexing benefit when the user didn’t build the package before opening in VS Code but should have a similar benefit on their first clean build time.
- Having a global module cache for SourceKit-LSP. I have a strong suspicion that most modules that take long to build are SDK modules that will be used between multiple packages, so this could be a real benefit for indexing, without being a benefit for build times, unless we also consider a global module cache for SwiftPM in general. The main question with a global module cache would likely be to decide how / if we can clean it. An idea would be to delete modules that are older than 1 month and trust that they will be re-generated if still needed.

loveucifer:
Reduces background indexing time by sharing module caches during target preparation.

When preparing targets for indexing, SourceKit-LSP now passes `-module-cache-path` to `swift build`. The cache location is determined by:

1. User-configured `index.swiftModuleCachePath` if set
2. User's build cache at `.build/<arch>/ModuleCache` if it exists
3. Global cache at `~/.cache/sourcekit-lsp/module-cache` (created if needed)

Stale modules older than 30 days are automatically cleaned from the global cache.

Fixes #2428


loveucifer:
the pr on the main repo swiftlang/swift isnt reviewed yet ig so I just picked this issue and can work on this for some time, after that gets a merge , ig ill work on the other compiler issue ( i think there were 2 issues ) lets see :D

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -766,6 +766,38 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+        // reuse existing user build cache
+        cacheToUse = userBuildCache
+      } else {
+        // fall back to global module cache
+        let globalCache = FileManager.default.homeDirectoryForCurrentUser
+          .appending(components: ".cache", "sourcekit-lsp", "module-cache")
+        try? FileManager.default.createDirectory(at: globalCache, withIntermediateDirectories: true)
+
+        // clean up stale modules older than 30 days
+        Self.cleanupStaleModules(in: globalCache, olderThan: 30)
```
ahoppen: I think we should do this in a background task. There’s no need to block getting build settings on deleting stale module files.

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -766,6 +766,38 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
     arguments += options.swiftPMOrDefault.buildToolsSwiftCompilerFlags?.flatMap { ["-Xbuild-tools-swiftc", $0] } ?? []
+
+    // add module cache path for faster preparation
+    if let customCachePath = options.indexOrDefault.swiftModuleCachePath {
+      if !customCachePath.isEmpty {
+        arguments += ["-Xswiftc", "-module-cache-path", "-Xswiftc", customCachePath]
+      }
+    } else {
+      // try to use the user's build module cache first
+      let userBuildCache =
```
ahoppen: I don’t think it makes sense to decide whether to use a global or a project-local module cache based on whether `.build` exists when SourceKit-LSP is started. This property may change while SourceKit-LSP is still alive. I would prefer if we just commit to either global or project-local module cache and stick with that (my personal preference is a global module cache).

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -766,6 +766,38 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+        .appending(component: "ModuleCache")
+
+      let cacheToUse: URL
+      if FileManager.default.fileExists(atPath: userBuildCache.path) {
+        // reuse existing user build cache
+        cacheToUse = userBuildCache
+      } else {
+        // fall back to global module cache
+        let globalCache = FileManager.default.homeDirectoryForCurrentUser
+          .appending(components: ".cache", "sourcekit-lsp", "module-cache")
```
ahoppen: Hmm, the storage location here is tricky. I would prefer if we do something similar to [how we discover](https://github.com/swiftlang/sourcekit-lsp/blob/29a0bdd15f360539f1f456fec943acd44654f860/Sources/sourcekit-lsp/SourceKitLSP.swift#L177-L208) the `.sourcekit-lsp/config.json` file. That is
- On macOS place this in `URL.cachesDirectory.appending(components: "org.swift.sourcekit-lsp", "ModuleCache")`
- On Linux, check if `XDG_CACHE_HOME` (https://wiki.archlinux.org/title/XDG_Base_Directory) exists and put the module cache in there and uses `~/.cache` as a fallback
- On Windows: Figure out how to get it’s cache directory, not sure where that is.

The advantage would be that OSs can clean up these files if they are low on disk space that way.

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -912,6 +944,28 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+      )
+    else { return }
+
+    for case let fileURL as URL in enumerator {
+      guard let resourceValues = try? fileURL.resourceValues(forKeys: [.contentModificationDateKey]),
+        let modificationDate = resourceValues.contentModificationDate,
+        modificationDate < cutoffDate
+      else { continue }
+
+      try? FileManager.default.removeItem(at: fileURL)
```
ahoppen: We should wrap this in a `orLog` and log if removing a file failed.

loveucifer:
Will send a pr  addressing all the comments , after we get a suggestion on what to do from @rintaro and @hamishknight so it dosent have unwanted additions :D

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -912,6 +944,28 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+      let enumerator = FileManager.default.enumerator(
+        at: cacheDir,
+        includingPropertiesForKeys: [.contentModificationDateKey],
+        options: [.skipsHiddenFiles]
+      )
+    else { return }
+
+    for case let fileURL as URL in enumerator {
+      guard let resourceValues = try? fileURL.resourceValues(forKeys: [.contentModificationDateKey]),
+        let modificationDate = resourceValues.contentModificationDate,
```
rintaro: Can we use `contentAccessDate` instead?

ahoppen:
As I commented in https://github.com/swiftlang/sourcekit-lsp/issues/2428#issuecomment-3938701596, I think the correct solution would be to add functionality in SwiftPM to use the existing global module cache that is used if no `-module-cache-path` is specified to the compiler instead of having a module cache that’s specific to SourceKit-LSP. At that point we also wouldn’t need to clean it up anymore.

loveucifer:
closing pr as it belongs to package manager ( for anyone who might stumble onto this later )

PR #2443, opened 2026-01-09
add support for multi-line string literals when converting string concatenation to string interpolation

### LINKED ISSUE #1568: Add support for multi-line string literals when converting string concatenation to string interpolation
### Description

#1551 added a code action for converting string concatenation to string interpolation, however multi-line string literals are not yet supported. We should add support for this, e.g.
```swift
"""
I
am
a
""" + value + 
"""
multi-line
string
literal
"""
```
will be converted to
```swift
"""
I
am
a\(value)multi-line
string
literal
"""
```

loveucifer:
fixes #1568
Thiis PR adds support for multi-line string literals in the "Convert String Concatenation to String Interpolation" code action.

Changes:

modified preflight to track hasMultilineString instead of rejecting multi-line strings
used multilineStringQuoteToken() when any input string is multi-line
updated test to verify multi-line string support works correctly

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
+++ b/Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(SourceKitLSP) import LanguageServerProtocol
+import SwiftBasicFormat
 import SwiftRefactor
 import SwiftSyntax
 
@@ -86,6 +87,10 @@ struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvide
         lastSegment.trailingTrivia = .newline
         segments = StringLiteralSegmentListSyntax(segments.dropLast() + [lastSegment])
       }
+
+      // Re-introduce the indentation of the last multi-line string literal's
+      // closing quote to the entire string literal.
+      segments = segments.indented(by: closingQuoteTrivia, indentFirstLine: true)
     }
 
     let quoteToken: TokenSyntax =
@@ -98,10 +103,7 @@ struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvide
       ? quoteToken.with(\.trailingTrivia, .newline)
       : quoteToken
 
-    let closingQuote: TokenSyntax =
-      hasMultilineString
-      ? quoteToken.with(\.leadingTrivia, closingQuoteTrivia)
-      : quoteToken
+    let closingQuote: TokenSyntax = quoteToken
 
     return syntax.with(
       \.elements,
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1150,9 +1150,9 @@ final class CodeActionTests: SourceKitLSPTestCase {
                   range: positions["1️⃣"]..<positions["2️⃣"],
                   newText: #"""
                     """
-                    a
-                    bc
-                    d
+                      a
+                      bc
+                      d
                       """
                     """#
                 )


Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1049,36 +1046,32 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
-          title: "Apply De Morgan's law, converting '!(a && b)' to '(!a || !b)'",
+          title: "Convert String Concatenation to String Interpolation",
           kind: .refactorInline,
           edit: WorkspaceEdit(
             changes: [
               uri: [
                 TextEdit(
                   range: positions["1️⃣"]..<positions["2️⃣"],
-                  newText: "(!a || !b)"
+                  newText: "\"\"\"I\nam\na\n\\( value )multi-line\nstring\nliteral\n\"\"\""
```
ahoppen: Could you us a raw multi-line string literal (starting with `#"""`) to represent that. That should be way more readable.

---

Also, I think the expected result should not have a newline before `\(value)` (please double check that I’m right though)

---

Finally, why are we adding spaces around `value`, shouldn’t this  be `\(value)`?

ahoppen (commented):
I think the tricky part will be concatenating multi-line string literals with different base indentation levels, eg.

```swift
"""
a
b
""" + """
  c
  d
  """
```
 
Should become

```swift
"""
a
bc
d
""" 
```

or 

```swift
  """
  a
  bc
  d
  """ 
```

Tests/SourceKitLSPTests/CodeActionTests.swift
loveucifer: semantically it shouldn't be there. however, I traced through the code and the segments array is correct (a with no trailing newline), but SwiftSyntax appears to be adding the newline when rendering the StringLiteralExprSyntax with multiline quotes. Looking at the segments before creating the result:
"I\n", "am\n", "a" (no trailing newline), (value), "multi-line\n", ...
butut the output still has a\n(value). this seems to be SwiftSyntax behavior ( i think ) when combining string segments with expression segments in multiline literals. I've added a comment noting this limitation - but i have no idea where to start from so this might take a while

loveucifer: more into this ->  a, \(x), and b are all on the same source line, and SwiftSyntax keeps them together.
But when we're combining segments from separate source strings (first string ends with a, then expression from a different component), SwiftSyntax treats each segment boundary as a potential line break. :/

ahoppen: If you look at the contents of the string literal carefully, you’ll notice that all but the last line have the trailing newline as part of their contents while the last line has it as trivia. This is how we indicate whether the newline is part of the represented string literal value or just for formatting purposes. You’ll need to handle that accordingly.

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
-              pounds: commonPounds,
-              expressions: [
-                LabeledExprSyntax(expression: component.singleLineTrivia)
-              ]
-            )
-          )
+        let expression =
+          hasMultilineString
+          ? component.with(\.leadingTrivia, []).with(\.trailingTrivia, [])
+          : component.singleLineTrivia
```
ahoppen: Why do we need to handle trivia of expressions differently depending on whether we generate a single-line or multi-line string literal? Shouldn’t they be treated the same?

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -179,6 +238,46 @@ private extension StringLiteralExprSyntax {
   }
 }
 
+/// strips base indentation from multiline string segments
+private func stripIndentation(
```
ahoppen: I think you should be able to use `IndentationReover`, introduced by https://github.com/swiftlang/sourcekit-lsp/pull/2420 for this.

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
 
+      var literalSegments = stringLiteral.segments
+
+      // strip base indentation for multiline strings
+      if hasMultilineString && !stringLiteral.isSingleLine {
+        let baseIndent = stringLiteral.closingQuote.indentationOfLine
+        literalSegments = stripIndentation(from: literalSegments, baseIndent: baseIndent)
+      }
+
+      if hasMultilineString && !isLastComponent && !stringLiteral.isSingleLine {
```
ahoppen: Nitpick

```suggestion
      if hasMultilineString, !isLastComponent, !stringLiteral.isSingleLine {
```

Same in a few other places. It’s just a little more swifty.

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
+        if let lastSeg = literalSegments.last, case let .stringSegment(s) = lastSeg {
+          let strippedContent = s.content.with(\.trailingTrivia, [])
+          let strippedSeg = s.with(\.content, strippedContent)
```
ahoppen: I think you can simplify this slightly.

```suggestion
        if var lastSegment = literalSegments.last {
          lastSegment.trailingTrivia = []
```

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
+      var literalSegments = stringLiteral.segments
+
+      // strip base indentation for multiline strings
+      if hasMultilineString && !stringLiteral.isSingleLine {
+        let baseIndent = stringLiteral.closingQuote.indentationOfLine
+        literalSegments = stripIndentation(from: literalSegments, baseIndent: baseIndent)
+      }
+
+      if hasMultilineString && !isLastComponent && !stringLiteral.isSingleLine {
+        if let lastSeg = literalSegments.last, case let .stringSegment(s) = lastSeg {
```
ahoppen: Could you add a short comment why the transform is necessary for future reference?

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
+          let strippedContent = s.content.with(\.trailingTrivia, [])
+          let strippedSeg = s.with(\.content, strippedContent)
+          literalSegments = StringLiteralSegmentListSyntax(
+            literalSegments.dropLast() + [.stringSegment(strippedSeg)]
+          )
+        }
+      }
+
+      // process segments with possible merging for multiline strings
+      var segmentsToAdd = literalSegments
```
ahoppen: Why do we need a new variable here? Couldn’t you just modify `literalSegments`?

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
+        case let .stringSegment(lastStrSeg) = lastSeg,
+        !lastStrSeg.content.text.hasSuffix("\n"),
+        let firstSeg = segmentsToAdd.first,
+        case let .stringSegment(firstStrSeg) = firstSeg
+      {
+        // merge last of previous with first of current
+        let merged = lastStrSeg.content.text + firstStrSeg.content.text
+        let mergedSeg = StringSegmentSyntax(content: .stringSegment(merged))
+        segments = StringLiteralSegmentListSyntax(segments.dropLast() + [.stringSegment(mergedSeg)])
+        segmentsToAdd = StringLiteralSegmentListSyntax(segmentsToAdd.dropFirst())
```
ahoppen: We shouldn’t be looking into the contents of the string literal here. Rather, I’d like to treat the contents as opaque and just concern ourselves with trivia because all the information about indentation and the last newline in the string literal should be represented in there. What I’d prefer is that we did something like the following: Before adding segments to `literalSegments`, check if we are in the last component of the concatenation chain. If not, remove the trailing newline from the last segment in the multi-line component. After that I think (or hope) that you should be able to just concatenate the segments without having to do any special merging anymore.

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,118 @@
...
       }
+
+      segments += segmentsToAdd
+    }
+
+    // ensure trailing newline for multiline strings
+    if hasMultilineString,
+      let lastSeg = segments.last,
+      case let .stringSegment(lastStrSeg) = lastSeg,
+      !lastStrSeg.content.text.hasSuffix("\n")
```
ahoppen: Similar here, we shouldn’t need to inspect the string literal’s contents, I’d prefer to inspect everything purely based on trivia.

ahoppen (commented):
Some comments inline, I also have a couple more test case that would be good to check:

```swift
"""
a
b
""" + "c" + """
d
"""
```

```swift
"""
a
b
""" + "c"
```

```swift
func test() {
  """
  a
  b
  """ + """
    c
    d
    """
}
```

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
loveucifer: hmm isnt  multiline string literal content (including the trailing newline before """) stored as text content ? so is trivia gonna work ?

ahoppen: The last trailing newline before the closing `"""` is stored as trivia. It’s carefully designed in a way that if you concatenate the text of all string components, you get the represented string literal value. For example, also consider how both the backslash and the newline are represented as trivia here:

```swift
"""
abc \
def
"""
```

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
loveucifer: i tried this and this is what happened 
`// This FAILED with 4 test failures:
let remover = IndentationRemover(indentation: baseIndent, indentFirstLine: true)
literalSegments = StringLiteralSegmentListSyntax(
  literalSegments.map { remover.rewrite($0).cast(StringLiteralSegmentListSyntax.Element.self) }
)`
i think this is beacause string content indentation is stored differently than statement indentation , idk if this is right but the tests do fail with `IndentationRemover`

loveucifer: more info -> with IndentationRemover, the actual output is: `newText: "\"\"\"\na\nb  c\n  d\n\"\"\""`
i think thee issue is that `IndentationRemover` only modifies trivia (leading/trailing whitespace around tokens), but in multiline string literals, the indentation within the string content is stored in the token text itself, pls do correct me if im wrong

ahoppen: Nope, all indentation that may be stripped without modifying the string’s content are stored as trivia. I think I left a comment somewhere else about how the text contents of the string segments form the represented string literal value and that we thus shouldn’t modify it.

loveucifer:
is this the right approach ? im sure something is wrong :/

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -11,56 +11,100 @@
...
-        segments.append(
-          .expressionSegment(
-            ExpressionSegmentSyntax(
-              pounds: commonPounds,
-              expressions: [
-                LabeledExprSyntax(expression: component.singleLineTrivia)
-              ]
-            )
-          )
+        // preserve comments as block comments
```
ahoppen: This comment doesn’t seem to match the code. Or am I missing something?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1083,19 +1083,237 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["1️⃣"]..<positions["2️⃣"],
+                  newText: #"""
+                    """
+                    a
+                    b  c
+                      d
+                    """
```
ahoppen: This should be the following, right?


```suggestion
                    """
                    a
                    bc
                    d
                    """
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1083,19 +1083,237 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testConvertStringConcatenationNestedIndentation() async throws {
```
ahoppen: Isn’t this exactly the same as `testConvertStringConcatenationToStringInterpolationDifferentIndentation`? Or am I missing something.

ahoppen (commented):
Nice, this looks like we’re on the right track. One of the test cases doesn’t seem to match but otherwise this looks good form a medium-level review perspective.

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -42,12 +41,34 @@ struct ConvertStringConcatenationToStringInterpolation: SyntaxRefactoringProvide
...
+        let indentPattern = baseIndent.description
+        // use IndentationRemover for trivia
         let remover = IndentationRemover(indentation: baseIndent, indentFirstLine: true)
         literalSegments = StringLiteralSegmentListSyntax(
-          literalSegments.map { remover.rewrite($0).cast(StringLiteralSegmentListSyntax.Element.self) }
+          literalSegments.map { segment in
+            var rewritten = remover.rewrite(segment).cast(StringLiteralSegmentListSyntax.Element.self)
+            // also strip base indentation from string segment text
+            if case var .stringSegment(strSeg) = rewritten {
+              var text = strSeg.content.text
```
ahoppen: As I said before, the contents of the string literal should remain unchanged to ensure we don’t change the underlying string literal content. All indentation-related spaces that we might remove should be in the trivia. Please ensure that we only modify trivia instead of the segment’s contents.

loveucifer: i was going to address this with another commit but u fortunately my mac has broken down , i will make the fixes in a week or so sorry 🥲💔

loveucifer:
i had some confusion finding a fix for this at first i tried a more complex approach: decomposing trivia pieces using `@_spi(RawSyntax)` to access `.decomposed`, then comparing them prefix-wise to strip the base indentation. This worked but required using a semi-private API and was more complex than necessar imo 
then i found a better solution after a while and i think this is good bcoz  in multiline string literals, SwiftSyntax stores each segment's base indentation (the indentation matching the closing `"""`) as leading trivia, while the content text contains only the actual string value. Since the leading trivia IS exactly the base indentation, we can simply clear it with `leadingTrivia = []`.
ill send a commit but im not fairly sure if this appraoch is better than the other

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -18,49 +18,96 @@ import SwiftSyntax
...
+      // For multiline strings, strip base indentation (trivia only).
+      // In multiline string literals, each segment's leading trivia contains
+      // exactly the base indentation (matching the closing """), so we clear it.
+      // Do not modify content - only trivia.
+      if hasMultilineString, !stringLiteral.isSingleLine {
+        literalSegments = StringLiteralSegmentListSyntax(
+          literalSegments.map { segment in
+            var updated = segment
+            updated.leadingTrivia = []
+            return updated
```
ahoppen:
```suggestion
            segment.with(\.leadingTrivia, [])
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1083,19 +1083,161 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+              ]
+            ]
+          )
+        )
+      ]
     }
   }
 
   func testApplyDeMorganLawNegatedAnd() async throws {
+
```
ahoppen: Unrelated change

Sources/SwiftLanguageService/CodeActions/ConvertStringConcatenationToStringInterpolation.swift
```diff
@@ -18,49 +18,96 @@ import SwiftSyntax
...
+    let quoteToken: TokenSyntax =
+      hasMultilineString
+      ? .multilineStringQuoteToken()
+      : .stringQuoteToken()
+
+    let openingQuote: TokenSyntax =
+      hasMultilineString
+      ? quoteToken.with(\.trailingTrivia, .newline)
+      : quoteToken
+
```
ahoppen, line 105: We should re-introduce the indentation of the last multi-line string literal’s closing quote. Eg.

```swift
let x = """
  a
  """ + 
  """
  b
  """
```

Should become 

```swift
let x = """
  ab
  """
```

ahoppen: You need to re-introduce the indentation of the last closing quote to the entire string literal, eg. right now `testConvertStringConcatenationToStringInterpolationDifferentIndentation` doesn’t constitute a valid string literal because the contents are indented less than the closing quote. I think you should be able to use `.indented(by:)` on `StringLiteralSegmentListSyntax` for that.

ahoppen:
@swift-ci Please test

PR #2444, opened 2026-01-09
Perf: Index only reachable products of package dependencies (#2427)

### LINKED ISSUE #2427: Decide if we want to index products of package dependencies
We currently index all products of a package’s dependencies. This matches the fact that you can build all all targets referenced by a dependency’s product on the command line (eg. you can run `swift build --product swift-driver` from within SourceKit-LSP. Most users probably never use this fact though and we already don’t index all code in package dependencies (eg. test targets of dependencies aren’t indexed). Based on initial measurements, we could save about another minute (relative to 2:45 min) on the total indexing time of SourceKit-LSP if we didn’t index products of dependencies.

If we want to go this way, we’d likely need to check if a product is in `graph.reachableProducts` here: https://github.com/swiftlang/swift-package-manager/blob/3a4ba00c1fbdf119512f2293a51d79521e664518/Sources/Build/BuildPlan/BuildPlan.swift#L980-L984

<details><summary>Stub for a test case</summary>

```swift
func testFoo() async throws {
  let project = try await MultiFileTestProject(
    files: [
      // MyLibrary
      "MyDependency/Sources/MyDependency/MyDependency.swift": """
      public func makeHello() -> String { "" }
      """,
      "MyDependency/Sources/swift-help/swift-help.swift": "",
      "MyDependency/Package.swift": """
      // swift-tools-version: 5.7

      import PackageDescription

      let package = Package(
        name: "MyDependency",
        products: [
          .library(name: "MyDependency", targets: ["MyDependency"]),
          .executable(name: "swift-help", targets: ["swift-help"]),
        ],
        targets: [
          .target(name: "MyDependency"),
          .executableTarget(name: "swift-help")
        ]
      )
      """,

      // MyPackage
      "MyPackage/Sources/MyPackage/MyPackage.swift": """
      import MyDependency

      func test() {
        let x: Void = makeHello()
      }
      """,
      "MyPackage/Package.swift": """
      // swift-tools-version: 5.7

      import PackageDescription

      let package = Package(
        name: "MyPackage",
        dependencies: [
          .package(path: "../MyDependency"),
        ],
        targets: [.target(name: "MyPackage", dependencies: [.product(name: "MyDependency", package: "MyDependency")])]
      )
      """,
    ],
    workspaces: { scratchDirectory in
      return [WorkspaceFolder(uri: DocumentURI(scratchDirectory.appending(component: "MyPackage")))]
    },
    enableBackgroundIndexing: true
  )
  try await project.testClient.send(SynchronizeRequest(index: true))
}
```

</details>

DPrakashhh:
Fixes #2427
Summary
This PR optimizes background indexing by filtering out unreachable dependency products.

1-**Logic**: Uses `ModulesGraph.reachableProducts` to identify only the dependency modules required by the root package.
2-**Safety**: Always indexes all modules in the root package, regardless of reachability.
3-**Implementation**: Replaces the generic `traverseModules` in `SwiftPMBuildServer.swift` with a filtered `traverseReachableModules` wrapper.
4-**Clean Code**: Moves filtering logic to a new `PackageGraph+Reachability.swift` .

### DIFF
--- a/Sources/BuildServerIntegration/PackageGraph+Reachability.swift
+++ b/Sources/BuildServerIntegration/PackageGraph+Reachability.swift
@@ -1,0 +1,64 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if !NO_SWIFTPM_DEPENDENCY
+@preconcurrency import PackageGraph
+@preconcurrency import SourceKitLSPAPI
+
+extension ModulesGraph {
+  /// Computes the set of module (target) names that are reachable from root package modules.
+  ///
+  /// A module is considered reachable if it is part of the dependency graph starting
+  /// from the root package's modules. This is used to filter out unreachable modules
+  /// from dependencies during background indexing to improve performance.
+  ///
+  /// - Returns: A set of module names that are reachable from root packages.
+  func reachableModuleNames() -> Set<String> {
+    Set(reachableModules.map(\.name))
+  }
+}
+
+extension SourceKitLSPAPI.BuildDescription {
+  /// Traverses modules while filtering out unreachable modules from dependencies.
+  ///
+  /// This method wraps the standard `traverseModules` call but filters out modules
+  /// that are not reachable from the root package. Modules from the root package
+  /// itself are always included, even if they are not technically reachable,
+  /// to ensure complete indexing of the user's own code.
+  ///
+  /// - Parameters:
+  ///   - modulesGraph: The package graph containing reachability information.
+  ///   - callback: Closure called for each reachable build target with its parent.
+  func traverseReachableModules(
+    in modulesGraph: ModulesGraph,
+    callback: (any BuildTarget, _ parent: (any BuildTarget)?) -> Void
+  ) {
+    let reachableModuleNames = modulesGraph.reachableModuleNames()
+
+    self.traverseModules { buildTarget, parent in
+      // Always include modules from the root package
+      guard !buildTarget.isPartOfRootPackage else {
+        callback(buildTarget, parent)
+        return
+      }
+
+      // For dependency modules, only include them if they are reachable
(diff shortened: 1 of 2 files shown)


ahoppen:
As mentioned in https://github.com/swiftlang/sourcekit-lsp/issues/2427, I’d like to hear @rintaro and @hamishknight’s opinion on this idea before deciding whether this is the right direction to go in.

DPrakashhh:
yeah, sure i have done this so they can review this implementation quality and based on this they can decide whether to continue this or leave it :)

Sources/BuildServerIntegration/PackageGraph+Reachability.swift
```diff
@@ -0,0 +1,64 @@
...
+  /// Computes the set of module (target) names that are reachable from root package modules.
+  ///
+  /// A module is considered reachable if it is part of the dependency graph starting
+  /// from the root package's modules. This is used to filter out unreachable modules
+  /// from dependencies during background indexing to improve performance.
+  ///
+  /// - Returns: A set of module names that are reachable from root packages.
+  func reachableModuleNames() -> Set<String> {
+    Set(reachableModules.map(\.name))
+  }
```
ahoppen, line 27: Since we only call this from a single location and the implementation is trivial, I think we can just inline it.

Sources/BuildServerIntegration/PackageGraph+Reachability.swift
```diff
@@ -0,0 +1,64 @@
...
+    callback: (any BuildTarget, _ parent: (any BuildTarget)?) -> Void
+  ) {
+    let reachableModuleNames = modulesGraph.reachableModuleNames()
+
+    self.traverseModules { buildTarget, parent in
+      // Always include modules from the root package
+      guard !buildTarget.isPartOfRootPackage else {
+        callback(buildTarget, parent)
+        return
+      }
```
ahoppen, line 52: Why is this needed?

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -484,7 +484,9 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     self.swiftPMTargets = [:]
     self.targetDependencies = [:]
 
-    buildDescription.traverseModules { buildTarget, parent in
+    // Use the filtered traversal to only index reachable modules from dependencies.
+    // All modules from the root package are included regardless of reachability.
```
ahoppen, line 488: I don’t think this comment provides much value because it just states what the next line does.

ahoppen (commented):
Sorry for only coming back to this now. It would be great to get this in. I left a few comments inline.

PR #2445, opened 2026-01-10
Implement textDocument/typeDefinition request

loveucifer:
follow up pr referenced from - [comment](https://github.com/swiftlang/sourcekit-lsp/pull/2436#discussion_r2677710970)
to be merged after #2436 

- Add TypeDefinition.swift with cursorInfo USR lookup
- Add cursorInfoFromTypeUSR and convertMangledTypeToUSR helpers
- Register typeDefinitionProvider capability and handler
- Add typeDefinition to LanguageService protocol
- Add tests for local and cross-module type definitions

### DIFF
--- a/Sources/SwiftLanguageService/CMakeLists.txt
+++ b/Sources/SwiftLanguageService/CMakeLists.txt
@@ -52,6 +52,7 @@ add_library(SwiftLanguageService STATIC
   SyntaxHighlightingTokens.swift
   SyntaxTreeManager.swift
   TestDiscovery.swift
+  TypeDefinition.swift
   VariableTypeInfo.swift
 )
 set_target_properties(SwiftLanguageService PROPERTIES


loveucifer:
i might have missed something because i was going back and forth , lets see

ahoppen:
I’ll start reviewing this after https://github.com/swiftlang/sourcekit-lsp/pull/2436 is merged and you can rebase this on top of `main` with all those changes in, so no need for you to keep this update until that time.

loveucifer:
i dont think this right T...T , let me check again

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,54 @@
...
+      keys.compilerArgs: compileCommand?.compilerArgs as [any SKDRequestValue]?,
+    ])
+
+    let dict = try await send(sourcekitdRequest: \.cursorInfo, skreq, snapshot: snapshot)
+
+    guard let typeUsr: String = dict[keys.typeUsr] else {
+      return nil
+    }
+
+    // Try cursorInfo with USR lookup first - it knows the current in-memory state
```
ahoppen: There’s no `second` here, so I think the comment no longer applies

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,54 @@
...
+    if let typeInfo = try await cursorInfoFromTypeUSR(typeUsr, in: snapshot),
+      let location = typeInfo.symbolInfo.bestLocalDeclaration
+    {
+      return .locations([location])
+    }
```
ahoppen, line 73: Similar to my comments in the inlay type hints PR, this doesn’t handle jumping to generated interfaces. Instead of doing another ad-hoc fix like you did in the inlay hints PR, I’d really like to see this unified with the logic in `indexBasedJumpToDefinition`. Could you look into how we can best re-use code between both that function, inlay hints resolve and the type definition request?

Tests/SourceKitLSPTests/TypeDefinitionTests.swift
```diff
@@ -0,0 +1,258 @@
...
+    guard case .locations(let locations) = response, let location = locations.first else {
+      XCTFail("Expected location response")
+      return
+    }
+
+    XCTAssertEqual(location.uri, uri)
+    XCTAssertEqual(location.range, Range(positions["1️⃣"]))
+  }
+
+  func testTypeDefinitionEnumType() async throws {
```
ahoppen: Classes and enums are just the same as a struct in the type system, so these two tests don’t really provide much value and I think we can delete them.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1146,6 +1148,7 @@ extension SourceKitLSPServer {
       completionProvider: completionOptions,
       signatureHelpProvider: signatureHelpOptions,
       definitionProvider: .bool(true),
+      typeDefinitionProvider: .bool(true),
```
ahoppen, line 1151: Can you also advertise the `typeDefinitionProvider` [here](https://github.com/swiftlang/sourcekit-lsp/blob/61ba0c92a49689eb3c64e86943ff205a46e6667b/Sources/SwiftLanguageService/SwiftLanguageService.swift#L393)?

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -438,6 +439,10 @@ package extension LanguageService {
     throw ResponseError.requestNotImplemented(DeclarationRequest.self)
   }
 
+  func typeDefinition(_ request: TypeDefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+    throw ResponseError.requestNotImplemented(TypeDefinitionRequest.self)
+  }
+
```
ahoppen, line 445: Should we also forward the request to clangd in the `ClangLanguageService`? Not sure if clangd supports the request right now but if it does in the future, we’ll pick it up that way.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -790,6 +790,8 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await self.handleRequest(for: request, requestHandler: self.declaration)
     case let request as RequestAndReply<DefinitionRequest>:
       await self.handleRequest(for: request, requestHandler: self.definition)
+    case let request as RequestAndReply<TypeDefinitionRequest>:
+      await self.handleRequest(for: request, requestHandler: self.typeDefinition)
```
ahoppen, line 794: This list is ordered alphabetically? Could insert this at `t`?

ahoppen (commented):
Very nice. I think the biggest thing will be to reconcile the resolving of the definition’s USR with `indexBasedDefinition` now.

ahoppen:
Instead of adding a new requirement to `LanguageService` like you did, which also doesn’t unify the implementation for the inlay type resolve, I was thinking to pull `definitionLocations` and `definitionInInterface` out to top-level functions (or maybe their own type if it makes sense, probably not but not sure) and then call it from `indexBasedDefinition` in `SourceKitLSPServer`, the type definition request in `SwiftLanguageService` and also the inlay hint resolve. I just checked and `indexBasedDefinition` mostly deals with logic to jump to definitions instead of declarations (only really applicable to the C family of languages), so I don’t think we need to generalize any of that except for potentially uniquing the results, which we can just copy to all three locations.

Sources/SourceKitLSP/DefinitionLocations.swift
ahoppen, line 1: Could you add this file to CMakeLists.txt?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import IndexStoreDB
+@_spi(SourceKitLSP) package import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+package import SemanticIndex
+
+/// converts a symbol index location to an LSP location
```
ahoppen: Could you make sure all functions have the lowes possible access level?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+func indexToLSPLocation(_ location: SymbolLocation) -> Location? {
+  guard !location.path.isEmpty else { return nil }
+  return Location(
+    uri: location.documentUri,
+    range: Range(
+      Position(
+        line: max(0, location.line - 1),
+        utf16index: max(0, location.utf8Column - 1)
```
ahoppen, line 33: Could you keep the comments from eh previous `indexToLSPLocation` implementation and remove the implementation in `SourceKitLSPServer` so we only have one copy of it?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+      )
+    )
+  )
+}
+
+/// returns the definition location for a symbol, handling generated interfaces for SDK types
+package func definitionLocations(
```
ahoppen, line 53: Could you remove the corresponding definition in SourceKitLSPServer so we only have one copy of it (same for all the other functions in this file).

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+  index: CheckedIndex?,
+  openGeneratedInterface:
+    @escaping (
+      _ document: DocumentURI,
+      _ moduleName: String,
+      _ groupName: String?,
+      _ symbolUSR: String?
+    ) async throws -> GeneratedInterfaceDetails?
+) async throws -> [Location] {
+  // module symbols generate a textual interface
```
ahoppen: Could you keep the comments of the current implementation? In general please copy the function with the least modification necessary to avoid accidental introduction of behavior changes.

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+      return [bestLocalDeclaration]
+    }
+    return []
+  }
+
+  guard let usr = symbol.usr else { return [] }
+  logger.info("Performing indexed jump-to-definition with USR \(usr)")
+
+  let occurrences = index.definitionOrDeclarationOccurrences(ofUSR: usr)
+
```
ahoppen, line 108: You’re missing the check for `symbol.isDynamic` here. Is this intentional?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,143 @@
...
+  for symbol: SymbolDetails,
+  originatorUri: DocumentURI,
+  index: CheckedIndex?,
+  openGeneratedInterface:
+    @escaping (
+      _ document: DocumentURI,
+      _ moduleName: String,
+      _ groupName: String?,
+      _ symbolUSR: String?
+    ) async throws -> GeneratedInterfaceDetails?
```
ahoppen: Instead of passing the closure in here, would it be easier to understand if we pass an `any LanguageService` in here? Same below.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -59,10 +59,7 @@ extension SwiftLanguageService {
     return hint
   }
 
-  /// Looks up the definition location for the type at the given position.
-  ///
-  /// This is used by inlay hint resolution to enable go-to-definition on type hints.
-  /// For SDK types, this returns a location in the generated interface.
+  /// looks up the definition location for the type at the given position
```
ahoppen: Why did you change this comment. Please ensure that you don’t have unrelated changes in your PR.

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -87,27 +84,20 @@ extension SwiftLanguageService {
...
+          moduleName: moduleName,
+          groupName: groupName,
+          symbolUSR: symbolUSR
+        )
       }
-    }
+    )
 
-    return nil
+    return locations.first
```
ahoppen: If we get multiple locations, the type might be ambiguous. Instead of giving the user a false sense of security by taking them to one of them, let’s return `nil` instead, ie. use `only` instead of `first`.

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,83 @@
...
+      guard let typeUsr: String = dict[keys.typeUsr] else {
+        return nil
+      }
+      let typeInfo = try await cursorInfoFromTypeUSR(typeUsr, in: snapshot)
+      symbol = typeInfo?.symbolInfo
+    }
+
+    guard let symbol else {
+      return nil
+    }
```
ahoppen: Shouldn’t everything in `symbol == nil` go into the `guard else` branch?

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+
+    // otherwise get the type of the symbol at this position
+    if symbol == nil {
+      guard let typeUsr: String = dict[keys.typeUsr] else {
+        return nil
+      }
+      let typeInfo = try await cursorInfoFromTypeUSR(typeUsr, in: snapshot)
+      symbol = typeInfo?.symbolInfo
```
ahoppen: Do we ever hit this case? If we can’t parse the cursor info response, shouldn’t we just exit with an error?

Tests/SourceKitLSPTests/TypeDefinitionTests.swift
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+    guard case .locations(let locations) = response, let location = locations.first else {
+      XCTFail("Expected location response")
+      return
+    }
+
+    XCTAssertEqual(location.uri, uri)
+    XCTAssertEqual(location.range, Range(positions["1️⃣"]))
+  }
```
ahoppen, line 174: Could you also add a test case that jumps to a generated interface, eg. jumping to `String`?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -0,0 +1,154 @@
...
+      let location = try await definitionInInterface(
+        moduleName: systemModule.moduleName,
+        groupName: systemModule.groupName,
+        symbolUSR: symbol.usr,
+        originatorUri: originatorUri,
+        openGeneratedInterface: openGeneratedInterface
+      )
+      return [location]
+    }
+  }
```
ahoppen, line 126: Did you drop the logic that checks for `symbol.isDynamic` etc?

Sources/SourceKitLSP/DefinitionLocations.swift
ahoppen: This is still outstanding.

Sources/SourceKitLSP/DefinitionLocations.swift
ahoppen: I think you still dropped some comments.

loveucifer:
Regarding the isDynamic check -  your comment  said that "indexBasedDefinition mostly deals with logic to jump to definitions instead of declarations (only really applicable to the C family of languages)", so I left tout the isDynamic handling out of the shared definitionLocations  function since:
Type definition and inlay hint resolve don't need dynamic symbol handling afaik they just need to find the type's location
Only indexBasedDefinition  in SourceKitLSPServer needs the isDynamic logic for proper jump-to-definition behavior

What i thought was to do is this 
keep the shared definitionLocations simple (for type definition + inlay hints)
keep indexBasedDefinition in SourceKitLSPServer with its own isDynamic handling
Is this the right way   or should the shared function also include the isDynamic logic?

ahoppen:
Oh, sorry, you‘re right about `isDynamic`. I think I was partially confused because the old `definitionLocations` still exists and I would have expected the `isDynamic` check to show up somewhere in the diff where it stuck around but it didn’t. That being said, could you still address all comments from my previous review? Eg. https://github.com/swiftlang/sourcekit-lsp/pull/2445#discussion_r2702707875 but all of them really.

loveucifer:
okie got it

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2142,11 +2036,47 @@ extension SourceKitLSPServer {
...
+            if let receiverUsrs = symbol.receiverUsrs {
+              return transitiveSubtypeClosure(
+                ofUsrs: receiverUsrs,
+                index: index
+              )
+            } else {
+              return nil
+            }
+          }()
+          let additionalOccurrences = index.definitionOrDeclarationOccurrences(ofUSR: usr)
```
ahoppen: We are already performing this index lookup in `definitionLocations`. Since index lookups may be somewhat expensive, could we avoid it, eg. by returning the array of occurrences from `definitionLocations` as well?

Sources/SwiftLanguageService/InlayHintResolve.swift
```diff
@@ -87,27 +88,13 @@ extension SwiftLanguageService {
...
-      )
-      if let details = interfaceDetails {
-        let position = details.position ?? Position(line: 0, utf16index: 0)
-        return Location(uri: details.uri, range: Range(position))
-      }
-    }
+    let locations = try await SourceKitLSP.definitionLocations(
+      for: typeInfo.symbolInfo,
+      originatorUri: snapshot.uri,
+      index: nil,
```
ahoppen: Are you intentionally not passing in an index here? I feel like we should use an index if we have one.

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    // Helper to get symbol from typeUsr
+    func symbolFromTypeUsr() async throws -> SymbolDetails? {
+      guard let typeUsr: String = dict[keys.typeUsr] else {
+        return nil
+      }
+      guard let typeInfo = try await cursorInfoFromTypeUSR(typeUsr, in: snapshot) else {
+        return nil
+      }
+      return typeInfo.symbolInfo
```
ahoppen:
```suggestion
      return try await cursorInfoFromTypeUSR(typeUsr, in: snapshot)?.symbolInfo
```

At which point I’m not sure if having the helper function is beneficial to have this helper function anymore, I’d just inline it.

Sources/SwiftLanguageService/TypeDefinition.swift
```diff
@@ -0,0 +1,83 @@
...
+      guard let typeSymbol = try await symbolFromTypeUsr() else {
+        return nil
+      }
+      symbol = typeSymbol
+    }
+
+    let locations = try await SourceKitLSP.definitionLocations(
+      for: symbol,
+      originatorUri: uri,
+      index: nil,
```
ahoppen: Similarly here, shouldn’t we be passing in an index?

Sources/SourceKitLSP/SourceKitLSPServer.swift
loveucifer: okay

ahoppen (approved):
Let’s 🚢  it

ahoppen:
@swift-ci Please test

loveucifer:
@swift-ci please test windows

Sources/SwiftLanguageService/TypeDefinition.swift
ahoppen, line 1: Could you add this file to CMakeLists.txt?

loveucifer:
@swift-ci please test

loveucifer:
@swift-ci please test windows

ahoppen:
@swift-ci Please test Windows

loveucifer:
yaay 🎊

PR #2448, opened 2026-01-11
Fix bugs in demorgan implementation

PhantomInTheWire:
fixes bugs reported here: https://github.com/swiftlang/sourcekit-lsp/issues/2441#issuecomment-3728695531

i figured since i was the one that implemented it i should be the one to fix the minor bugs and inconsistencies reported

#### bug report for ref:

> ##### 1. Negation should propagate into the complement of `try(!)` and `await` (and recursively), i.e.
>
> ```js
> true && try a // original
> !(false || try !a) // intended
> !(false || !(try a)) // actual
> ```
>
> ---
>
> ##### 2. Bitwise DeMorgan has inconsistent heuristics, I'm not sure if it's intended,
>
> ```js
> // bivalent
>
> // no inversion
> 1 | 2 // no D.M.
>
> // any inversion
> ~1 | 2
> ~(1 & ~2) // suggestion
> 1 | ~2
> ~(~1 & 2) // suggestion
>
> // multivalent
> 1 | 2 | 3
> ~(~1 & ~2 & ~3) // suggestion
> ```
>
> ---
>
> ###### 3. Forced Unwrapped should also be an atomic expression,
>
> ```js
> a! && false // original
> !(!a! && true) // intended
> !((!a!) || true) // actual
> ```
>
> **P.S.** Just a small bug report, trivia are not preserved in some cases,
>
> ```js
> [0, true /**/&& false][1] && false // source
> true /**/&& false // target
> !(false /**/|| true) // intended
> !(false|| true) // actual
> ```
>

1. This should be fixed now.

2. Suggestions are only emitted when at least one side can be simplified. This behavior is intentional, though I agree it can look odd, if this is concerning we should discuss the desired semantics separately.

3. Agreed. I have added forced unwrap and a few similar expressions to the atomic expression whitelist.

In addition, I fixed the trivia loss in the case mentioned above and addressed another issue where indentation was not being attached to the newly introduced negation operator. This PR also includes a few related minor fixes discovered along the way.

Finally, I added regression tests to in case the implementation is modified.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
+++ b/Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
@@ -261,7 +261,7 @@ package struct DeMorganTransformer {
     var strippedInfix = mappedInfix
     let leadingTrivia = strippedInfix.leadingTrivia
     if !leadingTrivia.isEmpty {
-      strippedInfix = strippedInfix.with(\.leadingTrivia, [])
+      strippedInfix.leadingTrivia = []
     }
 
     let innerExpr: ExprSyntax
@@ -452,7 +452,7 @@ package struct DeMorganTransformer {
       }
       var newExpr = negatedInner.expr
       // Ensure separation to prevent `try` and `!` from fusing into `try!`
-      if tryExpr.tryKeyword.trailingTrivia.isEmpty && newExpr.leadingTrivia.isEmpty {
+      if tryExpr.tryKeyword.trailingTrivia.isEmpty, newExpr.leadingTrivia.isEmpty {
         newExpr = newExpr.with(\.leadingTrivia, .spaces(1))
       }
       let newTry = tryExpr.with(\.expression, newExpr)
@@ -465,7 +465,7 @@ package struct DeMorganTransformer {
       }
       var newExpr = negatedInner.expr
       // Ensure separation for style (await! does not exist, but `await !a` is better than `await!a`)
-      if awaitExpr.awaitKeyword.trailingTrivia.isEmpty && newExpr.leadingTrivia.isEmpty {
+      if awaitExpr.awaitKeyword.trailingTrivia.isEmpty, newExpr.leadingTrivia.isEmpty {
         newExpr = newExpr.with(\.leadingTrivia, .spaces(1))
       }
       let newAwait = awaitExpr.with(\.expression, newExpr)


Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -256,21 +256,29 @@ package struct DeMorganTransformer {
 
     let mappedInfix = mapThroughParentheses(expr, replacement: ExprSyntax(newInfix))
 
+    // Transfer leading trivia from the inner expression to the new negation operator
+    // so that indentation is preserved and not trapped inside.
+    var strippedInfix = mappedInfix
+    let leadingTrivia = strippedInfix.leadingTrivia
+    if !leadingTrivia.isEmpty {
+      strippedInfix = strippedInfix.with(\.leadingTrivia, [])
```
ahoppen:
```suggestion
      strippedInfix.leadingTrivia = []
```

Sources/SwiftLanguageService/CodeActions/ApplyDeMorganLaw.swift
```diff
@@ -422,6 +439,42 @@ package struct DeMorganTransformer {
...
+    // Only handle plain `try`, not `try?` or `try!`.
+    // `try?` changes the type to Optional, so negation doesn't propagate.
+    // `try!` could propagate, but for simplicity we only handle plain `try`.
+    if let tryExpr = expr.as(TryExprSyntax.self), tryExpr.questionOrExclamationMark == nil {
+      guard let negatedInner = negateExpression(tryExpr.expression, exprType: exprType) else {
+        return nil
+      }
+      var newExpr = negatedInner.expr
+      // Ensure separation to prevent `try` and `!` from fusing into `try!`
+      if tryExpr.tryKeyword.trailingTrivia.isEmpty && newExpr.leadingTrivia.isEmpty {
```
ahoppen:
```suggestion
      if tryExpr.tryKeyword.trailingTrivia.isEmpty, newExpr.leadingTrivia.isEmpty {
```

ahoppen (commented):
Very nice. Thanks for fixing the bugs so quickly, @PhantomInTheWire 🙏🏽 Just two minor nitpicks, otherwise LGTM. And I agree with the assessments in your PR description.

PhantomInTheWire:
Pushed the nitpick fixes.

Happy to follow up on this since I originally implemented it.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PhantomInTheWire:
ci failure seems unrelated

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

PR #2449, opened 2026-01-11
Add position markers in the expected ranges

PhantomInTheWire:
this should make tests more readable, follow up to this pr https://github.com/swiftlang/sourcekit-lsp/pull/2435 
reference comment here https://github.com/swiftlang/sourcekit-lsp/pull/2435#discussion_r2670161613

### DIFF
--- a/Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
+++ b/Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
@@ -29,7 +29,7 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
   func testAddDocumentationRefactor() throws {
     try assertRefactor(
       """
-        1️⃣func refactor(syntax: DeclSyntax, in context: Void) -> DeclSyntax? { }
+        1️⃣func 2️⃣refactor(syntax: DeclSyntax, in context: Void) -> DeclSyntax? { }
       """,
       context: (),
       provider: AddDocumentation.self,
@@ -48,15 +48,14 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
               """
           )
         ]
-      },
-      checkMarkers: ["1️⃣"]
+      }
     )
   }
 
   func testAddDocumentationRefactorSingleParameter() throws {
     try assertRefactor(
       """
-        1️⃣func refactor(syntax: DeclSyntax) { }
+        1️⃣func 2️⃣refactor(syntax: DeclSyntax) { }
       """,
       context: (),
       provider: AddDocumentation.self,
@@ -71,8 +70,7 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
               """
           )
         ]
-      },
-      checkMarkers: ["1️⃣"]
+      }
     )
   }
 
@@ -80,7 +78,7 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
     try assertRefactor(
       """
       1️⃣{
-         "name": "Produce",
+         3️⃣"name": "Produce",
          "shelves": [
              {
                  "name": "Discount Produce",
@@ -119,7 +117,7 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
           )
         ]
       },
-      checkMarkers: ["1️⃣"]
+      checkMarkers: ["1️⃣", "3️⃣"]
     )
   }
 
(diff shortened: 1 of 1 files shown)


Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -23,52 +23,58 @@ final class SyntaxRefactorTests: SourceKitLSPTestCase {
...
-              ///
-              /// - Returns:
-              \("")
-            """
-        )
-      ]
+      expected: { positions in
+        [
+          SourceEdit(
+            range: positions["1️⃣"]!..<positions["1️⃣"]!,
```
ahoppen: Nitpick, we have a convenience initializer on `Range` for cases like this. Same in the other tests.

```suggestion
            range: Range(positions["1️⃣"]!),
```

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -300,7 +342,19 @@ func assertRefactor<R: EditRefactoringProvider>(
   var parser = Parser(textWithoutMarkers)
   let sourceFile = SourceFileSyntax.parse(from: &parser)
 
-  let markersToCheck = markers.isEmpty ? [("1️⃣", 0)] : markers.sorted { $0.key < $1.key }
+  let markersToCheck: [(String, Int)]
```
ahoppen: Let’s make this `AbsolutePosition` to start with and save ourselves the trouble to try and understand what the integer represents here, in particular the UTF-8 vs UTF-16 offset question.

```suggestion
  let markersToCheck: [(String, AbsolutePosition)]
```

And probably worth pushing that type change all the way into `extractMarkers` and make `SKTestSupport` depend on swift-syntax for that. I did a really short survey just now and it shouldn’t affect too many places. Worth doing it now since this is a refactoring PR anyway.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: cool done.
though i needed this to make it work, i wonder if we should patch this upstream.
```swift
fileprivate extension Range where Bound == AbsolutePosition {
  init(_ position: AbsolutePosition) {
    self = position..<position
  }
}
```

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: sure done.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -19,56 +19,68 @@ import SwiftRefactor
 import SwiftSyntax
 import XCTest
 
+fileprivate extension Range where Bound == AbsolutePosition {
```
ahoppen, line 22: Oh, interesting. I could have sworn that we had this extension defined without a requirement on `Bound`. But I was wrong. Thanks for adding it.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
```diff
@@ -19,56 +19,68 @@ import SwiftRefactor
...
+                ///   - syntax:
+                ///   - context:
+                ///
+                /// - Returns:
+                \("")
+              """
+          )
+        ]
+      },
+      checkMarkers: ["1️⃣"]
```
ahoppen: Sorry for not noticing this earlier but this reduces the test coverage, right? Previously, we also checked that this refactoring action correctly applies at the `2` marker and now we don’t. The same applies to all other tests that did not explicitly specify `checkMarkers` as well.

ahoppen (commented):
Sorry, missed one comment in last review.

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: cool

Tests/SourceKitLSPTests/SyntaxRefactorTests.swift
PhantomInTheWire: hmm, i did not think of it this way,  re-adding multiple trigger points and removing redundant checkMarkers calls.

ahoppen (approved):
Thanks for the cleanup

ahoppen:
@swift-ci Please test

PhantomInTheWire:
@ahoppen this should be good to merge

PR #2452, opened 2026-01-13
Code action to add missing import declarations

PhantomInTheWire:
fixes: https://github.com/swiftlang/sourcekit-lsp/issues/1089

This PR implements a new code action, AddMissingImports, which provides quick fixes to automatically add import statements for unresolved types or values. It leverages the semantic index to identify defining modules and intelligently inserts the necessary imports into the source file.

Key Features
- Automatically detects cannot find '...' in scope and cannot find type '...' in scope diagnostics.
- Queries the semantic index to find which modules define the missing type or value.
- Filtering strategy
  - Filters out modules that are already imported in the current file.
  - Excludes the current module to avoid self-imports.
- Insertion Strategy:
  - Inserts new imports after the last existing import statement if present.
  - If no imports exist, inserts at the beginning of the file (after leading trivia/headers).
  - Handles spacing and newlines correctly to maintain file formatting.
- Sorting: Provides multiple suggestions sorted alphabetically by module name when a type is defined in multiple accessible modules.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
+++ b/Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
@@ -101,13 +101,11 @@ extension SwiftLanguageService {
         .compactMap { $0.path.first?.name.text }
     )
 
-    // Get current module name
     var currentModule: String? = nil
     if let canonicalTarget = await workspace.buildServerManager.canonicalTarget(for: request.textDocument.uri) {
       currentModule = await workspace.buildServerManager.moduleName(for: request.textDocument.uri, in: canonicalTarget)
     }
 
-    // Find candidate modules and create commands
     var codeActions: [CodeAction] = []
 
     for diagnostic in relevantDiagnostics {
@@ -148,26 +146,23 @@ extension SwiftLanguageService {
     let snapshot = try await self.latestSnapshot(for: command.textDocument.uri)
     let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
 
+    let allImportDecls = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
     let insertPosition = AddMissingImportsHelper.importInsertionPosition(
       for: command.moduleName,
       in: syntaxTree,
       snapshot: snapshot
     )
 
-    let hasExistingImports = !syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }.isEmpty
-
-    let importText: String
-    if hasExistingImports {
-      let allImportDecls = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
-      let lastImportEnd = allImportDecls.last.map { snapshot.position(of: $0.endPosition) }
+    // Determine import text based on insertion position:
+    // - If inserting at end of last import: prepend newline
+    // - Otherwise (before import or no imports): append newline
+    let lastImportEnd = allImportDecls.last.map { snapshot.position(of: $0.endPosition) }
+    let importText =
       if lastImportEnd == insertPosition {
-        importText = "\nimport \(command.moduleName)"
+        "\nimport \(command.moduleName)"
       } else {
-        importText = "import \(command.moduleName)\n"
+        "import \(command.moduleName)\n"
       }
-    } else {
-      importText = "import \(command.moduleName)\n"
-    }
 
     guard let sourceKitLSPServer else {
       throw ResponseError.unknown("Connection to the editor closed")
@@ -252,7 +247,6 @@ private enum AddMissingImportsHelper {
 
     let importNames = importDecls.compactMap { $0.path.first?.name.text }
 
-    // Check if existing imports are already sorted
     let isSorted = zip(importNames, importNames.dropFirst()).allSatisfy { $0 <= $1 }
 
     if isSorted,
@@ -261,21 +255,16 @@ private enum AddMissingImportsHelper {
         return firstPath > newModule
       })
     {
-      // Insert before the first import that is lexicographically after the new module
       return snapshot.position(of: insertBeforeImport.position)
     } else if let lastImport = importDecls.last {
-      // Not sorted or new module is alphabetically last - insert after last import
       return snapshot.position(of: lastImport.endPosition)
     }
-    // No existing imports - insert before first statement but after its leading trivia (header comments)
     if let firstStatement = syntaxTree.statements.first {
       return snapshot.position(of: firstStatement.positionAfterSkippingLeadingTrivia)
     }
     return snapshot.position(of: AbsolutePosition(utf8Offset: 0))
   }
 }
-
-// MARK: - Diagnostic Extension
 
 private extension Diagnostic {
   var codeString: String? {
--- a/Tests/SourceKitLSPTests/AddMissingImportsTests.swift
+++ b/Tests/SourceKitLSPTests/AddMissingImportsTests.swift
@@ -61,7 +61,6 @@ class AddMissingImportsTests: SourceKitLSPTestCase {
 
     let (uri, positions) = try project.openDocument("main.swift")
 
-    // Get the diagnostics
     var diagnostic: Diagnostic?
     let diags = try await project.testClient.send(
       DocumentDiagnosticsRequest(textDocument: TextDocumentIdentifier(uri))
--- a/Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
+++ b/Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
@@ -23,7 +23,6 @@ import XCTest
 
 class AddMissingImportsUnitTests: XCTestCase {
 
-  /// Creates a syntax tree and snapshot from source code for testing.
   private func makeSyntaxTreeAndSnapshot(
     from source: String,
     uri: DocumentURI
@@ -234,7 +233,7 @@ class AddMissingImportsUnitTests: XCTestCase {
     let actions = SwiftLanguageService.findMissingImports(
       diagnostics: [diagnostic],
       existingImports: [],
-      currentModule: "MyModule",  // This is the current module
+      currentModule: "MyModule",
       syntaxTree: syntaxTree,
       snapshot: snapshot,
       uri: uri


Sources/SKTestSupport/SwiftPMTestProject.swift
```diff
@@ -242,7 +242,7 @@ package class SwiftPMTestProject: MultiFileTestProject {
   }
 
   /// Build a SwiftPM package package manifest is located in the directory at `path`.
-  package static func build(at path: URL, extraArguments: [String] = []) async throws {
+  package static func build(at path: URL, buildTests: Bool = true, extraArguments: [String] = []) async throws {
```
ahoppen: We should be able to achieve this by just adding `--build-tests` to `extraArguments`, no new parameter needed.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        )
+        """,
+      capabilities: clientCapabilitiesWithCodeActionSupport,
+      enableBackgroundIndexing: false
+    )
+
+    let (uri, positions) = try project.openDocument("main.swift")
+
+    // We build only `Lib` so that we get the index for it, but don't fail because `Exec` doesn't build.
+    try await SwiftPMTestProject.build(
```
ahoppen: This should be the same behavior that you get if you let background indexing run. It will fail to index `Exec` but that’s perfectly fine.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,122 @@
...
+      buildTests: false,
+      extraArguments: ["--target", "Lib"]
+    )
+
+    try await project.testClient.send(SynchronizeRequest(index: true))
+
+    // Wait for the diagnostic to appear. The compiler should complain about missing LibStruct.
+    var foundDiagnostic: Diagnostic?
+    // Try for up to 30 seconds
+    for _ in 0..<30 {
```
ahoppen: We have `repeateUntilExpectedResult` for this, don’t we? But why do we need to repeat this at all? We should have all necessary information to get diagnostics on the first go, the dependencies are already built and we should have build settings.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,122 @@
...
+      return
+    }
+
+    let request = CodeActionRequest(
+      range: Range(positions["1️⃣"]),
+      context: CodeActionContext(diagnostics: [diagnostic], only: [.quickFix]),
+      textDocument: TextDocumentIdentifier(uri)
+    )
+
+    var codeAction: CodeAction?
```
ahoppen: Shouldn’t the code action always show up if we have produced the diagnostic?

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,122 @@
...
+          codeAction = action
+          break
+        }
+      }
+      try await Task.sleep(for: .seconds(1))
+    }
+
+    XCTAssertNotNil(codeAction)
+
+    if let codeAction {
```
ahoppen: You could use `try XCTUnwrap` here instead.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,122 @@
...
+      guard let changes = codeAction.edit?.changes?[uri] else {
+        XCTFail("No edit changes found")
+        return
+      }
+      XCTAssertTrue(changes.contains { $0.newText.contains("import Lib") })
+    }
```
ahoppen, line 99: Could you also add a check for the code that the code action produces?

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+@_spi(SourceKitLSP) import SwiftLanguageService
+import SwiftParser
+import SwiftSyntax
+import XCTest
+
+class AddMissingImportsUnitTests: XCTestCase {
+
+  /// Creates a syntax tree and snapshot from source code for testing.
+  private func makeSyntaxTreeAndSnapshot(from source: String, uri: DocumentURI) -> (SourceFileSyntax, DocumentSnapshot)
+  {
```
ahoppen: Personal formatting preference

```suggestion
  private func makeSyntaxTreeAndSnapshot(
    from source: String, 
    uri: DocumentURI
  ) -> (SourceFileSyntax, DocumentSnapshot) {
```

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    let syntaxTree = Parser.parse(source: source)
+    let snapshot = DocumentSnapshot(uri: uri, language: .swift, version: 0, lineTable: LineTable(source))
+    return (syntaxTree, snapshot)
+  }
+
+  func testAddMissingImports() {
+    let diagnostic = Diagnostic(
+      range: Position(line: 0, utf16index: 10)..<Position(line: 0, utf16index: 19),
```
ahoppen: Could you use `DocumentPositions` and position markers here? I feel like that would make the diagnostics more readable.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+      severity: .error,
+      code: .string("cannot_find_in_scope"),
+      source: "sourcekitd",
+      message: "cannot find 'LibStruct' in scope"
+    )
+
+    let uri = try! DocumentURI(string: "file:///main.swift")
```
ahoppen: Can’t you use the standard way of creating mock DocumentURIs here?

```suggestion
    let uri = DocumentURI(for: .swift)
```

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+      currentModule: nil,
+      syntaxTree: syntaxTree,
+      snapshot: snapshot,
+      uri: uri,
+      lookup: lookup
+    )
+
+    XCTAssertEqual(actions.count, 1)
+
+    guard let action = actions.first else { return }
```
ahoppen: This can be `let action = try XCTUnwrap(actions.only)`. Same in a few more places of this test.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    XCTAssertEqual(action.title, "Import Lib")
+
+    guard let edit = action.edit, let changes = edit.changes?[uri] else {
+      XCTFail("No edit found")
+      return
+    }
```
ahoppen: This can also become `XCTUnwrap`

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+
+    XCTAssertEqual(changes.count, 1)
+    XCTAssertEqual(changes.first?.newText, "import Lib\n")
+    // Should insert at beginning since there are no imports
+    XCTAssertEqual(changes.first?.range, Position(line: 0, utf16index: 0)..<Position(line: 0, utf16index: 0))
```
ahoppen: Would this be easier to check for correctness mentally if we checked the final result that gets produced when running apply(edits:to:)? Same for the other tests (for all the comments above).

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+
+    XCTAssertEqual(actions.count, 1)
+    guard let action = actions.first, let edit = action.edit, let changes = edit.changes?[uri] else {
+      XCTFail("No edit found")
+      return
+    }
+
+    // Import should be inserted at the beginning since header comments are leading trivia in this parse
+    // (The multi-line string literal has leading indentation which makes the comments part of trivia)
+    XCTAssertEqual(changes.first?.range.lowerBound.line, 0)
```
ahoppen: I would have expected this to get inserted after the first block comment, ie. in line 6.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+  }
+
+  func testImportInsertionWithMultiLineHeader() {
```
ahoppen: How is this different than `testImportInsertionAfterFileHeader`?

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    guard let action = actions.first, let edit = action.edit, let changes = edit.changes?[uri] else {
+      XCTFail("No edit found")
+      return
+    }
+
+    // Import should be inserted at the beginning since header comments are leading trivia
+    XCTAssertEqual(changes.first?.range.lowerBound.line, 0)
+  }
+
+  func testImportInsertionWithNoImportsAndNoHeader() {
```
ahoppen: How is this different than `testAddMissingImports`?

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    // Import should be inserted at the end of the Foundation import (line 0)
+    XCTAssertEqual(changes.first?.range.lowerBound.line, 0)
+
+    // When inserting after existing imports, the new text should have a prepended newline
+    // to avoid concatenation like "import Foundationimport Lib"
+    XCTAssertEqual(
+      changes.first?.newText,
+      "\nimport Lib\n",
+      "Should prepend newline when inserting after existing imports"
+    )
```
ahoppen: I think this is exactly where it’s easier to express the expected behavior by applying the changes to the original source code and comparing that.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    // Diagnostic with proper code
+    let diagnostic = Diagnostic(
+      range: Position(line: 0, utf16index: 8)..<Position(line: 0, utf16index: 17),
+      severity: .error,
+      code: .string("cannot_find_in_scope"),
+      source: "sourcekitd",
+      message: "cannot find 'LibStruct' in scope"
+    )
+
+    let lookup: (String) -> Set<String> = { _ in ["Lib"] }
```
ahoppen: Would it be more readable to just define this in the call to `findMissingImports`? Could even be a trailing closure.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+      diagnostics: [diagnostic],
+      existingImports: [],
+      currentModule: nil,
+      syntaxTree: syntaxTree,
+      snapshot: snapshot,
+      uri: uri,
+      lookup: lookup
+    )
+
+    XCTAssertEqual(actions.count, 1, "Should match diagnostic by code")
```
ahoppen: I’m not entirely sure what this is intended to check.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
```diff
@@ -0,0 +1,444 @@
...
+      diagnostics: [diagnostic],
+      existingImports: [],
+      currentModule: nil,
+      syntaxTree: syntaxTree,
+      snapshot: snapshot,
+      uri: uri,
+      lookup: lookup
+    )
+
+    XCTAssertEqual(actions.count, 1, "Should match diagnostic by string fallback when code is absent")
```
ahoppen: Let’s just match by diagnostic ID. If we don’t have a diagnostic ID, let’s just not offer this refactoring action.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    guard let buildSettings = await self.compileCommand(for: request.textDocument.uri, fallbackAfterTimeout: true),
+      !buildSettings.isFallback
+    else {
+      return []
+    }
+
+    guard
+      let index = await self.sourceKitLSPServer?.workspaceForDocument(uri: request.textDocument.uri)?.index(
+        checkedFor: .modifiedFiles
+      )
```
ahoppen: Personal formatting preference

```suggestion
      let index = await self.sourceKitLSPServer?.workspaceForDocument(uri: request.textDocument.uri)?
        .index(checkedFor: .modifiedFiles)
```

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    else {
+      return []
+    }
+
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = Self.extractExistingImports(from: syntaxTree)
+    let currentModule = Self.extractModuleName(from: buildSettings.compilerArgs)
```
ahoppen: We have `BuildServerManager.moduleName(for:in:)` that does exactly that already, just need to get the canonical target for the file first.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    guard
+      let index = await self.sourceKitLSPServer?.workspaceForDocument(uri: request.textDocument.uri)?.index(
+        checkedFor: .modifiedFiles
+      )
+    else {
+      return []
+    }
+
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = Self.extractExistingImports(from: syntaxTree)
```
ahoppen: Does it really make sense to define this in a separate function? The implementation is quite trivial and isn’t called anywhere else.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+      lookup: { typeName in Self.findModulesDefining(typeName, in: index) }
+    )
+  }
+
+  /// Finds missing import code actions for diagnostics indicating unresolved types.
+  ///
+  /// This is the core logic separated from async/index dependencies for unit testing.
+  /// Filters diagnostics, extracts type names, looks up defining modules, and generates
+  /// code actions with proper import insertion positions.
+  package static func findMissingImports(
```
ahoppen: All of these static functions are very specific to this refactoring action, so I don’t think it makes sense to add them to `SwiftLangaugeService`. What do you think about creating an enum as a namespace for these functions?

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+package import SwiftSyntax
+
+extension SwiftLanguageService {
+  /// Generates code actions to add missing imports for unresolved type references.
+  ///
+  /// This method queries the semantic index to find which modules define the missing types,
+  /// then creates code actions to import those modules. It filters out already-imported modules
+  /// and self-imports to avoid redundant or invalid suggestions.
+  package func addMissingImports(_ request: CodeActionRequest) async throws -> [CodeAction] {
+    let snapshot = try await self.latestSnapshot(for: request.textDocument.uri)
```
ahoppen: Can we add an early exit here if we don’t have a missing type diagnostic in the request (which will be the case almost all the time). That way we don’t need to do any of the more expensive operations like getting the build settings and syntax tree.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+  /// testability and clarity. Only considers type-like symbols.
+  private static func findModulesDefining(_ typeName: String, in index: CheckedIndex) -> Set<String> {
+    var modules: Set<String> = []
+    index.forEachCanonicalSymbolOccurrence(byName: typeName) { occurrence in
+      guard IndexSymbolKind.typeKinds.contains(occurrence.symbol.kind) else {
+        return true
+      }
+
+      // Prefer container name (module), fall back to location module name
+      let moduleName = index.containerNames(of: occurrence).first ?? occurrence.location.moduleName
```
ahoppen: The container name doesn’t necessarily have to be the module name, eg. in the following the container name of `Inner` is `Outer`

```swift
struct Outer {
  struct Inner {}
}
```

Any reason why we can’t just always grab the module name. After all that’s what we’re after.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+      )
+    else {
+      return []
+    }
+
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = Self.extractExistingImports(from: syntaxTree)
+    let currentModule = Self.extractModuleName(from: buildSettings.compilerArgs)
+
+    return Self.findMissingImports(
```
ahoppen: Computing all the unused imports may be somewhat expensive. Could we only do this when we actually execute the code action? For example see how `RemoveUnsuedImports` only determines if the code action is applicable in `retrieveRemoveUnusedImportsCodeAction` and defers all the actual source transform to `removeUnusedImports`.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    return compilerArgs[moduleNameIndex + 1]
+  }
+
+  /// Finds all modules that define a given type by querying the semantic index.
+  ///
+  /// This abstracts away the slow index lookup operation, isolating it for better
+  /// testability and clarity. Only considers type-like symbols.
+  private static func findModulesDefining(_ typeName: String, in index: CheckedIndex) -> Set<String> {
+    var modules: Set<String> = []
+    index.forEachCanonicalSymbolOccurrence(byName: typeName) { occurrence in
```
ahoppen: I believe this will also return all references to a symbol with this name. A test case for this would be something like

LibA/LibA.swift
```swift
struct MyStruct {}
```

LibB/A.swift
```
import LibA
func test(x: MyStruct) {}
```

LibB/B.swift
```
func anotherTest(x: MyStruct) {} // Invoke code action here
```

This will currently likely report both `LibA` and `LibB` as modules of `MyStruct`. You need to filter by the occurrence’s role of `.definition` or `.declaration`.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+      guard IndexSymbolKind.typeKinds.contains(occurrence.symbol.kind) else {
```
ahoppen: We shouldn’t just offer this for types. This is also valid for top-level functions defined in other modules.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+  ///
+  /// Inserts after the last existing import if any exist, otherwise at the file start
+  /// (after any leading trivia like file headers).
+  ///
+  /// - Returns: A tuple containing the insertion position and a boolean indicating
+  ///   whether imports already exist (true if inserting after existing imports).
+  private static func importInsertionPosition(
+    in syntaxTree: SourceFileSyntax,
+    snapshot: DocumentSnapshot
+  ) -> (position: Position, hasExistingImports: Bool) {
```
ahoppen: I don’t think we need `hasExistingImports` if we properly adjust returned position to be right after the newline following the last import statement.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    let importDecls = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
+
+    if let lastImport = importDecls.last {
```
ahoppen: What do you think about trying to insert the import in alphabetical order. Ie. insert it before the first import statement that is lexicographically after this new import.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    return (snapshot.position(of: startPosition), false)
+  }
+
+  /// Extracts the type name from a diagnostic message like "cannot find 'Foo' in scope".
+  private static func extractTypeName(from message: String) -> String? {
+    for prefix in ["cannot find type '", "cannot find '"] {
+      if let typeName = extractQuotedText(from: message, after: prefix, before: "' in scope") {
+        return typeName
+      }
+    }
```
ahoppen: Could we do this using Regex-matching?

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    else {
+      return nil
+    }
+    return String(text[prefixRange.upperBound..<suffixRange.lowerBound])
+  }
+
+  /// Checks if a diagnostic indicates a missing type or value in scope.
+  private static func isMissingTypeOrValueDiagnostic(_ diagnostic: Diagnostic) -> Bool {
+    if let code = diagnostic.codeString {
+      return code == "cannot_find_type_in_scope" || code == "cannot_find_in_scope"
```
ahoppen: Let’s also add `cannot_find_type_in_scope_did_you_mean` and `cannot_find_in_scope_corrected` in here.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+    }
+    return diagnostic.message.localizedCaseInsensitiveContains("cannot find")
+      && diagnostic.message.localizedCaseInsensitiveContains("in scope")
```
ahoppen: As I mentioned in the test cases, I don’t think we should have a fallback here unless absolutely necessary. The diagnostic code should always be there and it’s way more stable.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,223 @@
...
+      )
+    }
+  }
+
+  /// Extracts existing imports from the syntax tree.
+  private static func extractExistingImports(from syntaxTree: SourceFileSyntax) -> Set<String> {
+    Set(
+      syntaxTree.statements
+        .compactMap { $0.item.as(ImportDeclSyntax.self) }
+        .compactMap { $0.path.first?.name.text }
```
ahoppen, line 101: Let’s only consider imports here that unconditionally import an entire module, eg. if we have `import struct MyLibrary.MyStruct`, we could still suggest to import `MyLibrary` for a missing type.

Sources/SKTestSupport/SwiftPMTestProject.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
PhantomInTheWire: sure makes sense.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
PhantomInTheWire: yeah that makes sense I can just get the diagnostics directly then

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
PhantomInTheWire: yep. removed the ?

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: sure. done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: yeah I forgot to fix that I think. it should work now.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: sounds like a good idea.

PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: done.

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: sure trailing closure it is

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: should be more clearer now

Tests/SourceKitLSPTests/AddMissingImportsUnitTests.swift
PhantomInTheWire: this took a while because we weren't really getting any id in any case that I tested, should be fixed now.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: fixed.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: yeah I don't know why I did that, occurrence.location.moduleName should work here always

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: makes sense. done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: makes sense. done.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
PhantomInTheWire: makes sense. done.

PhantomInTheWire:
ping @ahoppen in case this got lost

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+internal import BuildServerIntegration
```
ahoppen, line 13: You can remove the `internal` access modifiers in this file.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+  package init?(fromLSPDictionary dictionary: [String: LanguageServerProtocol.LSPAny]) {
+    guard case .string(let title)? = dictionary[CodingKeys.title.stringValue] else {
+      return nil
+    }
+    guard case .dictionary(let documentDict)? = dictionary[CodingKeys.textDocument.stringValue] else {
+      return nil
+    }
+    guard let textDocument = TextDocumentIdentifier(fromLSPDictionary: documentDict) else {
+      return nil
+    }
```
ahoppen, line 53:
```suggestion
    guard let textDocument = TextDocumentIdentifier(fromLSPAny: dictionary[CodingKeys.textDocument.stringValue]) else {
      return nil
    }
```

Sources/SwiftLanguageService/Diagnostic.swift
```diff
@@ -270,6 +274,11 @@ extension Diagnostic {
       }
     }
 
+    // If no educational note available, fall back to diagnostic ID (e.g., "cannot_find_in_scope")
+    if code == nil, let diagnosticID {
+      code = .string(diagnosticID)
```
ahoppen, line 279: The LSP documentation for `code` says 

> The diagnostic's code, which might appear in the user interface.

I don’t want to show the diagnostic IDs in the user interface, so I think we should store the diagnostic ID in the `data` dictionary.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+  ///
+  /// This is the initial code action request handler. It identifies which modules
+  /// define missing symbols and returns commands for each candidate. The actual
+  /// import insertion is deferred to `executeAddMissingImport`.
+  package func addMissingImports(_ request: CodeActionRequest) async throws -> [CodeAction] {
+    // Early exit if no relevant diagnostics
+    let relevantDiagnostics = request.context.diagnostics.filter(AddMissingImportsHelper.isMissingTypeOrValueDiagnostic)
+    guard !relevantDiagnostics.isEmpty else { return [] }
+
+    guard let buildSettings = await self.compileCommand(for: request.textDocument.uri, fallbackAfterTimeout: true),
```
ahoppen, line 80: We should only fall back to fallback build settings after timeout if we listen for updated build settings to refresh results (eg. for diagnostics). This isn’t the case here.


```suggestion
    guard let buildSettings = await self.compileCommand(for: request.textDocument.uri, fallbackAfterTimeout: false),
```

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+      !buildSettings.isFallback
+    else {
+      return []
+    }
+
+    guard
+      let workspace = await self.sourceKitLSPServer?.workspaceForDocument(uri: request.textDocument.uri),
+      let index = await workspace.index(checkedFor: .modifiedFiles)
```
ahoppen, line 88: Even if a file is modified, it’s reasonable to assume that the declarations within it are still the same, so we can relax the index check here a little bit.

```suggestion
      let index = await workspace.index(checkedFor: .deletedFiles)
```

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+          textDocument: request.textDocument,
+          moduleName: moduleName
+        )
+        codeActions.append(
+          CodeAction(
+            title: command.title,
+            kind: .quickFix,
+            diagnostics: [diagnostic],
+            edit: nil,
+            command: command.asCommand()
```
ahoppen, line 132:
```suggestion
          CodeAction(
            title: command.title,
            kind: .quickFix,
            diagnostics: [diagnostic],
            command: command.asCommand()
```

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+        reason = ""
+      }
+      logger.error("client refused to apply edit for adding import: \(reason)")
+    }
+  }
+}
+
+/// Helper enum providing static functions for the add missing imports refactoring.
+private enum AddMissingImportsHelper {
+
```
ahoppen, line 193: Superfluous newline

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+  /// Diagnostic codes that indicate a missing type or value.
+  private static let missingSymbolDiagnosticCodes: Set<String> = [
+    "cannot_find_type_in_scope",
+    "cannot_find_in_scope",
+    "cannot_find_type_in_scope_did_you_mean",
+    "cannot_find_in_scope_corrected",
+  ]
+
+  /// Regex pattern for extracting type/symbol name from diagnostic messages.
+  nonisolated(unsafe) private static let symbolNameRegex = /[Cc]annot find (?:type )?'(\w+)' in scope/
```
ahoppen, line 203: We shouldn’t suppress concurrency warnings with `nonisolated(unsafe)` here. Maybe we’d need to add a global actor that can own this regex to make it concurrency safe.

ahoppen, line 203: I think this regex might be easier to read if we used `RegexBuilder` to construct it.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+    if isSorted,
+      let insertBeforeImport = importDecls.first(where: { importDecl in
+        guard let firstPath = importDecl.path.first?.name.text else { return false }
+        return firstPath > newModule
+      })
+    {
+      return snapshot.position(of: insertBeforeImport.position)
+    } else if let lastImport = importDecls.last {
+      return snapshot.position(of: lastImport.endPosition)
+    }
```
ahoppen, line 261: I think we can always use the `isSorted` case here. That way we try and find some reasonable sorted-ish location even if a single import statement doesn’t match alphabetical order.

Sources/SwiftLanguageService/CodeActions/AddMissingImports.swift
```diff
@@ -0,0 +1,331 @@
...
+    case .number(let code): return String(code)
+    case nil: return nil
+    }
+  }
+}
+
+extension SwiftLanguageService {
+  /// Finds missing import code actions for diagnostics indicating unresolved symbols.
+  /// Exposed for unit testing.
+  package static func findMissingImports(
```
ahoppen, line 282: Am I blind or is this only called by tests? In which case we aren’t testing the actual implementation in the tests this one. I would have probably relied a lot more on tests that use `SwiftPMTestProject` instead of writing tests that exercise this alternate code path. 

That being said, if you manage to write more unit-style tests with minimal amount of alternative code, that would of course be great.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,105 @@
...
+    )
+    codeActionCapabilities.codeActionLiteralSupport = codeActionLiteralSupport
+    documentCapabilities.codeAction = codeActionCapabilities
+    return ClientCapabilities(workspace: nil, textDocument: documentCapabilities)
+  }
+
+  func testAddMissingImport() async throws {
+    // This test relies on the index, so it needs to run where we can build and index.
+    try SkipUnless.longTestsEnabled()
+
```
ahoppen, line 34: I don’t think this test is any slower than other tests that we have, relying on the index. Generally a slow test is a test that takes more than 1 or 2 seconds to execute, which I’d be surprised if this one does. So, I’d remove this check.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,105 @@
...
+    )
+
+    let (uri, positions) = try project.openDocument("main.swift")
+
+    var diagnostic: Diagnostic?
+    let diags = try await project.testClient.send(
+      DocumentDiagnosticsRequest(textDocument: TextDocumentIdentifier(uri))
+    )
+    diagnostic = diags.fullReport?.items.first { $0.message.contains("LibStruct") }
+    XCTAssert(diagnostic != nil)
```
ahoppen, line 69: Not necessary because we `XCTUnwrap` below.

Tests/SourceKitLSPTests/AddMissingImportsTests.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    // Validate the edit inline when executing the command
+    project.testClient.handleSingleRequest { (request: ApplyEditRequest) -> ApplyEditResponse in
+      XCTAssertEqual(
+        request.edit.changes,
+        [
+          uri: [
+            TextEdit(range: Range(positions["1️⃣"]), newText: "import Lib\n")
+          ]
+        ]
```
ahoppen, line 96:
```suggestion
        [
          uri: [TextEdit(range: Range(positions["1️⃣"]), newText: "import Lib\n")]
        ]
```

PR #2457, opened 2026-01-18
Handles unused imports 

### LINKED ISSUE #2335: Remove unused imports should be able to remove unused import declarations in active `#if` clauses
Currently, the `Remove unused imports` code action only removes `import` declarations that are at the top level and not nested inside `#if` clauses. We should be able to also remove `import` declarations from those `#if` branches that are active in the current build configuration. For that, SourceKit-LSP needs to know which `#if` branches are active. While we could use the active regions sourcekitd requests, I think the cleaner design would be to implement a `SwiftIfConfig.BuildConfiguration` that gets its results from sourcekitd, similar to `CompilerBuildConfiguration` in the compiler’s repo.

DPrakashhh:
Fixes #2335 
1-**Integrated** `SwiftIfConfig`: Added and linked the library in `Package.swift` and `CMakeLists.txt` to enable conditional compilation evaluation.
2-New Build Configuration: Implemented `SourceKitLSPBuildConfiguration` to provide the necessary compiler arguments and target triple for `#if` evaluation.
3-Platform-Aware Visitor: Refactored `ImportCollectorVisitor` to use `activeClause(in:)`, ensuring only active code blocks are scanned for unused imports.
4-**Verified Logic**: Added unit tests confirming correct behavior for both active and inactive branches, passing all 71 tests in the suite.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
+++ b/Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
@@ -205,7 +205,6 @@ extension SwiftLanguageService {
         snapshot: snapshot,
         activeRegions: activeRegions ?? []
       )
-      logger.debug("remove-unused-imports: collected import decls count = \(importDecls.count)")
 
       var declsToRemove: [ImportDeclSyntax] = []
 
@@ -330,52 +329,44 @@ extension SwiftLanguageService {
 
     let dict = try await send(sourcekitdRequest: \.activeRegions, skreq, snapshot: snapshot)
 
-    // Newer sourcekitd returns `results` with entries that include an `offset` and optional `is_active`.
     // Build ranges by pairing each entry with the next entry's offset; include only active ones.
     guard let results = dict[keys.results] as SKDResponseArray? else {
       return nil
     }
 
-    // Collect entries into an ordered array
-    var entries: [(offset: Int, isActive: Bool?)] = []
-    results.forEach { _, entry in
-      let off = entry[keys.offset] as Int?
+    let entries: [(offset: Int, isActive: Bool?)] = results.compactMap { entry in
+      guard let off = entry[keys.offset] as Int? else {
+        return nil
+      }
       let activeBool = entry[keys.isActive] as Bool?
       let activeInt = entry[keys.isActive] as Int?
-      if let off {
-        entries.append((offset: off, isActive: activeBool ?? activeInt.map { $0 != 0 }))
-      }
-      return true
+      return (offset: off, isActive: activeBool ?? activeInt.map { $0 != 0 })
     }
     guard !entries.isEmpty else { return [] }
-    entries.sort { $0.offset < $1.offset }
+    let sortedEntries = entries.sorted { $0.offset < $1.offset }
 
     var ranges: [SourceRange] = []
     let fileEnd = snapshot.text.utf8.count
-    // sourcekitd omits `is_active` for inactive regions; default to inactive when it is missing.
-    var currentIsActive = entries.first?.isActive ?? false
-    for i in 0..<entries.count {
-      let start = entries[i].offset
-      let end = i + 1 < entries.count ? entries[i + 1].offset : fileEnd
-      if let explicit = entries[i].isActive {
-        currentIsActive = explicit
-      } else {
-        currentIsActive = false
-      }
-
-      logger.debug(
-        "activeRegions: entry offset=\(start) isActive=\(String(describing: entries[i].isActive)) end=\(end)"
-      )
-
-      if currentIsActive, start < end {
-        ranges.append(SourceRange(start: start, end: end))
+    for (current, next) in zip(sortedEntries, sortedEntries.dropFirst()) {
+      let start = current.offset
+      let end = next.offset
+      let isActive = current.isActive ?? false
+      if isActive, start < end {
+        ranges.append(AbsolutePosition(utf8Offset: start)..<AbsolutePosition(utf8Offset: end))
       }
     }
-    logger.debug("activeRegions: computed \(ranges.count) active ranges")
+    if let last = sortedEntries.last {
+      let start = last.offset
+      let end = fileEnd
+      let isActive = last.isActive ?? false
+      if isActive, start < end {
+        ranges.append(AbsolutePosition(utf8Offset: start)..<AbsolutePosition(utf8Offset: end))
+      }
+    }
     return ranges
   }
 
-  /// Collects all import declarations that are accessible.
+  /// Collects all import declarations that are active.
   /// This includes top-level imports and imports from #if clauses.
   ///
   /// - Parameters:
@@ -400,8 +391,6 @@ private class ImportCollectorVisitor: SyntaxVisitor {
   private(set) var collectedImports: [ImportDeclSyntax] = []
   private let activeRegions: [SourceRange]
   private let snapshot: DocumentSnapshot
-  /// Track the depth of #if nesting - 0 means we're at the top level
-  private var ifConfigDepth = 0
 
   init(activeRegions: [SourceRange], snapshot: DocumentSnapshot, viewMode: SyntaxTreeViewMode = .sourceAccurate) {
     self.activeRegions = activeRegions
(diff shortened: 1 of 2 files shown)


DPrakashhh:
Thanks for the feedback! I implemented the previous version following the earlier suggestion to use a configuration evaluator, but I completely agree that relying on `sourcekitd` for `activeRegions` is a much more robust and cleaner architectural path.

I’ll refactor the implementation to use the `activeRegions` request to identify live source ranges instead. I'll update the PR as soon as it's ready.

DPrakashhh:
Hi, I've refactored the implementation to use `sourcekitd’s` `active_regions` API, removing the manual flag-parsing logic and ensuring all `CodeActionTests` (including `canImport` cases) now pass correctly.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -197,9 +197,15 @@ extension SwiftLanguageService {
...
-      let importDecls = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
+      // Fetch active regions from sourcekitd before collecting imports
+      let activeRegions = try await fetchActiveRegions(snapshot: snapshot, compileCommand: compileCommand)
+
+      let importDecls = try await collectImportDecls(
+        from: syntaxTree,
+        snapshot: snapshot,
+        activeRegions: activeRegions ?? []
+      )
+      logger.debug("remove-unused-imports: collected import decls count = \(importDecls.count)")
```
ahoppen: I don’t see how the number of unused imports will be useful log output. If anything, you’ll want to know which ones were identified as unused. But that should already be recorded in the LSP response that we’ll send to the client, so I don’t think this provides much value.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+    var entries: [(offset: Int, isActive: Bool?)] = []
+    results.forEach { _, entry in
+      let off = entry[keys.offset] as Int?
+      let activeBool = entry[keys.isActive] as Bool?
+      let activeInt = entry[keys.isActive] as Int?
+      if let off {
+        entries.append((offset: off, isActive: activeBool ?? activeInt.map { $0 != 0 }))
+      }
+      return true
+    }
```
ahoppen: You should be able to use `map` here, which will make this look a little nicer.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+      let end = i + 1 < entries.count ? entries[i + 1].offset : fileEnd
+      if let explicit = entries[i].isActive {
+        currentIsActive = explicit
+      } else {
+        currentIsActive = false
+      }
+
+      logger.debug(
+        "activeRegions: entry offset=\(start) isActive=\(String(describing: entries[i].isActive)) end=\(end)"
+      )
```
ahoppen: This will already be included in the logged sourcekitd response, so it’s superfluous.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+      return true
+    }
+    guard !entries.isEmpty else { return [] }
+    entries.sort { $0.offset < $1.offset }
+
+    var ranges: [SourceRange] = []
+    let fileEnd = snapshot.text.utf8.count
+    // sourcekitd omits `is_active` for inactive regions; default to inactive when it is missing.
+    var currentIsActive = entries.first?.isActive ?? false
+    for i in 0..<entries.count {
```
ahoppen: You should be able to handle this without an index using something like `zip(entries, entries.dropFirst())`. You’d just need to handle the last one outside the loop but you already have a special case for that, so that shouldn’t be an issue.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+    let skreq = sourcekitd.dictionary([
+      keys.sourceFile: snapshot.uri.sourcekitdSourceFile,
+      keys.primaryFile: snapshot.uri.primaryFile?.pseudoPath,
+      keys.compilerArgs: compileCommand.compilerArgs as [any SKDRequestValue],
+    ])
+
+    let dict = try await send(sourcekitdRequest: \.activeRegions, skreq, snapshot: snapshot)
+
+    // Newer sourcekitd returns `results` with entries that include an `offset` and optional `is_active`.
+    // Build ranges by pairing each entry with the next entry's offset; include only active ones.
```
ahoppen, line 332: AFAIK the result structure has always been the same, so I don’t think this comment is correct. Or am I missing something?

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+
+      logger.debug(
+        "activeRegions: entry offset=\(start) isActive=\(String(describing: entries[i].isActive)) end=\(end)"
+      )
+
+      if currentIsActive, start < end {
+        ranges.append(SourceRange(start: start, end: end))
+      }
+    }
+    logger.debug("activeRegions: computed \(ranges.count) active ranges")
```
ahoppen: Similarly here, this is superfluous information that just adds noise to the log.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+    return ranges
+  }
+
+  /// Collects all import declarations that are accessible.
```
ahoppen:
```suggestion
  /// Collects all import declarations that are active.
```

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+    }
+    return false
+  }
+}
+
+/// Represents a source range with start and end byte offsets.
+private struct SourceRange {
+  let start: Int
+  let end: Int
+}
```
ahoppen: Could you use `Range<AbsolutePosition>` instead of defining your custom type for this?

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
```diff
@@ -301,4 +307,143 @@ extension SwiftLanguageService {
...
+
+  init(activeRegions: [SourceRange], snapshot: DocumentSnapshot, viewMode: SyntaxTreeViewMode = .sourceAccurate) {
+    self.activeRegions = activeRegions
+    self.snapshot = snapshot
+    super.init(viewMode: viewMode)
+  }
+
+  override func visit(_ node: ImportDeclSyntax) -> SyntaxVisitorContinueKind {
+    let startOffset = snapshot.utf8Offset(of: snapshot.position(of: node.positionAfterSkippingLeadingTrivia))
+    if ifConfigDepth == 0 || isOffsetInActiveRegion(startOffset) {
```
ahoppen: Do we need to keep track if the `ifConfigDepth`? Even for `import` statements outside of `#if`, they should be in an active region, right?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1545,6 +1545,92 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+      // Should remove Darwin and LibA/LibB imports, but NOT the Glibc import (which is in an inactive #if clause)
+      guard let changesDict = request.edit.changes,
+        let changes = changesDict[uri]
+      else {
+        XCTFail("Expected edits for the test file")
+        return ApplyEditResponse(applied: false, failureReason: "No edits")
+      }
+
+      // Verify we have 3 edits (Darwin, LibA, LibB)
+      XCTAssertEqual(changes.count, 3, "Expected 3 import removals")
```
ahoppen, line 1611: Let’s have an explicit `XCTAssert` check for `changes`. I think that should cover everything.

Sources/SwiftLanguageService/CodeActions/RemoveUnusedImports.swift
DPrakashhh: I'll update the comment to simply describe the current structure of the response and focus on the offset-pairing logic instead.

DPrakashhh:
I’ve refactored the implementation to address all feedback:

1-**Simplified Architecture:** the visitor now uses parent lookup to gate `#if` imports against `activeRegions`.
2-**Swifty Refinements:** Adopted `compactMap` for parsing and `zip(entries, entries.dropFirst())` for index-free range building.
3-**Cleanup**: Removed all superfluous debug logging, updated doc comments, and swapped to native `Range<AbsolutePosition>`.
4-**Explicit Tests**: Added detailed assertions to verify the specific imports targeted for removal.

Ready for final review!

ahoppen:
I will continue reviewing this PR after [your PR to indexstore-db](https://github.com/swiftlang/indexstore-db/pull/266) is ready to be merged.

PR #2458, opened 2026-01-18
Report convert-to-computed and convert-to-stored refactoring actions

ahoppen:
These were implemented in swift-syntax but never added to the list of refactoring actions reported by SourceKit-LSP.

While at this, also fix a few other things:
- Report an error if we non-exhaustively checked that we had an empty array of code actions, since this was a tautology
- Allow ignoring code actions that we don’t want to check for (needed to write sensible exhaustive checks)
- Fixed a bug that caused convert-to-zero-parameter-function to be reported when on a return statement in an accessor block
- Make it easier to implement `SyntaxRefactoringCodeActionProvider` for refactoring actions that take a non-Void context. Not needed at the moment but may be useful in the future. I briefly needed this when trying to bring the expand-placeholder code action to LSP before I realized that there are no `<##>` placeholders created by SourceKit-LSP – they get translated to LSP placeholders.
- Sort the list of extensions that conform refactoring actions in swift-syntax to `SyntaxRefactoringCodeActionProvider` alphabetically

### DIFF
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1931,7 +1931,7 @@ final class CodeActionTests: SourceKitLSPTestCase {
           )
         }
         XCTAssert(
-          codeActions.contains(expected),
+          Set(codeActions).isSuperset(of: expected),
           """
           Code actions did not contain expected at range \(startMarker)-\(endMarker):
           \(codeActions)


ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1734,18 +1827,28 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+        if expected.isEmpty {
+          XCTFail(
+            "exhaustive: false is incompatible with an empty list of expected code actions",
+            file: file,
+            line: line
+          )
+        }
         XCTAssert(
-          codeActions.contains(expected(uri, positions)),
+          codeActions.contains(expected),
```
rintaro: This is not about this PR but `Array.contains(_: some Collection)` checks for a subsequence. I.e. `[1,2,3,4].contains([1,3])` is false. I don't think that's what was intended here. Instead, it should:
```swift
Set(codeActions).isSuperset(of: expected)
```
Could you fix it while you are here?

ahoppen: Oh, good catch. Thank you!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

PR #2460, opened 2026-01-18
Match enabled upcoming features from Package manifest in CMakeLists.txt

ahoppen:
These drifted out-of-sync. Bring them in sync again.


ahoppen:
@swift-ci Please test

PR #2462, opened 2026-01-27
FreeBSD: Include freebsd runtimes in rpath

etcwilde:
Running sourcekit-lsp built by Swift CI on FreeBSD results in the following error:

```
ld-elf.so.1: Shared object "libswiftCore.so" not found, required by "sourcekit-lsp"
```

The runpath is pointing at a location in CI rather than the binary-relative location where we can find the runtime libraries.


etcwilde:
@swift-ci please test

Issue #2381, opened 2025-12-08
Option to have inlay hints for trailing closures


RowSalmon:
### Description

Hello, im hoping for inlay hints to be shown for trailing closures. I often have trouble working out what parameter the trailing closure is for - and an inlay hint would be a great way to help me out here. I'm sure other beginners learning swift would also appreciate this feature.

An example (from Apples' NavigationCookBook example project):

```swift
        .sheet(isPresented: $navigationModel.showExperiencePicker) {
            ExperiencePicker(experience: $experience)
        }

```

The function signature for `.sheet` is

```swift
func sheet<Content>(
    isPresented: Binding<Bool>,
    onDismiss: (() -> Void)? = nil,
    @ViewBuilder content: @escaping () -> Content
) -> some View where Content : View
```

Its not immediately clear to a beginner whether the trailing closure relates to `onDismiss` or `content`.

The enhancement here would be to add an inlay hint as follows:

```swift
        .sheet(isPresented: $navigationModel.showExperiencePicker) : content {
            ExperiencePicker(experience: $experience)
        }

```

Those proficient with swift may find inlay hints here unnecessarily verbose - and that is why i'm proposing this be an optional toggle feature (consistent with how other LSPs allow granular control over what type of inlay hints are shown and which ones are not)

Thank you for your consideration.

PR #2464, opened 2026-01-27
feat: add inlay hints for trailing closure labels

DPrakashhh:
**Overview**
Implements Inlay Hints for trailing closures (issue #2381). This adds optional visual labels like `: body` or `: content` to improve readability in DSLs and SwiftUI.

**Key Changes**
1-**Opt-in Toggle**: Added `trailingClosureLabels` to `SourceKitLSPOptions` (disabled by default).
2-**Logic**: Added `TrailingClosureInlayHints.swift` to resolve parameter names from function signatures.
3-**Performance**: Uses the asynchronous `InlayHint` protocol to ensure no impact on editor latency.

**Verification**
1-**Automated Tests**: 12/12 tests passed in `TrailingClosureInlayHintTests.swift`. Covers `forEach`, SwiftUI modifiers, and nested closures.
2-**Manual Verification**: Confirmed the `trailingClosureLabels` setting correctly toggles the feature in the IDE.

**Note**: Full verification is provided by the unit tests; local IDE indexing was still finalizing during manual UI capture.

### DIFF
--- a/Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
+++ b/Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
@@ -83,7 +83,7 @@ extension SwiftLanguageService {
         // Check if the closure is within the requested range
         if let range {
           let openingBracePosition = snapshot.position(
-            of: closureInfo.openingBrace.endPositionBeforeTrailingTrivia
+            of: closureInfo.openingBrace.positionAfterSkippingLeadingTrivia
           )
           guard openingBracePosition >= range.lowerBound && openingBracePosition < range.upperBound else {
             continue
@@ -95,7 +95,7 @@ extension SwiftLanguageService {
           for: closureInfo.functionCall,
           in: snapshot
         ) {
-          let hintPosition = snapshot.position(of: closureInfo.openingBrace.endPositionBeforeTrailingTrivia)
+          let hintPosition = snapshot.position(of: closureInfo.openingBrace.positionAfterSkippingLeadingTrivia)
           let label = ": \(parameterLabel)"
 
           let hint = InlayHint(
@@ -118,8 +118,9 @@ extension SwiftLanguageService {
 
   /// Retrieves the parameter label for a trailing closure in a function call.
   ///
-  /// This queries sourcekitd to determine the function's signature and identifies
-  /// the parameter that the trailing closure satisfies.
+  /// Uses sourcekitd's signatureHelp to get structured parameter information,
+  /// then determines which parameter the trailing closure satisfies based on
+  /// the number of labeled arguments.
   ///
   /// - Parameters:
   ///   - functionCall: The function call expression containing the trailing closure.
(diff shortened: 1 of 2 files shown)


Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
```diff
@@ -0,0 +1,229 @@
...
+            continue
+          }
+        }
+
+        // Try to get the parameter label from the function signature
+        if let parameterLabel = await getTrailingClosureParameterLabel(
+          for: closureInfo.functionCall,
+          in: snapshot
+        ) {
+          let hintPosition = snapshot.position(of: closureInfo.openingBrace.endPositionBeforeTrailingTrivia)
```
PhantomInTheWire: Use closureInfo.openingBrace.positionAfterSkippingLeadingTrivia instead of endPositionBeforeTrailingTrivia to position the hint before the opening brace.

Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
```diff
@@ -0,0 +1,229 @@
...
+        let paramRange = Range(match.range(at: 1), in: signature)
+      {
+        let paramName = String(signature[paramRange])
+        return paramName
+      }
+    }
+
+    return nil
+  }
+}
```
PhantomInTheWire: this seems like a very brittle and unreliable approach to do things here.
 
1. It checks for hardcoded common names like content, label, body, completion, etc. in order of priority. If a function has both a content and completion parameter, it would always return content regardless of which is the trailing closure.
2. The regex patterns doesn't account for all closure type variations (e.g., optional 	closures (() -> Void)?, autoclosures @ autoclosure, or closures with throws/async).

3. The approach parses XML/annotated declaration strings with regex rather than using structured data from sourcekitd, this is inherently brittle and very hard to maintain.

Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
```diff
@@ -0,0 +1,229 @@
...
+
+
+func testHint() {
+  let numbers = [1, 2]
+  numbers.forEach { number in  
+    print(number)
+  }
+}
```
PhantomInTheWire:
```suggestion
```

DPrakashhh:
Thank you for the detailed feedback. I completely see your point about the brittleness of the regex approach and the risk of misidentifying parameters based on hardcoded names. I initially used this to verify the UI/UX flow, but I agree it should be replaced with structured data from sourcekitd to be reliable.

Sources/SwiftLanguageService/TrailingClosureInlayHints.swift
DPrakashhh: You're absolutely right. That `testHint()` function was a local sandbox I used for manual verification and was not intended to be part of the final implementation. I have removed the entire block to keep the source file clean. Thank you for catching that!

PhantomInTheWire:
can you please tell me why this can't be structured then? why is it this way?

DPrakashhh:
I hear you on the previous approach—it was definitely brittle. I’ve gone ahead and completely refactored the implementation to move away from string parsing and heuristics entirely.

**Here’s the new approach:**

1-**Structured Data**: I’m now using `sourcekitd.signatureHelp` to get the actual `parameters` array directly from the compiler. This handles all the edge cases you mentioned (async, throws, optional closures) automatically because we're relying on the compiler's own model of the function.
2-**Semantic Mapping**: Instead of guessing the name, I’m resolving the label by matching the trailing closure’s position against the `labeledArguments` count. This follows Swift’s actual rules for which parameter a trailing closure satisfies.
3-**Cleaner Positioning**: I switched to `positionAfterSkippingLeadingTrivia` so the hint is anchored directly to the opening brace `{` node in the syntax tree, making it much more resilient to whitespace or trivia changes.

I’ve verified this with 13 tests covering everything from basic SwiftUI modifiers to nested closures and complex generics, and they're all green. I also made sure to clean up the `testHint()` sandbox code I had in there for manual testing.

Let me know if this looks better to you now?

DPrakashhh:
That makes total sense. I moved to `signatureHelp` specifically to address the correctness issues and 'brittleness' flagged earlier in the review, as it was the only structured way I found to map labels to trailing closures accurately.

I definitely see the performance risk with per-closure requests in large files. Since a bulk request for this doesn't exist yet, would you recommend I stick with this for now as a baseline, or should I look into a different approach—perhaps leveraging something like `source.request.variable.type` or another batch-friendly request to keep things performant?

ahoppen:
I will continue reviewing this PR after [your PR to indexstore-db](https://github.com/swiftlang/indexstore-db/pull/266) is ready to be merged.

Issue #2463, opened 2026-01-27
Support textDocument/selectionRange


Steffeeen:
### Description

The [textDocument/selectionRange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) capability allows the language server to provide language specific selection ranges when using the [Shrink/expand selection](https://code.visualstudio.com/docs/editing/codebasics#_shrinkexpand-selection) shortcuts. Due this capability currently not being implemented editors have to fall back to guessing the ranges by looking at the tokens. This leads to suboptimal results in some cases.

Examples:

https://github.com/user-attachments/assets/c3bc6592-0104-47c9-aa22-68a3761bbd34

Here, an additional selection range for the parameter label and value would be useful.

https://github.com/user-attachments/assets/7956735d-51d8-4c6f-9e59-f83339d4918a

Additional selection ranges for `abc.def` and `let test = abc.def` before selecting the entire if expression would be useful.

These two examples illustrate the problem, but there are a lot more cases where the selection behavior is currently suboptimal.

I have already started implementing this and want to provide a PR as soon as I'm done.

Steffeeen:
Implemented in #2467

PR #2467, opened 2026-01-29
Implement support for textDocument/selectionRange

Steffeeen:
This PR implement the [textDocument/selectionRange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) capability providing better support for the [Shrink/expand selection](https://code.visualstudio.com/docs/editing/codebasics#_shrinkexpand-selection) shortcuts and addresses #2463.

# Comparison
Before:

https://github.com/user-attachments/assets/743a51e0-afbe-483e-8dcf-8d6bec132c76

After:

https://github.com/user-attachments/assets/a42c99f8-a0fa-4418-8168-e5e283917b1a

Before:

https://github.com/user-attachments/assets/d188bfe2-509e-4fc5-9ee5-0f4284326969

After:

https://github.com/user-attachments/assets/00b3630f-04c3-44c7-b2a2-195cb3ecbfa4

# Implementation

The feature works by getting the AST node under the cursor and then creating selection ranges for nodes by walking up the AST to the root. A range that is the same as the last one that was added is automatically filtered out. Additionally, there are special cases for a lot of AST nodes, depending on whether they should never create a selection range or need special handling in that they may create more than one selection range.

I also added handling for when the cursor is directly at a boundary token, e.g. `test(a: 1, b: 2|)`, in this case the user most likely wants to select the `2` even tough the cursor is technically at the `)` token. For this we check if we are at a boundary token and if we are, we use the previous token as the start of our AST traversal.

Most of the unit tests where initially generated by Claude Sonnet 4.5. I have, however, looked over all of them to ensure that they are correct and match the behavior I intended.

# Additional Notes

I am aware that this is a big PR, note however that ~2300 of the added lines are for tests.

I intend to squash a lot of the commits, however I'm unsure if everything should be squashed into one big commit or if multiple commits would be helpful. One approach could be to modify the first commit to only contain the bare-bones logic for walking up the AST nodes and creating selection ranges for all nodes and the basic setup for the unit tests. A second commit could then contain all the unit tests while the third commit introduces all the special cases for the different AST nodes. Commits two and three could also be swapped to avoid having a commit (the commit with the tests) in the history with a bunch of failing unit tests.  

As I was not familiar with the swift-syntax API before working on this I may not have used the API in the most ideal way in some cases. I'm open to feedback on that.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -820,6 +820,8 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await request.reply { try await executeCommand(request.params) }
     case let request as RequestAndReply<FoldingRangeRequest>:
       await self.handleRequest(for: request, requestHandler: self.foldingRange)
+    case let request as RequestAndReply<SelectionRangeRequest>:
+      await self.handleRequest(for: request, requestHandler: self.selectionRange)
     case let request as RequestAndReply<GetReferenceDocumentRequest>:
       await request.reply { try await getReferenceDocument(request.params) }
     case let request as RequestAndReply<HoverRequest>:
@@ -1180,6 +1182,7 @@ extension SourceKitLSPServer {
       typeHierarchyProvider: .bool(true),
       semanticTokensProvider: semanticTokensOptions,
       inlayHintProvider: inlayHintOptions,
+      selectionRangeProvider: .bool(true),
       experimental: .dictionary(experimentalCapabilities)
     )
   }
@@ -1804,6 +1807,14 @@ extension SourceKitLSPServer {
     languageService: any LanguageService
   ) async throws -> [FoldingRange]? {
     return try await languageService.foldingRange(req)
+  }
+
+  func selectionRange(
+    _ req: SelectionRangeRequest,
+    workspace: Workspace,
+    languageService: any LanguageService
+  ) async throws -> [SelectionRange] {
+    return try await languageService.selectionRange(req)
   }
 
   func documentSymbol(
--- a/Sources/SwiftLanguageService/SelectionRange.swift
+++ b/Sources/SwiftLanguageService/SelectionRange.swift
@@ -1,0 +1,605 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) package import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+import SourceKitLSP
+import SwiftOperators
+import SwiftSyntax
+
+extension SwiftLanguageService {
+  package func selectionRange(_ req: SelectionRangeRequest) async throws -> [SelectionRange] {
+    let snapshot = try self.documentManager.latestSnapshot(req.textDocument.uri)
+    let sourceFile = await syntaxTreeManager.syntaxTree(for: snapshot)
+
+    try Task.checkCancellation()
+
+    return req.positions.map { position in
+      let absolutePosition = snapshot.absolutePosition(of: position)
+
+      guard let (token, newPosition) = findIntuitiveToken(in: sourceFile, at: absolutePosition) else {
+        return SelectionRange(range: position..<position)
+      }
+
+      guard
+        let selectionRange = computeSelectionRangeFor(
+          position: newPosition,
+          snapshot: snapshot,
+          node: Syntax(token)
+        )
+      else {
+        return SelectionRange(range: position..<position)
+      }
+
+      return selectionRange
+    }
+  }
+}
+
+private func findIntuitiveToken(
+  in sourceFile: SourceFileSyntax,
+  at position: AbsolutePosition
+) -> (TokenSyntax, AbsolutePosition)? {
+  guard let currentToken = sourceFile.token(at: position) else {
+    if (sourceFile.endPositionBeforeTrailingTrivia...sourceFile.endPosition).contains(position),
+      // The last token is EOF, so we use the token just before EOF
+      let newToken = sourceFile.lastToken(viewMode: .sourceAccurate)?.previousToken(viewMode: .sourceAccurate)
+    {
+      return (newToken, newToken.endPosition.advanced(by: -1))
+    }
+    return nil
+  }
+
+  let boundaryTokens: [TokenKind] = [
+    .leftParen, .rightParen,
+    .leftBrace, .rightBrace,
+    .leftSquare, .rightSquare,
+    .leftAngle, .rightAngle,
+    .comma, .period,
+    .semicolon, .colon,
+  ]
+
+  if position == currentToken.position && boundaryTokens.contains(currentToken.tokenKind) {
+    // The cursor is at the start of a boundary token (e.g. `test(a: 3, b: 2|)`)
+    // here the user most likely wants to select the `2` and then `b: 2` instead of
+    // selecting the entire function call, so we use the previous token
+
+    let newToken = currentToken.previousToken(viewMode: .sourceAccurate) ?? currentToken
+
+    if newToken.endPositionBeforeTrailingTrivia == position {
+      return (newToken, newToken.endPosition.advanced(by: -1))
+    }
+  }
+
+  return (currentToken, position)
+}
+
+private func computeSelectionRangeFor(
+  position: AbsolutePosition,
+  snapshot: DocumentSnapshot,
+  node: Syntax
+) -> SelectionRange? {
+  var ranges: [Range<AbsolutePosition>] = []
+
+  var current = node
+  while true {
+    let rangesForNode = calculateRangesFor(node: current, snapshot: snapshot, position: position)
+
+    for range in rangesForNode {
+      if ranges.last == range {
+        // Some AST nodes have the exact same range, we just skip creating ranges for them
+        continue
+      }
+
+      ranges.append(range)
+    }
+
+    guard let parent = current.parent else {
+      break
+    }
+
+    current = parent
+  }
+
+  var selectionRange: SelectionRange? = nil
+  for range in ranges.reversed() {
+    let start = snapshot.position(of: range.lowerBound)
+    let end = snapshot.position(of: range.upperBound)
+    selectionRange = SelectionRange(range: start..<end, parent: selectionRange)
+  }
+
+  return selectionRange
+}
+
+private func calculateRangesFor(
+  node: Syntax,
+  snapshot: DocumentSnapshot,
+  position: AbsolutePosition
+) -> [Range<AbsolutePosition>] {
+  if let stringSegment = node.as(StringSegmentSyntax.self) {
+    return calculateSelectionRangesForStringSegment(stringSegment: stringSegment, position: position)
+  }
+
+  if node.isProtocol((any DeclGroupSyntax).self)
+    || node.is(TypeAliasDeclSyntax.self)
+    || node.is(FunctionDeclSyntax.self)
+  {
+    let name = Syntax(node.asProtocol((any NamedDeclSyntax).self)?.name)
+    let type = Syntax(node.as(ExtensionDeclSyntax.self)?.extendedType)
+    let genericParameterClause = node.asProtocol((any WithGenericParametersSyntax).self)?.genericParameterClause
+
+    if let nameOrType = name ?? type {
+      return calculateRangesForTypeOrFunctionDeclaration(
+        declaration: node,
+        position: position,
+        nameOrType: nameOrType,
+        genericParameters: genericParameterClause
+      )
+    }
+  }
+
+  if let provider = node.asProtocol((any SyntaxProtocol).self) as? (any SelectionRangeProvider) {
+    return provider.calculateSelectionRanges(position: position)
+  }
+
+  return []
+}
+
+private func calculateSelectionRangesForStringSegment(
+  stringSegment: StringSegmentSyntax,
+  position: AbsolutePosition
+) -> [Range<AbsolutePosition>] {
+  // For string segments we first want to select just the word under the cursor.
+  // To determine words we use a simple heuristic: expand the selection until we hit any non-letter character.
+  let offsetInString = position.utf8Offset - stringSegment.positionAfterSkippingLeadingTrivia.utf8Offset
+
+  let text = stringSegment.content.text
+  let index = text.utf8.index(text.startIndex, offsetBy: offsetInString)
+
+  if !text[index].isLetter {
+    return []
+  }
+
+  let start = text[..<index].lastIndex(where: { !$0.isLetter }).flatMap { text.index(after: $0) } ?? text.startIndex
+  let end = text[index...].firstIndex(where: { !$0.isLetter }) ?? text.endIndex
+
+  let startOffsetInString = text.utf8.distance(from: text.startIndex, to: start)
+  let endOffsetInString = text.utf8.distance(from: text.startIndex, to: end)
+
+  let startPosition = stringSegment.positionAfterSkippingLeadingTrivia.advanced(by: startOffsetInString)
+  let endPosition = stringSegment.positionAfterSkippingLeadingTrivia.advanced(by: endOffsetInString)
+
+  return [startPosition..<endPosition]
+}
+
+private func calculateRangesForTypeOrFunctionDeclaration(
+  declaration: some SyntaxProtocol,
+  position: AbsolutePosition,
+  nameOrType: some SyntaxProtocol,
+  genericParameters: GenericParameterClauseSyntax?,
+) -> [Range<AbsolutePosition>] {
+  // If we started the selection in either the name (or type in the case of extensions)
+  // or the generic parameter clause of a declaration we want to have a selection range
+  // for just the name (type) and generic parameter clause.
+  //
+  // Example: given `struct Test<|T> {}` we want to have a selection range for `Test<T>`
+  //
+  // As the name (type) and generic parameter clause are siblings in the declaration nodes
+  // we have to special-case them.
+
+  var ranges: [Range<AbsolutePosition>] = []
+
+  if let genericParameters = genericParameters,
+    nameOrType.range.contains(position) || genericParameters.range.contains(position)
+  {
+    let start = nameOrType.positionAfterSkippingLeadingTrivia
+    let end = genericParameters.endPositionBeforeTrailingTrivia
+    ranges.append(start..<end)
+  }
+
+  ranges.append(declaration.trimmedRange)
+
+  return ranges
+}
+
+private protocol SelectionRangeProvider: SyntaxProtocol {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>]
+}
+
+extension FunctionCallExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    if let memberAccess = self.calledExpression.as(MemberAccessExprSyntax.self),
+      !(self.parent?.is(ExpressionPatternSyntax.self) ?? false),
+      (memberAccess.declName.position..<self.additionalTrailingClosures.endPosition).contains(position)
+    {
+      // Special case for adding an extra range including the function name and parameters/trailing closures
+      // this is needed for chained method calls
+      // Example:
+      // numbers
+      //  .filter { $0 > 0 }
+      //  .map { $0 * 2 }
+      //  .reduce(0,| +)
+      //
+      // When starting a selection from | we want to have a selection for `reduce(0, +)` in addition to selecting
+      // the entire function call (starting from `numbers`)
+      return [
+        memberAccess.declName.positionAfterSkippingLeadingTrivia..<self.endPositionBeforeTrailingTrivia,
+        self.trimmedRange,
+      ]
+    }
+
+    // the default case: just create a range for the function call node
+    return [self.trimmedRange]
+  }
+}
+
+extension SubscriptCallExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For subscript calls we want to have a selection range for the entire subscript operator
+    // including the `[]`
+    // Example: given `matrix[2, |3]` we want to be able to select `[2, 3]`
+
+    if self.arguments.range.contains(position) {
+      let start = self.leftSquare.positionAfterSkippingLeadingTrivia
+      let end = self.rightSquare.endPositionBeforeTrailingTrivia
+      return [start..<end, self.trimmedRange]
+    }
+
+    return [self.trimmedRange]
+  }
+}
+
+extension LabeledExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For labeled expressions we want to be able to select just the label and expression without the comma
+    let start = self.positionAfterSkippingLeadingTrivia
+    let end = self.expression.endPositionBeforeTrailingTrivia
+
+    return [start..<end]
+  }
+}
+
+extension GenericParameterSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // Don't include the trailing comma in the selection, except if the parameter is the only one
+    if let parameterList = self.parent?.as(GenericParameterListSyntax.self),
+      parameterList.count == 1
+    {
+      return [self.trimmedRange]
+    }
+
+    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+    return [self.positionAfterSkippingLeadingTrivia..<end]
+  }
+}
+
+extension FunctionParameterSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // Function parameters have two special cases:
+    // - If the cursor is in the type and the type includes an ellipsis we want to have a range for the type and ellipsis
+    // - If the parameter has two names we want to have a range for selecting both names
+    let start = self.positionAfterSkippingLeadingTrivia
+    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+    let rangeWithoutComma = start..<end
+
+    if type.range.contains(position) {
+      if let ellipsis = self.ellipsis {
+        // Add an additional range for selecting the ellipsis of variadic parameters.
+        let range = self.type.positionAfterSkippingLeadingTrivia..<ellipsis.endPositionBeforeTrailingTrivia
+        return [range, rangeWithoutComma]
+      }
+      return [rangeWithoutComma]
+    }
+
+    if let defaultValue = self.defaultValue, defaultValue.range.contains(position) {
+      return [rangeWithoutComma]
+    }
+
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let secondName = self.secondName {
+      // If the parameter has two names, add an additional range for selecting both names
+      let range = self.firstName.positionAfterSkippingLeadingTrivia..<secondName.endPositionBeforeTrailingTrivia
+      if range.contains(position) {
+        ranges.append(range)
+      }
+    }
+
+    ranges.append(rangeWithoutComma)
+    return ranges
+  }
+}
+
+extension ClosureExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let signature = self.signature, signature.range.contains(position) {
+      let start = signature.positionAfterSkippingLeadingTrivia
+      let end = self.statements.endPositionBeforeTrailingTrivia
+      ranges.append(start..<end)
+    }
+
+    ranges.append(self.trimmedRange)
+    return ranges
+  }
+}
+
+extension EnumCaseParameterSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For enum case parameters we also add a range for selecting the names of the parameter, similar to function parameters
+
+    // This implementation is really similar to the one for FunctionParameterSyntax,
+    // except that we don't have to deal with ellipses and have to deal with unlabeled parameters
+    let start = self.positionAfterSkippingLeadingTrivia
+    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+    let rangeWithoutComma = start..<end
+
+    if self.type.range.contains(position) {
+      return [rangeWithoutComma]
+    }
+
+    if let defaultValue = self.defaultValue, defaultValue.range.contains(position) {
+      return [rangeWithoutComma]
+    }
+
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let firstName = self.firstName,
+      let secondName = self.secondName
+    {
+      // The parameter has two names, add a selection range for selecting both names
+      let range = firstName.positionAfterSkippingLeadingTrivia..<secondName.endPositionBeforeTrailingTrivia
+      if range.contains(position) {
+        ranges.append(range)
+      }
+    }
+
+    ranges.append(rangeWithoutComma)
+
+    return ranges
+  }
+}
+
+extension ExprListSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // Expression lists are more complex to deal with as we first have to convert the expression list into the
+    // corresponding tree. We can then find the node the cursor is on in the tree and walk the tree up to its root
+    guard let sequenceExpression = self.parent?.as(SequenceExprSyntax.self) else {
+      return [self.trimmedRange]
+    }
+
+    let table = OperatorTable.standardOperators
+    let foldedTree = orLog("Folding ExprListSyntax") {
+      try table.foldSingle(sequenceExpression)
+    }
+
+    guard let foldedTree = foldedTree else { return [] }
+
+    let foldedTreeOffset = SourceLength(utf8Length: sequenceExpression.position.utf8Offset)
+    let offsetInTree = position - foldedTreeOffset
+
+    guard var operandNode = foldedTree.token(at: offsetInTree)?.parent else {
+      return []
+    }
+
+    // Walk up from the token to the operand node
+    // This is needed to avoid processing everything below the operand node two times, as everything below has already
+    // been processed by the normal logic before hitting the ExprListSyntax
+    while let parent = operandNode.parent {
+      if parent.is(InfixOperatorExprSyntax.self) {
+        break
+      }
+
+      operandNode = parent
+    }
+
+    var ranges: [Range<AbsolutePosition>] = []
+
+    for node in sequence(first: operandNode, next: \.parent) {
+      let startPosition = node.positionAfterSkippingLeadingTrivia + foldedTreeOffset
+      let endPosition = node.endPositionBeforeTrailingTrivia + foldedTreeOffset
+      ranges.append(startPosition..<endPosition)
+    }
+
+    return ranges
+  }
+}
+
+extension PatternBindingSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For pattern bindings we special-case depending on whether we have a single pattern binding or multiple
+    guard let patternBindingList = self.parent?.as(PatternBindingListSyntax.self) else {
+      return []
+    }
+
+    if patternBindingList.children(viewMode: .sourceAccurate).count > 1 {
+      // Special case for pattern bindings like this: `let x = 1, y = 2, z = 3`
+      // Here we want to be able to select only `y = 2`
+      let start = self.positionAfterSkippingLeadingTrivia
+      let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+      return [start..<end]
+    }
+
+    // By default we don't want to create ranges for pattern bindings to avoid selecting `x = 0` in `let x = 0`
+    return []
+  }
+}
+
+extension IfExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let elseKeyword = self.elseKeyword, let elseBody = self.elseBody {
+      // When inside the else block add a range for selecting `else {...}`
+      let range = elseKeyword.positionAfterSkippingLeadingTrivia..<elseBody.endPositionBeforeTrailingTrivia
+      if range.contains(position) {
+        ranges.append(range)
+      }
+    }
+
+    ranges.append(self.trimmedRange)
+    return ranges
+  }
+}
+
+extension ForStmtSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For statements get an extra range for selecting from the pattern until the sequence,
+    // i.e. selecting `i in 1...3` in `for i in 1...3 {}`
+    // As the for statement can have a lot of immediate children, more special cases can be added here in the future,
+    // for example for the keywords before the pattern
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if (self.pattern.position..<self.sequence.endPosition).contains(position) {
+      let range = self.pattern.positionAfterSkippingLeadingTrivia..<self.sequence.endPositionBeforeTrailingTrivia
+      ranges.append(range)
+    }
+
+    ranges.append(self.trimmedRange)
+
+    return ranges
+  }
+}
+
+extension DictionaryElementSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    let start = self.positionAfterSkippingLeadingTrivia
+    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+    return [start..<end]
+  }
+}
+
+extension MemberAccessExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    if self.parent?.is(FunctionCallExprSyntax.self) ?? false {
+      // If the member access is part of a function call, we don't return any range as this case is handled in the
+      // FunctionCallExprSyntax extension
+      return []
+    }
+
+    return [self.trimmedRange]
+  }
+}
+
+extension AvailabilityArgumentSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    if let trailingComma = self.trailingComma {
+      let start = self.positionAfterSkippingLeadingTrivia
+      let end = trailingComma.positionAfterSkippingLeadingTrivia
+      return [start..<end]
+    }
+
+    return [self.trimmedRange]
+  }
+}
+
+extension TokenSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    switch self.tokenKind {
+    case .identifier where self.parent?.keyPathInParent == \AttributeSyntax.attributeName:
+      // For attributes we don't want to create a range for just the attribute name but rather always include the `@`
+      return []
+
+    case .identifier where self.keyPathInParent == \MacroExpansionExprSyntax.macroName:
+      // For macro expansions we don't want to create a range for just the macro name but rather always include the `#`
+      return []
+
+    case .identifier where self.keyPathInParent == \GenericParameterSyntax.name:
+      // For generic parameters we want to handle the identifier in the `GenericParameter` node as we may have to
+      // include or exclude the trailing comma
+      return []
+
+    case .keyword(let keyword) where keyword == .as:
+      // The `as` keyword should always be handled by the `UnresolvedAsExpr` node as it also includes the `!` or `?`
+      return []
+
+    case .binaryOperator, .dollarIdentifier, .floatLiteral, .identifier, .integerLiteral, .keyword:
+      return [self.trimmedRange]
+
+    default:
+      return []
+    }
+  }
+}
+
+// Default implementation used by all the nodes declared below
+private extension SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    return [self.trimmedRange]
+  }
+}
+
+extension AccessorBlockSyntax: SelectionRangeProvider {}
+extension AccessorDeclSyntax: SelectionRangeProvider {}
+extension ArrayElementListSyntax: SelectionRangeProvider {}
+extension ArrayExprSyntax: SelectionRangeProvider {}
+extension AsExprSyntax: SelectionRangeProvider {}
+extension AssociatedTypeDeclSyntax: SelectionRangeProvider {}
+extension AttributedTypeSyntax: SelectionRangeProvider {}
+extension AttributeListSyntax: SelectionRangeProvider {}
+extension AttributeSyntax: SelectionRangeProvider {}
+extension AvailabilityArgumentListSyntax: SelectionRangeProvider {}
+extension AwaitExprSyntax: SelectionRangeProvider {}
+extension ClassDeclSyntax: SelectionRangeProvider {}
+extension ClosureShorthandParameterListSyntax: SelectionRangeProvider {}
+extension ClosureShorthandParameterSyntax: SelectionRangeProvider {}
+extension ClosureSignatureSyntax: SelectionRangeProvider {}
+extension CompositionTypeElementListSyntax: SelectionRangeProvider {}
+extension ConditionElementListSyntax: SelectionRangeProvider {}
+extension ConditionElementSyntax: SelectionRangeProvider {}
+extension ConformanceRequirementSyntax: SelectionRangeProvider {}
+extension DeclReferenceExprSyntax: SelectionRangeProvider {}
+extension DeinitializerDeclSyntax: SelectionRangeProvider {}
+extension DictionaryElementListSyntax: SelectionRangeProvider {}
+extension DictionaryExprSyntax: SelectionRangeProvider {}
+extension DoStmtSyntax: SelectionRangeProvider {}
+extension EnumCaseDeclSyntax: SelectionRangeProvider {}
+extension EnumCaseElementListSyntax: SelectionRangeProvider {}
+extension ExpressionSegmentSyntax: SelectionRangeProvider {}
+extension FunctionEffectSpecifiersSyntax: SelectionRangeProvider {}
+extension FunctionParameterListSyntax: SelectionRangeProvider {}
+extension GenericParameterClauseSyntax: SelectionRangeProvider {}
+extension GenericParameterListSyntax: SelectionRangeProvider {}
+extension GenericRequirementSyntax: SelectionRangeProvider {}
+extension GenericWhereClauseSyntax: SelectionRangeProvider {}
+extension GuardStmtSyntax: SelectionRangeProvider {}
+extension ImplicitlyUnwrappedOptionalTypeSyntax: SelectionRangeProvider {}
+extension InheritanceClauseSyntax: SelectionRangeProvider {}
+extension InheritedTypeListSyntax: SelectionRangeProvider {}
+extension InitializerDeclSyntax: SelectionRangeProvider {}
+extension KeyPathExprSyntax: SelectionRangeProvider {}
+extension LabeledExprListSyntax: SelectionRangeProvider {}
+extension MacroExpansionExprSyntax: SelectionRangeProvider {}
+extension OperatorDeclSyntax: SelectionRangeProvider {}
+extension PlatformVersionSyntax: SelectionRangeProvider {}
+extension PrefixOperatorExprSyntax: SelectionRangeProvider {}
+extension RepeatStmtSyntax: SelectionRangeProvider {}
+extension ReturnClauseSyntax: SelectionRangeProvider {}
+extension ReturnStmtSyntax: SelectionRangeProvider {}
+extension StringLiteralExprSyntax: SelectionRangeProvider {}
+extension StringLiteralSegmentListSyntax: SelectionRangeProvider {}
+extension SubscriptDeclSyntax: SelectionRangeProvider {}
+extension SwitchCaseItemListSyntax: SelectionRangeProvider {}
+extension SwitchCaseItemSyntax: SelectionRangeProvider {}
+extension SwitchCaseSyntax: SelectionRangeProvider {}
+extension SwitchExprSyntax: SelectionRangeProvider {}
+extension ThrowStmtSyntax: SelectionRangeProvider {}
+extension TryExprSyntax: SelectionRangeProvider {}
+extension TupleExprSyntax: SelectionRangeProvider {}
+extension TuplePatternElementListSyntax: SelectionRangeProvider {}
+extension TuplePatternSyntax: SelectionRangeProvider {}
+extension TypeAnnotationSyntax: SelectionRangeProvider {}
+extension UnresolvedAsExprSyntax: SelectionRangeProvider {}
+extension VariableDeclSyntax: SelectionRangeProvider {}
+extension VersionTupleSyntax: SelectionRangeProvider {}
+extension WhileStmtSyntax: SelectionRangeProvider {}
--- a/Tests/SourceKitLSPTests/SelectionRangeTests.swift
+++ b/Tests/SourceKitLSPTests/SelectionRangeTests.swift
@@ -1,0 +1,2162 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SKTestSupport
+import SKUtilities
+import XCTest
+
+class SelectionRangeTests: XCTestCase {
+
+  // MARK: - Strings And Expressions
+
+  func testStringLiteralWithCursorInWord() async throws {
+    try await assertSelectionRanges(
+      markedSource: """
+        let a = "Hel1️⃣lo, World!"
+        """,
+      expectedSelections: [
+        "Hello",
+        "Hello, World!",
+        #""Hello, World!""#,
+      ]
+    )
+  }
+
+  func testStringLiteralWithCursorInWord2() async throws {
+    try await assertSelectionRanges(
+      markedSource: """
+        let a = "Hello, Wor1️⃣ld!"
+        """,
+      expectedSelections: [
+        "World",
+        "Hello, World!",
+        #""Hello, World!""#,
+      ]
+    )
+  }
+
+  func testStringLiteralWithCursorInWord3() async throws {
+    try await assertSelectionRanges(
+      markedSource: """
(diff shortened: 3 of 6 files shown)


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1177,6 +1179,7 @@ extension SourceKitLSPServer {
       typeHierarchyProvider: .bool(true),
       semanticTokensProvider: semanticTokensOptions,
       inlayHintProvider: inlayHintOptions,
+      selectionRangeProvider: .bool(true),
```
ahoppen, line 1185: Could you also advertise this capability in SwiftLanguageService.swift?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+    let sourceFile = await syntaxTreeManager.syntaxTree(for: snapshot)
+
+    try Task.checkCancellation()
+
+    var results: [SelectionRange] = []
+    for position in req.positions {
+      let absolutePosition = snapshot.absolutePosition(of: position)
+
+      guard let token = findIntuitiveToken(in: sourceFile, at: absolutePosition), let initialNode = token.parent else {
+        continue
```
ahoppen: I don’t think this is correct. If we don’t find an intuitive token for a position, the result array will have fewer elements than the request and thus the client has no way of associating the selection ranges with the requested positions.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+
+extension SwiftLanguageService {
+  package func selectionRange(_ req: SelectionRangeRequest) async throws -> [SelectionRange] {
+    let snapshot = try self.documentManager.latestSnapshot(req.textDocument.uri)
+    let sourceFile = await syntaxTreeManager.syntaxTree(for: snapshot)
+
+    try Task.checkCancellation()
+
+    var results: [SelectionRange] = []
+    for position in req.positions {
```
ahoppen: This could be a `.map` instead.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+      results.append(
+        computeSelectionRangeFor(position: absolutePosition, snapshot: snapshot, initialNode: initialNode, token: token)
+      )
+    }
+
+    return results
+  }
+}
+
+private func findIntuitiveToken(in sourceFile: SourceFileSyntax, at position: AbsolutePosition) -> TokenSyntax? {
```
ahoppen: Is this intended to do the same thing as `tokenForRefactoring`? I feel like it might be, but I could also be missing something.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+private func computeSelectionRangeFor(
+  position: AbsolutePosition,
+  snapshot: DocumentSnapshot,
+  initialNode: Syntax,
+  token: TokenSyntax
+) -> SelectionRange {
+  var ranges: [Range<AbsolutePosition>] = []
+
+  let parents = sequence(first: initialNode, next: { $0.parent })
+  let childParentPairs = zip(parents, [Syntax(token)] + parents)
```
ahoppen: I believe we have a pattern like the following in a few places and I think it would be good to follow that for consistency.

```swift
var current: Syntax = …
while let parent = current.parent {
  // code here
  current = parent
}
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+  }
+
+  var selectionRange: SelectionRange? = nil
+  for range in ranges.reversed() {
+    let start = snapshot.position(of: range.lowerBound)
+    let end = snapshot.position(of: range.upperBound)
+    selectionRange = SelectionRange(range: start..<end, parent: selectionRange)
+  }
+
+  return selectionRange!
```
ahoppen: I would prefer to throw an error / log a fault instead of crashing here if for some reason `selectionRange` is `nil`.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+
+private func calculateRangesFor(
+  node: Syntax,
+  previousNode: Syntax,
+  snapshot: DocumentSnapshot,
+  position: AbsolutePosition
+) -> [Range<AbsolutePosition>] {
+  switch node.as(SyntaxEnum.self) {
+
+  case .classDecl(let classDeclaration):
```
ahoppen: You should be able to handle all the nominal type declarations as `if let declGroup = node.as(DeclGroupSyntax.self)`

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+    let end = snapshot.position(of: range.upperBound)
+    selectionRange = SelectionRange(range: start..<end, parent: selectionRange)
+  }
+
+  return selectionRange!
+}
+
+private func calculateRangesFor(
+  node: Syntax,
+  previousNode: Syntax,
```
ahoppen: Isn’t `previousNode` always `node.parent`? If so, why do we need to pass it at all?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+      memberAccess.declName.positionAfterSkippingLeadingTrivia..<functionCall.endPositionBeforeTrailingTrivia,
+      functionCall.trimmedRange,
+    ]
+  }
+
+  // the default case: just create a range for the function call node
+  return [functionCall.trimmedRange]
+}
+
+private func calculateRangesInside(
```
ahoppen: For all of these functions, could you add short comments that explain how the range is being computed, maybe with an example? I think that would really help readers understand what’s happening.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2385 @@
...
+              return _temperature
+            }
+          """,
+        ]
+      )
+    }
+
+    @Test
+    func testPropertyWithWillSetDidSet() async throws {
+      try await testSelectionRange(
```
ahoppen: I’m personally doubtful that returning 7 different selection ranges will be useful, but I also haven’t used this feature in any other language. Do you have examples of what other LSP servers return for the selection ranges?

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2385 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SKTestSupport
+import SKUtilities
+import Testing
+
+@Suite(.serialized)
```
ahoppen: Almost all of our infrastructure is built around XCTest and I would prefer to stick this test case to XCTest as well. I’m just not sure if something in `SourceKitTestClient` or elsewhere in `SKTestSupport` may call `XCTFail`, which wouldn’t get picked up by swift-testing.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,659 @@
...
+  return selectionRange!
+}
+
+private func calculateRangesFor(
+  node: Syntax,
+  previousNode: Syntax,
+  snapshot: DocumentSnapshot,
+  position: AbsolutePosition
+) -> [Range<AbsolutePosition>] {
+  switch node.as(SyntaxEnum.self) {
```
ahoppen: General design decision: Wouldn’t it be cleaner to define a new protocol `SelectionRangeProvider` (or something like that) with a `selectionRanges: [Range<AbsolutePosition>]` property and then just conform all types that provide selection ranges to it? That way you also can’t get into a situation where you define one of the `calculateRanges` functions but forget to add it here.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: You're right, I will change it to return a selection range with just the current position

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I looked into this, and it seems like both functions do something similar, however when I tried using `tokenForRefactoring` instead a few of the tests started to fail. This is caused by `tokenForRefactoring` also applying the logic in cases like `matrix[|2, 3]` where it moves the cursor to before the `[` which leads to the entire `[2, 3]` being selected immediately.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I have changed this to return a `SelectionRange` with just the position in it, similar to what I have done for the continue case above. I can also add a log message for both cases if desired.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: `previousNode` is always one of the child nodes of `node`. It is useful to know which child we came from to be able to create a range from multiple child nodes in some cases. For example for `struct Foo<T|> {}` we can create a range for `Foo<T>`. However, when the cursor is in the `struct` keyword or in the body of the declaration we don't want to create this range.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: I mostly tried to follow what IntelliJ does for Kotlin as that is what I'm used to for the selection behavior. In this case IntelliJ provides even more selection ranges. Some of the difference can be explained by `willSet` and `didSet` being normal functions in the Kotlin version and not language built-ins.

https://github.com/user-attachments/assets/816d0713-9e97-4662-a5a0-82d945101a45

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: I just saw that some tests were using `Testing` so I assumed it was ok to use. I will change the tests to XCTest.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: That seems like a great idea. I'm used to writing Kotlin code where that is not possible, so I didn't think of that approach. But to me, it seems like the better approach, so I will change the code to use this design. During development, I actually ran into the case a few times where I implemented the `calculateRangesInside` function but forgot to add in the switch 😄

Steffeeen:
> Overall, I feel like we are maybe returning too many selection ranges. If you have examples of what other LSP servers return for the selection ranges, that would be a great reference point.

As mentioned in the other comment I mostly tried to follow what IntelliJ does for Kotlin code as that is what I'm most familiar with, and it generally selected what I intended. But then again maybe I'm just biased from using IntelliJ for a long time. Furthermore, I would argue that invoking the expand selection action one or two additional times is easier than having it skip over something and then having to manually select the code.

ahoppen:
> As mentioned in the other comment I mostly tried to follow what IntelliJ does for Kotlin code as that is what I'm most familiar with, and it generally selected what I intended

That’s very good feedback. Thanks for sharing the screen recording. I just wasn’t sure how other language servers handled the `textDocument/selectionRange` request at all. I just checked the TypeScript LSP server as well and it behaves similarly, so let’s stick with what you’re proposing in this PR.

Sources/SwiftLanguageService/SelectionRange.swift
ahoppen: Hmm, OK. Let me look at this closer in the next review.

Sources/SwiftLanguageService/SelectionRange.swift
ahoppen: I’m not entirely sure I follow. Wouldn’t something like the following work?
- First you’re at the generic parameter -> not sure if this returns a selection range, probably does
- Then you walk up the parent chain and reach a `GenericParameterClauseSyntax` -> this produces a selection range.
- Next parent is `StructDeclSyntax` -> produces a selection range again

And when started from the `struct` keyword we just never reach the `GenericParameterClauseSyntax`, so no selection range is returned.

Honestly, now that I think about it, we might just need a list of node types for which we want to return selection ranges and if a syntax node is in that list, we return its `trimmedRange` as a selection range. Might be missing something though. Would like to see this PR implemented with https://github.com/swiftlang/sourcekit-lsp/pull/2467#discussion_r2743156970, which should make it easier for me to review.

Steffeeen: What you describes works in principle, but the next selection after `<T>` would already be the entire struct declaration `struct Foo<T> {}`. This is because the `GenericParameterClauseSyntax` and the `identifier` token are siblings in the `StructDeclSyntax`.

I also checked what IntelliJ does in this case and interestingly for Java it creates the selection ranges like I do, i.e. having a separate one for just the name and generic parameter, while for Kotlin it directly jumps to selecting the entire class after selecting the generic parameter.

The list of node types approach is what I started with and is still the foundation of the current implementation. All the cases in the switch, except the last two, are special cases. The last two implement this node list approach: we have the case for all the nodes which should not get a range and the default case which creates a range if the node is not special-cased above or listed as not creating a special case.

I think we cannot avoid special-casing a bunch of nodes due to the way the AST is structured. Another example are variadic parameters: the ellipsis is part of the `FunctionParameterSyntax` while, for me at least, semantically it is part of the type. I would thus expect to be able to select just `Int...` in `func test(a: In|t...) {}` before selecting the entire parameter.

As mentioned below, I will change the implementation to the approach you suggested, it seems sensible.

Steffeeen:
I changed the implementation to use the protocol and extension approach. This is just a new commit for now as I did not want to through all the commits and change each one. I have also tried using the `DeclGroupSyntax` protocol like you suggested but could not get it to work with the extension approach. This may just be an issue on my site as I haven't really grasped the specifics of how extensions and protocols interact in this case.

I still intend to change the tests to use XCTest and add the comment for each extension explaining why it is needed.

 As I need to squash the commits anyhow before this PR gets merged, I thought about squashing now. Should I just put everything in one big commit or split this PR up into smaller commits? What do you think @ahoppen?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      previousNode: previousNode,
+      nameOrType: Syntax(self.name),
+      genericParameters: self.genericParameterClause
+    )
+  }
+}
+
+extension ProtocolDeclSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+
```
ahoppen: I think empty lines at the start of a function are superfluous 😉 Could you remove them?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+}
+
+private func computeSelectionRangeFor(
+  position: AbsolutePosition,
+  snapshot: DocumentSnapshot,
+  token: TokenSyntax
+) -> SelectionRange? {
+  var ranges: [Range<AbsolutePosition>] = []
+
+  // The first node we call calculateRangesFor() for is never a token. This forces us to sometimes explicitly create
```
ahoppen: I might be missing something but couldn’t we do something like the following so we call `calculateRangesFor` for the token itself?

```swift
while true {
   …
  guard let parent = current.parent else { 
    break
  }
  current = parent
}
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+}
+
+private func calculateRangesFor(
+  node: Syntax,
+  previousNode: Syntax,
+  snapshot: DocumentSnapshot,
+  position: AbsolutePosition
+) -> [Range<AbsolutePosition>] {
+  switch node.as(SyntaxEnum.self) {
+
```
ahoppen: Superfluous empty line

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+    // We cannot provide the position in the protocol as it may not always be correct due to the logic in findIntuitiveToken().
+    // For the string segment this does not matter as when we encounter a string segment node we can be sure that
+    // findIntuitiveToken() always returned the original token.
+    return calculateSelectionRangesForStringSegment(stringSegment: stringSegmentSyntax, position: position)
+
+  case .patternBindingList, .initializerClause, .matchingPatternCondition, .sequenceExpr,
+    .accessorDeclList, .functionParameterClause, .functionSignature, .switchCaseLabel, .switchCaseList,
+    .inheritedType, .memberBlockItemList, .memberBlock, .enumCaseParameterClause,
+    .optionalChainingExpr, .tuplePatternElement, .arrayElement, .keyPathComponent, .keyPathComponentList:
+    return []
```
ahoppen: Do we need this clause at all? I think it would be easier if we just relied on the `SelectionRangeProvider` cast below so we automatically return selection ranges for new types that add a `SelectionRangeProvider` conformance.

At which point the entire cast to `SyntaxEnum` could become just a check for `StringSegmentSyntax`.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+) -> [Range<AbsolutePosition>] {
+  // For string segments we first want to select just the word under the cursor.
+  // To determine words we use a simple heuristic: expand the selection until we hit any whitespace character.
+
+  let offsetInString = position.utf8Offset - stringSegment.positionAfterSkippingLeadingTrivia.utf8Offset
+
+  let text = stringSegment.content.text
+  let index = text.index(text.startIndex, offsetBy: offsetInString)
+
+  if text[index].isWhitespace {
```
ahoppen: Should we be checking for `.isLetter` instead? That would also cover eg. `,` as a word boundary.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+  }
+
+  var start = index
+  while text.startIndex < start {
+    let prev = text.index(before: start)
+    if text[prev].isWhitespace {
+      break
+    }
+    start = prev
+  }
```
ahoppen: Couldn’t this be something like `text[..<start].lastIndex(where: { !$0.isWhitespace })`. Same below.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      break
+    }
+    end = text.index(after: end)
+  }
+
+  let startOffsetInString = text.distance(from: text.startIndex, to: start)
+  let endOffsetInString = text.distance(from: text.startIndex, to: end)
+
+  let startPosition = stringSegment.positionAfterSkippingLeadingTrivia.advanced(by: startOffsetInString)
+  let endPosition = stringSegment.positionAfterSkippingLeadingTrivia.advanced(by: endOffsetInString)
```
ahoppen: This is not correct because the distance in the string indices will be in Unicode grapheme clusters while `AbsolutPosition` is measured in terms of UTF-8 bytes. I think you should be able to fix this by using `text.utf8.distance` instead.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+  // for just the name (type) and generic parameter clause.
+  //
+  // Example: given `struct Test<|T> {}` we want to have a selection range for `Test<T>`
+  //
+  // As the name (type) and generic parameter clause are siblings in the declaration nodes
+  // we have to special-case them.
+
+  var ranges: [Range<AbsolutePosition>] = []
+
+  if previousNode.id == nameOrType.id {
```
ahoppen: I still don’t think that passing in the previous node into this function is the correct design here. Instead, I’d prefer to report the range of `nameOrType` when we are traversing it. If we need to check that a `TokenSyntax` is a function name, you can check `token.keyPathInParent == \FunctionDeclSyntax.name` (or something along those lines).

Or am I missing something here why that wouldn’t work?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+    .inheritedType, .memberBlockItemList, .memberBlock, .enumCaseParameterClause,
+    .optionalChainingExpr, .tuplePatternElement, .arrayElement, .keyPathComponent, .keyPathComponentList:
+    return []
+
+  default:
+    if let provider = node.asProtocol((any SyntaxProtocol).self) as? (any SelectionRangeProvider) {
+      return provider.calculateSelectionRanges(previousNode: previousNode)
+    }
+
+    return [node.trimmedRange]
```
ahoppen: Instead of returning the node’s trimmed range as a selection range for any nodes that may be added in the future, I’d prefer to have an allow list of nodes we want to report, ideally by just conforming them to `SelectionRangeProvider`.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      nameOrType: Syntax(self.name),
+      genericParameters: self.genericParameterClause
+    )
+  }
+}
+
+private func calculateRangesForTypeOrFunctionDeclaration(
+  declaration: Syntax,
+  previousNode: Syntax,
+  nameOrType: Syntax,
```
ahoppen: You can save yourself a little bit of casting on the call side if you make these `some SyntaxProtocol`

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+
+  ranges.append(declaration.trimmedRange)
+
+  return ranges
+}
+
+extension FunctionCallExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    if let memberAccess = self.calledExpression.as(MemberAccessExprSyntax.self),
+      self.parent?.as(ExpressionPatternSyntax.self) == nil,
```
ahoppen: Since we don’t actually care about the value, we should use `is` instead.

```suggestion
      !(self.parent?.is(ExpressionPatternSyntax.self) ?? false),
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      previousNode.id == self.arguments.id
+        || previousNode.id == self.trailingClosure?.id
```
ahoppen: I’ve now been able to identify why I feel uneasy about the `previousNode` checks: We wouldn’t hit this case if the selection range was started from the right parenthesis or if some new node got added in between the arguments and trailing closure in the future. I think what would be a cleaner design is if the requested position was passed in here. We could then check something like `(arguments.position..<self.endPosition).contains(requestedPosition)`.

Also realized that you should probably also handle additional trailing closures here.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+    {
+      // Special case for adding an extra range including the function name and parameters/trailing closures
+      // this is needed for chained method calls
+      // Example:
+      // numbers
+      //  .filter { $0 > 0 }
+      //  .map { $0 * 2 }
+      //  .reduce(0,| +)
+      //
+      // when starting a selection from | we want to have a selection for `reduce(0, +)` in addition to selecting
```
ahoppen: Should this also be the case when starting from `red|uce`?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+
+extension LabeledExprSyntax: SelectionRangeProvider {
+  // for labeled expressions we want to be able to select just the label and expression without the comma
+
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let label = self.label,
+      previousNode.id == label.id
+    {
```
ahoppen:
```suggestion
    if let label = self.label, previousNode.id == label.id {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
+    return [self.positionAfterSkippingLeadingTrivia..<end]
+  }
+}
+
+extension FunctionParameterSyntax: SelectionRangeProvider {
+  // Function parameters have two special cases:
+  // - If the cursor is in the type and the type includes an ellipsis we want to have a range for the type and ellipsis
+  // - If the parameter has two names we want to have a range for selecting both names
+
```
ahoppen: I would place these comments inside the function body. Putting them in front of it makes it seem like they are meant to be doc comments while they actually describe the implementation.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    let start = self.positionAfterSkippingLeadingTrivia
+    let end =
+      if let comma = self.trailingComma {
+        comma.position
+      } else {
+        self.endPositionBeforeTrailingTrivia
+      }
```
ahoppen:
```suggestion
    let end = self.trailingComma?.position ?? self.endPositionBeforeTrailingTrivia
```

Same further below again

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+
+    return ranges
+  }
+
+  private func findCorrespondingOperandIn(
+    foldedTree: Syntax,
+    operandStart: AbsolutePosition,
+    operandEnd: AbsolutePosition
+  ) -> Syntax {
+    var current = foldedTree
```
ahoppen: Would it be simpler to use `token(at:)` to find the token at the give position and then check if its parent is an `InfixOperatorExprSyntax`?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+
+    let operandNode = findCorrespondingOperandIn(
+      foldedTree: Syntax(foldedTree),
+      operandStart: startInTree,
+      operandEnd: endInTree
+    )
+
+    var ranges: [Range<AbsolutePosition>] = []
+
+    for node in sequence(first: operandNode, next: { $0.parent }) {
```
ahoppen: Nice use of `sequence`. Never thought of it that way 👍🏽 

```suggestion
    for node in sequence(first: operandNode, next: \.parent) {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      if node.is(InfixOperatorExprSyntax.self) {
+        let startPosition =
+          sequenceExpression.position + SourceLength(utf8Length: node.positionAfterSkippingLeadingTrivia.utf8Offset)
```
ahoppen: Would it make sense to store `SourceLength(utf8Length: sequenceExpression.position.utf8Offset)` in a variable somewhere above and then flip this around to

```swift
node.positionAfterSkippingLeadingTrivia + foldedTreeOffset
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+      } else {
+        self.positionAfterSkippingLeadingTrivia
+      }
+    return [start..<end]
+  }
+}
+
+extension OperatorDeclSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    if previousNode.id == self.name.id {
```
ahoppen: I think this wouldn’t be necessary if we started tree traversal at the token node, right?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+
+extension MemberAccessExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    return if self.parent?.is(FunctionCallExprSyntax.self) == true {
+      // If the member access is part of a function call, we don't return any range as this case is handled in the
+      // FunctionCallExprSyntax extension
+      []
+    } else {
+      [self.trimmedRange]
+    }
```
ahoppen:
```suggestion
    if self.parent?.is(FunctionCallExprSyntax.self) ?? false {
      // If the member access is part of a function call, we don't return any range as this case is handled in the
      // FunctionCallExprSyntax extension
      return []
    }
    return [self.trimmedRange]
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+  }
+}
+
+extension IdentifierTypeSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>] {
+    if self.parent?.is(AttributeSyntax.self) == true {
```
ahoppen: I generally prefer using `??` for these `Optional<Bool>` comparisons because it clearly assigns a boolean value to the `nil` value.

```suggestion
    if self.parent?.is(AttributeSyntax.self) ?? false {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,695 @@
...
+  let endPosition = stringSegment.positionAfterSkippingLeadingTrivia.advanced(by: endOffsetInString)
+
+  return [startPosition..<endPosition]
+}
+
+private protocol SelectionRangeProvider: SyntaxProtocol {
+  func calculateSelectionRanges(previousNode: Syntax) -> [Range<AbsolutePosition>]
+}
+
+extension ClassDeclSyntax: SelectionRangeProvider {
```
ahoppen: I think you currently don’t handle `ActorDeclSyntax`. I would really prefer if all of these nodes were handled using the `DeclGroupSyntax` protocol. Since we can’t retroactively conform `DeclGroupSyntax` to `SelectionRangeProvider`, we might just need to add a special case in `calculateRangesFor` that checks for `DeclGroupSyntax` in addition to `SelectionRangeProvider`.

ahoppen (commented):
Thanks for making the adjustments, @Steffeeen. Left a few more comments inline, the biggest one being that I still feel uneasy about the `previousNode` parameter to these functions. One comment explains why and I’d prefer to just pass in the position at which the selection ranges were requested. Once that’s addressed, I think it should be down to just one final round of nitpicks or so.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I agree, that the position would be better, an earlier version actually used the position instead of the previous node. I changed the code to passing the previous node as the logic in `findIntuitiveToken` may determine another token to be the "correct" token to use, for example in this case: `foo(1, 2|)`. Here `findIntuitiveToken` determines that we want to start the selection from the `5` token instead of the `)` token the cursor is actually at. This however causes checks like `trimmedRange.contains(position)` to fail for the `LabeledExprList` for example. I guess we could also adjust the cursor position to be the end position of the token minus one. I remember also considering this when I changed to passing the previous node, however I don't remember if this caused other issues. I will try this approach and see if I run into any issues.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: Yes, `token(at:)` is a good idea. I was able to get rid of the `findCorrespondingOperandIn` function entirely.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I fixed this and also added a test that explicitly uses utf-8 characters in a string literal. This uncovered a bug in the supporting test code which I have also fixed.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I changed the code to use `.isLetter`. This also better matches what IntelliJ does.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: Yes, good point. I adjusted this when changing the code to pass the position instead of the previous node.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I changed the code to use the position instead of the previous node. That seems to work fine.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: With the old design of using the switch statement it was a conscious decision not to call `calculateRangesFor` for the token as that would have needed further special casing. For some tokens we don't want to create a selection range while for others do. With the new design (using the `SelectionRangeProvider` protocol and explicit allow list) this is no longer an issue as I just didn't conform `TokenSyntax` to `SelectionRangeProvider`. I thus now use the loop like you described to allow adding `SelectionRangeProvider` conformance to `TokenSyntax` in the future.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I switched to the explicit allow list approach and replaced the switch with three ifs: one for `StringSegmentSyntax`, one for `DeclGroupSyntax` and one for `SelectionRangeProvider`.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: With the old design of creating a selection range by default, this would be unnecessary. However, there are other cases where we don't want to create a selection range for tokens. For example for attributes we want to immediately select the entire attribute including the `@`. We thus don't want to have a range for the token.

In the new design, `TokenSyntax` currently does not conform to `SelectionRangeProvider` as I kept the previous special cases (like this one) for explicitly creating the range for tokens that need it. But I guess this logic could be moved into a `TokenSyntax` extension which looks at its parents to decide if the token should create a range.

Steffeeen: I moved the logic into a `TokenSyntax` extension.

Steffeeen:
I have addressed all of your comments @ahoppen.

Currently, the logic for handling trailing commas is still duplicated in some places. I have tried to unify this handling but sometimes further logic is needed (for example for `FunctionParameterSyntax`) which prevents nice uniform handling of trailing commas.

The extension list at the bottom of the file is sorted alphabetically. The other extensions are currently not sorted and in a more or less random order. Should I also sort them alphabetically?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+    // We have to use custom logic for string segments as they need the position.
+    // We cannot provide the position in the protocol as it may not always be correct due to the logic in findIntuitiveToken().
+    // For the string segment this does not matter as when we encounter a string segment node we can be sure that
+    // findIntuitiveToken() always returned the original token.
+    return calculateSelectionRangesForStringSegment(stringSegment: stringSegment, position: position)
+  }
+
+  if node.isProtocol((any DeclGroupSyntax).self)
+    || node.is(TypeAliasDeclSyntax.self)
+    || node.is(FunctionDeclSyntax.self)
```
ahoppen, line 134: Should we also handle the following nodes here as they also have a name (just a keyword but kind of qualifies as the same) and a generic where clause?
- `InitializerDeclSyntax`
- `MacroDeclSyntax`
- `SubscriptDeclSyntax`

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+      return [start..<end, self.trimmedRange]
+    }
+
+    return [self.trimmedRange]
+  }
+}
+
+extension LabeledExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // for labeled expressions we want to be able to select just the label and expression without the comma
```
ahoppen: Just for consistency, let’s start all comments with an uppercase letter. 

```suggestion
    // For labeled expressions we want to be able to select just the label and expression without the comma
```

Same in at least one more position.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // For closure signatures we add a selection for the signature + body content
+    var ranges: [Range<AbsolutePosition>] = []
+    ranges.append(self.trimmedRange)
+
+    if let closureExpression = self.parent?.as(ClosureExprSyntax.self) {
+      let start = self.positionAfterSkippingLeadingTrivia
+      let end = closureExpression.statements.endPositionBeforeTrailingTrivia
+      ranges.append(start..<end)
+    }
```
ahoppen: In all the previous cases `calculateSelectionRanges` only ever report additional restricted instead of extended ranges, which I also find more intuitive. As such, would it make sense to implement this on `ClosureExprSyntax`.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let firstName = self.firstName {
+      if let secondName = self.secondName {
+        // The parameter has two names, first select the one with the cursor in it and then both
+        let range = firstName.positionAfterSkippingLeadingTrivia..<secondName.endPositionBeforeTrailingTrivia
+        if firstName.range.contains(position) {
+          ranges.append(firstName.trimmedRange)
```
ahoppen: Do we need to report these? Shouldn’t they already be reported by the conformance of `TokenSyntax` to `SelectionRangeProvider`? Same for `FunctionParameterSyntax`.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+extension ExprListSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    // Expression lists are more complex to deal with as we first have to convert the expression list into the
+    // corresponding tree. We can then find the node the cursor is on in the tree and walk the tree up to its root
+    guard let sequenceExpression = self.parent?.as(SequenceExprSyntax.self) else {
+      return [self.trimmedRange]
+    }
+
+    let table = OperatorTable.standardOperators
+    guard let foldedTree = try? table.foldSingle(sequenceExpression) else { return [] }
```
ahoppen: Could you put this in an `orLog` statement so we emit a log message if folding fails, which may help debug issues.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+
+    let foldedTreeOffset = SourceLength(utf8Length: sequenceExpression.position.utf8Offset)
+    let offsetInTree = position - foldedTreeOffset
+
+    guard let token = foldedTree.token(at: offsetInTree)?.parent else {
```
ahoppen: If we have `?.parent`, the resulting node isn’t a token anymore though, right? It’s already of type `Syntax`. Should this bind `operandNode` directly?


```suggestion
    guard var operandNode = foldedTree.token(at: offsetInTree)?.parent else {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+}
+
+extension CodeBlockSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    if let ifExpression = self.parent?.as(IfExprSyntax.self),
+      let elseKeyword = ifExpression.elseKeyword,
+      ifExpression.elseBody?.id == self.id
+    {
+      // special case for if expression: when inside the else block add a range for selecting `else {...}`
+      return [elseKeyword.positionAfterSkippingLeadingTrivia..<self.endPositionBeforeTrailingTrivia]
```
ahoppen: Similar to my comment on `ClosureSignatureSyntax` of only reporting more restricted ranges, should this be implemented on `IfExprSyntax` so we can report a more restricted range for `else { … }` there?

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+      return [start..<end]
+    }
+
+    return [self.trimmedRange]
+  }
+}
+
+extension TokenSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    let parentTypes: [any SyntaxProtocol.Type] = [
```
ahoppen: I might totally be missing something, but shouldn’t we report a selection range for every identifier, keyword, operator and number literal? Feel free to search through `TokenKind` as well in case I missed something. All of them seem worth highlighting to me and then I don’t think we’d need any of these rules anymore.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,613 @@
...
+extension GenericRequirementSyntax: SelectionRangeProvider {}
+extension GenericWhereClauseSyntax: SelectionRangeProvider {}
+extension GuardStmtSyntax: SelectionRangeProvider {}
+extension IdentifierPatternSyntax: SelectionRangeProvider {}
+extension IfExprSyntax: SelectionRangeProvider {}
+extension ImplicitlyUnwrappedOptionalTypeSyntax: SelectionRangeProvider {}
+extension InheritanceClauseSyntax: SelectionRangeProvider {}
+extension InheritedTypeListSyntax: SelectionRangeProvider {}
+extension InitializerDeclSyntax: SelectionRangeProvider {}
+extension IntegerLiteralExprSyntax: SelectionRangeProvider {}
```
ahoppen: This wouldn’t be needed if we report integer tokens as selection ranges as I suggested in the `TokenSyntax` extension.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+      markedSource: markedSource,
+      expectedSelections: [expectedSelections],
+      file: file,
+      line: line
+    )
+  }
+
+  func testSelectionRange(
+    markedSource: String,
+    expectedSelections: [[String]],
```
ahoppen: Should this be `[String: [String]]` so we can be explicit about which selection marker corresponds to which expected ranges?

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+  }
+
+  func testActorDeclarationWithGenerics() async throws {
+    try await testSelectionRange(
+      markedSource: "actor Test<T1️⃣> {}",
+      expectedSelections: ["T", "<T>", "Test<T>", "actor Test<T> {}"]
+    )
+  }
+
+  func testSelectionRange(
```
ahoppen: To follow the naming patterns in this repo, could these be called `assertSelectionRanges` and be top-level `private` functions?

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+      expectedSelections.count,
+      documentPositions.allMarkers.count,
+      "The number of markers and expected selections differ: \(documentPositions.allMarkers.count) markers vs \(expectedSelections.count) selections",
+      file: file,
+      line: line
+    )
+    let flatMappedSelections = expectedSelections.flatMap { $0 }
+    XCTAssert(
+      flatMappedSelections.allSatisfy { text.contains($0) },
+      "The following expected selections are not contained in the source:\n \(flatMappedSelections.filter { !text.contains($0) }.joined(separator: "\n"))",
```
ahoppen: Think this reads a little nicer as a multi-line literal

```suggestion
      """
      The following expected selections are not contained in the source:
      \(flatMappedSelections.filter { !text.contains($0) }.joined(separator: "\n"))
      """,
```

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+
+    // check the actual returned ranges
+    let testClient = try await TestSourceKitLSPClient()
+    let uri = DocumentURI(for: .swift)
+    testClient.openDocument(text, uri: uri)
+
+    for (index, marker) in documentPositions.allMarkers.enumerated() {
+      let position = documentPositions[marker]
+      let request = SelectionRangeRequest(textDocument: TextDocumentIdentifier(uri), positions: [position])
+      let response: SelectionRangeRequest.Response = try await testClient.send(request)
```
ahoppen: We usually just do the following in tests.

```suggestion
      let response = try await testClient.send(SelectionRangeRequest(textDocument: TextDocumentIdentifier(uri), positions: [position]))
```

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+
+      let lineTable = LineTable(text)
+
+      let range = response[0]
```
ahoppen: We shouldn’t crash the tests if the selection ranges are empty.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+      let expected = expectedSelections[index]
+
+      var rangeIndex = 0
+      var currentRange: SelectionRange? = range
+      while rangeIndex < expected.count {
```
ahoppen: This won’t fail the test if the selection range response contains more elements than expected, right? Looking at the tests, I think that’s expected but it would be worth documenting this behavior.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+          line: line
+        )
+
+        currentRange = currentRange?.parent
+        rangeIndex += 1
+      }
+    }
+  }
+
+  func selectionRangeMismatchMessage(rangeIndex: Int, expected: String, actual: String) -> String {
```
ahoppen: If we use position markers as the keys for `expectedSelections`, we can report the position marker here instead of `rangeIndex`.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        """
+    }
+  }
+
+  func getStringOfSelectionRange(lineTable: LineTable, selectionRange: SelectionRange?) -> String {
+    guard let selectionRange = selectionRange else {
+      return "<no selection range>"
+    }
+
+    let lowerBoundOffset = lineTable.utf8OffsetOf(
```
ahoppen: I think you can do `stringIndexOf` directly, no need to go through UTF-8 offsets.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+  }
+
+  func testVariableWithExplicitType() async throws {
+    try await testSelectionRange(
+      markedSource: """
+        let name: Str1️⃣ing = "Swift"
+        """,
+      expectedSelections: [
+        "String",
+        ": String",
```
ahoppen, line 196: Do you think that reporting this type annotation is useful? I wouldn’t have thought so but if you think we should report it, I’m happy to join your opinion.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+  func testComputedProperty() async throws {
+    try await testSelectionRange(
+      markedSource: """
+        var fullName: String {
+          return first1️⃣Name + " " + lastName
+        }
+        """,
+      expectedSelections: [
+        "firstName",
+        "firstName + \" \"",
```
ahoppen: Would these read easier as raw string literals with `#"`? Same for the other tests that contain string literals or backslashes in the expected selections.

```suggestion
        #"firstName + " ""#,
```

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        """
+        func compare<T>(a: T, b: T) -> Bool where T: Comparable {
+          return a < b
+        }
+        """,
+      ]
+    )
+  }
+
+  // MARK: - ClosuresAndFunctionTypes
```
ahoppen: You can use spaces in `MARK` comments 😉 

```suggestion
  // MARK: - Closures and function types
```

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+    try await testSelectionRange(
+      markedSource: """
+        numbers.map { nu1️⃣m in
+          return num * 2
+        }
+        """,
+      expectedSelections: [
+        "num",
+        "num in",
+        "num in\n  return num * 2",
```
ahoppen:
```suggestion
        """
        num in
          return num * 2
        """,
```

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        case .option2:
+          print("two")
+        default:
+          break
+        }
+        """,
+      expectedSelections: [
+        "option1",
+        ".option1",
+        "case .option1:\n  print(\"one\")",
```
ahoppen:
```suggestion
        """
        case .option1:
          print("one")
        """,
```

Similarly, I’d use multi-line string literals for all expected selections that contain newlines.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+  func testProtocolInheritance() async throws {
+    try await testSelectionRange(
+      markedSource: """
+        protocol TextRepresentable: CustomString1️⃣Convertible {
+          var text: String { get }
+        }
+        """,
+      expectedSelections: [
+        "CustomStringConvertible",
+        ": CustomStringConvertible",
```
ahoppen: Do you think it makes sense to report this range? Similar to the type annotation, I probably wouldn’t have reported it but happy to follow your judgment.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        """
+        public func apiMethod() {
+          print("Public API")
+        }
+        """,
+      ]
+    )
+  }
+
+  func testOpenClass() async throws {
```
ahoppen: I think `testPublicModifier` and `testOpenClass` test the exact same code paths as `testPrivateModifier` and are thus redundant.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        weak var dele1️⃣gate: MyDelegate?
+        """,
+      expectedSelections: [
+        "delegate",
+        "weak var delegate: MyDelegate?",
+      ]
+    )
+  }
+
+  func testUnownedReference() async throws {
```
ahoppen: Isn’t this testing the exact same thing as `testWeakReference`?

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+      expectedSelections: [
+        "name",
+        "\\Person.name",
+        "let keyPath = \\Person.name",
+      ]
+    )
+  }
+
+  func testKeyPathSubscript() async throws {
+    try await testSelectionRange(
```
ahoppen: Doesn’t this just test functionality that’s already tested by the key path test case and the subscript call test?

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2383 @@
...
+        "[keyPath: \\.name]",
+        "person[keyPath: \\.name]",
+        "let value = person[keyPath: \\.name]",
+      ]
+    )
+  }
+
+  // MARK: - DynamicMemberLookup
+
+  func testDynamicMemberLookup() async throws {
```
ahoppen: Isn’t this structurally equivalent to `testPropertyWrapperDeclaration`?

If you could do another pass to see if there are any more effectively duplicated test cases that I’ve missed, I’d appreciate that to make sure we don’t spend too much time testing the same thing over and over again.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: I changed the code to create a range for `binaryOperator`, `dollarIdentifier`, `floatLiteral`, `identifier`, `integerLiteral` and `keyword` tokens. I had to add four special cases where we don't want to create a range for one of these tokens. But overall this implementation seems cleaner to me as the range creation for tokens is now all in a single place.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: Seems like a good idea, I adjusted the code.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: I changed the code, but I'm curious if there is any specific reason for doing it like this? I originally split the line into two statements as I didn't like the long line.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: I think it's useful for cases where you want to remove the explicit type annotation. IntelliJ also does the same for Kotlin.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: A similar reasoning as for the type annotation applies: this range is useful for removing the protocol inheritance.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: Yes, it seems like `testPrivateModifier` also already covers `testUnownedReference` and `testWeakReference`. I have removed both.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
Steffeeen: I looked through all the tests and removed a bunch of them. I measured the test coverage before and after and it's still roughly 96%.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
ahoppen: For me personally it’s one less variable that needs to be named, especially if you send multiple requests in a test case. Maybe it’s just a habit. There’s no strong reason.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
ahoppen: Sounds like a good reason to me. Thanks for sharing your rationale.

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,618 @@
...
+  }
+}
+
+extension ClosureExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let signature = self.signature,
+      signature.range.contains(position)
+    {
```
ahoppen: Nitpick: Fits on one line

```suggestion
    if let signature = self.signature, signature.range.contains(position) {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,618 @@
...
+  }
+}
+
+extension IfExprSyntax: SelectionRangeProvider {
+  func calculateSelectionRanges(position: AbsolutePosition) -> [Range<AbsolutePosition>] {
+    var ranges: [Range<AbsolutePosition>] = []
+
+    if let elseKeyword = self.elseKeyword,
+      let elseBody = self.elseBody
+    {
```
ahoppen:
```suggestion
    if let elseKeyword = self.elseKeyword, let elseBody = self.elseBody {
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,618 @@
...
+      return []
+
+    case .identifier where self.parent?.is(MacroExpansionExprSyntax.self) ?? false:
+      // For macro expansions we don't want to create a range for just the macro name but rather always include the `#`
+      return []
+
+    case .identifier where self.parent?.is(GenericParameterSyntax.self) ?? false:
+      // For generic parameters we want to handle the identifier in the `GenericParameter` node as we may have to
+      // include or exclude the trailing comma
+      return []
```
ahoppen: What do you think of changing this to the following so we are more explicit about which child node of these we care about (in case one of these gains a new identifier child in the future)

```diff
    case .identifier where self.parent?.keyPathInParent == \AttributeSyntax.attributeName:
      // For attributes we don't want to create a range for just the attribute name but rather always include the `@`
      return []

    case .identifier
    where self.keyPathInParent == \MacroExpansionDeclSyntax.macroName
      || self.keyPathInParent == \MacroExpansionExprSyntax.macroName:
      // For macro expansions we don't want to create a range for just the macro name but rather always include the `#`
      return []

    case .identifier where self.keyPathInParent == \GenericParameterSyntax.name:
      // For generic parameters we want to handle the identifier in the `GenericParameter` node as we may have to
      // include or exclude the trailing comma
      return []
```

Sources/SwiftLanguageService/SelectionRange.swift
```diff
@@ -0,0 +1,618 @@
...
+    case .binaryOperator, .dollarIdentifier, .floatLiteral, .identifier, .integerLiteral, .keyword:
+      return [self.trimmedRange]
+
+    case _
+    where self.parent?.isProtocol((any DeclGroupSyntax).self) ?? false
+      || self.parent?.is(TypeAliasDeclSyntax.self) ?? false
+      || self.parent?.is(FunctionDeclSyntax.self) ?? false
+      || self.parent?.is(InitializerDeclSyntax.self) ?? false
+      || self.parent?.is(MacroDeclSyntax.self) ?? false
+      || self.parent?.is(SubscriptDeclSyntax.self) ?? false:
```
ahoppen: I don’t fully understand why this is needed. I feel like I’m missing something but if I don’t understand it now, it would be good to explain what’s going on here in a comment.

Tests/SourceKitLSPTests/SelectionRangeTests.swift
```diff
@@ -0,0 +1,2163 @@
...
+    line: selectionRange.range.lowerBound.line,
+    utf16Column: selectionRange.range.lowerBound.utf16index
+  )
+  let upperBoundIndex = lineTable.stringIndexOf(
+    line: selectionRange.range.upperBound.line,
+    utf16Column: selectionRange.range.upperBound.utf16index
+  )
+
+  let slice = lineTable.content[lowerBoundIndex..<upperBoundIndex]
+  return String(slice)
```
ahoppen:
```suggestion
  return String(lineTable.content[lowerBoundIndex..<upperBoundIndex])
```

ahoppen (commented):
Just a few more comments, otherwise this looks great to me. Thanks a lot for adding new support for this LSP request. 🚀

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: Seems like a good idea, especially with regard to future proofing this code.

Sources/SwiftLanguageService/SelectionRange.swift
Steffeeen: Yeah, this is no longer needed as this is now covered by the more general `.identifier` case.

Steffeeen:
If there are no further comments, I would go ahead and squash everything into a single commit for merging.

Steffeeen:
I squashed everything into a single commit. Let's ship it 🚀

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SwiftLanguageService/SelectionRange.swift
ahoppen, line 1: Can you add this file to [CMakeLists.txt](https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SwiftLanguageService/CMakeLists.txt) so it gets picked up by the Windows build?

Steffeeen: Done.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Steffeeen:
Thank you for merging this 🎉

ahoppen:
Thank you for implementing it!

PR #2471, opened 2026-01-30
Add a test for the topic section in the DocC preview

Steffeeen:
cc @matthewbastien

### DIFF
--- a/Tests/SourceKitLSPTests/DoccDocumentationTests.swift
+++ b/Tests/SourceKitLSPTests/DoccDocumentationTests.swift
@@ -885,6 +885,44 @@ final class DoccDocumentationTests: SourceKitLSPTestCase {
     )
   }
 
+  func testMarkdownExtensionTopics() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "MyLibrary/Foo.swift": """
+          public struct  1️⃣ Foo {
+            public func methodA() {}
+            public func methodB() {}
+          }
+        """,
+
+        "MyLibrary/MyLibrary.docc/Extensions/Foo.md": """
+        # ``MyLibrary/Foo``
+        Overview for Foo.
+
+        ## Topics
+        ### Group One
+        - ``methodA()``
+
+        ### Group Two
+        - ``methodB()``
+        """,
+      ],
+      enableBackgroundIndexing: true
+    )
+
+    try await renderDocumentation(
+      fileName: "Foo.swift",
+      project: project,
+      expectedResponses: [
+        ("1️⃣", .renderNode(kind: .symbol, containing: "Overview for Foo.")),
+        ("1️⃣", .renderNode(kind: .symbol, containing: "Group One")),
+        ("1️⃣", .renderNode(kind: .symbol, containing: "Group Two")),
+        ("1️⃣", .renderNode(kind: .symbol, containing: "methodA()")),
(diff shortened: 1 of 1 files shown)


matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows

Tests/SourceKitLSPTests/DoccDocumentationTests.swift
```diff
@@ -885,6 +885,59 @@ final class DoccDocumentationTests: SourceKitLSPTestCase {
     )
   }
 
+  func testMarkdownExtensionTopics() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "MyLibrary/Foo.swift": """
+          public struct  1️⃣2️⃣3️⃣4️⃣5️⃣ Foo {
```
ahoppen: Having 5 location markers for the same location seems a little odd to me. Would it be possible to change this to only have a single location marker and then adjust `renderDocumentation` to take tuples of location marker and expected node instead, accounting for the fact that there may be multiple location markers at every given location?

Steffeeen: I adjusted `renderDocumentation` to take a list of tuples and added fallback versions of `renderDocumentation` to avoid having to change all other test cases.

Steffeeen:
Could you please take a look at the changes @ahoppen? Is there anything else I should do to get this merged?

Tests/SourceKitLSPTests/DoccDocumentationTests.swift
```diff
@@ -990,82 +1044,104 @@ private func renderDocumentation(
...
+  testClient: TestSourceKitLSPClient,
+  uri: DocumentURI,
+  positions: DocumentPositions,
+  expectedResponses: [(String, PartialConvertResponse)],
+  file: StaticString = #filePath,
+  line: UInt = #line
 ) async {
   for marker in positions.allMarkers {
-    guard let expectedResponse = expectedResponses[marker] else {
+    let expectedResponsesForMarker = expectedResponses.filter({ $0.0 == marker }).map({ $0.1 })
```
ahoppen: If you add labels to `expectedResponses: [(marker: String, response: PartialConvertResponse)]`, this can become the following, which is a little more readable


```suggestion
    let expectedResponsesForMarker = expectedResponses.filter({ $0.marker == marker }).map(\.response)
```

Tests/SourceKitLSPTests/DoccDocumentationTests.swift
```diff
@@ -990,82 +1044,104 @@ private func renderDocumentation(
...
+
+    if expectedResponsesForMarker.isEmpty {
       XCTFail("No expected response was given for marker \(marker)", file: file, line: line)
```
ahoppen: If we no longer return, this could become an `XCTAssertFalse`, right?

```suggestion
    XCTAssertFalse(expectedResponsesForMarker.isEmpty, "No expected response was given for marker \(marker)", file: file, line: line)
```

Steffeeen:
Should I squash this into a single commit or do we want to keep both for clarity?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2474, opened 2026-02-04
Fix inverted log check when updating the index store

rockbruno:
I think this check seems inverted; it says the unit is up-to-date but it checks for the opposite. I bumped into this when investigating a case where the logs would claim something would not be indexed and then follow-up by indexing it.


plemarquand:
@swift-ci please test

Issue #2472, opened 2026-02-02
Semantic highlighting does not distinguish escaped characters in string literals


Steffeeen:
### Swift version

swift-6.2.3-RELEASE

### Platform

macOS Sequoia 15.4.1

### Editor

Visual Studio Code with Swift extension 2.16.0

### Description

When using semantic highlighting, escaped characters in string literals are not highlighted in a separate color. The (default) TextMate grammar-based syntax highlighting creates separate tokens for escaped characters.

With semantic highlighting:

<img width="577" height="104" alt="Image" src="https://github.com/user-attachments/assets/3c82c5c4-01eb-4919-a46b-86778392a47f" />

Without semantic highlighting, using the TextMate grammar-based highlighting:

<img width="486" height="102" alt="Image" src="https://github.com/user-attachments/assets/8a0560b7-5f7d-4212-80a1-2befb54ef957" />

VSCode token debug hover showing the different token types:

<img width="557" height="507" alt="Image" src="https://github.com/user-attachments/assets/dfeaca97-8711-4d04-906f-355dabd8fd18" />

I’m aware that sourcekit-lsp currently only forwards semantic tokens provided by sourcekitd / swift-syntax, and does not perform additional tokenization itself.

That said, it seems possible for this behavior to be implemented at the LSP level by augmenting or post-processing semantic tokens, even if the upstream tokens do not currently distinguish escaped characters.

However, I’m unsure whether this is the correct layer to address this issue. It may be more appropriate to implement this upstream. This would avoid splitting the tokenization logic across multiple projects. But I don't know which other projects consume the tokens and if it is thus possible / feasible to change this upstream.

I am open to working on this and creating a PR.

### Steps to Reproduce

1. Ensure semantic highlighting is turned on in VSCode: `"editor.semanticHighlighting.enabled": true`
2. Open a Swift file with a string containing escaped characters, for example `let a = "test \n \r \\ \' \""`

RedCMD:
Semantic highlighting should only be used as an enhancement, for when TextMate falls short, not a replacement

so basic tokens like strings, comments etc should be dropped
which `sourcekit-lsp` could prob easily do?
and maybe add a setting to enable/disable it

ahoppen:
[...]

PR #2485, opened 2026-02-17
Only create semantic tokens in cases where they are actually needed

Steffeeen:
This PR explores generating semantic tokens only in cases where they are actually needed. Currently, semantic tokens are generated for nearly everything which can lead to issues as seen in #2472. Note that the linked issue is originally only about string literals, however in this PR I have removed other semantic tokens which seemed unnecessary.

In most cases not reporting the semantic tokens does not change the colors in the editor. This is for example the case for comments and number literals.

By not reporting semantic tokens for keywords, the color of non-control-flow (e.g., `func` or `struct` or `let`) keywords changed from `#c586c0` to `#569cd6` due this being the default color for non-control-flow keywords in VSCode. Something similar happens for regex literals where the color changed from `#646695` to a few different ones based on the syntax highlighting of the regex expression.

For string literals the color stays mostly the same except for escaped characters. These are now highlighted separately inside the string literal.

Before:
Screenshot 2026-02-17 at 17 35 58

After:
Screenshot 2026-02-17 at 17 36 33

As comparing screenshots is really annoying and cannot be done automatically, I created https://github.com/Steffeeen/vscode-syntax-highlighting-tester. It uses the TextMate grammar and language server to reproduce the syntax highlighting of VSCode and produce an HTML file showing the differences between two different versions. I generated example code using AI to have a comparison of how the changes impact different syntax constructs. This comparison can be found here: https://semantic-token-changes-diff.netlify.app. I can easily generate a diff for any swift code and can provide help if anyone wants to use the tool themselves to compare changes.

I'm open to discussing if we want to remove the semantic tokens for keywords as it causes significant changes in the syntax highlighting. Existing users are used to `#c586c0` being the color for all keywords. However, using `#569cd6` is closer to what VSCode does for other languages, for example TypeScript, where the `function` and `class` keywords are in `#569cd6`, while control-flow keywords like `if` or `return` are in `#c586c0`. Additionally, if the semantic token calculation takes a bit of time the keywords also take a short time to get their correct color. For this example I added a 2-second delay before the semantic tokens are returned to better illustrate this:

https://github.com/user-attachments/assets/499eb5f3-e9b7-4b1a-b231-0b7b765800c9

There are also some other syntax constructs that this PR currently changes in a way that is not ideal. However, these could probably be fixed in the TextMate grammar. This is mostly the case for `#if` and macros like `#expect`. 

Before:
Screenshot 2026-02-17 at 18 01 26

After:
Screenshot 2026-02-17 at 18 01 57

This can also be seen in the linked diff at the bottom of the file. 

The code is also not final, I just replaced the returns with `return nil` to make the diff more clear and to be able to easily toggle between generating semantic tokens or not for different syntax.


Steffeeen:
> I thought about this a little more and we already have precedence for where we deviate from the LSP spec in order to work around potential bugs in the editor with the `cancelTextDocumentRequestsOnEditAndClose` configuration option. What do you think about adding a new configuration option `reportSyntacticHighlightInSemanticTokens` to the configuration options that defaults to `false`. When set to `true`, we maintain the current behavior. Otherwise we don’t report syntactic tokens.

This seems like a good idea to allow users to easily switch back to the old behavior, if they encounter a bug or other unwanted behavior.

> Also, if you could file issues to the TextMate grammar (or just fix them) for any issues that you find in the TextMate grammar, I’d really appreciate that.

👍

> And one last thing that I’d like to explore is what happens if we don’t generate any tokens using swift-syntax if `reportSyntacticHighlightInSemanticTokens == false` (see `mergedAndSortedTokens`). Maybe those are exactly the tokens that we should stop reporting in that mode, I don’t exactly remember what gets reported by sourcekitd’s semantic tokens request and what doesn’t.

I tried this out, and it seems like your intuition was correct. The only difference between using the changes in this PR and just removing all swift-syntax tokens seems to be in how macros are highlighted. This diff is between the changes in this PR on the left and removing the swift-syntax tokens on the right: https://semantic-tokens-without-swift-syntax.netlify.app

ahoppen:
In that case, let’s just not run the swift-syntax based highlighting if `reportSyntacticHighlightInSemanticTokens == false` and let’s consider any highlight issues from that as issues in the grammar.

hamishknight:
> I don’t exactly remember what gets reported by sourcekitd’s semantic tokens request and what doesn’t.

AFAIK we only report semantic tokens for decl references, everything else is syntactic

> What do you think about adding a new configuration option `reportSyntacticHighlightInSemanticTokens` to the configuration options that defaults to `false`

That seems reasonable to me, I guess my only concern here is for complicated bits of grammar where attempting to match the exact parsing logic in swift-syntax is likely to be a challenge for TextMate grammars, regex literals are the main thing that come to mind. e.g it seems like the Swift TextMate grammar will consider `x/y/` a regex literal, but this is actually a binary and postfix operator. It looks like it also does the same for `x/y/z`. That being said, the behavior we have today is already insufficient since given we don't report a regex literal, the TextMate grammar will still "win" (at least for tokens where we don't report the decl reference).

Steffeeen:
I added the configuration option and removed the other commit which added a bunch of `return nil`s. I still need to adjust the tests to match the new behavior.

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -71,10 +72,14 @@ extension SwiftLanguageService {
     try Task.checkCancellation()
 
     let tokens =
-      await tree
-      .classifications(in: range)
-      .map { $0.highlightingTokens(in: snapshot) }
-      .reduce(into: SyntaxHighlightingTokens(tokens: [])) { $0.tokens += $1.tokens }
+      if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
```
ahoppen: We shouldn’t need to create the syntax tree at all if `reportSyntacticHighlightInSemanticTokensOrDefault` is `false`. I suspect that it’ll be easiest to have two separate branches here depending on whether syntactic tokens are enabled. Then we also don’t need to go through `mergingTokens` anymore.

Sources/SKOptions/SourceKitLSPOptions.swift
```diff
@@ -385,6 +385,14 @@ public struct SourceKitLSPOptions: Sendable, Codable, Equatable {
     return cancelTextDocumentRequestsOnEditAndClose ?? true
   }
 
+  /// Whether the response for `textDocument/semanticTokens` should include semantic tokens for syntactic and semantic
+  /// highlighting or just semantic highlighting.
```
ahoppen: Could you include a short rationale when this should be enabled by clients (ie. if their syntactic grammars are insufficient)?

ahoppen (commented):
I hope updating the tests isn’t too much work. Likely, we want to just enable this setting for any tests that mostly tested syntactic tokens and have a few tests that only test semantic tokens.

Steffeeen:
One idea I had for the tests is adding a `isSemantic` Bool to the `TokenSpec` struct. We could then mark all tokens which come from SourceKit in the expected results with this field. This would allow us to reuse all existing tests and just run each test twice, once with `reportSyntacticHighlightInSemanticTokens` set to true and once with it set to false and filtering out the `TokenSpec`s which are not semantic tokens from SourceKit. What do you think about that approach?

Steffeeen:
I implemented the approach I described above. For now, I have kept it in a separate commit. I intend to squash the commits before merging.

ahoppen (approved):
I really like the approach to add a property to `TokenSpec`. PR looks good to me. Do you want to squash your commits and we can merge this?

Steffeeen:
Squashed the commits.

Steffeeen:
@swift-ci Please test

Steffeeen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Looks good, I’m just holding this back because I’d like to hear @rintaro and @hamishknight’s opinion on the changed behavior and the new configuration option.

Steffeeen:
I have generated a diff for the final changes: https://semantic-tokens-sourcekit-only.netlify.app.

I also noticed that my tool caused the `assert()` in SyntaxHighlightingTokens.swift:42 to fail because the semantic tokens from SourceKit do not seem to be sorted, adding `.sorted({ $0.start < $1.start })` to SemanticTokens.swift:70 fixed this. The tests did not fail however before this change.

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -58,8 +60,23 @@ extension SwiftLanguageService {
...
     try Task.checkCancellation()
 
+    let semanticTokens =
+      await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+      ?? SyntaxHighlightingTokens(tokens: [])
+
+    if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
+      return try await calculateSyntacticHighlightingTokens(for: snapshot, in: range, semanticTokens: semanticTokens)
+    } else {
+      return semanticTokens
```
rintaro: Yeah, we need `.sorted { $0.start < $1.start }` here.

Alternatively, could you do something like this? I think this makes it a bit easier to see how the behavior differs depending on the options:

```swift
    async let semanticTokens = ... // 'async' so we can get semantic and syntactic tokens simultaneously.

    let tokens = if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
      await syntacticTokens(for: snapshot, in: range).mergingTokens(with: semanticTokens)
    } else{
      await semanticTokens
    }
    return tokens.sorted { $0.start < $1.start }
  }
```
where `syntacticHighlightingTokens(for:in:)` is only returning unsorted syntactic tokens.
```swift
  private func syntacticTokens(
    for snapshot: DocumentSnapshot,
    in range: Range<Position>?,
  ) async -> SyntaxHighlightingTokens {
    let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
    let range = if let range {
      snapshot.byteSourceRange(of: range)
    } else {
      tree.range
    }

    return tree
      .classifications(in: range)
      .map { $0.highlightingTokens(in: snapshot) }
      .reduce(into: SyntaxHighlightingTokens(tokens: [])) { $0.tokens += $1.tokens }
}
```

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -58,8 +60,23 @@ extension SwiftLanguageService {
   ) async throws -> SyntaxHighlightingTokens {
     try Task.checkCancellation()
 
+    let semanticTokens =
+      await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
```
rintaro, line 64: I feel strange for ignoring `range` completely for `semanticHighlightingTokens(for:)` here. Maybe we should pass range to it, and filter out tokens outside. But I guess that's not the scope of this PR.

rintaro (commented):
Apologies for not reviewing sooner.
I think this is a reasonable change, including the options and the default value.
Just a few code style comments 🙏

Sources/SwiftLanguageService/SemanticTokens.swift
Steffeeen: Yeah, I agree, this also already came up in #2497. So I did not do anything about it. I can provide a follow-up PR that addresses this. But maybe the better approach would be to add the range to the request to SourceKit and let SourceKit handle this directly. For large files I imagine this would be a bit better in terms of performance as SourceKit does not have to return semantic tokens for the entire file, only to have a potentially large amount of them be filtered out by SourceKit-LSP.

Sources/SwiftLanguageService/SemanticTokens.swift
Steffeeen: Adjusted the code 👍

rintaro:
@swift-ci Please test

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -58,30 +60,37 @@ extension SwiftLanguageService {
...
 
     let range =
       if let range {
         snapshot.byteSourceRange(of: range)
       } else {
-        await tree.range
+        tree.range
       }
 
-    try Task.checkCancellation()
```
hamishknight, line 71: Any reason to remove the check for cancellation here?

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -543,7 +547,9 @@ final class SemanticTokensTests: SourceKitLSPTestCase {
   }
 
   func testReplaceUntilEndOfToken() async throws {
-    let testClient = try await TestSourceKitLSPClient()
+    var options = try await SourceKitLSPOptions.testDefault()
+    options.reportSyntacticHighlightInSemanticTokens = true
```
hamishknight: Would it make sense for this to be the default for `testDefault`?

Sources/SwiftLanguageService/SemanticTokens.swift
Steffeeen: No, not really. I just assumed it was no longer needed, as it also wasn't in the suggestion by @rintaro. I can add it back in, if you want.

Tests/SourceKitLSPTests/SemanticTokensTests.swift
Steffeeen: As the default value for the option is false, I would argue that it should also be false in all tests, except the ones that explicitly enable it.

We could also adjust the few test cases which explicitly set it to only use semantic tokens reported by SourceKit to not have to set the option. Most of these test cases test the incremental updating of the tokens after an edit to the document. We could maybe also introduce a helper for such tests and then also do the test twice, once with the option set to true and once with the option set to false. But I guess that would be in a future PR.

Steffeeen:
I'm not really sure why the Linux CI failed. The `build-script` invocation failed with exit code 1, but I don't see any error message.

Tests/SourceKitLSPTests/SemanticTokensTests.swift
hamishknight: My main thinking here is that it's better to default to over-testing the behavior than to under-testing the behavior, even if it's disabled by default

Sources/SwiftLanguageService/SemanticTokens.swift
hamishknight: I think it makes sense to still have given we may suspend to retrieve the syntax tree

hamishknight:
The test output is hard to read, but this was the actual failure, it's a known non-deterministic failure:

```
[2026-03-12T03:57:18.407Z] Test Case 'CrashRecoveryTests.testPreventClangdCrashLoop' failed (32.087 seconds)
[2026-03-12T03:57:18.407Z] Test Suite 'CrashRecoveryTests' failed at 2026-03-12 03:56:43.214
[2026-03-12T03:57:18.407Z] 	 Executed 1 test, with 1 failure (1 unexpected) in 32.087 (32.087) seconds
[2026-03-12T03:57:18.407Z] Test Suite 'Selected tests' failed at 2026-03-12 03:56:43.215
[2026-03-12T03:57:18.407Z] 	 Executed 1 test, with 1 failure (1 unexpected) in 32.087 (32.087) seconds
```

Tests/SourceKitLSPTests/SemanticTokensTests.swift
Steffeeen: I see your point and will change the tests and `testDefault()`.

Sources/SwiftLanguageService/SemanticTokens.swift
Steffeeen: Added it back in.

Steffeeen:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test Linux

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

PR #2488, opened 2026-02-20
Add test to verify RemoveImports works inside active #if clauses

amanthatdoescares:
### This PR expands test coverage for the Remove Unused Imports code action.
Fix: https://github.com/swiftlang/sourcekit-lsp/issues/2335
#### Existing tests verify that:
Unused top-level import declarations are removed correctly
The code action is not available when the source file contains compilation errors

This PR begins adding test coverage for import declarations nested inside conditional compilation blocks. In particular, it verifies that unused imports inside active #if clauses are considered by the code action.

This helps guard against regressions when extending RemoveUnusedImports to account for conditional compilation regions in future changes.

### DIFF
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1145,11 +1145,318 @@ final class CodeActionTests: SourceKitLSPTestCase {
 
   // DONE: testRemoveUnusedImportsFromActiveIfClause
   // DONE: testRemoveUnusedImportsDoesNotRemoveImportsFromInactiveIfClause
-  // TODO: testRemoveUnusedImportsDoesNotRemoveImportsInIfFalseBlock
-  // TODO: testRemoveUnusedImportsFromIfTrueBlock
-  // TODO: testRemoveUnusedImportsMixedActiveInactiveRegions
-  // TODO: testRemoveUnusedImportsNestedInactiveRegions
-  // TODO: testRemoveUnusedImportsTopLevelAndConditional
+  // DONE: testRemoveUnusedImportsDoesNotRemoveImportsInIfFalseBlock
+  // DONE: testRemoveUnusedImportsFromIfTrueBlock
+  // DONE: testRemoveUnusedImportsMixedActiveInactiveRegions
+  // DONE: testRemoveUnusedImportsNestedInactiveRegions
+  // DONE: testRemoveUnusedImportsTopLevelAndConditional
+
+  func testRemoveUnusedImportsFromIfTrueBlock() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "LibA/LibA.swift": "",
+        "Test/Test.swift": """
+        #if true
+        1️⃣import LibA
+        #endif
+        """,
+      ],
+      manifest: """
+        let package = Package(
+          name: "MyLibrary",
+          targets: [
+            .target(name: "LibA"),
+            .target(
+              name: "Test",
+              dependencies: ["LibA"],
+              swiftSettings: [
+                .enableUpcomingFeature("MemberImportVisibility")
+              ]
+            )
+          ]
+        )
+        """,
+      capabilities: clientCapabilitiesWithCodeActionSupport,
+      enableBackgroundIndexing: true
+    )
+
+    let (uri, positions) = try project.openDocument("Test.swift")
+
+    let importResult = try await project.testClient.send(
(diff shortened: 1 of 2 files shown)


Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1649,6 +1143,118 @@ final class CodeActionTests: SourceKitLSPTestCase {
     )
   }
 
+  func testRemoveUnusedImportsFromActiveIfClause() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Test/Test.swift": """
+        1️⃣#if canImport(Darwin)
```
rintaro: `canImport(Darwin)` is not active on non-Darwin environment. Can we use more reliable tests e.g.
In `Package.swift`
```swift
swiftSettings: [.define("FLAG")]
```
then in `Test/Test.swift`
```
#if FLAG
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1649,6 +1143,118 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+        arguments: removeUnusedImportsCommand.arguments
+      )
+    )
+  }
+
+  func testRemoveUnusedImportsDoesNotRemoveImportsFromInactiveIfClause() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Test/Test.swift": """
+        1️⃣#if canImport(Darwin)
```
rintaro: Ditto. Please avoid using `canImport`

Tests/SourceKitLSPTests/CodeActionTests.swift
rintaro: Also, do you know why this is rendered as "1□" but the `CodeActionResult` call below is displayed as "1️⃣"  in my GitHub UI? 🤔 The UTF8 code sequence is different?

Image

amanmaurya92: image
The marker is the keycap emoji (1️⃣), which is composed of multiple Unicode code points. GitHub’s diff viewer likely doesn’t fully render the combining keycap, so it shows up as 1□. The UTF-8 sequence is indeed different, but this is expected.

Tests/SourceKitLSPTests/CodeActionTests.swift
amanmaurya92: Thanks for the suggestion — I’ve replaced canImport(...) with a defined flag to make the test deterministic across environments. Please let me know if there are additional scenarios or edge cases you’d like to see covered in this PR to make the tests more robust.

amanmaurya92:
@rintaro I’m also interested in working on improving the RemoveUnusedImports code action itself, and thought it would be useful to add these tests first to lock in the expected behaviour around conditional compilation before making any implementation changes.

rintaro:
@swift-ci Please test

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1649,6 +1143,125 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+      )
+    )
+  }
+
+  func testRemoveUnusedImportsDoesNotRemoveImportsFromInactiveIfClause() async throws {
+    let project = try await SwiftPMTestProject(
+      files: [
+        "Test/Test.swift": """
+        1️⃣#if FLAG
+        import Darwin
```
rintaro: I guess this still fails in non-Darwin environments because `import Darwin` causes compilation errors.

amanmaurya92: yes, i changed it to Foundation now.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

ahoppen:
@amanthatdoescares The tests you added don’t pass because they test functionality that doesn’t exist yet. Please ensure that the test you add pass locally before putting them up for PR review.

ahoppen:
@DPrakashhh started fixing this issue in https://github.com/swiftlang/sourcekit-lsp/pull/2457. The comments in that PR would also apply here. Also, I don’t want to have two concurrent PRs implementing the same feature and since https://github.com/swiftlang/sourcekit-lsp/pull/2457 already has already gone through some iterations, I’d prefer that.

PR #2489, opened 2026-02-21
[Test/PlaygroundsDiscovery] Add "/refresh" style entry point requests

rintaro:
(Based on #2500)

When the client opts in to `workspace/tests/refresh` or `workspace/playgrounds/refresh` via experimental client capabilities, SourceKit-LSP now maintains a proactive cache of the current test and
playground lists and sends the corresponding `workspace/.../refresh` notification whenever the cache changes. `workspaceTests()`/`workspacePlaygrounds()` then serve subsequent fetch requests directly from the cache.

Add `EntryPointManager`: runs background scans, stores the results, fires callbacks on changes:
 - Start scanning when build targets are updated including initial updates, any watched files are changed, and index is updated.
 - Send `*/refresh` server initiated requests when the cache has changed.
 - Coalesces rapid invalidations by cancelling any in-flight refresh task.

NOTE: `/refresh` is not sent when the user edits the open documents. But the scanning for the cache takes in-memory documents into account. This is just a design decision. The cache should capture as newest status as possible, but `/refresh` shouldn't be sent for every key stroke.

Also:
- Simplify `SourceKitIndexDelegate` from an `actor` with `AtomicInt32` to a plain `class`, since it is only called from IndexStoreDB's internal serial dispatch queue.

### DIFF
--- a/Sources/SKTestSupport/TestSourceKitLSPClient.swift
+++ b/Sources/SKTestSupport/TestSourceKitLSPClient.swift
@@ -391,6 +391,33 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
     requestHandlers.value.append((requestHandler: requestHandler, isOneShot: false))
   }
 
+  /// Executes `operation` and waits until a request of the specified type is received from the server.
+  package func withWaitingFor<R: RequestType, Value>(
+    _: R.Type,
+    timeout: Duration = defaultTimeoutDuration,
+    pollingInterval: Duration = .milliseconds(10),
+    operation: () async throws -> Value,
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws -> Value where R.Response == VoidResponse {
+    // Register a one-shot handler that records when the request arrives.
+    let received = AtomicBool(initialValue: false)
+    self.handleSingleRequest { (_: R) in
+      received.value = true
+      return VoidResponse()
+    }
+
+    let result = try await operation()
+    try await repeatUntilExpectedResult(
+      timeout: timeout,
+      sleepInterval: pollingInterval,
+      { received.value },
+      file: file,
+      line: line
+    )
+    return result
+  }
+
   // MARK: - Conformance to MessageHandler
 
   /// - Important: Implementation detail of `TestSourceKitLSPServer`. Do not call from tests.
--- a/Sources/SourceKitLSP/EntryPointManager.swift
+++ b/Sources/SourceKitLSP/EntryPointManager.swift
@@ -1,0 +1,111 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SKUtilities
+
+/// Manages discovered tests and playgrounds.
+actor EntryPointManager {
+  weak let sourceKitLSPServer: SourceKitLSPServer?
+
+  // Implicitly-unwrapped because initializing it requires capturing `self`.
+  private nonisolated(unsafe) var refreshDebouncer: Debouncer<Void>!
+  private var currentRefreshTask: Task<Void, any Error>?
+
+  // Collected entry points in the workspaces.
+  private(set) var latestWorkspaceTests: [TestItem] = []
+  private(set) var latestPlaygrounds: [Playground] = []
+
+  // Callback functions when the list has changed. Non-nil means the client wants that entry point kind.
+  var onWorkspaceTestsChanged: (() -> Void)? = nil
+  var onWorkspacePlaygroundsChanged: (() -> Void)? = nil
+
+  init(sourceKitLSPServer: SourceKitLSPServer) {
+    self.sourceKitLSPServer = sourceKitLSPServer
+    self.refreshDebouncer = nil
+    self.refreshDebouncer = Debouncer(debounceDuration: .seconds(2)) { [weak self] in
+      await self?.refreshImpl()
+    }
+  }
+
+  /// Sets up the callbacks for each entry point kind.
+  /// If `nil`, the entry point of the kind will not be monitored.
+  func setCallbacks(
+    onWorkspaceTestsChanged: (() -> Void)?,
+    onWorkspacePlaygroundsChanged: (() -> Void)?,
+  ) {
+    self.onWorkspaceTestsChanged = onWorkspaceTestsChanged
+    self.onWorkspacePlaygroundsChanged = onWorkspacePlaygroundsChanged
+  }
+
+  /// Schedules a new refresh of the cached entry point lists.
+  ///
+  /// - Parameters:
+  ///   - debounce: If `true`, the refresh is debounced to coalesce rapid successive refresh requests,
+  ///     such as when multiple index changes arrive in quick succession. If `false` (default), the
+  ///     refresh starts immediately.
+  func refresh(debounce: Bool = false) {
+    guard onWorkspaceTestsChanged != nil || onWorkspacePlaygroundsChanged != nil else {
+      // If there's no listener, no need to scan things.
+      return
+    }
+
+    Task {
+      await refreshDebouncer.scheduleCall()
+      if !debounce {
+        await refreshDebouncer.flush()
+      }
+    }
+  }
+
+  private func refreshImpl() {
+    self.currentRefreshTask?.cancel()
+
+    self.currentRefreshTask = Task { [weak self] in
+      guard let self else {
+        return
+      }
+      try Task.checkCancellation()
+
+      async let testsTask = self.refreshTestsImpl()
+      async let playgroundsTask = self.refreshPlaygroundsImpl()
+      _ = await (testsTask, playgroundsTask)
+
+      // No need to clear 'self.currentRefreshTask'; the next call to `refresh()` will cancel and replace it.
+    }
+  }
+
+  private func refreshTestsImpl() async {
+    guard let onWorkspaceTestsChanged, let sourceKitLSPServer else {
+      // 'onWorkspaceTestsChanged == nil' means the client is not interested in 'workspace/tests/refresh' request.
+      return
+    }
+    let newTests = await TestDiscovery(sourceKitLSPServer: sourceKitLSPServer).workspaceTests()
+    // Never modify the result here. modify 'workspaceTests()' if needed so other clients can gent the same result.
+    if !Task.isCancelled, newTests != latestWorkspaceTests {
+      self.latestWorkspaceTests = newTests
+      onWorkspaceTestsChanged()
+    }
+  }
+
+  private func refreshPlaygroundsImpl() async {
+    guard let onWorkspacePlaygroundsChanged, let sourceKitLSPServer else {
+      return
+    }
+    let newPlaygrounds = await PlaygroundDiscovery(sourceKitLSPServer: sourceKitLSPServer).workspacePlaygrounds()
+    // Never modify the result here. modify 'workspacePlaygrounds()' if needed so other clients can get the same results.
+    if !Task.isCancelled, newPlaygrounds != latestPlaygrounds {
+      self.latestPlaygrounds = newPlaygrounds
+      onWorkspacePlaygroundsChanged()
+    }
+  }
+}
--- a/Sources/SourceKitLSP/PlaygroundDiscovery.swift
+++ b/Sources/SourceKitLSP/PlaygroundDiscovery.swift
@@ -17,25 +17,22 @@ import SemanticIndex
 import SwiftExtensions
 @_spi(SourceKitLSP) import ToolsProtocolsSwiftExtensions
 
-extension SourceKitLSPServer {
+struct PlaygroundDiscovery {
+  let sourceKitLSPServer: SourceKitLSPServer
 
-  /// Return all the playgrounds in the given workspace.
-  ///
-  /// The returned list of playgrounds is not sorted. It should be sorted before being returned to the editor.
+  init(sourceKitLSPServer: SourceKitLSPServer) {
+    self.sourceKitLSPServer = sourceKitLSPServer
+  }
+
   private func playgrounds(in workspace: Workspace) async -> [Playground] {
-    // If files have recently been added to the workspace (which is communicated by a `workspace/didChangeWatchedFiles`
-    // notification, wait these changes to be reflected in the build server so we can include the updated files in the
-    // playgrounds.
-    await workspace.buildServerManager.waitForUpToDateBuildGraph()
-
     let playgroundsFromSyntacticIndex = await workspace.syntacticIndex.playgrounds()
 
-    // We don't need to sort the playgrounds here because they will get sorted by `workspacePlaygrounds` request handler
+    // We don't need to sort the playgrounds here because they will get sorted by `workspacePlaygrounds`
     return playgroundsFromSyntacticIndex
   }
 
-  func workspacePlaygrounds(_ req: WorkspacePlaygroundsRequest) async throws -> [Playground] {
-    return await self.workspaces
+  func workspacePlaygrounds() async -> [Playground] {
+    return await sourceKitLSPServer.workspaces
       .concurrentMap { await self.playgrounds(in: $0) }
       .flatMap { $0 }
       .sorted { $0.location < $1.location }
--- a/Sources/SourceKitLSP/SourceKitIndexDelegate.swift
+++ b/Sources/SourceKitLSP/SourceKitIndexDelegate.swift
@@ -18,44 +18,41 @@ import SwiftExtensions
 @_spi(SourceKitLSP) import ToolsProtocolsSwiftExtensions
 
 /// `IndexDelegate` for the SourceKit workspace.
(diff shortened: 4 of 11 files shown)


Sources/SourceKitLSP/SourceKitIndexDelegate.swift
```diff
@@ -18,44 +18,41 @@ import SwiftExtensions
...
-  package init(mainFilesChangedCallback: @escaping @Sendable () async -> Void) {
-    self.mainFilesChangedCallback = mainFilesChangedCallback
+  package init(callback: @escaping @Sendable () async -> Void) {
+    self.callback = callback
   }
 
-  nonisolated package func processingAddedPending(_ count: Int) {
-    pendingUnitCount.value += Int32(count)
+  package func processingAddedPending(_ count: Int) {
+    pendingUnitCount += count
```
rintaro, line 34: This is safe because `IndexStoreDB`'s delegate methods are called in a specific serial queue. Also the previous implementation wasn't thread safe either because `pendingUnitCount.value += Int32(count)` caused 2 separate read and write operations non-atomically.

Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift
```diff
@@ -1045,12 +1036,12 @@ final class WorkspaceTestDiscoveryTests: SourceKitLSPTestCase {
         TestItem(
           id: "MyLibraryTests.MyTests",
           label: "MyTests",
-          location: Location(uri: uri, range: positions["1️⃣"]..<positions["4️⃣"]),
+          location: Location(uri: uri, range: positions["1️⃣"]..<positions["1️⃣"]),
```
rintaro, line 1035: Only `DocumentTestsRequest` returns the correct range to XCTest test cases. Previously tests in in-memory documents are retrieved with internal `DocumentTestsRequest`, but since this PR stops doing that, `workspace/tests` only uses the symbol position from the index.

Sources/SourceKitLSP/PlaygroundDiscovery.swift
```diff
@@ -17,26 +17,23 @@ import SemanticIndex
...
+  private func discoverPlaygrounds(in workspace: Workspace) async -> [Playground] {
     let playgroundsFromSyntacticIndex = await workspace.syntacticIndex.playgrounds()
 
     // We don't need to sort the playgrounds here because they will get sorted by `workspacePlaygrounds` request handler
     return playgroundsFromSyntacticIndex
   }
 
-  func workspacePlaygrounds(_ req: WorkspacePlaygroundsRequest) async throws -> [Playground] {
-    return await self.workspaces
-      .concurrentMap { await self.playgrounds(in: $0) }
```
rintaro, line 39: This `concurrentMap { }` was actually meaningless because `self` here was an actor `SourceKitLSPServer`. So they were only processed one-by-one.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -316,19 +317,12 @@ package protocol LanguageService: AnyObject, Sendable {
...
-  /// This is used as a fallback to show the test cases in a file if the index for a given file is not up-to-date.
-  ///
-  /// A return value of `nil` indicates that this language service does not support syntactic test discovery.
-  func syntacticDocumentTests(for uri: DocumentURI, in workspace: Workspace) async throws -> [AnnotatedTestItem]?
-
   /// Syntactically scans the file at the given URL for tests declared within it.
   ///
   /// Does not write the results to the index.
   ///
   /// The order of the returned tests is not defined. The results should be sorted before being returned to the editor.
```
rintaro, line 326:
```swift
/// A return value of `nil` indicates that this language service does not support syntactic test discovery.
```

ahoppen (commented):
> The practical impact should be minimal because:
> - Open documents are visible in the editor, so clients should use textDocument/tests for those files to reflect real-time edits

I really dislike this direction, because
1. I’d like us to move towards being more friendly to in-memory documents, treating in-memory documents just the same as on-disk documents and being in a state where in-memory edits are reflected for global requests like this as well. And this is going in the opposite direction. 
2. Source files will typically be compiled before the tests are run automatically and we should be reflecting the state the source files will be in once they are compiled than trying to represent an old state.
3. I don’t think we’ve documented anywhere that clients should merge the results from `textDocument/tests` with `workspace/tests`. I know that VS Code does it but in general clients shouldn’t need to do the fairly complex merging logic here.

How noise, in practice, is it if we debounce `workspace/tests/refresh` by, say 3 or 5 seconds and which fields change. If it’s mostly the location that causes the changes, should we add a `test/resolve` method to which a test ID can be given and that resolves the location or something like that?

rintaro:
Okay. I will reconsider the strategy around in-memory documents :)

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test Linux

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test macOS

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1122,6 +1146,7 @@ extension SourceKitLSPServer {
 
     var experimentalCapabilities: [String: LSPAny] = [
       WorkspaceTestsRequest.method: .dictionary(["version": .int(2)]),
+      WorkspaceTestsRefreshRequest.method: .dictionary(["version": .int(1)]),
```
ahoppen, line 1151: Do we also need to advertise `WorkspacePlaygroundsRefreshRequest` here?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2708,7 +2736,13 @@ extension SourceKitLSPServer {
...
 
   func workspaceTests(_ req: WorkspaceTestsRequest) async throws -> [TestItem] {
-    return await TestDiscovery(sourceKitLSPServer: self).workspaceTests()
+    if self.capabilityRegistry?.clientHasWorkspaceTestsRefreshSupport ?? false {
+      // If the client supports 'workspace/tests/refresh', return the pre-populated tests.
+      return await self.entryPointManager.latestWorkspaceTests
+    } else {
+      // Otherwise, retrieve tests from the current index.
+      return await TestDiscovery(sourceKitLSPServer: self).workspaceTests()
+    }
```
ahoppen, line 2769: Instead of having two pretty different code paths here (the way we compute `workspaceTests` in `EntryPointManager` may at some point diverge from this, should we always have an entry point manager that is always responsible for caching the latest result and we always use that here?

Sources/SourceKitLSP/EntryPointManager.swift
```diff
@@ -0,0 +1,87 @@
...
+import LanguageServerProtocol
+
+/// Manages discovered tests and playgrounds.
+actor EntryPointManager {
+  weak let sourceKitLSPServer: SourceKitLSPServer?
+  private var currentRefreshTask: Task<Void, any Error>?
+
+  // Collected entry points in the workspaces.
+  private(set) var latestWorkspaceTests: [TestItem] = []
+  private(set) var latestPlaygrounds: [Playground] = []
```
ahoppen, line 26: This may return stale results if `refresh` is called but the refresh hasn’t completed. I generally like a pattern where these variables store a `Task<[TestItem], any Error>?` that callers can await. Refreshing then replaces the task and clients can await it. If you also want cancellation behavior, you can make it a `RefCountedTask`, which automatically gets cancelled once nobody is interested in it anymore.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -389,7 +387,13 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
       createMainFilesProvider: { (initializationData, mainFilesChangedCallback) -> (any MainFilesProvider)? in
         await createIndex(
           initializationData: initializationData,
-          mainFilesChangedCallback: mainFilesChangedCallback,
+          indexChangedCallback: { [weak sourceKitLSPServer] in
+            // Notify that main files may have changed.
+            await mainFilesChangedCallback()
+
+            // Schedule updating entry point cache.
+            await sourceKitLSPServer?.entryPointManager.refresh()
```
ahoppen: The index changed callback is called *very often*. Should we debounce the refresh?

Sources/SKTestSupport/TestSourceKitLSPClient.swift
```diff
@@ -380,6 +380,29 @@ package final class TestSourceKitLSPClient: MessageHandler, Sendable {
...
+  package func withWaitingFor<R: RequestType, Value>(
+    _: R.Type,
+    timeout: Duration = defaultTimeoutDuration,
+    pollingInterval: Duration = .milliseconds(10),
+    operation: () async throws -> Value,
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) async throws -> Value where R.Response == VoidResponse {
+    // Register a one-shot handler that records when the request arrives.
+    let received = AtomicBool(initialValue: false)
```
ahoppen, line 404: Could we use an `XCTestExpectation` here so we don’t need `repeatUntilExpectedResult` at all?

Sources/SourceKitLSP/EntryPointManager.swift
rintaro: IMO, this should return immediately even if it's technically stale. The `/refresh` request for the current refresh task will be sent eventually (if something has changed) and the client should refetch it anyway.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I don't think `EntryManager` should background cache refresh at all unless client support `/refresh`.
If the client doesn't support `/refresh`, it probably either don't want `workspace/tests` at all, or has its own timing for re-fetching `workspace/tests` (by monitoring the file system etc.). So there'd be no much point to cache the results in the server.

As for the possibility of the divergence, I will add a comment in `TestDiscovery.refresh(Tests|Playgrounds)Impl()` not to diverge.

Sources/SKTestSupport/TestSourceKitLSPClient.swift
rintaro: One reason I didn't use `XCTestExpectation` was `fulfillmentOfOrThrow` doesn't report the file/line while `repeatUntilExpectedResult` does. Comparing these two I didn't find much benefit of it.

Maybe we should fix `fulfillmentOfOrThrow` to report `XCTFail` with correct file/line in it. But that's out of scope of this PR.

Sources/SourceKitLSP/Workspace.swift
rintaro: Added `debouneDuration` parameter to `EntryPointManager.refresh()` and set it to wait for 1 sec only for the index change callbacks. For file change callbacks, I think we should start the task immediately so we can report the latest results asap.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test macOS

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/37
@swift-ci Please test

Sources/SKTestSupport/TestSourceKitLSPClient.swift
ahoppen: You just fixed `fulfillmentOfOrThrow` to report file  + line though, right? https://github.com/swiftlang/sourcekit-lsp/blob/f1635f1209bda73de5fdbb39b70318abfbe9f521/Sources/SKTestSupport/Assertions.swift#L185

To be clear, I don’t think that’s blocking this PR for now but we should unify this if possible.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I personally think there also isn’t too much harm in caching the results, even if the client doesn’t support the `/refresh` request. I’m mostly just concerned in subtle differences in behaviors depending on which capabilities the client supports, which makes it harder to reproduce these issues if, eg. an engineer working on SourceKit-LSP is more familiar with VS Code but the issue only reproduces in neovim.

Sources/SourceKitLSP/EntryPointManager.swift
```diff
@@ -0,0 +1,97 @@
...
+      // If there's no listener, no need to scan things.
+      return
+    }
+
+    self.currentRefreshTask?.cancel()
+
+    self.currentRefreshTask = Task { [weak self] in
+      if debounceDuration != .zero {
+        try await Task.sleep(for: debounceDuration)
+      }
```
ahoppen: If we get a new call to `refresh` every 0.5 seconds with a debounce duration of 1s, we will never actually get to the refresh. `Debouncer` handles all of this. Should we use that instead?

Sources/SourceKitLSP/EntryPointManager.swift
ahoppen: I personally disagree. If you ask for the workspace tests, then you’re asking for the workspace tests for the current state of the project, not whatever SourceKit-LSP happens to have computed up until now.

rintaro: But most requests don't wait for up-to-date semantic index. I don't think it's worth to wait for it. Especially for this case, `/refresh` will be sent out when it's refreshed and anything has actually changed.

Sources/SourceKitLSP/EntryPointManager.swift
rintaro: Except it didn't 😅  Fixing it in https://github.com/swiftlang/sourcekit-lsp/pull/2581

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I see where you're coming from, but IMO it's just a matter of where we branch. As long as we have two different modes, there's always a risk of divergence.

Sources/SourceKitLSP/PlaygroundDiscovery.swift
```diff
@@ -17,25 +17,22 @@ import SemanticIndex
...
+  init(sourceKitLSPServer: SourceKitLSPServer) {
+    self.sourceKitLSPServer = sourceKitLSPServer
+  }
 
+  private func playgrounds(in workspace: Workspace) async -> [Playground] {
     let playgroundsFromSyntacticIndex = await workspace.syntacticIndex.playgrounds()
 
-    // We don't need to sort the playgrounds here because they will get sorted by `workspacePlaygrounds` request handler
+    // We don't need to sort the playgrounds here because they will get sorted by `workspacePlaygrounds`
     return playgroundsFromSyntacticIndex
```
hamishknight, line 31: Any reason for the intermediate binding?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1597,6 +1623,9 @@ extension SourceKitLSPServer {
     for languageService in languageServices.values.flatMap(\.self) {
       await languageService.filesDidChange(notification.changes)
     }
+
+    // Schedule updating entry point cache.
+    await entryPointManager.refresh()
```
hamishknight, line 1628: Refreshing the list of tests/playgrounds for every workspace on any file updating seems like it may be expensive, would it make more sense to at least limit it to the workspace that had the file update?

Sources/SourceKitLSP/EntryPointManager.swift
```diff
@@ -0,0 +1,113 @@
...
+
+  /// Schedules a new refresh of the cached entry point lists.
+  ///
+  /// Cancels any previously scheduled refresh task. Only the most recent refresh request will run.
+  ///
+  /// - Parameters:
+  ///   - debounceDuration: The duration to wait before starting the refresh.
+  ///     Defaults to `.zero` for immediate refresh. Use a non-zero duration to debounce
+  ///     rapid successive refresh requests, such as when multiple index change.
+  func refresh(debounce: Bool = false) {
```
hamishknight: Nit: doc comment no longer matches

rintaro:
https://github.com/swiftlang/swift/pull/88156
@swift-ci Please test

Sources/SourceKitLSP/PlaygroundDiscovery.swift
rintaro: It's not necessary. I _guess_ it was to clarify that we don't need to do anything here.

rintaro:
https://github.com/swiftlang/swift/pull/88156
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Filed https://github.com/swiftlang/sourcekit-lsp/issues/2582

Issue #549, opened 2019-05-30
[SR-10806] Implement workspace/symbol request


benlangmuir:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-10806      |
   |Radar             | None         |
   |Original Reporter | @benlangmuir      |
   |Type              | New Feature    |
   |Status            | Resolved        |
   |Resolution        | Done    |
        
   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | SourceKit-LSP    |
   |Labels            | New Feature, StarterBug        |
   |Assignee          | literalpie (JIRA)      |
   |Priority          | Medium      |

   

   md5: 88e6c20bbce2e4e811331802adca7a81

  </details>

**Issue Description:**

Implement the `workspace/symbol` request, which given a query string returns all symbols that match.

<https://microsoft.github.io/language-server-protocol/specification#workspace_symbol>

To implement this, I suggest exposing the following methods from [indexstore-db](https://github.com/apple/indexstore-db) to the Swift wrapper code:

-   `foreachSymbolName`

-   `foreachCanonicalSymbolOccurrenceByName`

Then SourceKit-LSP can use the `foreachSymbolName` to get all the symbol names and perform matching against them, followed by `foreachCanonicalSymbolOccurrenceByName` to get the symbol information.

Issue #543, opened 2019-08-09
[SR-11276] Sourcekit-LSP returns wrong position for protocol method implementations


swift-ci:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-11276      |
   |Radar             | None         |
   |Original Reporter | cukier (JIRA User)      |
   |Type              | Bug    |

<details>
  <summary>Environment</summary>

commit c0d2ad0eb8a2c371ca2c5c9f65b4f71baece24dc of sourcekit-lsp

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | SourceKit-LSP    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 1f8f35a85d2991c3ff9d52e044c6482d

  </details>

**Issue Description:**

``` java
protocol Aaa { 
    func bbb()
}
class Ccc { 
    func bbb() { } //5
}
extension Ccc: Aaa { } //7
class Ddd: Ccc { } //8
```

If you ask for implementations for function \`bbb\` in the protocol, Sourcekit-LSP will return positions at lines 7 and 8. I would expect it to return the position of the method at line 5.

ahoppen:
rdar://95597776

ahoppen:
This no longer reprocess for me.

PR #2492, opened 2026-02-22
Strip -working-directory from compiler args for non-file URI documents

loveucifer:
addresses  #549


loveucifer:
> `Also, I believe you meant that this fixes #539 instead of #543, right?

oh my bad , fixin it

loveucifer:
hmm if we need the fix in sourcekitd im just closing this pr :/

Issue #1277, opened 2024-05-13
Merge index-based and syntactic test discovery results


ahoppen:
In the `textDocument/tests` request, we can use syntactic test discovery to discover tests that weren’t built during indexing.

- Run an active regions sourcekitd request
- If there are inactive regions, discover tests from the inactive regions
- Add the syntactically discovered tests from the inactive regions to the tests discovered from the index

rdar://125496140

PR #2494, opened 2026-02-23
Merge index-based and syntactic test discovery results

loveucifer:
addresses #1277

### DIFF
--- a/Sources/SourceKitLSP/TestDiscovery.swift
+++ b/Sources/SourceKitLSP/TestDiscovery.swift
@@ -344,14 +344,22 @@ extension SourceKitLSPServer {
         index.unitTests(referencedByMainFiles: [mainFileUri.pseudoPath])
         .filter { $0.canBeTestDefinition }
 
+      // Syntactically discovered XCTest items that are not covered by the semantic index.
+      // `filterUsing(semanticSymbols:)` already passes through any item whose name doesn't appear in the
+      // semantic index (e.g. classes inside inactive `#if` regions), so we just need to include the
+      // syntactic XCTest results here rather than discarding them when the index branch is taken.
+      var syntacticXCTestItems: [AnnotatedTestItem] {
+        syntacticTests?.filter { $0.testItem.style == TestStyle.xcTest } ?? []
+      }
+
       if !testSymbols.isEmpty {
         let documentSymbols = await orLog("Getting document symbols for test ranges") {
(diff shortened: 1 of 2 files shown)


loveucifer:
first , I did a new `syntacticXCTestsInInactiveRegions` protocol method on `LanguageService` that fired a `source.request.activeregions` sourcekitd request to map out which byte ranges were active, then filtered the syntax-tree scan to only return items outside those ranges , i think it worked, but it was quite complicate imo , When I re-read `filterUsing(semanticSymbols:)` more carefully, I realised it already lets inactive-region tests pass through untouchedd (since the semantic index simply has no entry for symbols the compiler never saw), meaning the syntactic results we already had in hand contained exactly what we needed. The only  bug was that `documentTestsWithoutMergingExtensions` was silently discarding the XCTest items from `syntacticTests` whenever the semantic-index branch was taken, so I replaced it with a small deduplication step that appends any syntactic XCTest items whose ID isn't already covered by the index

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -362,12 +370,16 @@ extension SourceKitLSPServer {
...
           }
-        ) + syntacticSwiftTestingTests
+        )
+        let semanticXCTestIDs = Set(semanticXCTestItems.map { $0.testItem.id })
+        let unindexedXCTests = syntacticXCTestItems.filter { !semanticXCTestIDs.contains($0.testItem.id) }
+        return semanticXCTestItems + syntacticSwiftTestingTests + unindexedXCTests
       }
       if index.hasAnyUpToDateUnit(for: mainFileUri) {
         // The semantic index is up-to-date and doesn't contain any tests. We don't need to do a syntactic fallback for
         // XCTest. We do still need to return swift-testing tests which don't have a semantic index.
```
rintaro, line 380: Could you update this comment?

rintaro:
@swift-ci Please test macOS

rintaro:
Nice find that we don't really need `source.request.activeregions` for this!
At this point, I'm not really sure why we're still using the semantic index as the "main" data source, given that they don't have correct range info. After I merge this PR, I think I'm going to address it as a part of https://github.com/swiftlang/sourcekit-lsp/pull/2489

loveucifer:
Hmm failed might be a formatting error let me see !

loveucifer:
hmm it failed :(

PR #2495, opened 2026-02-23
Fix multiline DocC Returns rendering by preserving indentation

vanshkamra12:
Fixes swiftlang/vscode-swift#2119

Preview documentation wasn't rendering multiline Returns values correctly. Indented bullets appeared in the Discussion section instead of remaining part of the Return Value section.

## Root Cause
The bug was in `Sources/SwiftLanguageService/SymbolGraph.swift` where `.trimmingCharacters(in: .whitespaces)` was stripping ALL whitespace from documentation comments, including the leading indentation that DocC uses to parse hierarchical structures.

## Solution
- Removed `.trimmingCharacters(in: .whitespaces)` from both `docLineComment` and `docBlockComment` parsing
- Now only strips the `///` or `/** */` delimiters
- Preserves leading spaces/tabs so DocC can identify sub-items

### Before (broken):
```swift
/// - Returns: One of the following
///     * 0: field doesn't exist  ← Goes to Discussion section 
///     * 1: field exists         ← Goes to Discussion section

### DIFF
--- a/Tests/SourceKitLSPTests/DoccDocumentationTests.swift
+++ b/Tests/SourceKitLSPTests/DoccDocumentationTests.swift
@@ -66,6 +66,72 @@ final class DoccDocumentationTests: SourceKitLSPTestCase {
     )
   }
 
+  func testMultilineReturnsInDocLineComment() async throws {
+    // Regression test for https://github.com/swiftlang/vscode-swift/issues/2119
+    let testClient = try await TestSourceKitLSPClient()
+    let uri = DocumentURI(for: .swift)
+    let positions = testClient.openDocument(
+      """
+      /// Checks whether a field exists.
+      ///
+      /// - Returns: One of the following
+      ///     * 1️⃣0: field does not exist
+      ///     * 1: field exists
+      public func hexists() -> Int { return 1 }
+      """,
+      uri: uri
+    )
+    let response = try await testClient.send(
+      DoccDocumentationRequest(
+        textDocument: TextDocumentIdentifier(uri),
+        position: positions["1️⃣"]
+      )
+    )
+    let renderNodeString = response.renderNode
+    XCTAssertTrue(
+      renderNodeString.contains("0: field does not exist"),
+      "Expected continuation text in render node"
+    )
+    XCTAssertFalse(
+      renderNodeString.contains("\"kind\" : \"content\""),
+      "Continuation lines should not create a separate Discussion section"
+    )
+  }
+
+  func testMultilineReturnsInDocBlockComment() async throws {
+    // Regression test for https://github.com/swiftlang/vscode-swift/issues/2119
+    let testClient = try await TestSourceKitLSPClient()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by matthewbastien** [comment]
Please add a test case to verify your change.

**Comment by vanshkamra12** [comment]
@matthewbastien done sir i added the test cases!

File: Tests/SourceKitLSPTests/DoccDocumentationTests.swift [hunk 1]
```diff
@@ -66,6 +66,72 @@ final class DoccDocumentationTests: SourceKitLSPTestCase {
...
+        position: positions["1️⃣"]
+      )
+    )
+    let renderNodeString = response.renderNode
+    XCTAssertTrue(
+      renderNodeString.contains("0: field does not exist"),
+      "Expected continuation text in render node"
+    )
+    XCTAssertFalse(
+      renderNodeString.contains("\"kind\" : \"content\""),
```
matthewbastien, line 96: I don't see how this tests whether or not a discussion section exists in the output. Also, looking for the absence of a very specific string in serialized JSON output is very brittle.

Tests/SourceKitLSPTests/DoccDocumentationTests.swift
```diff
@@ -66,6 +66,72 @@ final class DoccDocumentationTests: SourceKitLSPTestCase {
...
+      uri: uri
+    )
+    let response = try await testClient.send(
+      DoccDocumentationRequest(
+        textDocument: TextDocumentIdentifier(uri),
+        position: positions["1️⃣"]
+      )
+    )
+    let renderNodeString = response.renderNode
+    XCTAssertTrue(
```
matthewbastien, line 91: Use `renderDocumentation()` like the other tests in this suite. It has an expectation called `contains` that does exactly this.

matthewbastien (changes requested):
The test cases you added don't actually test for the expected behaviour: I tried running the tests locally and they pass regardless of whether or not I have applied your fix.

The string `"0: field does not exist"` will always be present in the render node. The important part to check is **where** it is located. The tests should be updated to actually check that the parameters are in the correct section.

vanshkamra12:
Thanks for the review! I looked into using renderDocumentation() with containing:, but the challenge is that "0: field does not exist" appears in the render node JSON in both cases — with the bug it's in the Discussion section, and with the fix it's in the Return Value section. So a simple containing check passes regardless of whether the fix is applied.

To properly catch the regression, I've updated the tests to parse the render node JSON using JSONSerialization, locate the content section identified by anchor: "return-value" (the Return Value section), and assert the continuation lines appear specifically inside that section — not in a separate Discussion section. This way the tests fail without the fix and pass with it.

Does this approach sound reasonable? If so, I'll push the updated tests.

vanshkamra12:
@matthewbastien can you please tell what should i do next ? I mentioned the issue which i am facing..

matthewbastien:
You should push up your new change so I can review it.

PR #2496, opened 2026-02-24
Add type annotation to convertStoredPropertyToComputed

Padmashree06:
**Issue**: https://github.com/swiftlang/sourcekit-lsp/issues/2459

This PR implements the `ConvertStoredPropertyToComputed` refactoring in SourceKit-LSP, adopting the new API recently merged in swift-syntax ([#3249](https://github.com/swiftlang/swift-syntax/pull/3249)).

## Summary

It integrates `cursorInfo` into the `SyntaxCodeActionScope`, allowing the refactoring to resolve types for inferred properties (e.g., `let x = 5`) by extracting the semantic type from SourceKit's annotated declaration.
### Changes

- Updated `SyntaxCodeActionScope` to include `[CursorInfo]`
- Updated the code action request handler to:
  - Perform a `cursorInfo` request
  - Add the result to `SyntaxCodeActionScope`
- Added `ConvertStoredPropertyToComputed` to `allSyntaxCodeActionProviders` in `SyntaxCodeActions.swift`

### Implementation Details

Implemented `SyntaxCodeActionProvider` for `ConvertStoredPropertyToComputed` which:

1. Attempts to resolve the property type via `cursorInfo`
2. Falls back to explicit type annotations if semantic info is unavailable
3. Passes the resolved type to the refactoring context


Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
```
ahoppen, line 13: Could you add a copyright header to this file?

ahoppen, line 13: Mostly out of curiosity: Why do we need to import Foundation here?

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,60 @@
...
+      CodeAction(
+        title: "Convert to computed property",
+        kind: .refactorInline,
+        edit: WorkspaceEdit(changes: [scope.snapshot.uri: [edit]])
+      )
+    ]
+  }
+
+  private static func extractType(from annotatedDecl: String) -> String? {
+
```
ahoppen: Superfluous newline

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+extension ConvertStoredPropertyToComputed: SyntaxCodeActionProvider {
+
```
ahoppen: Superfluous newline

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -44,14 +44,18 @@ struct SyntaxCodeActionScope {
...
     snapshot: DocumentSnapshot,
     syntaxTree file: SourceFileSyntax,
-    request: CodeActionRequest
+    request: CodeActionRequest,
+    cursorInfo: [CursorInfo] = []
   ) {
     self.snapshot = snapshot
     self.request = request
     self.file = file
+    self.cursorInfo = cursorInfo
```
ahoppen, line 58: I don’t think we set `cursorInfo` anywhere, do we? Also, we should implement this in a way where the cursor info is only retrieved when it’s actually needed by a refactoring action instead of being set for every syntactic refactoring action.

ahoppen, line 58: Could you add a test that tests the code action with and  without a type annotation?

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,60 @@
...
+        title: "Convert to computed property",
+        kind: .refactorInline,
+        edit: WorkspaceEdit(changes: [scope.snapshot.uri: [edit]])
+      )
+    ]
+  }
+
+  private static func extractType(from annotatedDecl: String) -> String? {
+
+    guard let start = annotatedDecl.range(of: "<type>"),
```
ahoppen: Instead of parsing the annotated description, we should use the `typename` key.

ahoppen (commented):
A few preliminary review comments before I dive deeper into the implementation.

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
Padmashree06: Yes sure!

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
Padmashree06: Thank you for pointing that out!

You’re right, I missed that `cursorInfo` is not being initialized anywhere. Triggering a `cursorInfo` request for every syntactic refactoring would indeed be unnecessary overhead.

I’ll remove `cursorInfo` from `SyntaxCodeActionScope`. Instead, I’ll first perform a syntactic check to determine whether the type information is missing. Only if semantic resolution is actually required for the refactoring will I trigger a `cursorInfo` request within the refactoring operation itself, using `SwiftLanguageService.cursorInfo`.

Please let me know if this approach aligns with your expectations or if there’s anything else I should consider.

Padmashree06: Yes sure!

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
Padmashree06: Yes thank you! I'll  use `typename` key instead!

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
Padmashree06: I imported `Foundation` because I was using `String.range(of:)` in the `extractType` helper. However, if we'd prefer to keep the dependencies minimal, I'm happy to switch to native Swift String methods like `firstRange(of:)` and remove the import entirely. Let me know what you'd prefer!

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
Padmashree06: Thank you for the feedback! I'll remove the unnecessary lines.

ahoppen:
If you have addressed my comments, it seems like you haven’t pushed your latest changes.

Padmashree06:
I haven’t pushed the changes yet. I was waiting for confirmation before doing so. I’ll push them shortly.

Padmashree06:
@ahoppen I’ve updated the implementation based on your feedback.

- Cursor info is now fetched only when the type is not explicitly declared.
- When a type annotation is present, the refactoring is applied inline.
- Otherwise, the `semantic.refactor.convertStoredPropertyToComputed` command is executed to retrieve cursor info, infer the type semantically, and perform the conversion to a computed property.
- I’ve added corresponding CodeAction tests for both cases.

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,64 @@
...
+      ]
+    }
+
+    return [
+      CodeAction(
+        title: "Convert Stored Property to Computed Property",
+        kind: .refactorInline,
+        command: Command(
+          title: "Convert Stored Property to Computed Property",
+          command: "semantic.refactor.convertStoredPropertyToComputed",
```
ahoppen: Instead of implementing a custom command, do you think you could use the support for `codeAction/resolve` that @Tabonx added recently in https://github.com/swiftlang/sourcekit-lsp/pull/2549?

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
ahoppen: You’r no longer using that anymore though, right?

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
ahoppen: Yes, that’s a lot better now with the asynchronous command and it’s the approach I would have preferred until a few weeks ago but now we have `codeAction/resolve`, which is designed for exactly this workflow. Could you adopt it?

Padmashree06:
Yes, I will modify the implementation to use  `codeAction/resolve`

Padmashree06:
@ahoppen  I've updated the implementation to use `codeAction/resolve`

ahoppen:
Sorry for the delayed response, I was on vacation for the last couple of weeks.

I just took a deeper dive into the implementation and was a little uneasy with how ad-hoc the implementation was for this particular refactoring action because I think we need similar support for more refactoring actions in the future. I would have really liked to make some suggestion for a better design but needed to really dive into it myself to explore alternatives and once I arrived at one, I pretty much had a working implementation ready, so I ended up opening https://github.com/swiftlang/sourcekit-lsp/pull/2622. I would appreciate your feedback on that PR, let’s continue the conversation over there. Thanks for the work you did here to get this started.

Padmashree06:
@ahoppen Thanks for the review and for taking a deeper look into this. I understand your concerns about the implementation being too specific. I’ll go through #2622 and add my inputs there!

PR #2497, opened 2026-02-24
Emit proper semantic token kinds for declaration sites

evan-templeton:
## Issue
SourceKit's `source.request.semantic_tokens` API returns reference tokens only (`refVarLocal`, `refStruct`, etc.) and never declaration tokens, so declaration names were all falling back to the `.identifier` classification. This meant `let asdf: Int = 0` showed `asdf` as a plain `identifier` instead of a variable with a declaration modifier, making it impossible for editors to recognize declarations through semantic token modifiers.

## Changes
Introduce `DeclarationHighlightingVisitor,` a `SyntaxVisitor` that walks the AST and emits properly-typed tokens (`.variable`, `.property`, `.struct`, `.class`, `.function`, `.method`, `.enum`, `.enumMember`, etc.) with the `.declaration` modifier for every named declaration site. These are merged into `semanticHighlightingTokens` alongside SourceKit reference tokens so that SourceKit still takes precedence for usage/reference sites.

Also fix `mergedAndSortedTokens` to filter the merged result to the requested range for range-based requests, preventing whole-file semantic tokens from leaking into partial-range results.

### DIFF
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -19,8 +19,9 @@ import SwiftParser
 import SwiftSyntax
 
 extension SwiftLanguageService {
-  /// Requests the semantic highlighting tokens for the given snapshot from sourcekitd.
-  private func semanticHighlightingTokens(for snapshot: DocumentSnapshot) async throws -> SyntaxHighlightingTokens? {
+  /// Returns semantic reference tokens from SourceKit for the given snapshot, or `nil` if
+  /// no real compile command is available or the request fails.
+  private func sourceKitReferenceTokens(for snapshot: DocumentSnapshot) async throws -> SyntaxHighlightingTokens? {
     guard let compileCommand = await self.compileCommand(for: snapshot.uri, fallbackAfterTimeout: false),
       !compileCommand.isFallback
     else {
@@ -58,30 +59,43 @@ extension SwiftLanguageService {
   ) async throws -> SyntaxHighlightingTokens {
     try Task.checkCancellation()
 
-    async let tree = syntaxTreeManager.syntaxTree(for: snapshot)
-    let semanticTokens = await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+    let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
 
-    let range =
+    let byteRange =
       if let range {
         snapshot.byteSourceRange(of: range)
       } else {
-        await tree.range
+        tree.range
       }
 
     try Task.checkCancellation()
 
-    let tokens =
-      await tree
-      .classifications(in: range)
+    // Syntactic classification tokens (keywords, literals, comments, etc.)
+    let syntaxTokens =
+      tree
+      .classifications(in: byteRange)
       .map { $0.highlightingTokens(in: snapshot) }
       .reduce(into: SyntaxHighlightingTokens(tokens: [])) { $0.tokens += $1.tokens }
 
+    // Declaration name tokens derived from the syntax tree
+    let declarationVisitor = DeclarationHighlightingVisitor(snapshot: snapshot)
+    declarationVisitor.walk(tree)
+    let declarationTokens = SyntaxHighlightingTokens(tokens: declarationVisitor.tokens)
+    
     try Task.checkCancellation()
 
-    return
-      tokens
-      .mergingTokens(with: semanticTokens ?? SyntaxHighlightingTokens(tokens: []))
+    let skTokens = await orLog("Loading SourceKit reference tokens") {
+      try await sourceKitReferenceTokens(for: snapshot)
+    }
+
+    let merged =
+      syntaxTokens
+      .mergingTokens(with: declarationTokens)
+      .mergingTokens(with: skTokens ?? SyntaxHighlightingTokens(tokens: []))
       .sorted { $0.start < $1.start }
+
+    guard let range else { return merged }
+    return SyntaxHighlightingTokens(tokens: merged.tokens.filter { range.overlaps($0.range) })
   }
 
   package func documentSemanticTokens(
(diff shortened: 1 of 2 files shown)


Steffeeen:
The change looks like an improvement to me. It also seems easy to integrate with what I'm doing in #2485. In VSCode this causes a few changes in the highlighting. I have generated a diff here: https://proper-semantic-token-declarations.netlify.app/. I like that variable/property declarations are now in the same blue as variable/property references. `associatedtype` type names turning green is also nice, as the changes in #2485 turn the `associatedtype` keyword into the same blue which the type name used previously.

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -36,12 +48,15 @@ extension SwiftLanguageService {
     let dict = try await send(sourcekitdRequest: \.semanticTokens, skreq, snapshot: snapshot)
```
hamishknight, line 37: If this throws an error we'll lose the syntactic decl tokens, should we add the syntactic tokens in `mergedAndSortedTokens` instead?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -72,16 +87,19 @@ extension SwiftLanguageService {
...
     try Task.checkCancellation()
 
-    return
+    let merged =
       tokens
       .mergingTokens(with: semanticTokens ?? SyntaxHighlightingTokens(tokens: []))
       .sorted { $0.start < $1.start }
+
+    guard let range else { return merged }
+    return SyntaxHighlightingTokens(tokens: merged.tokens.filter { range.overlaps($0.range) })
```
hamishknight, line 98: Is there a test case we could add for this?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+      emit(node.name, kind: .interface, modifiers: [.declaration])
+      scopeStack.append(.typeBody)
+      return .visitChildren
+    }
+    override func visitPost(_ node: ProtocolDeclSyntax) { scopeStack.removeLast() }
+
+    // MARK: Type aliases and associated types
+
+    override func visit(_ node: TypeAliasDeclSyntax) -> SyntaxVisitorContinueKind {
+      emit(node.name, kind: .typeParameter, modifiers: [.declaration])
```
hamishknight, line 254: `typeParameter` for a `typealias` seems odd to me, although it does match what we currently do in `parseKindAndModifiers`. Seems like clangd picks `class` for a `typedef`, wondering if we should just pick `struct` or something instead. Any thoughts @ahoppen?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+  /// (e.g. `.variable`, `.property`, `.struct`) with the `.declaration` modifier for every named
+  /// declaration so that editors can highlight them distinctly from plain identifiers.
+  private final class DeclarationHighlightingVisitor: SyntaxVisitor {
+    var tokens: [SyntaxHighlightingToken] = []
+    private let snapshot: DocumentSnapshot
+
+    /// Tracks the innermost scope kind so we can distinguish instance properties (`.property`)
+    /// from local or global variables (`.variable`).
+    private enum ScopeKind {
+      case typeBody  // inside a struct / class / enum / actor / protocol member block
```
hamishknight: Extensions should also be considered type scopes

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+      case codeBody  // inside a function, initializer, or closure
+    }
+    private var scopeStack: [ScopeKind] = []
```
hamishknight: Rather than maintaining this state, could we just look at the syntax node parents instead? Then you can just check for a `DeclGroupSyntax` parent

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+    override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
+      scopeStack.append(.codeBody)
+      return .visitChildren
+    }
+    override func visitPost(_ node: ClosureExprSyntax) { scopeStack.removeLast() }
+
+    // MARK: Variable and property declarations
+
+    override func visit(_ node: PatternBindingSyntax) -> SyntaxVisitorContinueKind {
+      guard let idPattern = node.pattern.as(IdentifierPatternSyntax.self) else {
```
hamishknight: How about things like `let (a, b) = `?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+      }
+      var modifiers: SemanticTokenModifiers = [.declaration]
+      if let varDecl = node.parent?.parent?.as(VariableDeclSyntax.self),
+        varDecl.modifiers.contains(where: {
+          $0.name.tokenKind == .keyword(.static) || $0.name.tokenKind == .keyword(.class)
+        })
+      {
+        modifiers.insert(.static)
+      }
+      emit(idPattern.identifier, kind: isInTypeMemberScope ? .property : .variable, modifiers: modifiers)
```
hamishknight: Do we want to represent `let`s as `.constant`?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+      return .visitChildren
+    }
+
+    // MARK: Optional binding conditions (if let x = ..., guard let x = ...)
```
hamishknight: What about variables bound in `case` statements e.g `case let x`?

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -906,17 +906,61 @@ final class SemanticTokensTests: SourceKitLSPTestCase {
...
       ]
     )
   }
 
+  func testOptionalBindingDeclarations() async throws {
+    // if let x = expr: the bound variable should be a declaration, not a plain identifier
+    try await assertSemanticTokens(
+      markedContents: """
+        1️⃣let 2️⃣x: 3️⃣Int? = 4️⃣nil
+        5️⃣if 6️⃣let 7️⃣y = 8️⃣x {}
```
hamishknight, line 1066: Can you also add a test case for shorthand `if let`? e.g `if let x {}`

Tests/SourceKitLSPTests/SemanticTokensTests.swift
```diff
@@ -906,17 +906,61 @@ final class SemanticTokensTests: SourceKitLSPTestCase {
...
+        TokenSpec(marker: "7️⃣", length: 1, kind: .variable, modifiers: .declaration),
+        TokenSpec(marker: "8️⃣", length: 1, kind: .variable),
+      ]
+    )
+
+    // guard let x = expr: the bound variable should be a declaration, not a plain identifier
+    try await assertSemanticTokens(
+      markedContents: """
+        1️⃣let 2️⃣x: 3️⃣Int? = 4️⃣nil
+        5️⃣guard 6️⃣let 7️⃣y = 8️⃣x 9️⃣else { 🔟return }
```
hamishknight, line 1086: Ditto for `guard`

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+
+    override func visit(_ node: EnumCaseElementSyntax) -> SyntaxVisitorContinueKind {
+      emit(node.name, kind: .enumMember, modifiers: [.declaration])
+      return .visitChildren
+    }
+
+    // MARK: Macro declarations
+
+    override func visit(_ node: MacroDeclSyntax) -> SyntaxVisitorContinueKind {
+      emit(node.name, kind: .macro, modifiers: [.declaration])
```
hamishknight, line 322: Can you add a test for this?

Sources/SwiftLanguageService/SemanticTokens.swift
hamishknight: Also can you add a test case for this?

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -169,3 +187,159 @@ extension SyntaxClassification {
...
+
+    // MARK: Type aliases and associated types
+
+    override func visit(_ node: TypeAliasDeclSyntax) -> SyntaxVisitorContinueKind {
+      emit(node.name, kind: .typeParameter, modifiers: [.declaration])
+      return .visitChildren
+    }
+
+    override func visit(_ node: AssociatedTypeDeclSyntax) -> SyntaxVisitorContinueKind {
+      emit(node.name, kind: .typeParameter, modifiers: [.declaration])
```
hamishknight, line 259: Can you add a test case for this?

evan-templeton: added

Sources/SwiftLanguageService/SemanticTokens.swift
evan-templeton: Will add tuple support (and test) in the next revision

Sources/SwiftLanguageService/SemanticTokens.swift
evan-templeton: 👍

Sources/SwiftLanguageService/SemanticTokens.swift
ahoppen: I think we picked `typeParameter` here because `typealias` satisfying `associatedtype` requirements could be considered parameters. Though I don’t support that mapping anymore. How about we just use the generic `type` semantic token type. The LSP spec specifies that as
*Represents a generic type. Acts as a fallback for types which can't be mapped to a specific type like class or enum.*, which matches this pretty accurately. Also, let’s change `parseKindAndModifiers` while we’re at it.

hamishknight: Yeah that sounds good to me 👍

ahoppen:
@evan-templeton I just merged https://github.com/swiftlang/sourcekit-lsp/pull/2497, which generated merge conflict. Would you be able to resolve them?

PR #2498, opened 2026-02-25
new feature to move methods and fields between types

amanthatdoescares:
## Add Move Member Refactoring Code Action (`refactor.move`)

This PR introduces a new **Move Member** Code Action to Swift LSP that allows users to move a member declaration (e.g. property, method, enum case) from one type to another.

The action is surfaced under the **Refactor…** menu when the cursor is placed inside a member of a nominal type (`struct`, `class`, or `enum`). It is not offered for top-level declarations.

The implementation adds a new provider in:

Sources/SwiftLanguageService/CodeActions/MoveMember.swift

which detects `MemberBlockItemSyntax` at the cursor position and returns a command-based Code Action (`swift.moveMember`) of kind `"refactor.move"`.

### Tests Added

Added test coverage in:

Tests/SourceKitLSPTests/CodeActionTests.swift

- `testMoveMemberCodeAction`
- `testMoveMemberCodeActionNotOfferedForTopLevelCode`

feature: https://github.com/swiftlang/sourcekit-lsp/issues/2384

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/MoveMember.swift
+++ b/Sources/SwiftLanguageService/CodeActions/MoveMember.swift
@@ -1,0 +1,52 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftSyntax
+
+extension CodeActionKind {
+  static let refactorMove = CodeActionKind(rawValue: "refactor.move")
+}
+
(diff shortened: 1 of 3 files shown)


amanmaurya92:
> What client implements the `swift.moveMember` command? How do you choose what type to move the method/field to?

Thanks for raising this. At the moment, the PR only adds the server-side code action and command, and the client interaction needed to select the destination type is not implemented yet.

I’ll complete the expected LSP flow by adding support to query valid destination types, allowing the client to prompt the user, and then execute `swift.moveMember` with the selected target. I’ll put this up for review again once the full interaction is implemented end-to-end.

plemarquand:
What does the list of destination types look like? Is it only ones in the same file? same module? entire project? How do you avoid overwhelming the user with potentially 100s of options? What is the UX in the client for picking a destination type?

amanmaurya92:
> What does the list of destination types look like? Is it only ones in the same file? same module? entire project? How do you avoid overwhelming the user with potentially 100s of options? What is the UX in the client for picking a destination type?

That makes sense. I’ll take this into account while revisiting the design and will avoid a broad project-wide destination list.

Before finalizing the approach, I plan to start a discussion on the Swift forums outlining the proposed behavior (limiting destinations to structurally valid containers in the immediate lexical hierarchy) to gather feedback and ensure alignment on the intended UX and scope.

I’ll proceed with implementation once there’s clarity and consensus around the direction.

PR #2500, opened 2026-02-26
[TestDiscovery] Prefer syntactic scan results over semantic index

rintaro:
(Separated from #2489)

Previously, test discovery used the semantic index as the primary source and fell back to the syntactic index only for files where the semantic index was out-of-date. This meant test locations came from the semantic index, which only records a point position rather than the full symbol range. Also tests in inactive regions were not emitted at all.

Flip the priority: use syntactic scan results as the primary source (which have correct location ranges) and supplement with semantic index results. The semantic results are range-fixed via `textDocument/documentSymbol` before being returned.

Introduce unified `combineTests()` helper shared by both `workspaceTests` and `documentTests`.

Also fix `SyntacticSwiftXCTestScanner` to emit extensions as proper `AnnotatedTestItem` nodes (with `isExtension: true`) rather than a flat list of methods, so extension test methods are correctly merged into their class via `mergingTestsInExtensions`. The class and extension visitors are unified through a shared `handleClassOrExtension` helper.

### DIFF
--- a/Sources/SourceKitLSP/TestDiscovery.swift
+++ b/Sources/SourceKitLSP/TestDiscovery.swift
@@ -185,51 +185,6 @@ struct TestDiscovery {
       .map { testItem(for: $0, context: []) }
   }
 
-  /// Fix the 'location.range' of test cases from the semantic index using 'textDocument/symbol' results.
-  private func fixSemanticTestRanges(
-    tests: [AnnotatedTestItem],
-    workspace: Workspace
-  ) async -> [AnnotatedTestItem] {
-
-    // Cached 'textDocument/symbol' result per document.
-    var _documentSymbols: [DocumentURI: DocumentSymbolResponse?] = [:]
-    func documentSymbols(in uri: DocumentURI) async throws -> DocumentSymbolResponse? {
-      if let cached = _documentSymbols[uri] {
-        return cached
-      }
-      guard let languageService = workspace.primaryLanguageService(for: uri) else {
-        return nil
-      }
-      let symbols = try await languageService.documentSymbol(
-        DocumentSymbolRequest(textDocument: TextDocumentIdentifier(uri))
-      )
-      _documentSymbols[uri] = symbols
-      return symbols
-    }
-
-    // Recursively fix up 'TestItem.location.range'.
-    func fixupLocation(item: inout TestItem, using documentSymbols: DocumentSymbolResponse) {
-      let fixedRange = findInnermostSymbolRange(
-        containing: item.location.range.lowerBound,
-        documentSymbolsResponse: documentSymbols
-      )
-      if let fixedRange {
-        item.location.range = fixedRange
-      }
-      for idx in item.children.indices {
-        fixupLocation(item: &item.children[idx], using: documentSymbols)
-      }
-    }
-
-    return await tests.asyncMap { (item) async -> AnnotatedTestItem in
-      var item = item
-      if let symbols = try? await documentSymbols(in: item.testItem.location.uri) {
-        fixupLocation(item: &item.testItem, using: symbols)
-      }
-      return item
-    }
-  }
-
   /// Combine 'syntacticTests' and 'semanticTests'.
   ///
   ///  * Use 'syntacticTests' primarily
@@ -238,8 +193,7 @@ struct TestDiscovery {
   private func combineTests(
     syntacticTests: [AnnotatedTestItem],
     semanticTests: [AnnotatedTestItem]?,
-    index: CheckedIndex?,
-    workspace: Workspace
+    index: CheckedIndex?
   ) async throws -> [AnnotatedTestItem] {
     guard let semanticTests else {
       return syntacticTests
@@ -265,13 +219,7 @@ struct TestDiscovery {
       )
     }
 
-    // 'semanticTestsMap.values' now contains the results to include. Fix-up the range info.
-    let semanticTestsFixed = await fixSemanticTestRanges(
-      tests: semanticTestsMap.values.flatMap(\.self),
-      workspace: workspace
-    )
-
-    return syntacticTests + semanticTestsFixed
+    return syntacticTests + semanticTestsMap.values.flatMap(\.self)
   }
 
   /// Return all the tests in the given workspace.
@@ -357,8 +305,7 @@ struct TestDiscovery {
     return try await combineTests(
       syntacticTests: syntacticTests,
       semanticTests: semanticTests,
-      index: index,
-      workspace: workspace
+      index: index
     )
   }
 
@@ -416,15 +363,45 @@ struct TestDiscovery {
       syntacticTests == nil ? .deletedFiles : .inMemoryModifiedFiles(documentManager)
     let index = await workspace.index(checkedFor: indexCheckLevel)
 
-    return try await combineTests(
+    let tests = try await combineTests(
       syntacticTests: syntacticTests ?? [],
       semanticTests: try testItems(
-        for: index?.unitTests(referencedByMainFiles: [mainFileUri.pseudoPath]) ?? [],
+        for: index?.unitTests(referencedByMainFiles: [mainFileUri.pseudoPath]).filter(\.canBeTestDefinition) ?? [],
         index: index
       ),
-      index: await workspace.index(checkedFor: .deletedFiles),
-      workspace: workspace
+      index: await workspace.index(checkedFor: .deletedFiles)
     )
+
+    // Recursively fix up 'TestItem.location.range' from semantic index.
+    func fixupLocation(item: inout TestItem, using documentSymbols: DocumentSymbolResponse) {
+      if item.location.range.isEmpty {
+        let fixedRange = findInnermostSymbolRange(
+          containing: item.location.range.lowerBound,
+          documentSymbolsResponse: documentSymbols
+        )
+        if let fixedRange {
+          item.location.range = fixedRange
+        }
+      }
+      for idx in item.children.indices {
+        fixupLocation(item: &item.children[idx], using: documentSymbols)
+      }
+    }
+
+    let symbols = await orLog("Getting 'textDocument/symbols' for \(uri)") {
+      try await languageService.documentSymbol(
+        DocumentSymbolRequest(textDocument: TextDocumentIdentifier(uri))
+      )
+    }
+    if let symbols {
+      return tests.map { item in
+        var item = item
+        fixupLocation(item: &item.testItem, using: symbols)
+        return item
+      }
+    }
+
+    return tests
   }
 }
 
--- a/Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift
+++ b/Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift
@@ -992,10 +992,10 @@ final class WorkspaceTestDiscoveryTests: SourceKitLSPTestCase {
         @interface MyTests : XCTestCase
         @end
 
-        1️⃣@implementation MyTests
-        2️⃣- (void)testSomething {}3️⃣
+        @implementation 1️⃣MyTests
+        - (void)2️⃣testSomething {}
         0️⃣
-        @4️⃣end
+        @end
         """
       ],
       manifest: """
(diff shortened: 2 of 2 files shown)


rintaro:
@swift-ci Please test

Tests/SourceKitLSPTests/DocumentTestDiscoveryTests.swift
```diff
@@ -1925,4 +1925,77 @@ final class DocumentTestDiscoveryTests: SourceKitLSPTestCase {
     let tests = try await project.testClient.send(DocumentTestsRequest(textDocument: TextDocumentIdentifier(uri)))
     XCTAssertEqual(tests, [])
   }
+
+  func testXCTestsInInactiveIfRegionsAreDiscovered() async throws {
```
rintaro, line 1929: I stole this test case from #2494 @loveucifer 🙇‍♂️

Sources/SwiftLanguageService/SyntacticSwiftXCTestScanner.swift
```diff
@@ -81,46 +81,59 @@ final class SyntacticSwiftXCTestScanner: SyntaxVisitor {
...
-    let superclassName = superclass.type.as(IdentifierTypeSyntax.self)?.name.text
-    if superclassName == "NSObject" {
-      // We know that the class can't be an subclass of `XCTestCase` so don't visit it.
-      // We can't explicitly check for the `XCTestCase` superclass because the class might inherit from a class that in
-      // turn inherits from `XCTestCase`. Resolving that inheritance hierarchy would be semantic.
-      return .visitChildren
-    }
-    let testMethods = findTestMethods(in: node.memberBlock.members, containerName: node.name.text)
-    guard !testMethods.isEmpty || superclassName == "XCTestCase" else {
+  func handleClassOrExtenstion(
```
rintaro:
```suggestion
  func handleClassOrExtension(
```

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -199,112 +288,71 @@ extension SourceKitLSPServer {
...
+    )
   }
 
-  func workspaceTests(_ req: WorkspaceTestsRequest) async throws -> [TestItem] {
-    return await self.workspaces
-      .concurrentMap { await self.tests(in: $0).prefixTestsWithModuleName(workspace: $0) }
-      .flatMap { $0 }
-      .sorted { $0.testItem.location < $1.testItem.location }
+  /// Collect all test cases from all the workspaces and merge it into a sorted list of the test cases.
+  func workspaceTests() async -> [TestItem]? {
```
rintaro:
```suggestion
  func workspaceTests() async -> [TestItem] {
```

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -182,10 +182,99 @@ extension SourceKitLSPServer {
...
+    // Cached 'textDocument/symbol' result per document.
+    var _documentSymbols: [DocumentURI: DocumentSymbolResponse?] = [:]
+    func documentSymbols(in uri: DocumentURI) async throws -> DocumentSymbolResponse? {
+      if let cached = _documentSymbols[uri] {
+        return cached
+      }
+      guard let languageService = workspace.primaryLanguageService(for: uri) else {
+        return nil
+      }
+      let symbols = try await languageService.documentSymbol(
```
ahoppen: Do we know that this document has been opened by the client? Ie. when retrieving workspace tests, could it be that we need to fix up test ranges for documents that aren’t open and thus `documentSymbols` will fail? Also, I wouldn’t want to get document symbols for all documents in the workspace.

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -182,10 +182,99 @@ extension SourceKitLSPServer {
...
+    syntacticTests: [AnnotatedTestItem],
+    semanticTests: [AnnotatedTestItem]?,
+    maybeOutdatedIndex: CheckedIndex?,
+    workspace: Workspace
+  ) async -> [AnnotatedTestItem] {
+    guard let semanticTests else {
+      return syntacticTests
+    }
+
+    var semanticTestsMap = [String: [AnnotatedTestItem]](grouping: semanticTests, by: { $0.testItem.id })
```
ahoppen:
```suggestion
    var semanticTestsMap = [String: [AnnotatedTestItem]](grouping: semanticTests, by: \.testItem.id)
```

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -182,10 +182,99 @@ extension SourceKitLSPServer {
...
+      // a single XCTestCase subclass, so caching doesn't make sense here.
+      // Also, this is only called for files containing test cases but for which the semantic index is out-of-date.
+      return item.filterUsing(
+        semanticSymbols: maybeOutdatedIndex?.symbols(inFilePath: item.testItem.location.uri.pseudoPath)
+      )
+    }
+
+    // 'semanticTestsMap.values' now contains the results to include. Fix-up the range info.
+    let semanticTestsFixed = await fixSemanticTestRanges(
+      tests: semanticTestsMap.values.flatMap({ $0 }),
```
ahoppen:
```suggestion
      tests: semanticTestsMap.values.flatMap(\.self),
```

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -199,179 +288,119 @@ extension SourceKitLSPServer {
...
-    //    index.
-    //  - All files that have in-memory modifications are syntactically scanned for tests here.
-    let index = await workspace.index(checkedFor: .inMemoryModifiedFiles(documentManager))
+    let documentManager = sourceKitLSPServer.documentManager
 
-    // TODO: Remove this workaround once https://github.com/swiftlang/swift/issues/75600 is fixed
-    func documentManagerHasInMemoryModifications(_ uri: DocumentURI) -> Bool {
-      return documentManager.fileHasInMemoryModifications(uri)
-    }
+    var testFiles = (try? await workspace.buildServerManager.projectTestFiles()) ?? Array(documentManager.openDocuments)
```
ahoppen: Falling back to `documentManager.openDocuments` if the build server doesn’t provide any test files seems like an odd choice to me because now the `workspace/tests` request is dependent on which files you currently have open in your editor. Why is it needed?

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -199,179 +288,119 @@ extension SourceKitLSPServer {
...
-      } ?? []
-    }.flatMap { $0 }
-
-    let semanticTestSymbolOccurrences = index?.unitTests().filter { return $0.canBeTestDefinition } ?? []
-
-    let testsFromSyntacticIndex = await workspace.syntacticIndex.tests()
+    // Collect tests from semantic index.
+    var symbolOccurrences: [SymbolOccurrence] = []
+    if let index, let maybeOutdatedIndex {
+      // FIXME: Instead of querying two times, get all the symbolOccurences and filter them here.
```
ahoppen: Could you either fix this or file an issue for it + reference it? I find that FIXME comments in the code base tend to just get lost.

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -199,179 +288,119 @@ extension SourceKitLSPServer {
...
+      symbolOccurrences += index.unitTests()
+      symbolOccurrences += maybeOutdatedIndex.unitTests(
+        referencedByMainFiles: filesToUseSemanticIndex.map(\.pseudoPath)
+      )
```
ahoppen: Don’t we add symbol occurrences twice for files that are currently open but don’t have any modifications?

Sources/SourceKitLSP/TestDiscovery.swift
```diff
@@ -199,179 +288,119 @@ extension SourceKitLSPServer {
...
-      .concurrentMap { await self.tests(in: $0).prefixTestsWithModuleName(workspace: $0) }
-      .flatMap { $0 }
-      .sorted { $0.testItem.location < $1.testItem.location }
+  /// Collect all test cases from all the workspaces and merge it into a sorted list of the test cases.
+  func workspaceTests() async -> [TestItem] {
+    return await self.sourceKitLSPServer.workspaces
+      .concurrentMap {
+        await self.tests(in: $0).prefixTestsWithModuleName(workspace: $0)
+      }
+      .flatMap({ $0 })
```
ahoppen:
```suggestion
      .flatMap { $0 }
```

or 

```suggestion
      .flatMap(\.self)
```

Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift
```diff
@@ -916,6 +902,7 @@ final class WorkspaceTestDiscoveryTests: SourceKitLSPTestCase {
       class NotQuiteTest: SomeClass {
         func testMyLibrary() {}
       }
+
```
ahoppen, line 905: Unrelated change?

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: Non-open documents don't have the `languageService` registered in the `Workspace`, so no `DocumentSymbols` requests are sent to any, and it's resolved to `nil`. I admit it sounds brittle.

Tbh, I'm not a fan of this feature (fix up ranges) and I'm all for removing this entirely. I.e. all `XCTest` tests in ObjC and non-direct subclass of `XCTestCase` in Swift would only have index-locations, consistently regardless of the open document situation.

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: Good point, I think I should fix the FIXME above now after all.

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: For workspaces without a build server. I.e. when opening a random file in the editor.
see `testInMemoryFileWithFallbackBuildServer` test case https://github.com/swiftlang/sourcekit-lsp/blob/ce86aae5042cdafbadc02c0070adef8d5c70ea2a/Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift#L704

I personally don't think we need to return anything in such cases, but I'm just keeping the existing behavior.

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: Decided to do the range fix-up only in `textDocument/tests` request for now.

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: I added a comment in the code.

Sources/SourceKitLSP/TestDiscovery.swift
rintaro: Now `index` itself uses `.deletedFiles` checking, and the filtering is done using the each file status.

Tests/SourceKitLSPTests/WorkspaceTestDiscoveryTests.swift
rintaro: More like drive-by change. Since `"""` doesn't add newline at the end of the literal, the updated content by `(originalContents + addedTest)` was like:
```swift
class NotQuiteTest: SomeClass {
  func testMyLibrary() {}
}class OtherNotQuiteTest: SomeClass {
  func testSomethingElse() {}
}
```
I thought having a newline at the EOL was more realistic scenario.

Sources/SourceKitLSP/SyntacticIndex.swift
```diff
@@ -285,12 +285,19 @@ package actor SyntacticIndex: Sendable {
...
-  /// Gets all the tests in the syntactic index.
+  /// Gets the syntactically indexed tests for the given files.
+  /// Returns tuple of the list of the tests found and the document URIs that wasn't in the index.
   ///
   /// This waits for any pending document updates to be indexed before returning a result.
-  nonisolated package func tests() async -> [AnnotatedTestItem] {
-    let readTask = indexingQueue.async(metadata: .read) {
-      return await self.indexedSources.values.flatMap(\.tests)
+  ///
+  nonisolated package func tests(in files: [DocumentURI]) async -> [DocumentURI: [AnnotatedTestItem]] {
```
rintaro: Changed this to return a dictionary so we can know which file doesn't support syntactic scan (i.e. written in C family languages).

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Issue #2004, opened 2025-02-25
Pass experimental features to SwiftParser once we have build settings


ahoppen:
When we have build settings for a file, we should be able to use those when parsing the SwiftSyntax tree. Currently, we always create the SwiftSyntax tree with all experimental features disabled.

loveucifer:
heyy @ahoppen, I took a stab at this.i think the approach is to extract -enable-experimental-feature flags from the compiler arguments in SwiftCompileCommand, store them per-document in SyntaxTreeManager, ( is this right ? ) and use the @_spi(ExperimentalLanguageFeatures) parser initializer to pass them through , and thenfeatures get set whenever build settings arrive (in openDocument nd reopenDocument), and cached trees are invalidated if the features change.
also is storing the features as state in SyntaxTreeManager the right callor does it makes more sense to have it threaded explicitly through each syntaxTree(for:) call?

ahoppen:
Conceptually, the tricky part to get this right is to make sure that we generate a syntax tree without any experimental features set while we don’t have build settings and only incorporate the enabled experimental features once we get build settings. The approach I would start with is storing this in the `SyntaxTreeManager` (what you suggested) and transition from a base syntax tree (without experimental features enabled) to one that takes experimental features into account when we get build settings, similar to how we handle document changes.

PR #2501, opened 2026-02-26
Pass experimental features to SwiftParser once we have build settings

loveucifer:
addressing #2004

this pr should be merged only after https://github.com/swiftlang/swift-syntax/pull/3286 this

### DIFF
--- a/Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
@@ -826,6 +826,11 @@ final class SwiftPMIntegrationTests: SourceKitLSPTestCase {
       newMarkedContents: newManifest
     )
 
+    // Wait for the build server to update the build graph after Package.swift changes.
+    // SynchronizeRequest(index: true) calls buildServerManager.waitForUpToDateBuildGraph()
+    // which ensures SwiftPM has finished re-parsing Package.swift and updated build settings.
+    try await project.testClient.send(SynchronizeRequest(index: true))
+
     // After removing the experimental feature from Package.swift, the parser should
     // produce a proper syntax tree, and DocumentSymbolRequest should find the symbols.
     let response = try await project.testClient.send(


Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -95,6 +96,28 @@ package struct SwiftCompileCommand: Sendable, Equatable, Hashable {
...
+  /// with the same experimental features that the compiler would use.
+  var experimentalFeatures: Parser.ExperimentalFeatures {
+    var features: Parser.ExperimentalFeatures = []
+    var iterator = compilerArgs.makeIterator()
+    while let arg = iterator.next() {
+      if arg == "-enable-experimental-feature", let featureName = iterator.next() {
+        // The feature name from the compiler flag may include a colon-separated
+        // availability suffix (e.g. "FeatureName:adoption"). Strip it before
+        // looking up the parser feature.
+        let baseName = featureName.split(separator: ":").first.map(String.init) ?? featureName
```
rintaro: This avoids creating intermediate `Array`
```suggestion
        let baseName = featureName.fistIndex(of: ":").map{ String(featureName[..<$0]) } ?? featureName
```

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
```diff
@@ -15,9 +15,11 @@ import BuildServerIntegration
 import SKLogging
 import SKTestSupport
 import SourceKitLSP
-import SwiftLanguageService
+@_spi(ExperimentalLanguageFeatures) import SwiftParser
 import XCTest
 
+@testable import SwiftLanguageService
```
rintaro: We don't use `@testable` import. Could you mark the APIs `@_spi(Testing) public` if needed?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -102,15 +102,16 @@ package struct SwiftCompileCommand: Sendable, Equatable, Hashable {
   /// This scans the compiler arguments for `-enable-experimental-feature <name>` flags and maps them
   /// to `Parser.ExperimentalFeatures` values so that the SwiftParser can parse the file correctly
   /// with the same experimental features that the compiler would use.
-  var experimentalFeatures: Parser.ExperimentalFeatures {
+  @_spi(Testing)
+  public var experimentalFeatures: Parser.ExperimentalFeatures {
```
rintaro: Ah sorry I think `package` is enough. Not even `@_spi(Testing)` is needed.

loveucifer: oh ok cool

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
```diff
@@ -40,4 +41,57 @@ final class SwiftCompileCommandsTest: SourceKitLSPTestCase {
...
+        "a", "-enable-experimental-feature", "TrailingComma",
+        "-enable-experimental-feature", "ThenStatements", "b",
+      ],
+      language: .swift
+    )
+    let compileCommand = SwiftCompileCommand(settings)
+    let features = compileCommand.experimentalFeatures
+    XCTAssert(features.contains(.trailingComma), "Expected .trailingComma to be extracted")
+    XCTAssert(features.contains(.thenStatements), "Expected .thenStatements to be extracted")
+    XCTAssertFalse(features.contains(.doExpressions), "Expected .doExpressions to NOT be present")
```
rintaro: Can't we just `XCTAssertEqual(features, [.trailingComma, .thenStatements, .doExpressions])`

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
```diff
@@ -40,4 +41,57 @@ final class SwiftCompileCommandsTest: SourceKitLSPTestCase {
     let compileCommand = SwiftCompileCommand(settings)
     XCTAssertEqual(compileCommand.compilerArgs, ["a", "b", "-working-directory", "/custom-root"])
   }
+
+  func testExperimentalFeaturesExtraction() {
+    let settings = FileBuildSettings(
+      compilerArguments: [
+        "a", "-enable-experimental-feature", "TrailingComma",
+        "-enable-experimental-feature", "ThenStatements", "b",
```
rintaro: Using the actual feature names sounds fragile. Experimental features evolves both in the compiler and SwiftParser. Maybe we should have some permanent/stable "experimental" feature that actually affects the parser behavior.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -25,7 +26,7 @@ import SemanticIndex
 package import SourceKitD
 package import SourceKitLSP
 import SwiftExtensions
-import SwiftParser
+@_spi(ExperimentalLanguageFeatures) public import SwiftParser
```
rintaro, line 29: Is `public` needed?

loveucifer: ig without the public import, the @_spi(ExperimentalLanguageFeatures) imported types remain internal to SwiftLanguageService, causing this  specific build error: "property cannot be declared package because its type uses an internal type".

loveucifer:
> Is it possible to add some e2e test cases? Though, I'm not sure any existing experimental features actually affect the sourcekit-lsp functionalities that uses `SyntaxTreeManager`.
>

i actually tried  attempting  this by adding an integration test with .enableExperimentalFeature("TrailingComma") in a SwiftPMTestProject  but it resulted in a sourcekitd error: "experimental feature 'TrailingComma' cannot be enabled in production compiler".

loveucifer:
implementing an E2E test would require a custom-built toolchain with experimental features enabled would it not ?

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
loveucifer: the tests use real experimental feature names because swift-syntax doesn't  really provide us test-only stable features, ( i think )

rintaro:
> implementing an E2E test would require a custom-built toolchain with experimental features enabled would it not ?

Yes, the toolchain should be built with asserts to actually use the feature. But for our testing purpose, we can ignore the compiler diagnostics. As long as we can test a syntactic functionalities that depends on `SyntaxTreeManager`, it should be fine.

> swift-syntax doesn't really provide us test-only stable features, ( i think )

We can add one :)

loveucifer:
i hope i didnt mess this up since i was going back and forth between both the repos and i might have gotten something mixed up but tests did pass so i hope it works

rintaro:
https://github.com/swiftlang/swift-syntax/pull/3286
@swift-ci Please test

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
```diff
@@ -40,4 +41,55 @@ final class SwiftCompileCommandsTest: SourceKitLSPTestCase {
...
+        "-enable-experimental-feature", "SomeUnknownFeature",
+        "-enable-experimental-feature", "TrailingComma",
+      ],
+      language: .swift
+    )
+    let compileCommand = SwiftCompileCommand(settings)
+    let features = compileCommand.experimentalFeatures
+    XCTAssert(features.contains(.trailingComma), "Expected .trailingComma to be extracted")
+    // Unknown features should be silently ignored
+    XCTAssertEqual(features, .trailingComma, "Expected only .trailingComma, unknown features should be ignored")
```
rintaro: I'm just not a fan of writing `OptionSet` with bare element when describing it as a "set".
```suggestion
    XCTAssertEqual(features, [.trailingComma], "Expected only .trailingComma, unknown features should be ignored")
```
Also the test above (`features.contains(.trailingComma)`) is not necessary

Tests/SourceKitLSPTests/SwiftCompileCommandsTest.swift
```diff
@@ -40,4 +41,55 @@ final class SwiftCompileCommandsTest: SourceKitLSPTestCase {
     let compileCommand = SwiftCompileCommand(settings)
     XCTAssertEqual(compileCommand.compilerArgs, ["a", "b", "-working-directory", "/custom-root"])
   }
+
+  func testExperimentalFeaturesExtraction() {
+    let settings = FileBuildSettings(
+      compilerArguments: [
+        "a", "-enable-experimental-feature", "TrailingComma",
```
rintaro: Now that we have a stable feature name, so we could use that. 
Maybe adding another, for example `_test_Dummy`. So that we can avoid any real feature?

loveucifer:
@swift-ci please test

rintaro:
(This requires cross-repo until the corresponding swift-syntax PR is merged. So the current testing will fail)

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SwiftLanguageService/SyntaxTreeManager.swift
```diff
@@ -90,15 +131,20 @@ actor SyntaxTreeManager {
...
       )
-      return Parser.parseIncrementally(source: postEditSnapshot.text, parseTransition: parseTransition)
     }
     self.setComputation(for: postEditSnapshot.id, computation: incrementalParseComputation)
   }
 
   /// Remove all cached syntax trees for the given document, eg. when the document is closed.
   func clearSyntaxTrees(for uri: DocumentURI) {
     syntaxTreeComputations.removeAll(where: { $0.uri == uri })
+    experimentalFeaturesPerDocument.removeValue(forKey: uri)
```
ahoppen, line 155: There is quite a subtle interaction here how you set the build settings using `setExperimentalFeatures` but they get cleared implicitly when calling `clearSyntaxTrees`. Wouldn’t it be more explicit if we explicitly set the experimental features for the document to `nil` from `closeDocument`. Then we’d have a nice match how `openDocument`/`reopenDocument` set the build settings and `closeDocument` clears them again.

Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
```diff
@@ -762,4 +762,96 @@ final class SwiftPMIntegrationTests: SourceKitLSPTestCase {
...
+
+    let packageManifestPath = project.scratchDirectory.appending(component: "Package.swift")
+    try await newManifest.writeWithRetry(to: packageManifestPath)
+    project.testClient.send(
+      DidChangeWatchedFilesNotification(changes: [
+        FileEvent(uri: DocumentURI(packageManifestPath), type: .changed)
+      ])
+    )
+    // Ensure that the DidChangeWatchedFilesNotification is handled before we continue.
+    _ = try await project.testClient.send(SynchronizeRequest())
```
ahoppen: I think you should be able to use `project.changeFileOnDisk`, which handles this entire dance.

Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
```diff
@@ -762,4 +762,96 @@ final class SwiftPMIntegrationTests: SourceKitLSPTestCase {
...
+
+    // After removing the experimental feature from Package.swift, the parser should
+    // produce a proper syntax tree, and DocumentSymbolRequest should find the symbols.
+    // Wait for build settings to update and verify symbols are now found.
+    try await repeatUntilExpectedResult {
```
ahoppen: If we send a `SynchronizeRequest` above, do we really still need to do `repeatUntilExpectedResult`? Shouldn’t all changes be reflected by now?

ahoppen (approved):
Looks good, thank you!

ahoppen:
@swift-ci Please test

ahoppen:
Looks like the tests failed on Linux

```
/home/build-user/sourcekit-lsp/Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift:839: error: SwiftPMIntegrationTests.testExperimentalFeaturesPassedToSyntaxTreeManager : XCTUnwrap failed: expected non-nil value of type "DocumentSymbol" - 
```

loveucifer:
hmm the ci did run fine before the previous commit

loveucifer:
@ahoppen  I think On Linux (and possibly on macOS), SwiftPM probably takes time to re-parse Package.swift and update build settings , even after DidChangeWatchedFilesNotification + SynchronizeRequest  i think thats why the error came on the second CI trigger , should we go back to repeatUntilExpectedResult ?

loveucifer:
i think this will work now

loveucifer:
@swift-ci please test

loveucifer:
@swift-ci   Please Test Windows

loveucifer:
huh failed again :0

loveucifer:
pinging in case this was missed :) @ahoppen @rintaro

rintaro:
> I think On Linux (and possibly on macOS), SwiftPM probably takes time to re-parse Package.swift and update build settings , even after DidChangeWatchedFilesNotification + SynchronizeRequest 

`changeFileOnDisk()` only sends the default `SynchronizeRequest()` which doesn't do `buildServerManager.waitForUpToDateBuildGraph()`.
Could you `try await project.testClient.send(SynchronizeRequest(index: true))` after `changeFileOnDisk()`, instead of `repeatUntilExpectedResult`?

loveucifer:
@swift-ci Please Test

loveucifer:
failed in like 2 mins whaaat :0

rintaro:
The failure is unrelated, let's wait for it being fixed.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

loveucifer:
windows CI has some issues ?

rintaro:
@swift-ci Please test Windows

loveucifer:
🎊  letsgo

rintaro (approved):
Thank you!

PR #2503, opened 2026-02-27
SemanticIndex | Add deterministic IndexStoreDB cleanup on shutdown

sean7218:
## Summary

Refactors UncheckedIndex to wrap IndexStoreDB in a ThreadSafeBox, enabling explicit lifecycle management. A new close() method sets the underlying instance to nil, triggering deterministic deinitialization rather than relying on ARC-driven deallocation.

Introduces a close() method on the MainFilesProvider protocol (with a default no-op) so BuildServerManager can tear down the index store without a circular dependency on SemanticIndex.
Calls mainFilesProvider.close() as the first step of BuildServerManager.shutdown(), ensuring the IndexStoreDB is released before the build server connection is torn down.

## Motivation

Previously, the IndexStoreDB instance was only released when UncheckedIndex was deallocated, which depends on ARC timing and can lead to non-deterministic resource cleanup. In test scenarios this caused race conditions where the connection could be reported as a leak. This change gives the shutdown path explicit control over when the index store is closed.

## Changes

* CheckedIndex.swift — UncheckedIndex.underlyingIndexStoreDB is now a computed property backed by a ThreadSafeBox<UncheckedSendableOptionalIndexStoreDB>. Added close() to nil out the stored instance. Accessing the index after close triggers a preconditionFailure.

* MainFilesProvider.swift — Added func close() async to the protocol with a default empty implementation.
BuildServerManager.swift — shutdown() now calls await self.mainFilesProvider?.value?.close() before proceeding with build server teardown.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -271,6 +271,8 @@ var targets: [Target] = [
     dependencies: [
       "SemanticIndex",
       "SKTestSupport",
+      "ToolchainRegistry",
+      .product(name: "IndexStoreDB", package: "indexstore-db"),
       .product(name: "SKLogging", package: "swift-tools-protocols"),
       .product(name: "ToolsProtocolsSwiftExtensions", package: "swift-tools-protocols"),
     ],
--- a/Tests/SemanticIndexTests/TaskSchedulerTests.swift
+++ b/Tests/SemanticIndexTests/TaskSchedulerTests.swift
@@ -10,10 +10,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+@preconcurrency import IndexStoreDB
 @_spi(SourceKitLSP) import SKLogging
 import SKTestSupport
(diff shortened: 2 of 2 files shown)


ahoppen:
Maybe I missed something but does this contain anything that’s not also covered by https://github.com/swiftlang/sourcekit-lsp/pull/2483? I think the problem that https://github.com/swiftlang/sourcekit-lsp/issues/2455#issuecomment-3907291720 mentioned was that if SourceKit-LSP is not idle, there may still be lingering references to the `IndexStoreDB` object, so really what we need is a proper `close` method inside the indexstore-db repository instead of relying on ref-counting there.

sean7218:
You are absolutely right! 😅 I misunderstood the problem. Let me fix it on the upstream side

PR #2528, opened 2026-03-02
Feature/convert comments code action

### LINKED ISSUE #2523: Add Convert between line and block comments code action
### Description

Convert `//` line comments to `/* */` block comments and vice versa.

**Before:**

```swift
// This is a comment
// that spans multiple
// lines
func example() {}
```

**After:**

```swift
/*
 This is a comment
 that spans multiple
 lines
 */
func example() {}
```

loveucifer:
Resolves #2523

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
+++ b/Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
@@ -12,6 +12,8 @@
 
 @_spi(SourceKitLSP) import LanguageServerProtocol
 import SourceKitLSP
+import SwiftBasicFormat
+import SwiftExtensions
 import SwiftSyntax
 
 /// Syntactic code action provider to convert between `//` line comments and
@@ -27,20 +29,10 @@ struct ConvertComments: SyntaxCodeActionProvider {
     if let action = codeAction(
       for: token.leadingTrivia,
       startPosition: token.position,
+      token: token,
       scope: scope
     ) {
       return [action]
-    }
-
-    // Also check trailing trivia of the previous token
-    if let previousToken = token.previousToken(viewMode: .sourceAccurate) {
-      if let action = codeAction(
-        for: previousToken.trailingTrivia,
-        startPosition: previousToken.endPositionBeforeTrailingTrivia,
-        scope: scope
-      ) {
-        return [action]
-      }
     }
 
     return []
@@ -49,11 +41,12 @@ struct ConvertComments: SyntaxCodeActionProvider {
   private static func codeAction(
     for trivia: Trivia,
     startPosition: AbsolutePosition,
+    token: TokenSyntax,
     scope: SyntaxCodeActionScope
   ) -> CodeAction? {
     var position = startPosition
     var lineComments: [(start: AbsolutePosition, text: String)] = []
-    var lineCommentGroupStart: AbsolutePosition?
+    var consecutiveNewlines = 0
 
     for piece in trivia {
       let pieceStart = position
@@ -62,16 +55,14 @@ struct ConvertComments: SyntaxCodeActionProvider {
 
       switch piece {
       case .lineComment(let text):
-        if lineCommentGroupStart == nil {
-          lineCommentGroupStart = pieceStart
-        }
         lineComments.append((pieceStart, text))
+        consecutiveNewlines = 0
 
       case .blockComment(let text):
         // If cursor is in this block comment, offer to convert to line comments
         let pieceRange = pieceStart..<position
         if pieceRange.overlaps(scope.range) {
-          let indentation = leadingIndentation(at: pieceStart, in: scope.snapshot.text)
+          let indentation = token.indentationOfLine.description
           return CodeAction(
             title: "Convert Block Comment to Line Comments",
             kind: .refactorInline,
@@ -85,10 +76,17 @@ struct ConvertComments: SyntaxCodeActionProvider {
             ])
           )
         }
+        consecutiveNewlines = 0
 
-      case .newlines, .carriageReturns, .carriageReturnLineFeeds:
-        // Newline continues a potential line comment group
-        continue
+      case .newlines(let count), .carriageReturns(let count), .carriageReturnLineFeeds(let count):
+        consecutiveNewlines += count
+        // Break line comment group if there's an empty line (2+ newlines)
+        if consecutiveNewlines >= 2 && !lineComments.isEmpty {
+          if let action = lineCommentsAction(lineComments, token: token, scope: scope) {
+            return action
+          }
+          lineComments.removeAll()
+        }
 
       case .spaces, .tabs:
         // Whitespace is allowed between line comments
@@ -96,28 +94,28 @@ struct ConvertComments: SyntaxCodeActionProvider {
 
       default:
         // Other trivia breaks the line comment group
-        if let action = lineCommentsAction(lineComments, groupStart: lineCommentGroupStart, scope: scope) {
+        if let action = lineCommentsAction(lineComments, token: token, scope: scope) {
           return action
         }
         lineComments.removeAll()
-        lineCommentGroupStart = nil
+        consecutiveNewlines = 0
       }
     }
 
     // Check remaining line comments at end of trivia
-    return lineCommentsAction(lineComments, groupStart: lineCommentGroupStart, scope: scope)
+    return lineCommentsAction(lineComments, token: token, scope: scope)
   }
 
   private static func lineCommentsAction(
     _ comments: [(start: AbsolutePosition, text: String)],
-    groupStart: AbsolutePosition?,
+    token: TokenSyntax,
     scope: SyntaxCodeActionScope
   ) -> CodeAction? {
-    guard !comments.isEmpty, let start = groupStart else {
+    guard let firstComment = comments.first, let lastComment = comments.last else {
       return nil
     }
 
-    let lastComment = comments.last!
+    let start = firstComment.start
     let end = lastComment.start.advanced(by: lastComment.text.utf8.count)
     let groupRange = start..<end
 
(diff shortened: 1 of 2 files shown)


loveucifer:
the way i implemented this only examines the token at the cursor position and its trivia, rather than iterating through all tokens in the file ( i think this is better but i might be wrong )

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+    // Check leading trivia of the token (comments typically appear here)
+    if let action = codeAction(
+      for: token.leadingTrivia,
+      startPosition: token.position,
+      scope: scope
+    ) {
+      return [action]
+    }
+
+    // Also check trailing trivia of the previous token
```
ahoppen: Why is this needed? Shouldn’t `token(at:)` always return the token that contains the trivia under the current cursor?

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+    var lineCommentGroupStart: AbsolutePosition?
+
+    for piece in trivia {
+      let pieceStart = position
+      let pieceLength = piece.sourceLength.utf8Length
+      position = position.advanced(by: pieceLength)
+
+      switch piece {
+      case .lineComment(let text):
+        if lineCommentGroupStart == nil {
```
ahoppen: Should we also check if the cursor is inside the comment here? So we convert the correct comment if we have something like the following. Also, we probably shouldn’t offer the refactoring action if the cursor isn’t inside a comment.

```swift
// Line 1
/* line 2 */
// Line 3
```

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+                  range: scope.snapshot.absolutePositionRange(of: pieceRange),
+                  newText: blockToLineComments(text, indentation: indentation)
+                )
+              ]
+            ])
+          )
+        }
+
+      case .newlines, .carriageReturns, .carriageReturnLineFeeds:
+        // Newline continues a potential line comment group
```
ahoppen: Do you think we should break a line comment group if there’s an empty line in between them? Ie.

```swift
// Line 1

// Line 2
func foo() {}
```

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+  private static func lineCommentsAction(
+    _ comments: [(start: AbsolutePosition, text: String)],
+    groupStart: AbsolutePosition?,
+    scope: SyntaxCodeActionScope
+  ) -> CodeAction? {
+    guard !comments.isEmpty, let start = groupStart else {
+      return nil
+    }
+
+    let lastComment = comments.last!
```
ahoppen: We can avoid the force unwrap here if you use this instead of `comment.isEmpty` above.

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    for piece in trivia {
+      let pieceStart = position
+      let pieceLength = piece.sourceLength.utf8Length
+      position = position.advanced(by: pieceLength)
+
+      switch piece {
+      case .lineComment(let text):
+        if lineCommentGroupStart == nil {
+          lineCommentGroupStart = pieceStart
```
ahoppen: Do we need `lineCommentGroupStart` at all? Isn’t this the same as `lineComments.first`?

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    guard groupRange.overlaps(scope.range) else {
+      return nil
+    }
+
+    let indentation = leadingIndentation(at: start, in: scope.snapshot.text)
+    let texts = comments.map { String($0.text.dropFirst(2)) }  // Remove "//"
+
+    return CodeAction(
+      title: "Convert Line Comments to Block Comment",
```
ahoppen: Should this be *Comment* without the *s* if there’s only a single line comment?

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+            range: scope.snapshot.absolutePositionRange(of: groupRange),
+            newText: lineToBlockComment(texts, indentation: indentation)
+          )
+        ]
+      ])
+    )
+  }
+
+  private static func lineToBlockComment(_ lines: [String], indentation: String) -> String {
+    if lines.count == 1 {
```
ahoppen: You could use `if let line = lines.only` here, which means that you wouldn’t need the `[0]` access below.

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+      return "/*\(lines[0])*/"
+    }
+    return "/*\n" + lines.map { "\(indentation)\($0)\n" }.joined() + "\(indentation)*/"
```
ahoppen: Would this look nicer as a multi-line string literal? What do you think?


```suggestion
    return """
    /*
    \(lines.map { "\(indentation)\($0)\n" }.joined())
    \(indentation)*/
    """
```

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  private static func blockToLineComments(_ text: String, indentation: String) -> String {
+    let body = String(text.dropFirst(2).dropLast(2))  // Remove /* and */
+    var lines = body.split(separator: "\n", omittingEmptySubsequences: false).map(String.init)
+
+    // Trim empty first/last lines from multiline block comments
+    if lines.count > 1 {
```
ahoppen: Is this check needed? Wouldn’t this all be covered by `first` or `last` being `nil`?

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+      if lines.first?.allSatisfy(\.isWhitespace) == true { lines.removeFirst() }
```
ahoppen: I generally prefer to spell it like this because it explicitly assigns a boolean value to `nil`.

```suggestion
      if lines.first?.allSatisfy(\.isWhitespace) ?? false { lines.removeFirst() }
```

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    return lines.map { line in
+      let trimmed = line.hasPrefix(indentation) ? String(line.dropFirst(indentation.count)) : line
+      return "//\(trimmed)"
+    }.joined(separator: "\n")
+  }
+
+  private static func leadingIndentation(at position: AbsolutePosition, in source: String) -> String {
+    let index = source.utf8.index(source.startIndex, offsetBy: position.utf8Offset)
+    let lineStart = source[..<index].lastIndex(of: "\n").map { source.index(after: $0) } ?? source.startIndex
```
ahoppen: We would also need to handle `\r` and `\r\n` here. Would it be possible to instead call `Trivia.indentation(isOnNewline:)`, maybe by just dropping a few trivia pieces or something like that? My experience is just that getting the indentation at a given location is more tricky than you think it would be.

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,174 @@
...
+    return String(source[lineStart..<index].prefix { $0 == " " || $0 == "\t" })
+  }
+}
```
ahoppen: Could you run swift-format to ensure there’s a trailing newline here?

ahoppen (commented):
Nice, thanks for implementing this, @loveucifer

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,185 @@
...
+        consecutiveNewlines += count
+        // Break line comment group if there's an empty line (2+ newlines)
+        if consecutiveNewlines >= 2 && !lineComments.isEmpty {
+          if let action = lineCommentsAction(lineComments, token: token, scope: scope) {
+            return action
+          }
+          lineComments.removeAll()
+        }
+
+      case .carriageReturns(let count):
```
ahoppen: Can’t you implement all of these in one `case` block?

```suggestion
      case .newlines(let count), .carriageReturns(let count), .carriageReturnLineFeeds(let count):
        consecutiveNewlines += count
        // Break line comment group if there's an empty line (2+ newlines)
        if consecutiveNewlines >= 2 && !lineComments.isEmpty {
          if let action = lineCommentsAction(lineComments, token: token, scope: scope) {
            return action
          }
          lineComments.removeAll()
        }
```

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,185 @@
...
+      let pieceLength = piece.sourceLength.utf8Length
+      position = position.advanced(by: pieceLength)
+
+      switch piece {
+      case .lineComment(let text):
+        lineComments.append((pieceStart, text))
+        consecutiveNewlines = 0
+
+      case .blockComment(let text):
+        // If cursor is in this block comment, offer to convert to line comments
```
ahoppen, line 62: Should we also reset `consecutiveNewlines` here?

swift-syntax
ahoppen, line 1: Unrelated change?

Sources/SwiftLanguageService/CodeActions/ConvertComments.swift
```diff
@@ -0,0 +1,185 @@
...
+    )
+  }
+
+  private static func lineToBlockComment(_ lines: [String], indentation: String) -> String {
+    if let line = lines.only {
+      return "/*\(line)*/"
+    }
+    return """
+      /*
+      \(lines.map { "\(indentation)\($0)\n" }.joined())\(indentation)*/
```
ahoppen: It should behave the same but I feel like the intention is clearer if this is written as

```suggestion
      \(lines.map { "\(indentation)\($0)" }.joined(separator: "\n"))
      \(indentation)*/
```

ahoppen (commented):
A couple more small things, otherwise looks good.

swift-syntax
loveucifer: oh lol , i was reading that repo on the side , my bad

loveucifer:
@swift-ci Please Test

rintaro:
We decided not to accept this code action
https://github.com/swiftlang/sourcekit-lsp/issues/2523#issuecomment-4033203317

PR #2529, opened 2026-03-02
Add Inline temp variable code action

### LINKED ISSUE #2507: Add Inline temp variable code action
### Description

## Example

**Before:**

```swift
func example() {
    let basePrice = item.price
}
```

**After:**

```swift
func example() {
}
```

amanmaurya92:
### Summary
Implements the **Inline temporary variable** refactor action: replaces occurrences of a local `let` binding with its initializer expression in the enclosing scope and removes the declaration.

Fixes #2507

#### Before:
```swift
func example() {
    let basePrice = item.price
}
```

#### After:
```swift
func example() {
}
```

### Implementation Details
- **Syntax Provider**: Implemented as `InlineTempVariable: SyntaxCodeActionProvider` in `SwiftSyntaxCodeActions`.
- **Lexical Lookup**: Uses `SwiftLexicalLookup` during reference collection to accurately resolve references to the declaration and avoid inlining identifiers shadowed by inner scopes or closure parameters.
- **Precedence Preservation**: Wraps composite expressions in parentheses where necessary when inlining into higher-precedence contexts (e.g., `let x = 1 + 2; let y = x * 3` -> `let y = (1 + 2) * 3`).
- **Safety**: Only offered when the cursor is on the `let` binding and there is at least one usage to inline.

### Testing
- Added unit and integration tests in `InlineTempVariableTests.swift`:
  - Basic variable inlining.
  - Multiple references in the same scope.
  - Parentheses insertion for precedence preservation.
  - Outer scope inlining with inner shadowed variable preservation.
  - Inner shadowed variable inlining.
  - Closure parameter shadowing.
  - Ensuring the action is not offered when no usages exist.


scripts/build-linux.sh
ahoppen, line 1: This is a change that’s unrelated to the issue you’re fixing. Could you exclude it from this PR?

Sources/SwiftSyntaxCodeActions/InlineTempVariable.swift
```diff
@@ -0,0 +1,232 @@
...
+  private let afterPosition: AbsolutePosition
+  private(set) var references: [DeclReferenceExprSyntax] = []
+
+  init(name: String, afterPosition: AbsolutePosition) {
+    self.name = name
+    self.afterPosition = afterPosition
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  override func visit(_ node: DeclReferenceExprSyntax) -> SyntaxVisitorContinueKind {
```
ahoppen, line 216: This is trying to re-implement lexical lookup and won’t take shadowing and other name lookup rules into account. Luckily all of this is implemented in `SwiftLexicalLookup`. Could you use that library instead?

ahoppen (commented):
Thanks for starting this. I have one high-level comment and will take a closer look once that’s addressed.

Sources/SwiftSyntaxCodeActions/InlineTempVariable.swift
amanmaurya92: Reference collection now uses SwiftLexicalLookup, i used the library.

amanmaurya92:
@rintaro can I keep working on this? Are we still adding this as a code action?

ahoppen:
Yes, we think this code action provides functionality that is somewhat common, tedious and somewhat hard to get right, which qualifies it.

Sources/SwiftSyntaxCodeActions/InlineTempVariable.swift
ahoppen: As far as I can tell, you are still only comparing there base name here, not actually calling int `SwiftLexicalLookup`.

ahoppen:
Hi @amanmaurya92,

https://github.com/swiftlang/sourcekit-lsp/pull/2754 picked this refactoring up after inactivity in this PR and is close to being merged. I would thus go with that PR’s implementation instead of this one. Sorry for the conflict here. I would encourage you to look for other issues labeled *good first issue* that don’t currently have an active PR associated with them if you want to contribute.

Thanks!

amanmaurya92:
Ohh, Thanks @ahoppen for the closure.

PR #2530, opened 2026-03-02
Add Remove unused parameter code action

amanmaurya92:
Adds an LSP refactor that removes an unused function/initializer parameter and updates all call sites in the same file.

### Example: 
```
func greet(name: String, title: String) { 
print("Hello, \(name)") 
} 
With the cursor on the title → parameter and its argument at call sites are removed, so the signature becomes 
(name: String) 
and calls become greet(name: "Alice").
```

#### RemoveUnusedParameter.swift (new): 
Detects “unused” by checking that the parameter name does not appear as a DeclReferenceExpr in the body (syntactic only). 

Builds a single declaration edit (range ends at rightParen.endPositionBeforeTrailingTrivia so the space before { is kept) and per-call-site edits by resolving relatedIdentifiers at the function name and removing the matching argument (and comma) from each FunctionCallExprSyntax.

#### SwiftLanguageService.swift: 
Registers retrieveRemoveUnusedParameterCodeActions in the code action pipeline with kind .refactor.
#### CMakeLists.txt: 
Adds the new file to the SwiftLanguageService target.
#### CodeActionTests.swift: 
Adds testRemoveUnusedParameterCodeAction (request at parameter, apply edit to unmarked source, assert exact result).


ahoppen (commented):
Same as in https://github.com/swiftlang/sourcekit-lsp/pull/2529#discussion_r2901616190, we should use `SwiftLexcialLookup` instead of writing our own `DeclReferenceCollector` here. 

Also, in order to remove the arguments from the call sides, we’d need to find all cables in the entire project, not just in the current file and matching the arguments to the function’s parameters is more tricky as I described in https://github.com/swiftlang/swift-syntax/issues/2047. 

I think the best path forward here would be to first ensure we have a solid argument to parameter matching implemented in swift-syntax and then start using it here.

PR #2531, opened 2026-03-02
Add flip binary operands code action

### LINKED ISSUE #2520: Add Flip operands of binary expression code action
### Description

**Before:**

```swift
if 5 < count {
    process()
}
```

**After:**

```swift
if count > 5 {
    process()
}
```

**Before:**

```swift
let sum = 1 + value
```

**After:**

```swift
let sum = value + 1
```

corvid-agent:
Fixes #2520

Adds a new `FlipBinaryOperands` syntactic code action that swaps the left and right operands of a binary expression, adjusting the operator when needed.

## Behavior

**Commutative operators** (`+`, `*`, `==`, `!=`, `&&`, `||`, etc.) — operands are simply swapped:
```swift
let sum = 1 + value  →  let sum = value + 1
```

**Comparison operators** — the operator is also flipped to preserve semantics:
| Before | After |
|--------|-------|
| `5 < count` | `count > 5` |
| `a > b` | `b < a` |
| `a <= b` | `b >= a` |
| `a >= b` | `b <= a` |

## Changes

- `FlipBinaryOperands.swift` — new code action provider
- `SyntaxCodeActions.swift` — register the provider
- `CMakeLists.txt` — add the new file
- `CodeActionTests.swift` — 7 tests covering commutative, comparison, and equality operators

## Test plan

- [x] `swift build --target SwiftLanguageService` passes
- [x] CI tests pass
- [x] @0xLeif review

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
+++ b/Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
@@ -10,8 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(SourceKitLSP) import LanguageServerProtocol
-import SourceKitLSP
+import SwiftRefactor
 import SwiftSyntax
 
 /// Syntactic code action provider to swap the left and right operands of a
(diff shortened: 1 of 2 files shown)


Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
```diff
@@ -0,0 +1,117 @@
...
+    // and the trailing trivia of the original right on the new right.
+    let newLeft = rightOperand
+      .with(\.leadingTrivia, leftOperand.leadingTrivia)
+      .with(\.trailingTrivia, leftOperand.trailingTrivia)
+    let newRight = leftOperand
+      .with(\.leadingTrivia, rightOperand.leadingTrivia)
+      .with(\.trailingTrivia, rightOperand.trailingTrivia)
+
+    let newOperator: ExprSyntax
+    if flippedOperator != operatorText {
```
ahoppen: Is this check actually needed? If `flippedOperator == operatorText` then overwriting the operator doesn’t do any harm.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
```diff
@@ -0,0 +1,117 @@
...
+      newOperator = ExprSyntax(
+        binaryOp.with(\.operator, .binaryOperator(flippedOperator))
```
ahoppen: Do we need to preserve trivia of the original operator here?

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
```diff
@@ -0,0 +1,117 @@
...
+/// For commutative operators the operands are simply swapped:
+/// ```swift
+/// let sum = 1 + value  →  let sum = value + 1
+/// ```
+///
+/// For comparison operators the operator is also flipped:
+/// ```swift
+/// if 5 < count { }  →  if count > 5 { }
+/// ```
+struct FlipBinaryOperands: SyntaxCodeActionProvider {
```
ahoppen: Could we implement this as an `SyntaxRefactoringCodeActionProvider` similar to `AddDocumentation`? That gives us some convenience functionality to construct the `CodeAction` and would allow us to move this to swift-syntax if we want to in the future.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
```diff
@@ -0,0 +1,117 @@
...
+
+  /// Returns the flipped operator for comparison operators, or the same
+  /// operator text for commutative operators.
+  private static func flippedOperatorText(for op: String) -> String {
+    switch op {
+    case "<": return ">"
+    case ">": return "<"
+    case "<=": return ">="
+    case ">=": return "<="
+    default: return op
```
ahoppen, line 87: We shouldn’t assume that all operators not listed above are commutative. Users may, for example, have their own non-commutative operators. And eg. `+` on strings or `-` is not commutative. Instead, we should have an allow-list of commutative operators.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+}
+
+private extension SyntaxProtocol {
+  /// Walk up the tree to find the nearest `InfixOperatorExprSyntax` that
+  /// contains the current node.
+  func findEnclosingInfixOperator() -> InfixOperatorExprSyntax? {
+    var current: Syntax? = Syntax(self)
```
ahoppen: You should be able to user `ancestorOrSelf` here.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1564,6 +1564,199 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
   }
 
+  // MARK: - Flip binary operands
+
+  func testFlipBinaryOperandsCommutative() async throws {
+    try await assertCodeActions(
+      """
+      let sum = 1️⃣1 + value2️⃣
+      """,
+      ranges: [("1️⃣", "2️⃣")],
```
ahoppen: Do we need to specify the ranges here? Doesn’t the default invoke the code actions on all markers and w should return the actions on all markers?

ahoppen (commented):
Thanks for your contribution @corvid-agent. I have a few comments inline but the overall implementation looks solid.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
corvid-agent: Good point — removed the guard. The new operator token is always constructed now, even when the text is the same.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
corvid-agent: Fixed — the flipped operator token now preserves both leading and trailing trivia from the original.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
corvid-agent: Done — refactored to match the `AddDocumentation` pattern: `EditRefactoringProvider` with `textRefactor` + `SyntaxRefactoringCodeActionProvider` extension with `nodeToRefactor` and `title`.

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
corvid-agent: Great catch — replaced the blocklist with an explicit allowlist: `==`, `\!=`, `&&`, `||`, `&`, `|`, `^`, `<`, `>`, `<=`, `>=`. Comparison operators still get flipped (e.g. `<` → `>`).

Sources/SwiftLanguageService/CodeActions/FlipBinaryOperands.swift
corvid-agent: Switched to `findParentOfSelf(ofType:stoppingIf:)` — same pattern `AddDocumentation` uses for its ancestor walk.

Tests/SourceKitLSPTests/CodeActionTests.swift
corvid-agent: You're right — removed the explicit `ranges:` from all tests. The default behavior covers all markers.

corvid-agent:
Thanks for the thorough review @ahoppen\! All 6 items addressed in the latest push:

1. Refactored to `EditRefactoringProvider` + `SyntaxRefactoringCodeActionProvider` (matching `AddDocumentation` pattern)
2. Replaced operator blocklist with explicit allowlist (`==`, `\!=`, `&&`, `||`, `&`, `|`, `^`, `<`, `>`, `<=`, `>=`)
3. Removed the unnecessary `flippedOperator == operatorText` guard
4. Operator token now preserves original trivia
5. Using `findParentOfSelf(ofType:stoppingIf:)` for ancestor walk
6. Removed explicit `ranges:` from all tests

rintaro:
We decided not to accept this code action https://github.com/swiftlang/sourcekit-lsp/issues/2520#issuecomment-4033276729

rintaro:
Reopened  https://github.com/swiftlang/sourcekit-lsp/issues/2520

PhantomInTheWire:
@rintaro @ahoppen out of curiosity what is the policy on non human contributors since it seems @corvid-agent is an ai agent per their profile

rintaro:
We recently established an official policy on AI-assisted contributions: https://github.com/swiftlang/project-operations/blob/main/contributing/ai-tools.md

@corvid-agent, since you describe yourself as an “AI software engineer,” I’m going to close this PR for now. If you are in fact a human contributor and are able to comply with the policy, please feel free to reopen it.

Issue #2525, opened 2026-02-28
Add @Sendable annotation code action


Tabonx:
### Description

Add `@Sendable` to closures and function types for concurrency safety.

**Before:**

```swift
func perform(callback: @escaping () -> Void) {
    Task {
        callback()
    }
}
```

**After:**

```swift
func perform(callback: @Sendable @escaping () -> Void) {
    Task {
        callback()
    }
}
```

rintaro:
I don't think it's worth to have this as a code action. It's easy to do by hand.

PR #2532, opened 2026-03-03
Add @Sendable annotation code action

amanthatdoescares:
## Add @Sendable annotation code action (#2525)

Adds an LSP code action that inserts `@Sendable` on closure and function types for concurrency safety.

### Behavior

- **Trigger:** Cursor or selection on a function/closure type (e.g. parameter type).
- **Action:** Inserts `@Sendable ` at the start of that type.
- **Skip:** Types that already have `@Sendable` do not get the action.

### Example

**Before:**
```
func perform(callback: @escaping () -> Void) {
    Task {
        callback()
    }
}
```

**After:**
```
func perform(callback: @Sendable @escaping () -> Void) {
    Task {
        callback()
    }
}
```

#### New: 
Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift — syntactic provider that finds the innermost function type (or attributed function type) under the selection, ensures it does not already have @Sendable, and offers a single refactor that inserts "@Sendable " at the type’s start.

#### Registration: 
Provider added to SyntaxCodeActions.swift and CMakeLists.txt.

#### Tests: 
testAddSendableAnnotation and testAddSendableAnnotationNotOfferedWhenAlreadySendable in CodeActionTests.swift.

fixes: #2525

### DIFF
--- a/Tests/SourceKitLSPTests/CodeActionTests.swift
+++ b/Tests/SourceKitLSPTests/CodeActionTests.swift
@@ -1778,6 +1778,9 @@ final class CodeActionTests: SourceKitLSPTestCase {
       []
     }
   }
+  func testAddSendableAnnotation() async throws {
+    try await assertCodeActions(
+      """
       func perform(callback: 1️⃣@escaping () -> Void) {
       }
       """,


Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
```diff
@@ -0,0 +1,92 @@
...
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftSyntax
+
+/// Code action that adds `@Sendable` to closures and function types for concurrency safety.
+///
+/// ## Example
+/// Before: `func perform(callback: @escaping () -> Void) { ... }`
+/// After:  `func perform(callback: @Sendable @escaping () -> Void) { ... }`
+struct AddSendableAnnotation: SyntaxCodeActionProvider {
```
ahoppen: Could you implement this as a `SyntaxRefactoringCodeActionProvider`. It gives us some convenience methods and would allow us to move this to the swift-syntax repo in the future if we want to do so.

Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
```diff
@@ -0,0 +1,92 @@
...
+  if let attributed = type.as(AttributedTypeSyntax.self) {
+    return typeIsOrContainsFunctionType(attributed.baseType)
+  }
+  return false
+}
+
+private func attributedTypeHasSendable(_ attributed: AttributedTypeSyntax) -> Bool {
+  attributed.attributes.contains { element in
+    guard let attr = element.as(AttributeSyntax.self) else { return false }
+    return attributeNameIsSendable(attr.attributeName)
```
ahoppen: We should be able to just inline the function here, which isn’t really any more complicated.

```suggestion
    return attributeName.as(IdentifierTypeSyntax.self).identifier?.name.text == "Sendable"
```

Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
```diff
@@ -0,0 +1,92 @@
...
+/// A type node (function type or attributed function type) that can have @Sendable added.
+private struct SendableAnnotationTarget {
+  let type: TypeSyntax
+}
+
+/// Walks up from the scope's innermost node to find a function type (or attributed function type)
+/// that does not already have @Sendable and returns the type node to edit.
+private func findFunctionTypeToAnnotate(in scope: SyntaxCodeActionScope) -> SendableAnnotationTarget? {
+  var node: Syntax? = scope.innermostNodeContainingRange
+  while let current = node, !current.is(CodeBlockSyntax.self), !current.is(MemberBlockSyntax.self) {
```
ahoppen: I think you should be able to use `ancestorOrSelf` here.

Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
```diff
@@ -0,0 +1,92 @@
...
+            scope.snapshot.uri: [edit]
+          ]
+        )
+      )
+    ]
+  }
+}
+
+/// A type node (function type or attributed function type) that can have @Sendable added.
+private struct SendableAnnotationTarget {
```
ahoppen: Is it really worth having this wrapper type? Can’t we just return `TypeSyntax` instead?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1778,6 +1778,44 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    }
+  }
+
+  func testAddSendableAnnotationNotOfferedWhenAlreadySendable() async throws {
+    try await assertCodeActions(
+      """
+      func perform(callback: 1️⃣@Sendable @escaping () -> Void) { }
+      """,
+      markers: ["1️⃣"],
+      exhaustive: false
```
ahoppen: This doesn’t check anything. You will want too et `exhaustive: true` (ie. omit it) to negatively check that a code action doesn’t exist.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1778,6 +1778,44 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
     }
   }
+  func testAddSendableAnnotation() async throws {
+    try await assertCodeActions(
+      """
+      func perform(callback: 1️⃣@escaping () -> Void) {
+          Task {
+              callback()
+          }
+      }
```
ahoppen, line 1785: You can simplify this a little bit because the code action doesn’t actually care about the function’s body.

```suggestion
      func perform(callback: 1️⃣@escaping () -> Void) {
      }
```

Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
amanmaurya92: I’ll switch to ancestorOrSelf(mapping:) here, Thank you

Sources/SwiftLanguageService/CodeActions/AddSendableAnnotation.swift
amanmaurya92: Yes @ahoppen , i have tried that now.

rintaro:
We decided not to accept this code action https://github.com/swiftlang/sourcekit-lsp/issues/2525#issuecomment-4033175376

PR #2534, opened 2026-03-04
[6.3] Fix issue that caused the target preparation status to not be cleared when workspace root did not have trailing slash

ahoppen:
- **Explanation**: If a workspace was opened as `/path/to/workspace` instead of `/path/to/workspace/` (notice the trailing slash), the check for `url.deletingLastPathComponent() == self.projectRoot` failed because deleting the last path component always produces a URL with a trailing slash and URL considers two path different if they mismatch in their trailing slash usage.
Use `DocumentURI` for equality checking, which is a little more lenient with declaring equality, including with regard to trailing slashes, and whose equality definition aligns better with what we want in SourceKit-LSP.
- **Scope**: Files did change events for Package manifest files
- **Issue**: rdar://171744166
- **Original PR**: https://github.com/swiftlang/sourcekit-lsp/pull/2363
- **Risk**: Low, makes the URL equality comparison for package manifest changes a little more relaxed
- **Reviewer**: @rintaro

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -844,7 +844,9 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     else {
       return false
     }
-    return url.deletingLastPathComponent() == self.projectRoot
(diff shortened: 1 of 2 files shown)


nkcsgexi (approved):

rintaro:
@swift-ci Please test

PR #2536, opened 2026-03-05
Add explicit enum case raw values code action

### LINKED ISSUE #2516: Add explicit enum case raw values code action
### Description

Add explicit raw values to all cases of an enum that has an implicit raw value type.

**Before:**

```swift
enum Status: Int {
    case active
    case inactive
    case pending = 10
    case archived
}
```

**After:**

```swift
enum Status: Int {
    case active = 0
    case inactive = 1
    case pending = 10
    case archived = 11
}
```

william-laverty:
## Summary

Adds a new syntactic code action that makes implicit raw values explicit for enums with `Int` or `String` raw value types.

### Before

```swift
enum Status: Int {
    case active
    case inactive
    case pending = 10
    case archived
}
```

### After

```swift
enum Status: Int {
    case active = 0
    case inactive = 1
    case pending = 10
    case archived = 11
}
```

Also works for `String` enums:

### Before

```swift
enum Direction: String {
    case north
    case south = "down"
    case east
}
```

### After

```swift
enum Direction: String {
    case north = "north"
    case south = "down"
    case east = "east"
}
```

## Details

- Supports `Int`, `Int8`, `Int16`, `Int32`, `Int64`, `UInt`, `UInt8`, `UInt16`, `UInt32`, `UInt64`, and `String` raw value types
- Follows Swift's implicit numbering rules: starts at 0, increments from the last explicit value
- Only offers the action when at least one case is missing an explicit raw value

Resolves #2516

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
+++ b/Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
@@ -10,8 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(SourceKitLSP) import LanguageServerProtocol
-import SourceKitLSP
+import SwiftRefactor
 import SwiftSyntax
 
 /// Syntactic code action provider to add explicit raw values to enum cases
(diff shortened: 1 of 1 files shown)


Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,145 @@
...
+  case string
+}
+
+private extension EnumDeclSyntax {
+  /// Determine the raw value type if it's `Int` or `String`.
+  var rawValueType: RawValueKind? {
+    guard let inheritanceClause = self.inheritanceClause else {
+      return nil
+    }
+    for inheritance in inheritanceClause.inheritedTypes {
```
ahoppen: Only the first type in the inheritance clause may specify a raw value, so we can just look at `inheritanceClause.inheritedTypes.first` here.

Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,145 @@
...
+    // Build the edits.
+    var edits: [TextEdit] = []
+
+    switch rawValueType {
+    case .int:
+      var nextValue = 0
+      for element in caseElements {
+        if let rawValue = element.rawValue {
+          // Parse the existing raw value to determine the next implicit value.
+          let rawText = rawValue.value.description.filter { !$0.isWhitespace }
```
ahoppen: Could you check if the value is an `IntegerLiteralExprSyntax` here? Also, if there is a raw value but we don’t understand it, we should fail the refactoring action instead of potentially generating code that’s doesn’t match the pre-refactoring semantics.

Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,145 @@
...
+///
+/// ```swift
+/// enum Status: Int {
+///     case active = 0
+///     case inactive = 1
+///     case pending = 10
+///     case archived = 11
+/// }
+/// ```
+struct AddExplicitEnumRawValues: SyntaxCodeActionProvider {
```
ahoppen: Could you implement this as a `SyntaxRefactoringCodeActionProvider`?

Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,145 @@
...
+  /// Determine the raw value type if it's `Int` or `String`.
+  var rawValueType: RawValueKind? {
+    guard let inheritanceClause = self.inheritanceClause else {
+      return nil
+    }
+    for inheritance in inheritanceClause.inheritedTypes {
+      let typeName = inheritance.type.trimmedDescription
+      switch typeName {
+      case "Int", "Int8", "Int16", "Int32", "Int64",
+        "UInt", "UInt8", "UInt16", "UInt32", "UInt64":
```
ahoppen: We have `Int128` and `UInt128` as well now 😉

Sources/SwiftLanguageService/CodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,145 @@
...
+        let insertionText = " = \"\(element.name.text)\""
+        let position = scope.snapshot.position(
+          of: element.name.endPositionBeforeTrailingTrivia
+        )
+        edits.append(TextEdit(range: position..<position, newText: insertionText))
+      }
+    }
+
+    guard !edits.isEmpty else {
+      return []
```
ahoppen: If there are no cases to transform we should throw a `RefactoringNotApplicableError` so the refactoring action doesn’t show up as a code action. Also applies to the other `return []`.

william-laverty:
Closing — no longer contributing to swiftlang repos. Thanks for the reviews!

Issue #2515, opened 2026-02-28
Add Sort imports code action


Tabonx:
### Description

Alphabetically sort all `import` statements.

**Before:**

```swift
import UIKit
import Foundation
import Combine
import SwiftUI
```

**After:**

```swift
import Combine
import Foundation
import SwiftUI
import UIKit
```

`textDocument/format` will invoke `swift-format` which handles import sorting, but there is no code action to trigger just sorting imports on demand.

ahoppen:
We need to ensure that any sorting we do here matches the formatting done by swift-format. We should be able to just copy-paste the code for that if it’s not too much.

rintaro:
This should be covered by `textDocument/formatting` or `textDocument/rangeFormatting` rather than a code action. Closing

PR #2537, opened 2026-03-05
Add Sort imports code action 

amanthatdoescares:
## Add Sort imports code action (#2515)

Adds an LSP code action that alphabetically sorts the leading block of `import` statements. Matches the import order produced by swift-format.

### Behavior

- **Trigger:** Cursor or selection overlaps the import block at the top of the file.
- **Scope:** Only the contiguous leading imports (stops at the first non-import statement).
- **Action:** Replaces that block with the same imports in alphabetical order by path.
- **Skip:** Not offered when there is only one import or when imports are already sorted.

### Example

**Before:**
```
import UIKit
import Foundation
import Combine
import SwiftUI
```

**After:**
```
import Combine
import Foundation
import SwiftUI
import UIKit

```
#### Implementation

New: Sources/SwiftLanguageService/CodeActions/SortImports.swift 
syntactic provider that collects the leading import block from SourceFileSyntax.statements, 
checks overlap with the request range and that there are ≥2 imports and they are unsorted, 
then returns a single “Sort imports” refactor that replaces the block with sorted imports (by path.trimmedDescription).

#### Registration: 
Provider added in SyntaxCodeActions.swift and CMakeLists.txt.

#### Tests: 
`testSortImports` and `testSortImportsNotOfferedWhenOnlyOneImport` in CodeActionTests.swift.

feature: #2515

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/SortImports.swift
+++ b/Sources/SwiftLanguageService/CodeActions/SortImports.swift
@@ -11,29 +11,16 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(SourceKitLSP) import LanguageServerProtocol
+import Foundation
 import SourceKitLSP
 import SwiftSyntax
 
-/// Code action that alphabetically sorts the leading block of import statements.
+/// Code action that sorts the leading block of import statements to match swift-format's
+/// OrderedImports rule: lexicographic order within groups, with groups 1) regular imports,
+/// 2) declaration imports, 3) @_implementationOnly imports, 4) @testable imports (each
+/// group separated by a blank line). Preserves file header comments and per-import comments.
 ///
-/// Matches the import order produced by swift-format. Only considers the contiguous
-/// import block at the top of the file (stops at the first non-import statement).
-///
-/// ## Example
-/// Before:
-/// ```swift
-/// import UIKit
-/// import Foundation
-/// import Combine
-/// import SwiftUI
-/// ```
-/// After:
-/// ```swift
-/// import Combine
-/// import Foundation
-/// import SwiftUI
-/// import UIKit
-/// ```
+/// See: https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/Rules/OrderedImports.swift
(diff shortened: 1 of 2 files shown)


rintaro:
As @ahoppen states in the issue https://github.com/swiftlang/sourcekit-lsp/issues/2515#issuecomment-3983285161, this code action should align with what swift-format does, which is:
https://github.com/swiftlang/swift-format/blob/7ef767c7cbedff519c799233afe505abface2ef7/Sources/SwiftFormat/Rules/OrderedImports.swift#L15-L18
```
/// Imports must be lexicographically ordered and logically grouped at the top of each source file.
/// The order of the import groups is 1) regular imports, 2) declaration imports, 3) @_implementationOnly
/// imports, and 4) @testable imports. These groups are separated by a single blank line. Blank lines in
/// between the import declarations are removed.
```

Also it needs to preserve the file header comments. e.g.
```swift
// This is part of the file header.

// So is this.

// Top comment
import Bimport
import Aimport
```
should be reordered as:
```swift
// This is part of the file header.

// So is this.

import Aimport
// Top comment
import Bimport
```
as swift-format does https://github.com/swiftlang/swift-format/blob/7ef767c7cbedff519c799233afe505abface2ef7/Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift#L223-L260

amanmaurya92:
Thanks for the pointer to swift-format’s OrderedImports rule. I’ve updated the implementation to align with it.

amanmaurya92:
@swift-ci please test

ahoppen:
Hmm, this has quickly become a little more complex than I would have thought. @rintaro Do you think we should restructure swift-format in a way so we’re able to re-use the order import logic here, which would also mean adding a dependency from SourceKit-LSP to swift-format, which we currently don’t have, but at least ensuring that the ordering doesn’t diverge between the two tools? Or should we just copy-paste it here?

rintaro:
We decided not to accept this code action https://github.com/swiftlang/sourcekit-lsp/issues/2515#issuecomment-4033322688

PR #2542, opened 2026-03-06
Add "Flip range expression" code action (fixes #2521)

### LINKED ISSUE #2521: Add Flip range expression code action
### Description

Reverse the bounds of a range expression.

**Before:**

```swift
for i in stride(from: 0, to: 10, by: 1) {
    print(i)
}
```

**After:**

```swift
for i in stride(from: 10, to: 0, by: -1) {
    print(i)
}
```

Also for simple ranges:

**Before:**

```swift
let range = 1...10
```

**After:**

```swift
let range = (1...10).reversed()
```

amanthatdoescares:
### Summary
Adds a syntactic code action **"Flip range expression"** that reverses the direction of range/stride expressions so they iterate backward.

### Behavior
| Expression | Result |
|------------|--------|
| `stride(from: 0, to: 10, by: 1)` | `stride(from: 10, to: 0, by: -1)` |
| `1...10` | `(1...10).reversed()` |
| `1..<10` | `(1..<10).reversed()` |

The action is offered when the selection overlaps the stride call or the range expression (`..<` or `...`).

### Implementation
- **New:** `Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift`
  - Conforms to `SyntaxCodeActionProvider`.
  - **Stride:** Finds `FunctionCallExprSyntax` for `stride(from:to:by:)`, swaps `from`/`to`, and negates `by` (handles integer/float literals and wraps other expressions in `-(...)` when needed).
  - **Ranges:** Supports both representations used by SwiftSyntax:
    - `InfixOperatorExprSyntax` (after operator folding).
    - `SequenceExprSyntax` with three elements (left, `BinaryOperatorExpr` `..<`/`...`, right) when no folding is applied (e.g. single-file or no workspace).
  - Uses the token at the start of the selection to find the right expression and avoid including the next token when the range ends on the expression boundary.
- **Registration:** `FlipRangeExpression.self` added to `allSyntaxCodeActions` in `SyntaxCodeActions.swift`; file added to `SwiftLanguageService/CMakeLists.txt`.

### Tests
- `testFlipRangeExpressionStride` – stride flip.
- `testFlipRangeExpressionClosedRange` – `1...10` → `(1...10).reversed()`.
- `testFlipRangeExpressionHalfOpenRange` – `1..<10` → `(1..<10).reversed()`.

Tests use a shared helper `assertFlipRangeCodeAction(_:expectedNewText:...)` and follow existing code-action test style (unlabeled first argument, multiline strings, `testName: #function`).

### Notes
- Range expressions are offered in both folded (`InfixOperatorExprSyntax`) and unfolded (`SequenceExprSyntax`) forms so the action works in standalone files and in full projects.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
+++ b/Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
@@ -21,20 +21,28 @@ import SwiftSyntax
 /// - `stride(from: 0, to: 10, by: 1)` → `stride(from: 10, to: 0, by: -1)`
 /// - `1...10` → `(1...10).reversed()`
 /// - `1..<10` → `(1..<10).reversed()`
+///
+/// Implemented as `SyntaxCodeActionProvider` (not `SyntaxRefactoringCodeActionProvider`)
+/// because there is no corresponding refactoring in SwiftRefactor; adding one would require
+/// a new refactoring in the swift-syntax package.
 struct FlipRangeExpression: SyntaxCodeActionProvider {
   static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
     // Use the token at the start of the range so we find the expression that contains
     // the selection even when the range end is exactly at the expression boundary
     // (otherwise the right token can be the next token and the common ancestor is too high).
-    guard let startToken = scope.file.token(at: scope.range.lowerBound) else {
-      return []
+    if let startToken = scope.file.token(at: scope.range.lowerBound) {
+      let node = Syntax(startToken)
+      if let action = flipStride(scope: scope, from: node) {
+        return [action]
+      }
+      if let action = flipRangeOperator(scope: scope, from: node) {
+        return [action]
+      }
     }
-    let node = Syntax(startToken)
-
-    if let action = flipStride(scope: scope, from: node) {
-      return [action]
-    }
-    if let action = flipRangeOperator(scope: scope, from: node) {
+    // Fallback: selection may have made the common ancestor the range or a tuple containing it.
+    if let inner = scope.innermostNodeContainingRange,
+      let action = flipRangeOperator(scope: scope, from: inner)
+    {
       return [action]
     }
     return []
@@ -86,23 +94,17 @@ private func findArgument(label: String, in arguments: LabeledExprListSyntax) ->
   arguments.first { $0.label?.text == label }
 }
 
-/// Returns the source text for the negated stride step (e.g. "1" → "-1", "x" → "-(x)").
+/// Returns the source text for the negated stride step. Toggles a leading "-" when present
+/// so that applying the action twice restores the original (e.g. "1" → "-1", "-1" → "1").
+/// For other expressions, prepends "-" (with parens if needed for precedence).
 private func negateStrideStepText(_ expr: ExprSyntax) -> String {
   let inner = expr.lookingThroughParentheses
+  let text = inner.description
 
-  if let intLit = inner.as(IntegerLiteralExprSyntax.self) {
-    let raw = intLit.literal.text.filter { $0 != "_" }
-    if let value = Int(raw, radix: 10) {
-      return (-value).description
-    }
+  // Add "-" if there isn't one, remove it if there is (no Int/float conversion).
+  if text.hasPrefix("-") {
+    return String(text.dropFirst(1).drop(while: { $0.isWhitespace }))
   }
-
-  if let floatLit = inner.as(FloatLiteralExprSyntax.self),
-    let value = Double(floatLit.literal.text)
-  {
-    return (-value).description
-  }
-
   let exprText = expr.description
   if expr.needParensForPrefixMinus {
     return "-(\(exprText))"
@@ -118,12 +120,17 @@ private extension ExprSyntax {
     return self
   }
 
+  /// True if the expression needs parentheses when wrapped in a prefix "-" to preserve meaning.
+  /// Uses a whitelist of expression kinds that are safe without parens; errs on the side of adding parens.
   var needParensForPrefixMinus: Bool {
     switch self.kind {
-    case .infixOperatorExpr, .sequenceExpr, .ternaryExpr, .binaryOperatorExpr:
-      return true
-    default:
+    case .integerLiteralExpr, .floatLiteralExpr, .booleanLiteralExpr, .nilLiteralExpr,
+      .stringLiteralExpr, .regexLiteralExpr, .declReferenceExpr, .memberAccessExpr,
+      .functionCallExpr, .subscriptCallExpr, .optionalChainingExpr, .arrayExpr,
+      .dictionaryExpr, .tupleExpr, .keyPathExpr, .macroExpansionExpr, .superExpr:
       return false
+    default:
+      return true
     }
   }
 }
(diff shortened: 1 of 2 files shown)


Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+import SwiftExtensions
+import SwiftSyntax
+
+/// A code action that flips range expressions: reverses the bounds of a stride
+/// or wraps a range in `.reversed()`.
+///
+/// - `stride(from: 0, to: 10, by: 1)` → `stride(from: 10, to: 0, by: -1)`
+/// - `1...10` → `(1...10).reversed()`
+/// - `1..<10` → `(1..<10).reversed()`
+struct FlipRangeExpression: SyntaxCodeActionProvider {
```
ahoppen, line 28: Could you implement this as a `SyntaxRefactoringCodeActionProvider`? Or is that not possible?

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+      ]
+    )
+  )
+}
+
+private func isStrideCall(_ call: FunctionCallExprSyntax) -> Bool {
+  guard let ref = call.calledExpression.as(DeclReferenceExprSyntax.self) else {
+    return false
+  }
+  return ref.baseName.text == "stride"
```
ahoppen: You should be able to just inline this as

```suggestion
  call.calledExpression.as(DeclReferenceExprSyntax.self)?.baseName.text == "stride"
```

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Returns the source text for the negated stride step (e.g. "1" → "-1", "x" → "-(x)").
+private func negateStrideStepText(_ expr: ExprSyntax) -> String {
+  let inner = expr.lookingThroughParentheses
+
+  if let intLit = inner.as(IntegerLiteralExprSyntax.self) {
+    let raw = intLit.literal.text.filter { $0 != "_" }
+    if let value = Int(raw, radix: 10) {
+      return (-value).description
```
ahoppen: Do we need to go through an `Int` conversion here? Can’t we just add a `-` if there isn’t one and remove the `-` if there currently is one?

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+    let value = Double(floatLit.literal.text)
+  {
+    return (-value).description
+  }
+
+  let exprText = expr.description
+  if expr.needParensForPrefixMinus {
+    return "-(\(exprText))"
+  }
+  return "-\(exprText)"
```
ahoppen, line 112: Should we also transform `-1` to `1` here so that applying the refactoring action twice results in the original code?

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+  var lookingThroughParentheses: ExprSyntax {
+    if let tuple = self.as(TupleExprSyntax.self), let single = tuple.singleElementExpression {
+      return single.lookingThroughParentheses
+    }
+    return self
+  }
+
+  var needParensForPrefixMinus: Bool {
+    switch self.kind {
+    case .infixOperatorExpr, .sequenceExpr, .ternaryExpr, .binaryOperatorExpr:
```
ahoppen: I would prefer to flip this around and have a whitelist of expressions that don’t need parentheses. Adding parentheses is always the safe option, so I would prefer to error on the side of caution if in doubt.

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
```diff
@@ -0,0 +1,211 @@
...
+// MARK: - Range operator
+
+/// Range expressions like `1..<10` or `1...10` can appear as either
+/// `InfixOperatorExprSyntax` (after operator folding) or `SequenceExprSyntax`
+/// with three elements [left, BinaryOperatorExpr(..< or ...), right].
+private func flipRangeOperator(scope: SyntaxCodeActionScope, from node: Syntax) -> CodeAction? {
+  if let action = flipRangeInfix(scope: scope, from: node) {
+    return action
+  }
+  return flipRangeSequence(scope: scope, from: node)
```
ahoppen: If you run `SwiftOperators` on the syntax tree, you could guarantee that we always have a `BinaryOperatorExpr`.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1333,6 +1333,68 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    try await assertFlipRangeCodeAction(
+      "let range = 1️⃣1...102️⃣",
+      expectedNewText: "(1...10).reversed()"
+    )
+  }
+
+  func testFlipRangeExpressionHalfOpenRange() async throws {
+    try await assertFlipRangeCodeAction(
+      "let range = 1️⃣1..<102️⃣",
+      expectedNewText: "(1..<10).reversed()"
```
ahoppen, line 1398: Open question: Should we drop `.reversed()` when invoking the code action here, so we end up with `1..<10` again?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1333,6 +1333,68 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
   }
 
+  func testFlipRangeExpressionStride() async throws {
+    try await assertFlipRangeCodeAction(
+      """
+      for i in 1️⃣stride(from: 0, to: 10, by: 1)2️⃣ {
+          print(i)
+      }
+      """,
+      expectedNewText: "stride(from: 10, to: 0, by: -1)"
```
ahoppen, line 1341: Could we test that invoking the code action on this result ends up with the original text again?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1333,6 +1333,68 @@ final class CodeActionTests: SourceKitLSPTestCase {
     )
   }
 
+  func testFlipRangeExpressionStride() async throws {
+    try await assertFlipRangeCodeAction(
+      """
+      for i in 1️⃣stride(from: 0, to: 10, by: 1)2️⃣ {
+          print(i)
```
ahoppen: We don’t care about the body of the for loop, so you can just remove it here.

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
amanmaurya92: There’s no “flip range” refactoring in SwiftRefactor, that's why i did not used SyntaxRefactoringCodeActionProvider

Tests/SourceKitLSPTests/CodeActionTests.swift
amanmaurya92: i will add a new test for that, if that's okay

Tests/SourceKitLSPTests/CodeActionTests.swift
amanmaurya92: yes, i have added a test for that now

Sources/SwiftLanguageService/CodeActions/FlipRangeExpression.swift
amanmaurya92: @ahoppen should i implement this as SyntaxRefactoringCodeActionProvider ? or keep it as it is?

rintaro:
We decided not to accept this code action https://github.com/swiftlang/sourcekit-lsp/issues/2521#issuecomment-4033258734

PR #2544, opened 2026-03-07
Add Generate enum associated value accessors code action

### LINKED ISSUE #2522: Add Generate enum associated value accessors code action
### Description

Generate computed properties to extract associated values from enum cases.

**Before:**

```swift
enum Value {
    case text(String)
    case number(Int)
}
```

**After:**

```swift
enum Value {
    case text(String)
    case number(Int)

    var asText: String? {
        if case let .text(v) = self { return v }
        return nil
    }

    var isText: Bool {
        if case .text = self { return true }
        return false
    }

    var asNumber: Int? {
        if case let .number(v) = self { return v }
        return nil
    }

    var isNumber: Bool {
        if case .number = self { return true }
        return false
    }
}
```

See also: [rust-analyzer `generate_enum_as_method`](https://rust-analyzer.github.io/book/assists.html#generate_enum_as_method)

william-laverty:
### Description

Adds a syntax-based code action that generates computed properties for enums with associated values.

For each case with associated values, generates:
- `asX: T?` — extracts the associated value via pattern matching, or `nil` for other cases
- `isX: Bool` — returns `true` if the value matches the case

**Before:**
```swift
enum Value {
    case text(String)
    case number(Int)
}
```

**After:**
```swift
enum Value {
    case text(String)
    case number(Int)

    var asText: String? {
        if case let .text(v) = self { return v }
        return nil
    }

    var isText: Bool {
        if case .text = self { return true }
        return false
    }

    var asNumber: Int? {
        if case let .number(v) = self { return v }
        return nil
    }

    var isNumber: Bool {
        if case .number = self { return true }
        return false
    }
}
```

Handles single and multiple associated values (tuples). Skips generation for accessors that already exist. Inspired by [rust-analyzer's `generate_enum_as_method`](https://rust-analyzer.github.io/book/assists.html#generate_enum_as_method).

### Tests

- Test for enum with two cases having associated values
- Test that no action is offered for enums without associated values

Resolves #2522

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
+++ b/Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
@@ -1,0 +1,157 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftSyntax
+
+/// A code action that generates computed properties to extract associated
+/// values and check cases for an enum.
+///
+/// For each case with associated values, generates:
+/// - `asX: T?` — extracts the associated value, or `nil` if a different case
+/// - `isX: Bool` — returns `true` if the value matches that case
+///
+/// Example:
+/// ```swift
+/// enum Value {
+///     case text(String)
+///     case number(Int)
+/// }
+/// ```
+/// Generates `asText`, `isText`, `asNumber`, `isNumber` computed properties.
+struct GenerateEnumAssociatedValueAccessors: SyntaxCodeActionProvider {
+  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard let node = scope.innermostNodeContainingRange else {
+      return []
+    }
+
+    guard let enumDecl = node.findParentOfSelf(
+      ofType: EnumDeclSyntax.self,
+      stoppingIf: { _ in false }
+    ) else {
+      return []
+    }
+
+    // Collect all cases with associated values.
+    let casesWithAssociatedValues = enumDecl.memberBlock.members.compactMap { member -> EnumCaseElementSyntax? in
+      guard let caseDecl = member.decl.as(EnumCaseDeclSyntax.self),
+            let element = caseDecl.elements.first,
+            caseDecl.elements.count == 1,
+            element.parameterClause != nil
+      else {
+        return nil
+      }
+      return element
+    }
+
+    if casesWithAssociatedValues.isEmpty {
+      return []
+    }
+
+    // Scan existing member names to avoid duplicates.
+    let existingMembers = Set(
+      enumDecl.memberBlock.members.compactMap { member -> String? in
+        guard let varDecl = member.decl.as(VariableDeclSyntax.self),
+              let binding = varDecl.bindings.first,
+              let pattern = binding.pattern.as(IdentifierPatternSyntax.self)
+        else {
+          return nil
+        }
+        return pattern.identifier.text
+      }
+    )
+
+    var accessors: [String] = []
+
+    for element in casesWithAssociatedValues {
+      let caseName = element.name.text
+      let capitalizedName = caseName.prefix(1).uppercased() + caseName.dropFirst()
+      let asName = "as\(capitalizedName)"
+      let isName = "is\(capitalizedName)"
+
+      guard let paramClause = element.parameterClause else { continue }
+      let params = Array(paramClause.parameters)
+
+      if params.count == 1 {
+        let typeText = params[0].type.trimmedDescription
+
+        if !existingMembers.contains(asName) {
+          accessors.append(
+            """
+                var \(asName): \(typeText)? {
+                    if case let .\(caseName)(v) = self { return v }
+                    return nil
+                }
+            """
+          )
+        }
+      } else {
+        let tupleTypes = params.map { $0.type.trimmedDescription }
+        let returnType = "(\(tupleTypes.joined(separator: ", ")))"
+        let bindingVars = (0..<params.count).map { "v\($0)" }
+        let bindingPattern = bindingVars.joined(separator: ", ")
+
+        if !existingMembers.contains(asName) {
+          accessors.append(
+            """
+                var \(asName): \(returnType)? {
+                    if case let .\(caseName)(\(bindingPattern)) = self { return (\(bindingPattern)) }
+                    return nil
+                }
+            """
+          )
+        }
+      }
+
+      if !existingMembers.contains(isName) {
+        accessors.append(
+          """
+              var \(isName): Bool {
+                  if case .\(caseName) = self { return true }
+                  return false
+              }
+          """
+        )
+      }
+    }
+
+    if accessors.isEmpty {
+      return []
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+      ofType: EnumDeclSyntax.self,
+      stoppingIf: { _ in false }
+    ) else {
+      return []
+    }
+
+    // Collect all cases with associated values.
+    let casesWithAssociatedValues = enumDecl.memberBlock.members.compactMap { member -> EnumCaseElementSyntax? in
+      guard let caseDecl = member.decl.as(EnumCaseDeclSyntax.self),
+            let element = caseDecl.elements.first,
```
ahoppen, line 48: Could you run swift-format on your PR as described in CONTRIBUTING.md

ahoppen, line 48: We have `.only` for this

```suggestion
            let element = caseDecl.elements.only,
```

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+///
+/// Example:
+/// ```swift
+/// enum Value {
+///     case text(String)
+///     case number(Int)
+/// }
+/// ```
+/// Generates `asText`, `isText`, `asNumber`, `isNumber` computed properties.
+struct GenerateEnumAssociatedValueAccessors: SyntaxCodeActionProvider {
```
ahoppen, line 32: Would it be possible to implement this as a `SyntaxRefactoringCodeActionProvider`?

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard let node = scope.innermostNodeContainingRange else {
+      return []
+    }
+
+    guard let enumDecl = node.findParentOfSelf(
+      ofType: EnumDeclSyntax.self,
+      stoppingIf: { _ in false }
```
ahoppen, line 40: We should stop at a `CodeBlockSyntax` same as in other code actions so we don’t offer this while the cursor is inside a function implementation. We should probably also stop at `DeclSyntax` so we don’t offer this while you’re in a nested type.

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+            caseDecl.elements.count == 1,
+            element.parameterClause != nil
+      else {
+        return nil
+      }
+      return element
+    }
+
+    if casesWithAssociatedValues.isEmpty {
+      return []
```
ahoppen, line 58: You should throw a `RefactoringNotApplicableError` if the refactoring action doesn’t apply because that will hide it in SourceKit-LSP.

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+    for element in casesWithAssociatedValues {
+      let caseName = element.name.text
+      let capitalizedName = caseName.prefix(1).uppercased() + caseName.dropFirst()
+      let asName = "as\(capitalizedName)"
+      let isName = "is\(capitalizedName)"
+
+      guard let paramClause = element.parameterClause else { continue }
+      let params = Array(paramClause.parameters)
+
+      if params.count == 1 {
```
ahoppen, line 85: You can use `.only` here, which means that you also don’t need the `[0]` below.

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+                    return nil
+                }
+            """
+          )
+        }
+      } else {
+        let tupleTypes = params.map { $0.type.trimmedDescription }
+        let returnType = "(\(tupleTypes.joined(separator: ", ")))"
+        let bindingVars = (0..<params.count).map { "v\($0)" }
+        let bindingPattern = bindingVars.joined(separator: ", ")
```
ahoppen, line 102: Why can’t this general case also handle the case where there’s a single associated value?

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+                    if case let .\(caseName)(v) = self { return v }
+                    return nil
+                }
+            """
+          )
+        }
+      } else {
+        let tupleTypes = params.map { $0.type.trimmedDescription }
+        let returnType = "(\(tupleTypes.joined(separator: ", ")))"
+        let bindingVars = (0..<params.count).map { "v\($0)" }
```
ahoppen, line 101: We should attempt to pick more descriptive variable names here. If the associated value has a label, we should pick that. Otherwise, I’d use `value2` instead of `v2`.

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,157 @@
...
+        let bindingPattern = bindingVars.joined(separator: ", ")
+
+        if !existingMembers.contains(asName) {
+          accessors.append(
+            """
+                var \(asName): \(returnType)? {
+                    if case let .\(caseName)(\(bindingPattern)) = self { return (\(bindingPattern)) }
```
ahoppen, line 108: We should infer the indentation of the source file (see how other code actions do this) instead of hardcoding it to 4 spaces.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1925,3 +1925,52 @@ private func assertDeMorganTransform(
...
+      [
+        CodeAction(
+          title: "Generate enum associated value accessors",
+          kind: .refactorInline,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["2️⃣"]..<positions["2️⃣"],
+                  newText: "\n    var asText: String? {\n        if case let .text(v) = self { return v }\n        return nil\n    }\n\n    var isText: Bool {\n        if case .text = self { return true }\n        return false\n    }\n\n    var asNumber: Int? {\n        if case let .number(v) = self { return v }\n        return nil\n    }\n\n    var isNumber: Bool {\n        if case .number = self { return true }\n        return false\n    }\n"
```
ahoppen, line 1952: I think this would be a lot more readable as a multi-line string literal.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1925,3 +1925,52 @@ private func assertDeMorganTransform(
...
+  func testGenerateEnumNoAssociatedValues() async throws {
+    try await assertCodeActions(
+      ##"""
+      1️⃣enum Direction {
+          case north
+          case south
+      2️⃣}
+      """##,
+      ranges: [("1️⃣", "2️⃣")],
+      exhaustive: false
```
ahoppen, line 1971: You need to pass `exhaustive: true` if you want to check that a code action doesn’t exist.

Sources/SwiftLanguageService/CodeActions/GenerateEnumAssociatedValueAccessors.swift
ahoppen, line 58: Is there any reason why we shouldn’t generate the `is` accessors for cases without associated values?

PR #2547, opened 2026-03-08
Add Inline temp variable code action

### LINKED ISSUE #2507: Add Inline temp variable code action

william-laverty:
### Description

Add a syntactic code action that inlines a temporary variable by replacing all references to it with its initializer expression, then removing the declaration.

### Changes

- New `InlineTempVariable` code action in `Sources/SwiftLanguageService/CodeActions/`
- Tests for basic inlining, parenthesization of binary expressions, and non-offering for unused variables

### Behavior

The action is offered when the cursor is on a `let`/`var` declaration that:
- Has a single binding with an initializer
- Has no type annotation
- Is referenced in subsequent statements
- Is not used as an `inout` argument or reassigned

Smart parenthesization wraps the inlined expression when it's a binary operation, ternary, `try`/`await`, or cast expression to preserve precedence.

**Before:**
```swift
let basePrice = item.price
```

**After:**
```swift
```

Resolves #2507

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/InlineTempVariable.swift
+++ b/Sources/SwiftLanguageService/CodeActions/InlineTempVariable.swift
@@ -1,0 +1,215 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


ahoppen:
@William-Laverty I appreciate your enthusiasm but you now have 7 open PRs. I would encourage you to leave it at that for now so that you have the time and energy to address review feedback on those.

ahoppen:
Here, again, we already have a PR that implements this.  https://github.com/swiftlang/sourcekit-lsp/pull/2529

PR #2549, opened 2026-03-08
Add codeAction/resolve support

Tabonx:
Mentioned in #1539, `codeAction/resolve` support is missing in sourcekit-lsp. This aims to add that support.

### DIFF
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -256,6 +256,7 @@ package protocol LanguageService: AnyObject, Sendable {
   ) async throws -> DocumentSemanticTokensResponse?
   func colorPresentation(_ req: ColorPresentationRequest) async throws -> [ColorPresentation]
   func codeAction(_ req: CodeActionRequest) async throws -> CodeActionRequestResponse?
+  func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction
   func inlayHint(_ req: InlayHintRequest) async throws -> [InlayHint]
   func inlayHintResolve(_ req: InlayHintResolveRequest) async throws -> InlayHint
   func codeLens(_ req: CodeLensRequest) async throws -> [CodeLens]
@@ -481,6 +482,10 @@ package extension LanguageService {
 
   func codeAction(_ req: CodeActionRequest) async throws -> CodeActionRequestResponse? {
     throw ResponseError.requestNotImplemented(CodeActionRequest.self)
+  }
+
+  func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
+    return req.codeAction
   }
 
   func inlayHint(_ req: InlayHintRequest) async throws -> [InlayHint] {
--- a/Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
@@ -43,19 +43,62 @@ package struct SourceKitLSPCommandMetadata: Codable, Hashable {
   }
 }
 
+/// Metadata injected into CodeAction.data to support routing codeAction/resolve requests.
+package struct CodeActionResolveMetadata: Codable, Hashable {
+  package var textDocument: TextDocumentIdentifier
+  package var underlyingData: LSPAny?
+
+  package init(textDocument: TextDocumentIdentifier, underlyingData: LSPAny? = nil) {
+    self.textDocument = textDocument
+    self.underlyingData = underlyingData
+  }
+}
+
+extension CodeActionResolveMetadata {
+  package init?(fromLSPDictionary dictionary: [String: LSPAny]) {
+    guard
+      case .dictionary(let textDocumentDict)? = dictionary[CodingKeys.textDocument.stringValue],
+      let textDocument = TextDocumentIdentifier(fromLSPDictionary: textDocumentDict)
+    else {
+      return nil
+    }
+    self.init(
+      textDocument: textDocument,
+      underlyingData: dictionary[CodingKeys.underlyingData.stringValue]
+    )
+  }
+
+  package func encodeToLSPAny() -> LSPAny {
+    var dict: [String: LSPAny] = [
+      CodingKeys.textDocument.stringValue: textDocument.encodeToLSPAny()
+    ]
+    if let underlyingData {
+      dict[CodingKeys.underlyingData.stringValue] = underlyingData
+    }
+    return .dictionary(dict)
+  }
+}
+
 extension CodeActionRequest {
   package func injectMetadata(toResponse response: CodeActionRequestResponse?) -> CodeActionRequestResponse? {
-    let metadata = SourceKitLSPCommandMetadata(textDocument: textDocument)
-    let metadataArgument = metadata.encodeToLSPAny()
+    let commandMetadata = SourceKitLSPCommandMetadata(textDocument: textDocument)
+    let commandMetadataArgument = commandMetadata.encodeToLSPAny()
     switch response {
     case .codeActions(var codeActions)?:
       for i in 0..<codeActions.count {
-        codeActions[i].command?.arguments?.append(metadataArgument)
+        codeActions[i].command?.arguments?.append(commandMetadataArgument)
+
+        if codeActions[i].edit == nil && codeActions[i].command == nil {
+          codeActions[i].data = CodeActionResolveMetadata(
+            textDocument: textDocument,
+            underlyingData: codeActions[i].data
+          ).encodeToLSPAny()
+        }
       }
       return .codeActions(codeActions)
     case .commands(var commands)?:
       for i in 0..<commands.count {
-        commands[i].arguments?.append(metadataArgument)
+        commands[i].arguments?.append(commandMetadataArgument)
       }
       return .commands(commands)
     case nil:
--- a/Sources/SwiftLanguageService/SwiftLanguageService.swift
+++ b/Sources/SwiftLanguageService/SwiftLanguageService.swift
@@ -949,6 +949,10 @@ extension SwiftLanguageService {
     }.flatMap { $0 }
   }
 
+  package func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
+    return req.codeAction
+  }
+
   func retrieveRefactorCodeActions(_ params: CodeActionRequest) async throws -> [CodeAction] {
     let additionalCursorInfoParameters: ((SKDRequestDictionary) -> Void) = { skreq in
       skreq.set(self.keys.retrieveRefactorActions, to: 1)
--- a/Sources/ClangLanguageService/ClangLanguageService.swift
+++ b/Sources/ClangLanguageService/ClangLanguageService.swift
@@ -608,6 +608,10 @@ extension ClangLanguageService {
     return try await forwardRequestToClangd(req)
   }
 
+  package func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
+    return try await forwardRequestToClangd(req)
+  }
+
   package func inlayHint(_ req: InlayHintRequest) async throws -> [InlayHint] {
     return try await forwardRequestToClangd(req)
   }
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -776,6 +776,8 @@ extension SourceKitLSPServer: QueueBasedMessageHandler {
       await self.handleRequest(for: request, requestHandler: self.prepareCallHierarchy)
     case let request as RequestAndReply<CodeActionRequest>:
       await self.handleRequest(for: request, requestHandler: self.codeAction)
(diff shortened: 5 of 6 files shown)


Tabonx:
@ahoppen Sorry for pinging you. I'm not sure how GitHub handles draft PRs. Could you please take a look at this and let me know if it's something that would be interesting for the LSP? I've checked, and the VSCode client is already prepared to handle `codeAction/resolve`.

Although I'm not sure about the implementation. It feels like I had to bend a few things to get it working, so I would appreciate a little guidance there.

I tried to get a very basic version of #1320 working with this, and to make it work across files, I had to add a bunch of things to the context. I’m not entirely sure if this is correct. Maybe the whole action shouldn't even be implemented in the LSP since it requires some semantic information, and `codeAction/resolve` should simply get the same scope as `textDocument/codeAction`. But that feels very limiting and would kind of defeat the point of resolving the action lazily, since there wouldn't be much to do.

Sources/SwiftLanguageService/CodeActions/ResolvableCodeActionProvider.swift
```diff
@@ -0,0 +1,46 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import LanguageServerProtocol
+import SourceKitLSP
+
+/// A code action provider that can defer expensive work to `codeAction/resolve`
+protocol ResolvableCodeActionProvider {
+
```
ahoppen: Superfluous newline

Sources/SwiftLanguageService/CodeActions/ResolvableCodeActionProvider.swift
```diff
@@ -0,0 +1,46 @@
...
+struct SyntaxCodeActionResolutionContext: Sendable {
+  let scope: SyntaxCodeActionScope
+  let workspace: Workspace
+  let documentManager: DocumentManager
+  let languageService: any LanguageService
+}
+
+extension ResolvableCodeActionProvider {
+  static func makeResolveData(
```
ahoppen: Could we make this return a type that conforms to `LSPAnyCodable` instead? Then we can also use that to deserialize the data again.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -488,6 +484,10 @@ package extension LanguageService {
     throw ResponseError.requestNotImplemented(CodeActionRequest.self)
   }
 
+  func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
```
ahoppen, line 487: Could you override this in `ClangLanguageService` and just forward it?

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActions.swift
```diff
@@ -40,3 +39,11 @@ let allSyntaxCodeActions: [any SyntaxCodeActionProvider.Type] = {
 let supersededSourcekitdRefactoringActions: Set<String> = [
   "source.refactoring.kind.simplify.long.number.literal"  // Superseded by AddSeparatorsToIntegerLiteral
 ]
+
+extension [any SyntaxCodeActionProvider.Type] {
+  subscript(resolveIdentifier id: String) -> (any ResolvableCodeActionProvider.Type)? {
```
ahoppen: This seems like a very odd subscript to have to me. If we want this, I would make it a method instead.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -976,9 +944,94 @@ extension SwiftLanguageService {
...
+
+    let actions = await allSyntaxCodeActions.concurrentMap { provider in
+      provider.codeActions(in: scope)
     }.flatMap { $0 }
+
+    let clientSupportsResolve =
+      capabilityRegistry.clientCapabilities
+      .textDocument?.codeAction?.resolveSupport != nil
+
+    guard !clientSupportsResolve else {
```
ahoppen: I usually consider `guard` as some kind of early exit in error scenarios. I think this reads better as an `if`.

```suggestion
    if clientSupportsResolve {
```

Sources/SwiftLanguageService/CodeActions/ResolvableCodeActionProvider.swift
ahoppen: This isn’t called from anywhere, right?

Tabonx: Not in the PR. I thought it would be like a function that encodes the ID of the action, range, and file URI as metadata, but I found out that I can't really do that since clangd would not have the `uri` encoded for me to route the request to the language server. So I needed to wrap the returned `data` to add the file URI there. However, the specific language server will still need to encode everything it needs into the `data` to perform the action and determine what action it actually is.

Tabonx: I've removed the `ResolvableCodeActionProvider` from this PR, and we can figure out the design in a later PR.

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
```diff
@@ -43,25 +43,71 @@ package struct SourceKitLSPCommandMetadata: Codable, Hashable {
...
-    let metadataArgument = metadata.encodeToLSPAny()
+    let commandMetadata = SourceKitLSPCommandMetadata(textDocument: textDocument)
+    let commandMetadataArgument = commandMetadata.encodeToLSPAny()
+    let resolveMetadata = CodeActionResolveMetadata(textDocument: textDocument)
     switch response {
     case .codeActions(var codeActions)?:
       for i in 0..<codeActions.count {
-        codeActions[i].command?.arguments?.append(metadataArgument)
+        codeActions[i].command?.arguments?.append(commandMetadataArgument)
+        if codeActions[i].data != nil {
```
rintaro: Do we make a convention where resolvable code actions must always have `data`?
If so, that's fine, but otherwise maybe `codeAction[i].edits == nil && codeAction[i].command == nil`

Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
```diff
@@ -43,25 +43,71 @@ package struct SourceKitLSPCommandMetadata: Codable, Hashable {
...
+  private static func mergeResolveMetadata(
+    _ metadata: CodeActionResolveMetadata,
+    into existingData: LSPAny?
+  ) -> LSPAny {
+    var dict: [String: LSPAny] = [
+      "_sourcekitlsp": metadata.encodeToLSPAny()
+    ]
+    if let existingData {
+      dict["_underlying"] = existingData
+    }
```
rintaro: I don't feel comfortable with the encoding/decoding logic being implemented in separated places.
Instead, how about `CodeActionResolveMetadata` hold the underlying data? I.e.
```swift
struct CodeActionResolveMetadata: Codable {
  var textDocument: TextDocumentIdentifier
  var underlyingData: LSPAny?
}

extension CodeActionResolveMetadata: LSPAnyCodable {
  init?(fromLSPDictionary dictionary: [String: LSPAny]) {
    guard case .dictionary(let textDocumentDict)? = dictionary[CodingKeys.textDocument.stringValue],
      let textDocument = TextDocumentIdentifier(fromLSPDictionary: textDocumentDict)
    else {
      return nil
    }
    self.init(
      textDocument: textDocument,
      underlyingMetadata: dictionary[CodingKeys.underlyingData.stringValue]
    )
  }

  func encodeToLSPAny() -> LSPAny {
    return .dictionary([
      CodingKeys.textDocument.stringValue: textDocument.encodeToLSPAny()
      CodingKeys.underlyingData.stringValue: underlyingData
    ])
  }
}
```
Then in `injectMetadata`:
```swift
codeActions[i].data = CodeActionResolveMetadata(
  textDocument: textDocument,
  underlyingData: codeActions[i].data
).encodeToLSPAny()
```
and in `codeActionResolve(_:)`
```swift
guard
  let data = req.codeAction.data,
  let metadata = CodeActionResolveMetadata(fromLSPAny: data)
else {
  return req.codeAction
}
let uri = metadata.textDocument.uri
...
forwardedReq.codeAction.data = metadata.underlyingData
```

I know you kinda followed `SourceKitLSPCommandMetadata` scheme, but I don't think we need to do that.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
Tabonx: I thought that not having an empty data field was the requirement for a code action to be considered resolvable, so that the client can send it back to the server. If that is not the case, I think it's better to do it the way you suggested. Clangd might not send `data` but would still consider it resolvable.

Sources/SourceKitLSP/SourceKitLSPCommandMetadata.swift
```diff
@@ -90,3 +133,24 @@ extension ExecuteCommandRequest {
     return arguments?.dropLast()
   }
 }
+
+extension CodeAction {
```
rintaro: This might just be a style preference, but I'm not sure this extension is adding much value.

The resolveMetadata accessor isn't trivial, and we don't want to call it multiple times. I'd lean toward inlining the logic in `codeActionResolve(_:)` to make that clearer and keep the control flow explicit.

Tabonx: You are right. I was also double decoding the metadata, which was unnecessary and slightly hidden by that extension.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test Windows

rintaro:
```
 Test Case 'CodeActionTests.testCodeActionResponseCommandMetadataInjection' started at 2026-03-20 23:58:21.555
 /home/build-user/sourcekit-lsp/Tests/SourceKitLSPTests/CodeActionTests.swift:170: error: CodeActionTests.testCodeActionResponseCommandMetadataInjection : XCTAssertEqual failed: ("Optional(LanguageServerProtocol.CodeActionRequestResponse.codeActions([LanguageServerProtocol.CodeAction(title: "1", kind: nil, diagnostics: nil, isPreferred: nil, disabled: nil, edit: nil, command: nil, data: Optional(LanguageServerProtocol.LSPAny.dictionary(["textDocument": LanguageServerProtocol.LSPAny.dictionary(["uri": LanguageServerProtocol.LSPAny.string("file:///a.swift")])]))), LanguageServerProtocol.CodeAction(title: "2", kind: nil, diagnostics: nil, isPreferred: nil, disabled: nil, edit: nil, command: Optional(LanguageServerProtocol.Command(title: "Title", command: "Command", arguments: Optional([LanguageServerProtocol.LSPAny.int(1), LanguageServerProtocol.LSPAny.string("text"), LanguageServerProtocol.LSPAny.double(2.2), LanguageServerProtocol.LSPAny.null, LanguageServerProtocol.LSPAny.dictionary(["sourcekitlsp_textDocument": LanguageServerProtocol.LSPAny.dictionary(["uri": LanguageServerProtocol.LSPAny.string("file:///a.swift")])])]))), data: nil)]))") is not equal to ("Optional(LanguageServerProtocol.CodeActionRequestResponse.codeActions([LanguageServerProtocol.CodeAction(title: "1", kind: nil, diagnostics: nil, isPreferred: nil, disabled: nil, edit: nil, command: nil, data: nil), LanguageServerProtocol.CodeAction(title: "2", kind: nil, diagnostics: nil, isPreferred: nil, disabled: nil, edit: nil, command: Optional(LanguageServerProtocol.Command(title: "Title", command: "Command", arguments: Optional([LanguageServerProtocol.LSPAny.int(1), LanguageServerProtocol.LSPAny.string("text"), LanguageServerProtocol.LSPAny.double(2.2), LanguageServerProtocol.LSPAny.null, LanguageServerProtocol.LSPAny.dictionary(["sourcekitlsp_textDocument": LanguageServerProtocol.LSPAny.dictionary(["uri": LanguageServerProtocol.LSPAny.string("file:///a.swift")])])]))), data: nil)]))") - 
 Test Case 'CodeActionTests.testCodeActionResponseCommandMetadataInjection' failed (0.011 seconds)
```
I think the actual result is fine because `CodeAction(title: "1")` doesn't have `edit` nor  `command`. Please update the test case.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/38
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

Issue #2533, opened 2026-03-03
Add Implementations CodeLens Support


Tabonx:
### Description

It would be great if SourceKit-LSP could support showing the number of implementations as a CodeLens above protocols and methods. This feature would improve navigation and code understanding.

ahoppen:
Just to clarify: What do you mean by *Implementations* here? Function implementations that satisfy a protocol requirement?

Tabonx:
Yes, I think so. Maybe the right word would be conformances in Swift.

For example, when I have:

```swift
protocol Greetable {
    func greet(_ name: String)
}

struct Person: Greetable {
    func greet(_ name: String) {
        print("hello: \(name)")
    }
}
```

The LSP would show references (I’m not sure what the exact term is) for the protocol and its methods/variables:

```swift
// 1 Conformance (Person)
protocol Greetable {
    // 1 Conformance / implementation (Person.greet)
    func greet(_ name: String)
}

struct Person: Greetable {
    func greet(_ name: String) {
        print("hello: \(name)")
    }
}
```

ahoppen:
We should have the information about this in the index (the `overrideOf` relationship). I’ve never really worked with code lenses much and there isn’t really any documentation in the LSP spec what they are intended for. Do you know if other LSP servers use code lenses for functionality like that?

Tabonx:
C# LSP has this feature. It looks like this

<img width="416" height="203" alt="Image" src="https://github.com/user-attachments/assets/21474eca-e92c-45f5-88f3-9bb12e9786e3" />

ahoppen:
Interesting. What happens if you click on the code lens? 11 references sounds more like there are 11 places where this is called instead of how many functions override/implement this.

Tabonx:
Sorry, I just realized it’s not from the C# LSP, but from Rider, which is what I had in mind. When you click on it, it basically opens the same list as "Go to References" or "Go to Implementations"

<img width="399" height="107" alt="Image" src="https://github.com/user-attachments/assets/0fe3456d-6ab1-4c2f-846f-3cca567c9aef" />

ahoppen:
If you want to see references, you can invoke it from the context menu or your editor’s equivalent. I don’t see a lot of value in taking up valuable screen space to give you shortcuts to add buttons for these. What do you think?

Tabonx:
[...]

PR #2552, opened 2026-03-11
Add references code lens for symbol declarations

Tabonx:
Adds a references code lens above symbol declarations by querying the index for reference counts.

Supported declarations: classes, structs, enums, actors, functions, protocols, type aliases, initializers, subscripts, and member level variables.

I have also created an updated vscode swift extension so you can test it on my fork or in the repository: https://github.com/Tabonx/vscode-swift/tree/add-references-code-lens

There is a limitation where, if indexing is still in progress, the lens returns "0 references" instead of not appearing at all. However, I could not find a way to check whether indexing is still in progress.

This is related to #2533.

This is how it looks in VS Code:

image

Depends on https://github.com/swiftlang/swift-tools-protocols/pull/39

### DIFF
--- a/Sources/SwiftLanguageService/SwiftCodeLensScanner.swift
+++ b/Sources/SwiftLanguageService/SwiftCodeLensScanner.swift
@@ -12,12 +12,16 @@
 
 internal import BuildServerIntegration
 import BuildServerProtocol
+internal import IndexStoreDB
 @_spi(SourceKitLSP) import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+import SemanticIndex
 import SourceKitLSP
 import SwiftSyntax
 import ToolchainRegistry
 
-/// Scans a source file for classes or structs annotated with `@main` and returns a code lens for them.
+/// Scans a source file for code lenses including `@main` run/debug actions,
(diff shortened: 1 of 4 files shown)


Sources/SwiftLanguageService/SwiftCodeLensScanner.swift
```diff
@@ -12,132 +12,275 @@
...
+        CodeLens(
+          range: range,
+          command: Command(title: "Debug" + suffix, command: debugCommand, arguments: arguments)
         )
+      )
+    }
+  }
+
+  /// Queries the index for the number of references to a symbol and appends a code lens with the count.
+  private func captureReferenceLens(for nameToken: TokenSyntax, at displayRange: Range<AbsolutePosition>) async {
```
ahoppen: I think invoking a cursor info for every declaration in the file is not acceptable because of the overhead it generates. We would probably need to add a request like `CollectVariableType` to sourcekitd that returns the USRs of all declarations in the file as a batch.

Tabonx: Ok, I'll take a look at sourcekitd to see if it supports something like that. If not, I will create an issue and maybe try to add it if I can figure it out.

PR #2555, opened 2026-03-15
Use Bundle.module when looking up test resources

owenv:
The current implementation makes some assumptions about products directory layout which don't hold when building the tests with --build-system swiftbuild. Instead, use Bundle.module where possible to avoid tying the tests to build system implementation details

### DIFF
--- a/Sources/SKTestSupport/ExternalBuildServerTestProject.swift
+++ b/Sources/SKTestSupport/ExternalBuildServerTestProject.swift
@@ -26,24 +26,13 @@ private let sdkArgs =
 
 /// The path to the INPUTS directory of shared test projects.
 private let skTestSupportInputsDirectory: URL = {
-  #if os(macOS)
-  var resources =
-    productsDirectory
-    .appending(components: "SourceKitLSP_SKTestSupport.bundle", "Contents", "Resources")
-  if !FileManager.default.fileExists(at: resources) {
-    // Xcode and command-line swiftpm differ about the path.
-    resources.deleteLastPathComponent()
-    resources.deleteLastPathComponent()
+  guard let resourceURL = Bundle.module.resourceURL else {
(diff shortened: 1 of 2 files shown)


owenv:
Draft for now because I haven't verified this manually on every platform

owenv:
@swift-ci test

owenv:
Given this passes CI and I tested it locally with Xcode and the SwiftPM CLI, I think it should be safe, and going forward their behavior should be aligned

Sources/SKTestSupport/ExternalBuildServerTestProject.swift
```diff
@@ -26,24 +26,13 @@ private let sdkArgs =
 
 /// The path to the INPUTS directory of shared test projects.
 private let skTestSupportInputsDirectory: URL = {
-  #if os(macOS)
-  var resources =
-    productsDirectory
```
rintaro, line 31: `SwiftPMTestProject.macroPackageManifest` also uses `productsDirectory`. Is it not affected by `-build-system swiftbuild`?

rintaro (commented):
I wondered why we didn't use `Bundle.module` in the first place. But it was because of https://github.com/swiftlang/swift-package-manager/issues/4541. Since it's fixed, I think this change is fine.

Sources/SKTestSupport/ExternalBuildServerTestProject.swift
rintaro: Even if it's not affected, does it make sense to make `productsDirectory` use `Bundle.module.bundleURL` instead of the flimsy `testBundle` lookup?

owenv: It is affected, but resolving that is a little more complicated since it's actually grabbing Swift Syntax object files from the build dir, so I was going to leave it for a follow-up PR

PR #2556, opened 2026-03-16
Explicitly pass --build-system native when preparing using the builtin SwiftPMBuildServer

owenv:
In practice, the builtin preparation functionality in source kit-lsp only supports "--build-system native" as "--build-system swiftbuild" clients are expected to use the out of process BSP instead. Specify this explicitly to avoid issues as we stage in the new default build system backend


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test Windows

PR #2558, opened 2026-03-16
Always return the toolchain from SOURCEKIT_TOOLCHAIN_PATH in ToolchainRegistry.preferredToolchain()

Steffeeen:
Sadly, by default the logged error is not shown when running tests. When `SOURCEKIT_LSP_LOG_LEVEL=error` and  `SOURCEKIT_LSP_FORCE_NON_DARWIN_LOGGER=1` are used the error is shown, but it may be buried between a lot of other errors that are also logged. This is for example the case when running the `InlayHintTests`. 

This could be fixed by having `preferredToolchain()` throw an error. However, this also causes issues as we then need to propagate this error up. I have also tried this, but in `UpdateIndexStoreTaskDescription` we would be forced to use `orLog()` to handle the error thus losing the benefits of throwing an error in the first place.

I also included the two error logging improvements in `ClientPlugin.swift` and `SourceKitD.swift` in this PR.


Steffeeen:
@swift-ci please test

Steffeeen:
@swift-ci please test windows

PR #2565, opened 2026-03-18
Skip tests for the experimental SwiftPM build server if it is not installed in the user's toolchain

ahoppen:
Currently, `testBasicSwiftArgs` fails if the user’s toolchain is not recent enough to have `swift package experimental-build-server`. Skip testing the SwiftPM BSP server integration in this case, similar to what we do in other tests using `SkipUnless`.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

PR #2569, opened 2026-03-20
Always wait for real build settings in `testExperimentalFeaturesPassedToSyntaxTreeManager`

ahoppen:
If the SwiftPM build server is slow to respond, we may get document symbols based on a `SyntaxTreeManager` with fallback build settings, which would not contain the `_test_EverythingUnexpected` experimental feature and which would thus fail.


ahoppen:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #2570, opened 2026-03-22
Fix rpath handling on OpenBSD.

3405691582:
The default works fine unless the binary is actually being installed, in that case, more is required.


3405691582:
@swift-ci please test.

ahoppen:
@swift-ci Please test

PR #2571, opened 2026-03-22
Use SymbolLink.destination directly in MarkdownTitleFinder

Bhushan-Aditya:
While reading through the DocC documentation code, I noticed that
`MarkdownTitleFinder.visitHeading(_:)` was doing unnecessary work to extract
a symbol name from a heading like `# ``MyClass`` `.
The code was calling `symbolLink.plainText` to get the text, and then manually
stripping the surrounding backticks off to recover the actual symbol name.
The problem is that `SymbolLink.plainText` is literally defined as:
    "``\(destination ?? "")``"
So it adds the backticks in the first place. Using `symbolLink.destination` directly gives us the clean symbol name without any string manipulation needed.


rintaro:
@swift-ci Please test

Bhushan-Aditya:
Tested locally on macOS - all 29 DoccDocumentationTests pass with 0 failures (26s).

PR #2573, opened 2026-03-22
Fix testEmptyActor using class instead of actor declaration

Bhushan-Aditya:
Fixes an incorrect declaration in testEmptyActor so it actually tests an actor.

Changes
Bug fix: testEmptyActor was declaring public class Actor instead of public actor Actor.
The test still passed because the symbol name was "Actor" in both cases, but it was effectively testing an empty class instead of an empty actor.
This updates the declaration to use the actor keyword, consistent with testActor above it.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test macOS

Bhushan-Aditya:
Tested locally on macOS - testEmptyActor passes with 0 failures (0.355s).

PR #2579, opened 2026-03-24
Add option to install without building.

3405691582:
This allows the build action to be decoupled from the install action, so packaging systems that expect to be able to do discrete installation steps post-build can do so cheaply, without having to effectively start the build over from scratch.

To match existing behavior, this defaults to being disabled. This was introduced previously in swiftlang/swift-package-manager#9050, and the flag has the same name used there.


rintaro:
@swift-ci Please test

PR #2580, opened 2026-03-24
[ContributorDoc] Add detailed docs for using a local build of SourceKit

Steffeeen:
As I have [struggled](https://forums.swift.org/t/running-sourcekit-lsp-tests-with-local-build-of-sourcekit/85205) a lot with this, I decided to add dedicated documentation for using a local build of SourceKit. As I haven't been able to get tests which use `SwiftPMTestProject` to run I'm also looking for input on that. Ideally, I would like to change the documentation to include general instructions which work for all tests or if that's not easily possible, at least layout the steps to get these tests to work.

I have already worked out that the toolchain needs to have `clang` and `clangd`. The default swift build does not have them in `swift-macosx-arm64/bin`. However, creating symlinks to the Xcode toolchain `clang` and `clangd` works. Another issue is that the swift build does not have the `PackageDescription` module which causes the parsing of the `Package.swift` to fail.

Both of these issues can be fixed by including `--swift-testing`, `--swift-testing-macros`, `--llbuild`, `--llbuild`, `--install-all` in the `build-script` invocation. This creates a `toolchain-macosx-arm64` directory which contains an Xcode toolchain with `clang`, `clangd` and the `PackageDescription`. When using this toolchain, the tests work. However, I have not been able to figure how to get ninja to update this toolchain when I recompile SourceKit.

## New Environment Variable to force in-process Sourcekit

I added a new environment variable `SOURCEKIT_LSP_RUN_SOURCEKIT_IN_PROCESS` to explicitly force SourceKit to be run in-process on macOS. For this to work correctly, we also have to make a small change in the cmake setup in the main Swift repo. https://github.com/swiftlang/swift/blob/main/cmake/modules/SwiftComponents.cmake#L87-L91 needs to be changed to be:
```cmake
if(NOT CMAKE_SYSTEM_NAME MATCHES "Darwin")
  list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-xpc-service")
endif()
```
instead of
```cmake
if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
  list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-inproc")
else()
  list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-xpc-service")
endif()
```

This is because we need `sourcekit-inproc`, even when using XPC. Without this change we don't get a `.framework` directory for `sourcekitdInProc`. I haven't opened a PR for this yet, as I'm unsure if this is the correct place and/or way to fix this.

In my opinion, this is better than forcing this in the `build-script` invocation by adding `--extra-cmake-options='-DSWIFT_SOURCEKIT_USE_INPROC_LIBRARY:BOOL=FALSE'` as the environment variable approach avoids a reconfigure. The `build-script` approach would also need a few changes to different `CMakeLists.txt` to make it work correctly.

I can also move the commit for the environment variable to a new PR, if that's preferred.

### DIFF
--- a/Contributor Documentation/Using a Local Build of SourceKit.md
+++ b/Contributor Documentation/Using a Local Build of SourceKit.md
@@ -1,0 +1,79 @@
+# Using a Local Build of SourceKit
+
+This guide explains how to run SourceKit-LSP against a local build of SourceKit from the Swift monorepo.
+
+Examples in this document use the checkout layout from Swift's [How to Set Up an Edit-Build-Test-Debug Loop](https://github.com/swiftlang/swift/blob/main/docs/HowToGuides/GettingStarted.md), ie.`swift-project/swift` and `swift-project/sourcekit-lsp` as sibling directories.
+This layout is only for convenience.
+`sourcekit-lsp` and `swift` can live in different directories as long as you pass the correct absolute paths.
+
+## 1. Build Swift
+
+Follow the [How to Set Up an Edit-Build-Test-Debug Loop](https://github.com/swiftlang/swift/blob/main/docs/HowToGuides/GettingStarted.md) guide to clone Swift and get the required dependencies. Then, add the following additional flags to the `build-script` invocation: `--swift-testing`, `--swift-testing-macros`, `--llbuild`, `--swiftpm` and `--install-all`.
+These flags are necessary to build the components of Swift that SourceKit-LSP depends on, and to ensure that a toolchain with all the necessary components is created.
+When using the `build-script` command from the guide the entire command becomes:
+
+- macOS:
+
+  ```sh
+  utils/build-script --skip-build-benchmarks \
+    --swift-darwin-supported-archs "$(uname -m)" \
+    --release-debuginfo --swift-disable-dead-stripping \
+    --bootstrapping=hosttools \
+    --swift-testing \
+    --swift-testing-macros \
+    --llbuild \
+    --swiftpm \
+    --install-all
+  ```
+
+- Linux:
+
+  ```sh
+  utils/build-script --release-debuginfo \
+    --swift-testing \
+    --swift-testing-macros \
+    --llbuild \
+    --swiftpm \
+    --install-all
+  ```
+
+## 2. Point SourceKit-LSP to that toolchain
+
+Set the `SOURCEKIT_TOOLCHAIN_PATH` environment variable when running
+tests or launching SourceKit-LSP to point to the `.xctoolchain` directory of the locally built toolchain.
+The path should look something like this: `.../swift-project/build/Ninja-RelWithDebInfoAssert/toolchain-macosx-arm64/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain`.
+You may need to adjust the path based on the build configuration and platform.
+
+Example (assuming `sourcekit-lsp` and `swift` are sibling directories):
+
+```bash
+SOURCEKIT_TOOLCHAIN_PATH="$PWD/../build/Ninja-RelWithDebInfoAssert/toolchain-macosx-arm64/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain" \
+swift test
+```
+
+(you may need to adjust the path based on the build configuration and platform)
+
+## 3. Making Changes to SourceKit
+
+When making changes to SourceKit, you can iterate faster by only rebuilding SourceKit and its dependencies instead of rebuilding the entire Swift toolchain.
+You need to ensure that the toolchain gets updated with the new build artifacts.
+This can be done by running the following command from the `build/Ninja-RelWithDebInfoAssert/swift-macosx-arm64` directory after building Swift with the `build-script` command from step 1:
+
+```bash
+DESTDIR=../toolchain-macosx-arm64 ninja sourcekitd sourcekit-inproc install-sourcekit-xpc-service install-sourcekit-inproc
+```
+
+(you may need to adjust the path based on the build configuration and platform)
+
+## SourceKit modes on macOS
+
+Normally, SourceKit-LSP uses XPC on macOS to communicate with SourceKit, while on Linux SourceKit runs in the same process as SourceKit-LSP.
+However, sometimes it may be desirable to run SourceKit in-process on macOS as well.
+Running SourceKit in-process can be useful when using the `SOURCEKIT_LOGGING` environment variable, as the output will include the log messages from SourceKit itself instead of just the messages from the SourceKit wrapper in SourceKit-LSP.
+Additionally, running SourceKit in-process can be useful for profiling as the SourceKit symbols will be directly visible in the profile.
+
+To force SourceKit-LSP to run SourceKit in process, set the `SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS` environment variable.
+
+```bash
+SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS=1 swift test
+```
--- a/Sources/ToolchainRegistry/Toolchain.swift
+++ b/Sources/ToolchainRegistry/Toolchain.swift
@@ -263,9 +263,14 @@ public final class Toolchain: Sendable {
   /// bin/clang
   ///    /clangd
   ///    /swiftc
-  /// lib/sourcekitd.framework/sourcekitd
-  ///    /libsourcekitdInProc.{so,dylib}
-  ///    /libIndexStore.{so,dylib}
+  /// lib/libIndexStore.{so,dylib}
+  /// ```
+  /// and one of:
+  /// ```
+  /// lib/sourcekitd.framework/sourcekitd (macOS)
+  /// lib/sourcekitdInProc.framework/sourcekitdInProc (macOS, if preferInProcessSourceKitD is true)
+  /// lib/libsourcekitdInProc.{so,dylib} (Linux or macOS, if the options above don't exist)
+  /// lib/sourcekitdInProc.dll (Windows)
   /// ```
   ///
   /// The above directory layout can found relative to `path` in the following ways:
@@ -275,7 +280,7 @@ public final class Toolchain: Sendable {
   ///
   /// If `path` contains an ".xctoolchain", we try to read an Info.plist file to provide the
   /// toolchain identifier, etc.  Otherwise this information is derived from the path.
-  convenience package init?(_ path: URL) {
+  convenience package init?(_ path: URL, preferInProcessSourceKitD: Bool = false) {
     // Properties that need to be initialized
     let identifier: String
     let displayName: String
@@ -360,16 +365,17 @@ public final class Toolchain: Sendable {
         dylibExtension = ".so"
       }
 
-      func findDylib(named name: String, searchFramework: Bool = false) -> URL? {
+      func findDylib(named name: String) -> URL? {
         let libSearchPath = libPath.appending(component: "lib\(name)\(dylibExtension)")
         if FileManager.default.isFile(at: libSearchPath) {
           return libSearchPath
         }
+        #if os(macOS)
         let frameworkPath = libPath.appending(components: "\(name).framework", name)
-        if searchFramework, FileManager.default.isFile(at: frameworkPath) {
+        if FileManager.default.isFile(at: frameworkPath) {
           return frameworkPath
         }
-        #if os(Windows)
+        #elseif os(Windows)
         let binSearchPath = binPath.appending(component: "\(name)\(dylibExtension)")
         if FileManager.default.isFile(at: binSearchPath) {
           return binSearchPath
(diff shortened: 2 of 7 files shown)


rintaro:
I understand having `sourcekitdInProc` is convenient for debugging.

> ```cmake
> if(NOT CMAKE_SYSTEM_NAME MATCHES "Darwin")
>   list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-xpc-service")
> endif()
> ```
> instead of
> ```cmake
> if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
>   list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-inproc")
> else()
>   list(REMOVE_ITEM _SWIFT_DEFAULT_COMPONENTS "sourcekit-xpc-service")
> endif()
> ```
> 
> This is because we need `sourcekit-inproc`, even when using XPC. Without this change we don't get a `.framework` directory for `sourcekitdInProc`. I haven't opened a PR for this yet, as I'm unsure if this is the correct place and/or way to fix this.

Yes this is the correct way to do it. We might want to briefly discuss this internally, but feel free to open a PR for it now.

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -378,9 +379,16 @@ public final class Toolchain: Sendable {
         return nil
       }
 
-      if let sourcekitdPath = findDylib(named: "sourcekitd", searchFramework: true)
-        ?? findDylib(named: "sourcekitdInProc")
-      {
+      let sourceKitdPath: URL? = if ProcessEnv.block["SOURCEKIT_LSP_RUN_SOURCEKIT_IN_PROCESS"] != nil {
```
rintaro: It's called `sourcekitd`, `SourceKitD` or `SOURCEKITD` 🙂
```suggestion
      let sourcekitdPath: URL? = if ProcessEnv.block["SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS"] != nil {
```

Steffeeen:
I have opened https://github.com/swiftlang/swift/pull/88159 for the CMake change.

Sources/ToolchainRegistry/Toolchain.swift
```diff
@@ -378,9 +379,18 @@ public final class Toolchain: Sendable {
...
+          findDylib(named: "sourcekitdInProc", searchFramework: Platform.current == .darwin)
+        } else if let sourcekitdPath = findDylib(named: "sourcekitd", searchFramework: true)
+          ?? findDylib(named: "sourcekitdInProc", searchFramework: Platform.current == .darwin)
+        {
+          sourcekitdPath
+        } else {
+          nil
+        }
+
+      if let sourcekitdPath = sourceKitdPath {
```
rintaro: Can be
```suggestion
      let sourcekitdPath: URL? =
        if Platform.current != .darwin || ProcessEnv.block["SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS"] != nil {
          findDylib(named: "sourcekitdInProc", searchFramework: true)
        } else {
          findDylib(named: "sourcekitd", searchFramework: true)
        }

      if let sourcekitdPath {
```
`searchFramework` doesn't harm much in non-Darwin platforms. I don't think it's worth to change it.

Contributor Documentation/Using a Local Build of SourceKit.md
```diff
@@ -0,0 +1,45 @@
...
+## SourceKit modes on macOS
+
+Normally, SourceKit-LSP uses XPC on macOS to communicate with SourceKit, while on Linux SourceKit runs in the same process as SourceKit-LSP.
+However, sometimes it may be desirable to run SourceKit in-process on macOS as well.
+Running SourceKit in-process can be useful when using the `SOURCEKIT_LOGGING` environment variable, as the output will include the log messages from SourceKit itself instead of just the messages from the SourceKit wrapper in SourceKit-LSP.
+
+To force SourceKit-LSP to run SourceKit in process, set the `SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS` environment variable.
+
+```bash
+SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS=1 swift test
```
rintaro, line 78: Since you mentioned above, is it worth to add `SOURCEKIT_LOGGING=1` here as well?

Contributor Documentation/Using a Local Build of SourceKit.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Tests which use `SwiftPMTestProject`
+
+Tests which use `SwiftPMTestProject` currently cannot be run with a local build of SourceKit.
+This is due to a multitude of factors:
+
+- `SwiftPMTestProject` requires a toolchain that contains `clang` and `clangd`, which are not present in the local build of swift
+- TODO: expand on this with more details about the issues and potential solutions
```
rintaro: (Haven't tried)
Maybe this can be resolved by `build-script -r --install-llvm --install-swift` (which installs `clangd`, `clang`, `swiftc` etc into e.g. `${build_dir}/toolchain-macosx-arm64/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain`.

```bash
SOURCEKIT_TOOLCHAIN_PATH="$PWD/../build/Ninja-RelWithDebInfoAssert/toolchain-macosx-arm64/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr" \
swift test
```

Sources/ToolchainRegistry/Toolchain.swift
Steffeeen: I'm inclined to agree, all other uses are also `searchFramework: true` so we could remove the parameter entirely. However, this code will be called for all paths that are searched for a toolchain, which includes all of `$PATH`. On Linux and Windows we thus may be doing an additional 50-100 `isFile` calls that are unnecessary. But I guess those are negligible.

Maybe, the `SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS` environment variable name should also be a parameter that is passed to the `ToolchainRegistry`, similar to `SOURCEKIT_TOOLCHAIN_PATH`. Otherwise, some of the toolchain registry tests could potentially break if `SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS` is set, but a test creates a test toolchain with `makeToolchain(..., sourcekitdInProc: false)` (which is the default value). All the tests always explicitly set the `environmentVariables` parameter of the `ToolchainRegistry` initializer, presumably to avoid this.

Contributor Documentation/Using a Local Build of SourceKit.md
Steffeeen: I guess it would make sense to note in the docs that one probably wants to also specify `SOURCEKIT_LOGGING` when using `SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS` even tough it is not strictly required.

The `SOURCEKIT_LOGGING` environment variables also does not seem to documented anywhere. In SourceKit-LSP I only found it in my docs and in the swift repo the only reference I found was in the actual code where it is read. It should probably be documented somewhere (in the SourceKit readme?), and we can then also link to that documentation in the SourceKit-LSP docs.

Contributor Documentation/Using a Local Build of SourceKit.md
Steffeeen: Building the toolchain works, I now also finally got ninja to correctly update the toolchain. I currently just set the `DESTDIR` environment variable to point to the toolchain directory before calling ninja:
```bash
DESTDIR=../toolchain-macosx-arm64 ninja sourcekitd sourcekit-inproc install-sourcekit-xpc-service install-sourcekit-inproc
```
But it should probably also be possible to make the `install-sourcekit-xpc-service` and `install-sourcekit-inproc` targets install into the toolchain if a toolchain was created. But I currently don't know how.

Sources/ToolchainRegistry/Toolchain.swift
Steffeeen: I just realized that we can just put the logic for searching for `.framework` inside a `#if os(macOS)` block and remove the `searchFramework` parameter.

Sources/ToolchainRegistry/ToolchainRegistry.swift
Steffeeen, line 143: When this initializer is used, the environment variable would not work currently, as it is not passed to the `Toolchain` initializer. However, as far as I can tell, this initializer is currently never used and could be removed entirely.

Sources/ToolchainRegistry/ToolchainRegistry.swift
```diff
@@ -170,6 +170,7 @@ package final actor ToolchainRegistry {
     xcodes: [URL] = [_currentXcodeDeveloperPath].compactMap({ $0 }),
     libraryDirectories: [URL] = FileManager.default.urls(for: .libraryDirectory, in: .allDomainsMask),
     pathEnvironmentVariables: [ProcessEnvironmentKey] = ["SOURCEKIT_PATH", "PATH"],
+    forceInProcessSourcekitdEnvironmentVariable: ProcessEnvironmentKey? = "SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS",
```
rintaro: Is this because you don't want to pollute the environment vars in the test?
How about making this
```suggestion
    preferInProcessSourceKitD: Bool = ProcessEnv.block["SOURCEKIT_LSP_RUN_SOURCEKITD_IN_PROCESS"] ?? false,
```
and passing `preferInProcessSourceKitD: true` in the test?

Steffeeen: I changed the variable name to be passed as a parameter to allow running the ToolchainRegistry tests even when the environment variable is set. Otherwise, by setting the environment variable some of the tests may fail due the toolchain selecting inproc but the tests expecting the xpc version. This is similar to what is done for the `SOURCEKIT_TOOLCHAIN_PATH` environment variable. I also considered changing the parameter to be a boolean, like you suggested, but decided against that to actually test reading an environment variable in the test. However, I guess in the end this isn't such a big difference. I will change the parameter to be a boolean.

Sources/ToolchainRegistry/ToolchainRegistry.swift
ahoppen: I think that’s fine to not read the environment variable here. And let’s keep the initializer.

ahoppen (approved):
Very nice. Thanks a lot for the detailed instructions @Steffeeen 🙏🏽 Just want to wait for @rintaro’s review before merging this.

Steffeeen:
@rintaro could we get this merged? What are your thoughts on my two inline comments regarding `SOURCEKIT_LOGGING` and rebuilding the toolchain with ninja without manually setting `DESTDIR`?

rintaro (approved):

rintaro:
@swift-ci Please test

rintaro:
@Steffeeen Sorry for delay 🙇‍♂️ this completely slipped my mind

rintaro:
@swift-ci Please test Windows

Steffeeen:
The test failures on Linux and Windows are because of the change of only looking for `.framework` on macOS. For the failing tests `makeToolchain()` creates a toolchain with `sourcekitd.framework` even on Linux and Windows. The tests worked previously, because we always did a `findDylib(named: "sourcekitd", searchFramework: true)` first before falling back to looking for `sourcekitdInProc.dylib/so/dll`.

IMO `makeToolchain` should not create `sourcekitd.framework` on Linux and Windows and rather just fall back to creating `sourcekitdInProc.dylib/so/dll` if the `sourcekitd` is true as that matches what an actual non-macOS toolchain looks like. On the other hand, having `sourcekitd.framework` on Linux and Windows allows testing the `.framework` finding logic on these platforms even though it is only relevant for macOS. What do you think?

rintaro:
Let's just simulate the actual toolchain in the platform.
_Conceptually_ `makeToolchain(..., sourcekitd: true, ...)` should do:
```swift
if sourcekitd{
  #if os(macOS)
  mkdir("sourcekitd.framekwork")
  mkdir("sourcekitdInProc.framekwork")
  #elseif os(Windos)
  touch("sourcekitdInProc\(dylibSuffix)")
  #else
  touch("libsourcekitdInProc\(dylibSuffix)")
  #endif
}
```
And I don't think we need separate `sourcekitdInProc` parameter here. Let me know if any test fails with this change.

Steffeeen:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Tests/ToolchainRegistryTests/ToolchainRegistryTests.swift
```diff
@@ -759,14 +808,18 @@ private func makeToolchain(
...
     )
     try Data().write(to: libPath.appending(components: "sourcekitd.framework", "sourcekitd"))
-  }
-  if sourcekitdInProc {
-    #if os(Windows)
+    try FileManager.default.createDirectory(
+      at: libPath.appending(component: "sourcekitdInProc.framework"),
+      withIntermediateDirectories: true
+    )
+    try Data().write(to: libPath.appending(components: "sourcekitdInProc.framework", "sourcekitdInProc"))
```
rintaro, line 821: `sourcekitdInProc.framework` should be a directory in macOS

Steffeeen: `sourcekitdInProc.framework` is a directory, it is created in the line above. This line just creates the `sourcekitdInProc` file inside that directory. Or am I missing something?

rintaro: Oh I misread. No worries. Lgtm

Steffeeen:
@swift-ci Please test Windows

PR #2588, opened 2026-04-01
Add shared cursorInfo infrastructure and forEach-to-forIn code action

### LINKED ISSUE #2509: Add Convert forEach to for-in loop code action

MoonMao42:
Closes #2584
Closes #2509

Adds `SharedCursorInfo` so code action providers can share a single cursorInfo request. `SyntaxCodeActionProvider.codeActions(in:)` is now async, and the scope object (renamed to `CodeActionScope`) is created once in `codeAction()` and passed to all retrieve methods.

Also adds a code action to convert `.forEach` to `for-in` loops, as the first provider that uses this.

Before:

```swift
names.forEach { name in
  print(name)
}
```

After:

```swift
for name in names {
  print(name)
}
```

Only handles closures with explicit named parameters (`$0` shorthand is not supported).


Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -61,6 +66,20 @@ struct SyntaxCodeActionScope {
...
+    guard let shared = sharedCursorInfo else {
+      return nil
+    }
+    do {
+      let result = try await shared.get()
+      return result.cursorInfo.first
+    } catch {
+      logger.error("cursorInfo request failed: \(error)")
+      return nil
+    }
```
rintaro: Why not just throw it? The function is already `throws`.

This whole function could simply be:
```swift
func cursorInfo() async throws -> CursorInfo? {
  sharedCursorInfo?.get()?.cursorInfo.first
}
```

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -61,6 +66,20 @@ struct SyntaxCodeActionScope {
     self.range = left.position..<right.endPosition
     self.innermostNodeContainingRange = findCommonAncestorOrSelf(Syntax(left), Syntax(right))
   }
+
+  /// Returns the first `CursorInfo` from the shared cache, or `nil` if unavailable.
+  package func cursorInfo() async throws -> CursorInfo? {
```
rintaro: The struct itself is internal. I don't think `package` is needed.

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,44 @@
...
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import SKLogging
+import SourceKitD
+
+/// Lazy, shared cache for a single `cursorInfo` request across concurrent code action providers.
+///
+/// The first call to `get()` triggers the sourcekitd request; subsequent calls await the same `Task`.
+actor SharedCursorInfo {
+  private var cachedTask: Task<(cursorInfo: [CursorInfo], refactorActions: [SemanticRefactorCommand], symbolGraph: String?), any Error>?
```
rintaro: While you are here could you make a dedicated struct for the result tuple?
```swift
struct SourceKitDCursorInfoResponse {
  var cursorInfo: [CursorInfo]
  var refactorActions: [SemanticRefactorCommand]
  var symbolGraph: String?
}
```

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -44,14 +44,19 @@ struct SyntaxCodeActionScope {
   /// The innermost node that contains the entire selected source range
   var innermostNodeContainingRange: Syntax?
 
+  /// Shared lazy cursorInfo cache, if semantic information is available for this request.
+  var sharedCursorInfo: SharedCursorInfo?
```
rintaro: I don't see a case this can be `nil`

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  func get(
+    additionalParameters: ((SKDRequestDictionary) -> Void)? = nil
+  ) async throws -> (cursorInfo: [CursorInfo], refactorActions: [SemanticRefactorCommand], symbolGraph: String?) {
+    if let task = cachedTask {
+      return try await task.value
+    }
+    let task = fetchCursorInfo(additionalParameters)
+    cachedTask = task
```
rintaro: This can cause confusing behaviors depending on how this is used. I don't think this should accept `additionalParameters`

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import SKLogging
+import SourceKitD
+
+/// Lazy, shared cache for a single `cursorInfo` request across concurrent code action providers.
+///
+/// The first call to `get()` triggers the sourcekitd request; subsequent calls await the same `Task`.
+actor SharedCursorInfo {
```
rintaro: I think this whole actor can be generalized as something like:
```swift
actor LazyTask<Success: Sendable> {
  private let operation: @Sendable () async throws -> Success

  init(_ operation: @escaping @Sendable () async throws -> Success) {
    self.operation = operation
  }

  private lazy var task: Task<Success, any Error> = Task {
    try await operation()
  }

  var value: Success {
    get async throws {
      try await task.value
    }
  }
}
```
Maybe we already have something like it?

Tests/SourceKitLSPTests/SharedCursorInfoIntegrationTests.swift
rintaro, line 1: I'm not sure what these new tests (except `ForEachToForInCodeActionTests.swift`) are verifying. Could you elaborate?

rintaro (commented):
Looks pretty good already!

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
MoonMao42: Done.

Tests/SourceKitLSPTests/SharedCursorInfoIntegrationTests.swift
MoonMao42: You're right. The ForEachToForIn tests already cover the infrastructure.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
```
rintaro:
```suggestion
// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
```

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
...
+  let collection: ExprSyntax
+  let closure: ClosureExprSyntax
+}
+
+private func findForEachCall(in scope: CodeActionScope) -> Match? {
+  guard let node = scope.innermostNodeContainingRange else {
+    return nil
+  }
+
+  let visitor = ForEachCallVisitor(rangeToMatch: scope.range)
```
rintaro: Is `SyntaxVisitor` really relevant? For example, for:
```
1️⃣func test() {
  2️⃣foo.bar.3️⃣forEach4️⃣ { item in
    print(item)
  }5️⃣
}6️⃣
```
I think we should only offer the code action when the curosr/selection is `2️⃣..<4️⃣`, `2️⃣..<5️⃣`, `3️⃣..<3️⃣`, `3️⃣..<4️⃣`, and `3️⃣..<5️⃣`. 
I think  using a visitor and returning `.visitChildren` for every failure case makes it available for `1️⃣..<6️⃣` or even for the entire source file.
Please add test cases as needed.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
...
+
+    guard let info = try? await scope.cursorInfo(), isStdlibSequenceForEach(info) else {
+      return []
+    }
+
+    guard !hasReturnWithValue(match.closure) else {
+      return []
+    }
+
+    guard !hasTryOrAwait(match.closure) else {
```
rintaro: Could you explain what is a case we can't convert it to for-in?

I understand having `try?` outside the `forEach` call expression is not easy:
```swift
try? foo.forEach { ...; try ...; }
```
to:
```swift
do {
  for item in foo { ...; try ...; }
} catch { /*ignore*/ }
```
But I'm not sure having `try` in the closure body prevents anything.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
...
+  guard parameters.parameters.count == 1 else {
+    return nil
+  }
+
+  guard let param = parameters.parameters.first else {
+    return nil
+  }
+
+  // Reject if parameter is shorthand $0 syntax
+  if param.firstName.text == "$0" {
```
rintaro: I don't think this handles anonymous closure parameters. `$0` does not appear in the closure parameter syntax tree.

It would be great this also rewrites `$0`. I.e. rewrite
```swift
foo.forEach {
  print($0)
}
```
with
```swift
for element in foo {
  print(element)
}
```
It's not super obvious whether we can use `element` though, we should check if it's not already used in the body.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
...
+private class ReturnToContinueRewriter: SyntaxRewriter {
+  override func visit(_ node: ReturnStmtSyntax) -> StmtSyntax {
+    guard node.expression == nil else {
+      return StmtSyntax(node)
+    }
+    return StmtSyntax(ContinueStmtSyntax())
+  }
+
+  override func visit(_ node: FunctionDeclSyntax) -> DeclSyntax {
+    DeclSyntax(node)
```
rintaro: I don't think we should dig into any non-`StmtSyntax`.
Are there any cases where `return` in an expression or decl escapes the surrounding function scope?

Sources/SwiftSyntaxCodeActions/ForEachToForInCodeAction.swift
rintaro, line 1: Please add this file and `SharedCursorInfo.swift` to https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SwiftLanguageService/CMakeLists.txt

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
```
rintaro:
```suggestion
// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
```

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,238 @@
...
+  visitor.walk(closure)
+  return visitor.found
+}
+
+private func extractParameterName(from closure: ClosureExprSyntax) -> String? {
+  guard let signature = closure.signature else {
+    return nil
+  }
+
+  guard let parameters = signature.parameterClause?.as(ClosureParameterClauseSyntax.self) else {
```
rintaro: What about `ClosureShortParameterSyntax`? Ideally the conversion should be:
`foo.forEach { item in ... }` -> `for item in foo { ... }`
`foo.forEach { (item) in ... }` -> `for item in foo { ... }`
`foo.forEach { (item: Ty) in ... }` -> `for item: Ty in foo { ... }`

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -927,24 +927,56 @@ extension SwiftLanguageService {
...
+    if wantedActionKinds == nil || (wantedActionKinds?.contains(.quickFix) ?? false) {
+      allCodeActions += try await retrieveQuickFixCodeActions(scope)
+    }
+
+    if wantedActionKinds == nil || (wantedActionKinds?.contains(.sourceOrganizeImports) ?? false) {
+      allCodeActions += try await retrieveRemoveUnusedImportsCodeAction(scope)
+    }
+
     let codeActionCapabilities = capabilityRegistry.clientCapabilities.textDocument?.codeAction
-    let codeActions = try await retrieveCodeActions(req, providers: providers)
```
rintaro, line 975: `func retrieveCodeActions(_:providers:)` below is now unused.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -927,24 +927,56 @@ extension SwiftLanguageService {
...
+    ) else {
+      return nil
     }
+
+    var allCodeActions: [CodeAction] = []
+    let wantedActionKinds = req.context.only
+
+    if wantedActionKinds == nil {
+      allCodeActions += await retrieveSyntaxCodeActions(scope)
+    }
```
rintaro, line 1008: I _think_ `retrieveSyntaxCodeActions` was added even if `wantedActionKinds` exist. Am I missing something?

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A lazily-evaluated, shared async value that is computed at most once.
+///
+/// The first access triggers the operation; subsequent accesses await the same `Task`.
+actor LazyValue<Success: Sendable> {
```
rintaro: Could you use a different name?
We already have `LazyValue` with different semantics here https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SwiftExtensions/LazyValue.swift

MoonMao42:
Done, thanks for the review.

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -61,6 +70,18 @@ struct SyntaxCodeActionScope {
...
     self.innermostNodeContainingRange = findCommonAncestorOrSelf(Syntax(left), Syntax(right))
   }
+
+  /// Returns the first `CursorInfo` from the shared cache, or `nil` if unavailable.
+  func cursorInfo() async throws -> CursorInfo? {
+    try await sharedCursorInfo.value.cursorInfo.first
+  }
+
+  /// Returns cursorInfo at the position of the given syntax node.
+  func cursorInfo(at absolutePosition: AbsolutePosition) async throws -> CursorInfo? {
```
rintaro: Ah sorry I didn't think about what information is in the cursor-info result when I was suggesting the ranges to offer actions.

We'd like to avoid requesting non-shared `cursorInfo` in `textDocument/codeAction` request. Could you remove this? And I think it's okay for now to suggest this action only when the cursor is on `forEach`.

I.e. for
```swift
      1️⃣func test() {
        let array = [1, 2, 3]
        2️⃣array.reversed().3️⃣forEach4️⃣ { item in
          print(item)
        }5️⃣
      }6️⃣
```
The action is active only for `3️⃣..<3️⃣` and `3️⃣..<4️⃣`.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,313 @@
...
+}
+
+private func findForEachCall(in scope: CodeActionScope) -> Match? {
+  guard let node = scope.innermostNodeContainingRange else {
+    return nil
+  }
+
+  // Walk up from the innermost node to find a forEach call expression,
+  // stopping at function/closure boundaries to avoid matching distant calls.
+  guard let callExpr = node.findParentOfSelf(
```
rintaro: If it's only active at `forEach`, the check can be something  like:
```swift
guard
  let token = node.as(TokenSyntax.self),
  token.text = "forEach"
  let memberName = token.parent?.as(DeclReferenceExprSyntax.self),
  let memberAccessExpr = member.parent?.as(MemberAccessExprSyntax.self),
  memberAccessExpr.declName.id == memberName.id,
  let callExpr = memberAccessExpr.parent?.as(FunctionCallExprSyntax.self),
  callExpr.calledExpression.id == memberAccessExpr,
else {
  return nil
}

let closure: ClosureExprSyntax?
...
```

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  guard let closure else {
+    return nil
+  }
+
+  return Match(
+    callExpr: callExpr,
+    memberAccess: memberAccess,
+    collection: collection,
```
rintaro: `forEach(_:)` is available on `Sequence`, so could you name this `sequence` instead of `collection`?

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,313 @@
...
+/// Uses `cursorInfo` to verify that `.forEach` refers to `Sequence.forEach` from the
+/// Swift stdlib before suggesting conversion.
+struct ForEachToForInCodeAction: SyntaxCodeActionProvider {
+  static func codeActions(in scope: CodeActionScope) async -> [CodeAction] {
+    guard let match = findForEachCall(in: scope) else {
+      return []
+    }
+
+    let forEachPosition = match.memberAccess.declName.baseName.positionAfterSkippingLeadingTrivia
+    guard let info = try? await scope.cursorInfo(at: forEachPosition),
```
rintaro: And this can be `scope.cursorInfo()`

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,313 @@
...
+      return []
+    }
+
+    guard !hasReturnWithValue(match.closure) else {
+      return []
+    }
+
+    guard !hasAwait(match.closure) else {
+      return []
+    }
```
rintaro: Could you make a single function/visitor checking both? So we can check the eligibility in single pass.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,313 @@
...
+  let name: String
+  let typeAnnotation: TypeAnnotationSyntax?
+  /// Whether the body needs `$0` references rewritten to `name`.
+  let needsDollarZeroRewrite: Bool
+}
+
+private func extractParameter(from closure: ClosureExprSyntax) -> ClosureParam? {
+  guard let signature = closure.signature else {
+    // No signature → anonymous $0 usage.
+    let name = generateUniqueName(avoiding: collectIdentifiers(in: closure.statements))
```
rintaro: I don't think we need to separate function for this.
Also, could you avoid inserting every identifier in the `Set`, we're only interested in identifiers starting with `element`

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
rintaro: I don't see you addressed this? Also applies to other visitors/rewriters. Please avoid dig into unnecessary nodes.

MoonMao42:
Since the `.forEach { ... }` pattern is already quite distinctive syntactically, I'm thinking of dropping the cursorInfo dependency entirely — keep this as a pure `SyntaxCodeActionProvider` (synchronous, no scope changes), which also resolves the trigger range question. If false positives turn out to be an issue later, we can add verification via `codeAction/resolve`. Want to make sure we're on the same page before I rework things again.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
MoonMao42: Addressed — all three visitor/rewriter classes (`ClosureEligibilityVisitor`, `ReturnToContinueRewriter`, `DollarZeroRewriter`) now consistently skip the same five scope-introducing constructs: `FunctionDeclSyntax`, `InitializerDeclSyntax`, `DeinitializerDeclSyntax`, `ClosureExprSyntax`, `AccessorBlockSyntax`.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
MoonMao42: Done — replaced with `selectedForEachToken(in:)` that validates directly against the request range positions.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
MoonMao42: Done — `containsIdentifier(named:in:)` now checks on demand instead of collecting all names.

Sources/SwiftLanguageService/CodeActions/SyntaxCodeActionProvider.swift
MoonMao42: Makes sense — the action now activates only when the cursor/selection is exactly on the `forEach` token, so shared cursorInfo at the request position works.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,298 @@
...
+// MARK: - Syntax Rewriting
+
+/// Rewrites bare `return` statements to `continue` within the closure body,
+/// skipping nested scopes where `return` has different semantics.
+private class ReturnToContinueRewriter: SyntaxRewriter {
+  override func visit(_ node: ReturnStmtSyntax) -> StmtSyntax {
+    guard node.expression == nil else {
+      return StmtSyntax(node)
+    }
+    return StmtSyntax(ContinueStmtSyntax())
```
rintaro: So I realized this is not enough for cases like
```swift
items.forEach { item in
  for elem in item.elements {
    if condition(elem) {
      return
    }
  }
  // ...
}
```
In this case, we need a label on the rewritten for-in:
```swift
ITEM: for item in items {
  for elem in item.elements {
    if condition(elem) {
      continue ITEM
    }
  }
  // ...
}
```
That said, I'm fine with ignoring such cases for this PR. But in that case, could you add `FIXME` and file an issue for that?

Sources/SwiftSyntaxCodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,298 @@
...
+    result.hasAwait = true
+    return result.isEligible ? .visitChildren : .skipChildren
+  }
+
+  // Don't dig into nested scopes.
+  override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind { .skipChildren }
+  override func visit(_ node: InitializerDeclSyntax) -> SyntaxVisitorContinueKind { .skipChildren }
+  override func visit(_ node: DeinitializerDeclSyntax) -> SyntaxVisitorContinueKind { .skipChildren }
+  override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind { .skipChildren }
+  override func visit(_ node: AccessorBlockSyntax) -> SyntaxVisitorContinueKind { .skipChildren }
```
rintaro, line 263: Any type declarations too. i.e. `struct`, `class`, `enum`, etc.

Sources/SwiftSyntaxCodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,298 @@
...
+
+private struct ClosureParam {
+  let name: String
+  let typeAnnotation: TypeAnnotationSyntax?
+  /// Whether the body needs `$0` references rewritten to `name`.
+  let needsDollarZeroRewrite: Bool
+}
+
+private func extractParameter(from closure: ClosureExprSyntax) -> ClosureParam? {
+  guard let signature = closure.signature else {
```
rintaro, line 277: Rare edge case:
```swift
arr.forEach { [] in
  print($0)
}
```
In this case `signature` is non-nil.
Even more edge case:
```swift
arr.forEach { [foo = bar] in
  print($0, foo)
}
```
I guess we should rewrite this like
```swift
do {
  let foo = bar 
  for element in arr {
    print(element, foo)
  }
}
```
But it's too much effort.
I'm fine with just rejecting (not offering the action) for closures with capture list.

Sources/SwiftLanguageService/CodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,298 @@
...
+private struct Match {
+  let callExpr: FunctionCallExprSyntax
+  let sequence: ExprSyntax
+  let closure: ClosureExprSyntax
+}
+
+/// Matches only when the cursor/selection is on the `forEach` token.
+private func findForEachCall(in scope: CodeActionScope) -> Match? {
+  guard let token = selectedForEachToken(in: scope),
+        token.text == "forEach",
```
rintaro: I'm not sure this is what `swift-format` suggests. Could you run `swift format -i -r .` at the package top-level directory? Otherwise the test will fail.

rintaro (commented):
It's getting close!
Just a few nit-picky edge case handling suggestions.

Sources/SwiftSyntaxCodeActions/ForEachToForInCodeAction.swift
```diff
@@ -0,0 +1,298 @@
...
+
+private struct Match {
+  let callExpr: FunctionCallExprSyntax
+  let sequence: ExprSyntax
+  let closure: ClosureExprSyntax
+}
+
+/// Matches only when the cursor/selection is on the `forEach` token.
+private func findForEachCall(in scope: CodeActionScope) -> Match? {
+  guard let token = selectedForEachToken(in: scope),
```
rintaro, line 108: Does `scope.innermostNodeContainingRange` not work?

MoonMao42: `tokenForRefactoring` shifts the cursor to the previous token at exact token boundaries, so `innermostNodeContainingRange` resolves to `.` instead of `forEach` when the cursor is at the start of `forEach`. `selectedForEachToken` bypasses this by matching against the raw request position.

rintaro:
@swift-ci Please test

rintaro:
@MoonMao42 Could you rebase on the current main and fix the build issue?
Also macOS failure indicates the formatting is still off. Try `swift format -i -r .` again right before pushing.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -927,79 +927,73 @@ extension SwiftLanguageService {
...
+    func wantActionKind(_ kind: CodeActionKind) -> Bool {
+      guard let only = wantedActionKinds else { return true }
+      return only.contains(kind)
     }
+
+    async let syntaxActions = retrieveSyntaxCodeActions(scope)
+    async let refactorActions = wantActionKind(.refactor) ? try await retrieveRefactorCodeActions(scope) : []
+    async let quickFixActions = wantActionKind(.quickFix) ? try await retrieveQuickFixCodeActions(scope) : []
+    async let unusedImportActions = wantActionKind(.sourceOrganizeImports)
+      ? try await retrieveRemoveUnusedImportsCodeAction(scope) : []
```
rintaro: Ah, I think `try await` shouldn't be here.

MoonMao42:
@rintaro It's OK?I format it.

rintaro:
@swift-ci Please test

rintaro:
```
[2026-04-23T05:43:05.440Z] /home/build-user/sourcekit-lsp/Sources/SwiftLanguageService/CodeActions/ConvertCommentToDocComment.swift:41:40: error: cannot find type 'SyntaxCodeActionScope' in scope
[2026-04-23T05:43:05.440Z] 39 |   static let title = "Convert Comment to Doc Comment"
[2026-04-23T05:43:05.440Z] 40 | 
[2026-04-23T05:43:05.440Z] 41 |   static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> DeclSyntax? {
[2026-04-23T05:43:05.440Z]    |                                        `- error: cannot find type 'SyntaxCodeActionScope' in scope
[2026-04-23T05:43:05.440Z] 42 |     let cursorPosition = scope.snapshot.absolutePosition(of: scope.request.range.lowerBound)
[2026-04-23T05:43:05.440Z] 43 |     guard let token = scope.file.token(at: cursorPosition) else {
[2026-04-23T05:43:05.440Z] 
[2026-04-23T05:43:05.440Z] /home/build-user/sourcekit-lsp/Sources/SwiftLanguageService/CodeActions/ConvertCommentToDocComment.swift:56:62: error: cannot infer contextual base in reference to member 'sourceAccurate'
[2026-04-23T05:43:05.440Z] 54 |     else { return nil }
[2026-04-23T05:43:05.440Z] 55 | 
[2026-04-23T05:43:05.440Z] 56 |     guard let firstToken = declaration.firstToken(viewMode: .sourceAccurate),
[2026-04-23T05:43:05.440Z]    |                                                              `- error: cannot infer contextual base in reference to member 'sourceAccurate'
[2026-04-23T05:43:05.440Z] 57 |       firstToken.id == token.id
[2026-04-23T05:43:05.440Z] 58 |     else { return nil }
```
I think `ConvertCommentToDocComment.swift` was added after you opened this PR. Please run tests locally again and fix the issues.

MoonMao42:
> ```
> [2026-04-23T05:43:05.440Z] /home/build-user/sourcekit-lsp/Sources/SwiftLanguageService/CodeActions/ConvertCommentToDocComment.swift:41:40: error: cannot find type 'SyntaxCodeActionScope' in scope
> [2026-04-23T05:43:05.440Z] 39 |   static let title = "Convert Comment to Doc Comment"
> [2026-04-23T05:43:05.440Z] 40 | 
> [2026-04-23T05:43:05.440Z] 41 |   static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> DeclSyntax? {
> [2026-04-23T05:43:05.440Z]    |                                        `- error: cannot find type 'SyntaxCodeActionScope' in scope
> [2026-04-23T05:43:05.440Z] 42 |     let cursorPosition = scope.snapshot.absolutePosition(of: scope.request.range.lowerBound)
> [2026-04-23T05:43:05.440Z] 43 |     guard let token = scope.file.token(at: cursorPosition) else {
> [2026-04-23T05:43:05.440Z] 
> [2026-04-23T05:43:05.440Z] /home/build-user/sourcekit-lsp/Sources/SwiftLanguageService/CodeActions/ConvertCommentToDocComment.swift:56:62: error: cannot infer contextual base in reference to member 'sourceAccurate'
> [2026-04-23T05:43:05.440Z] 54 |     else { return nil }
> [2026-04-23T05:43:05.440Z] 55 | 
> [2026-04-23T05:43:05.440Z] 56 |     guard let firstToken = declaration.firstToken(viewMode: .sourceAccurate),
> [2026-04-23T05:43:05.440Z]    |                                                              `- error: cannot infer contextual base in reference to member 'sourceAccurate'
> [2026-04-23T05:43:05.440Z] 57 |       firstToken.id == token.id
> [2026-04-23T05:43:05.440Z] 58 |     else { return nil }
> ```
> 
> I think `ConvertCommentToDocComment.swift` was added after you opened this PR. Please run tests locally again and fix the issues.

All right.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SwiftLanguageService/CodeActions/SharedCursorInfo.swift
```diff
@@ -0,0 +1,34 @@
...
+///
+/// The first access triggers the operation; subsequent accesses await the same `Task`.
+actor AsyncLazy<Success: Sendable> {
+  private let operation: @Sendable () async throws -> Success
+
+  init(_ operation: @escaping @Sendable () async throws -> Success) {
+    self.operation = operation
+  }
+
+  private lazy var task: Task<Success, any Error> = Task {
```
ahoppen: This task won’t get cancelled if the `codeAction` request is cancelled. Could you adopt cancellation here? We could eg. cancel the task if `AsyncLazy` gets reinitialized.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -927,79 +927,76 @@ extension SwiftLanguageService {
...
+
+    let uri = req.textDocument.uri
+    let range = req.range
+    let sharedCursorInfo = SharedCursorInfo { [weak self] in
+      guard let self = self else {
+        throw CancellationError()
       }
+      return try await self.cursorInfo(
+        uri,
+        range,
```
ahoppen: Do we really want to execute the cursor info request at `range`? I just implemented something similar in https://github.com/swiftlang/sourcekit-lsp/pull/2622 and there eg. when you have

```swift
var x = 2
```

And you invoke the code action on `var`, then the code action to convert the stored property to a computed property is applicable but we need to run the cursor info at `x` to get the type of `x` to add a type annotation.

MoonMao42: I think this should meet expectations.

rintaro: We already request `cursorInfo` for `range` in `retrieveRefactorCodeActions`. The idea was to reuse that if it is sufficient for determining applicability. Otherwise, I don't think we should allow each code action provider to request arbitrary `cursorInfo` during `textDocument/codeAction`.

MoonMao42:
@rintaro @ahoppen  Maybe we should align on the cursorInfo direction here.I updated the conflict resolution to follow rintaro’s suggestion for this PR. That seems cheaper, and the behavior is easier to reason about. The cases that need cursorInfo at a different position seem like they should go through the structured approach in #2622, rather than having each syntactic code action issue its own cursorInfo request.

ahoppen:
Sorry to keep you waiting here, @MoonMao42. I’ve been meaning to get back to my cursor info PR to align on a direction but just haven’t found the time. I hope that I can get to it sometime in the next weeks.

ahoppen:
Hi @MoonMao42 Sorry for the super long delay. I finally found time to come back to https://github.com/swiftlang/sourcekit-lsp/pull/2622. Once that’s reviewed and merged, you should be able to rebase your PR on top of it.

MoonMao42:
Looking forward to it.

PR #2596, opened 2026-04-06
[CMake] Move SwiftPM after SwiftSyntax

DougGregor:
We want to make sure that we import the SwiftSyntax modules rather than the namespace-prefixed ones from the Package Manager.


DougGregor:
@swift-ci please test

rintaro (approved):

DougGregor:
This didn't actually work, at least not by itself. I'm keeping the PR open in case I need it, but please don't merge on my behalf

PR #2597, opened 2026-04-07
Cache inlay hints to avoid flickering

### LINKED ISSUE #2468: Inlay hints flicker sometimes
### Swift version

Swift version 6.2.3 (swift-6.2.3-RELEASE)

### Platform

macOS Sequoia 15.4.1

### Editor

Visual Studio Code with the Swift Extension

### Description

Sometimes the inlay hints disappear for a short time before reappearing. This is probably due to the inlay hint calculation taking too long. When using soft wrapping this can cause the lines in the editor to visually shift. I have provided examples of what that looks like below. The first example shows the behavior as I normally observe it. For the second example I have added a 3-second delay at the end of the `inlayHint` method in the `SwiftLanguageService` extension in `InlayHints.swift` to exaggerate the effect.

Example:

https://github.com/user-attachments/assets/bfea6073-4b63-489a-b08c-27ff6ced98d4

Example with a 3-second delay before returning the inlay hints:

https://github.com/user-attachments/assets/4c531e70-8c5c-409b-bd4a-a896eaa6e2f2

I have also looked a bit into how other plugins deal with this, and it seems that they just recalculate the inlay hints fast enough. As far as I can tell this is also the intended solution as VSCode does not provide any API for caching the inlay hints or something similar. Caching the inlay hints in the language service may be an option, but it would probably be difficult to correctly invalidate this cache.

### Steps to Reproduce

1. Ensure soft wrapping is turned on (for VSCode: View -> Word Wrap)
2. Open an editor that contains swift source code with a long type hint, for example:
```swift
struct ReallyLongTypeNameStruct123456789 {}

let x = ReallyLongTypeNameStruct123456789()
```
3. Either make the VSCode window narrow enough or open a second editor next to the first one to force soft wrapping
4. Edit the file to force a recalculation of the inlay hints

### Logging

_No response_

Steffeeen:
Fixes #2468.

## Comparison
Before:

https://github.com/user-attachments/assets/abd4570b-2083-410e-809e-155adfe466f4

https://github.com/user-attachments/assets/13157ee3-d992-4a2e-9aa7-8efae977983a

After:

https://github.com/user-attachments/assets/b8e37145-927f-4447-a620-bc9ff1487956

https://github.com/user-attachments/assets/e9e510aa-7a2b-4abb-a4ac-85943bb1d774

## Description

Currently, we always recomputed the inlay hints by calling into SourceKit. If the document that we compute inlay hints for has recently changed, SourceKit may need a bit of time to perform parsing and semantic analysis. The inlay hints may thus need roughly 200-700ms to be computed. This causes flickering in VSCode as it removes the old inlay hints immediately and only displays them again when SourceKit-LSP returned them.

With this PR the inlay hints are cached and recomputed in the background. On each `textDocument/didChange` request the cached inlay hints are shifted according to the text edits to ensure their positions are still correct. This avoids the flickering as we always have inlay hints which we can return immediately. The returned hints may however temporarily show outdated type information until the background recompute is finished. This can be seen in the first example where the inlay hints need a short time to be updated.

## Open Questions / ToDo

There are still a few open questions which I need to think about, but I also wanted to get input on them as early as possible:

In the current version of this PR whenever the file changes, the inlay hints for the entire file are recomputed. I think this is not strictly necessary, but I'm also unsure how much of an impact it has compared to only computing inlay hints for a part of the document. The main thing that takes time for computing inlay hints is the parsing and semantic analysis of the file in SourceKit. I haven't looked into if it can do this incrementally or if it always does the semantic analysis for the entire file.

~~The inlay hints for `#if` directives are currently also cached, but I'm not sure that's necessary, they can maybe be recomputed each time.~~

~~It may currently also be possible to have a memory leak with the cache if the client disconnects/loses connection without sending a `didClose` request. This can maybe be handled by using a LRU cache or a timeout after which the cached inlay hints for a document are dropped if they were not accessed.~~

~~I also don't like the constant conversions between `Position`s and `AbsolutePosition`s in the code, but they probably can't be avoided.~~

## `DocumentSnapshot` changes 

I can also move the two commits for the `DocumentSnapshot` methods to a separate PR.

### DIFF
--- a/Sources/SwiftLanguageService/InlayHintManager.swift
+++ b/Sources/SwiftLanguageService/InlayHintManager.swift
@@ -1,0 +1,316 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(SourceKitLSP) package import LanguageServerProtocol
+@_spi(SourceKitLSP) import SKLogging
+import SKUtilities
+import SourceKitLSP
+import SwiftExtensions
+import SwiftSyntax
+
+package struct InlayHintResolveData: Codable, LSPAnyCodable {
+  package let uri: DocumentURI
+  package let position: Position
+  package let version: Int
+
+  package init(uri: DocumentURI, position: Position, version: Int) {
+    self.uri = uri
+    self.position = position
+    self.version = version
+  }
+}
+
+private struct InlayHintCacheEntry {
+  let version: Int
+  /// Marks hints as position-shifted but semantically stale; trigger a background recompute before treating cache as fully fresh.
+  let needsSemanticRefresh: Bool
+  /// The cached hints, sorted ascending by their position in the document.
+  let hints: [InlayHint]
+}
+
+private struct InFlightInlayHintRefreshTask {
+  let id: UUID
+  let expectedVersion: Int
+  let task: Task<(), any Error>
+  var sendRefreshRequest: Bool
+}
+
+actor InlayHintManager {
+  /// Cached inlay hints for each document.
+  ///
+  /// Each entry stores hints for the full document and the document version they were computed for.
+  /// - Note: The capacity has been chosen without scientific measurements. 20 seems like a resonable number of open documents a client may have.
+  private var cache = LRUCache<DocumentURI, InlayHintCacheEntry>(capacity: 20)
+
+  /// Documents that currently have a background inlay-hint recomputation in progress.
+  ///
+  /// Used to avoid scheduling multiple concurrent recomputations for the same document.
+  private var inFlightRefreshTasks: [DocumentURI: InFlightInlayHintRefreshTask] = [:]
+
+  func getCachedInlayHints(
+    swiftLanguageService service: SwiftLanguageService,
+    for snapshot: DocumentSnapshot,
+    range: Range<Position>?
+  ) -> [InlayHint]? {
+    guard let entry = cache[snapshot.uri] else {
+      scheduleInlayHintRefresh(swiftLanguageService: service, for: snapshot)
+      return nil
+    }
+
+    if entry.version != snapshot.version || entry.needsSemanticRefresh {
+      // The cached hints are stale. Schedule a refresh and return the stale hints for now. The stale hints are still in the correct position they may just contain outdated type information.
+      scheduleInlayHintRefresh(swiftLanguageService: service, for: snapshot)
+    }
+
+    return filterInlayHints(entry.hints, in: range)
+  }
+
+  private func filterInlayHints(_ hints: [InlayHint], in range: Range<Position>?) -> [InlayHint] {
+    guard let range else {
+      return hints
+    }
+
+    let lowerBoundIndex = hints.binarySearchFirst(where: { $0.position >= range.lowerBound })
+    let upperBoundIndex = hints.binarySearchFirst(where: { $0.position >= range.upperBound })
+    return Array(hints[lowerBoundIndex..<upperBoundIndex])
+  }
+
+  func processEdits(
+    for uri: DocumentURI,
+    contentChanges: [TextDocumentContentChangeEvent],
+    swiftLanguageService service: SwiftLanguageService,
+    preEditSnapshot: DocumentSnapshot,
+    postEditSnapshot: DocumentSnapshot
+  ) {
+    // Immediately schedule a refresh of the inlay hints using SourceKit, but don't send a `workspace/inlayHint/refresh`
+    // request to the client.
+    // If the time between the client sending `textDocument/didChange` and `textDocument/inlayHint` is longer than the
+    // time taken to recompute the inlay hints, we have them available immediately when the client requests inlay hints.
+    scheduleInlayHintRefresh(swiftLanguageService: service, for: postEditSnapshot, sendRefreshRequest: false)
+
+    guard let cachedEntry = cache[uri] else {
+      return
+    }
+
+    var currentHints = cachedEntry.hints
+    for change in contentChanges {
+      guard let range = change.range else {
+        // Full document replacement. Invalidate all cached hints for the document, since we don't know how to shift them.
+        cache[uri] = nil
+        return
+      }
+
+      let lineDelta = change.text.count(where: \.isNewline) - (range.upperBound.line - range.lowerBound.line)
+      let columnDelta =
+        if let lastNewlineIndex = change.text.lastIndex(where: \.isNewline) {
+          change.text.utf16.distance(from: change.text.index(after: lastNewlineIndex), to: change.text.endIndex)
+        } else {
+          change.text.utf16.count
+        }
+
+      func shiftedPosition(_ position: Position) -> Position {
+        let newUtf16Index =
+          if position.line == range.upperBound.line {
+            if lineDelta > 0 {
+              // The line does change, we thus have to calculate the offset of the hint in the new line
+              // This offset has the length of all characters added by the edit after the newline (columnDelta)
+              // + the length of the text between the end of the edit and the start of the hint
+              columnDelta + position.utf16index - range.upperBound.utf16index
+            } else {
+              // The line does not change, just add the column delta
+              position.utf16index + columnDelta
+            }
+          } else {
+            // The hint is on a different line than the edit, the column doesn't change
+            position.utf16index
+          }
+        return Position(line: position.line + lineDelta, utf16index: newUtf16Index)
+      }
+
+      let previousHints = currentHints
+      currentHints = []
+      let lowerBoundIndex = previousHints.binarySearchFirst(where: { $0.position >= range.lowerBound })
+      let upperBoundIndex = previousHints.binarySearchFirst(where: { $0.position >= range.upperBound })
+
+      // Hints before the edit range are unaffected.
+      currentHints.append(contentsOf: previousHints[..<lowerBoundIndex])
+      // Hints that overlap with the edit range are dropped and will be recomputed in the background.
+      // Hints after the edit range need to be shifted by the edit delta.
+      let shiftedHints: [InlayHint] = previousHints[upperBoundIndex...].compactMap { hint in
+        if hint.position == range.lowerBound,
+          hint.position == range.upperBound,
+          hint.labelAsString == change.text
+        {
+          // This change inserts this inlay hint, so we remove the inlay hint
+          return nil
+        }
+
+        let newPosition = shiftedPosition(hint.position)
+
+        let newTextEdits = hint.textEdits?.map { textEdit in
+          return TextEdit(
+            range: shiftedPosition(textEdit.range.lowerBound)..<shiftedPosition(textEdit.range.upperBound),
+            newText: textEdit.newText
+          )
+        }
+
+        return InlayHint(
+          position: newPosition,
+          label: hint.label,
+          kind: hint.kind,
+          textEdits: newTextEdits,
+          tooltip: hint.tooltip,
+          paddingLeft: hint.paddingLeft,
+          paddingRight: hint.paddingRight,
+          data: hint.data
+        )
+      }
+      currentHints.append(contentsOf: shiftedHints)
+    }
+
+    cache[uri] = InlayHintCacheEntry(version: postEditSnapshot.version, needsSemanticRefresh: true, hints: currentHints)
+  }
+
+  func scheduleInlayHintRefresh(
+    swiftLanguageService service: SwiftLanguageService,
+    for snapshot: DocumentSnapshot,
+    sendRefreshRequest: Bool = true
+  ) {
+    let uri = snapshot.uri
+    if var inFlightTask = inFlightRefreshTasks[uri] {
+      if inFlightTask.expectedVersion >= snapshot.version {
+        // We already have a task running for a newer version of the document, so we don't need to schedule another one.
+        // But we may have to update the `sendRefreshRequest` property, if the already running task did not need a refresh but the new one does
+        inFlightTask.sendRefreshRequest = inFlightTask.sendRefreshRequest || sendRefreshRequest
+        inFlightRefreshTasks[uri] = inFlightTask
+        return
+      }
+      // Cancel the currently running task for the older version of the document, since we will schedule a new one for the newer version below.
+      inFlightTask.task.cancel()
+    }
+
+    let taskID = UUID()
+    let task = Task(priority: .medium) { [self, service] in
+      try await run {
+        do {
+          try Task.checkCancellation()
+
+          // We recompute inlay hints for the whole document, even if only a range was requested. This is because edits
+          // can affect the validity of inlay hints outside of their edit range (e.g., an edit that changes a variable's
+          // type can make type hints for all other variables that use the edited variable stale). Caching and returning
+          // inlay hints for only a subrange of the document would add a lot of complexity, because we would need to track
+          // which hints are valid for which ranges and versions.
+          let updatedHints = try await computeTypeInlayHints(swiftLanguageService: service, for: snapshot, range: nil)
+
+          try Task.checkCancellation()
+
+          let updatedEntry = InlayHintCacheEntry(
+            version: snapshot.version,
+            needsSemanticRefresh: false,
+            hints: updatedHints
+          )
+          cache[uri] = updatedEntry
+
+          guard let inFlightTask = inFlightRefreshTasks[uri], inFlightTask.id == taskID else {
+            return
+          }
+
+          if inFlightTask.sendRefreshRequest {
+            let _ = try await service.sourceKitLSPServer?.sendRequestToClient(InlayHintRefreshRequest())
+          }
+        } catch is CancellationError {
+          return
+        } catch {
+          logger.error("Inlay hint refresh failed for \(uri.forLogging): \(error.forLogging)")
+        }
+      } cleanup: {
+        guard let inFlightTask = inFlightRefreshTasks[uri], inFlightTask.id == taskID else {
+          // This task was already replaced by another one so we should not remove the entry
+          return
+        }
+        inFlightRefreshTasks[uri] = nil
+      }
+    }
+
+    inFlightRefreshTasks[uri] = InFlightInlayHintRefreshTask(
+      id: taskID,
+      expectedVersion: snapshot.version,
+      task: task,
+      sendRefreshRequest: sendRefreshRequest
+    )
+  }
+
+  func computeTypeInlayHints(
+    swiftLanguageService service: SwiftLanguageService,
+    for snapshot: DocumentSnapshot,
+    range: Range<Position>?
+  ) async throws -> [InlayHint] {
+    let infos = try await service.variableTypeInfos(snapshot.uri, nil)
+    return infos
+      .lazy
+      .filter { !$0.hasExplicitType }
+      .map { info -> InlayHint in
+        let position = info.range.upperBound
+        let variableStart = info.range.lowerBound
+        let label = ": \(info.printedType)"
+        let textEdits: [TextEdit]?
+        if info.canBeFollowedByTypeAnnotation {
+          textEdits = [TextEdit(range: position..<position, newText: label)]
+        } else {
+          textEdits = nil
+        }
+        let resolveData = InlayHintResolveData(uri: snapshot.uri, position: variableStart, version: snapshot.version)
+        return InlayHint(
+          position: position,
+          label: .string(label),
+          kind: .type,
+          textEdits: textEdits,
+          data: resolveData.encodeToLSPAny()
+        )
+      }
+      .sorted { $0.position < $1.position }
+  }
+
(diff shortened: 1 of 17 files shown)


Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -12,10 +12,40 @@
...
+  let id: UUID
+  let expectedVersion: Int
+  let task: Task<Void, any Error>
+}
+
+struct InlayHintState {
+  /// Cached inlay hints for each document.
+  ///
+  /// Each entry stores hints for the full document and the document version they were computed for.
+  var cache: [DocumentURI: InlayHintCacheEntry] = [:]
```
ahoppen: I think I would prefer to make this an LRU cache similar to what we have for in the diagnostics manager and the syntax tree cache.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -30,28 +60,233 @@ package struct InlayHintResolveData: Codable, LSPAnyCodable {
...
   }
 }
 
+struct ShiftedInlayHint {
+  let inlayHint: InlayHint
+  let position: AbsolutePosition
+  let dataPosition: AbsolutePosition?
+  let textEdits: [ShiftedTextEdit]
+
+  func shiftBy(delta: SourceLength) -> ShiftedInlayHint {
```
ahoppen: This should be named `shiftedBy` to follow https://www.swift.org/documentation/api-design-guidelines/#name-according-to-side-effects.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -12,10 +12,40 @@
...
+  let hints: [InlayHint]
+}
+
+struct InFlightInlayHintRefreshTask {
+  let id: UUID
+  let expectedVersion: Int
+  let task: Task<Void, any Error>
+}
+
+struct InlayHintState {
```
ahoppen: Could we have an `InlayHintManager` similar to how we have a `SyntaxTreeManager` that takes care of all the position shifting etc. I think encapsulating all of that logic in one place will make it easier to reason about.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -30,28 +60,233 @@ package struct InlayHintResolveData: Codable, LSPAnyCodable {
...
+        guard snapshot.version == expectedVersion else {
+          return
+        }
+
+        let updatedEntry = InlayHintCacheEntry(version: snapshot.version, hints: updatedHints)
+        await self.storeInlayHintCache(updatedEntry, for: uri)
+
+        let _ = try await self.sourceKitLSPServer?.sendRequestToClient(InlayHintRefreshRequest())
+      } catch is CancellationError {
+        throw CancellationError()
```
ahoppen: I think this doesn’t do anything.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -30,28 +60,233 @@ package struct InlayHintResolveData: Codable, LSPAnyCodable {
...
+    guard let inFlightTask = inlayHintState.inFlightRefreshTasks[uri], inFlightTask.id == taskID else {
+      return
+    }
+    inlayHintState.inFlightRefreshTasks[uri] = nil
+
+    guard let latestVersion = try? documentManager.latestSnapshot(uri).version else {
+      return
+    }
+    if latestVersion > inFlightTask.expectedVersion {
+      scheduleInlayHintRefresh(for: uri, expectedVersion: latestVersion)
```
ahoppen: Could you add a comment why this is necessary?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -759,9 +763,21 @@ extension SwiftLanguageService {
       edits: concurrentEdits
     )
 
+    shiftCachedInlayHints(
+      for: notification.textDocument.uri,
+      edits: edits,
+      preEditSnapshot: preEditSnapshot,
+      postEditSnapshot: postEditSnapshot
+    )
+    scheduleInlayHintRefresh(for: notification.textDocument.uri, expectedVersion: postEditSnapshot.version)
```
ahoppen: Actually, I wonder if we should be re-computing inlay hints on every edit to the document if we had inlay hints cached for a previous version of the document. Chances are very high that the client will need inlay hints for the next document version as well and this way there is a chance that we’ll already have them computed when the client asks for them (or at least reduces the amount of time until they are available). What do you think?

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -30,28 +60,233 @@ package struct InlayHintResolveData: Codable, LSPAnyCodable {
...
+        try Task.checkCancellation()
+
+        guard snapshot.version == expectedVersion else {
+          return
+        }
+
+        let updatedEntry = InlayHintCacheEntry(version: snapshot.version, hints: updatedHints)
+        await self.storeInlayHintCache(updatedEntry, for: uri)
+
+        let _ = try await self.sourceKitLSPServer?.sendRequestToClient(InlayHintRefreshRequest())
```
ahoppen: My biggest concern with this approach is that this will cause the client to refresh inlay hints for *all* open documents, not just the one that we just re-computed the inlay hints for. I wonder if we could add an LSP extension to this request to only refresh a single document. That being said, I don’t think it’s a blocker and previous versions of inlay hints worked like this, so it’s probably fine – would just be a nice optimization.

@plemarquand Do you think it would be possible to add a middleware to the VS Code extension that captures `inlayHint/refresh` requests and only refreshes inlay hints for a single document as opposed to all?

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -96,7 +338,47 @@ extension SwiftLanguageService {
...
+      currentHints = []
+
+      let lowerBoundIndex = previousHints.binarySearchFirst(where: { $0.position >= edit.range.lowerBound })
+      let upperBoundIndex = previousHints.binarySearchFirst(where: { $0.position >= edit.range.upperBound })
+
+      // Hints before the edit range are unaffected.
+      currentHints.append(contentsOf: previousHints[..<lowerBoundIndex])
+      // Hints that overlap with the edit range are dropped and will be recomputed in the background.
+      // Hints after the edit range need to be shifted by the edit delta.
+      currentHints.append(contentsOf: previousHints[upperBoundIndex...].map { $0.shiftBy(delta: delta) })
```
ahoppen: Just a thought: Would it be possible to do the shifting in the `Position` line:column space instead of the `AbsolutePosition` offset-based space? I think it should work as well because we just need to add something to the line number if a line was inserted before and adjust the column if there was an edit on this line. Could be missing something though.

ahoppen (commented):
Thanks. Would be great to get rid of that flickering 😄

Sources/SwiftLanguageService/InlayHints.swift
Steffeeen: I removed this check as I found it to be unnecessary.

Sources/SwiftLanguageService/SwiftLanguageService.swift
Steffeeen: I don't quite get what you are suggesting. The code already starts a Task for updating the inlay hints using SourceKit immediately when it receives a `textDocument/didChange`.

Sources/SwiftLanguageService/InlayHints.swift
Steffeeen: Yeah, that works, I initially rejected this idea because the `SourceEdit`s were using `Range<AbsolutePosition>`. However, I now pass in the `TextDocumentContentChangeEvent`s which use `Range<Position>`.

Sources/SwiftLanguageService/InlayHints.swift
Steffeeen: Yeah, I also don't like that this causes the client to refresh the inlay hints in all documents. However, according to the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_inlayHint_refresh):
> Note that the client still has the freedom to delay the re-calculation of the inlay hints if for example an editor is currently not visible."

So the client may not request the inlay hints for all open documents immediately. Additionally, the inlay hints for the open documents are cached so it's just a matter of returning the cached inlay hints. So this shouldn't trigger a SourceKit computation for all the other documents.

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,270 @@
...
+
+    var currentHints = cachedEntry.hints
+    for change in contentChanges {
+      guard let range = change.range else {
+        // Full document replacement. Invalidate all cached hints for the document, since we don't know how to shift them.
+        cache[uri] = nil
+        return
+      }
+
+      let lineDelta = change.text.count(where: { $0.isNewline }) - (range.upperBound.line - range.lowerBound.line)
```
ahoppen: Nitpick to make this read a little nicer.

```suggestion
      let lineDelta = change.text.count(where: \.isNewline) - (range.upperBound.line - range.lowerBound.line)
```

Similar below

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,270 @@
...
+      let columnDelta =
+        if let lastNewlineIndex = change.text.lastIndex(where: { $0.isNewline }) {
+          change.text.distance(from: change.text.index(after: lastNewlineIndex), to: change.text.endIndex)
```
ahoppen: I think we need to adjust by UTF-16 positions here instead of numbers of grapheme clusters.

```suggestion
          change.text.utf16.distance(from: change.text.index(after: lastNewlineIndex), to: change.text.endIndex)
```

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,270 @@
...
+        } else {
+          change.text.count
```
ahoppen:
```suggestion
          change.text.utf16.count
```

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,270 @@
...
+      if inFlightTask.expectedVersion >= snapshot.version {
+        // We already have a task running for a newer version of the document, so we don't need to schedule another one.
+        return
+      }
+      // Cancel the currently running task for the older version of the document, since we will schedule a new one for the newer version below.
+      inFlightTask.task.cancel()
+    }
+
+    let taskID = UUID()
+    let task = Task(priority: .medium) { [weak self, service] in
```
ahoppen: Since we capture `self` strongly, immediately when the task starts executing, I think we can avoid the `weak` dance here. It only has an affect if `InlayHintManager` would get destroyed after the task was created but before it starts executing, and that’s usually a very short time period.

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,270 @@
...
+      guard let self else {
+        return
+      }
+      defer { Task { await finishInlayHintRefresh(swiftLanguageService: service, for: uri, taskID: taskID) } }
```
ahoppen: Instead of spawning a `Task` here that may execute whenever, it would be nicer to use `run(_:cleanup:)` here. (At least until we get async defers).

With that, we might also be able to inline the implementation of `finsihInlayHintRefresh` here, which would help match the removal of tasks from `inFlightRefreshTasks` to where they are created.

Sources/SwiftLanguageService/InlayHints.swift
```diff
@@ -12,71 +12,53 @@
...
-          textEdits: textEdits,
-          data: resolveData.encodeToLSPAny()
-        )
-      }
+    if let sourceKitLSPServer = self.sourceKitLSPServer,
+      let clientCapabilities = await sourceKitLSPServer.capabilityRegistry?.clientCapabilities,
+      !(clientCapabilities.workspace?.inlayHint?.refreshSupport ?? false)
+    {
+      // The client does not support workspace/inlayHint/refresh.
+      // We have to compute inlay hints on every request, because we cannot trigger a refresh when the document changes.
```
ahoppen:
```suggestion
      // We have to compute inlay hints on every request, because we cannot trigger a refresh when the inlay hints have been recomputed in the background.
```

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -19,25 +19,92 @@ import XCTest
...
+    func getBackgroundComputedInlayHints() async throws -> [InlayHint] {
+      let refreshRequestReceived = testCase.expectation(description: "Receive first inlay hint refresh request")
+      testClient.handleSingleRequest { (_: InlayHintRefreshRequest) in
+        refreshRequestReceived.fulfill()
+        return VoidResponse()
+      }
+
+      let request = InlayHintRequest(textDocument: TextDocumentIdentifier(uri), range: range)
+      let _ = try await testClient.send(request)
+      await testCase.fulfillment(of: [refreshRequestReceived], timeout: 10)
```
ahoppen: CI can be surprisingly slow in some cases. I’d prefer if we just stick with the default timeout here.

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -404,4 +509,202 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+    ) {
+      let hints = try await $0.getBackgroundComputedInlayHints()
+      assertHintsEqual(
+        hints,
+        [
+          makeInlayHint(position: $0.positions["3️⃣"], kind: .type, label: ": Int")
+        ]
+      )
+
+      $0.sendChange(
```
ahoppen: I generally prefer to only use the anonymous `$0` variables if the closure is a single line (or at least not significantly more), because it’s too easy to loose context of what `$0` is over multiple lines. Could you name the variable in the tests?

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -404,4 +509,202 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+          let b: Int = 10
+          """
+      )
+
+      let shiftedHints = try await $0.getCachedInlayHints()
+      assertHintsEqual(
+        shiftedHints,
+        [
+          makeInlayHint(position: $0.positions["1️⃣"], kind: .type, label: ": Int"),
+          makeInlayHint(position: Position(line: 3, utf16index: 5), kind: .type, label: ": Int"),
```
ahoppen: Could we use a position marker here instead of hard-coding the position? Same for a few more test cases below.

Sources/SwiftLanguageService/InlayHints.swift
ahoppen: OK, I guess it’s fine in practice. Still would like to hear from Paul if we can improve this in the long term.

Sources/SwiftLanguageService/SwiftLanguageService.swift
ahoppen: I was wondering if we should schedule a background re-computation of inlay hints on every edit (just without sending the `inlayHint/refresh` request to the client once this one is finished. That way, if the editor sends an inlay hints request, say, 1s after the edit, we might have already computed the new inlay hints and we don’t need to take the `inlayHint/refresh` path at all. And even if re-computing the inlay hints takes 2s, then we will be able to show the inlay hints 1s earlier than if we only triggered the refresh when the client sends the inlay hint request.

What do you think? I think regardless, this could be an improvement for a follow-up PR.

Sources/SwiftLanguageService/InlayHints.swift
plemarquand: @ahoppen I think this would be tricky to implement; there isn't a middleware for the refresh notification in the VS Code language client API.  

The vscode-languageclient [requests a range](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/inlayHint.ts#L69) when asking for inlay hints, so from the client's perspective its already asking for the bare minimum

ahoppen: OK, in that case I think we just stick with the current design. I was just hoping that there would be a way to avoid the following situation:
- The user has two tabs open side-by-side in VS Code
- We send a `inlayHints/refresh` request because we re-computed the tokens for the left-hand editor
- VS Code would now also refresh the inlay hints for the right-hand editor

But it probably isn’t all that bad in practice.

Sources/SwiftLanguageService/SwiftLanguageService.swift
Steffeeen: The `InlayHintManager` currently already calls `scheduleInlayHintRefresh` in `processEdits` which accomplishes what you are suggesting. However, that of course does always trigger the `inlayHint/refresh` path currently. I'm now also realizing that we probably should not schedule a refresh if the client does not support `inlayHint/refresh` as the results of this inlay hint computation will never be used.

I think implementing this should simply be adding a bool to `scheduleInlayHintRefresh` indicating whether to send the refresh request at the end. I will try it out.

ahoppen: Oh, I totally missed that. But doesn’t this mean that all the tests are also racy. If the inlay hint refresh triggered by the edit finishes before the tests send the `textDocument/inlayHints` requests, then we won’t trigger another `inlayHints/refresh` request and thus waiting for that will time out in the test.

Tests/SourceKitLSPTests/InlayHintTests.swift
Steffeeen: I changed the code to use position markers where possible. In `testInlayHintShiftingWithMultipleChanges` the manual positions cannot really be avoided, due to the changes being incremental.

Sources/SwiftLanguageService/SwiftLanguageService.swift
Steffeeen: I guess that could have happened, although I haven't had that happen when executing the tests locally.

I added the capability check for processing the edits. I also added the bool to `scheduleInlayHintRefresh` so that we don't send the refresh request when a recomputation of the inlay hints was trigged by a change.

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,316 @@
...
+
+          if inFlightTask.sendRefreshRequest {
+            let _ = try await service.sourceKitLSPServer?.sendRequestToClient(InlayHintRefreshRequest())
+          }
+        } catch {
+          if error is CancellationError {
+            return
+          }
+          logger.error("Inlay hint refresh failed for \(uri.forLogging): \(error.forLogging)")
+        }
```
ahoppen, line 235: Could be slightly simplified

```suggestion
        } catch is CancellationError {
          return
        } catch {
          logger.error("Inlay hint refresh failed for \(uri.forLogging): \(error.forLogging)")
        }
```

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,316 @@
...
+  package init(uri: DocumentURI, position: Position, version: Int) {
+    self.uri = uri
+    self.position = position
+    self.version = version
+  }
+}
+
+private struct InlayHintCacheEntry {
+  let version: Int
+  let needsSemanticRefresh: Bool
```
ahoppen, line 36: Could you add a comment what exactly this means?

Sources/SwiftLanguageService/InlayHintManager.swift
```diff
@@ -0,0 +1,316 @@
...
+}
+
+private extension InlayHint {
+  var labelAsString: String {
+    switch self.label {
+    case .string(let label):
+      return label
+
+    case .parts(let parts):
+      return parts.map { $0.value }.joined(separator: "")
```
ahoppen: I think `separator` defaults to an empty string.

```suggestion
      return parts.map { $0.value }.joined()
```

Sources/SwiftLanguageService/SwiftLanguageService.swift
ahoppen: I have seen CI failures for race conditions that seem a lot less likely in the past, so I suspect that we would hit this race condition at some point. 

I think what we’d need to do in `getBackgroundComputedInlayHints` is something like the following:
- Run `textDocument/inlayHints`
- Check if the result matches the expectation
- If not, wait for a `inlayHints/refresh` request
- Check if the results now match the expectations

Steffeeen: I renamed `getBackgroundComputedInlayHints` to `checkInlayHintsComputedInTheBackgroundMatch` and changed it to do what you suggested. I also verified that this fixes the potential timeout by adding a `Task.sleep()` call in one of the tests after the `sendChange()` call.

ahoppen (approved):
Looks good. Thank you 😄

ahoppen:
@swift-ci Please test

Steffeeen:
Seems like `fulfillment()` does not have a default value for `timeout` on Linux and Windows. `fulfillment()` was implemented here: https://github.com/swiftlang/swift-corelibs-xctest/pull/453. But I'm not sure if the default value for the timeout was just missed in the corelibs PR or if the default value in the Xcode version was introduced later and the corelibs version was not adjusted.

Should I just set the timeout manually for all the calls in the tests or do we want to fix it upstream?

ahoppen:
You should be able to use `fulfillmentOfOrThrow`, which handles the timeout already and also aborts the test if the expectation is not fulfilled.

Steffeeen:
@swift-ci Please test

Steffeeen:
@swift-ci Please test windows

PR #2598, opened 2026-04-07
Preserve language service when closing generated interface

edwardlee:
Closing a generated interface was removing the language service for the originating file because both share the same buildSettingsFile key. https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/SourceKitLSP/ReferenceDocumentURL.swift#L125
Guard the removal by when no other open document shares that key.
See #2209 and https://github.com/swiftlang/sourcekit-lsp/commit/7437d924538e229dab98e3b6d87ae51db467d196

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1442,7 +1442,9 @@ extension SourceKitLSPServer {
       await languageService.closeDocument(notification)
     }
 
-    workspace.removeLanguageServices(for: uri)
+    if !documentManager.openDocuments.contains(where: { $0.buildSettingsFile == uri.buildSettingsFile }) {
+      workspace.removeLanguageServices(for: uri)
+    }
 
     workspaceQueue.async {
       self.workspaceForUri[notification.textDocument.uri] = nil
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -395,6 +395,41 @@ final class SwiftInterfaceTests: SourceKitLSPTestCase {
     XCTAssertEqual(transformLocation.uri.scheme, "sourcekit-lsp")
     assertContains(transformLocation.uri.pseudoPath, "Foundation.NSAffineTransform.swiftinterface")
   }
+
+  func testClosingGeneratedInterfacePreservesOriginatingLanguageService() async throws {
+    let testClient = try await TestSourceKitLSPClient(
+      capabilities: ClientCapabilities(experimental: [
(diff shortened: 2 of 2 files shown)


rintaro:
Good find, thanks! Could you add a test case that fails without this change?

rintaro:
On second thought, what if the documentManager document is closed first while the related generated interfaces are still open? Do they lose the language service?
Perhaps the language service should be tied to the uri itself, not the buildSettingsFile?

edwardlee:
Added test similar to testNoDiagnosticsInGeneratedInterface. Since the logic checks for any remaining open file with the key before shutting down LSP, both cases work. Tying to the uri makes sense, this was just simpler.

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1457,7 +1442,9 @@ extension SourceKitLSPServer {
       await languageService.closeDocument(notification)
     }
 
-    workspace.removeLanguageServices(for: uri)
+    if !documentManager.openDocuments.contains(where: { $0.buildSettingsFile == uri.buildSettingsFile }) {
```
rintaro, line 1445: Using `.buildSettingsFile` as the key is an implementation detail of `Workspace`, so ideally this logic should remain encapsulated there.
Could you move this into `Workspace.removeLanguageServices(for:)`? You should be able to access the document manager via `self.sourceKitLSPServer?.documentManager`.
Also, maybe `removeLanguageServices(for:)` should be renamed to `unsetLanguageServices(for:)` to pair better with `setLanguageServices(for:_:)`.

rintaro (commented):
Thank you for the test case!
I think the logic is fine. Just a code placement concern.

rintaro:
@swift-ci Please test Windows

rintaro:
```
[2026-04-23T06:10:58.845Z] <EXPR>:0: error: SwiftInterfaceTests.testClosingGeneratedInterfacePreservesOriginatingLanguageService : threw error "ResponseError(code: LanguageServerProtocol.ErrorCode(rawValue: -32001), message: "Unknown error: Invalid Scheme for reference document", data: nil)"
[2026-04-23T06:10:58.845Z] Test Case 'SwiftInterfaceTests.testClosingGeneratedInterfacePreservesOriginatingLanguageService' failed (0.218 seconds)
```
🤔

rintaro:
@swift-ci Please test

rintaro:
Hi @edwardlee, again, great catch and thanks for the fix! Since this is a serious issue we need to fix asap, I've opened #2648 to get this merged sooner.
I made a small adjustment to move the guard into `Workspace` and updated the test to pass on CI (hopefully).

PR #2607, opened 2026-04-13
Make RenameLocation unique

DeeSee:
## Summary

When a file is indexed by multiple compilation units (e.g. multiple targets in a build system compile or import the same source file), `index.occurrences(ofUSR:roles:)` returns one entry per unit for the same source position. These duplicate `RenameLocation`s produce overlapping `TextEdit`s in the `WorkspaceEdit`, which editors reject — VS Code shows "Rename failed to apply edits" with no further detail.

## Changes

- Changed `locationsByFile` to collect rename locations in a `Set<RenameLocation>` instead of `[RenameLocation]`, effectively making duplicates impossible.
- Made `RenameLocation` and `RenameLocation.Usage` conform to `Hashable` so they can be stored in a `Set`.
- Sort the set into an array before passing to `editsToRename`, since the downstream API expects `[RenameLocation]` and some callers depend on ordering.

## Test

- I've added a test case that simulates the issue: it compiles single source file twice with different compilation conditions and then checks that rename doesn't produce duplicate edits. I've verified that this test fails if I rollback the fix.

### DIFF
--- a/Tests/SourceKitLSPTests/RenameTests.swift
+++ b/Tests/SourceKitLSPTests/RenameTests.swift
@@ -10,9 +10,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+@_spi(SourceKitLSP) import BuildServerProtocol
 @_spi(SourceKitLSP) import LanguageServerProtocol
 import SKLogging
 import SKTestSupport
+import SwiftExtensions
+import ToolchainRegistry
 import XCTest
 
 final class RenameTests: SourceKitLSPTestCase {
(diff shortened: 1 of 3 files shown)


Tests/SourceKitLSPTests/RenameTests.swift
```diff
@@ -1323,4 +1331,169 @@ final class RenameTests: SourceKitLSPTestCase {
     )
     XCTAssertEqual(result?.changes, [:])
   }
+
+  func testRenameDedupesMultiUnitIndexEntries() async throws {
+    let scratchDir = try testScratchDir()
```
ahoppen: Could you use `withTestScratchDir` so that the test directory gets deleted even if one of the function calls throws before `TestSourceKitLSPClient` gets created?

Tests/SourceKitLSPTests/RenameTests.swift
```diff
@@ -1323,4 +1331,169 @@ final class RenameTests: SourceKitLSPTestCase {
...
+    var sharedFlags: [String] = [
+      "-module-name", "MyLib",
+      "-index-store-path", indexPath,
+      "-module-cache-path", cachePath,
+    ]
+    if let sdk = defaultSDKPath {
+      sharedFlags += ["-sdk", sdk]
+    }
+
+    func runFrontend(
```
ahoppen: I think a simpler alternative than invoking the compilers here manually would be to generate a `CustomBuildServerTestProject` that reports two targets that both reference the same source file. That would also more closely replicate a real-life situation.

Alternatively, if we want to stick with this approach: We have a very similar code path to index a file in `IndexedSingleSwiftFileTestProject.init`. Could you check if you could extract that to eg. a static methods on `IndexedSingleSwiftFileTestProject` so that we can just re-use it from here?

ahoppen (commented):
Thank, @DeeSee. The change looks good to me. I just have a few suggestions to improve the test case.

Tests/SourceKitLSPTests/RenameTests.swift
DeeSee: Seems like after moving to CustomBuildServerTestProject, test scratch files are now managed internally

Tests/SourceKitLSPTests/RenameTests.swift
DeeSee: I've tried CustomBuildServerTestProject and it looks great! Please check out how it turned out now

Tests/SourceKitLSPTests/RenameTests.swift
```diff
@@ -1323,4 +1326,123 @@ final class RenameTests: SourceKitLSPTestCase {
...
+
+    for (uri, edits) in changes {
+      let ranges = edits.map(\.range)
+      let uniqueRanges = Set(ranges)
+      XCTAssertEqual(
+        ranges.count,
+        uniqueRanges.count,
+        "Duplicate edits in \(uri): \(edits)"
+      )
+    }
```
ahoppen: I think it would make sense to actually compare the list of changes to check that we get exactly two changes at the locations.

ie. something like `XCTASsertEqual(changes, …)`. And try to use location markers for all the locations.

ahoppen (commented):
Just one nitpick, otherwise LGTM.

Tests/SourceKitLSPTests/RenameTests.swift
DeeSee: Good catch, made check more structured

ahoppen (approved):
Thanks for the fix @DeeSee. And congrats to your first Swift contribution 🙂

ahoppen:
@swift-ci Please test

DeeSee:
Thank you! Actually, it's second contribution if you count [lldb](https://github.com/swiftlang/llvm-project/pull/1659/changes) 😊 I'm huge fan of deduplication you know 😅

ahoppen:
I did not see that. Let’s re-word it as congrats to you first SourceKit-LSP contribution then 😉

PR #2612, opened 2026-04-20
[SwiftPM] Ignore file events from scratch directory

rintaro:
Binary targets backed by zip archives cause SwiftPM to extract the artifact bundle into the scratch directory on every package load. The resulting file-change events triggered another reload, causing an infinite loop.

Add `isInScratchDirectory` to `fileEventShouldTriggerPackageReload` to filter out events before the build-settings check. It covers both the configured scratch directory and the default `.build/` directory (which receives regular `swift build` output even when a custom scratch path is configured).

https://github.com/swiftlang/sourcekit-lsp/issues/2615

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -122,6 +122,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
 
   private let toolchain: Toolchain
   private let swiftPMWorkspace: Workspace
+  private let defaultScratchDirectory: AbsolutePath
 
   private let pluginConfiguration: PluginConfiguration
   private let traitConfiguration: TraitConfiguration
@@ -250,6 +251,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     }
 
     let absProjectRoot = try AbsolutePath(validating: projectRoot.filePath)
+    self.defaultScratchDirectory = Workspace.DefaultLocations.scratchDirectory(forRootPackage: absProjectRoot)
     self.toolsets =
       try options.swiftPMOrDefault.toolsets?.map {
         try AbsolutePath(validating: $0, relativeTo: absProjectRoot)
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -1242,6 +1244,193 @@ struct SwiftPMBuildServerTests {
...
+      await server.didChangeWatchedFiles(
+        notification: OnWatchedFilesDidChangeNotification(
+          changes: [
+            FileEvent(uri: DocumentURI(extractedInfoJSON), type: .deleted),
+            FileEvent(uri: DocumentURI(extractedInfoJSON), type: .created),
+          ]
+        )
+      )
+
+      #expect(!unexpectedReloadStarted.value)
```
hamishknight, line 1366: Is this guaranteed to catch this case? Seems like we async dispatch to `packageLoadingQueue` before calling `reloadPackageDidStart`

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -919,6 +921,23 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+    guard let filePath = try? AbsolutePath(validating: url.filePath) else {
+      return false
+    }
+    if filePath.isDescendantOfOrEqual(to: self.swiftPMWorkspace.location.scratchDirectory) {
+      return true
+    }
+
+    // Also ignore the default '.build' directory. SourceKit-LSP uses '.build/index-build' as its scratch
+    // directory by default, but users can configure a custom scratch path outside of '.build'. In that case,
+    // regular 'swift build' output still lands in '.build', so we want to ignore events there too.
```
hamishknight, line 934: Do we care about cases where the user is using a custom `--scratch-path` with `swift build`?

hamishknight (commented):
I'm not familiar enough with the logic here to say for sure, but this looks reasonable to me. I would like @ahoppen to take a look though since he's worked on fixing issues here before.

One question is what happens when a build plugin is edited to produce a new output, do we still handle that correctly? (though it's not actually clear to me what actually triggers the plugin to re-run today)

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
rintaro: sourcekit-lsp can't actually know the exact build directory. Actually, this `.build` exclusion is unnecessary because `fileAffectsSwiftOrClangBuildSettings` should return `false` for non-`.build/index-build` files. But it's just fast to check and `swift build` is likely using the default path.

hamishknight: Right yeah, I was just wondering if there are potential issues with `.build` whether there could also be issues for a custom scratch path too. But as you say there's nothing we can really do there 🤷

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
rintaro: Added 
```swift
await server.waitForBuildSystemUpdates(request: WorkspaceWaitForBuildSystemUpdatesRequest())
```
to wait for it.

rintaro:
@swift-ci Please test Windows

hamishknight (approved):

MaxDesiatov:
@swift-ci test windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

ahoppen:
> I'm not familiar enough with the logic here to say for sure, but this looks reasonable to me. I would like @ahoppen to take a look though since he's worked on fixing issues here before.

This looks good to me. I can’t think of any reason why we might need to listen for changes to the build directory.

PR #2614, opened 2026-04-21
[Test] Disable a test case while investigating

rintaro:
rdar://175272642


finagolfin:
I'm guessing this was some merged compiler or other dependency change, as this test started failing on this repo at the exact same commit where it was passing before, ie nothing changed in this repo itself to break this test.

PR #2617, opened 2026-04-22
[Test] Fix testCompilerArgumentsForFileThatContainsPlusCharacterURLEcoded

rintaro:
The test was constructing the percent-encoded file URL from `filePath`, which uses native OS path separators. On Windows this produced a backslash-separated path inside a `file://` URL, making the URL malformed. Fix by replacing `+` with `%2B` in `absoluteString` instead, which is already a well-formed URL string with forward slashes on all platforms.

rdar://175272642


rintaro:
@swift-ci Please test

hamishknight (approved):

PR #2618, opened 2026-04-22
[6.3][SwiftPM] Ignore file events in scratch directory

rintaro:
Cherry-pick #2612 into `release/6.3`

- **Explanation**: Binary targets backed by remote `.artifactbundleindex` URLs cause SwiftPM to re-extract the artifact bundle into the scratch directory on every package load. The resulting file-change events (delete + create) for the extracted files triggered another reload, creating an infinite loop. This fix ignores file system events from the  scratch directory because it shouldn't contain user-editable source code.
- **Scope**: Only affects packages with binary targets backed by remote `.artifactbundleindex` URLs. No oberservable behavior change for other packages. Cannot suppress legitimate package reloads — package manifests and source files are never inside the scratch directory.
- **Risk**: Low. The change only adds an early-return guard for scratch-directory paths in the file-event filter. Scratch directories cannot contain package manifests or source files.
- **Issues**: https://github.com/swiftlang/sourcekit-lsp/issues/2615
- **Original PRs**: https://github.com/swiftlang/sourcekit-lsp/pull/2612
- **Testing**: Added regression test cases.
- **Reviewers**: Hamish Knight (@hamishknight)

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -122,6 +122,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
 
   private let toolchain: Toolchain
   private let swiftPMWorkspace: Workspace
+  private let defaultScratchDirectory: AbsolutePath
 
   private let pluginConfiguration: PluginConfiguration
   private let traitConfiguration: TraitConfiguration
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

hamishknight (approved):

rintaro:
Decided not to include this

PR #2619, opened 2026-04-24
Add `workspace/symbolNames` and `workspace/symbolInfo` LSP extensions

rintaro:
- **`sourcekit/workspace/symbolNames`** — returns a flat, deduplicated list of every symbol name in the workspace index (source and indexed system modules). Clients use this to drive their search UI locally.

- **`sourcekit/workspace/symbolInfo`** — given a list of exact symbol names, returns `WorkspaceSymbolItem` for each occurrence across all workspaces, for display in the search result list. Source-file symbols get `SymbolInformation` with a `file://` location. SDK/stdlib symbols get a `WorkspaceSymbol` with `location: .uri(…)` The client must call `workspaceSymbol/resolve` after the user selects an SDK/stdlib symbol to obtain the concrete interface location.

- **`workspaceSymbol/resolve`** — resolves the deferred `WorkspaceSymbol` location from `sourcekit/workspace/symbolInfo`. Parses the `?module=` value into `moduleName`/`groupName`, finds a real source file via `mainFiles(containing:)`, calls `openGeneratedInterface`, and returns the symbol with `location` replaced by a full `sourcekit-lsp://generated-swift-interface/` URI + range (or a temp `file://` path for clients without `workspace/getReferenceDocument` support).


rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -980,142 +1133,41 @@ package actor BuildServerManager: QueueBasedMessageHandler {
     }
   }
 
+  // MARK: File mappings
+
   /// Check if the URI referenced by `location` has been copied during the preparation phase. If so, adjust the URI to
   /// the original source file.
   package func locationAdjustedForCopiedFiles(_ location: Location) -> Location {
```
ahoppen: Is it worth still having these shims or should callers just call into `CopiedFileMap`?

Sources/BuildServerIntegration/BuildServerManager.swift
ahoppen, line 1: I think the refactoring of `CopiedFileMap` would make a good standalone PR.

Sources/SemanticIndex/CheckedIndex.swift
```diff
@@ -96,6 +96,10 @@ package final class CheckedIndex {
     self.checker = IndexOutOfDateChecker(checkLevel: checkLevel)
   }
 
+  package func allSymbolNames() throws -> [String] {
+    return try index.allSymbolNames()
```
ahoppen: Since we don’t do any validation of whether the symbol names still belong to up-to-date documents, I think this should be on `UncheckedIndex`.

Sources/SourceKitLSP/GeneratedInterfaceDocumentURLData.swift
```diff
@@ -63,24 +75,37 @@ package struct GeneratedInterfaceDocumentURLData: Hashable, ReferenceURLData {
     }
   }
 
-  package init(moduleName: String, groupName: String?, sourcekitdDocumentName: String, primaryFile: DocumentURI) {
+  /// Creates a fully resolved instance, synthesizing `sourcekitdDocumentName` from the module name and a hash of the
+  /// build-settings file so that the sourcekitd buffer name is unique per build context.
+  package init(moduleName: String, groupName: String?, primaryFile: DocumentURI) {
+    let buildSettingsFile = primaryFile.buildSettingsFile
+    let hash = "\(abs(buildSettingsFile.stringValue.hashValue))"
```
ahoppen: Should we use SHA256 or something like that to get stable hash values across launches of SourceKit-LSP. I just have a feeling that having stable values might help with debugging in the long run

Sources/SwiftExtensions/Array+SortAndDedupe.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Array where Element: Comparable {
+  /// Whether the array's elements are in strictly ascending order with no duplicates.
+  public var isSortedAndUnique: Bool {
+    @_specialize(where Element == String)
```
ahoppen, line 16: Mostly out of curiosity: Does `@_specialize` have any effect here if we build LSP with package CMO?

Sources/SwiftExtensions/Array+SortAndDedupe.swift
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Array where Element: Comparable {
+  /// Whether the array's elements are in strictly ascending order with no duplicates.
+  public var isSortedAndUnique: Bool {
```
ahoppen: Let’s make this `package` instead of `public`. Same below.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1767,124 @@ extension SourceKitLSPServer {
     return try await languageService.signatureHelp(req)
   }
 
+  /// Handle a workspace/symbolNames request, returning the name list.
+  func workspaceAllSymbolNames(_ req: WorkspaceAllSymbolNamesRequest) async throws -> WorkspaceAllSymbolNamesResponse {
+    var symbols = try await self.workspaces.asyncFlatMap { workspace in
+      try await workspace.index(checkedFor: .deletedFiles)?.allSymbolNames() ?? []
+    }
+    if !symbols.isSortedAndUnique {
```
ahoppen, line 1781: Should we just check for `isSortedAndUnique` in `sortAndDedupe`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1767,124 @@ extension SourceKitLSPServer {
...
+
+  /// Map a `SymbolOccurrence` from the index to a `WorkspaceSymbolItem` suitable for returning in a
+  /// `workspace/symbolInfo` response.
+  private func workspaceSymbolItem(
+    for symbolOccurrence: SymbolOccurrence,
+    in index: CheckedIndex,
+    copiedFileMap: CopiedFileMap
+  ) throws -> WorkspaceSymbolItem {
+    let location: Location
+    if capabilityRegistry!.clientHasWorkspaceGetReferenceDocumentSupport,
```
ahoppen: Let’s not force unwrap `capabilityRegistry` here.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1767,124 @@ extension SourceKitLSPServer {
...
+        uri: DocumentURI(components.url!),
+        range: Range(Position(line: 0, utf16index: 0))
+      )
+    } else {
+      let line = symbolOccurrence.location.line
+      let col = symbolOccurrence.location.utf8Column
+      let symbolPosition = Position(
+        line: line > 0 ? line - 1 : 0,  // 1-based -> 0-based
+        // Technically we would need to convert the UTF-8 column to a UTF-16 column. This would require reading the
+        // file. In practice they almost always coincide, so we accept the incorrectness here to avoid the file read.
```
ahoppen: We have 3 places to which we copied this comment now. Should we add a conversion function (eg. as a static function on `DocumentSnapshot` so it lives with the other conversion functions) that does this?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1767,124 @@ extension SourceKitLSPServer {
...
+        moduleFilePath: symbolOccurrence.location.documentUri
+      )
+      var components = URLComponents(url: try urlData.uri.arbitrarySchemeURL, resolvingAgainstBaseURL: false)!
+      let usr = symbolOccurrence.symbol.usr
+      if !usr.isEmpty {
+        components.fragment = usr
+      }
+      location = Location(
+        uri: DocumentURI(components.url!),
+        range: Range(Position(line: 0, utf16index: 0))
```
ahoppen: If we used `WorkspaceSymbol` instead of `SymbolInformation` the we could return a URI without a location, removing any ambiguity on the client side whether they should jump to the location returned here or whether they should  use the location from the `getReferenceDocument` request.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1983,7 +2087,40 @@ extension SourceKitLSPServer {
...
+        primaryFile: primaryFile
+      )
+      guard let newBaseURL = try? newData.uri.arbitrarySchemeURL else { return nil }
+      var components = URLComponents(url: newBaseURL, resolvingAgainstBaseURL: false)!
+      components.fragment = req.uri.arbitrarySchemeURL.fragment
+      return GetReferenceDocumentRequest(uri: DocumentURI(components.url!))
+    }
+
+    for workspace in workspaces {
+      guard let mainFile = await workspace.buildServerManager.mainFiles(containing: data.buildSettingsFrom).first else {
```
ahoppen: Should we sort this before calling `.first` so that we get consistent results?

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,198 @@
...
+      public func myFunction() {}
+      """,
+      indexSystemModules: true
+    )
+
+    let response = try await project.testClient.send(WorkspaceAllSymbolNamesRequest())
+
+    XCTAssert(response.names.contains("MyStruct"), "Expected 'MyStruct' in allSymbolNames")
+    XCTAssert(response.names.contains("myFunction()"), "Expected 'myFunction()' in allSymbolNames")
+    XCTAssert(response.names.contains("String"), "Expected stdlib 'String' in allSymbolNames")
```
ahoppen: You can use `assertContains`, which will produce nicer failure messages if the string is not contained.

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  func testWorkspaceSymbolInfoUsesGeneratedInterfaceURIForStdlibSymbol() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      "let x: String = \"\"",
```
ahoppen:
```suggestion
      """
      let x: String = ""
      """,
```

Same below

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,198 @@
...
+      "Expected a 'String' item with a sourcekit-lsp:// location"
+    )
+    XCTAssertEqual(info.location.uri.scheme, "sourcekit-lsp")
+    XCTAssertNotNil(
+      info.location.uri.arbitrarySchemeURL.fragment,
+      "Expected USR embedded as URI fragment"
+    )
+  }
+
+  func testWorkspaceSymbolInfoGetReferenceDocumentHasSymbolPosition() async throws {
```
ahoppen: Wouldn’t it make more sense to merge this with `testWorkspaceSymbolInfoUsesGeneratedInterfaceURIForStdlibSymbol` by just adding a few more assertions to that test case?

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,198 @@
...
+  func testBinarySwiftModuleSymbolsNotIndexed() async throws {
+    guard let swiftc = await ToolchainRegistry.forTesting.default?.swiftc else {
+      throw XCTSkip("swiftc not found")
+    }
+
+    // Compile a Swift module to binary .swiftmodule only — no -index-store-path,
+    // so its symbols are never written to any index store.
+    let binaryModuleDir = FileManager.default.temporaryDirectory
+      .appendingPathComponent("BinaryModuleTest-\(UUID().uuidString)")
+    try FileManager.default.createDirectory(at: binaryModuleDir, withIntermediateDirectories: true)
```
ahoppen: If you use `withTestScratchDir`, we would get some nice functionality like being able to keep the temporary files by setting the `SOURCEKIT_LSP_KEEP_TEST_SCRATCH_DIR` environment variable.

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,198 @@
...
+    // so the binary module is importable in the consumer project regardless of SDK version.
+    #if os(macOS)
+    #if arch(arm64)
+    args += ["-target", "arm64-apple-macosx10.13"]
+    #elseif arch(x86_64)
+    args += ["-target", "x86_64-apple-macosx10.13"]
+    #endif
+    #endif
+    args += [sourceFile.path]
+    let process = Process()
```
ahoppen: Can’t you use `checkNonZeroExit` here?

Sources/BuildServerIntegration/BuildServerManager.swift
rintaro: It's worth to have shims only to keep the current call sites as-is.

Sources/SwiftExtensions/Array+SortAndDedupe.swift
rintaro: I am not really sure tbh. This is just to be on the safe side.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I wanted an option to perform `sortAndDedupe` without checking, because `isSortedAndUnique` is not free and there're case the caller _knows_ it's not isSortedAndUnique.

Sources/BuildServerIntegration/BuildServerManager.swift
rintaro: Agreed👍 https://github.com/swiftlang/sourcekit-lsp/pull/2624

Sources/SourceKitLSP/GeneratedInterfaceDocumentURLData.swift
rintaro: I was just moving the logic into the initializer. Modified the PR not to touch this for now.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: PR separated https://github.com/swiftlang/sourcekit-lsp/pull/2625

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Now we use `WorkspaceSymbol` for SDK symbols. And `workspaceSymbol/resolve` to open the generated interface, and resolve the location.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Oh I see this is created from `Set`. Updated to sort by `\.arbitrarySchemeURL.absoluteString` because that's seems to be the fastest (from `swift-foundation` source code)

Sources/SwiftExtensions/Array+SortAndDedupe.swift
ahoppen: OK, no objections to keeping it. Just wanted to understand it.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: Fair point 👍🏽

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1774,7 +1774,7 @@ extension SourceKitLSPServer {
   /// Handle a workspace/symbolNames request, returning the name list.
   func workspaceSymbolNames(_ req: WorkspaceSymbolNamesRequest) async throws -> WorkspaceSymbolNamesResponse {
     var symbols = try await self.workspaces.asyncFlatMap { workspace in
-      try await workspace.index(checkedFor: .deletedFiles)?.allSymbolNames() ?? []
+      try await workspace.index(checkedFor: .deletedFiles)?.unchecked.allSymbolNames() ?? []
```
ahoppen: Maybe it’s worth exposing `Workspace.uncheckedIndex` to avoid this dance. There’s one more location where we currently do it (when closing the index). That could be cleaned up as well.

Sources/SemanticIndex/CheckedIndex.swift
rintaro: Exposed `Workspace.uncheckedIndex`

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test Windows

Contributor Documentation/LSP Extensions.md
```diff
@@ -389,6 +389,65 @@ export interface IsIndexingResult {
...
 ```
 
+## `sourcekit/workspace/symbolInfo`
+
+New request that returns structured location information for a list of exact symbol names.
+
+Unlike the standard `workspace/symbol` request (which accepts a fuzzy query string), this request takes exact names — typically obtained from `sourcekit/workspace/symbolNames` — and returns all index occurrences for each name across every workspace.
+
+For each name the response contains zero or more `WorkspaceSymbolItem` values:
+- Source-file symbols carry a `SymbolInformation` with a `file://` URI and the exact 0-based line/column.
```
ahoppen: No need to explicitly call out the zero-based line:column here. It’s like that for every LSP position.

Contributor Documentation/LSP Extensions.md
```diff
@@ -389,6 +389,65 @@ export interface IsIndexingResult {
...
+- SDK/stdlib symbols carry a `WorkspaceSymbol` with `location: .uri(...)` pointing at the `file://` URI of the `.swiftinterface` or `.swiftmodule` file, with the fully-qualified module name as a `?module=` query parameter. The symbol's USR is stored in `data["usr"]`. Call `workspaceSymbol/resolve` to obtain the exact `Location` within the generated interface. The client must advertise `workspace.symbol.resolveSupport`; without it, the raw `file://` URI is returned as `SymbolInformation` instead.
```
ahoppen, line 400: Do we really want to make the structure of the URI part of the API contract? I would prefer to keep the URI opaque and just define that clients must call `workspaceSymbol/resolve` to resolve the location. This gives us more flexibility going forward.

Contributor Documentation/LSP Extensions.md
```diff
@@ -389,6 +389,65 @@ export interface IsIndexingResult {
...
+
+Every requested name is present in the response as a flat array; items carry their name in the `name` field of the `SymbolInformation` or `WorkspaceSymbol`.
+
+> [!IMPORTANT]
+> This request is experimental and may be modified or removed in future versions of SourceKit-LSP without notice. Do not rely on it.
```
ahoppen, line 405: Other new experimental LSP requests have an associated experimental feature that the client needs to enable. The idea is that using these requests should really be a deliberate choice and you shouldn’t start to accidentally call them. Maybe we should do the same here.

Contributor Documentation/LSP Extensions.md
```diff
@@ -389,6 +389,65 @@ export interface IsIndexingResult {
 }
 ```
 
+## `sourcekit/workspace/symbolInfo`
```
ahoppen, line 392: I just realized that `sourceKit/_isIndexing` capitalizes the `K`. We are very inconsistent in the naming of experimental requests at the moment but just thinking if that would be a reason to converge to a capitalized `K`. I leave the final decision to you.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -114,6 +114,18 @@ package final actor CapabilityRegistry {
...
 
+  package nonisolated var clientHasWorkspaceGetReferenceDocumentSupport: Bool {
+    return clientHasExperimentalCapability(GetReferenceDocumentRequest.method)
+  }
+
+  /// Whether the client supports `workspaceSymbol/resolve` for lazy location resolution.
+  /// Requires `workspace.symbol.resolveSupport.properties` to contain `"location"` or `"location.range"`.
+  package nonisolated var clientSupportsWorkspaceSymbolResolve: Bool {
+    return clientCapabilities.workspace?.symbol?.resolveSupport?.properties.contains(where: {
+      $0 == "location" || $0.hasPrefix("location.")
```
ahoppen: Doc comment says that it should contain `location.range` but you are just checking for `location.` here.

Sources/SourceKitLSP/MessageHandlingDependencyTracker.swift
```diff
@@ -244,6 +244,12 @@ package enum MessageHandlingDependencyTracker: QueueBasedMessageHandlerDependenc
       self = .freestanding
     case is WorkspaceSymbolResolveRequest:
       self = .freestanding
+    case is WorkspaceSymbolNamesRequest:
+      self = .freestanding
+    case is WorkspaceSymbolInfoRequest:
+      self = .freestanding
+    case is WorkspaceSymbolResolveRequest:
+      self = .freestanding
```
ahoppen, line 250: We already have this case above.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
     return try await languageService.signatureHelp(req)
   }
 
+  /// Handle a workspace/symbolNames request, returning the name list.
+  func workspaceSymbolNames(_ req: WorkspaceSymbolNamesRequest) async throws -> WorkspaceSymbolNamesResponse {
+    var symbols = try await self.workspaces.asyncFlatMap { workspace in
```
ahoppen: Would it make sense to `concurrentMap` here?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
...
+          name: symbolOccurrence.symbol.name,
+          kind: symbolOccurrence.symbol.kind.asLspSymbolKind(),
+          containerName: containerName,
+          location: .uri(.init(uri: DocumentURI(locationURL))),
+          data: data
+        )
+      )
+    }
+
+    guard let symbolLocation = symbolOccurrence.location.lspLocation else { return nil }
```
ahoppen, line 1842: What happens if the location is inside a `.swiftmodule`. Do we start opening that? Users won’t be able to see any useful information in it, I think.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
...
+        URLQueryItem(name: "module", value: symbolOccurrence.location.moduleName)
+      ]
+      guard let locationURL = urlComponents.url else {
+        return nil
+      }
+
+      // Data: {"usr": "<usr>"}
+      // Used by `workspaceSymbol/resolve` to find the symbol in the interface.
+      let usr = symbolOccurrence.symbol.usr
+      let data: LSPAny? = usr.isEmpty ? nil : .dictionary(["usr": .string(usr)])
```
ahoppen: Instead of this string-based serialization and deserialization in the resolve request, I’d prefer to introduce a new struct that conforms to `LSPAnyCodable` and which then represents the data.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
...
+  ///   `WorkspaceSymbol` with `location: .uri(file:// URL?module=...)` and the USR in `data`, provided
+  ///   the client advertises `workspace.symbol.resolveSupport`. The client should call
+  ///   `workspaceSymbol/resolve` to obtain the exact location within the interface.
+  ///   Without that capability the raw `file://` URI from the index record is returned instead.
+  ///
+  /// Every requested name is present as a key in the response, mapping to an empty array when there
+  /// are no occurrences.
+  func workspaceSymbolInfo(_ req: WorkspaceSymbolInfoRequest) async throws -> WorkspaceSymbolInfoResponse {
+    var result: [WorkspaceSymbolItem] = []
+    for workspace in workspaces {
```
ahoppen: Should we loop through the workspaces concurrently?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
...
+  /// If the symbol has a `location: .uri(moduleFileURL?module=...)` (as emitted by
+  /// `workspace/symbolInfo` for SDK/stdlib symbols), opens the generated Swift interface, resolves
+  /// the symbol position using `data["usr"]`, and returns the symbol with `location: .location(...)`.
+  /// Symbols with an already-resolved `location: .location(...)` are returned unchanged.
+  func workspaceSymbolResolve(_ req: WorkspaceSymbolResolveRequest) async throws -> WorkspaceSymbol {
+    var symbol = req.workspaceSymbol
+    guard
+      case .uri(let uriOnly) = symbol.location,
+      let urlComponents = URLComponents(url: uriOnly.uri.arbitrarySchemeURL, resolvingAgainstBaseURL: false),
+      let fullModuleName = urlComponents.queryItems?.first(where: { $0.name == "module" })?.value
```
```suggestion
      let fullModuleName = urlComponents.queryItems?.last(where: { $0.name == "module" })?.value
```

I think we should always make the last `module` query parameter take precedence if there are multiple.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1763,6 +1771,187 @@ extension SourceKitLSPServer {
...
+    }
+
+    let usr: String? =
+      if case .dictionary(let dataDict) = symbol.data, case .string(let usr) = dataDict["usr"] {
+        usr
+      } else {
+        nil
+      }
+
+    let moduleFileURI = DocumentURI(filePath: urlComponents.path, isDirectory: false)
```
ahoppen: `path` will give you the path representation as it’s written in the URL, which will not necessarily match how the file would be written for the file system on Windows. I can’t think of any issue here but I know that these sort of transformations are more finicky than you think.

Instead, I’d prefer to modify `URLComponents` to drop the `query` part and then construct a URL from it again. That should guarantee that all the other parts of the URL are retained verbatim.

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,209 @@
...
+      queryItems?.first(where: { $0.name == "module" })?.value,
+      "URI should contain a ?module= query parameter"
+    )
+    XCTAssertFalse(moduleParam.isEmpty, "?module= query parameter should be non-empty")
+    XCTAssertNil(urlComponents.fragment, "URI should not contain a fragment")
+
+    guard case .dictionary(let dataDict) = symbol.data,
+      case .string(let usr) = dataDict["usr"]
+    else {
+      XCTFail("Expected data[\"usr\"] string, got \(String(describing: symbol.data))")
```
ahoppen: Do we really need to check this? I would consider the exact structure of `data` as an implementation detail and everything is fine as long as `/resolve` works.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I considered that, but we already have prior arts using `LSPAny`, such as `inlayHintResolve` and `completionItemResolve`. Since it's just `"usr"`, I don't think it's worth creating a `struct`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: It's up to the client. The editor probably shows some errors:

Screenshot 2026-04-29 at 6 41 35 AM

Not much we can do here.

Although this function is used from the standard `workspace/symbol` request, it probably won't to reach `.swiftmodule` because we don't index user binary modules.
If the client request our `workspace/symbolInfo` extension, it _should_ support `workspaceSymbol/resolve`, so I don't think this would be an issue.

Contributor Documentation/LSP Extensions.md
rintaro: Let's rename it too :) https://github.com/swiftlang/swift-tools-protocols/pull/48

Contributor Documentation/LSP Extensions.md
rintaro: Some clients want to show module/group names in the candidate list. So we need to include that information somehow in the response in a way clients can read. Yes it's an API contract. We have seveal options: 
* Use `.data` - in general `data` should be opaque to client.
* Add dedicated (extension) field (e.g. `module`) to `WorkspaceSymbol` - I want to avoid this at all costs. 
* Include module names to `WorkspaceSymbol.containerName` - currently we're using this field to show  nesting types, and this field is meant to be displayed as-is in the client. I didn't come up with nice format for including module names. 
* Use `URL` - I think this is the least bad option

Contributor Documentation/LSP Extensions.md
rintaro: We're not doing that for every extension requests. I don't think we need that for these requests either.

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
rintaro: Removed. It's not a part of the API surface.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Agreed.

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test macOS

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: IIRC in `inlayHintResolve` and `completionItemResolve`, we only deserialize the URI using a string-based deinitializer because we don’t want to make assumptions about other fields to find the language service to handle the requests.

Since we control the entire encoding and decoding side here, I think a struct would make sense, if only to communicate that it’s really only the `usr` that gets included in the data here.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I would argue that sending the editor to a file we are pretty certain it can’t handle because it’s not a text file is a fairly bad user experience. My preference would be to not show the symbol at all in this case. The user won’t be able to get any useful information out if that `.swiftmodule` file.

Sources/SourceKitLSP/CapabilityRegistry.swift
ahoppen: This is still open.

Contributor Documentation/LSP Extensions.md
ahoppen: I personally would guard every experimental request behind an experimental feature. The fact that some aren’t is because they pre-date the experimental features, I think. Would like to clean that up at some point as well, if possible.

Contributor Documentation/LSP Extensions.md
ahoppen: I would argue that the `data` is more structured than the URI and I’d prefer that. Also, I personally don’t see a big issue with extending `WorkspaceSymbol` and would have probably done that.

If we want to go the URI route: Could we just say that the returned URI may contain a `module` query parameter and keep the rest of the URI opaque?

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I personally think _not_ showing the symbol is worse. 

>  The user won’t be able to get any useful information out if that .swiftmodule file.

Users can still know the symbol exists in that `.swiftmodule`.

ahoppen: I just think that it’ll feel broken to the user. But I see your point. and if you prefer to open the `.swiftmodule`, I don’t have super strong objections against it.

Contributor Documentation/LSP Extensions.md
rintaro: I disagree. `ExperimentalFeature` is a part of `SourceKitLSPOptions`. Although it can technically be specified by `IntializeRequest.initializationOptions`, usually it's specified via `.sourcekit-lsp/config.json`. It's not under control of the client editor.

In the current code base, we only gate `sourcekit/isIndexing`, `workspace/_setOptions`, `workspace/_sourceKitOptions`, `workspace/_outputPaths`, and a couple of fields in `workspace/synchronize`. They are all _control_ requests and are only intended to be used by our test cases.

I don't think we should gate the request itself with the feature flag, unless the request is a server control nob or really internal testing requests.

Sources/SourceKitLSP/CapabilityRegistry.swift
rintaro: I added some explanation for we're doing this.

Contributor Documentation/LSP Extensions.md
rintaro: The file path is actually a contract. Clients may look into the file path and derive the characteristics of the SDK E.g. "macOS 26.0 SDK". The way of the presentation is up-to the client so we don't want to return the exact text to show.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Type/call hierarchy item data as well.
Anyway, I decided to go with `struct`, here's the groundwork https://github.com/swiftlang/sourcekit-lsp/pull/2637 . I will add the `struct` to the added `MessageMetadata.swift` there.

Contributor Documentation/LSP Extensions.md
rintaro: We decided not to guard these requests with "experimental features"

Sources/SourceKitLSP/CapabilityRegistry.swift
rintaro: We decided to require `location` or both `location.uri` and `location.range`

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I'm going with bare `.swiftmodule` interface for now. VS Code or other clients wouldn't see it anyway for now because regular `workspace/symbol` request doesn't SDK symbols at this point.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Updated

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test Windows

Contributor Documentation/LSP Extensions.md
```diff
@@ -389,6 +389,65 @@ export interface IsIndexingResult {
...
+
+New request that returns structured location information for a list of exact symbol names.
+
+Unlike the standard `workspace/symbol` request (which accepts a fuzzy query string), this request takes exact names — typically obtained from `sourcekit/workspace/symbolNames` — and returns all index occurrences for each name across every workspace.
+
+For each name the response contains zero or more `WorkspaceSymbolItem` values:
+- Source-file symbols carry a `SymbolInformation` with a `file://` URI and the exact position.
+- SDK/stdlib symbols carry a `WorkspaceSymbol` with `location: .uri(...)` pointing at the `file://` URI of the `.swiftinterface` or `.swiftmodule` file, with the fully-qualified module name as a `?module=` query parameter. The symbol's USR is stored in `data["usr"]`. Call `workspaceSymbol/resolve` to obtain the exact `Location` within the generated interface. The client must advertise `workspace.symbol.resolveSupport`; without it, the raw `file://` URI is returned as `SymbolInformation` instead.
```
ahoppen, line 400: As discussed offline, I would like to not make the type of URLs that are returned for specific symbols part of the API contract. Instead, we should a general sentence like:

> URLs may contain a `module` query parameter that informs the client of the name that is represented by the file.

Documentation/Jump to Definition.md
```diff
@@ -0,0 +1,176 @@
...
+  │                                          │   when it reaches 0]
+```
+
+1. **Definition** — the client requests the definition of the symbol
+   at the cursor. For source-defined symbols the server returns a
+   `file://` URI with the exact source location. For SDK/stdlib
+   symbols it returns a `sourcekit-lsp://` URI and sets `range` to
+   the symbol's position within the generated interface (computed
+   server-side via `editor.find_usr`).
+2. **Content retrieval** — the client fetches the generated interface
```
ahoppen: Let’s highlight that this step only happens for `sourcekit-lsp://` URLs.

Documentation/Jump to Definition.md
```diff
@@ -0,0 +1,176 @@
...
+   via `workspace/getReferenceDocument` to display its content. The
+   client scrolls to `range` from the definition response.
+3. **Open notification** — once the client opens the tab it sends
+   `textDocument/didOpen` with the interface content (already fetched
+   in step 2) as `text`, `languageId` set to `"swift"`, and `version`
+   set to `1`. The server increments the ref count in
```
ahoppen: The client doesn’t necessarily need to set the version to 1. It just needs to be monotonically increasing as the file is modified (which it isn’t in this case). I’d just remove the mention of version here.

Documentation/Jump to Definition.md
```diff
@@ -0,0 +1,176 @@
...
+  groupName:     <from SymbolDetails.systemModule>,
+  symbolUSR:     <symbol.usr>,
+  originatorUri: <the file the cursor is in>
+)
+```
+
+### 2. `openGeneratedInterface`
+
+`definitionInInterface` delegates to
+`SwiftLanguageService.openGeneratedInterface`, which:
```
ahoppen: General feedback on this document: It breaks lines very aggressively. Since the rest of the code base allows 120 columns, I think it would be good to also use 120 columns here or just not have a column limit for Markdown files at all. Makes it more readable in its raw format.

Documentation/Jump to Definition.md
```diff
@@ -0,0 +1,176 @@
...
+   - `buildSettingsFrom` is set to `originatorUri.buildSettingsFile`
+     — the build settings file of the **requesting source file**, not
+     the module file. This ensures sourcekitd uses the same compiler
+     arguments as the file that triggered the request.
+2. Calls `generatedInterfaceManager.position(ofUsr:in:)` to find the
+   symbol's position within the generated interface (see below).
+3. Returns `GeneratedInterfaceDetails(uri: sourcekit-lsp://...,
+   position: <symbol position>)`.
+
+The URI has no USR fragment. The position is returned separately and
```
ahoppen: I don’t think that we have any URIs with USR fragments anymore, right? So this is a moot point. 

Same below.

Documentation/Jump to Definition.md
```diff
@@ -0,0 +1,176 @@
...
+sourcekit-lsp://generated-swift-interface/Swift.String.swiftinterface
+  ?moduleName=Swift
+  &groupName=String
+  &sourcekitdDocument=Swift.String.12345678
+  &buildSettingsFrom=file:///path/to/main.swift
+```
+
+The `range` in the returned `Location` carries the symbol's position
+in the interface, so the client knows where to scroll without calling
+`workspace/getReferenceDocument` first.
```
ahoppen: This is also outdated.

Documentation/Open Quickly.md
ahoppen, line 1: I think these should live in Contributor Documentation. Users of SourceKit-LSP don’t really care about these.

Documentation/Open Quickly.md
```diff
@@ -0,0 +1,247 @@
...
+  }
+```
+
+Without that capability, the raw `file://` URI of the `.swiftinterface` or `.swiftmodule` file from the index record is returned as `SymbolInformation` instead.
+
+The response is a flat array of `WorkspaceSymbolItem` values. Each item carries the symbol name in its `name` field.
+
+### `workspaceSymbol/resolve` — Location Resolution
+
+Resolves the lazy location of a `WorkspaceSymbol` returned by `sourcekit/workspace/symbolInfo`. The server parses `moduleName` and `groupName` from the `?module=` query parameter of the location URL, reads `usr` from the `data` dictionary, opens the generated Swift interface for the symbol's module, finds the symbol's position using the USR, and returns the same symbol with `location` replaced by a full `Location` (URI + range).
```
ahoppen: I find it weird if the server users information from both the query parameter *and* the data, ie. from two different sources. Could we store `moduleName` and `groupName` in `data` as well, so that 
1. We don’t need to destruct it again
2. We store all information used for `resolve` in data and thus in one data structure.

That would also make the `module` query parameter purely for display purposes, which may open up improvements in the future if we want to modify it.

ahoppen: I think it’s worth highlighting here that the URI gets replaced.

Documentation/Open Quickly.md
ahoppen, line 1: Does not have to be part of this PR but I feel like both of these overview documents are somewhat repetitive and also sometimes go too deep into how the actual implementation is written, which leaves the susceptible for becoming out-of-date. It would be great if we could condense them somewhat.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+  ///   the client advertises `workspace.symbol.resolveSupport`. The client should call
+  ///   `workspaceSymbol/resolve` to obtain the exact location within the interface.
+  ///   Without that capability the raw `file://` URI from the index record is returned instead.
+  ///
+  /// Every requested name is present as a key in the response, mapping to an empty array when there
+  /// are no occurrences.
+  func workspaceSymbolInfo(_ req: WorkspaceSymbolInfoRequest) async throws -> WorkspaceSymbolInfoResponse {
+    let canUseWorkspaceSymbolResolve = self.capabilityRegistry?.clientSupportsWorkspaceSymbolResolve ?? false
+
+    var groupedResultPerWorkspace = await workspaces.concurrentMap { workspace -> [String: [WorkspaceSymbolItem]] in
```
ahoppen, line 1871: Do we need to have the grouped results here at all or could we just collect a flat list here already?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+      guard let index = await workspace.index(checkedFor: .deletedFiles) else {
+        return [:]
+      }
+      var result: [String: [WorkspaceSymbolItem]] = [:]
+      let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
+      for name in req.names {
+        if Task.isCancelled { return [:] }
```
ahoppen, line 1878: Could this be `try Task.checkCancellation()`. I think at the moment we might return incorrect results if the users cancels the request, we thus return an empty array here but never check cancellation again, so we’d return an incomplete list instead of an error.

Same below.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+        var symbols: [SymbolOccurrence] = []
+        _ = orLog("getting symbol information") {
```
ahoppen: `orLog` strings typically start with an uppercase letter in the codebase.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+  /// If the symbol has a `location: .uri(moduleFileURL?module=...)` (as emitted by
+  /// `workspace/symbolInfo` for SDK/stdlib symbols), opens the generated Swift interface, resolves
+  /// the symbol position using `data["usr"]`, and returns the symbol with `location: .location(...)`.
+  /// Symbols with an already-resolved `location: .location(...)` are returned unchanged.
+  func workspaceSymbolResolve(_ req: WorkspaceSymbolResolveRequest) async throws -> WorkspaceSymbol {
+    var symbol = req.workspaceSymbol
+    guard
+      case .uri(let uriOnly) = symbol.location,
+      let urlComponents = URLComponents(url: uriOnly.uri.arbitrarySchemeURL, resolvingAgainstBaseURL: false),
+      let fullModuleName = urlComponents.queryItems?.last(where: { $0.name == "module" })?.value
```
ahoppen, line 1926: As I said in the overview documentation, I think it would be nicer if we stored all information that’s used for the resolve requests in data instead of getting some information from `data` and some from the URI. Also so we don’t need to parse the dot-separated module name again.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+      moduleName = fullModuleName
+      groupName = nil
+    }
+
+    let usr = WorkspaceSymbolData(fromLSPAny: symbol.data)?.usr
+
+    let moduleFileURI = DocumentURI(
+      {
+        var components = urlComponents
+        components.fragment = nil
```
ahoppen, line 1946: Remind me, do we set the fragment anywhere anymore?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1761,6 +1769,213 @@ extension SourceKitLSPServer {
...
+        symbol.location = .location(
+          Location(uri: details.uri, range: Range(details.position ?? Position(line: 0, utf16index: 0)))
+        )
+      }
+      return symbol
+    }
+
+    throw ResponseError.requestFailed(
+      "No source file found that imports \(fullModuleName); cannot open generated interface"
+    )
```
ahoppen, line 1978: Doesn’t have to be in this PR but I think we’d hit this if, say, a user used to import `CryptoKit` in their code base, then removes the import but `CryptoKit` is now still present in the index. Should we be able to open the module using fallback build settings in that case? Though we’d need to figure out how exactly those fallback builds settings should get computed.

Tests/SourceKitLSPTests/WorkspaceSymbolInfoTests.swift
```diff
@@ -0,0 +1,201 @@
...
+  func testBinarySwiftModuleSymbolsNotIndexed() async throws {
+    guard let swiftc = await ToolchainRegistry.forTesting.default?.swiftc else {
+      throw XCTSkip("swiftc not found")
+    }
+
+    // Compile a Swift module to binary .swiftmodule only — no -index-store-path,
+    // so its symbols are never written to any index store.
+    try await withTestScratchDir { binaryModuleDir in
+      let sourceFile = binaryModuleDir.appendingPathComponent("BinaryLib.swift")
+      try "public struct BinaryOnlyStruct {}".write(to: sourceFile, atomically: true, encoding: .utf8)
```
ahoppen: Use `writeWithRetry` here. Windows file writing is sometimes surprisingly finicky.

ahoppen (approved):
A few more comments, mostly in the documentation, otherwise looks good to me. Feel free to merge after addressing them.

Documentation/Jump to Definition.md
rintaro: I'm keeping this as-is. This is a _guidance_ for client implementation, they need to specify _some_ value.

Documentation/Open Quickly.md
rintaro: It's not really for "contributors" though. It's for client developers. Maybe we want another folder?

Documentation/Open Quickly.md
rintaro: `workspaceSymbol/resolve` also needs the `.swiftinterface` file path (for getting the main file). 
I don't like duplicated information, so I'm going to keep this as-is.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: We could, but I just want to keep items for the symbol "grouped" in the list. Not a API contract though.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: It's not easy because this is in `concurrentMap`. Let's just add `try Task.checkCancellation()` before creating the final flat list.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: No, this is  "just in case".

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: We don't parse the dot-separated module name in `workspace/symbolInfo`. we use  `symbolOccurrence.location.moduleName` as-is
```
      urlComponents.queryItems = [
        URLQueryItem(name: "module", value: symbolOccurrence.location.moduleName)
      ]
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Yeah, that's going to be a nice improvement.

Documentation/Jump to Definition.md
rintaro: This document is for jump to definition to SDK symbols that has `sourcekit-lsp://` scheme. I'm renaming the doc.

rintaro:
https://github.com/swiftlang/indexstore-db/pull/287
https://github.com/swiftlang/swift-tools-protocols/pull/47
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

Documentation/Open Quickly.md
ahoppen: Maybe we can rename the folders but the way they have been used so far is:
- Documentation: Documentation for *users* of SourceKit-LSP 
- Contributor Documentation: Documentation for anyone who needs to get into the details of how SourceKit-LSP operates, which includes client developers. I think there’s a significant overlap of details that client developers and SourceKit-LSP developers may need to understand, so I don’t think we should split this one up into two folders.

My main point is that this shouldn’t live in the general user documentation folder.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I thought `concurrentMap` also allowed throwing but maybe I mis-remembered.

`try Task.checkCancellation()` afterwards sounds good to me.

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I’d personally just keep it simple and remove it. There are plenty other cases where we also don’t remove fragments “just in case”. This code just makes me start to wonder what would add a fragment.

PR #2622, opened 2026-04-27
Add refactoring action to convert stored to computed properties

ahoppen:
This is the first syntactic refactoring action that needs to perform a cursor info request on `codeAction/resolve`, so the majority of this PR is to add infrastructure for that.

Based on https://github.com/swiftlang/sourcekit-lsp/pull/2496.

### DIFF
--- a/Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
+++ b/Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
@@ -1,0 +1,55 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftExtensions
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+extension ConvertStoredPropertyToComputed: SyntaxRefactoringCodeActionProvider {
+  static let title: String = "Convert Stored Property to Computed Property"
+
+  static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> VariableDeclSyntax? {
+    return scope.innermostNodeContainingRange?.as(VariableDeclSyntax.self)
+      ?? scope.innermostNodeContainingRange?.parent?.as(VariableDeclSyntax.self)
+  }
+
+  static func refactoringContext(
+    for node: VariableDeclSyntax,
+    in scope: SyntaxCodeActionScope
+  ) async -> SyntaxCodeActionContextResult<Context> {
+    guard node.bindings.contains(where: { $0.typeAnnotation?.type == nil }) else {
+      // All types are syntactically specified, we don't need to resolve the semantic type
+      return .context(Context())
+    }
+    guard let binding = node.bindings.only,
+      let identifier = binding.pattern.as(IdentifierPatternSyntax.self)?.identifier
+    else {
+      // We can only resolve type information for a single variable binding at the moment. If this is variable decl with multiple bindings, still
+      // offer the refactoring action and introduce placeholders for the type annotation.
+      return .context(Context())
+    }
+    if scope.resolveSupport?.canResolveEdit ?? false {
+      return .resolveEditLazily
+    }
+    // Cursor info reports type as `_` if it cannot determine the type.
+    if let type = try? await scope.cursorInfo(at: scope.snapshot.position(of: identifier.position)).only?.typeName,
+      type != "_"
+    {
+      return .context(Context(type: "\(raw: type)"))
+    }
+    return .context(Context())
+  }
+}
--- a/Sources/SwiftLanguageService/SwiftLanguageService.swift
+++ b/Sources/SwiftLanguageService/SwiftLanguageService.swift
@@ -25,7 +25,7 @@ import SemanticIndex
 package import SourceKitD
 package import SourceKitLSP
 import SwiftExtensions
-@_spi(ExperimentalLanguageFeatures) public import SwiftParser
+@_spi(ExperimentalLanguageFeatures) package import SwiftParser
 import SwiftParserDiagnostics
 package import SwiftSyntax
 package import ToolchainRegistry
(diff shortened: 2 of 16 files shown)


Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1276,6 +1276,139 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    )
+    XCTAssertNil(convertAction.edit)
+
+    let resolved = try await testClient.send(CodeActionResolveRequest(codeAction: convertAction))
+    XCTAssertEqual(
+      resolved.edit,
+      WorkspaceEdit(
+        changes: [
+          uri: [
+            TextEdit(range: positions["1️⃣"]..<positions["3️⃣"], newText: "\n  var x :Int{ 25 }")
```
ahoppen: @Padmashree06: It looks like the refactoring action in swift-syntax doesn’t produce correct whitespaces here (it should be `x: Int`, not `x :Int`). Would you be able to work on a fix for that in swift-syntax?

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/CodeActionTests.swift
Padmashree06: Yes sure! I will fix it!

Padmashree06: @ahoppen I have fixed the whitespace error in swift-syntax.

Sources/SwiftLanguageService/CodeActions/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,55 @@
...
+    else {
+      // We can only resolve type information for a single variable binding at the moment. If this is variable decl with multiple bindings, still
+      // offer the refactoring action and introduce placeholders for the type annotation.
+      return .context(Context())
+    }
+    if scope.resolveSupport?.canResolveEdit ?? false {
+      return .resolveEditLazily
+    }
+    // Cursor info reports type as `_` if it cannot determine the type.
+    if let type = try? await scope.cursorInfo(at: scope.snapshot.position(of: identifier.position)).only?.typeName,
```
rintaro: As I mentioned in https://github.com/swiftlang/sourcekit-lsp/pull/2588#discussion_r3206051684, I don't think we should allow arbitrary `cursorInfo` in `textDocument/codeAction` request.
If the client don't support `codeAction/resolve` for `.edit`, imo we should drop this action.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -973,16 +973,59 @@ extension SwiftLanguageService {
...
     }
     return await allSyntaxCodeActions.concurrentMap { provider in
-      return provider.codeActions(in: scope)
+      return await provider.codeActions(in: scope)
     }.flatMap { $0 }
   }
 
   package func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
-    return req.codeAction
+    guard let data = UnresolvedCodeActionData(fromLSPAny: req.codeAction.data) else {
```
rintaro, line 1020: IMO the whole mechanism should be something as simple as:

```swift
protocol SyntaxCodeActionProvider: SendableMetatype {
  // Provider name. default impl: { String(describing: Self.self) }
  static var name: String

  static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction]
  static func resolve(_ codeAction: CodeAction, in snapshot: DocumentSnapshot) -> CodeAction
}

struct UnresolvedCodeActionData: LSPAnyCodable {
  // Provider name generated the code action.
  var provider: String

  /// The document on which the code action should be applied.
  var document: VersionedTextDocumentIdentifier

  // Original `CodeAction.data` attached.
  var orignalData: LSPAny?
}
```

```swift
func codeAction(_ req: CodeActionRequest) -> CodeActionRequestResponse {
  let actions = [(provider: SyntaxCodeActionProvider.Type, action: CodeAction)] = ...

  return .codeActions(
    actions.map { item
      if shouldResolve(item.action) {
        var action = item.action
        action = UnresolvedCodeActionData(provider: provider.name, codeAction: action.data)
        return action
      } else {
        return item.action
      }
    }
  )
}

func codeActionResolve(_ req: CodeActionResolveRequest) -> CodeAction {
  var codeAction = req.codeAction 
  let data = UnresolvedCodeActionData(fromLSPAny: codeAction.data) else {
    return codeAction
  }
  let provider = getCodeActionProvider(name: data.provider)
  codeAction.data = data.originalData
  return provider.resolve(
    codeAction, in: getSnapshot(data.document) // Probably with some more context to request `cursorInfo`.
  ) 
}
```

Each `SyntaxCodeActionProvider` should be able to store abitrary `LSPAny` information necessary for resolving the action, as the `CodeAction.data`.

ahoppen:
Finally got back to this and updated it to have two distinct code action + resolve phases. Ready for another review.

ahoppen:
https://github.com/swiftlang/swift-tools-protocols/pull/72

@swift-ci Please test

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1010,7 +1017,40 @@ extension SwiftLanguageService {
...
+    guard let action = allSyntaxCodeActions.filter({ "\($0)" == data.action }).only else {
+      throw ResponseError.unknown("Could not find syntax action '\(data.action)' to resolve code action")
+    }
+    let snapshot = try documentManager.latestSnapshot(data.document.uri)
+    guard snapshot.version == data.document.version else {
+      throw ResponseError.unknown("Document was modified since between code action and resolve request")
+    }
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+
+    // Replay the code action without resolve support. This should now resolve all properties.
```
rintaro: This comment is not relevant anymore

Sources/SwiftSyntaxCodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -16,27 +16,87 @@ package import SourceKitLSP
...
+}
+
+extension SyntaxCodeActionProvider {
+  package static func resolve(
+    _ codeAction: CodeAction,
+    in scope: SyntaxCodeActionScope,
+    unresolvedData: LSPAny,
+    symbolInfo: (_ position: Position) async throws -> [SymbolDetails]
+  ) async throws -> CodeAction {
+    throw ResponseError.internalError("Resolve not implemented for '\(Self.self)'")
```
rintaro: This should return `codeAction` as-is. Clients may request `codeAction/resolve` regardless.

Or, make `resolve()` a requirement of `ResolvableSyntaxRefactoringCodeActionProvider` and cast the  provider type to it. If it's not a resolvable provider, just return `codeAction`.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1010,7 +1017,39 @@ extension SwiftLanguageService {
   }
 
   package func codeActionResolve(_ req: CodeActionResolveRequest) async throws -> CodeAction {
-    return req.codeAction
+    guard let data = UnresolvedCodeActionData(fromLSPAny: req.codeAction.data) else {
+      // We don't have any data to resolve the code action.
+      return req.codeAction
+    }
+    guard let action = allSyntaxCodeActions.filter({ "\($0)" == data.action }).only else {
```
rintaro: The name `action` makes me feel this is a `CodeAction`. Let's use `provider` as we do in `retrieveSyntaxCodeActions`
```suggestion
    guard let provider = allSyntaxCodeActions.filter({ "\($0)" == data.action }).only else {
```

Maybe we should rename `allSyntaxCodeActions` to `allSyntaxCodeActionProviders`

Sources/SwiftSyntaxCodeActions/SyntaxCodeActionProvider.swift
ahoppen: Changed it to return `codeAction`. I wanted to keep `ResolvableSyntaxRefactoringCodeActionProvider` as an internal type of `SwiftSyntaxCodeActions`.

Sources/SwiftLanguageService/SwiftLanguageService.swift
ahoppen: Renamed `allSyntaxCodeActions` as part of this PR. Good idea

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxCodeActions/SyntaxCodeActionProvider.swift
rintaro: I don't see this change (i.e. this is still throwing an error).

ahoppen: Not sure why this went missing. Sorry.

ahoppen:
https://github.com/swiftlang/swift-tools-protocols/pull/72

@swift-ci Please test

rintaro (approved):

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/72
@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift-tools-protocols/pull/72
@swift-ci Please test Windows

PR #2624, opened 2026-04-28
Factor out CopiedFileMap as a Sendable value type

rintaro:
Extract the copied-file adjustment logic from `BuildServerManager` into a `Sendable` value type `CopiedFileMap`. Call sites snapshot `cachedCopiedFileMap` once per request (hoisted before loops) instead of awaiting the actor on every iteration.

### DIFF
--- a/Sources/BuildServerIntegration/BuildServerManager.swift
+++ b/Sources/BuildServerIntegration/BuildServerManager.swift
@@ -346,6 +346,165 @@ private extension BuildServerSpec {
   }
 }
 
+/// Maps build-directory copies of source files back to their original source URIs.
+///
+/// During target preparation the build system may copy source files into the build directory
+/// (recorded as `copyDestinations` in `SourceFileInfo`). Index records and diagnostics can
+/// therefore reference the copy rather than the original. `CopiedFileMap` holds the reverse
+/// mapping — copy URI → original source URI — so that the server can translate those locations
+/// back to the files the user is actually editing before returning them to the client.
+package struct CopiedFileMap: Sendable {
+  private var map: [DocumentURI: DocumentURI]
+
+  init(map: [DocumentURI: DocumentURI]) {
+    self.map = map
+  }
+
+  /// Returns `true` if `uri` is a copy of a source file created during target preparation.
+  package func isCopiedFile(_ uri: DocumentURI) -> Bool {
+    return map[uri] != nil
+  }
+
+  /// Returns the original source URI for the given copied file URI, or `nil` if the URI is not a known copy.
+  package func originalURI(for uri: DocumentURI) -> DocumentURI? {
+    return map[uri]
+  }
+
+  /// Check if the URI referenced by `location` has been copied during the preparation phase. If so, adjust the URI to
+  /// the original source file.
+  package func locationAdjustedForCopiedFiles(_ location: Location) -> Location {
+    guard let originalUri = map[location.uri] else {
+      return location
+    }
+
+    if let fileUrl = originalUri.fileURL, !FileManager.default.fileExists(at: fileUrl) {
+      return location
+    }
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

rintaro:
(This will conflict with https://github.com/swiftlang/sourcekit-lsp/pull/2625)

rintaro:
@swift-ci Please test macOS

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -346,6 +346,159 @@ private extension BuildServerSpec {
...
 
+/// Maps build-directory copies of source files back to their original source URIs.
+///
+/// During target preparation the build system may copy source files into the build directory
+/// (recorded as `copyDestinations` in `SourceFileInfo`). Index records and diagnostics can
+/// therefore reference the copy rather than the original. `CopiedFileMap` holds the reverse
+/// mapping — copy URI → original source URI — so that the server can translate those locations
+/// back to the files the user is actually editing before returning them to the client.
+package struct CopiedFileMap: Sendable {
+  var map: [DocumentURI: DocumentURI]
```
ahoppen: Can this be `private`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1773,62 +1773,63 @@ extension SourceKitLSPServer {
...
       try index.forEachCanonicalSymbolOccurrence(
         containing: req.query,
         anchorStart: false,
         anchorEnd: false,
         subsequence: true,
         ignoreCase: true
       ) { symbol in
-        if Task.isCancelled {
-          return false
-        }
```
ahoppen, line 1790: Why did you remove this?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2255,7 +2257,9 @@ extension SourceKitLSPServer {
 
       return occurrences.compactMap { indexToLSPLocation($0.location) }
     }
-    let remappedLocations = await workspace.buildServerManager.locationsAdjustedForCopiedFiles(locations)
+    let remappedLocations = await workspace.buildServerManager.cachedCopiedFileMap.locationsAdjustedForCopiedFiles(
+      locations
+    )
```
ahoppen: I think this would be formatted better as follows. Same in a couple more cases (also in other files).

```suggestion
    let remappedLocations = await workspace.buildServerManager.cachedCopiedFileMap
      .locationsAdjustedForCopiedFiles(locations)
```

Sources/BuildServerIntegration/BuildServerManager.swift
rintaro: `private(set)`

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Added back

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/BuildServerIntegration/BuildServerManager.swift
ahoppen: I’d prefer if `map` really was a private implementation detail and we have dedicated accessors for all operations. That way we can also document more clearly what the key and the value of the dictionary is / make access to it read nicer.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

ahoppen (approved):

rintaro:
@swift-ci Please test Windows

PR #2633, opened 2026-05-06
Allow injection of a pre-initialized sourcekitd connection

rintaro:
Introduce `SourceKitDCore` as the protocol boundary between dylib lifecycle management and the high-level `SourceKitD` API. Its single lifecycle entry point, `initializeService(api:notificationCallback:)`, receives the already-loaded `sourcekitd_api_functions_t` from `SourceKitD.init(core:)`.

`SourceKitDCoreImpl` is the standard implementation: `init` opens the dylib; `initializeService` registers any plugin paths, calls `api.initialize()`, and wires the notification handler; `deinit` calls `shutdown()` and closes the handle. Pre-initialized conformances implement `initializeService` as a no-op.

Wire a `sourcekitdCoreInjector` hook through `Hooks` so an embedding host can return a pre-initialized `SourceKitDCore` for a given toolchain path, preventing `sourcekitd_initialize()` from being called a second time.

Declare `SourceKitDCoreForPlugin` at its use sites so each call site can express the exact deinit behavior it needs: `dlclose` for handles acquired via `RTLD_NOLOAD`, and `leak` for externally-owned handles.

### DIFF
--- a/Sources/SourceKitD/dlopen.swift
+++ b/Sources/SourceKitD/dlopen.swift
@@ -39,6 +39,14 @@ package final class DLHandle: Sendable {
   package static let rtldDefault = DLHandle(rawValue: Handle(handle: UnsafeMutableRawPointer(bitPattern: -2)!))
   #endif
 
+  /// Wrap a raw handle that is owned by an external system. The caller must call `leak()` before
+  /// `deinit` gets called prevent the assertion, since the external owner manages the lifecycle.
+  #if !os(Windows)
+  package static func adoptExisting(_ rawHandle: UnsafeMutableRawPointer) -> DLHandle {
+    DLHandle(rawValue: Handle(handle: rawHandle))
+  }
+  #endif
+
   fileprivate let rawValue: ThreadSafeBox<Handle?>
 
   fileprivate init(rawValue: Handle) {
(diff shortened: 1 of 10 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SourceKitD/dlopen.swift
```diff
@@ -39,6 +39,14 @@ package final class DLHandle: Sendable {
   package static let rtldDefault = DLHandle(rawValue: Handle(handle: UnsafeMutableRawPointer(bitPattern: -2)!))
   #endif
 
+  /// Wrap a raw handle that is owned by an external system. The caller must call `leak()` in
+  /// `deinit` to prevent the assertion, since the external owner manages the lifecycle.
```
rintaro: This is unused, but required for injecting custom `SourceKitDCore` via `sourcekitdCoreInjector` using externally `dlopen`ed handle.

ahoppen:
```suggestion
  /// Wrap a raw handle that is owned by an external system. The caller must call `leak()` before
  /// `deinit` gets called to prevent the assertion, since the external owner manages the lifecycle.
```

Tests/SourceKitLSPTests/SourcekitdCoreInjectorTests.swift
```diff
@@ -0,0 +1,125 @@
...
+
+    // Verify the SourceKitD instance held by SwiftLanguageService actually uses our core.
+    let usedCore = await swiftLS.sourcekitd.core
+    XCTAssert(usedCore as AnyObject === injectedCore)
+
+    // Verify that sourcekitd requests work through the injected core.
+    let response = try await testClient.send(
+      HoverRequest(textDocument: TextDocumentIdentifier(url), position: positions["1️⃣"])
+    )
+    _ = response  // discard - we just need it not to throw
```
ahoppen:
```suggestion
    // discard - we just need it not to throw
    _ = try await testClient.send(
      HoverRequest(textDocument: TextDocumentIdentifier(url), position: positions["1️⃣"])
    )
```

ahoppen (approved):
Two nitpicks, otherwise LGTM.

hamishknight (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #2637, opened 2026-05-07
Replace manual LSPAny data handling with typed structs

rintaro:
Introduce `HierarchyItemData` and `ResolveItemData` structs in `MessageMetadata.swift` conforming to `LSPAnyCodable`, replacing manual `LSPAny` dictionary construction and pattern matching for `CallHierarchyItem.data`, `TypeHierarchyItem.data`, `CompletionItem.data`, and `InlayHint.data`.

To use `HierarchyItemData`...

Move `CopiedFileMap` location/response adjustment logic out of `BuildServerIntegration` and into `adjusted(for: CopiedFileMap)` extensions on each LSP type (`Location`, `[Location]`, `LocationsOrLocationLinksResponse`, `WorkspaceEdit`, `CallHierarchyItem`, `TypeHierarchyItem`) in a new `LSP+CopiedFileMap.swift`.

The primitive URI remapping is extracted into `CopiedFileMap.adjustedURI(for:)`, which remains in `BuildServerIntegration`. This removes the awkward layering where a build-system module encoded knowledge of high-level LSP message shapes.

### DIFF
--- a/Sources/SourceKitLSP/LSP+CopiedFileMap.swift
+++ b/Sources/SourceKitLSP/LSP+CopiedFileMap.swift
@@ -1,0 +1,122 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+package import BuildServerIntegration
+@_spi(SourceKitLSP) package import LanguageServerProtocol
+
+extension Location {
+  package func adjusted(for copiedFileMap: CopiedFileMap) -> Location {
+    return Location(uri: copiedFileMap.adjustedURI(for: uri), range: range)
+  }
+}
+
+extension [Location] {
+  package func adjusted(for copiedFileMap: CopiedFileMap) -> [Location] {
+    return self.map { $0.adjusted(for: copiedFileMap) }
+  }
+}
+
+extension WorkspaceEdit {
+  package func adjusted(for copiedFileMap: CopiedFileMap) -> WorkspaceEdit {
+    var edit = self
+    if let changes = self.changes {
+      var newChanges: [DocumentURI: [TextEdit]] = [:]
+      for (uri, edits) in changes {
+        newChanges[copiedFileMap.originalURI(for: uri) ?? uri, default: []] += edits
+      }
+      edit.changes = newChanges
+    }
+    if let documentChanges = self.documentChanges {
+      edit.documentChanges = documentChanges.map { change in
+        switch change {
+        case .textDocumentEdit(var textEdit):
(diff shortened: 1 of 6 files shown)


Sources/SourceKitLSP/LSP+CopiedFileMap.swift
```diff
@@ -0,0 +1,121 @@
...
+    let adjustedSelectionLocation = Location(uri: uri, range: selectionRange).adjusted(for: copiedFileMap)
+    return TypeHierarchyItem(
+      name: name,
+      kind: kind,
+      tags: tags,
+      detail: detail,
+      uri: adjustedLocation.uri,
+      range: adjustedLocation.range,
+      selectionRange: adjustedSelectionLocation.range,
+      data: data
```
rintaro, line 98: I realized this `data` wasn't adjusted unlike `CallHierarchyItem`. Preserving the behavior for now.

Sources/SourceKitLSP/LSP+CopiedFileMap.swift
```diff
@@ -0,0 +1,121 @@
...
+        newChanges[copiedFileMap.originalURI(for: uri) ?? uri, default: []] += edits
+      }
+      edit.changes = newChanges
+    }
+    if let documentChanges = self.documentChanges {
+      edit.documentChanges = documentChanges.map { change in
+        switch change {
+        case .textDocumentEdit(var textEdit):
+          textEdit.textDocument.uri =
+            copiedFileMap.originalURI(for: textEdit.textDocument.uri) ?? textEdit.textDocument.uri
```
rintaro, line 43: Unlike others, `WorkspaceEdit` doesn't go through `adjustedURI(for:)` which checks the original file existence. This is intentional to keep the current behavior.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

Sources/SourceKitLSP/LSP+CopiedFileMap.swift
ahoppen: Is that the behavior we want? Or was it simply an oversight that this adjustment doesn’t check the original’s file existence?

Sources/SourceKitLSP/LSP+CopiedFileMap.swift
ahoppen: Oh, we should probably fix that. But makes sense to do that in a follow-up PR.

ahoppen (approved):
Nice improvements and well spotted for the two code paths that currently diverge. I think it makes sense to adjust them in a follow-up PR.

Sources/SourceKitLSP/LSP+CopiedFileMap.swift
rintaro: Not 100% sure but I assume this is just an oversight. I don't see a reason to emit `TextEdit` for non-existent files.

ahoppen: Would be great to address it in a follow-up PR if you get the chance.

rintaro: https://github.com/swiftlang/sourcekit-lsp/pull/2640

PR #2640, opened 2026-05-11
Fix and simplify CopiedFileMap adjustments

rintaro:
- `WorkspaceEdit.adjusted(for:)` was using `originalURI(for:) ?? uri`, which skips the file-existence check. Switch to `adjustedURI(for:)` to match others
- `TypeHierarchyItem.adjusted(for:)` was not adjusting the URI stored in `data`, unlike `CallHierarchyItem`.
- Simplify `TypeHierarchyItem`, `CallHierarchyItem`, and `LocationsOrLocationLinksResponse` adjusted implementations: since URI adjustment never changes the range, there's no need to construct intermediate `Location` values just to extract the URI.

### DIFF
--- a/Sources/SourceKitLSP/LSP+CopiedFileMap.swift
+++ b/Sources/SourceKitLSP/LSP+CopiedFileMap.swift
@@ -31,7 +31,7 @@ extension WorkspaceEdit {
     if let changes = self.changes {
       var newChanges: [DocumentURI: [TextEdit]] = [:]
       for (uri, edits) in changes {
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
> since URI adjustment never changes the range

Not 100% sure we can consider this as an invariant, though. But it is *copied*, the range should not change.

rintaro:
@swift-ci Please test Windows

ahoppen (approved):

Issue #2638, opened 2026-05-08
References response includes the definition itself


tothambrus11:
### Swift version

6.2.4

### Platform

Ubuntu

### Editor

VSCode

### Description

The references request should only contain uses of the definition, not the definition itself. It can add noise, and sometimes prevent immediate jumping to a single use, because we need to select between the definition itself and the usage).
<img width="909" height="310" alt="Image" src="https://github.com/user-attachments/assets/d2e13a92-14d7-4aa9-8664-3e3c0531e293" />

YooGyeongMo:
Hi @tothambrus11  and @ahoppen ,i'm demian from South korea.
This is my first contribution to sourcekit-lsp: picking this up as a great way to learn how the LSP server works internally. Thanks for filing the issue, and for the "good first issue" label that helped me find it!

I've opend a PR: #2644 

Took the approach of removing `.definition` from the roles set when `includeDeclaration` is `true`, and updated `testReferencesInMacro` to apply the same principle consistently across the macro case.                                                           

The PR description also covers two alternative approaches I tried (filtering by cursor position) and why I went with this one.                                                              
Would appreciate any feedback on the direction.

YooGyeongMo:
> Hi [@tothambrus11](https://github.com/tothambrus11) and [@ahoppen](https://github.com/ahoppen) ,i'm demian from South korea. This is my first contribution to sourcekit-lsp: picking this up as a great way to learn how the LSP server works internally. Thanks for filing the issue, and for the "good first issue" label that helped me find it!
> 
> I've opend a PR: [#2644](https://github.com/swiftlang/sourcekit-lsp/pull/2644)
> 
> Took the approach of removing `.definition` from the roles set when `includeDeclaration` is `true`, and updated `testReferencesInMacro` to apply the same principle consistently across the macro case.
> 
> The PR description also covers two alternative approaches I tried (filtering by cursor position) and why I went with this one. Would appreciate any feedback on the direction.

Update after looking into this more carefully:

I checked the LSP spec for `textDocument/references` and verified SourceKit-LSP's behavior with tests:

- `includeDeclaration: false` excludes the declaration from the result
- `includeDeclaration: true` includes the declaration in the result

Based on this, SourceKit-LSP appears to already be following the LSP behavior. The behavior reported here may be caused by the VS Code Swift extension sending `includeDeclaration: true` for "Find All References", which would make this more of a client-side issue.

[...]

PR #2644, opened 2026-05-12
[Tests] Add coverage for ReferencesContext.includeDeclaration

YooGyeongMo:
## Description

Following the discussion in #2638, SourceKit-LSP's `textDocument/references` behavior appears to already comply with the LSP specification. When `includeDeclaration` is `true`, the declaration is included. When `includeDeclaration` is `false`, the declaration is excluded.

The behavior reported in #2638, where the definition appears in "Find All References" results in VS Code, appears to be caused by the VS Code Swift extension sending `includeDeclaration: true` for that command, rather than by a server-side bug ([verified here](https://github.com/swiftlang/sourcekit-lsp/issues/2638#issuecomment-4432518139)).

This PR reverts the earlier behavior change and instead adds test coverage for the `includeDeclaration: false` case, which wasn't previously covered. The `includeDeclaration: true` case is already covered by existing tests `testReferencesInMacro`, `CopyDestinationTests`, `IndexTests`, `SwiftPMIntegrationTests`.

## Tests added

- `testReferencesWithoutDeclaration`: verifies that `includeDeclaration: false` returns only use sites

The test passes against the current behavior.

## Note on the original issue

The UX concern, where the definition appears in the results, is best addressed on the client side. For example, the VS Code Swift extension could send `includeDeclaration: false` for "Find All References". I would be happy to follow up on `vscode-swift` after this PR is reviewed.

Related to #2638

### DIFF
--- a/Tests/SourceKitLSPTests/ReferencesTests.swift
+++ b/Tests/SourceKitLSPTests/ReferencesTests.swift
@@ -40,8 +40,34 @@ final class ReferencesTests: SourceKitLSPTestCase {
     XCTAssertEqual(
       response,
       [
-        Location(uri: project.fileURI, range: Range(project.positions["1️⃣"])),
+        Location(uri: project.fileURI, range: Range(project.positions["1️⃣"]))
+      ]
+    )
+  }
+
+  func testReferencesExcludesOwnDefinition() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      func 1️⃣foo() {}
+
+      func bar() {
+        2️⃣foo()
+        3️⃣foo()
+      }
+      """
+    )
+    let response = try await project.testClient.send(
+      ReferencesRequest(
+        textDocument: TextDocumentIdentifier(project.fileURI),
+        position: project.positions["1️⃣"],
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+    XCTAssertEqual(
+      response,
+      [
         Location(uri: project.fileURI, range: Range(project.positions["2️⃣"])),
+        Location(uri: project.fileURI, range: Range(project.positions["3️⃣"])),
       ]
     )
   }
--- a/Tests/SourceKitLSPTests/ReferencesTests.swift
+++ b/Tests/SourceKitLSPTests/ReferencesTests.swift
@@ -72,32 +72,4 @@ final class ReferencesTests: SourceKitLSPTestCase {
       ]
     )
   }
-
-  func testReferencesWithDeclaration() async throws {
-    let project = try await IndexedSingleSwiftFileTestProject(
-      """
-      func 1️⃣foo() {}
-
-      func bar() {
-        2️⃣foo()
-        3️⃣foo()
-      }
-      """
-    )
-    let response = try await project.testClient.send(
-      ReferencesRequest(
-        textDocument: TextDocumentIdentifier(project.fileURI),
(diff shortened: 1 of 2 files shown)


Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -40,8 +40,34 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+        position: project.positions["1️⃣"],
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+    XCTAssertEqual(
+      response,
+      [
         Location(uri: project.fileURI, range: Range(project.positions["2️⃣"])),
+        Location(uri: project.fileURI, range: Range(project.positions["3️⃣"])),
       ]
```
tothambrus11: If I understood the LSP specification correctly, the declarartion should be part of this response if and only if `includeDeclaration` is `true`. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references

You could check if this is already the behavior (write tests if not yet covered). Maybe the VSCode extension is sending the wrong request with `includeDeclaration: true`.

ahoppen:
The changes in your PR do not seem to match your PR description. In particular, it currently only contains test changes. Could you please make sure that they match?

YooGyeongMo:
Hi @ahoppen @tothambrus11 ,

Thanks for the feedback. Apologies for the mismatch — I've updated the description to match the current changes: test coverage only, with no server behavior change.

For context on how I arrived here:

I initially tried removing `.definition` from the roles set so that references would always exclude the symbol's own definition. After @tothambrus11 pointed out the LSP spec — that the declaration should be excluded only when `includeDeclaration` is `false` — I wrote tests for both modes to verify SourceKit-LSP's actual behavior:

- `includeDeclaration: false` excludes the declaration from the references result
- `includeDeclaration: true` includes the declaration in the references result

The `includeDeclaration: true` test failed with my earlier change, which made it clear that my original direction was breaking the spec rather than fixing a bug. So I reverted that change and turned the verification tests into this PR.

Based on this, SourceKit-LSP appears to already be following the LSP behavior for `textDocument/references`. The behavior reported in #2638 may be caused by the VS Code Swift extension sending `includeDeclaration: true` for "Find All References", which would make this more of a client-side issue.

Does this direction make sense to you?

YooGyeongMo:
@ahoppen Thanks for the update in #2638 

I saw that this now appears to be related to the VS Code extension sending `includeDeclaration: true`, while SourceKit-LSP is handling the request according to the LSP behavior.

Given that, would it still be useful to keep this PR as test-only coverage for the existing `includeDeclaration` behavior?

ahoppen:
Yeah, a test would be good here but one of the tests you added covers the default behavior, which should already be tested by other tests. Could you reduce this to only add a test that covers new behavior?

YooGyeongMo:
Done, thanks @ahoppen !

I removed the redundant `testReferencesWithDeclaration` test and kept only the `includeDeclaration: false` coverage.

The `includeDeclaration: true` behavior is already covered by existing tests, including `testReferencesInMacro`.

Please let me know if this looks good or if anything else is needed.

ahoppen (approved):
LGTM. Thanks.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2648, opened 2026-05-14
Preserve language service when closing generated interface

rintaro:
Fixes a bug where closing a generated interface document (e.g. a `sourcekit-lsp://` Swift interface) would remove the language service for the originating source file, breaking subsequent requests like hover.
Both the generated interface and its originating source file share the same `buildSettingsFile` key in `Workspace.languageServices`, so closing the interface caused the service to be removed even while the source document was still open.

Based on [#2598](https://github.com/swiftlang/sourcekit-lsp/pull/2598) by @edwardlee. The test has been updated to use `IndexedSingleSwiftFileTestProject` with `-ignore-module-source-info` to avoid CI failures on source-built toolchains, where sourcekitd resolves stdlib types to real source files rather than generated interfaces.

### DIFF
--- a/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
+++ b/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
@@ -395,6 +395,41 @@ final class SwiftInterfaceTests: SourceKitLSPTestCase {
     XCTAssertEqual(transformLocation.uri.scheme, "sourcekit-lsp")
     assertContains(transformLocation.uri.pseudoPath, "Foundation.NSAffineTransform.swiftinterface")
   }
+
+  func testClosingGeneratedInterfacePreservesOriginatingLanguageService() async throws {
+    let testClient = try await TestSourceKitLSPClient(
+      capabilities: ClientCapabilities(experimental: [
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

hamishknight (approved):

Tests/SourceKitLSPTests/SwiftInterfaceTests.swift
```diff
@@ -395,6 +395,43 @@ final class SwiftInterfaceTests: SourceKitLSPTestCase {
...
+
+  func testClosingGeneratedInterfacePreservesOriginatingLanguageService() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      func test(x: 1️⃣String) {
+        let y: 2️⃣Int = 1
+      }
+      """,
+      capabilities: ClientCapabilities(experimental: [
+        GetReferenceDocumentRequest.method: .dictionary(["supported": true])
```
ahoppen, line 407: Should be able to use `LSPAny` conformance to `ExpressibleByDictionaryLiteral`.

```suggestion
        GetReferenceDocumentRequest.method: ["supported": true]
```

ahoppen (commented):
LGTM, one small nitpick.

PR #2651, opened 2026-05-14
[6.4] Preserve language service when closing generated interface

rintaro:
Cherry-pick #2648 into `release/6.4.x`

* **Explanation**: Closing a generated document (e.g. a Swift interface opened via jump-to-definition) was incorrectly removing the language service for the originating source file. Both the generated interface and its originating file share the same `buildSettingsFile` key in `Workspace.languageServices`, so closing the interface evicted the service while the source document was still open. Subsequent requests (hover, completion, etc.) on the source file would silently fail. The fix guards `removeLanguageServices(for:)` so it only removes the service when no other open document shares the same key.
* **Scope**: Language service management
* **Risk**: Low. At worst, a language service that should have been released is retained slightly longer. This fix is a strict improvement over the status quo where the service was incorrectly dropped.. 
* **Issues**: rdar://176574004
* **Testing**: Added regression test cases
* **Reviewer**: Hamish Knight (@hamishknight)

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1457,7 +1457,9 @@ extension SourceKitLSPServer {
       await languageService.closeDocument(notification)
     }
 
-    workspace.removeLanguageServices(for: uri)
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

hamishknight (approved):

rintaro:
@swift-ci Please test Windows

PR #2653, opened 2026-05-17
[WIP]: experimental closure isolation inlay hints

jamieQ:
WIP.

depends on: https://github.com/swiftlang/swift/pull/89198

### DIFF
--- a/Sources/SKOptions/ExperimentalFeatures.swift
+++ b/Sources/SKOptions/ExperimentalFeatures.swift
@@ -50,6 +50,9 @@ public enum ExperimentalFeature: String, Codable, Sendable, CaseIterable {
   /// - Note: Internal option, for testing only
   case synchronizeCopyFileMap = "synchronize-copy-file-map"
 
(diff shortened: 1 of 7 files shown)


Steffeeen:
I just quickly looked at this, as I recently changed inlay hint handling in #2597.

The actual SourceKit request should probably be in `computeTypeInlayHints()` in `InlayHintManager.swift`. This allows doing both SourceKit requests concurrently. It also ensures that the results are cached to avoid flickering (see the linked PR and corresponding issue for more info).

For running the tests with a local build of SourceKit have a look at the [developer docs](https://github.com/swiftlang/sourcekit-lsp/blob/main/Contributor%20Documentation/Using%20a%20Local%20Build%20of%20SourceKit.md). They should describe your use case exactly. Feel free to ping me if you have any questions. After moving the SourceKit request to `computeTypeInlayHints()` you should also be able to reuse the test infrastructure of the other inlay hint tests. The easiest approach is probably to just add the type inlay hints into the asserts instead of trying to filter them out.

PR #2654, opened 2026-05-17
Scope language service lifetime to workspace

rintaro:
Previously `SourceKitLSPServer` owned all language service instances and managed an N-workspace-to-M-service mapping itself. This meant the server had to track which services belonged to which workspaces, decide when to shut them down, and special-case services that should survive workspace closure (`isImmortal`). At the same time, services held workspace-specific state — `DocumentationLanguageService` held a reference to `Workspace.buildServerManager`, `SwiftLanguageService` held `Workspace.semanticIndexManagerTask` — so sharing them across workspaces was never truly safe anyway.

This PR resolves both problems by introducing a clean ownership hierarchy: one server owns N workspaces, each workspace owns its own language services. Services are created on demand within a workspace and shut down when that workspace closes. Because each service now belongs to exactly one workspace, holding workspace-specific state is natural and correct.

Simplified `LanguageService` protocol:
- `initialize(_:)` and `clientInitialized(_:)` are removed; each service's `init` now performs its own initialization and capability registration.
- `isImmortal` is removed; each service is shut down with its workspace.
- The workspace parameter is removed from `canHandle(toolchain:)`; since each workspace has its own service pool, a service only needs to check the toolchain to decide whether it can be reused for a new document.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -74,8 +74,6 @@ package actor SourceKitLSPServer {
 
   let languageServiceRegistry: LanguageServiceRegistry
 
-  var languageServices: [LanguageServiceType: [any LanguageService]] = [:]
-
   package nonisolated let documentManager = DocumentManager()
 
   /// The `TaskScheduler` that schedules all background indexing tasks.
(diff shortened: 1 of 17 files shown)


Sources/SwiftLanguageService/SwiftCommand.swift
```diff
@@ -55,11 +55,4 @@ extension SwiftLanguageService {
       command.identifier
     }
   }
-
-  /// `SwiftLanguageService` is immortal because sourcekitd uses global state.
-  ///
-  /// Since all instances of `SwiftLanguageService` share the same underlying sourcekitd process,
-  /// shutting down and restarting would cause unnecessary overhead as the new instance would
-  /// just reinitialize the same global state. Instead, we keep the service alive.
-  package static var isImmortal: Bool { true }
```
rintaro, line 64: > shutting down and restarting would cause unnecessary overhead

`SourceKitD` is globally cached (`SourceKitDRegistry.shared`). In reality, the only overhead is setting up the stored properties e.g. `self.macroExpansionManager` or `self.generatedInterfaceManager`. and the instance memory.

Tests/SourceKitLSPTests/LocalSwiftTests.swift
```diff
@@ -1380,6 +1368,11 @@ final class LocalSwiftTests: SourceKitLSPTestCase {
       """,
       uri: uri
     )
+    let swiftLanguageService = try await unwrap(testClient.primaryLanguageService(for: uri) as? SwiftLanguageService)
```
rintaro, line 1371: Moved this because `primaryLanguageService(for: uri)` is populated only in `openDocument()`

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1571,12 +1393,19 @@ extension SourceKitLSPServer {
     for docUri in self.documentManager.openDocuments {
       preChangeWorkspaces[docUri] = await self.workspaceForDocument(uri: docUri)
     }
+    // Capture the workspaces that will be removed so we can shut down their services after.
+    var removedWorkspaces: [Workspace] = []
     await workspaceQueue.async {
       if let removed = notification.event.removed {
-        self.workspacesAndIsImplicit.removeAll { workspace in
+        let removedEntries = self.workspacesAndIsImplicit.filter { entry in
```
ahoppen: Do we need the `removedEntries` variable at all or could we immediately assign to `removedWorkspaces`?

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -449,20 +460,148 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
...
+    guard let sourceKitLSPServer else { return [] }
+    var result: [any LanguageService] = []
+    for serviceType in languageServiceRegistry.languageServices(for: language) {
+      if let service = existingLanguageService(serviceType, toolchain: toolchain) {
+        result.append(service)
+        continue
+      }
+
+      // Start a new service.
+      let service: (any LanguageService)? = await orLog("failed to start language service") {
```
ahoppen: Let’s uppercase the first letter of the log message for consistency.

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -449,20 +460,148 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
...
+      )
+    }
+
+    return services
+  }
+
+  /// Find or create the primary language service for the given URI and language.
+  ///
+  /// Convenience wrapper around the async `languageServices(for:_:)`. Use this only when the
+  /// document may not have been opened yet. Throws if no service is available.
```
ahoppen: Let’s not mention `async` here since this is also async, so it’s not worth posting attention to.

```suggestion
  /// Convenience wrapper around `languageServices(for:_:)` that throws if no service is available.. Use this only when the document may not have been opened yet. 
```

Sources/SourceKitLSP/Workspace.swift
```diff
@@ -449,20 +460,148 @@ package final class Workspace: Sendable, BuildServerManagerDelegate {
...
     return languageServices.value[uri.buildSettingsFile] ?? []
   }
 
-  /// The language service with the highest precedence that can handle the given document.
-  func primaryLanguageService(for uri: DocumentURI) -> (any LanguageService)? {
+  /// The primary language service for an open document.
+  ///
+  /// Convenience wrapper around the sync `languageServices(for:)`. Returns `nil` if the document
+  /// has not been opened or has already been closed.
+  package func primaryLanguageService(for uri: DocumentURI) -> (any LanguageService)? {
```
ahoppen, line 604: It’s a very subtle distinction that passing a `language` to `primaryLanguageService` may create a new language service while you only get existing language services when not passing a language.

Maybe all the functions that only return existing language services should be called `existing*`

ahoppen (approved):
Thanks. Nice clean-up. Just a few small comments

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: :+1: Refined to do this one-pass.

Sources/SourceKitLSP/Workspace.swift
rintaro: I agree this is pretty confusing. In fact, I see several places using `languageServices(for:_:)`/`primaryLanguageService(for:_:)` where they should be able to use other.

I'm going to address it in a follow up.

rintaro:
@swift-ci Please test

Sources/SourceKitLSP/Workspace.swift
ahoppen: Thanks 👍🏽

ahoppen (approved):

rintaro:
@swift-ci Please test Windows

PR #2655, opened 2026-05-17
provide proper start and end positions in the LSP protocol

femaref:
vscode provides an alternative action if the target of goto definition is in the same character range we requested. this requires the start and end in the LSP response to be properly set, so vscode can see that the current character is actually part of the request. Currently, this functionality only works when invoking "go to definition" on the start of the character range.

Screenshot 2026-05-18 at 00 09 05

https://github.com/user-attachments/assets/cdcda3c7-eed5-420f-9681-5904d7101ea1

https://github.com/user-attachments/assets/ac5909a1-65d0-4981-91ff-0623830ac25a

### DIFF
--- a/Sources/SwiftLanguageService/CursorInfo.swift
+++ b/Sources/SwiftLanguageService/CursorInfo.swift
@@ -77,8 +77,10 @@ struct CursorInfo {
           documentManager.latestSnapshotOrDisk(uri, language: .swift)
         }
       if let snapshot {
-        let position = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
-        location = Location(uri: uri, range: Range(position))
+        let length = (dict[keys.length] ?? 0)
+        let positionStart = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
+        let positionEnd = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column + length - 1)
+        location = Location(uri: uri, range: positionStart..<positionEnd)
       } else {
         logger.error("Failed to get snapshot for \(uri.forLogging) to convert position")
         location = nil
--- a/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
+++ b/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
@@ -1079,7 +1079,7 @@ final class BackgroundIndexingTests: SourceKitLSPTestCase {
     let project = try await SwiftPMTestProject(
       files: [
         "Lib/MyFile.swift": """
-        public func 1️⃣foo() {}
+        public func 1️⃣foo3️⃣() {}
         """,
         "MyExec/MyExec.swift": """
         import Lib
@@ -1119,7 +1119,7 @@ final class BackgroundIndexingTests: SourceKitLSPTestCase {
     let definition = try await project.testClient.send(
       DefinitionRequest(textDocument: TextDocumentIdentifier(uri), position: positions["2️⃣"])
     )
-    XCTAssertEqual(definition, .locations([try project.location(from: "1️⃣", to: "1️⃣", in: "MyFile.swift")]))
+    XCTAssertEqual(definition, .locations([try project.location(from: "1️⃣", to: "3️⃣", in: "MyFile.swift")]))
   }
 
   func testCrossModuleFunctionalityEvenIfLowLevelModuleHasErrors() async throws {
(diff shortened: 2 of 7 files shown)


rintaro:
@swift-ci please test

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -77,8 +78,12 @@ struct CursorInfo {
...
         }
       if let snapshot {
-        let position = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
-        location = Location(uri: uri, range: Range(position))
+        let positionStart = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column - 1)
+        let positionEnd = snapshot.positionOf(zeroBasedLine: line - 1, utf8Column: column + length - 1)
+        location = Location(
+          uri: uri,
+          range: Range(uncheckedBounds: (lower: positionStart, upper: positionEnd))
+        )
```
rintaro: Could you use `..<`? There's no reason to use "unchecked" version here.
```suggestion
        location = Location(uri: uri, range: positionStart..<positionEnd)
```

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -67,7 +67,8 @@ struct CursorInfo {
     let location: Location?
     if let filepath: String = dict[keys.filePath],
       let line: Int = dict[keys.line],
-      let column: Int = dict[keys.column]
+      let column: Int = dict[keys.column],
+      let length: Int = dict[keys.length]
```
rintaro: We know `length` is always there in cursorInfo response, but just to be on the safe side, could you do make it defaulted to zero? i.e. (`dict[keys.length] ?? 0`), so it's not going to be a part of `if` conditions.

rintaro (commented):
Nice fix.
Please update the test cases 🙏

femaref:
> Nice fix. Please update the test cases 🙏

apologies, I'm inexperienced with the tooling - it showed green checkmarks at the end, but that doesn't actually seem to be a global summary.

rintaro:
@swift-ci Please test

Sources/SwiftLanguageService/CursorInfo.swift
```diff
@@ -65,6 +65,7 @@ struct CursorInfo {
     }
 
     let location: Location?
+    let length = (dict[keys.length] ?? 0)
```
rintaro: Please move this inside `if let snapshot {` branch.

femaref:
I just realized it's more than that test failing.

femaref:
I think I have now caught everything. I'm still getting two more failures, but those also happen on `main`:


> Test Case '-[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithoutIndex]' started.
> /Users/femaref/assembly/sourcekit-lsp/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift:269: error: -[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithoutIndex] : XCTAssertTrue failed - Full line was: '@inlinable public func filter(_ isIncluded: (Element) throws -> Bool) rethrows -> [Element]
> '
> Test Case '-[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithoutIndex]' failed (1.413 seconds).
> Test Case '-[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithIndex]' started.
> /Users/femaref/assembly/sourcekit-lsp/Tests/SourceKitLSPTests/SwiftInterfaceTests.swift:288: error: -[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithIndex] : XCTAssertTrue failed - Full line was: '@inlinable public func filter(_ isIncluded: (Element) throws -> Bool) rethrows -> [Element]
> '
> Test Case '-[SourceKitLSPTests.SwiftInterfaceTests testJumpToSynthesizedExtensionMethodInSystemModuleWithIndex]' failed (0.829 seconds).
>

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
>  I'm still getting two more failures, but those also happen on main:

Those 2 tests fails if you're testing against older SDK where https://github.com/swiftlang/swift/pull/87020 has not been applied. We probably should update the tests to accept both signatures...

rintaro (approved):

PR #2656, opened 2026-05-17
[Test] Fix deployment target mismatch in SwiftPMBuildServerTests

rintaro:
- `testToolsets`: sourcekit-lsp is built against a just-built SwiftPM, which can use macOS 10.13 as the minimum deployment target for  packages with no explicit `platforms` requirement. The preparation  step runs the toolchain's `swift build`, which can use a higher macOS  minimum. The resulting module embeds a higher deployment target,  causing a spurious mismatch diagnostic when the other target imports  it.
- `testBasicSwiftArgs` (swiftbuild variant): `swift-build` can clamp  the deployment target to the SDK's `MinimumDeploymentTarget`, which  can be above the 10.13 `PackageModel.Platform.macOS.oldestSupportedVersion`

Fix these by explicit set the deployment target to 13.0


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2657, opened 2026-05-18
Use 'weak let' for immutable weak references (SE-0481)

rintaro:
Replace `weak var` with `weak let` for weak reference properties that are set in init and never reassigned, using the Swift 6.3 feature introduced by SE-0481.


rintaro:
@swift-ci Please test

hamishknight (approved):

ahoppen (approved):

PR #2659, opened 2026-05-19
Fix rename target selection for initializer calls

### LINKED ISSUE #2646: Rename on initializer reference does not work
### Swift version

Swift version 6.4-dev (LLVM d2079213f1d4451, Swift 82b7720768ba875)

### Platform

Arch Linux

### Editor

Visual Studio Code + Swift, latest versions

### Description

With the setup below, if
- the text caret is on `NewType` of `let myNew: NewType` and a rename is invoked, then the name in `file_two.swift` is updated as expected
- the text caret is on `NewType` in the rhs of the declaration `= NewType()` and a rename is invoked, then only the references in `file_one.swift` are updated.

Renaming from the type reference does work.

Renaming from the initializer invocation does not update the class name. It should update in the same way as the other rename.

**file_one.swift:**
```swift
func main() {
    let myNew: NewType = NewType()
}
```

**file_two.swift**
```swift
class NewType {
}
```

I tested this with Swift 6.3.1 and also the latest main-snapshot (retrieved via swiftly).
As editor I used the latest VSCode+Swift setup and also my own (Noctule).

The response to the `textDocument/rename` request did not contain the change to `file_two.swift.`

### Steps to Reproduce

Create the two files and start the rename in `file_one.swift` on the rhs of the declaration.

**file_one.swift:**
```swift
func main() {
    let myNew: NewType = NewType()
}
```

**file_two.swift**
```swift
class NewType {
}
```

### Logging

_No response_

Steffeeen:
Fixes #2646.

For the case described in the issue `symbolInfo()` returns multiple symbols:
- The actual struct/class/actor type
- The initializers of that type

With this PR, the struct/class/actor type is selected and used for the renaming.

I added two tests to check this behavior and added an additional marker in `testRenameInitializerLabels` to ensure that placing the cursor inside the arguments still renames the initializer.

### DIFF
--- a/Sources/SwiftLanguageService/Rename.swift
+++ b/Sources/SwiftLanguageService/Rename.swift
@@ -607,6 +607,19 @@ extension SwiftLanguageService {
     )
 
     guard let functionLikeRange = await findFunctionLikeRange(of: startOfIdentifierPosition, in: snapshot) else {
+      if let symbolInfo, symbolInfo.count == 2,
+        symbolInfo.first(where: { $0.kind == .constructor }) != nil,
+        let structClassEnumActorSymbolInfo = symbolInfo.first(where: {
+          // There is no SymbolKind for actors, they just use `.class`.
+          $0.kind == .struct || $0.kind == .class || $0.kind == .enum
+        })
+      {
+        // We have multiple symbols at the position but none of them have a function-like range. This can happen when
+        // invoking a rename from the name of a initializer call like `MyStruct(x: 1)` with the cursor inside `MyStruct`.
+        // In this case, we want to rename the struct/class/actor instead of the initializer.
+        return (startOfIdentifierPosition, structClassEnumActorSymbolInfo.usr, nil)
+      }
+
       return (startOfIdentifierPosition, symbolInfo?.only?.usr, nil)
     }
     if let onlySymbol = symbolInfo?.only, onlySymbol.kind == .constructor {
--- a/Tests/SourceKitLSPTests/RenameTests.swift
+++ b/Tests/SourceKitLSPTests/RenameTests.swift
@@ -258,7 +258,7 @@ final class RenameTests: SourceKitLSPTestCase {
       struct Foo {
         1️⃣init(4️⃣x: Int) {}
       }
-      Foo(x: 1)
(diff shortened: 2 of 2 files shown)


Sources/SwiftLanguageService/Rename.swift
```diff
@@ -607,6 +607,15 @@ extension SwiftLanguageService {
     )
 
     guard let functionLikeRange = await findFunctionLikeRange(of: startOfIdentifierPosition, in: snapshot) else {
+      if let symbolInfo, symbolInfo.count > 1 {
+        // We have multiple symbols at the position but none of them have a function-like range. This can happen when
+        // invoking a rename from the name of a initializer call like `MyStruct(x: 1)` with the cursor inside `MyStruct`.
+        // In this case, we want to rename the struct/class/actor instead of the initializer.
+        // There is no SymbolKind for actors, they just use `.class`.
+        let structClassActorSymbolInfo = symbolInfo.first { $0.kind == .struct || $0.kind == .class }
```
ahoppen: We might also have multiple symbol infos if the code is ambiguous (ie. doesn’t currently compile). In that case, we probably don’t want to offer the rename because we aren’t sure which of the two types the user wants to rename. So, I would prefer to really special case the type + initializer case here and only in that case pick the type’s symbol.

Also: `enum` can have initializers as well, so we should handle them here too.

ahoppen (commented):
Thanks for the fix. Overall looks good.

Sources/SwiftLanguageService/Rename.swift
Steffeeen: I added enum handling and reworked the code explicitly check for a type and initializer. However, I wasn't able to get multiple symbol infos by having non-compiling code. I always got no symbol infos back instead.

ahoppen (commented):
Thanks you. Looks good. 

You can get ambiguous results eg. when doing a cursor info on `x` here. The ambiguous handling logic in cursor logic is, unfortunately, only half-baked.

```swift
func foo() -> String { "a" }
func foo() -> Int { 1 }
func test() {
  let x = foo()
}
```

ahoppen:
@swift-ci Please test

PR #2660, opened 2026-05-20
Allow configuring `--pkg-config-path`

ahoppen:
Add a SourceKit-LSP configuration option equivalent to SwiftPM’s `--pkg-config-path`.


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #2661, opened 2026-05-20
Add `forOpenDocument` label to language service lookup methods

rintaro:
Rename `languageServices(for:)`, `primaryLanguageService(for:)`, and their internal counterparts to use the `forOpenDocument` label, so the precondition that the document must already be open is visible at call sites.

Also make `primaryLanguageService(forOpenDocument:)` throw instead of returning an optional, and switch several resolve-style handlers from the find-or-create `primaryLanguageService(for:_:)` to `primaryLanguageService(forOpenDocument:)`, since those call sites already have an open document in hand.


rintaro:
@swift-ci Please test

ahoppen (approved):
Nice improvement. Thanks

PR #2662, opened 2026-05-21
Migrate atomic call sites to Synchronization.Atomic

rintaro:
Replace `AtomicBool`/`UInt8`/`UInt32`/`Int32` from `ToolsProtocolsSwiftExtensions` with `Synchronization.Atomic<T>` where the storage is a static, module-level let, or class stored property.

For values captured by `@Sendable` closures (where Atomic's `~Copyable` nature prevents capture), use `ThreadSafeBox<T>` instead.

Also relax atomic memory orderings to `.relaxed` for ID counters and standalone flags, since nothing reads other memory based on the atomic's value.

`MultiEntrySemaphore.signaled` is the exception: it is used as a signal/wait primitive where waiters proceed to use state set up before `signal()`. Use `.releasing` on store and `.acquiring` on load so that pre-signal writes are visible to waiters that observe `true`.

related: https://github.com/swiftlang/swift-tools-protocols/pull/51

### DIFF
--- a/Sources/Diagnose/DiagnoseCommand.swift
+++ b/Sources/Diagnose/DiagnoseCommand.swift
@@ -222,8 +222,8 @@ package struct DiagnoseCommand: AsyncParsableCommand {
     let outputFileUrl = bundlePath.appending(component: "log.txt")
     try FileManager.default.createFile(at: outputFileUrl, contents: nil)
     let fileHandle = try FileHandle(forWritingTo: outputFileUrl)
-    let bytesCollected = AtomicInt32(initialValue: 0)
-    let processExited = AtomicBool(initialValue: false)
+    let bytesCollected = ThreadSafeBox<Int32>(initialValue: 0)
+    let processExited = ThreadSafeBox<Bool>(initialValue: false)
     // 50 MB is an average log size collected by sourcekit-lsp diagnose.
     // It's a good proxy to show some progress indication for the majority of the time.
     let expectedLogSize = 50_000_000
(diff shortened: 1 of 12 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/Diagnose/DiagnoseCommand.swift
```diff
@@ -222,8 +222,8 @@ package struct DiagnoseCommand: AsyncParsableCommand {
     let outputFileUrl = bundlePath.appending(component: "log.txt")
     try FileManager.default.createFile(at: outputFileUrl, contents: nil)
     let fileHandle = try FileHandle(forWritingTo: outputFileUrl)
-    let bytesCollected = AtomicInt32(initialValue: 0)
-    let processExited = AtomicBool(initialValue: false)
+    let bytesCollected = ThreadSafeBox<Int32>(initialValue: 0)
+    let processExited = ThreadSafeBox<Bool>(initialValue: false)
```
ahoppen, line 226: We could also switch `ThreadSafeBox` to `Mutex` and remove `ThreadSafeBox`. Doesn’t have to be this PR though.

ahoppen, line 226: What I’ve done before to share an atomic is the following:

```swift
package final class RefBox<Value: ~Copyable> {
    private nonisolated(unsafe) var value: Value?

    package init(value: consuming Value) {
        self.value = consume value
    }

    package consuming func unbox() -> Value? {
        return value.take()
    }
}

extension RefBox: Sendable where Value: Sendable & ~Copyable {}
```

And then store the atomic as `RefBox<Atomic<Int>>`

Just putting this out there as an idea.

ahoppen (approved):

rintaro:
@swift-ci Please test Windows

PR #2664, opened 2026-05-22
SourceKitLSP: fallback to Swift source for SDK symbol resolve

compnerd:
When resolving a workspace symbol whose location points at an SDK .swiftmodule/.swiftinterface, SourceKit-LSP currently asks IndexStoreDB for main files containing that module file and uses one of those files to provide build settings for opening the generated interface.

That relationship is not guaranteed to exist for SDK module files. The module file can appear as the indexed symbol location, while mainFiles(containing:) still returns no source file. This causes workspaceSymbol/resolve to fail before calling sourcekitd, with errors like:

  No source file found that imports Swift.String; cannot open generated interface

If no main file is found for the module file, fall back to a buildable Swift source file from the workspace. The file is only used as the source of compiler arguments for generating the interface, so any buildable Swift source in the workspace is sufficient.

This fixes resolving stdlib/SDK workspace symbols such as Swift.String on Windows SDK layouts where the stdlib location is a .swiftmodule under the SDK.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1733,10 +1733,21 @@ extension SourceKitLSPServer {
       }()
     )
     for workspace in workspaces {
-      let mainFile = await workspace.buildServerManager
+      var candidateMainFiles = await workspace.buildServerManager
         .mainFiles(containing: moduleFileURI)
         .sorted(by: { $0.arbitrarySchemeURL.absoluteString < $1.arbitrarySchemeURL.absoluteString })
-        .first
+      if candidateMainFiles.isEmpty {
+        // SDK module files don't necessarily have a recorded main-file relationship in the index. We only need a Swift
+        // source file from the workspace to provide compiler arguments for generating the interface.
+        candidateMainFiles =
+          await orLog("Getting Swift source files for workspaceSymbol/resolve generated interface") {
+            try await workspace.buildServerManager.projectSourceFiles()
+              .keys
+              .filter { $0.fileURL?.pathExtension == "swift" }
+              .sorted(by: { $0.arbitrarySchemeURL.absoluteString < $1.arbitrarySchemeURL.absoluteString })
+          } ?? []
+      }
+      let mainFile = candidateMainFiles.first
       guard let mainFile else {
         continue
       }


compnerd:
@swift-ci please test

rintaro:
> The file is only used as the source of compiler arguments for generating the interface, so any buildable Swift source in the workspace is sufficient.

Not always. Some build server can support multiple targets for different platforms. 

IMO, the right fix is, when no workspace doesn't have a main file for it, use `.swiftmodule`/`.swiftinterface` as the `.buildSettingsFile`, then _infer_ the fallback compiler arguments in `openGeneratedInterface()`.

How? 
 * `-target`: If the module is named `<Name>.swiftmodule/<arch>.{swiftmodule|swiftinterface}`, use the triple to derive  it
 * `-sdk`: infer it from the path of the module, fallback to relative path to a toolchain used in the workspace
 * `-F {path}` for `{path}/<Name>.framework/Modules/...` , or `-I {path}` for `{path}/<Name>.swiftmodule/...`

rintaro:
If this is blocking something for Windows, feel free to disable it for Windows for now.

compnerd:
The current desire is for testing, but it seems that the test suite passed on CI, so I can hold off on this until someone has time to look into your recommendation

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1733,10 +1733,21 @@ extension SourceKitLSPServer {
...
+        // source file from the workspace to provide compiler arguments for generating the interface.
+        candidateMainFiles =
+          await orLog("Getting Swift source files for workspaceSymbol/resolve generated interface") {
+            try await workspace.buildServerManager.projectSourceFiles()
+              .keys
+              .filter { $0.fileURL?.pathExtension == "swift" }
+              .sorted(by: { $0.arbitrarySchemeURL.absoluteString < $1.arbitrarySchemeURL.absoluteString })
+          } ?? []
+      }
+      let mainFile = candidateMainFiles.first
```
ahoppen, line 1750: That doesn’t seem correct. A project may contain some files that are built for eg. the iOS SDK while others that are built for the watchOS SDK and you can’t open a module from the iOS SDK with build settings from the watchOS SDK.

What’s the exact issue that you’re seeing here? Do you have a test case that reproduces it?

PR #2666, opened 2026-05-24
Revert "Migrate atomic call sites to Synchronization.Atomic"

airspeedswift:
Speculatively revert swiftlang/sourcekit-lsp#2662 to see if it's the cause of windows failures on swift repo PRs.


airspeedswift:
@swift-ci Please test

rintaro (changes requested):
`AtomicUInt32`, `AtomicBool` etc. has gone away in `swift-tools-protocols`. We'd need to revert several PRs for reverting this.

PR #2667, opened 2026-05-25
Insert array type instead of variadic type in inlay hints

### LINKED ISSUE #2609: Inlay hints may wrongly insert variadic type
### Swift version

swiftlang-6.2.3.3.21

### Platform

macOS 15

### Editor

Visual Studio Code, main branch of SourceKit-LSP

### Description

For a function like this:
```swift
func test(x: Int...) {
  let y = x
}
```
SourceKit-LSP shows the inferred type for `y` to be `Int...`. Because the shown type is also the one that is inserted by double-clicking, SourceKit-LSP inserts `Int...` as the type for `y` which is not valid as variadic types are only allowed in function parameters.

I know that this issue probably has to be fixed in SourceKit, as SourceKit-LSP just displays the types that `source.request.variable.type` returns. However, as the `source.request.variable.type` request may also be used by other tools I'm not sure if we can just change the way variadic types are printed in the result. As a workaround, we could also just detect variadic types in the inlay hint code in SourceKit-LSP and replace them by the corresponding array type.

I'm open to providing a PR to fix this directly in SourceKit or for the workaround in SourceKit-LSP.

### Steps to Reproduce

Inlay hint test case that reproduces this issue:
```swift
func testInferredVariadicParameter() async throws {
  let (positions, hints) = try await performInlayHintRequest(
    markedText: """
      func test(x: Int...) {
        let y1️⃣ = x
      }
      """
  )
  assertHintsEqual(
    hints,
    [
      makeInlayHint(
        position: positions["1️⃣"],
        kind: .type,
        label: ": [Int]"
      ),
    ]
  )
}
```

### Logging

_No response_

adityasingh2400:
For a variable that is bound to a variadic parameter, sourcekitd's `source.request.variable.type` prints the type with a trailing `...`. For example, in

```swift
func test(x: Int...) {
  let y = x
}
```

the printed type of `y` is `Int...`. The inlay hint code used this string verbatim, both for the displayed label and for the text edit that is applied when the hint is accepted (double-clicked). A trailing `...` is only grammatical on a function parameter, so accepting the hint inserts `let y: Int... = x`, which does not compile.

The variadic spelling is also misleading semantically: inside the function body `y` is not a variadic, it is an `Array`. So the correct annotation is the array type.

The fix rewrites a top-level variadic type `T...` to `[T]` before it is used as the hint label and text edit. The detection is a `hasSuffix("...")` check, which is precise here because the variadic ellipsis is the only Swift type construct that ends in `...` (closure types containing a variadic parameter such as `(Int...) -> Void` end in their return type and are left untouched). Wrapping the element in `[...]` is always valid because it is array sugar that accepts any element type. The transformation lives on `VariableTypeInfo` so the printed type is still available unchanged for any other use.

I verified the change by building `SwiftLanguageService` and running the inlay hint tests. The added regression test fails before the fix (the hint is `: Int...`) and passes after (`: [Int]`), and the rest of `InlayHintTests` continues to pass.

Fixes #2609

### DIFF
--- a/Sources/SwiftLanguageService/InlayHintManager.swift
+++ b/Sources/SwiftLanguageService/InlayHintManager.swift
@@ -262,7 +262,7 @@ actor InlayHintManager {
       .map { info -> InlayHint in
         let position = info.range.upperBound
         let variableStart = info.range.lowerBound
-        let label = ": \(info.printedType)"
+        let label = ": \(info.annotationType)"
         let textEdits: [TextEdit]?
         if info.canBeFollowedByTypeAnnotation {
           textEdits = [TextEdit(range: position..<position, newText: label)]
--- a/Sources/SwiftLanguageService/VariableTypeInfo.swift
+++ b/Sources/SwiftLanguageService/VariableTypeInfo.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Dispatch
+import Foundation
 @_spi(SourceKitLSP) import LanguageServerProtocol
 import SourceKitD
 import SourceKitLSP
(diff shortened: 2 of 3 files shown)


ahoppen (commented):
Thanks for opening the PR, @adityasingh2400. Instead of fixing up the type annotations in SourceKit-LSP, it would be cleaner to fix [sourcekitd](https://github.com/swiftlang/swift/tree/main/tools/SourceKit) to report the variable type as an array already, in which case no code change is needed in SourceKit-LSP.

adityasingh2400:
That makes sense to me. Fixing it at the source in sourcekitd so the variable type comes back as an array is cleaner than patching the annotation in SourceKit-LSP, and it keeps the LSP layer free of compensating logic. I am happy to take that on. I will look into where sourcekitd builds the variable type for inlay hints under tools/SourceKit and have it report the array element type directly. Would you prefer I close this PR now, or leave it open as a reference until the sourcekitd change lands? Either works for me.

ahoppen:
Thanks for looking into a sourcekitd fix. I think we can close this PR as we’re not pursuing this solution anymore. And we can also always re-open it if we change our mind later.

adityasingh2400:
Sounds good, closing it. Thanks for the context, and I am happy to revisit the sourcekitd side if the team decides to pick this approach up later.

PR #2668, opened 2026-05-26
Add 'Add Explicit Raw Values' code action

### LINKED ISSUE #2516: Add explicit enum case raw values code action

ayush-that:
resolves #2516.

adds a syntactic refactoring action that converts implicit raw values to explicit ones for enums whose first inherited type is an integer (any width up to Int128/UInt128) or String.

before:

    enum Status: Int {
        case active
        case inactive
        case pending = 10
        case archived
    }

after:

    enum Status: Int {
        case active = 0
        case inactive = 1
        case pending = 10
        case archived = 11
    }

design follows the feedback @ahoppen left on #2536: implemented as `SyntaxRefactoringCodeActionProvider`, only the first inherited type is consulted for the raw value type, raw value expressions are validated as integer literals and the action is suppressed when they cannot be parsed, Int128 and UInt128 are accepted in the type list, and the action is suppressed when there is nothing to transform.

beyond #2536, the action also handles:
- negative integer raw values (`case low = -5` followed by implicit cases continues at -4)
- alternate-base integer literals with underscore separators (`case bit0 = 0x10` continues at 17)
- backticked case identifiers in String enums (``case `default`` becomes `= "default"`, not `= "`default`"`)
- integer overflow during continuation (suppresses the action rather than trapping)
- enums containing `#if` directives (rintaro flagged this on the original issue)
- enums containing freestanding macro expansions
- cases with associated values (suppresses rather than producing `case payload = 0(String)`)

### motivation

making implicit raw values explicit is a common refactor before serialising values or relying on their numeric identity. doing it by hand is error-prone once any cases already have explicit values, because the continuation rule must be reapplied for every gap.

### modifications

- new `Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift`
- registered in `Sources/SwiftSyntaxCodeActions/SyntaxCodeActions.swift` and `Sources/SwiftSyntaxCodeActions/CMakeLists.txt`
- 12 LSP-level tests in `Tests/SourceKitLSPTests/CodeActionTests.swift` covering positive cases (Int, String, negative continuation, hex continuation, backtick stripping) and negative cases (no raw value type, raw value type not first, all cases explicit, unsupported raw value expression, `#if` directives, freestanding macros, associated value cases)

### result

when the cursor is on an enum declaration with a supported raw value type and at least one case missing an explicit raw value, sourcekit-lsp offers "Add Explicit Raw Values" and inserts the missing values in place. existing positive and negative cases continue to behave as before.

all 12 new tests pass locally.

### DIFF
--- a/Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
+++ b/Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
@@ -1,0 +1,203 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftRefactor
+import SwiftSyntax
+
+/// Syntactic code action provider that adds explicit raw values to all cases of
+/// an enum whose first inherited type is `Int`/`UInt` (any width) or `String`.
+///
+/// For an enum with an integer raw value type, the action computes raw values
+/// following Swift's implicit numbering rules: counting starts at zero and
+/// continues from the last explicit value. For a `String` raw value type, each
+/// case without an explicit value is given a raw value equal to the case name.
+///
+/// ### Example
+///
+/// Before:
+///
+/// ```swift
+/// enum Status: Int {
+///     case active
+///     case inactive
+///     case pending = 10
+///     case archived
+/// }
+/// ```
+///
+/// After:
+///
+/// ```swift
+/// enum Status: Int {
+///     case active = 0
+///     case inactive = 1
+///     case pending = 10
+///     case archived = 11
+/// }
+/// ```
+struct AddExplicitEnumRawValues: EditRefactoringProvider {
+  static func textRefactor(syntax: EnumDeclSyntax, in context: Void) throws -> [SourceEdit] {
+    guard let rawValueType = syntax.rawValueType else {
+      throw RefactoringNotApplicableError("Enum does not have an Int, UInt, or String raw value type")
+    }
+
+    // Attached macros applied to the enum can introduce new cases via
+    // attribute-level macro expansion, which is not visible to a syntactic
+    // walk. Bail conservatively whenever there are any attributes; a follow-up
+    // can narrow this once we have a reliable way to tell apart attribute
+    // macros from non-macro attributes such as `@available`.
+    if !syntax.attributes.isEmpty {
+      throw RefactoringNotApplicableError("Enum has attributes that may be attached macros")
+    }
+
+    // `#if` blocks and freestanding macro expansions can introduce or hide
+    // enum cases without that being visible to a purely syntactic walk. In
+    // either case the implicit raw value counter cannot be computed safely,
+    // so bail rather than risk emitting values that disagree with the active
+    // configuration or expanded source.
+    for member in syntax.memberBlock.members {
+      if member.decl.is(IfConfigDeclSyntax.self) {
+        throw RefactoringNotApplicableError("Enum contains #if directives")
+      }
+      if member.decl.is(MacroExpansionDeclSyntax.self) {
+        throw RefactoringNotApplicableError("Enum contains a freestanding macro expansion")
+      }
+    }
+
+    let caseElements = syntax.memberBlock.members.flatMap { member -> [EnumCaseElementSyntax] in
+      guard let caseDecl = member.decl.as(EnumCaseDeclSyntax.self) else { return [] }
+      return Array(caseDecl.elements)
+    }
+
+    // Cases with associated values cannot have raw values; mixing them inside a
+    // raw-value enum is invalid Swift, but a user might be editing toward that
+    // state. Don't attempt the refactor in that case.
+    if caseElements.contains(where: { $0.parameterClause != nil }) {
+      throw RefactoringNotApplicableError("Enum has cases with associated values")
+    }
+
+    guard caseElements.contains(where: { $0.rawValue == nil }) else {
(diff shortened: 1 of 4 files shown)


Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,193 @@
...
+    // `#if` blocks and freestanding macro expansions can introduce or hide
+    // enum cases without that being visible to a purely syntactic walk. In
+    // either case the implicit raw value counter cannot be computed safely,
+    // so bail rather than risk emitting values that disagree with the active
+    // configuration or expanded source.
+    for member in syntax.memberBlock.members {
+      if member.decl.is(IfConfigDeclSyntax.self) {
+        throw RefactoringNotApplicableError("enum contains #if directives")
+      }
+      if member.decl.is(MacroExpansionDeclSyntax.self) {
```
ahoppen, line 71: I think we also need to check for attributes on the enum because they might be attached member macros that introduce cases. 

CC @rintaro in case I missed something why we might not need to consult attached macros.

Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,193 @@
...
+
+    // Cases with associated values cannot have raw values; mixing them inside a
+    // raw-value enum is invalid Swift, but a user might be editing toward that
+    // state. Don't attempt the refactor in that case.
+    if caseElements.contains(where: { $0.parameterClause != nil }) {
+      throw RefactoringNotApplicableError("enum has cases with associated values")
+    }
+
+    guard caseElements.contains(where: { $0.rawValue == nil }) else {
+      throw RefactoringNotApplicableError("all cases already have explicit raw values")
```
ahoppen: Could you uppercase the first letter of all error messages for consistency?

ahoppen (commented):
This looks great. Thank you @ayush-that. I just have two

Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
ayush-that: okie. added a bail out for any non empty attribute list. attached macros and ordinary attributes are not distinguishable at the syntactic level, so the bail is conservative (an enum with only `@available` will lose the action) but safe.

Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,193 @@
...
+  /// with `RefactoringNotApplicableError`. Such literals are rare in practice
+  /// and the action simply does not appear in those cases.
+  var rawValueType: RawValueKind? {
+    guard let firstType = inheritanceClause?.inheritedTypes.first else {
+      return nil
+    }
+    switch firstType.type.trimmedDescription {
+    case "Int", "Int8", "Int16", "Int32", "Int64", "Int128",
+      "UInt", "UInt8", "UInt16", "UInt32", "UInt64", "UInt128":
+      return .int
```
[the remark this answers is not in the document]
ayush-that: okie. split into signed/unsigned, unsigned rejects negative literals.

Sources/SwiftSyntaxCodeActions/AddExplicitEnumRawValues.swift
```diff
@@ -0,0 +1,193 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftRefactor
+import SwiftSyntax
+
+/// Syntactic code action provider that adds explicit raw values to all cases of
+/// an enum whose first inherited type is `Int` (any width) or `String`.
```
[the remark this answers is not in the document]
ayush-that: fixed

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2087,6 +2343,35 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+        textDocument: TextDocumentIdentifier(uri)
+      )
+    )
+    let codeActions = try XCTUnwrap(result?.codeActions, file: file, line: line)
+    XCTAssertFalse(
+      codeActions.contains(where: { $0.title == title }),
+      "did not expect code action titled '\(title)' at marker \(marker), got: \(codeActions.map(\.title))",
+      file: file,
+      line: line
+    )
```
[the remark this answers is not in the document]
ayush-that: fixed

ahoppen (approved):
This looks great. Thank you 🙏🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
One thing I forgot during the review: Could you open a follow-up PR to add this new refactoring action to

https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Refactoring%20Actions.md?

ayush-that:
> One thing I forgot during the review: Could you open a follow-up PR to add this new refactoring action to
> 
> https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Refactoring%20Actions.md?

okie, i'll do that. i'm very curious about why did the ci fail the first two times and worked the third time when i changed nothing. can you please educate me on that. thank you.

ahoppen:
I don’t remember what the exact failures were but I think CI was just flaky. It tests the entire Swift compiler suite and sometimes some random test in another project fails.

PR #2675, opened 2026-05-29
Bound task-scheduler shutdown wait to 10s and log process completion

rintaro:
Wrap each `task.waitToFinish()` in `TaskScheduler.shutDown()` with a 10s `withTimeout` so a stuck QueuedTask cannot block sourcekit-lsp's shutdown indefinitely; on timeout, log the task description so the offender is identifiable. The QueuedTask is already cancelled via `task.cancel()` and continues winding down in the background.

Add debug logs after `Process.waitUntilExit()` returns and after `Process.run` returns inside `prepare(singleTarget:)`, so a future shutdown hang can be triaged: did the swift build process actually exit, did `Process.run` propagate the result, or is the stall higher up in the prepare chain.

Motivated by a recurring Windows CI hang in
`BackgroundIndexingTests.testSymlinkedTargetReferringToSameSourceFile` where `prepareForExit` seemed to stall for the full test timeout with no log output during shutdown.

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -873,6 +873,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
         stderr: { @Sendable bytes in stderrHandler.handleDataFromPipe(Data(bytes)) }
       )
     )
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

hamishknight (approved):

PR #2676, opened 2026-05-29
Support `InsertReplaceEdit` and keep existing arguments when completing function calls

Steffeeen:
This PR implements two closely related features for code completion. 

## `InsertReplaceEdit` Support
First, it adds support for [`InsertReplaceEdit`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertReplaceEdit) allowing the user to choose between inserting (Enter in VSCode) the function name or replacing (Tab in VSCode) the existing rest of the identifier if the completion was invoked in the middle of an identifier. This often happens when changing which function is called. Example showing the difference:

https://github.com/user-attachments/assets/77e34a8d-6444-4dcd-8ea7-a7ee501d2294

## Reusing Existing Parentheses

The second feature this PR implements is not inserting parentheses or braces if they are already present in the document. This is useful in general, but was especially needed for the use case of changing which function is called. This example shows what happens with just the `InsertReplaceEdit` changes:

https://github.com/user-attachments/assets/b9bf1921-68ef-4fd4-b0ed-7040fe7c97b0

Compare this to the behavior when using the changes from both commits:

https://github.com/user-attachments/assets/abb41db1-c969-4f66-be07-827a7fcf8151

Below is an example of the general case of completing a function call with already present parentheses. Previously, the parentheses and argument labels would always be inserted and the user would end up with `testabc(x: , y: )(x: 3, y: 5)`.

https://github.com/user-attachments/assets/ed39ae5e-0aeb-4b02-8038-57d4df975c5d

If the parentheses already contain arguments, they are kept, even if the labels don't match the function call. I also thought about using an `InsertReplaceEdit` to let the user decide whether to replace the arguments or keep the existing ones. This however leads to a problem in situations like this: `foo|bar(x: 1, y: 2)`. Here, the replace operation could be for replacing just `bar` or replacing `bar(x: 1, y: 2)`. I also could not find another language server that does it that way, so I opted for not implementing this for now. But then again, no other language (that I know of) has mandatory argument labels where autocomplete is a real benefit. In Rust for example, the only benefit of autocomplete in this case would be inserting the commas between the arguments.

If the parentheses are empty, they are overwritten with the argument labels:

https://github.com/user-attachments/assets/cf852ae2-0095-4124-babd-2fc7f3ff8ff0

While this feature was primarily intended for the use case of completing function calls, I also ensured that it works for other similar cases, like enum case completion and macros.

### DIFF
--- a/Sources/SwiftLanguageService/CodeCompletionContext.swift
+++ b/Sources/SwiftLanguageService/CodeCompletionContext.swift
@@ -19,11 +19,20 @@ struct CompletionRequestContext {
   /// If the completion is triggered in the middle of an identifier, this is the position after the end of the
   /// identifier. This is needed to compute the correct text edit for completions that replace an existing identifier.
   let identifierEndPosition: Position?
+  let followingCallContext: FollowingCallContext
 
   init(req: CompletionRequest, snapshot: DocumentSnapshot, tree: SourceFileSyntax) {
     self.cursorPosition = req.position
     self.identifierEndPosition = findEndOfIdentifier(position: req.position, snapshot: snapshot, tree: tree)
+    self.followingCallContext = findFollowingCallContext(position: req.position, snapshot: snapshot, tree: tree)
   }
+}
+
+enum FollowingCallContext {
+  case none
+  case emptyParens(positionAfterClosingParen: Position)
+  case nonEmptyParens
+  case trailingClosure
 }
 
 private func findEndOfIdentifier(
@@ -40,6 +49,35 @@ private func findEndOfIdentifier(
   return snapshot.position(of: token.endPositionBeforeTrailingTrivia)
 }
 
+private func findIdentifierUnderCursor(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> TokenSyntax? {
+  let absolutePosition = snapshot.absolutePosition(of: position)
+  let token = tree.token(at: absolutePosition)
+  guard let token else {
+    return nil
+  }
+
+  // We also consider keywords here for cases where the user is trying to complete a function that also contains a
+  // keyword in its name. If only the keyword part of the function name is typed, we still want to show the completion
+  // for that function.
+  // For example, if the user has a function `func selfTest()`, we want to show the completion for `selfTest` even if
+  // the user has only typed `self` so far.
+  if token.isIdentifierOrKeyword {
+    return token
+  }
+
+  // If the cursor is right after an identifier, we also want to consider that identifier for completion.
+  let previousToken = token.previousToken(viewMode: .sourceAccurate)
+  guard let previousToken, previousToken.isIdentifierOrKeyword, previousToken.endPosition == absolutePosition else {
+    return nil
+  }
+
+  return previousToken
+}
+
 private extension TokenSyntax {
   var isIdentifierOrKeyword: Bool {
     switch tokenKind {
@@ -48,5 +86,77 @@ private extension TokenSyntax {
     default:
       return false
     }
+  }
+}
+
+private func findFollowingCallContext(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> FollowingCallContext {
+  guard let identifier = findIdentifierUnderCursor(position: position, snapshot: snapshot, tree: tree),
+    let parent = identifier.parent
+  else {
+    return .none
+  }
+
+  let functionCall: FunctionCallExprSyntax
+  if let declReference = parent.as(DeclReferenceExprSyntax.self) {
+    if let call = declReference.parent?.as(FunctionCallExprSyntax.self) {
+      functionCall = call
+    } else if let memberAccess = declReference.parent?.as(MemberAccessExprSyntax.self),
+      let call = memberAccess.parent?.as(FunctionCallExprSyntax.self)
+    {
+      functionCall = call
+    } else {
+      return .none
+    }
+  } else if let macroExpansion = parent.as(MacroExpansionExprSyntax.self), let leftParen = macroExpansion.leftParen {
+    return determineFollowingCallContextParenthesisKind(
+      leftParen: leftParen,
+      rightParen: macroExpansion.rightParen,
+      snapshot: snapshot
+    )
+  } else {
+    return .none
+  }
+
+  if let leftParen = functionCall.leftParen,
+    // We don't want to report empty parentheses for code like this: `foo() { doSomethingInClosure() }` as that would
+    // lead to the completion logic trying to replace the parentheses with the call argument labels
+    functionCall.trailingClosure == nil
+  {
+    return determineFollowingCallContextParenthesisKind(
+      leftParen: leftParen,
+      rightParen: functionCall.rightParen,
+      snapshot: snapshot
+    )
+  }
+
+  // Braces on the next line are also parsed into a ClosureExprSyntax, so we need to ensure that the left brace is
+  // actually on the same line as the cursor to consider it a trailing closure context.
+  if let trailingClosure = functionCall.trailingClosure,
+    snapshot.position(of: trailingClosure.leftBrace.positionAfterSkippingLeadingTrivia).line == position.line
+  {
+    return .trailingClosure
+  }
+
+  return .none
+}
+
+private func determineFollowingCallContextParenthesisKind(
+  leftParen: TokenSyntax,
+  rightParen: TokenSyntax?,
+  snapshot: DocumentSnapshot
+) -> FollowingCallContext {
+  if let rightParen, leftParen.nextToken(viewMode: .sourceAccurate) == rightParen,
+    // Only actually consider the parens as empty if there is no comment or newline in between. A comment should not be
+    // overwrittten and replacing parens across multiple lines is likely not what the user intended.
+    !leftParen.trailingTrivia.contains(where: { $0.isComment || $0.isNewline }),
+    !rightParen.leadingTrivia.contains(where: { $0.isComment || $0.isNewline })
+  {
+    return .emptyParens(positionAfterClosingParen: snapshot.position(of: rightParen.endPositionBeforeTrailingTrivia))
+  } else {
+    return .nonEmptyParens
   }
 }
--- a/Sources/SwiftLanguageService/CodeCompletionContext.swift
+++ b/Sources/SwiftLanguageService/CodeCompletionContext.swift
@@ -1,0 +1,52 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftSyntax
+
+struct CompletionRequestContext {
+  let cursorPosition: Position
+  /// If the completion is triggered in the middle of an identifier, this is the position after the end of the
+  /// identifier. This is needed to compute the correct text edit for completions that replace an existing identifier.
+  let identifierEndPosition: Position?
+
+  init(req: CompletionRequest, snapshot: DocumentSnapshot, tree: SourceFileSyntax) {
+    self.cursorPosition = req.position
+    self.identifierEndPosition = findEndOfIdentifier(position: req.position, snapshot: snapshot, tree: tree)
+  }
+}
+
+private func findEndOfIdentifier(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> Position? {
+  let token = tree.token(at: snapshot.absolutePosition(of: position))
+
+  guard let token, token.isIdentifierOrKeyword else {
+    return nil
+  }
+
+  return snapshot.position(of: token.endPositionBeforeTrailingTrivia)
+}
+
+private extension TokenSyntax {
+  var isIdentifierOrKeyword: Bool {
+    switch tokenKind {
+    case .identifier, .keyword:
+      return true
+    default:
+      return false
+    }
+  }
+}
--- a/Sources/SwiftLanguageService/CodeCompletionSession.swift
+++ b/Sources/SwiftLanguageService/CodeCompletionSession.swift
@@ -140,7 +140,7 @@ class CodeCompletionSession {
     options: SourceKitLSPOptions,
     indentationWidth: Trivia?,
     completionPosition: Position,
-    cursorPosition: Position,
+    requestContext: CompletionRequestContext,
     compileCommand: SwiftCompileCommand?,
     clientCapabilities: ClientCapabilities,
     filterText: String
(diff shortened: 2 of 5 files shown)


Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,118 @@
...
+  case emptyParens(positionAfterClosingParen: Position)
+  case nonEmptyParens
+  case trailingClosure
+}
+
+func createCompletionRequestContext(
+  req: CompletionRequest,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> CompletionRequestContext {
```
ahoppen: Should this just be a context on `CompletionRequestContext`?

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,118 @@
...
+}
+
+private func findEndOfIdentifier(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> Position? {
+  let token = tree.token(at: snapshot.absolutePosition(of: position))
+
+  guard let token = token, case .identifier(_) = token.tokenKind else {
```
ahoppen:
```suggestion
  guard let token, case .identifier = token.tokenKind else {
```

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,118 @@
...
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> TokenSyntax? {
+  let token = tree.token(at: snapshot.absolutePosition(of: position))
+  guard let token else {
+    return nil
+  }
+
+  if case .identifier(_) = token.tokenKind {
```
ahoppen:
```suggestion
  if case .identifier = token.tokenKind {
```

Same below

ahoppen: Should we also check for keywords here?

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,118 @@
...
+    return token
+  }
+
+  let previousToken = token.previousToken(viewMode: .sourceAccurate)
+  guard let previousToken, case .identifier(_) = previousToken.tokenKind else {
+    return nil
+  }
+
+  return previousToken
```
ahoppen, line 78: Is this for the case where the cursor is at the end of an identifier? In that, should we check that the previous’s token end position is equal to `position`?

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,118 @@
...
+}
+
+private func findFollowingCallContext(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> FollowingCallContext {
+  guard let identifier = findIdentifierUnderCursor(position: position, snapshot: snapshot, tree: tree),
+    let leftParenOrBrace = identifier.nextToken(viewMode: .sourceAccurate)
```
ahoppen: I think it would be nicer to inspect the syntax tree structure instead of re-implementing the current vs line matching logic from the parser. Is there any reason why we can’t do that?

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -519,7 +522,41 @@ class CodeCompletionSession {
...
-          textEdit.range = textEdit.range.lowerBound..<snapshot.position(of: nextIndex)
+          insertEnd = snapshot.position(of: nextIndex)
+        }
+      }
+
+      let followingCallContext = requestContext.followingCallContext
+      // .value is for macros that are completed with their call pattern, e.g. `#myMacro(x: , y: )`.
+      if [.method, .function, .constructor, .enumMember, .value].contains(completionKind),
+        // This is check is only a heuristic to avoid processing completions of kind .value other than macros
+        insertText.contains("(") || insertText.contains("{")
```
ahoppen, line 523: Should we check the sourcekitd insert text before doing any LSP snippet transformations to it so that we wouldn’t match a `{` from an [LSP snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax) of the form `${name:default}`?

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -638,20 +680,36 @@ class CodeCompletionSession {
...
-      utf8CodeUnitsToErase: utf8CodeUnitsToErase,
-      snapshot: snapshot
-    )
-    return TextEdit(range: textEditRangeStart..<requestPosition, newText: newText)
+    start: Position,
+    insertEnd: Position,
+    replaceEnd: Position?
+  ) -> CompletionItemEdit {
+    if clientSupportsInsertReplaceEdits, let replaceEnd {
+      let normalizedReplaceEnd = replaceEnd < insertEnd ? insertEnd : replaceEnd
```
ahoppen: When do we end up with `replaceEnd < insertEnd`?

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -519,7 +522,41 @@ class CodeCompletionSession {
...
+        // For example, if the completion item is `foo(x: , y: )` and the user invoked completion on `fo|(x: 5, y: 3)`,
+        // then we want to trim the `(x: , y: )` from the completion insert text since those parentheses are already
+        // present in the code and the user likely wants to keep them instead of ending up with
+        // `foo(x: , y: )(x: 5, y: 3)`.
+        // Note, that this currently does not look at the actual contents between the parentheses or braces to determine
+        // whether the labels match
+        switch followingCallContext {
+        case .nonEmptyParens, .trailingClosure:
+          insertText = trimCallPatternIfPresent(in: insertText, context: followingCallContext)
+        case let .emptyParens(positionAfterClosingParen) where insertText.contains("("):
```
ahoppen: Instead of inferring whether this is a call from the presence of `(`, should we look at `key.kind`?

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
```diff
@@ -1524,6 +1627,558 @@ final class SwiftCompletionTests: SourceKitLSPTestCase {
...
+    let completion = try XCTUnwrap(completions.items.first { $0.label.contains("foobar") })
+    guard case let .textEdit(textEdit)? = completion.textEdit else {
+      XCTFail("Expected text edit")
+      return
+    }
+    XCTAssertEqual(textEdit.range, positions["1️⃣"]..<positions["3️⃣"])
+    XCTAssertEqual(textEdit.newText, "foobar(x: , y: )")
+  }
+
+  func testFunctionCompletionDoesNotOverwriteMultilineParens() async throws {
```
ahoppen, line 1900: A lot of these tests have the exact same structure. Would it make sense to add a test helper like `assertOnlyCompletionEdit` that takes a source file, a position marker at which to invoke completion and checks that it matches an expected text edit?

I haven’t fully looked into it but it might also be possible to design a helper in such a way that you don’t need the local `assert` functions below.

ahoppen (commented):
Very nice improvement. Thanks, Steffen!

Sources/SwiftLanguageService/CodeCompletionContext.swift
Steffeeen: I suppose this would be relevant when trying to complete a function that contains a keyword as part of its name, e.g. `selfTest` and the editor only contains the `self` part? Or is there any other use case that I'm missing?

Sources/SwiftLanguageService/CodeCompletionSession.swift
Steffeeen: That seems to never happen, I removed the normalization.

Sources/SwiftLanguageService/CodeCompletionSession.swift
Steffeeen: Yeah, good point, I will also add a test case for that.

Sources/SwiftLanguageService/CodeCompletionSession.swift
Steffeeen: It seems like this check is not needed at all, I removed it.

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
Steffeeen: I agree that currently the tests all look really similar. I didn't introduce a newer, more general, helper as most of the other existing tests should also be changed to use that new helper and I considered that to be out of scope for this PR. Maybe I will look into this in a follow-up PR.

But I now added a helper to remove the inner `assert...` functions.

Sources/SwiftLanguageService/CodeCompletionContext.swift
ahoppen: Yeah, something like that. Or you type `await` and it gets parsed as a contextual keyword but you want to call a function called `await` and just haven’t added the backticks yet. I just can’t see a reason not to check for keywords.

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,158 @@
...
+    let parent = identifier.parent
+  else {
+    return .none
+  }
+
+  let functionCall: FunctionCallExprSyntax
+  if let declReference = parent.as(DeclReferenceExprSyntax.self) {
+    if let call = declReference.parent?.as(FunctionCallExprSyntax.self) {
+      functionCall = call
+    } else if let memberAccess = declReference.parent?.as(MemberAccessExprSyntax.self),
```
ahoppen, line 107: This could be arbitrary levels of `MemberAccessExprSyntax`, right? Ie. something like `foo.bar.baz.boom()`. You should be able to use ancestorOrSelf` for that.

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,158 @@
...
+private func determineFollowingCallContextParenthesisKind(
+  leftParen: TokenSyntax,
+  rightParen: TokenSyntax?,
+  snapshot: DocumentSnapshot
+) -> FollowingCallContext {
+  if let rightParen, leftParen.nextToken(viewMode: .sourceAccurate) == rightParen,
+    // Only actually consider the parens as empty if there is no comment or newline in between. A comment should not be
+    // overwrittten and replacing parens across multiple lines is likely not what the user intended.
+    !leftParen.trailingTrivia.contains(where: { $0.isComment }),
+    !rightParen.leadingTrivia.contains(where: { $0.isNewline })
```
ahoppen: I would play safe and check both leading and trailing trivia for newline and comment.

```suggestion
    !leftParen.trailingTrivia.contains(where: { $0.isComment || $.0isNewline }),
    !rightParen.leadingTrivia.contains(where: { $0.isComment || $.0isNewline })
```

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,158 @@
...
+    return determineFollowingCallContextParenthesisKind(
+      leftParen: leftParen,
+      rightParen: macroExpansion.rightParen,
+      snapshot: snapshot
+    )
+  } else {
+    return .none
+  }
+
+  if let leftParen = functionCall.leftParen {
```
ahoppen: Does this mean that we return `emptyParens` for the following function call? Do we want that?

```swift
foo() {
  doSomethingInTrailingClosure()
}
```

Sources/SwiftLanguageService/CodeCompletionSession.swift
ahoppen: Seems like it’s still there, right?

Sources/SwiftLanguageService/CodeCompletionSession.swift
```diff
@@ -638,20 +682,46 @@ class CodeCompletionSession {
...
-      snapshot: snapshot
-    )
-    return TextEdit(range: textEditRangeStart..<requestPosition, newText: newText)
+    start: Position,
+    insertEnd: Position,
+    replaceEnd: Position?
+  ) -> CompletionItemEdit {
+    if clientSupportsInsertReplaceEdits, let replaceEnd {
+      let normalizedReplaceEnd = replaceEnd < insertEnd ? insertEnd : replaceEnd
+      return .insertReplaceEdit(
```
ahoppen, line 693: Just to make sure I understand correctly: The following interaction will now insert `printabc` instead of `printabc(x: Int)`, right?

https://github.com/user-attachments/assets/6caa94cf-255e-49d1-829a-2d76dbb60cff

Tests/SourceKitLSPTests/SwiftCompletionTests.swift
ahoppen: Thanks, that’s a great simplification already.

Sources/SwiftLanguageService/CodeCompletionContext.swift
Steffeeen: Due to the way the syntax tree is built, the `MemberAccessExprSyntax` of `boom` always has the `FunctionCallExprSyntax` as its parent, all other `MemberAccessExprSyntax` nodes are in the base part of the `boom` `MemberAcessExprSyntax`. We would only get nested `MemberAccessExprSyntax` for any of the identifiers in the middle of the chain and in that case we would just return `none`.

Sources/SwiftLanguageService/CodeCompletionContext.swift
Steffeeen: Yes, good point. For this case `trailingClosure` is returned now. Maybe we should have a new entry for this case in the `FollowingCallContext` enum? But currently, its behavior would not differ from that of `trailingClosure`.

Sources/SwiftLanguageService/CodeCompletionSession.swift
Steffeeen: Yes, that is the motivation for the second commit.

Sources/SwiftLanguageService/CodeCompletionContext.swift
ahoppen: Oh, true. I forgot that it was built this way round. 🤦🏽

Sources/SwiftLanguageService/CodeCompletionContext.swift
ahoppen: I think `trailingClosure` is good enough for now.

Sources/SwiftLanguageService/CodeCompletionContext.swift
```diff
@@ -0,0 +1,162 @@
...
+}
+
+private func findEndOfIdentifier(
+  position: Position,
+  snapshot: DocumentSnapshot,
+  tree: SourceFileSyntax
+) -> Position? {
+  let token = tree.token(at: snapshot.absolutePosition(of: position))
+
+  guard let token, case .identifier = token.tokenKind else {
```
ahoppen: Should we also check for keywords here?

Steffeeen:
@swift-ci Please test

ahoppen (commented):
Sorry, I reviewed this last week but then forgot to submit the review.

One small comment, otherwise LGTM.

Steffeeen:
@swift-ci Please test

ahoppen (approved):
Thank you, Steffen!

PR #2677, opened 2026-05-29
Fix task leaks in TaskScheduler shutdown

rintaro:
`shutDown` only canceled `currentlyExecutingTasks`, leaving `pendingTasks` untouched. `QueuedTask` were leaked via the `self` retain cycles in with the `resultTask` closure, and the task never completes itself.

And possibly external callers awaiting `waitToFinish()` on a pending task hang forever.

Cancel pending tasks too: `task.cancel()` cancels `resultTask`, the `for await` over the stream exits immediately on cancellation, and `waitToFinish()` returns immediately.

(In reality, I don't believe this affects anything because we only shutdown `SourceKitLSPServer` when process exits)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

hamishknight (approved):

ahoppen (approved):

PR #2679, opened 2026-05-30
Fix references for local variables and parameters

### LINKED ISSUE #2583: Resolve references of local bindings
### Description

It seems that simple local variables don't resolve their references in the `textdocument/references` request. If Sourcekit-LSP doesn't build indices on this level, it's probably reasonable to visit all nodes in the function with a name, and collect each on that resolves to the requested node.

somiljain2006:
Fixes #2583 

Use relatedIdentifiers as a fallback when indexed references are unavailable, allowing textDocument/references to resolve local variables and parameters. Add regression tests covering local variable references, shadowed locals, parameter references, and includeDeclaration behavior.

### DIFF
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -565,7 +565,8 @@ package extension LanguageService {
     logger.error("\(Self.self) cannot be crashed")
   }
 
-  func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location] {
+  func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location]
+  {
     return []
   }
 }
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -567,6 +567,6 @@ package extension LanguageService {
 
   func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location]
   {
-    return []
+    throw ResponseError.internalError("\(#function) not implemented in \(Self.self)")
   }
 }
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -340,10 +340,9 @@ package protocol LanguageService: AnyObject, Sendable {
   /// Returns references for symbols within the current document using
   /// language-specific semantic analysis.
   ///
-  /// This is used by `textDocument/references` to provide up-to-date
-  /// references for the current document, including local symbols that
-  /// are not available through the index and references in documents
-  /// with in-memory edits.
+  /// This is used by `textDocument/references` as a fallback to provide
+  /// references for local symbols (like local variables and parameters)
+  /// that are not available through the global index.
   ///
   /// - Parameters:
   ///   - position: The position of the queried symbol.
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -336,6 +336,8 @@ package protocol LanguageService: AnyObject, Sendable {
 
   /// Crash the language server. Should be used for crash recovery testing only.
   func crash() async
+
+  func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location]
 }
 
 /// Default implementations for methods that satisfy the following criteria:
@@ -548,5 +550,15 @@ package extension LanguageService {
 
   func crash() async {
     logger.error("\(Self.self) cannot be crashed")
+  }
+}
+
+extension LanguageService {
+  package func localReferences(
+    at position: Position,
+    in uri: DocumentURI,
+    includeDeclaration: Bool
+  ) async throws -> [Location] {
+    return []
   }
 }
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -337,6 +337,17 @@ package protocol LanguageService: AnyObject, Sendable {
   /// Crash the language server. Should be used for crash recovery testing only.
   func crash() async
 
+  /// Returns references for local symbols that are not available through the index.
+  ///
+  /// This is used as a fallback for `textDocument/references` when indexed lookup
+  /// returns no results. Implementations may use language-specific semantic analysis
+  /// to resolve references within the current document.
+  ///
+  /// - Parameters:
+  ///   - position: The position of the queried symbol.
+  ///   - uri: The document containing the symbol.
+  ///   - includeDeclaration: Whether the declaration occurrence should be included.
+  /// - Returns: The locations of matching references in the current document.
   func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location]
 }
 
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -337,11 +337,13 @@ package protocol LanguageService: AnyObject, Sendable {
   /// Crash the language server. Should be used for crash recovery testing only.
   func crash() async
 
-  /// Returns references for local symbols that are not available through the index.
+  /// Returns references for symbols within the current document using
+  /// language-specific semantic analysis.
   ///
-  /// This is used as a fallback for `textDocument/references` when indexed lookup
-  /// returns no results. Implementations may use language-specific semantic analysis
-  /// to resolve references within the current document.
+  /// This is used by `textDocument/references` to provide up-to-date
+  /// references for the current document, including local symbols that
+  /// are not available through the index and references in documents
+  /// with in-memory edits.
   ///
   /// - Parameters:
   ///   - position: The position of the queried symbol.
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -564,14 +564,8 @@ package extension LanguageService {
   func crash() async {
     logger.error("\(Self.self) cannot be crashed")
   }
-}
 
-extension LanguageService {
-  package func localReferences(
-    at position: Position,
-    in uri: DocumentURI,
-    includeDeclaration: Bool
-  ) async throws -> [Location] {
+  func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location] {
     return []
   }
 }
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2245,9 +2245,7 @@ extension SourceKitLSPServer {
         position: req.position
       )
     )
-    let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(
-      checkedFor: .inMemoryModifiedFiles(documentManager)
-    )
+    let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(checkedFor: .deletedFiles)
     let indexLocations = try symbols.flatMap { symbol -> [Location] in
       guard let usr = symbol.usr, let index else { return [] }
       logger.info("Finding references for USR \(usr)")
@@ -2264,12 +2262,7 @@ extension SourceKitLSPServer {
       includeDeclaration: req.context.includeDeclaration
     )
 
-    var locations = indexLocations
-    var knownLocationStarts = Set(indexLocations.map(ReferenceLocationStart.init))
-    for localLocation in localLocations where knownLocationStarts.insert(ReferenceLocationStart(localLocation)).inserted
-    {
-      locations.append(localLocation)
-    }
+    let locations = indexLocations.filter { $0.uri != req.textDocument.uri } + localLocations
 
     let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
     let remappedLocations = locations.adjusted(for: copiedFileMap)
@@ -2898,16 +2891,6 @@ fileprivate extension Sequence where Element: Hashable {
   var unique: [Element] {
     var set = Set<Element>()
     return self.filter { set.insert($0).inserted }
-  }
-}
-
-private struct ReferenceLocationStart: Hashable {
-  var uri: DocumentURI
-  var position: Position
-
-  init(_ location: Location) {
-    self.uri = location.uri
-    self.position = location.range.lowerBound
   }
 }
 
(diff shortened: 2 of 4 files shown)


somiljain2006:
@ahoppen @hamishknight @rintaro Can you review this pr?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2257,6 +2258,16 @@ extension SourceKitLSPServer {
       }
       return try index.occurrences(ofUSR: usr, roles: roles).compactMap { $0.location.lspLocation }
     }
+
+    if locations.isEmpty {
+      locations =
+        (try? await languageService.localReferences(
```
rintaro: I think it's fine to throw errors. This should always succeed as long as sourcekitd is alive.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -1420,3 +1420,31 @@ extension SwiftLanguageService {
...
+    let response = try await self.relatedIdentifiers(
+      at: position,
+      in: snapshot,
+      includeNonEditableBaseNames: false
+    )
+
+    var identifiers = response.relatedIdentifiers
+
+    if !includeDeclaration {
+      identifiers = Array(identifiers.dropFirst())
```
rintaro: Is it guaranteed that the first element is the declaration? If so, could you add comment here?

rintaro (commented):
Looks nice!

Sources/SwiftLanguageService/SwiftLanguageService.swift
somiljain2006: Relying on lexical source order was a fragile assumption and didn't clearly express the intent of the code.
I looked at RelatedIdentifiers.swift and saw that sourcekitd already provides a RenameLocation.Usage for each occurrence. I've updated the implementation to remove the ordering assumption and filter out occurrences with .definition when includeDeclaration is false.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -336,6 +336,8 @@ package protocol LanguageService: AnyObject, Sendable {
 
   /// Crash the language server. Should be used for crash recovery testing only.
   func crash() async
+
+  func localReferences(at position: Position, in uri: DocumentURI, includeDeclaration: Bool) async throws -> [Location]
```
ahoppen, line 352: Could you add a doc comment to this declaration?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2257,6 +2258,15 @@ extension SourceKitLSPServer {
       }
       return try index.occurrences(ofUSR: usr, roles: roles).compactMap { $0.location.lspLocation }
     }
+
+    if locations.isEmpty {
```
ahoppen: I think we can expand this to always use the local references for the file from which the references request is invoked. When the file has in-memory edits, the index data will have stale location information while the references request will contain up-to-date information. You should be able to test this by

- Opening an indexed document
- Inserting an empty line at the beginning of the file
- Invoking the references request
- This will currently show locations based on the indexed on-disk version of the file but if we used `localReferences`, we would have up-to-date locations.

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,184 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+      )
+    )
+
+    let outerRefPositions = outerRefs.map { $0.range.lowerBound }
+    XCTAssertEqual(outerRefPositions.count, 3)
+    XCTAssertTrue(outerRefPositions.contains(project.positions["1️⃣"]))
+    XCTAssertTrue(outerRefPositions.contains(project.positions["2️⃣"]))
+    XCTAssertTrue(outerRefPositions.contains(project.positions["5️⃣"]))
+    XCTAssertFalse(outerRefPositions.contains(project.positions["3️⃣"]))
+    XCTAssertFalse(outerRefPositions.contains(project.positions["4️⃣"]))
```
ahoppen: I think this would read nicer if written like


```swift
XCTAssertEqual(Set(outerRefPositions), [project.positions["1️⃣"], project.positions["2️⃣"], project.positions["5️⃣"]])
```

Which, I think, you are doing in a test below.

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,184 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+      )
+    )
+
+    XCTAssertEqual(
+      Set(refs.map(\.range.lowerBound)),
+      Set([
+        project.positions["1️⃣"],
+        project.positions["2️⃣"],
+        project.positions["3️⃣"],
+      ])
```
ahoppen: You shouldn’t need to explicitly spell out `Set`. It should be inferred from the context.

```suggestion
      [
        project.positions["1️⃣"],
        project.positions["2️⃣"],
        project.positions["3️⃣"],
      ]
```

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,184 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+    )
+  }
+
+  func testLocalVariableReferencesWithoutDeclaration() async throws {
```
ahoppen, line 172: I might be missing something but aren’t you already testing this above with `outerRefsWithoutDecl`?

ahoppen (commented):
Nice improvement, @somiljain2006.

Sources/SourceKitLSP/SourceKitLSPServer.swift
somiljain2006: I tried this, but it causes tests to fail because localReferences misses macro expansions that the index catches. It looks like appending the results together creates duplicate entries because the index uses point ranges and relatedIdents uses full token ranges, which bypasses the .unique check.

Tests/SourceKitLSPTests/ReferencesTests.swift
somiljain2006: The declaration position case was already covered by outerRefsWithoutDecl in testLocalVariableReferences. I've simplified testLocalVariableReferencesWithoutDeclaration to only cover the usage-position scenario, which wasn't being exercised elsewhere.

Sources/SourceKitLSP/SourceKitLSPServer.swift
somiljain2006: I looked into this and found that replacing the current-file index results with localReferences caused some tests to fail because relatedIdentifiers misses some index-only references. The final approach uses the index with .inMemoryModifiedFiles, always queries localReferences, and merges both result sets while deduplicating by (uri, range.lowerBound). I also added the in-memory edit regression test you suggested, and all existing tests continue to pass.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2245,18 +2245,32 @@ extension SourceKitLSPServer {
         position: req.position
       )
     )
-    guard let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(checkedFor: .deletedFiles) else {
-      return []
-    }
-    let locations = try symbols.flatMap { (symbol) -> [Location] in
-      guard let usr = symbol.usr else { return [] }
+    let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(
+      checkedFor: .inMemoryModifiedFiles(documentManager)
```
ahoppen: We’d need to still check for `.deletedFiles` and just filter out the current document because with the current implementation, we would not be reporting any references for edited documents that aren’t the current document.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -336,6 +336,19 @@ package protocol LanguageService: AnyObject, Sendable {
 
   /// Crash the language server. Should be used for crash recovery testing only.
   func crash() async
+
+  /// Returns references for local symbols that are not available through the index.
+  ///
+  /// This is used as a fallback for `textDocument/references` when indexed lookup
+  /// returns no results. Implementations may use language-specific semantic analysis
+  /// to resolve references within the current document.
```
ahoppen: Does the doc comment now need updating?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2245,18 +2245,32 @@ extension SourceKitLSPServer {
...
     }
+
+    let localLocations = try await languageService.localReferences(
+      at: req.position,
+      in: req.textDocument.uri,
+      includeDeclaration: req.context.includeDeclaration
+    )
+
+    var locations = indexLocations
+    var knownLocationStarts = Set(indexLocations.map(ReferenceLocationStart.init))
```
ahoppen: Why is this duplicate check needed? I wouldn’t expect to receive duplicate symbols since we shouldn’t consult the index for the current file and also this should be covered by the `.unique` below.

somiljain2006:
Thanks for the feedback. I've updated the implementation accordingly: switched back to .deletedFiles, filtered out only current-document index results. I also added a test covering references from edited non-current documents are still reported. All tests are passing.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -550,3 +565,13 @@ package extension LanguageService {
     logger.error("\(Self.self) cannot be crashed")
   }
 }
+
+extension LanguageService {
+  package func localReferences(
```
ahoppen: Could you move this to the extension above that already contains default implementations?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2245,18 +2245,25 @@ extension SourceKitLSPServer {
...
       return try index.occurrences(ofUSR: usr, roles: roles).compactMap { $0.location.lspLocation }
     }
+
+    let localLocations = try await languageService.localReferences(
+      at: req.position,
+      in: req.textDocument.uri,
+      includeDeclaration: req.context.includeDeclaration
+    )
+
+    let locations = indexLocations.filter { $0.uri != req.textDocument.uri } + localLocations
```
ahoppen: Sorry, I should have thought about this earlier: We need to check whether `localReferences` is actually supported by the language service so that we don’t remove references in the current file for eg. C files, where `localReferences` isn’t implemented in the language service. 

I think what would be best is to have the default implementation for `localReferences` throw a `ResponseError.internalError`, similar to `openGeneratedInterface`, then check whether `localReferences` returns a value here and only filter the index locations when `localReferences` didn’t throw.

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -37,13 +37,7 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
       )
     )
-    XCTAssertEqual(
-      response,
-      [
-        Location(uri: project.fileURI, range: Range(project.positions["1️⃣"])),
-        Location(uri: project.fileURI, range: Range(project.positions["2️⃣"])),
-      ]
-    )
+    XCTAssertEqual(response.map(\.range.lowerBound), [project.positions["2️⃣"]])
```
ahoppen: Hmm, I didn’t realize it would change behavior like this.

@rintaro Do you have any opinion on whether we should use the possibly/regularly out-of-date index data to also cover eg. macro expansions or use from the `relatedIdents` data here, which is guaranteed to match the current document state but may produce different results?

rintaro: Hmm, reporting references from macro expansions in other files, but not from the current file, would be rather surprising. Until `relatedIdents` (or perhaps a new sourcekitd request) gains the ability to report them, I would argue that we should continue using index data, and use `relatedIdents` only when the index doesn't have any.

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,265 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+
+    let outerRefs = try await project.testClient.send(
+      ReferencesRequest(
+        textDocument: TextDocumentIdentifier(project.fileURI),
+        position: project.positions["1️⃣"],
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+
+    _ = outerRefs.map { $0.range.lowerBound }
```
ahoppen: Leftover statement?

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,265 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+      ReferencesRequest(
+        textDocument: TextDocumentIdentifier(project.fileURI),
+        position: project.positions["4️⃣"],
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+
+    XCTAssertEqual(
+      Set(innerRefs.map(\.range.lowerBound)),
+      Set([
```
ahoppen: You don’t need to specify `Set` here. 

```suggestion
      [
```

Same in a few more cases.

ahoppen (commented):
Sorry for the back-and-forth between Rintaro and myself here. The implementation looks good, just have a few nitpicks on the tests.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

somiljain2006:
@ahoppen The CI failure turned out to be unrelated to the reference's logic itself. The build was failing in the logging statements because SKLogging couldn't interpolate the error value directly. I've updated the logging code to avoid that interpolation issue and pushed a follow-up fix.

ahoppen:
@swift-ci Please test

Tests/SourceKitLSPTests/ReferencesTests.swift
```diff
@@ -72,4 +72,264 @@ final class ReferencesTests: SourceKitLSPTestCase {
...
+        textDocument: TextDocumentIdentifier(project.fileURI),
+        position: shiftedDefPos,
+        context: ReferencesContext(includeDeclaration: true)
+      )
+    )
+
+    // Since we prioritize the index to preserve macro references, we currently
+    // expect indexed symbols to return stale locations during in-memory edits.
+    // This failure is expected until sourcekitd gains macro support in relatedIdents.
+    XCTExpectFailure(
```
somiljain2006: @ahoppen The CI is currently failing because XCTExpectFailure isn't available in the XCTest version used there.

For the testReferencesWithInMemoryEdits case, would you prefer that I:

1. Remove the test entirely, since it's documenting a known limitation rather than validating the current behavior, or
2. Keep the test, but isolate the XCTExpectFailure block with an #if canImport(ObjectiveC) compilation guard

I also searched the codebase for similar uses, but XCTExpectFailure doesn't appear to be used anywhere else.

ahoppen: I would propose that you just assert on the current behavior and add a the doc comment that you wrote to it.

ahoppen:
@swift-ci Please test

somiljain2006:
@ahoppen The macOS CI failure looks unrelated to this PR. The job timed out while building, and I don't see any build errors or test failures in the log.

rintaro:
@swift-ci Please test

PR #2680, opened 2026-06-01
Add 'Generate enum associated value accessors' code action

ayush-that:
Implements #2522.

Adds a syntactic code action on an enum that generates, per case, an `is<Case>` Boolean property and (for cases with associated values) an `as<Case>` property returning the value(s) as an optional.

For example, on:

```swift
enum Value {
    case text(String)
    case number(Int)
}
```

it generates `asText` / `isText` / `asNumber` / `isNumber`.

Question (from #2544): should `is<Case>` be generated for every case(current behavior), or only for cases with associated values?

### Testing
Adds `testGenerateEnumAssociatedValueAccessors`.


Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,243 @@
...
+/// ```swift
+/// enum Value {
+///     case text(String)
+///     case number(Int)
+/// }
+/// ```
+///
+/// After:
+///
+/// ```swift
```
ahoppen: I think we should have two separate refactorings to generate the `as` and `is` names.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,243 @@
...
+/// ```
+struct GenerateEnumAssociatedValueAccessors: EditRefactoringProvider {
+  static func textRefactor(syntax: EnumDeclSyntax, in context: Void) throws -> [SourceEdit] {
+    // Attached macros, `#if` blocks, and freestanding macro expansions can
+    // introduce cases or members that are not visible to a purely syntactic
+    // walk, which could produce incomplete accessors or collide with members
+    // added by macro expansion. Bail conservatively in those cases.
+    if !syntax.attributes.isEmpty {
+      throw RefactoringNotApplicableError("Enum has attributes that may be attached macros")
+    }
```
ahoppen: I don’t think we need this bail-out. If we fail to generate one `is` accessor, that’s not a correctness issue and if it collides with a member generated by the macro, the user will get a compilation issue, which they can also decide how to fix, we shouldn’t try to make assumptions on their behalf. 

Similarly, we can generate accessors if the enum contains `#if` since we also won’t produce a correctness  issue.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,243 @@
...
+
+    // Names we must not collide with: existing members, plus the accessors we
+    // generate as we go (so e.g. `case foo` and `case Foo` don't both emit
+    // `isFoo`, which would be invalid).
+    var usedNames = existingMemberNames(in: syntax)
+
+    // Infer indentation from the source rather than hardcoding it: members are
+    // indented one step deeper than the enum, and property bodies one step
+    // deeper than the members. Deriving the step from the existing members
+    // honors whatever width (or tabs) the file already uses.
```
ahoppen: I don’t think this comment provides much value

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,243 @@
...
+    let enumIndentation = syntax.firstToken(viewMode: .sourceAccurate)?.indentationOfLine.description ?? ""
+    let memberIndentation =
+      syntax.memberBlock.members.first?.firstToken(viewMode: .sourceAccurate)?.indentationOfLine.description
+      ?? (enumIndentation + "    ")
```
ahoppen: We shouldn’t assume 4 space indentation. Use `BasicFormat.inferIndentation` to determine the file’s indentation.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
```diff
@@ -0,0 +1,243 @@
...
+
+    var accessors: [String] = []
+    for element in caseElements {
+      let caseName = element.name.identifier?.name ?? element.name.text
+      let capitalized = caseName.prefix(1).uppercased() + caseName.dropFirst()
+      let casePattern = element.name.text
+
+      if let parameters = element.parameterClause?.parameters, !parameters.isEmpty {
+        let asName = "as\(capitalized)"
+        if usedNames.insert(asName).inserted {
```
ahoppen: We shouldn’t generate the same name twice, so I think it’s sufficient to check `usedNames.contains(asName)`.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
ayush-that: okie. split into two.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
ayush-that: removed.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
ayush-that: removed.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,215 @@
...
+    let indentation = accessorIndentation(for: syntax)
+
+    var accessors: [String] = []
+    for element in enumCaseElements(of: syntax) {
+      guard let parameters = element.parameterClause?.parameters, !parameters.isEmpty else {
+        continue
+      }
+      let name = "as\(capitalizedCaseName(of: element))"
+      guard !usedNames.contains(name) else { continue }
+      usedNames.insert(name)
```
ahoppen: You should be able to combine this in one call as follows.

```suggestion
      guard !usedNames.insert(name).inserted else { continue }
```

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,215 @@
...
+    var usedNames = existingMemberNames(in: syntax)
+    let indentation = accessorIndentation(for: syntax)
+
+    var accessors: [String] = []
+    for element in enumCaseElements(of: syntax) {
+      guard let parameters = element.parameterClause?.parameters, !parameters.isEmpty else {
+        continue
+      }
+      let name = "as\(capitalizedCaseName(of: element))"
+      guard !usedNames.contains(name) else { continue }
```
ahoppen: Why do we need to check for duplicate names here? Do you have an example?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2332,6 +2013,343 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                    """
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
```
ahoppen, line 3172: Could you add a test case for the following source code?

```swift
enum Foo {
    case value(int: Int)
    case value(string: String)
}
```

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,215 @@
...
+  ) -> String {
+    let bindings = parameters.enumerated().map { index, parameter -> String in
+      if let label = parameter.firstName, label.text != "_" {
+        return label.text
+      }
+      return parameters.count == 1 ? "value" : "value\(index + 1)"
+    }
+    let types = parameters.map { $0.type.trimmedDescription }
+    let returnType: String
+    if parameters.count == 1 {
```
ahoppen: If you write it as follows, you don’t need the `[0]` access below.

```suggestion
    if let parameter = parameters.only {
```

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    return try sourceEdits(insertingAccessors: accessors, into: syntax)
+  }
+
+  private static func makeAsAccessor(
+    name: String,
+    casePattern: String,
+    parameters: [EnumCaseParameterSyntax],
+    indentation: (member: String, body: String)
```
ahoppen: Instead of passing the member and body indentation here, it would be cleaner to pass a base indentation and the indentation step so that we can compute the effective indentation in here. Otherwise you need to understand how `indentation.member` and `indentation.body` are used to correctly set it.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,215 @@
...
+    return """
+      \(indentation.member)var \(name): \(returnType) {
+      \(indentation.body)if case let .\(casePattern)(\(boundValues)) = self { return \(returnValue) }
+      \(indentation.body)return nil
+      \(indentation.member)}
+      """
+  }
+}
+
+extension GenerateEnumCaseAsAccessors: SyntaxRefactoringCodeActionProvider {
```
ahoppen: Instead of having extensions of the same type in the same file, I think it’s clearer if we just throw all members into on `struct` declaration.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2332,6 +2013,343 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+          diagnostics: nil,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["2️⃣"]..<positions["2️⃣"],
+                  newText: """
+
+
+                        var asPoint: (Int, Int)? {
```
ahoppen: I think this should also contain the tuple labels, ie. this should be


```suggestion
                        var asPoint: (x: Int, y: Int)? {
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2332,6 +2013,343 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+          diagnostics: nil,
+          edit: WorkspaceEdit(
+            changes: [
+              uri: [
+                TextEdit(
+                  range: positions["2️⃣"]..<positions["2️⃣"],
+                  newText: """
+
+
+                        var isFoo: Bool {
```
ahoppen, line 3111: I would rather generate duplicate cases here and have a compiler error that the user needs to fix instead of implicitly assuming that `isFoo` should refer to the underscore version of `foo`.

ayush-that:
1. the name tracking is gone completely
2. is accessors are generated for every case, as accessors for cases with associated values
3. duplicate names end up as compiler errors
4. tuple return types keep their labels
5. added overloaded test case

Tests/SourceKitLSPTests/CodeActionTests.swift
ayush-that: i agree. i've removed the name tracking, both `isFoo` accessors get generated and the `foo`/`Foo` test checks the duplicate output.

Tests/SourceKitLSPTests/CodeActionTests.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: done.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: done.

Tests/SourceKitLSPTests/CodeActionTests.swift
ayush-that: done. i've added `testGenerateEnumCaseAsAccessorsForOverloadedCaseNames`.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: no. i removed the whole name tracking, including the skip for existing members. duplicate names now end up as compiler errors.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: this code is gone now, along with the rest of the name tracking.

ahoppen (commented):
Thanks for the update. Now, if an accessor already exists, we will create a second copy of it, right? I don’t think we should do that. Also: If all accessors already exist, we shouldn’t suggest the refactoring action.

ayush-that:
You're right 😓. It duplicated, so I've added back the skip for accessor names that already exist, and when all of them already exist the action isn't offered anymore.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,199 @@
...
+
+private func enumCaseElements(of enumDecl: EnumDeclSyntax) -> [EnumCaseElementSyntax] {
+  enumDecl.memberBlock.members.flatMap { member -> [EnumCaseElementSyntax] in
+    guard let caseDecl = member.decl.as(EnumCaseDeclSyntax.self) else { return [] }
+    return Array(caseDecl.elements)
+  }
+}
+
+/// The names of the enum's existing properties, so an accessor whose name is already taken isn't generated again.
+private func existingMemberNames(of enumDecl: EnumDeclSyntax) -> Set<String> {
```
ahoppen: Let’s name this `existingVariableNames` because it only reports variables, not all members, which would also include functions.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
```diff
@@ -0,0 +1,199 @@
...
+    return """
+      \(memberIndentation)var \(name): \(returnType) {
+      \(bodyIndentation)if case let .\(casePattern)(\(boundValues)) = self { return \(returnValue) }
+      \(bodyIndentation)return nil
+      \(memberIndentation)}
+      """
+  }
+}
+
+/// Syntactic code action that adds an `is<Case>` Boolean computed property for
```
ahoppen, line 95: Could you also document these new refactoring actions in https://github.com/swiftlang/sourcekit-lsp/blob/main/Documentation/Refactoring%20Actions.md?

ahoppen (commented):
LGTM, just two small additional comments left.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: Done.

Sources/SwiftSyntaxCodeActions/GenerateEnumCaseAccessors.swift
ayush-that: Done.

ahoppen (approved):

ahoppen:
LGTM. Thanks for the contribution again, @ayush-that!

@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@ayush-that Would you be able to resolve the merge conflict introduced by https://github.com/swiftlang/sourcekit-lsp/pull/2671?

ayush-that:
Done @ahoppen

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ayush-that:
@ahoppen Resolved the merge conflicts again.

ahoppen:
Thank you

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2683, opened 2026-06-01
Add diagnostic logging for indexing scheduling investigation

rintaro:
After repeated CI issues where indexing tasks seems to stall in the scheduling pipeline, debug log:

- when scheduleIndexing reaches the post-preparation withTaskGroup
- when schedulePreparation creates a QueuedTask, with priorities
- when updateIndexStore creates a QueuedTask, with priorities
- when poke() rejects a pending task because of budget
- when QueuedTask.resultTask body starts, capturing both the stored and runtime priorities to detect priority elevation races

Together these distinguish whether the indexTask body wedged, a QueuedTask was never created, priority escalation never fired poke() to retry, or the resultTask body started after an elevation had already happened.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -698,6 +698,9 @@ package final actor SemanticIndexManager {
       }
       self.indexProgressStatusDidChange()
     }
+    logger.debug(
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

ahoppen (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #2684, opened 2026-06-02
Index targets top-down instead of bottom up

ahoppen:
We originally started preparing and indexing the low-level targets in the hope that we can unblock the files of those targets as early as possible for indexing and thus get best indexing performance. Revisiting this design decision, we can actually get a ~2x speed up (SourceKit-LSP gets indexed in 2:45 instead of 5:49 minutes on my machine) in initial indexing performance by preparing indexing top-level targets first. This speed-up is for two reasons:
- When preparing a top-level target, all of its dependencies are implicitly also prepared. That means that we don’t need to send any prepare requests to the build server for those targets, which is very advantages for SwiftPM projects, in which the constant cost in the order of 500ms (based on my memory) even for a null build.
- The underlying build server is able to prepare dependencies of the top-level target in parallel instead of implicitly serializing them by preparing all targets in the package in dependency order.


Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -915,7 +915,9 @@ package final actor SemanticIndexManager {
     // The targets sorted in reverse topological order, low-level targets before high-level targets. If topological
     // sorting fails, sorted in another deterministic way where the actual order doesn't matter.
     var sortedTargets: [BuildTargetIdentifier] =
-      await orLog("Sorting targets") { try await buildServerManager.topologicalSort(of: Array(filesByTarget.keys)) }
+      await orLog("Sorting targets") {
+        try await buildServerManager.targetsSortedForIndexing(Array(filesByTarget.keys))
```
rintaro, line 964: If `batchSize` is greater than 1, the targets end up being partitioned in a rather odd way. I don't believe `preparationBatchingStrategy` is widely used, but this change puts it in a somewhat awkward position.

ahoppen: Why do you think it sorts them in an odd way? It shouldn’t be any more odd than it is today, right?

ahoppen:
@swift-ci Please test

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -1450,24 +1449,53 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
       return lhs.uri.stringValue < rhs.uri.stringValue
     }
+
+    // Now walk through the list of targets in the work list. For each target from the work list, index all of its
+    // transitive dependencies next. We do this because preparing a top-level target likely also prepared all of its
+    // dependencies, so we should be able to index all files in the target's dependencies without needing to perform any
+    // target preparation.
+    var sorted: [BuildTargetIdentifier] = []
+    while !workList.isEmpty {
+      let target = workList.removeFirst()
```
hamishknight: This is quadratic since all the elements of `workList` need to be moved down the buffer. Could we make `workList` an `ArraySlice` instead?

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -1450,24 +1449,53 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
+      sorted.append(target)
+
+      let transitiveDependencies = transitiveClosure(of: [target]) { Set(buildTargets[$0]?.target.dependencies ?? []) }
+      let dependenciesInWorkList = workList.filter { transitiveDependencies.contains($0) }
+      sorted += dependenciesInWorkList
+      workList.removeAll(where: { dependenciesInWorkList.contains($0) })
```
hamishknight: Rather than doing an O(n) remove on each iteration, would it make more sense to just collect all the processed elements into a set, and skip them when visiting them in the loop? Actually you then don't even need to do the `removeFirst`, you can just make this a regular `for` loop

Sources/BuildServerIntegration/BuildServerManager.swift
```diff
@@ -1450,24 +1449,53 @@ package actor BuildServerManager: QueueBasedMessageHandler {
...
       if lhsDepth != rhsDepth {
-        return lhsDepth > rhsDepth
+        return lhsDepth < rhsDepth
       }
+      // Use the target's name as a tie-breaker
       return lhs.uri.stringValue < rhs.uri.stringValue
     }
+
+    // Now walk through the list of targets in the work list. For each target from the work list, index all of its
+    // transitive dependencies next. We do this because preparing a top-level target likely also prepared all of its
```
hamishknight, line 1501: Hmm this does feel like it's undermining the previous sorting a bit, since now this means we'll potentially index files from dependency packages before files in the root package, right? Should we be maintaining the root/dependency package split?

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -771,10 +767,15 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
 
   package func prepare(request: BuildTargetPrepareRequest) async throws -> BuildTargetPrepareResponse {
     // TODO: Support preparation of multiple targets at once. (https://github.com/swiftlang/sourcekit-lsp/issues/1262)
+    var implicitlyPreparedTargets: [BuildTargetIdentifier] = []
     for target in request.targets {
-      await orLog("Preparing") { try await prepare(singleTarget: target) }
+      await orLog("Preparing") {
+        try await prepare(singleTarget: target)
+        implicitlyPreparedTargets += transitiveClosure(of: [target], successors: { targetDependencies[$0] ?? [] })
```
hamishknight, line 788: Do we have to worry about BSP servers that may choose a different platform for the target vs one of its dependencies? e.g

```
A (iOS) -> B (iOS + macOS)
```

If the current run destination is macOS and we try to prepare A, a BSP might reasonably decide to fallback and prepare it using iOS instead, since that's the only thing it supports. B then also gets prepared for iOS. But if you ask for preparation of B with a macOS run destination, naturally you'd expect macOS.

cc @AnthonyLatsis who might also have thoughts here

Sources/BuildServerIntegration/BuildServerManager.swift
ahoppen: Good idea 👍🏽

Sources/BuildServerIntegration/BuildServerManager.swift
ahoppen: Yes, it does. But in practice I have found that the bottleneck to indexing is exclusively target preparation. So, if we chose to not index these files, a processor core would likely just be sitting idle. Furthermore, since all the preparation tasks are scheduled before the file index tasks, we should always prefer a target preparation over file indexing.

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
ahoppen: Targets in BSP are always configured targets for one particular platform. So we can’t have a target that is for multiple platforms here.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
hamishknight: Yeah, I guess I'm just wondering if that's the right long-term model. In any case, this is a future thing

hamishknight (approved):

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
ahoppen: I believe it is, but as you said, future discussion.

g54555848-cmyk:
ا

Yahoo Mail: Search, Organize, Conquer 
 
  On الاثنين, يوليو ١٣, ٢٠٢٦ at ١٧:٤٢, Alex ***@***.***> wrote:   


@ahoppen commented on this pull request.

In Sources/BuildServerIntegration/SwiftPMBuildServer.swift:
>      for target in request.targets {
-      await orLog("Preparing") { try await prepare(singleTarget: target) }
+      await orLog("Preparing") {
+        try await prepare(singleTarget: target)
+        implicitlyPreparedTargets += transitiveClosure(of: [target], successors: { targetDependencies[$0] ?? [] })

I believe it is, but as you said, future discussion.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -960,7 +960,9 @@ package final actor SemanticIndexManager {
     // The targets sorted in reverse topological order, low-level targets before high-level targets. If topological
```
rintaro: This comment looks outdated.

rintaro (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
ahoppen: Good catch. Updated.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2685, opened 2026-06-02
Fix build failures in a different build environment

rintaro:
* **`Workspace`: explicit weak captures**
  The outer `Task` in `Workspace.init` captures `sourceKitLSPServer` weakly, which makes it a `var` inside the closure. Three inner `@Sendable` closures captured it implicitly, which triggers the `SendableClosureCaptures` diagnostic in compilers without `ImmutableWeakCaptures` enabled. Add explicit `[weak sourceKitLSPServer]` capture lists so each inner closure binds to a fresh weak local instead of the enclosing `var`.

* **`UncheckedIndex`: `sending` parameter**
  `ThreadSafeBox.init` requires `consuming sending Value`. The `IndexStoreDB?` parameter on `UncheckedIndex.init` wasn't marked `sending`, so callers couldn't satisfy the region-isolation requirement under stricter build configurations and the call site failed with `SendingRisksDataRace`. Mark the parameter as `sending` to propagate the requirement up to the call site, where the freshly-constructed `IndexStoreDB` is naturally region-isolated.

### DIFF
--- a/Sources/SemanticIndex/CheckedIndex.swift
+++ b/Sources/SemanticIndex/CheckedIndex.swift
@@ -344,7 +344,7 @@ package final actor UncheckedIndex: Sendable {
   /// The set of unit output paths that are currently registered in the underlying `IndexStoreDB`.
   private var unitOutputPaths: Set<String> = []
 
-  package init?(_ index: IndexStoreDB?, usesExplicitOutputPaths: Bool) {
+  package init?(_ index: sending IndexStoreDB?, usesExplicitOutputPaths: Bool) {
     guard let index else {
       return nil
     }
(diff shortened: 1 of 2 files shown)


rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/64
@swift-ci Please test

ahoppen (approved):

hamishknight (approved):

rintaro:
https://github.com/swiftlang/swift-foundation/pull/2015
https://github.com/swiftlang/swift-tools-protocols/pull/64
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/64
@swift-ci Please test macOS

PR #2692, opened 2026-06-17
Run more of SwiftPMBuildServerTests with the out of process BSP

owenv:
Parameterize more tests in SwiftPMBuildServerTests to run with both the in and out of process implementations, fixing a few issues along the way.

- Use the new --experimental-skip-resolving-package-paths option so the build server handles si=ymlinked sources the same way as the old implementation
- Default to a separate scratch path when background indexing
- Resolve config options relative to the project root

After this change there are just a few smaller changes needed to run the whole SourceKit-LSP test suite with the out of process BSP, but I'm planning on landing those separately.

Depends on https://github.com/swiftlang/swift-build/pull/1470#pullrequestreview-4519819234

### DIFF
--- a/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
+++ b/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
@@ -50,6 +50,10 @@ private var hostTriple: Triple {
 }
 
 fileprivate extension SourceKitLSPOptions {
+  static var forTestingNativeSwiftPMBuildServer: Self {
+    SourceKitLSPOptions(swiftPM: SwiftPMOptions(buildSystem: .native))
+  }
+
   static var forTestingExperimentalSwiftPMBuildServer: Self {
     SourceKitLSPOptions(swiftPM: SwiftPMOptions(buildSystem: .swiftbuild))
   }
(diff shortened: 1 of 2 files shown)


owenv:
https://github.com/swiftlang/swift-build/pull/1470
@swift-ci test

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -529,15 +536,14 @@ struct SwiftPMBuildServerTests {
...
-        try packageRoot
-          .appending(components: "Sources", "libC", "include")
-          .filePath,
-        arguments: arguments
-      )
+      let libCInclude = try packageRoot.appending(components: "Sources", "libC", "include").filePath
+      if options.swiftPMOrDefault.buildSystem == .swiftbuild {
+        #expect(
+          arguments.contains("-I\(libCInclude)"),
+        )
```
ahoppen: Should we use `expectArgumentsContain` here to be consistent with the check with the native build system?

Same for `#expect` below.

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test

ahoppen (approved):

PR #2694, opened 2026-06-18
Add swap binary operands code action

### LINKED ISSUE #2520: Add Flip operands of binary expression code action

somiljain2006:
Fixes #2520 

Implement swap binary operands refactoring. Add a syntax-based code action that swaps the left and right operands of a binary expression when the cursor is positioned on the operator. Comparison operators are inverted as needed (for example, `<` becomes `>`), while commutative operators are preserved. Add tests covering nested expressions, operator selection, unsupported operators, and trivia preservation.


somiljain2006:
@ahoppen @rintaro Can you review this pr?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftOperators
+import SwiftSyntax
+
+struct SwapBinaryOperands: SyntaxCodeActionProvider {
```
ahoppen: Could you implement this as a `SyntaxRefactoringCodeActionProvider` similar to https://github.com/swiftlang/sourcekit-lsp/pull/2680?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+    let finder = CursorInfixFinder(cursorRanges: [absoluteRange, relativeRange])
+    finder.walk(foldedExpr)
+
+    guard let infixExpr = finder.found else {
+      return []
+    }
+
+    let opExpr = infixExpr.operator
+    let currentOperatorText: String
+    if let declRef = opExpr.as(DeclReferenceExprSyntax.self) {
```
ahoppen: When does this case occur?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+      return []
+    }
+
+    let leftOperand = infixExpr.leftOperand
+    let rightOperand = infixExpr.rightOperand
+
+    // Ignore incomplete expressions produced while the user is typing.
+    guard !leftOperand.is(MissingExprSyntax.self), !rightOperand.is(MissingExprSyntax.self) else {
+      return []
+    }
```
ahoppen: What exactly are you trying to guard against here? If you’re trying to detect any kind of syntax errors, I would rather check `.hasError`.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+          ]
+        )
+      )
+    ]
+  }
+}
+
+/// UTF-8 offset range used for comparing cursor positions against
+/// operator token ranges without relying on source locations.
+private struct OffsetRange {
```
ahoppen: Why do we need this and can’t use `Range<AbsolutePosition>` here?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+    if let binOp = self.as(BinaryOperatorExprSyntax.self) {
+      return binOp.operator.tokenTextOffsetRange
+    }
+    return nil
+  }
+}
+
+/// Finds the innermost infix operator whose token range contains
+/// the cursor position.
+private final class CursorInfixFinder: SyntaxVisitor {
```
ahoppen: Can’t you achieve the same using `token(at:).parent` or similar?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,243 @@
...
+
+    let newInfix =
+      infixExpr
+      .with(\.leftOperand, newLeft)
+      .with(\.operator, newOperatorExpr)
+      .with(\.rightOperand, newRight)
+
+    // Replace only the selected infix expression while leaving the
+    // surrounding folded expression tree unchanged.
+    let finalExpr = SwapRewriter(targetId: infixExpr.id, replacement: newInfix).visit(foldedExpr)
```
ahoppen: Is this needed? Can’t you just replace `infixExpr` instead of `exprToFold` below?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2333,6 +2333,376 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+            changes: [
+              uri: [
+                TextEdit(range: positions["0️⃣"]..<positions["3️⃣"], newText: "b && a")
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
```
ahoppen, line 2669: Just to be sure: Could you add a test case for `1 + 2 - 3` and invoking the swap on `+`? Just to make sure it produces `2 + 1 - 3?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
somiljain2006: The intent is to avoid offering refactoring on incomplete binary expressions while the user is typing. I have added comments and test cases in my updated pr.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
somiljain2006: infixExpr is part of the detached tree returned by foldAll(), so it cannot be used directly to determine the edit range in the source file. Therefore, rewrite the folded tree and replace the original exprToFold range with the updated expression.

Tests/SourceKitLSPTests/CodeActionTests.swift
somiljain2006: Done

somiljain2006:
@ahoppen, can you review the updated changes?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    // Only offer the refactoring when the operator participates in a
+    // binary expression that can be folded into an InfixOperatorExprSyntax.
+    guard
+      opExpr.findParentOfSelf(
+        ofType: ExprSyntax.self,
+        stoppingIf: { $0.is(CodeBlockItemSyntax.self) },
+        matching: { $0.is(SequenceExprSyntax.self) || $0.is(InfixOperatorExprSyntax.self) }
+      ) != nil
```
ahoppen: I don’t think you want to walk all the way up the tree but only inspect the direct parent (or grandparent for `SequenceExprSyntax`). For example, you don’t want to offer the refactoring action when invoked on `+` in

```swift
2 * list.reduce(0, +)
```

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+
+    // Restrict the action to selections that fall within the operator token
+    // itself. This prevents offering the action when the cursor is placed on
+    // either operand or in surrounding trivia.
+    guard startPos >= tokenStart else { return nil }
+    if startPos == endPos {
+      guard startPos < tokenEnd else { return nil }
+    } else {
+      guard endPos <= tokenEnd else { return nil }
+    }
```
ahoppen: Instead of comparing positions here could you use something like `(startPos..<endPos).overlaps(token.trimmedRange)` here?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+
+    let token = opExpr.operator
+    let tokenStart = token.positionAfterSkippingLeadingTrivia
+    let tokenEnd = token.endPositionBeforeTrailingTrivia
+
+    let startPos = scope.snapshot.absolutePosition(of: scope.request.range.lowerBound)
+    let endPos = scope.snapshot.absolutePosition(of: scope.request.range.upperBound)
+
+    // Restrict the action to selections that fall within the operator token
+    // itself. This prevents offering the action when the cursor is placed on
```
ahoppen: I suppose you mean


```suggestion
    // Restrict the action to selections that overlap with the operator token
    // itself. This prevents offering the action when the cursor is placed on
```

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+      opExpr.findParentOfSelf(
+        ofType: ExprSyntax.self,
+        stoppingIf: { $0.is(CodeBlockItemSyntax.self) },
+        matching: { $0.is(SequenceExprSyntax.self) || $0.is(InfixOperatorExprSyntax.self) }
+      ) != nil
+    else {
+      return nil
+    }
+
+    let token = opExpr.operator
```
ahoppen: Let’s name this `operatorToken`. `token` could be anything really.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+    } else {
+      guard endPos <= tokenEnd else { return nil }
+    }
+
+    return ExprSyntax(opExpr)
+  }
+
+  package static func textRefactor(syntax opExpr: ExprSyntax, in context: Void) throws -> [SourceEdit] {
+    guard let binOp = opExpr.as(BinaryOperatorExprSyntax.self) else {
+      return []
```
ahoppen: Could you throw an error if the refactoring action is not applicable to match the doc comment of `textRefactor` on `EditRefactoringProvider`, which is what this implements.
 
Same below

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -0,0 +1,227 @@
...
+    // Calculate the relative UTF-8 offset.
+    let targetRelativeOffset =
+      binOp.operator.positionAfterSkippingLeadingTrivia.utf8Offset - exprToFold.position.utf8Offset
+
+    let finder = OperatorMatchFinder(
+      targetText: currentOperatorText,
+      targetRelativeOffset: targetRelativeOffset,
+      rootPosition: foldedExpr.position
+    )
+    finder.walk(foldedExpr)
```
ahoppen: Shouldn’t you be able to use something like `folderExpr.token(at: targetRelativeOffset)` to get the operator token here and then use `.parent` to get the `InfixOperatorExprSyntax`?

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
ahoppen: But you don’t need to replace all of `exprToFold`, right? You could just replace the range of `infixExpr`. You might just need to adjust its range similarly to what you do for `targetRelativeOffset`.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
ahoppen, line 1: Could you also add this refactoring action to `Refactoring Actions.md`?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2333,6 +2333,424 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+    try await assertNoCodeAction(
+      titled: "Swap operands",
+      in: """
+        let x = 1 + 1️⃣-2 * 5
+        """,
+      atMarker: "1️⃣"
+    )
+  }
+
+  func testSwapBinaryOperandsGreaterThanOrEqual() async throws {
```
ahoppen: I don’t think this test provides significant value in addition to `testSwapBinaryOperandsGreaterThan`, so I’d remove it.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2333,6 +2333,424 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+      titled: "Swap operands",
+      in: """
+        let x = a1️⃣ + b
+        """,
+      atMarker: "1️⃣"
+    )
+  }
+
+  func testSwapBinaryOperandsNotOfferedWhitespaceAfterOperator() async throws {
+    try await assertNoCodeAction(
```
ahoppen: Actually, I think we should offer the code action here. We have `adjustPositionToStartOfIdentifier` with a rationale for why it’s needed. You may be able to factor out the logic that operates on a syntax tree and reuse it.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -2333,6 +2333,424 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testSwapBinaryOperandsLogicalAnd() async throws {
```
ahoppen: I think this doesn’t provide significant value to what we’re already testing eg. with `+`.

ahoppen (commented):
Thank you. The results look good but I think the implementation can be simplified a bit. Left a few comments below.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
somiljain2006: done

Sources/SwiftSyntaxCodeActions/SyntaxCodeActionProvider.swift
```diff
@@ -88,3 +88,11 @@ private func tokenForRefactoring(
   }
   return token
 }
+
+extension TokenSyntax {
+  /// Returns whether the given position is within the token's text content,
+  /// inclusive of the start and end boundaries, ignoring leading and trailing trivia.
+  package func containsContent(at position: AbsolutePosition) -> Bool {
+    return (positionAfterSkippingLeadingTrivia...endPositionBeforeTrailingTrivia).contains(position)
+  }
```
ahoppen: I don’t think we need a special function for this. The caller should be able to just do `token.trimmedRange.contains(position)`.

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -39,50 +39,42 @@ struct SwapBinaryOperands: SyntaxRefactoringCodeActionProvider {
     // Only offer the refactoring when the operator participates in a
     // binary expression that can be folded into an InfixOperatorExprSyntax.
     guard
-      opExpr.findParentOfSelf(
-        ofType: ExprSyntax.self,
-        stoppingIf: { $0.is(CodeBlockItemSyntax.self) },
-        matching: { $0.is(SequenceExprSyntax.self) || $0.is(InfixOperatorExprSyntax.self) }
-      ) != nil
+      opExpr.parent?.is(InfixOperatorExprSyntax.self) == true
+        || opExpr.parent?.parent?.is(SequenceExprSyntax.self) == true
```
ahoppen: Let’s do `opExpr.parent?.as(ExprListSyntax.self)?.parent?.is(SequenceExprSyntax.self) == true`

I always feel like that better communicates what the intermediate parent is supposed to be.

Same below

Sources/SwiftSyntaxCodeActions/SwapBinaryOperands.swift
```diff
@@ -39,50 +39,42 @@ struct SwapBinaryOperands: SyntaxRefactoringCodeActionProvider {
...
+    // Only offer the refactoring when the cursor or selection targets the
+    // operator token. This prevents offering the action when the cursor is placed on
     // either operand or in surrounding trivia.
-    guard startPos >= tokenStart else { return nil }
-    if startPos == endPos {
-      guard startPos < tokenEnd else { return nil }
-    } else {
-      guard endPos <= tokenEnd else { return nil }
+    let tokenRange = opExpr.operator.trimmedRange
+    guard selectionRange.isEmpty ? opExpr.operator.containsContent(at: startPos) : selectionRange.overlaps(tokenRange)
```
ahoppen: Ould it make sense to use `overlapsOrTouches` here?

ahoppen (commented):
A couple small comments, otherwise looks good to me.

ahoppen (approved):
LGTM. Thank you @somiljain2006!

ahoppen:
@swift-ci Please test

ahoppen:
@somiljain2006 Would you be able to resolve the merge conflict introduced by https://github.com/swiftlang/sourcekit-lsp/pull/2671?

somiljain2006:
@ahoppen Done

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2698, opened 2026-06-20
Add go-to-definition for symbols in doc comments, Markdown and Tutorial files

Padmashree06:
This PR introduces go-to-definition support for Markdown and Tutorial files along with symbols referenced in documentation comments in Swift files.

To enable this functionality, symbol graphs are now emitted during both the initial build and background indexing. The generated symbol graph data is then used to resolve symbols referenced in documentation files and navigate to their corresponding source definitions.

### Changes
- Emit symbol graphs during the initial build.
- Emit symbol graphs during background indexing.
- Add go-to-definition support for Markdown and Tutorial files using the generated symbol graph data. 
- Add go-to-definition support for symbols  referenced in documentation comments in Swift files. 

### User Impact
Users can now invoke go-to-definition on symbols referenced in Markdown and Tutorial files, as well as Swift documentation comments, and navigate to the corresponding source definition.

### DIFF
--- a/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
+++ b/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
@@ -38,10 +38,11 @@ struct DocCSymbolInformation {
   }
 
   func matches(_ link: DocCSymbolLink) -> Bool {
-    guard link.components.count == components.count else {
+    guard link.components.count <= components.count else {
       return false
     }
-    return zip(link.components, components).allSatisfy { linkComponent, symbolComponent in
+    let matchingSuffix = components.suffix(link.components.count)
+    return zip(link.components, matchingSuffix).allSatisfy { linkComponent, symbolComponent in
       linkComponent.name == symbolComponent.name && symbolComponent.information.matches(linkComponent.disambiguation)
     }
   }
--- a/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
+++ b/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
@@ -38,6 +38,15 @@ struct DocCSymbolInformation {
   }
 
   func matches(_ link: DocCSymbolLink) -> Bool {
+    guard link.components.count == components.count else {
+      return false
+    }
+    return zip(link.components, components).allSatisfy { linkComponent, symbolComponent in
+      linkComponent.name == symbolComponent.name && symbolComponent.information.matches(linkComponent.disambiguation)
+    }
+  }
+
+  func matchesAsSuffix(_ link: DocCSymbolLink) -> Bool {
     guard link.components.count <= components.count else {
       return false
     }
--- a/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
+++ b/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
@@ -10,12 +10,17 @@
 //
 //===----------------------------------------------------------------------===//
 
+import BuildServerIntegration
 import Foundation
 import IndexStoreDB
 @_spi(SourceKitLSP) package import LanguageServerProtocol
+import Markdown
+@_spi(SourceKitLSP) import SKLogging
 package import SKOptions
+import SemanticIndex
 package import SourceKitLSP
 import SwiftExtensions
+import SwiftParser
 package import SwiftSyntax
 package import ToolchainRegistry
 
@@ -36,7 +41,8 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
 
   package static var experimentalCapabilities: [String: LSPAny] {
     return [
-      DoccDocumentationRequest.method: ["version": 1]
+      DoccDocumentationRequest.method: ["version": 1],
+      "definitionProvider": .bool(true),
     ]
   }
 
@@ -48,7 +54,7 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     workspace: Workspace
   ) async throws {
     self.sourceKitLSPServer = sourceKitLSPServer
-    self.documentationManager = DocCDocumentationManager(buildServerManager: workspace.buildServerManager)
+    self.documentationManager = DocCDocumentationManager(buildServerManager: workspace.buildServerManager);
   }
 
   package nonisolated func canHandle(toolchain: Toolchain) -> Bool {
@@ -99,5 +105,319 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     edits: [SwiftSyntax.SourceEdit]
   ) async {
     // The DocumentationLanguageService does not do anything with document events
+  }
+
+  package func definition(_ req: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+    let snapshot = try self.documentManager.latestSnapshot(req.textDocument.uri)
+    let clickedSymbol: String?
+    switch snapshot.language {
+    case .swift:
+      clickedSymbol = extractSymbolFromDocComment(snapshot: snapshot, at: req.position)
+    case .markdown, .tutorial:
+      clickedSymbol = extractSymbolFromText(snapshot.text, at: req.position)
+    default:
+      return nil
+    }
+
+    guard let clickedSymbol else {
+      logger.debug("No symbol link found at the cursor position")
+      return nil
+    }
+
+    guard let symbolLink = DocCSymbolLink(linkString: clickedSymbol) else {
+      logger.debug("Failed to parse DocC symbol link from '\(clickedSymbol)'")
+      return nil
+    }
+
+    guard let sourceKitLSPServer else {
+      logger.debug("sourceKit-LSP server could not be resolved")
+      return nil
+    }
+
+    guard let workspace = await sourceKitLSPServer.workspaceForDocument(uri: req.textDocument.uri) else {
+      logger.debug("No workspace found for document \(req.textDocument.uri)")
+      return nil
+    }
+
+    guard let index = await workspace.index(checkedFor: .deletedFiles) else {
+      logger.debug("No index available for workspace")
+      return nil
+    }
+
+    let occurrence = try await sourceKitLSPServer.withOnDiskDocumentManager { onDiskDocumentManager in
+      try await index.primaryDefinitionOrDeclarationOccurrence(
+        ofDocCSymbolLink: symbolLink,
+        fetchSymbolGraph: { location in
+          guard let uri = location.uri else { return nil }
+          return try await workspace.primaryLanguageService(
+            for: uri,
+            workspace.buildServerManager.defaultLanguageInCanonicalTarget(for: uri)
+          ).symbolGraph(forOnDiskContentsAt: location, in: workspace, manager: onDiskDocumentManager)
+        }
+      )
+    }
+
+    guard let targetLocation = occurrence?.location.lspLocation else {
+      logger.debug("No definition occurrence found in the index for symbol link '\(clickedSymbol)'")
+      return nil
+    }
+    return .locations([targetLocation])
+  }
+
+  /// Walks the Markdown/DocC AST looking for symbol link
+  /// that contains a given source position.
+  private struct SymbolLocator: MarkupWalker {
+    let target: Markdown.SourceLocation
+    var found: String?
+
+    init(target: Markdown.SourceLocation) {
+      self.target = target
+    }
+
+    private func contains(_ range: Markdown.SourceRange?) -> Bool {
+      guard let range else { return false }
+      return range.lowerBound <= target && target < range.upperBound
+    }
+
+    mutating func visitSymbolLink(_ symbolLink: SymbolLink) {
+      guard
+        found == nil,
+        contains(symbolLink.range),
+        let destination = symbolLink.destination,
+        let range = symbolLink.range
+      else {
+        return
+      }
+      // Symbol links are delimited by two backticks (``Foo/bar``),
+      // so the destination starts two columns past the start of the link.
+      let destinationStartColumn = range.lowerBound.column + 2
+      let relativeColumn = target.column - destinationStartColumn
+      guard relativeColumn >= 0 else {
+        return
+      }
+      let components = destination.split(separator: "/")
+      var currentLength = 0
+
+      for (index, component) in components.enumerated() {
+        let end = currentLength + component.utf8.count
+        if relativeColumn < end {
+          found = components[0...index].joined(separator: "/")
+          return
+        }
+        currentLength = end + 1  // Skip '/'
+      }
+      found = destination
+    }
+
+    mutating func defaultVisit(_ markup: any Markup) {
+      guard found == nil else { return }
+      descendInto(markup)
+    }
+  }
+
+  private func extractSymbolFromText(_ text: String, at position: Position) -> String? {
+    let lines = text.components(separatedBy: "\n")
+    var column = position.utf16index + 1
+    if position.line < lines.count {
+      column = utf8Offset(inLine: lines[position.line], forUTF16Offset: position.utf16index) + 1
+    }
+    // LSP positions are 0-based; swift-markdown SourceLocation is 1-based.
+    let target = Markdown.SourceLocation(
+      line: position.line + 1,
+      column: column,
+      source: nil
+    )
+
+    let document = Markdown.Document(parsing: text, options: [.parseSymbolLinks, .parseBlockDirectives])
+    var locator = SymbolLocator(target: target)
+    locator.visit(document)
+
+    guard let symbol = locator.found, !symbol.isEmpty else {
+      return nil
+    }
+    return symbol
+  }
+
+  private struct DocCCommentLine {
+    let text: String
+    let startLine: Int
+    let startColumn: Int
+  }
+
+  private enum DocTriviaGroup {
+    case lines(lines: [DocCCommentLine])
+    case block(text: String, startLine: Int, startColumn: Int)
+  }
+
+  /// Converts a UTF-16 offset within `line` into the corresponding UTF-8 byte offset.
+  /// Needed because DocumentSnapshot/LSP positions are UTF-16-based but
+  /// Markdown.SourceLocation.column is UTF-8-byte-based.
+  private func utf8Offset(inLine line: String, forUTF16Offset utf16Offset: Int) -> Int {
+    let utf16View = line.utf16
+    guard
+      let utf16Index = utf16View.index(utf16View.startIndex, offsetBy: utf16Offset, limitedBy: utf16View.endIndex),
+      let stringIndex = utf16Index.samePosition(in: line)
+    else {
+      return line.utf8.count
+    }
+    return line.utf8.distance(from: line.startIndex, to: stringIndex)
+  }
+
+  private func docCommentGroups(
+    in trivia: Trivia,
+    tokenStart: AbsolutePosition,
+    snapshot: DocumentSnapshot
+  ) -> [DocTriviaGroup] {
+    var groups: [DocTriviaGroup] = []
+    var pendingLines: [DocCCommentLine] = []
+    var offset = tokenStart.utf8Offset
+    var newlinesSinceLastDoc = 0
+
+    func flushPendingLines() {
+      guard !pendingLines.isEmpty else { return }
+      groups.append(.lines(lines: pendingLines))
+      pendingLines.removeAll()
+    }
+
+    for piece in trivia.pieces {
+      defer { offset += piece.sourceLength.utf8Length }
+      switch piece {
+      case .docLineComment(let text):
+        if newlinesSinceLastDoc > 1 { flushPendingLines() }
+        let position = snapshot.positionOf(utf8Offset: offset)
+        pendingLines.append(DocCCommentLine(text: text, startLine: position.line, startColumn: position.utf16index))
+        newlinesSinceLastDoc = 0
+      case .docBlockComment(let text):
+        flushPendingLines()
+        let position = snapshot.positionOf(utf8Offset: offset)
+        groups.append(.block(text: text, startLine: position.line, startColumn: position.utf16index))
+        newlinesSinceLastDoc = 0
+      case .newlines(let n), .carriageReturns(let n), .carriageReturnLineFeeds(let n):
+        newlinesSinceLastDoc += n
+      case .spaces, .tabs:
+        break
+      default:
+        flushPendingLines()
+        newlinesSinceLastDoc = 0
+      }
+    }
+    flushPendingLines()
+    return groups
+  }
+
+  private struct StrippedLine {
+    let text: String
+    let strippedPrefixCount: Int
+  }
+
+  /// Strips `///` and any leading whitespace from each line comment piece.
+  private func stripLineCommentDelimiters(
+    _ lines: [DocCCommentLine]
+  ) -> [StrippedLine] {
+    return lines.map { line in
+      var text = Substring(line.text)
+      var stripped = 0
+
+      if text.hasPrefix("///") {
+        text.removeFirst(3)
+        stripped += 3
+      }
+      let beforeIndent = text.utf16.count
+      let trimmed = text.drop { $0 == " " || $0 == "\t" }
+      stripped += beforeIndent - trimmed.utf16.count
+      text = trimmed
+      return StrippedLine(text: String(text), strippedPrefixCount: stripped)
+    }
+  }
+  /// Strips `/**`, `*/`, and per-line leading `*`/whitespace from a block comment.
+  private func stripBlockCommentDelimiters(_ text: String) -> [StrippedLine] {
+    let lines = text.components(separatedBy: "\n")
+    return lines.enumerated().map { index, rawLine in
+      var line = Substring(rawLine)
+      var stripped = 0
+
+      if index == 0, line.hasPrefix("/**") {
+        line.removeFirst(3)
+        stripped += 3
+      }
+      if index == lines.count - 1, line.hasSuffix("*/") {
+        line.removeLast(2)
+      }
+      let beforeStar = line.utf16.count
+      var afterStar = line.drop { $0 == " " || $0 == "\t" }
+      if afterStar.hasPrefix("*") {
+        afterStar.removeFirst()
+        if afterStar.hasPrefix(" ") { afterStar.removeFirst() }
+        stripped += beforeStar - afterStar.utf16.count
+        line = afterStar
+      }
+      let beforeIndent = line.utf16.count
+      let trimmed = line.drop { $0 == " " || $0 == "\t" }
+      stripped += beforeIndent - trimmed.utf16.count
+      line = trimmed
+      return StrippedLine(text: String(line), strippedPrefixCount: stripped)
+    }
+  }
+
+  private func resolveTarget(
+    for group: DocTriviaGroup,
+    cursorPosition: Position
+  ) -> (strippedLines: [StrippedLine], lineIndex: Int, targetColumn: Int)? {
+    switch group {
+    case .lines(let lines):
+      guard let lineIndex = lines.firstIndex(where: { $0.startLine == cursorPosition.line }) else {
+        return nil
+      }
+      let strippedLines = stripLineCommentDelimiters(lines)
+      let rawLine = lines[lineIndex].text
+      let relativeUTF16 = cursorPosition.utf16index - lines[lineIndex].startColumn
+      let utf8Rel = utf8Offset(inLine: rawLine, forUTF16Offset: relativeUTF16)
+      let targetColumn = max(1, utf8Rel - strippedLines[lineIndex].strippedPrefixCount + 1)
+
+      return (strippedLines, lineIndex, targetColumn)
+
+    case .block(let text, let startLine, let startColumn):
+      let strippedLines = stripBlockCommentDelimiters(text)
+      let lineIndex = cursorPosition.line - startLine
+      guard strippedLines.indices.contains(lineIndex) else {
+        return nil
+      }
+
+      let rawLines = text.components(separatedBy: "\n")
+      let rawLine = rawLines[lineIndex]
+      let columnBase = lineIndex == 0 ? startColumn : 0
+      let relativeUTF16 = cursorPosition.utf16index - columnBase
+      let utf8Rel = utf8Offset(inLine: rawLine, forUTF16Offset: relativeUTF16)
+      let targetColumn = max(1, utf8Rel - strippedLines[lineIndex].strippedPrefixCount + 1)
+
+      return (strippedLines, lineIndex, targetColumn)
+    }
+  }
+
+  private func extractSymbolFromDocComment(snapshot: DocumentSnapshot, at position: Position) -> String? {
+    let sourceFile = SwiftParser.Parser.parse(source: snapshot.text)
+    let absolutePosition = snapshot.absolutePosition(of: position)
+
+    guard let token = sourceFile.token(at: absolutePosition) else {
+      return nil
+    }
+
+    let cursorPosition = snapshot.positionOf(utf8Offset: absolutePosition.utf8Offset)
+    let groups = docCommentGroups(in: token.leadingTrivia, tokenStart: token.position, snapshot: snapshot)
+
+    for group in groups {
+      guard let (strippedLines, lineIndex, targetColumn) = resolveTarget(for: group, cursorPosition: cursorPosition)
+      else {
+        continue
+      }
+
+      let combinedText = strippedLines.map(\.text).joined(separator: "\n")
+      let target = Markdown.SourceLocation(line: lineIndex + 1, column: targetColumn, source: nil)
+      let document = Markdown.Document(parsing: combinedText, options: [.parseSymbolLinks])
+      var locator = SymbolLocator(target: target)
+      locator.visit(document)
+      return locator.found
+    }
+    return nil
   }
 }
--- a/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
+++ b/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
@@ -163,8 +163,10 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     return .locations([targetLocation])
   }
 
-  /// Walks the Markdown/DocC AST looking for symbol link
-  /// that contains a given source position.
+  /// Walks the Markdown/DocC AST looking for a symbol link that contains a
+  /// given source position, and sets `found` to the link's destination
+  /// truncated to the component under the cursor (e.g. `Foo/bar/baz` with
+  /// the cursor on `bar` sets `found` to `"Foo/bar"`), not the full destination.
   private struct SymbolLocator: MarkupWalker {
     let target: Markdown.SourceLocation
     var found: String?
@@ -239,13 +241,12 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
 
   private struct DocCCommentLine {
     let text: String
-    let startLine: Int
-    let startColumn: Int
+    let start: Position
   }
 
   private enum DocTriviaGroup {
-    case lines(lines: [DocCCommentLine])
-    case block(text: String, startLine: Int, startColumn: Int)
+    case lines([DocCCommentLine])
+    case block(text: String, start: Position)
   }
 
   /// Converts a UTF-16 offset within `line` into the corresponding UTF-8 byte offset.
@@ -274,7 +275,7 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
 
     func flushPendingLines() {
       guard !pendingLines.isEmpty else { return }
-      groups.append(.lines(lines: pendingLines))
+      groups.append(.lines(pendingLines))
       pendingLines.removeAll()
     }
 
@@ -284,12 +285,12 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
       case .docLineComment(let text):
         if newlinesSinceLastDoc > 1 { flushPendingLines() }
         let position = snapshot.positionOf(utf8Offset: offset)
-        pendingLines.append(DocCCommentLine(text: text, startLine: position.line, startColumn: position.utf16index))
+        pendingLines.append(DocCCommentLine(text: text, start: position))
         newlinesSinceLastDoc = 0
       case .docBlockComment(let text):
         flushPendingLines()
         let position = snapshot.positionOf(utf8Offset: offset)
-        groups.append(.block(text: text, startLine: position.line, startColumn: position.utf16index))
+        groups.append(.block(text: text, start: position))
         newlinesSinceLastDoc = 0
       case .newlines(let n), .carriageReturns(let n), .carriageReturnLineFeeds(let n):
         newlinesSinceLastDoc += n
@@ -303,7 +304,10 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     flushPendingLines()
     return groups
   }
-
+  
+  /// A line of documentation text after stripping its comment delimiters/prefix.
+  /// `strippedPrefixCount` is the number of UTF-16 code units removed from the
+  /// beginning of the original line to produce `text`.
   private struct StrippedLine {
     let text: String
     let strippedPrefixCount: Int
@@ -328,6 +332,7 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
       return StrippedLine(text: String(text), strippedPrefixCount: stripped)
     }
   }
+
   /// Strips `/**`, `*/`, and per-line leading `*`/whitespace from a block comment.
   private func stripBlockCommentDelimiters(_ text: String) -> [StrippedLine] {
     let lines = text.components(separatedBy: "\n")
@@ -361,59 +366,49 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
   private func resolveTarget(
     for group: DocTriviaGroup,
     cursorPosition: Position
-  ) -> (strippedLines: [StrippedLine], lineIndex: Int, targetColumn: Int)? {
+  ) -> (strippedLines: [StrippedLine], target: Position)? {
     switch group {
     case .lines(let lines):
-      guard let lineIndex = lines.firstIndex(where: { $0.startLine == cursorPosition.line }) else {
+      guard let lineIndex = lines.firstIndex(where: { $0.start.line == cursorPosition.line }) else {
         return nil
       }
       let strippedLines = stripLineCommentDelimiters(lines)
-      let rawLine = lines[lineIndex].text
-      let relativeUTF16 = cursorPosition.utf16index - lines[lineIndex].startColumn
-      let utf8Rel = utf8Offset(inLine: rawLine, forUTF16Offset: relativeUTF16)
-      let targetColumn = max(1, utf8Rel - strippedLines[lineIndex].strippedPrefixCount + 1)
+      let relativeUTF16 = cursorPosition.utf16index - lines[lineIndex].start.utf16index
+      let localColumn = max(0, relativeUTF16 - strippedLines[lineIndex].strippedPrefixCount)
+      return (strippedLines, Position(line: lineIndex, utf16index: localColumn))
 
-      return (strippedLines, lineIndex, targetColumn)
-
-    case .block(let text, let startLine, let startColumn):
+    case .block(let text, let start):
       let strippedLines = stripBlockCommentDelimiters(text)
-      let lineIndex = cursorPosition.line - startLine
-      guard strippedLines.indices.contains(lineIndex) else {
-        return nil
-      }
-
-      let rawLines = text.components(separatedBy: "\n")
-      let rawLine = rawLines[lineIndex]
-      let columnBase = lineIndex == 0 ? startColumn : 0
+      let lineIndex = cursorPosition.line - start.line
+      guard strippedLines.indices.contains(lineIndex) else { return nil }
+      let columnBase = lineIndex == 0 ? start.utf16index : 0
       let relativeUTF16 = cursorPosition.utf16index - columnBase
-      let utf8Rel = utf8Offset(inLine: rawLine, forUTF16Offset: relativeUTF16)
-      let targetColumn = max(1, utf8Rel - strippedLines[lineIndex].strippedPrefixCount + 1)
-
-      return (strippedLines, lineIndex, targetColumn)
+      let localColumn = max(0, relativeUTF16 - strippedLines[lineIndex].strippedPrefixCount)
+      return (strippedLines, Position(line: lineIndex, utf16index: localColumn))
     }
   }
 
   private func extractSymbolFromDocComment(snapshot: DocumentSnapshot, at position: Position) -> String? {
     let sourceFile = SwiftParser.Parser.parse(source: snapshot.text)
     let absolutePosition = snapshot.absolutePosition(of: position)
+    guard let token = sourceFile.token(at: absolutePosition) else { return nil }
 
-    guard let token = sourceFile.token(at: absolutePosition) else {
-      return nil
-    }
+    let leadingGroups = docCommentGroups(in: token.leadingTrivia, tokenStart: token.position, snapshot: snapshot)
+    let trailingGroups = docCommentGroups(
+      in: token.trailingTrivia,
+      tokenStart: token.endPositionBeforeTrailingTrivia,
+      snapshot: snapshot
+    )
 
-    let cursorPosition = snapshot.positionOf(utf8Offset: absolutePosition.utf8Offset)
-    let groups = docCommentGroups(in: token.leadingTrivia, tokenStart: token.position, snapshot: snapshot)
-
-    for group in groups {
-      guard let (strippedLines, lineIndex, targetColumn) = resolveTarget(for: group, cursorPosition: cursorPosition)
-      else {
-        continue
-      }
+    for group in leadingGroups + trailingGroups {
+      guard let (strippedLines, target) = resolveTarget(for: group, cursorPosition: position) else { continue }
 
       let combinedText = strippedLines.map(\.text).joined(separator: "\n")
-      let target = Markdown.SourceLocation(line: lineIndex + 1, column: targetColumn, source: nil)
+      let utf8Column = utf8Offset(inLine: strippedLines[target.line].text, forUTF16Offset: target.utf16index) + 1
+      let markdownTarget = Markdown.SourceLocation(line: target.line + 1, column: utf8Column, source: nil)
+
       let document = Markdown.Document(parsing: combinedText, options: [.parseSymbolLinks])
-      var locator = SymbolLocator(target: target)
+      var locator = SymbolLocator(target: markdownTarget)
       locator.visit(document)
       return locator.found
     }
--- a/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
+++ b/Sources/DocumentationLanguageService/DocumentationLanguageService.swift
@@ -41,8 +41,7 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
 
   package static var experimentalCapabilities: [String: LSPAny] {
     return [
-      DoccDocumentationRequest.method: ["version": 1],
-      "definitionProvider": .bool(true),
+      DoccDocumentationRequest.method: ["version": 1]
     ]
   }
 
@@ -54,7 +53,7 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     workspace: Workspace
   ) async throws {
     self.sourceKitLSPServer = sourceKitLSPServer
-    self.documentationManager = DocCDocumentationManager(buildServerManager: workspace.buildServerManager);
+    self.documentationManager = DocCDocumentationManager(buildServerManager: workspace.buildServerManager)
   }
 
   package nonisolated func canHandle(toolchain: Toolchain) -> Bool {
--- a/Sources/SourceKitLSP/LanguageService.swift
+++ b/Sources/SourceKitLSP/LanguageService.swift
@@ -99,6 +99,13 @@ package struct GeneratedInterfaceDetails: ResponseType, Hashable {
   }
 }
 
+/// Thrown by a language service's request handler to indicate that this
+/// service isn't the right one for the current request context, and the
+/// next language service should be tried instead.
+package struct FallThroughToNextLanguageService: Error {
+  package init() {}
+}
+
 /// Provides language specific functionality to sourcekit-lsp from a specific toolchain.
 ///
 /// For example, we may have a language service that provides semantic functionality for c-family using a clangd server,
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -392,7 +392,7 @@ package actor SourceKitLSPServer {
       @Sendable @escaping (
         RequestType, Workspace, any LanguageService
       ) async throws ->
-      RequestType.Response?
+      RequestType.Response
   ) async {
     await request.reply {
       let request = request.params
@@ -404,14 +404,26 @@ package actor SourceKitLSPServer {
       if languageServices.isEmpty {
         throw ResponseError.unknown("No language service for '\(request.textDocument.uri)' found")
       }
-      // Return the results from the first language service that doesn't throw a `requestNotImplemented` error.
-      for languageService in languageServices {
+      // Return the results from the first language service that produces a non-nil response.
+      // A language service can opt out of handling this request for the given service by throwing
+      // `ResponseError.requestNotImplemented` (it never implements this method) or
+      // `FallThroughToNextLanguageService` (it implements the method, but isn't the right service
+      // for this particular request), in which case the next language service is tried.
+      for (index, languageService) in languageServices.enumerated() {
         do {
-          guard let response = try await requestHandler(request, workspace, languageService) else {
-            continue
-          }
-          return response
+          return try await requestHandler(request, workspace, languageService)
         } catch let error as ResponseError where error.code == .requestNotImplemented {
+          continue
+        } catch is FallThroughToNextLanguageService {
+          if index + 1 < languageServices.count {
+            logger.debug(
+              "\(type(of: languageService)) is not the right service for \(type(of: request).method); falling through to \(type(of: languageServices[index + 1]))"
+            )
+          } else {
+            logger.debug(
+              "\(type(of: languageService)) is not the right service for \(type(of: request).method); no further language services to try"
+            )
+          }
           continue
         }
       }
@@ -1969,7 +1981,7 @@ extension SourceKitLSPServer {
     workspace: Workspace,
     languageService: any LanguageService
   ) async throws -> LocationsOrLocationLinksResponse? {
-    var indexBasedResponse: [Location] = []
+    let indexBasedResponse: [Location]
     do {
       indexBasedResponse = try await indexBasedDefinition(req, workspace: workspace, languageService: languageService)
     } catch {
@@ -1990,12 +2002,14 @@ extension SourceKitLSPServer {
       } catch let error as ResponseError where error.code == .requestNotImplemented {
         // Signal to handleRequest's loop to try the next language service
         throw error
+      } catch is FallThroughToNextLanguageService {
+        // Signal to handleRequest's loop to try the next language service
+        throw FallThroughToNextLanguageService()
       } catch {
         logger.info("Fallback definition request failed: \(error.forLogging)")
         return nil
       }
     }
-
     let remappedLocations = indexBasedResponse.adjusted(for: copiedFileMap)
     return .locations(remappedLocations)
   }
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -392,7 +392,7 @@ package actor SourceKitLSPServer {
       @Sendable @escaping (
         RequestType, Workspace, any LanguageService
       ) async throws ->
-      RequestType.Response
+      RequestType.Response?
   ) async {
     await request.reply {
       let request = request.params
@@ -407,7 +407,10 @@ package actor SourceKitLSPServer {
       // Return the results from the first language service that doesn't throw a `requestNotImplemented` error.
       for languageService in languageServices {
         do {
-          return try await requestHandler(request, workspace, languageService)
+          guard let response = try await requestHandler(request, workspace, languageService) else {
+            continue
+          }
+          return response
         } catch let error as ResponseError where error.code == .requestNotImplemented {
           continue
         }
@@ -1966,7 +1969,13 @@ extension SourceKitLSPServer {
     workspace: Workspace,
     languageService: any LanguageService
   ) async throws -> LocationsOrLocationLinksResponse? {
-    let indexBasedResponse = try await indexBasedDefinition(req, workspace: workspace, languageService: languageService)
+    var indexBasedResponse: [Location] = []
+    do {
+      indexBasedResponse = try await indexBasedDefinition(req, workspace: workspace, languageService: languageService)
+    } catch {
+      logger.info("indexBasedDefinition failed: \(error.forLogging)")
+      indexBasedResponse = []
+    }
     let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
     // If we're unable to handle the definition request using our index, see if the
     // language service can handle it (e.g. clangd can provide AST based definitions).
@@ -1975,11 +1984,18 @@ extension SourceKitLSPServer {
     // `SwiftLanguageService` will always respond with `unsupported method`. Thus, only log such a failure instead of
     // returning it to the client.
     if indexBasedResponse.isEmpty {
-      return await orLog("Fallback definition request", level: .info) {
+      do {
         let result = try await languageService.definition(req)
         return result?.adjusted(for: copiedFileMap)
+      } catch let error as ResponseError where error.code == .requestNotImplemented {
+        // Signal to handleRequest's loop to try the next language service
+        throw error
+      } catch {
+        logger.info("Fallback definition request failed: \(error.forLogging)")
+        return nil
       }
     }
+
     let remappedLocations = indexBasedResponse.adjusted(for: copiedFileMap)
     return .locations(remappedLocations)
   }
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2002,9 +2002,9 @@ extension SourceKitLSPServer {
       } catch let error as ResponseError where error.code == .requestNotImplemented {
         // Signal to handleRequest's loop to try the next language service
         throw error
-      } catch is FallThroughToNextLanguageService {
+      } catch let error as FallThroughToNextLanguageService {
         // Signal to handleRequest's loop to try the next language service
-        throw FallThroughToNextLanguageService()
+        throw error
       } catch {
         logger.info("Fallback definition request failed: \(error.forLogging)")
         return nil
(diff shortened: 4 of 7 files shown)


matthewbastien:
@swift-ci please test

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -830,7 +829,16 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
       arguments += ["--traits", traits.joined(separator: ",")]
     }
     arguments += toolsets.flatMap { ["--toolset", $0.pathString] }
-    arguments += pkgConfigDirectories.flatMap { ["--pkg-config-path", $0.pathString] }
```
ahoppen, line 838: Unintentionally removed?

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -830,7 +829,16 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
+
+    let symbolGraphPath = self.swiftPMWorkspace.location.scratchDirectory.parentDirectory.appending(
+      component: "symbol-graphs"
+    ).pathString
+    arguments += [
+      "-Xswiftc", "-Xfrontend", "-Xswiftc", "-emit-symbol-graph",
+      "-Xswiftc", "-Xfrontend", "-Xswiftc", "-emit-symbol-graph-dir",
+      "-Xswiftc", "-Xfrontend", "-Xswiftc", symbolGraphPath,
+    ]
+
```
ahoppen: We shouldn’t emit symbol graphs during target preparation. The goal of target preparation is to get `.swiftmodule` files out there as quickly as possible so that we can index the files depending on those modules. So, all the symbol graph generation should happen during indexing, ie. `UpdateIndexStoreTaskDescription`

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
   ) async {
     // The DocumentationLanguageService does not do anything with document events
   }
+
+  package func symbolInfo(_ req: SymbolInfoRequest) async throws -> [SymbolDetails] {
+    return []
+  }
+
+  package func definition(_ req: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+
```
ahoppen: Superfluous newline

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
   ) async {
     // The DocumentationLanguageService does not do anything with document events
   }
+
+  package func symbolInfo(_ req: SymbolInfoRequest) async throws -> [SymbolDetails] {
+    return []
```
ahoppen: Do we need this stub?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    let sourceFile = SwiftParser.Parser.parse(source: snapshot.text)
+    let absolutePosition = snapshot.absolutePosition(of: position)
+
+    guard let token = sourceFile.token(at: absolutePosition) else {
+      return nil
+    }
+
+    let converter = SourceLocationConverter(fileName: "", tree: sourceFile)
+    let cursorLine = converter.location(for: absolutePosition).line
+    let cursorColumn = converter.location(for: absolutePosition).column
```
ahoppen: Instead of creating a `SourceLocationConverter`, there should be some method on `DocumentSnapshot` that gives you the source location you need. That’s going to be faster because it already has a pre-populated line table.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+
+    var triviaOffset = token.position.utf8Offset
+
+    for piece in token.leadingTrivia.pieces {
+      defer { triviaOffset += piece.sourceLength.utf8Length }
+
+      switch piece {
+      case .docLineComment(let commentText):
```
ahoppen: Is it correct to only parse the line on which the user clicked? Couldn’t a previous line influence how the markdown gets parsed, eg. if the previous line starts a code block with three backticks.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+        var locator = SymbolLocator(target: target)
+        locator.visit(document)
+
+        if let found = locator.found {
+          return found
+        }
+
+      case .docBlockComment:
+        // Symbol link navigation within /** */ blocks is not yet supported.
+        continue
```
ahoppen: I feel like we should support his straight away. Shouldn’t be too difficult to add, right?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      guard found == nil else { return }
+      descendInto(markup)
+    }
+  }
+
+  private func extractSymbolFromText(_ text: String, at position: Position) -> String? {
+    // LSP positions are 0-based; swift-markdown SourceLocation is 1-based.
+    let target = Markdown.SourceLocation(
+      line: position.line + 1,
+      column: position.utf16index + 1,
```
ahoppen: Are Markdown positions also based on the UTF-16 index or do they use UTF-8 column indices? You should be able to see the difference if you have a line that contains a complex emoji like 🧑‍🧑‍🧒‍🧒 before the cursor.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+  }
+
+  package func definition(_ req: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+
+    let snapshot = try self.documentManager.latestSnapshot(req.textDocument.uri)
+    let clickedSymbol: String?
+    if snapshot.language == .swift {
+      clickedSymbol = extractSymbolFromDocComment(snapshot: snapshot, at: req.position)
+    } else {
+      clickedSymbol = extractSymbolFromText(snapshot.text, at: req.position)
```
ahoppen, line 117: Should this only be done for Markdown files? I assume you don’t want to parse eg. C files as Markdown.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+
+    let workspaceRoot = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+    let targetGraphURL =
+      workspaceRoot
+      .appendingPathComponent(".build/symbol-graphs")
+      .appendingPathComponent("\(moduleName).symbols.json")
+
+    guard let data = try? Data(contentsOf: targetGraphURL),
+      let jsonObject = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
+      let symbolsArray = jsonObject["symbols"] as? [[String: Any]]
```
ahoppen: Please use a `Codable` struct here instead of relying on `JSONSerialization`, which really dates back to Objective-C.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +107,207 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      let moduleName = await self.workspace.buildServerManager.moduleName(for: targetID)
+    else {
+      return nil
+    }
+
+    let workspaceRoot = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+    let targetGraphURL =
+      workspaceRoot
+      .appendingPathComponent(".build/symbol-graphs")
+      .appendingPathComponent("\(moduleName).symbols.json")
```
ahoppen: Does this also work if the referenced symbol is in a different module than the current file?

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+@_spi(SourceKitLSP) import ToolsProtocolsSwiftExtensions
+
+import struct TSCBasic.AbsolutePath
+import class TSCBasic.Process
+import struct TSCBasic.ProcessResult
+import enum TSCBasic.SystemError
+
+private let updateSymbolGraphIDForLogging = Atomic<UInt32>(1)
+
+package struct UpdateSymbolGraphTaskDescription: IndexTaskDescription {
```
ahoppen: Would it be possible to generate the index store during the current indexing operation `UpdateIndexStoreTaskDescription`? Both indexing and symbol graph generation need to parse and type-check the file and we should be able to share that work.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -793,6 +793,25 @@ extension SwiftLanguageService {
...
+        let isInDocComment = token.leadingTrivia.pieces.contains {
+          switch $0 {
+          case .docLineComment, .docBlockComment:
+            return true
+          default:
+            return false
+          }
+        }
+        if isInDocComment {
+          throw ResponseError.requestNotImplemented(DefinitionRequest.self)
```
ahoppen: I feel like `requestNotImplemented` is the wrong signal here because the method is indeed implemented. I’d rather return `nil` to indicate that the language service wasn’t able to resolve a symbol and then `SourceKitLSPServer` can try the next language service.

ahoppen (commented):
Sorry for taking so long to review this. I left some thoughts inline. Also: Could you add some test cases that test the new behavior?

matthewbastien:
@swift-ci please test

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
Padmashree06: Yes, that was unintentional. I'll add it back!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes, this stub is required. Without it, `DocumentationLanguageService` returns a "request not implemented" error when `SymbolInfo` is called through `IndexBasedDefinition`. As a result, `IndexBasedDefinition` never returns an empty array, so the fallback condition `if indexBasedResponse.isEmpty { ... }` is never reached.

With this stub, `IndexBasedDefinition` returns an empty array, allowing the fallback to execute and enables go-to-definition for DocC comments, tutorials, and Markdown files.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: No, Markdown positions are based on UTF-8 index. I'll add the conversion! Thanks for pointing that out!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes,will add it!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: The `else` block is only reached for Markdown and Tutorial files. Since `DocumentationLanguageService` is registered only for `.swift`, `.markdown`, and `.tutorial` files, it won't attempt to parse C files as Markdown.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: No, currently it does not.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
ahoppen: Maybe we should do the fallback path even if `indexBasedDefinition` throws then, similar to how we try more language services if one of them throws. I’d just like to log all the errors thrown by both `indexBasedDefinition` and all language services tried in case `definition` fails.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
ahoppen: I would still add the check. It’s not unreasonable to assume that at some point we’ll also want to support C-based languages for Docc comments.

Padmashree06: Yes! Will add it!

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: I tried combining them initially, but the symbol graph wasn't getting generated. From what I found, symbol graph generation requires a module-level build, while the index store task does a file-level build. So, to keep the module-level compilation cheap, `UpdateSymbolGraphTask` uses the `-experimental-skip-all-function-bodies` flag, but using that flag for the index store task wouldn't work well, since we'd need to parse and type-check the code inside function bodies too.
So I kept them separate for now, but if you see a way to combine them, I'd be happy to give it a try.

Sources/SwiftLanguageService/SwiftLanguageService.swift
Padmashree06: From my understanding returning `nil` won't fall back to the next registered language service. `SourceKitLSPServer.handleRequest` only continues to the next language service when it catches a `ResponseError` with the .`requestNotImplemented` code. A plain `nil` return is treated as a successful result, so the loop stops there, because of which the `DocumenationLanguageService` is never triggered for Swift files and hence the go-to-definition feature for DocC comments in Swift files will not work. 

If I'm missing something or there's a better way to approach this, please let me know.

ahoppen: > From my understanding returning `nil` won't fall back to the next registered language

I don’t think there’s a reason why we can’t change that behavior.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
ahoppen: > From what I found, symbol graph generation requires a module-level build

I can’t think of why that would be the case. During debug builds, the driver will already invoke multiple swift-frontend commands with only a subset of the files in them, so  we should be able to generate a symbol graph for individual files as well. It could be that symbol graph generation is incompatible with some of the flags we pass for indexing, but I’d like to understand what they are before making any decisions here.

Padmashree06: Yes, Symbol Graph Generation is incompatible with `-index-file`, `-index-store-path` and `-disable-batch-mode` flags.

Sources/SwiftLanguageService/SwiftLanguageService.swift
Padmashree06: Yes, changed the behavior to continue to next language service on returning nil.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
ahoppen: I just dug into the compiler and symbol graph extraction happens while writing out the module file [here](https://github.com/swiftlang/swift/blob/60c7f491d4520b1100c10c96b3d073e3474a28fe/lib/Frontend/Serialization.cpp#L180). Technically, I think all the information for symbol graph generation should be available after type checking, there’s no need to also run SILGen to write it out. That also explains why symbol graph extraction is incompatible with `-index-file`, which stops after type checking and doesn’t emit a Swift module.

Ideally, we should investigate whether we could pull the symbol graph generation to an earlier place in the compilation pipeline so that it can run together with `-index-file`. But I suppose that’s out of scope for your GSoC project.

With that in mind, I think we might be fine here because we use `-experimental-skip-all-function-bodies`, which should bound the time spent during symbol graph generation. Still, I would like to see measurements how how long symbol graph generation takes compared to indexing when running background indexing on a reasonably sized project like SourceKit-LSP, SwiftPM, SwiftSyntax or whatever your favorite is. Should symbol graph generation take more time than indexing, I would be extremely hesitant to take this change outside of an experimental feature flag, if it adds <20%, I would probably be fine, anything in between we’d need to weigh our options.

I’m also going to leave a few comments on the compiler argument processing that you’re doing, I think there’s some room for improvement.

Also, I’d like to hear @hamishknight and @rintaro’s opinions.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+
+      var baseArgs: [String] = ["-I", customModulesDir.pathString]
+
+      baseArgs += ["-module-name", moduleName]
+
+      var i = 0
+      while i < rawArgs.count {
+        let arg = rawArgs[i]
+
+        if arg == "-I" || arg == "-F" || arg == "-target" || arg == "-sdk"
```
ahoppen: Why are all of these transformations necessary when they aren’t necessary for background indexing? Generally, indexing only works for the contents of the files on disk, so we shouldn’t have any `vscode-local:` files.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+          baseArgs.append(arg)
+          i += 1
+          continue
+        }
+
+        i += 1
+      }
+
+      if !baseArgs.contains("-wmo") && !baseArgs.contains("-whole-module-optimization") {
+        baseArgs.append("-wmo")
```
ahoppen: Why do we need `-wmo`? Symbol graph generations should work without any kind of optimization, I would imagine.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+      }
+
+      // --- Clean environment ---
+      var cleanEnvironment = ProcessInfo.processInfo.environment
+      cleanEnvironment.removeValue(forKey: "SWIFT_DRIVER_SUPPLEMENTARY_OUTPUT_FILE_MAP")
+      cleanEnvironment.removeValue(forKey: "SWIFT_DRIVER_OUTPUT_FILE_MAP")
+      cleanEnvironment.removeValue(forKey: "SWIFT_DRIVER_TEMP_DIR")
+      cleanEnvironment.removeValue(forKey: "SWIFT_DRIVER_RESPONSE_FILE_PATH")
+      cleanEnvironment.removeValue(forKey: "SWIFT_DRIVER_TOOLNAME")
```
ahoppen: Why are these needed if they aren’t needed for background indexing?

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+
+  package func execute() async {
+    await withLoggingSubsystemAndScope(subsystem: indexLoggingSubsystem, scope: "update-symbolgraph-\(id % 100)") {
+      guard language == .swift else { return }
+
+      let buildSettings: FileBuildSettings
+      let swiftc: URL
+
+      do {
+        guard let firstFile = filesToIndex.first?.file.mainFile else { return }
```
ahoppen: We shouldn’t assume that all files in the batch have the same build settings. See what `UpdateIndexStoreTaskDescription` does to collect files that have the same build settings.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+      let taskId = "symbol-graph-\(id)"
+
+      logMessageToIndexLog(
+        "Updating symbol graph for target: \(target.uri.stringValue)",
+        .info,
+        .begin(StructuredLogBegin(title: "Symbol Graph Run", taskID: taskId))
+      )
+
+      // emit module and symbol graph
+      let args =
```
ahoppen: If I understand correctly, we are always emitting a symbol graph for the entire module here. This means that if indexing batches a module with files `a.swift`, `b.swift`, `c.swift` as `[a.swift, b.swift], [c.swift]`, we would be invoking symbol graph generation for the same module twice. What’s worse is that these can run concurrently and potentially interfere with their writes to disk. 

Furthermore, this means that an edit to a file in a module with, say, 100 files, means that we need to re-generate the symbol graph for the entire module, which could take a while (happy to be proven otherwise based on some real-world projects). During background indexing, we have `-index-file-path` to only index a subset of the files listed in the compiler invocation. Really, symbol graph generations should have the same.

The best solution I can think of right now is to make `UpdateSymbolGraphTaskDescription` operate on targets instead of source files (because it generates a symbol graph for all files in the target at once), but then that’s going to be incompatible with symbol graph generation for clang-based files, for which we can generate a symbol graph on a per-file basis.

But overall, this whole-module approach is giving me extremely uneasy feelings about this entire endeavor. Files get saved very frequently and having to re-generate a `.swiftmodule` for the file’s entire module is going to be a very significant cost.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+        [swiftc.path] + baseArgs + [
+          "-emit-module",
+          "-emit-module-path", moduleOutputPath,
+          "-Xfrontend", "-emit-symbol-graph",
+          "-Xfrontend", "-emit-symbol-graph-dir", "-Xfrontend", symbolGraphDir.pathString,
+          "-Xfrontend", "-experimental-skip-all-function-bodies",
```
ahoppen: Does this succeed in case a file in the current module contains a compilation error? If so, you could try adding `-experimental-allow-module-with-compiler-errors`.

ahoppen: Other flags that may be worth investigating are: `-experimental-lazy-typecheck`, `-experimental-skip-non-exportable-decls`. We set those during background preparation and they may apply here as well.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+    self.filesToIndex = filesToIndex
+    self.target = target
+    self.language = language
+    self.buildServerManager = buildServerManager
+    self.logMessageToIndexLog = logMessageToIndexLog
+    self.timeout = timeout
+  }
+
+  package func execute() async {
+    await withLoggingSubsystemAndScope(subsystem: indexLoggingSubsystem, scope: "update-symbolgraph-\(id % 100)") {
```
ahoppen: Similar to `UpdateIndexStoreTaskDescription`’s `indexStoreUpToDateTracker`, we should have something equivalent here.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+      guard language == .swift else { return }
```
ahoppen: Do you have plans to support c-based languages as well?

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
```diff
@@ -0,0 +1,270 @@
...
+        "Updating symbol graph for target: \(target.uri.stringValue)",
+        .info,
+        .begin(StructuredLogBegin(title: "Symbol Graph Run", taskID: taskId))
+      )
+
+      // emit module and symbol graph
+      let args =
+        [swiftc.path] + baseArgs + [
+          "-emit-module",
+          "-emit-module-path", moduleOutputPath,
```
ahoppen: Do we ever need these `.swiftmodule`s? If not, we should be able to delete them straight away (or stream them to `/dev/null`, but really, we shouldn’t be generating them at all in the compiler.

ahoppen (commented):
The more I think about this, the more I come to the conclusion that to support background symbol graph generation, we’d need to re-work how all of this is done in the compiler. 
1. Symbol graph generation should be possible after type checking, same as background indexing and not rely on module generation
2. Symbol graph generation should be possible on a per-file basis instead of requiring the entire module to be compiled
3. And once the two items above are done, it’s likely easy to make it compatible with `-index-file` and share one compiler invocation.

But this is likely going to be a huge endeavor on its own and I have no idea what the scope is, or if it’s even possible. Sorry that I can’t be more helpful here. 

Maybe @hamishknight and @rintaro have thoughts.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
matthewbastien: Only Swift documentation is planned for this GSoC project right now.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: We would require updated `.swiftmodule`s as Symbol graph generation relies on them. Only if the `.swiftmodule`s are updated the `symbol-gaphs` will be updated.

snprajwal:
> The more I think about this, the more I come to the conclusion that to support background symbol graph generation, we’d need to re-work how all of this is done in the compiler.
> 
>   1. Symbol graph generation should be possible after type checking, same as background indexing and not rely on module generation

The symbol graph extractor is designed to work with a `ModuleDecl` input from any source, so it's reasonable to wire up a path that allows symbol graph generation after typechecking. It's mostly straightforward to do AFAIU. As mentioned above, this may potentially be out of scope for this project.

>   2. Symbol graph generation should be possible on a per-file basis instead of requiring the entire module to be compiled
>   3. And once the two items above are done, it’s likely easy to make it compatible with -index-file and share one compiler invocation.

This is significantly harder, and I'm not sure if it's feasible. There are cross-file relationships that depend on the full module being compiled:

- Protocol conformances
- Extensions for types in other files
- Synthesised symbols
- Default implementation members

If the generation is done on a per-file basis, the information available would be incomplete, and the resulting symbol graph would also diverge from the contents of the module-level symbol graph. Since the relationships are cross-file, editing a file would require re-processing the file's dependents to correctly update the state of the graph.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: Thank you for pointing it out! Because `fallbackAfterTimeout` was `true`, when the build server didn't respond in time, it would fall back to locally-synthesized settings that could still carry `vscode-local:`-scheme paths, because of which the transformation was necessary. Now that `fallbackAfterTimeout` is set to `false` this transformation is not necessary. I've removed it!

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: Yes, symbol graph generation works without `-wmo` flag, but the generated symbol graph will be incorrect as the  cross-module symbol references (e.g. `String`, `Int`, `Void`, `Escapable`/`Copyable`) resolve to broken/incomplete precise identifiers .`-wmo` type-checks the entire module in a shared context rather than processing each file independently, which is necessary for these cross-file references to be resolved correctly.

snprajwal:
I looked into SymbolGraphGen and added support for emitting symbol graphs with `-typecheck` here: https://github.com/swiftlang/swift/pull/90799

> Some numbers comparing this with the existing symbol graph generation after serialisation (ran with [hyperfine](https://github.com/sharkdp/hyperfine), mean of 20 runs with `-Onone`):

  | Project               | Module       | Files | Serialised | Typecheck | Speedup      |
  |-----------------------|--------------|------:|----------------------:|---------------------:|-------------:|
  | sourcekit-lsp         | SourceKitLSP |    31 |       1.831 ± 0.037 s |        0.837 ± 0.027 s |  2.19 ± 0.08x |
  | swift-syntax          | SwiftParser  |    46 |       3.517 ± 0.025 s |        1.557 ± 0.022 s |  2.26 ± 0.04x |
  | swift-package-manager | PackageModel |    59 |       2.850 ± 0.029 s |        1.188 ± 0.054 s |  2.40 ± 0.11x |
  | swift-build           | SWBUtil      |   100 |       5.259 ± 0.092 s |        2.206 ± 0.068 s |  2.38 ± 0.08x |

This should help here in terms of the performance overhead of using symbol graphs. It does not alleviate our concerns about needing to typecheck the whole module instead of a single file, but it can make it somewhat less burdensome.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: There was previously a conflict between the `SWIFT_DRIVER_*` environment variables and the explicit compiler flags this task passes, for instance, `SWIFT_DRIVER_SUPPLEMENTARY_OUTPUT_FILE_MAP` conflicting with `-emit-module-path`, since both specified the module output location.  These conflicts no longer occur, so I've removed the environment cleanup that was working around it.

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: Yes,I've added the flag

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: Yes I have added `symbolGraphUpToDateTracker`!

Sources/SemanticIndex/UpdateSymbolGraphTaskDescription.swift
Padmashree06: **These trials were performed on SourceKit-LSP**

| Trial | Time taken for SymbolGraph generation (ms) | Time taken for Indexing (ms) | Overhead percentage (%) | Performance cost |
|:---:|:---:|:---:|:---:|:---:|
| 1 | 159,494 | 63,872.3 | 249.70 | 2.5× |
| 2 | 171,342 | 82,867.8 | 206.76 | 2× |
| 3 | 352,348 | 146,415 | 240.65 | 2.4× |

**Average overhead percentage =** 232.37% 
**Average Performance cost =** 2.3x

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -34,9 +39,12 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
     }
   }
 
+  let workspace: Workspace
```
snprajwal: `Workspace` stores `DocumentationLanguageService`, so storing `Workspace` here creates a reference cycle that leaks memory. We can instead fetch the workspace per request, the way `DoccDocumentationHandler` does it:

https://github.com/swiftlang/sourcekit-lsp/blob/4c1504af2af55fef3a6f188741d3eafc8faa9734/Sources/DocumentationLanguageService/DoccDocumentationHandler.swift#L28-L30

This also means that we use the right workspace in cases where multiple projects are open in the same session.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +109,295 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      return nil
+    }
+
+    guard let sourceKitLSPServer else {
+      return nil
+    }
+
+    guard let index = await workspace.index(checkedFor: .deletedFiles) else {
+      return nil
+    }
```
snprajwal, line 145: It would be good to use the logger to provide debug logs for all these cases where we return nil.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -794,6 +794,25 @@ extension SwiftLanguageService {
   // MARK: - Language features
 
   package func definition(_ request: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+
+    if let snapshot = try? await latestSnapshot(for: request.textDocument.uri) {
+      let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
+      let absolutePosition = snapshot.absolutePosition(of: request.position)
+
+      if let token = tree.token(at: absolutePosition) {
+        let isInDocComment = token.leadingTrivia.pieces.contains {
```
snprajwal: This checks if the token has doc comments, but do we also need to check whether the cursor is inside the doc comment?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -794,6 +794,25 @@ extension SwiftLanguageService {
   // MARK: - Language features
 
   package func definition(_ request: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+
+    if let snapshot = try? await latestSnapshot(for: request.textDocument.uri) {
```
snprajwal: [`latestSnapshot`](https://github.com/swiftlang/sourcekit-lsp/blob/4c1504af2af55fef3a6f188741d3eafc8faa9734/Sources/SwiftLanguageService/SwiftLanguageService.swift#L319) can throw, but we silently swallow errors here with `try?`.

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -438,6 +438,7 @@ package final actor SemanticIndexManager {
         outOfDateTargets.formUnion(changedTargets)
       }
     }
+
```
snprajwal:
```suggestion
```

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +109,295 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    locator.visit(document)
+
+    guard let symbol = locator.found, !symbol.isEmpty else {
+      return nil
+    }
+    return symbol
+  }
+
+  private enum DocTriviaGroup {
+    case lines(lines: [(text: String, startLine: Int, startColumn: Int)])
```
snprajwal: I see this type in multiple places, it might be better to extract it into a struct and use that instead.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +109,295 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    mutating func visitSymbolLink(_ symbolLink: SymbolLink) {
+      guard
+        found == nil,
+        contains(symbolLink.range),
+        let destination = symbolLink.destination,
+        let range = symbolLink.range
+      else {
+        return
+      }
+      let relativeColumn = target.column - range.lowerBound.column - 3
```
snprajwal: AFAICT, this should be 2, not 3. It might also be better to represent it using the additive form, to make it more obvious that we're computing the offset between the cursor and the first letter of the link:


```suggestion
      // Symbol links are delimited by two backticks (``Foo/bar``),
      // so the destination starts two columns past the start of the link.
      let destinationStartColumn = (range.lowerBound.column + 2)
      let relativeColumn = target.column - destinationStartColumn
```

The cursor can also sit on the backticks, so it would be good to check if `relativeColumn >= 0`.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -10,12 +10,17 @@
 //
 //===----------------------------------------------------------------------===//
 
+import BuildServerIntegration
 import Foundation
-import IndexStoreDB
+@preconcurrency import IndexStoreDB
```
snprajwal: Why are we suppressing concurrency diagnostics here?

snprajwal (commented):
Thanks for updating this! We would need to add tests before this can land.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -392,7 +392,7 @@ package actor SourceKitLSPServer {
       @Sendable @escaping (
         RequestType, Workspace, any LanguageService
       ) async throws ->
-      RequestType.Response
+      RequestType.Response?
```
snprajwal: At first glance, this looks like it has a pretty broad blast radius. I think the convention is to use `ResponseError.requestNotImplemented` to fallthrough, but if we don't want to do that, we can introduce a new error type for this instead. @ahoppen would know more on this front.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes, will modify!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes this is extracted into `DocCCommentLine` struct

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Removed `@preconcurrency`, it wasn't needed here!

Sources/SwiftLanguageService/SwiftLanguageService.swift
Padmashree06: Modified the code to check if the cursor is inside a doc comment

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test

matthewbastien:
@swift-ci please test windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
Padmashree06: Added a separate error called `FallThroughToNextLanguageService` to handle this!

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,187 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SKTestSupport
+import XCTest
+
+final class DocumentationDefinitionTests: SourceKitLSPTestCase {
```
snprajwal: Nit: this differs from the filename


```suggestion
final class DocCDefinitionTests: SourceKitLSPTestCase {
```

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -795,6 +795,32 @@ extension SwiftLanguageService {
   // MARK: - Language features
 
   package func definition(_ request: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+
```
snprajwal, line 819: Superfluous newline

```suggestion
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1975,9 +1996,18 @@ extension SourceKitLSPServer {
...
-      return await orLog("Fallback definition request", level: .info) {
+      do {
         let result = try await languageService.definition(req)
         return result?.adjusted(for: copiedFileMap)
+      } catch let error as ResponseError where error.code == .requestNotImplemented {
+        // Signal to handleRequest's loop to try the next language service
+        throw error
+      } catch is FallThroughToNextLanguageService {
+        // Signal to handleRequest's loop to try the next language service
+        throw FallThroughToNextLanguageService()
```
snprajwal: We can forward the error instead of constructing a new instance here:


```suggestion
      } catch let error as FallThroughToNextLanguageService {
        // Signal to handleRequest's loop to try the next language service
        throw error
```

snprajwal (commented):
Thanks for making the changes! Just some minor nits, mostly looking good (pending CI) :)

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,414 @@
...
+  @Test
+  func gotoDefinitionDisambiguatesOverloads() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      public struct Foo {
+        public func bar(x: Int) {}
+        public func 1️⃣bar() {}
+      }
+
+      /// See ``Foo/2️⃣bar()``
```
snprajwal: This does not test the overload disambiguation, since the method parameters vary, so DocC's symbol links would vary by name (`bar(x:)` vs `bar()`). You can instead use different types on the parameters to test this case:

```suggestion
        public func bar(x: Int) {}
        public func 1️⃣bar(x: String) {}
      }

      /// See ``Foo/2️⃣bar(x:)-(String)``
```

snprajwal (approved):
LGTM, just one comment :)

snprajwal:
@swift-ci test

snprajwal:
@ahoppen this is now ready to review

Sources/DocumentationLanguageService/DocCSymbolInformation.swift
```diff
@@ -38,10 +38,11 @@ struct DocCSymbolInformation {
   }
 
   func matches(_ link: DocCSymbolLink) -> Bool {
-    guard link.components.count == components.count else {
+    guard link.components.count <= components.count else {
```
ahoppen, line 50: Why did you need to change this?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+  private func extractSymbolFromDocComment(snapshot: DocumentSnapshot, at position: Position) -> String? {
+    let sourceFile = SwiftParser.Parser.parse(source: snapshot.text)
+    let absolutePosition = snapshot.absolutePosition(of: position)
+
+    guard let token = sourceFile.token(at: absolutePosition) else {
+      return nil
+    }
+
+    let cursorPosition = snapshot.positionOf(utf8Offset: absolutePosition.utf8Offset)
+    let groups = docCommentGroups(in: token.leadingTrivia, tokenStart: token.position, snapshot: snapshot)
```
ahoppen: A doc comment could theoretically also be in trailing trivia, eg. 

```swift
func foo() { /** doc comment */

}
```

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
   ) async {
     // The DocumentationLanguageService does not do anything with document events
   }
+
+  package func definition(_ req: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+    let snapshot = try self.documentManager.latestSnapshot(req.textDocument.uri)
+    let clickedSymbol: String?
```
ahoppen: LSP has no notion of clicking — how about `cursorSymbol`? The log message eleven lines down already says “at the cursor position”.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      logger.debug("No workspace found for document \(req.textDocument.uri)")
+      return nil
+    }
+
+    guard let index = await workspace.index(checkedFor: .deletedFiles) else {
+      logger.debug("No index available for workspace")
+      return nil
+    }
+
+    let occurrence = try await sourceKitLSPServer.withOnDiskDocumentManager { onDiskDocumentManager in
```
ahoppen: Since `DoccDocumentationHandler.markdownDocumentation` already has this exact `withOnDiskDocumentManager` + `primaryDefinitionOrDeclarationOccurrence(ofDocCSymbolLink:fetchSymbolGraph:)` + `symbolGraph(forOnDiskContentsAt:in:manager:)` block, can we pull it into a shared helper that both call instead of duplicating it?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    }
+
+    guard let targetLocation = occurrence?.location.lspLocation else {
+      logger.debug("No definition occurrence found in the index for symbol link '\(clickedSymbol)'")
+      return nil
+    }
+    return .locations([targetLocation])
+  }
+
+  /// Walks the Markdown/DocC AST looking for symbol link
```
ahoppen: This says it finds the symbol link that contains the position, but `visitSymbolLink` returns only the *prefix* of the components up to the one the cursor is on (`components[0...index].joined(...)`). Could you make that explicit? As written I would expect `found` to be the whole destination.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+  }
+
+  private struct DocCCommentLine {
+    let text: String
+    let startLine: Int
+    let startColumn: Int
+  }
+
+  private enum DocTriviaGroup {
+    case lines(lines: [DocCCommentLine])
```
ahoppen: The label just restates the case name. Can we drop it and write `case lines([DocCCommentLine])`? The pattern match at the call site is already `case .lines(let lines)`.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    case block(text: String, startLine: Int, startColumn: Int)
+  }
+
+  /// Converts a UTF-16 offset within `line` into the corresponding UTF-8 byte offset.
+  /// Needed because DocumentSnapshot/LSP positions are UTF-16-based but
+  /// Markdown.SourceLocation.column is UTF-8-byte-based.
+  private func utf8Offset(inLine line: String, forUTF16Offset utf16Offset: Int) -> Int {
```
ahoppen: `DocumentSnapshot` already has this: `snapshot.lineTable.utf8ColumnAt(line:utf16Column:)` is a drop-in replacement and it is faster because the line table is already pre-populated. That would let `extractSymbolFromText` take the snapshot and drop both this helper and the `components(separatedBy: "\n")` split.

Could you also add a test with a complex emoji like 🧑‍🧑‍🧒‍🧒 before the cursor? None of the new tests contain anything outside ASCII, so this conversion is not actually being exercised.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+        newlinesSinceLastDoc = 0
+      }
+    }
+    flushPendingLines()
+    return groups
+  }
+
+  private struct StrippedLine {
+    let text: String
+    let strippedPrefixCount: Int
```
ahoppen, line 313: Is this a UTF-16 or a UTF-8 count? It is accumulated from `.utf16.count` differences in both strip functions, but `resolveTarget` subtracts it from `utf8Rel`, which is a UTF-8 byte offset. That works today only because `///`, `/**`, `*/`, `*` and whitespace are all single-byte. Could you document the unit so this does not silently break later?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+        stripped += 3
+      }
+      let beforeIndent = text.utf16.count
+      let trimmed = text.drop { $0 == " " || $0 == "\t" }
+      stripped += beforeIndent - trimmed.utf16.count
+      text = trimmed
+      return StrippedLine(text: String(text), strippedPrefixCount: stripped)
+    }
+  }
+  /// Strips `/**`, `*/`, and per-line leading `*`/whitespace from a block comment.
```
ahoppen, line 336: Missing blank line above this doc comment.

Sources/SourceKitLSP/LanguageService.swift
```diff
@@ -99,6 +99,13 @@ package struct GeneratedInterfaceDetails: ResponseType, Hashable {
   }
 }
 
+/// Thrown by a language service's request handler to indicate that this
+/// service isn't the right one for the current request context, and the
+/// next language service should be tried instead.
+package struct FallThroughToNextLanguageService: Error {
```
ahoppen: Could you also mention this on the `definition` requirement of the `LanguageService` protocol (it currently only carries `- Note: Only called as a fallback if the definition could not be found in the index.`)? That is the place someone writing a new language service would look, and it says nothing about `FallThroughToNextLanguageService` being a valid way to opt out — or about intermediate request handlers having to rethrow it rather than swallow it.

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -795,6 +795,31 @@ extension SwiftLanguageService {
   // MARK: - Language features
 
   package func definition(_ request: DefinitionRequest) async throws -> LocationsOrLocationLinksResponse? {
+    let snapshot = try await latestSnapshot(for: request.textDocument.uri)
+    let tree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let absolutePosition = snapshot.absolutePosition(of: request.position)
+
+    if let token = tree.token(at: absolutePosition) {
```
ahoppen: This walks the leading trivia only to decide whether we are in a doc comment, and then `DocumentationLanguageService.extractSymbolFromDocComment` re-parses the entire file from scratch with `SwiftParser.Parser.parse` — rather than the cached tree we already have here — and walks the trivia again to do the real work. Could this be done once? We already have the token and the cached tree at this point.

Does not have to be this PR, but it seems worth a follow-up at least.

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,414 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SKTestSupport
+import Testing
+
+@Suite(.serialized)
```
ahoppen: Two things: should this be `@Suite(.serialized, .configureLogging)`, like `SwiftPMBuildServerTests`? Without the trait the logging in these tests does not get a configured subsystem.

And do they actually need to be serialized? Each test builds its own scratch project with no shared state that I can see, and serializing sixteen tests — several of which build a full SwiftPM package with background indexing — is a real wall-clock cost.

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,414 @@
...
+      """
+    )
+
+    let response = try await project.testClient.send(
+      DefinitionRequest(textDocument: TextDocumentIdentifier(project.fileURI), position: project.positions["1️⃣"])
+    )
+    #expect(response == nil)
+  }
+
+  // MARK: DocC catalog files (tutorials and makdown)
```
ahoppen: Typo: “makdown” → “markdown”.

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,414 @@
...
+
+  @Test
+  func gotoDefinitionInDocCTutorialFile() async throws {
```
ahoppen: This file uses three prefixes for the same operation: `jumpToDefinition*`, `gotoDefinition*` and `definitionIsNil*`. `DefinitionTests.swift` already established `jumpToDefinition` for this, so let us stick to one throughout.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      let relativeUTF16 = cursorPosition.utf16index - columnBase
+      let utf8Rel = utf8Offset(inLine: rawLine, forUTF16Offset: relativeUTF16)
+      let targetColumn = max(1, utf8Rel - strippedLines[lineIndex].strippedPrefixCount + 1)
+
+      return (strippedLines, lineIndex, targetColumn)
+    }
+  }
+
+  private func extractSymbolFromDocComment(snapshot: DocumentSnapshot, at position: Position) -> String? {
+    let sourceFile = SwiftParser.Parser.parse(source: snapshot.text)
```
ahoppen, line 392: Can we use `SyntaxTreeManager` here to avoid re-parsing the source file?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    let absolutePosition = snapshot.absolutePosition(of: position)
+
+    guard let token = sourceFile.token(at: absolutePosition) else {
+      return nil
+    }
+
+    let cursorPosition = snapshot.positionOf(utf8Offset: absolutePosition.utf8Offset)
```
ahoppen: Isn’t this exactly the same as `position`?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    guard let symbol = locator.found, !symbol.isEmpty else {
+      return nil
+    }
+    return symbol
+  }
+
+  private struct DocCCommentLine {
+    let text: String
+    let startLine: Int
+    let startColumn: Int
```
ahoppen: Can’t this be `Position`?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    guard
+      let utf16Index = utf16View.index(utf16View.startIndex, offsetBy: utf16Offset, limitedBy: utf16View.endIndex),
+      let stringIndex = utf16Index.samePosition(in: line)
+    else {
+      return line.utf8.count
+    }
+    return line.utf8.distance(from: line.startIndex, to: stringIndex)
+  }
+
+  private func docCommentGroups(
```
ahoppen, line 266: Could you explain what exactly a doc comment group is in a comment?

ahoppen: Why do we need to build up these groups at all? Don’t we just need to inspect the doc comment to find the link that the user clicked on? Why do we care how the comments may be grouped?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      stripped += beforeIndent - trimmed.utf16.count
+      line = trimmed
+      return StrippedLine(text: String(line), strippedPrefixCount: stripped)
+    }
+  }
+
+  private func resolveTarget(
+    for group: DocTriviaGroup,
+    cursorPosition: Position
+  ) -> (strippedLines: [StrippedLine], lineIndex: Int, targetColumn: Int)? {
```
ahoppen: Why return a line + column here instead of `Position`?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    var offset = tokenStart.utf8Offset
+    var newlinesSinceLastDoc = 0
+
+    func flushPendingLines() {
+      guard !pendingLines.isEmpty else { return }
+      groups.append(.lines(lines: pendingLines))
+      pendingLines.removeAll()
+    }
+
+    for piece in trivia.pieces {
```
ahoppen, line 282: This seems to be replicating logic from `Trivia.docCommentValue` in swift-syntax. It would be great if there was a way to unify this logic, eg. by having an `@_spi` function in swift-syntax that both `Trivia.docCommentValue` and this implementation can call into.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+    return lines.map { line in
+      var text = Substring(line.text)
+      var stripped = 0
+
+      if text.hasPrefix("///") {
+        text.removeFirst(3)
+        stripped += 3
+      }
+      let beforeIndent = text.utf16.count
+      let trimmed = text.drop { $0 == " " || $0 == "\t" }
```
ahoppen, line 329: This is not correct. It will for example flatten nested lists like

```swift
/// - top bullet
///   - nested
```

A lot of thought like this went into `Trivia.docCommentValue` (see comments on https://github.com/swiftlang/swift-syntax/pull/3230 and the preceding PRs). That’s why I would really like to re-use the logic implemented there.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
```diff
@@ -100,4 +105,318 @@ package actor DocumentationLanguageService: LanguageService, Sendable {
...
+      let relativeColumn = target.column - destinationStartColumn
+      guard relativeColumn >= 0 else {
+        return
+      }
+      let components = destination.split(separator: "/")
+      var currentLength = 0
+
+      for (index, component) in components.enumerated() {
+        let end = currentLength + component.utf8.count
+        if relativeColumn < end {
```
ahoppen, line 204: Why is this comparison needed? Don’t you always us the entire link’s text as the destination?

Sources/SwiftLanguageService/SwiftLanguageService.swift
```diff
@@ -795,6 +795,31 @@ extension SwiftLanguageService {
...
+      var offset = token.position.utf8Offset
+      var isInDocComment = false
+
+      for piece in token.leadingTrivia.pieces {
```
ahoppen: We should also check trailingTrivia here. Would probably also be worth pulling this out into a helper function to get the trivia piece at a given function that may be called by other parts of the codebase in the future, if we need this again.

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
```diff
@@ -0,0 +1,414 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SKTestSupport
+import Testing
+
+@Suite(.serialized)
+struct DocCDefinitionTests {
```
ahoppen: Please use XCTests for tests. See https://github.com/swiftlang/sourcekit-lsp/pull/2467#discussion_r2743140416

ahoppen (commented):
An initial review pass. My biggest point is that we should share the logic of extracting the doc comment value with swift-syntax’s `Trivia.docCommentValue` because it’s surprisingly subtle to do so. I also have a gut feeling that the position translations might not need to be this complicated but I haven’t verified it in detail – a lot of this may come out of sharing implementation with swift-syntax.

Sources/DocumentationLanguageService/DocCSymbolInformation.swift
Padmashree06: The `components` always includes the full resolved path including the module (e.g. [`"ModuleA", "Sloth"`]), where as a doc-comment link like `Sloth` only has one component. So even an unambiguous, correctly-named match fails the length check before the name/disambiguation comparison ever runs.
Using `<=` instead of `==` allows the check to proceed when the doc-comment link has fewer components than the full resolved path, so valid unambiguous matches are not rejected by the length check.

Tests/SourceKitLSPTests/DocCDefinitionTests.swift
Padmashree06: Yes, changed it to `jumpToDefinition`

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Addressed. That's now fixed: the subtraction happens against the UTF-16 offset in `resolveTarget`, and the UTF-8 conversion (for `Markdown.SourceLocation`) happens afterward, in `extractSymbolFromDocComment`. The `StrippedLine` struct is also documented!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Changed it to `Position`

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yup! Dropped it!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: - Each doc comment group is a  set of continuous single line comments  or a block comment. 
- Each doc comment group is parsed as a single markdown file 

```swift
/// Comment Line 1  - group1 
/// Comment Line 2 

/// Comment  - group2 

/** 

Block Comment  - group3

*/
```



Ex: When there is a symbol link is between two comments 
```swift
\\\ This is a ``Symbol/
\\\ Link``
```

and also for cases like 
```swift
\\\ ```swift
\\\ ``Symbol``
\\\  ```
```
This was added while addressing this comment https://github.com/swiftlang/sourcekit-lsp/pull/2698#discussion_r3535492863

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: `SyntaxTreeManager` is `internal` scoped can we convert it, specifically `syntaxTree(for snapshot: )` to `package` scope so that we can use it here ?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: I would really like to reuse the `Trivia.docCommentValue` but the only problem is that it doesn't track the position after stripping the comments which is important here as we need to retain the cursor position  even after stripping.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: No, we do not consider the entire link's text as the destination when we have nested links. 
Ex: When we have ` /// ``Foo/bar/link`` `  and the user clicks on `bar` the destination is resolved to `Foo/bar`. 

CC: @matthewbastien

Sources/DocumentationLanguageService/DocCSymbolInformation.swift
ahoppen: I’m a little worried that we relax this too much for the other callers of this function, which may now incorrectly match multiple results with the new relaxed check. Could you instead add a new function here that explicitly checks for a suffix?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
ahoppen: Could we add that tracking fairly cheaply to swift-syntax?

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
ahoppen: Oh, interesting. That’s pretty surprising behavior to me because it doesn’t match how other links work. If I have a link to `https://www.swift.org/contributing/` and click on `swift.org`, I still expect to be taken to `https://www.swift.org/contributing/`. 

Is this really the behavior we want?

matthewbastien: I had originally suggested to go this route because VS Code was highlighting only a segment of the full symbol link. For example, if the full symbol link was `Foo/bar()` then hovering would only highlight either `Foo` or `bar()` depending on where the cursor was positioned. I think it'd be equally confusing if VS Code highlighted only `Foo`, but then actually navigated to `Foo/bar()` on click.

If we can find a way to change that behaviour and make it highlight the whole link then we can just do go to definition on the full link.

snprajwal: Thinking from the perspective of a documentation author, it's reasonable to assume that one would want to navigate to the container of the documented type in a qualified symbol link. Personally, I also think it's unintuitive if I click on `Foo` in `Foo/bar` and it jumped to `bar` instead. A counterpart from Swift that I can think of is static method calls -- if you have some `Foo.staticMethod()`, go-to-definition with the cursor on `Foo` would go to the type rather than the method. I would think the same logic applies here, just with slashes instead of dots as separators.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: Yes!

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
ahoppen: Ah, I see. I think that’s a fair argument. Thanks for clarifying.

Sources/DocumentationLanguageService/DocumentationLanguageService.swift
Padmashree06: I've added a PR https://github.com/swiftlang/swift-syntax/pull/3430 addressing this review!

Sources/DocumentationLanguageService/DocCSymbolInformation.swift
Padmashree06: Yes! I've added a new function `matchesAsSuffix`

Sources/DocumentationLanguageService/DocCSymbolInformation.swift
```diff
@@ -45,6 +45,16 @@ struct DocCSymbolInformation {
       linkComponent.name == symbolComponent.name && symbolComponent.information.matches(linkComponent.disambiguation)
     }
   }
+
+  func matchesAsSuffix(_ link: DocCSymbolLink) -> Bool {
```
ahoppen: From this API name, it’s not really clear whether `self` or `link` is the suffix. How about changing it to something like the following?

```suggestion
  func matches(suffix: DocCSymbolLink) -> Bool {
```

PR #2699, opened 2026-06-22
Release in-flight QueuedTasks when TaskScheduler.shutDown returns

rintaro:
`TaskScheduler.shutDown()` cancelled and waited for in-flight tasks but never emptied `currentlyExecutingTasks` / `pendingTasks`. That left a retain cycle through the scheduler's own queues:

```
  TaskScheduler.{currentlyExecuting,pending}Tasks
    → QueuedTask
      → executionStateChangedCallback (captures self of SemanticIndexManager)
        → SemanticIndexManager.indexTaskScheduler
          → TaskScheduler
```

Once the external strong refs went away (`SourceKitLSPServer.indexTaskScheduler` and `Workspace.semanticIndexManagerTask`) the cycle kept both `TaskScheduler` and `SemanticIndexManager` rooted, which transitively kept `BuildServerManager`, the per-task `UpToDateTracker`s, `ToolchainRegistry`, and assorted caches alive.

PR #2700, opened 2026-06-23
Drop new tasks scheduled after TaskScheduler.shutDown

rintaro:
Follow-up to #2699. That PR drained the queues at shutdown but did not stop new work from being scheduled afterwards.

After a preparation task completes (or is cancelled), `SemanticIndexManager.scheduleIndexing` continues and schedules an `UpdateIndexStoreTaskDescription` via `indexTaskScheduler.schedule(...)`. If that `schedule()` call lands after `shutDown()` has cleared the arrays, the new `QueuedTask` goes into `pendingTasks` and leaks.

Make `schedule()` return `nil` when the scheduler is already shut down, and update the two `SemanticIndexManager` call sites to bail out on `nil`.

### DIFF
--- a/Sources/SemanticIndex/TaskScheduler.swift
+++ b/Sources/SemanticIndex/TaskScheduler.swift
@@ -425,6 +425,9 @@ package actor TaskScheduler<TaskDescription: TaskDescriptionProtocol> {
 
   /// Enqueue a new task to be executed.
   ///
+  /// Returns `nil` if the scheduler has already been shut down via `shutDown()`; the task is
+  /// not queued in that case.
+  ///
   /// - Important: A task that is scheduled by `TaskScheduler` must never be awaited from a task that runs on
   ///   `TaskScheduler`. Otherwise we might end up in deadlocks, eg. if the inner task cannot be scheduled because the
   ///   outer task is claiming all execution slots in the `TaskScheduler`.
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

Sources/SemanticIndex/TaskScheduler.swift
```diff
@@ -435,7 +435,12 @@ package actor TaskScheduler<TaskDescription: TaskDescriptionProtocol> {
     @_inheritActorContext executionStateChangedCallback: (
       @Sendable (QueuedTask<TaskDescription>, TaskExecutionState) async -> Void
     )? = nil
-  ) async -> QueuedTask<TaskDescription> {
+  ) async -> QueuedTask<TaskDescription>? {
```
ahoppen, line 441: Could you add a doc comment indicating what a `nil` return value signifies?

ahoppen (approved):

hamishknight (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

PR #2702, opened 2026-06-24
Replace version-0 sentinel on DocumentSnapshot with explicit Origin

rintaro:
`SwiftLanguageService.syntacticTestItems` used `snapshot.version == 0` as a "loaded from disk" signal to bypass the shared `syntaxTreeManager` cache. The LSP spec doesn't constrain the initial version (VS Code uses 1, 0 is legal too), so a client picking 0 would silently miss the cache for every test discovery scan.

Replace it with a required `DocumentSnapshot.Origin` (`openDocument` / `fromDisk` / `generated`). `DocumentManager` is the only `.openDocument` source; the other creators declare `.fromDisk` or `.generated`. `syntacticTestItems` branches on `origin == .openDocument`, which correctly covers both non-tracked cases.

### DIFF
--- a/Sources/SourceKitLSP/DocumentManager.swift
+++ b/Sources/SourceKitLSP/DocumentManager.swift
@@ -26,6 +26,20 @@ import Synchronization
 /// ``Document``. The purpose of a ``DocumentSnapshot`` is to be able to work
 /// with one version of a document without having to think about it changing.
 package struct DocumentSnapshot: Identifiable, Sendable {
+  /// How a `DocumentSnapshot` was produced. Determines whether the `version` is a real LSP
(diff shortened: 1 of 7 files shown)


hamishknight (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test macOS

PR #2703, opened 2026-06-26
Release SemanticIndexManager state on workspace shutdown

rintaro:
Release `SemanticIndexManager` (and the `TaskScheduler` / `BuildServerManager` / `ToolchainRegistry` chain it transitively retains) when its workspace tears down. Two related retain cycles kept it alive past shutdown, surfacing as leaks in tests that exit while indexing is paused or in-flight (e.g. `testPausingBackgroundIndexingDoesNotStopPreparation`). 
  
  - Add `SemanticIndexManager.shutDown` that cancels every Task in `inProgressIndexTasks`, `inProgressPreparationTasks`, `inProgressPrepareForEditorTask`, and `fileOrBuildTargetChangedTasks`, and drops those dictionaries. Without this, a `.low`-priority `UpdateIndexStoreTaskDescription` that sat in `pendingTasks` and was cancelled by `TaskScheduler.shutDown` without ever executing left its `inProgressIndexTasks` entry behind — the entry's `indexTask` captures `self`, forming a self-cycle. The scheduler's per-task `.finished` callback normally clears the entry, but it's only fired from `QueuedTask.execute` / `finalizeExecution`. Call it from `Workspace.shutdown`.
  
  - Gate `TaskScheduler.finalizeTaskExecution`'s re-add of `.cancelledToBeRescheduled` tasks on `!isShutDown`. After `shutDown` clears `pendingTasks`, a racing `cancelToBeRescheduled` (from a `setMaxConcurrentTasksByPriority`-driven `Task.sleep` that elapses mid-shutdown) would otherwise re-populate the array. The re-added entry never drains — `poke` early-returns on `isShutDown` — and keeps the `QueuedTask` alive.

### DIFF
--- a/Sources/SemanticIndex/SemanticIndexManager.swift
+++ b/Sources/SemanticIndex/SemanticIndexManager.swift
@@ -295,6 +295,37 @@ package final actor SemanticIndexManager {
     self.indexProgressStatusDidChange = indexProgressStatusDidChange
   }
 
+  /// Cancels any in-progress preparation/index Tasks and drops references to them so that
+  /// `SemanticIndexManager` (and the `TaskScheduler` it transitively retains) can be deallocated
+  /// once the workspace tears down.
+  ///
+  /// The `Task<Void, Never>` values stored in `inProgressIndexTasks` /
+  /// `fileOrBuildTargetChangedTasks` capture `self`, forming a self-cycle through these
+  /// dictionaries. The cycle is normally broken by the scheduler's per-task `.finished` callback
+  /// clearing the entry — but that callback is only invoked from `QueuedTask.execute` /
+  /// `finalizeExecution`, so a task that was paused in `pendingTasks` and cancelled by
+  /// `TaskScheduler.shutDown` without ever executing leaves its entry behind.
+  package func shutDown() {
+    for inProgress in inProgressIndexTasks.values {
+      switch inProgress.state {
+      case .creatingIndexTask:
+        break
+      case .waitingForPreparation(_, let indexTask),
+        .preparing(_, let indexTask),
+        .updatingIndexStore(_, let indexTask):
+        indexTask.cancel()
+      }
+    }
+    inProgressIndexTasks = [:]
+    inProgressPreparationTasks = [:]
+    inProgressPrepareForEditorTask?.task.cancel()
+    inProgressPrepareForEditorTask = nil
+    for task in fileOrBuildTargetChangedTasks.values {
+      task.cancel()
+    }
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -295,6 +295,37 @@ package final actor SemanticIndexManager {
...
+      case .waitingForPreparation(_, let indexTask),
+        .preparing(_, let indexTask),
+        .updatingIndexStore(_, let indexTask):
+        indexTask.cancel()
+      }
+    }
+    inProgressIndexTasks = [:]
+    inProgressPreparationTasks = [:]
+    inProgressPrepareForEditorTask?.task.cancel()
+    inProgressPrepareForEditorTask = nil
```
hamishknight, line 322: Could we run into a similar race where a new task could be scheduled after we perform shutdown?

Sources/SemanticIndex/SemanticIndexManager.swift
```diff
@@ -295,6 +295,37 @@ package final actor SemanticIndexManager {
...
+      case .creatingIndexTask:
+        break
+      case .waitingForPreparation(_, let indexTask),
+        .preparing(_, let indexTask),
+        .updatingIndexStore(_, let indexTask):
+        indexTask.cancel()
+      }
+    }
+    inProgressIndexTasks = [:]
+    inProgressPreparationTasks = [:]
```
hamishknight, line 320: Should we cancel these too?

hamishknight (approved):

Sources/SemanticIndex/SemanticIndexManager.swift
rintaro: `inProgressPreparationTasks` values are `OpaqueQueuedIndexTask` owned by the `TaskScheduler`, `TaskScheduler.shutDown()` should take care of them.

Sources/SemanticIndex/SemanticIndexManager.swift
rintaro: Good point. So we need `SemanticIndexManager.isShutDown` flag too, and block things in several places.
I'm going to fix in a follow-up.

rintaro: I think most of the possible cases are triggered by an LSP request/notification, which is okay because `SourceKitLSPServer` removes the `Workspace` after `shutDown()`.
But still, `buildTargetsChanged` can be called triggered by `BSP` notification.

PR #2704, opened 2026-07-03
Add missing import code action

GerardSerra18:
## Summary

Adds a new Swift code action that offers to insert a missing import when SourceKit reports an unresolved symbol diagnostic and the semantic index can resolve that symbol to a single module.

## Details

This change adds a new `SwiftLanguageService` quick fix provider that:

- queries SourceKit diagnostics for missing-symbol diagnostic IDs
- extracts the unresolved symbol name from the diagnostic message
- looks up canonical symbol occurrences in the semantic index
- suggests an import only when the symbol resolves to exactly one module
- avoids suggesting modules that are already imported
- deduplicates repeated suggestions for the same module

The provider is registered alongside the existing quick-fix providers and included in the SwiftLanguageService CMake target.

## Testing

```bash
swift build
SKIP_LONG_TESTS=1 SOURCEKIT_LSP_FORCE_NON_DARWIN_LOGGER=1 swift test --filter 'SourceKitLSPTests.CodeActionTests/testAddMissingImport'
SKIP_LONG_TESTS=1 SOURCEKIT_LSP_FORCE_NON_DARWIN_LOGGER=1 swift test --filter 'SourceKitLSPTests.CodeActionTests/testRemoveUnusedImports'
SKIP_LONG_TESTS=1 SOURCEKIT_LSP_FORCE_NON_DARWIN_LOGGER=1 swift test --filter 'SourceKitLSPTests.CodeActionTests/'

### CODE REVIEW

**Review by ahoppen — COMMENTED** [review]
Thanks for the PR @GerardSerra18. I think this will be very valuable. I have one high-level comment: Instead of having one code action that adds all missing imports, would it be possible to attach these code actions as quick fixes to the diagnostics that report the missing type? That way the fix is localized to the error that produces them.

**Comment by GerardSerra18** [comment]
Thanks for the review @ahoppen, that makes sense.
I’ll update the implementation so the missing-import actions are attached to the unresolved-symbol diagnostics themselves, and then let the existing `retrieveQuickFixCodeActions` path surface them as localized quick fixes. That should also let me remove the separate code action provider registration and make this fit better with the existing diagnostic quick-fix flow.

**Comment by GerardSerra18** [comment]
@ahoppen Updated.

The missing-import quick fixes are now attached directly to the unresolved-symbol diagnostics before retrieveQuickFixCodeActions maps diagnostic code actions into LSP code actions.
This removes the separate code action provider registration and keeps the action localized to the diagnostic that reports the missing symbol. I also updated the tests to request the quick fix through the diagnostic-based flow.

Thhanks again!

File: Sources/SwiftLanguageService/AddMissingImport.swift [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+    _ diagnosticReport: RelatedFullDocumentDiagnosticReport,
+    for snapshot: DocumentSnapshot,
+    document: DocumentURI
+  ) async throws -> RelatedFullDocumentDiagnosticReport {
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
+
+    var result = diagnosticReport
+
+    for index in result.items.indices {
```
ahoppen: Could we `.map` the items instead of working based on indices?

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+
+private func missingSymbolName(from diagnostic: Diagnostic) -> String? {
+  guard diagnostic.source == "SourceKit" else {
+    return nil
+  }
+
+  for prefix in missingImportDiagnosticMessagePrefixes {
+    guard diagnostic.message.hasPrefix(prefix) else {
```
ahoppen: Instead of doing a string match here, could we do the enrichment of diagnostics in `DiagnosticReportManager.requestReport`. In there, we’d have access to the diagnostic IDs returned by sourcekitd as well. We’d need to pass an index into `DiagnosticReportManager` but that should be fine.

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,161 @@
...
+]
+
+extension SwiftLanguageService {
+  func diagnosticReportWithMissingImportCodeActions(
+    _ diagnosticReport: RelatedFullDocumentDiagnosticReport,
+    for snapshot: DocumentSnapshot,
+    document: DocumentURI
+  ) async throws -> RelatedFullDocumentDiagnosticReport {
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
```
ahoppen: Ideally, we’d only request the syntax tree + walk it for imports if we actually have a diagnostic about a missing symbol.

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,161 @@
...
+  snapshot: DocumentSnapshot
+) -> TextEdit? {
+  if let lastImport = existingImports.last {
+    let insertionPosition = snapshot.position(of: lastImport.endPosition)
+    return TextEdit(
+      range: insertionPosition..<insertionPosition,
+      newText: "\nimport \(moduleName)"
+    )
+  } else if let firstStatement = syntaxTree.statements.first {
+    let insertionPosition = snapshot.position(of: firstStatement.positionAfterSkippingLeadingTrivia)
```
ahoppen: I think if we don’t have imports, we should just insert this at the start of the file. Right now, we’d insert it after the doc comment in

```swift
/// Do an amazing thing
struct Starstruck {}
```

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,161 @@
...
+    let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+    let existingImports = syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
+
+    var result = diagnosticReport
+
+    for index in result.items.indices {
+      guard
+        let missingSymbol = missingSymbolName(from: result.items[index]),
+        let moduleName = try? await self.uniqueModuleProvidingSymbol(named: missingSymbol, for: document),
+        !existingImports.imports(moduleName),
```
ahoppen: What was the motivation for this check? If we get the diagnostic and we have a unique module providing the symbol, we shouldn’t need to check if the import already exists.

ahoppen (commented):
Nice, this is going in the right direction. I left some initial comments inline. Looking forward to having this.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: Done. I moved the enrichment into `DiagnosticReportManager.requestReport`, so we now attach the quick fix while building each diagnostic instead of mutating `diagnosticReport.items` by index.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: Done. The missing-import enrichment now happens in `DiagnosticReportManager.requestReport`, where the raw sourcekitd diagnostics are still available.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: The syntax tree is now only requested if the sourcekitd diagnostics contain a missing-symbol diagnostic, if there are no relevant diagnostics, we skip computing importss.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: Done. If there are no existing imports, the quick fix now inserts the import at the start of the file instead of before the first non-trivia statement. I also added a test that covers the leading doc comment case too

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: Makes sense, I removed that check. If SourceKit is already reporting the missing-symbol diagnostic and we can resolve the symbol to a unique module, that should be enough for offer the fix

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,111 @@
...
+import SourceKitLSP
+import SwiftSyntax
+
+private let missingImportDiagnosticIDs: Set<String> = [
+  "cannot_find_type_in_scope",
+  "cannot_find_in_scope",
+  "use_of_unresolved_identifier",
+]
+
+struct MissingImportCodeActionContext {
```
ahoppen: Do we still need the context struct? I feel like it might be easier to just pass in the index + existing imports.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -149,15 +155,58 @@ actor DiagnosticReportManager {
...
+
+      return diagnosticID
+    }.isEmpty == false
+
+    let missingImportContext: MissingImportCodeActionContext?
+    if hasMissingImportDiagnostic, let uncheckedIndex {
+      let syntaxTree = await syntaxTreeManager.syntaxTree(for: snapshot)
+        missingImportContext = MissingImportCodeActionContext(
+            index: uncheckedIndex.checked(for: .deletedFiles),
+            existingImports: syntaxTree.statements.compactMap { $0.item.as(ImportDeclSyntax.self) }
```
ahoppen: Instead of letting the caller extract the import decls, should this happen inside `missingImportCodeAction` to localize the logic and we just pass in a syntax tree here?

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,111 @@
...
+      moduleNames.insert(moduleName)
+
+      return moduleNames.count <= 1
+    }
+
+    guard moduleNames.count == 1 else {
+      return nil
+    }
+
+    return moduleNames.first
```
ahoppen:
```suggestion
    return moduleNames.only
```

Sources/SwiftLanguageService/AddMissingImport.swift
ahoppen, line 1: Given how intertwined this is with `DiagnosticReportManager` now, I think it would make sense to move these functions over there and then maybe make some of the `private`/`fileprivate`.

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,111 @@
...
+  "cannot_find_in_scope",
+  "use_of_unresolved_identifier",
+]
+
+struct MissingImportCodeActionContext {
+  let index: CheckedIndex
+  let existingImports: [ImportDeclSyntax]
+}
+
+func isMissingImportDiagnosticID(_ diagnosticID: String) -> Bool {
```
ahoppen: Not sure if this function provides a lot of value compared to just inlining the `.contains` check.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1925,6 +1925,189 @@ final class CodeActionTests: SourceKitLSPTestCase {
     }
   }
 
+  func testAddMissingImport() async throws {
+    let project = try await SwiftPMTestProject(
+        files: [
+        "MyLibrary/ImportedType.swift": """
```
ahoppen, line 1931: Could you run `swift-format` on your changes? I think these aren’t formatted correctly.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -1925,6 +1925,189 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+        """,
+        ],
+        manifest: """
+        let package = Package(
+        name: "MissingImport",
+        targets: [
+            .target(name: "MyLibrary"),
+            .executableTarget(name: "MyApp", dependencies: ["MyLibrary"]),
+        ]
+        )
```
ahoppen, line 1947: Also make sure that the package manifest inside the string literal is formatted correctly.

Sources/SwiftLanguageService/AddMissingImport.swift
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  let edit = addImportEdit(
+    moduleName,
+    existingImports: context.existingImports,
+    snapshot: snapshot
+  )
+
+  return CodeAction(
+    title: "Add import \(moduleName)",
```
ahoppen:
```suggestion
    title: "Add import for '\(moduleName)'",
```

ahoppen (commented):
The overall shape looks good now. I’ve got a few more local comments inline.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: You're right. I removed the context struct and now pass the checked index and syntax tree directly.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: I change it,  `missingImportCodeAction` now takes the syntax tree and extracts the existing imports internally, so the import-specific logic stays localized there.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: I applied, thanks!

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: I applied too, thanks!

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: I moved the missing-import helpers into `DiagnosticReportManager.swift` and made them private.

Sources/SwiftLanguageService/AddMissingImport.swift
GerardSerra18: I removed the helper and inlined the `missingImportDiagnosticIDs.contains(...)` check.

Tests/SourceKitLSPTests/CodeActionTests.swift
GerardSerra18: I did it, thanks

Tests/SourceKitLSPTests/CodeActionTests.swift
GerardSerra18: Done. I also fixed the formatting of the package manifest string literals in the tests.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -149,15 +155,60 @@ actor DiagnosticReportManager {
...
+    let missingImportDiagnosticIDsInResponse: [String] =
+      rawDiagnostics?.compactMap { rawDiagnostic -> String? in
+        guard let diagnosticID: String = rawDiagnostic[keys.id] else {
+          return nil
+        }
+
+        return missingImportDiagnosticIDs.contains(diagnosticID) ? diagnosticID : nil
+      } ?? []
+
+    let hasMissingImportDiagnostic = !missingImportDiagnosticIDsInResponse.isEmpty
```
ahoppen:
```suggestion
    let hasMissingImportDiagnostic: [String] =
      rawDiagnostics?.contains { rawDiagnostic -> String? in
        guard let diagnosticID: String = rawDiagnostic[keys.id] else {
          return false
        }

        return missingImportDiagnosticIDs.contains(diagnosticID)
      } ?? false
```

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -201,4 +252,86 @@ actor DiagnosticReportManager {
...
+    "cannot_find_type_in_scope",
+    "cannot_find_in_scope",
+    "use_of_unresolved_identifier",
+  ]
+
+  private func missingSymbolName(from diagnosticDescription: String) -> String? {
+    let components = diagnosticDescription.split(separator: "'", omittingEmptySubsequences: false)
+    guard components.count >= 3 else {
+      return nil
+    }
```
ahoppen: Could we change this to `components.count == 3` so that if the diagnostic message should change to contain more `'` characters in the future, we don’t start looking up random text?

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -201,4 +252,86 @@ actor DiagnosticReportManager {
...
+
+    return CodeAction(
+      title: "Add import for '\(moduleName)'",
+      kind: .quickFix,
+      diagnostics: nil,
+      edit: WorkspaceEdit(changes: [snapshot.uri: [edit]])
+    )
+  }
+
+  private func uniqueModuleProvidingSymbol(named symbolName: String, using index: CheckedIndex) throws -> String? {
```
ahoppen: I think this would make sense as a `fileprivate extension` on `CheckedIndex` in this file because it doesn’t use any members from  `DiagnoticReportManager`.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -201,4 +252,86 @@ actor DiagnosticReportManager {
...
+    var moduleNames = Set<String>()
```
ahoppen: Do we even need to have a `Set` here? Since this should only ever contain one symbol, wouldn’t `String?` make more sense?

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -201,4 +252,86 @@ actor DiagnosticReportManager {
...
+
+      moduleNames.insert(moduleName)
+
+      return moduleNames.count <= 1
+    }
+
+    return moduleNames.only
+  }
+
+  private func addImportEdit(
```
ahoppen: Since this is only called from `missingImportCodeAction`, which isn’t terribly complicated so far, would it make sense to inline the code there and keep all functionality localized?

ahoppen (commented):
I just have a few stylistic nitpicks inline, otherwise LGTM.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: Thanks,I removed the intermediate array and compute `hasMissingImportDiagnostic` directly.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: Yes you're right, I changed it for `components.count == 3`

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: Done. I moved this into a `fileprivate` extension on `CheckedIndex`.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: I switched this to track a single optional module name instead of building a set.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
GerardSerra18: That's true, I inlined this into `missingImportCodeAction` to keep the edit construction localized there.

ahoppen (approved):
Looks good. Thanks for providing this, @GerardSerra18 🎉

ahoppen:
@swift-ci Please test

GerardSerra18:
> Looks good. Thanks for providing this, @GerardSerra18 🎉

Thanks a lot for the review and approval, @ahoppen! 
I’ll keep an eye on the CI results.

ahoppen:
@swift-ci Please test Windows

GerardSerra18:
@ahoppen I checked the CI logs and the failures look unrelated to the missing-import changes. The failing targets seem to be `SourceKitLSPTests` / `BuildServerIntegrationTests`, and I don’t see the new missing-import tests reported as failures.

Would it make sense to retrigger CI?

ahoppen:
The issue is that accessing `workspace.uncheckedIndex` in `SwiftLanguageService` doesn’t return until the SwiftPM package manifest has been loaded, which is blocked in the test to simulate slow package loading (eg. dependency resolution). 

I think what we should do is change `DiagnosticReportManager` to take an async closure that returns the index. That way, we don’t need to access the index at all if we only have swift-syntax based diagnostics. And while at it, we should probably also add a 500ms timeout to getting the index so that a missing (for whichever reason) doesn’t block diagnostics from being returned.

GerardSerra18:
@ahoppen thanks for tracking this down.

I updated `DiagnosticReportManager` to take an async index provider instead of eagerly resolving `workspace.uncheckedIndex` while constructing `SwiftLanguageService`. The index is now only requested when SourceKit returns a missing-import diagnostic, and the lookup is capped with a 500ms timeout so diagnostics can still be returned without the missing-import quick fix if the index is unavailable or still blocked on package loading.

Sources/SwiftLanguageService/DiagnosticReportManager.swift
```diff
@@ -300,6 +300,25 @@ actor DiagnosticReportManager {
       edit: WorkspaceEdit(changes: [snapshot.uri: [edit]])
     )
   }
+
+  private func uncheckedIndex(timeout: Duration) async -> UncheckedIndex? {
+    let uncheckedIndexProvider = self.uncheckedIndexProvider
+
+    return await withTaskGroup(of: UncheckedIndex?.self) { group in
```
ahoppen: We have `withTimeout`, which you  should  be able to use here.

GerardSerra18: Good point, updated to use `withTimeout` here.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
Looks like a file isn’t formatted correctly. Could you run `swift-format` as described in https://github.com/swiftlang/sourcekit-lsp/blob/main/CONTRIBUTING.md#formatting?

GerardSerra18:
@ahoppen Done, thanks for catching that

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Congrats and thank you for your first contribution, @GerardSerra18.

GerardSerra18:
@ahoppen Thank you! I really appreciate the review and guidance.

PR #2705, opened 2026-07-03
Add "Flip operands" code action for binary expressions

### LINKED ISSUE #2520: Add Flip operands of binary expression code action

krishnapermi:
Adds a new `FlipOperands` code action that swaps the left and right operands of a binary expression.

I noticed while browsing the open issues that #2520 had been filed for this — it's a handy refactoring that I reach for pretty often when reading conditions like `if 0 < count` and wanting them the more natural way around.

**What it does:**

For comparison operators, the operator direction is flipped to preserve semantics:
```swift
// Before
if 5 < count { process() }

// After
if count > 5 { process() }
```

For commutative operators (`+`, `*`, `==`, `!=`, `===`, `!==`, `&&`, `||`, `&`, `|`, `^`), only the operands are swapped:
```swift
// Before
let sum = 1 + value

// After
let sum = value + 1
```

Operators where swapping would change semantics (`-`, `/`, `%`, etc.) are not offered.

**Implementation:**
- `FlipOperands.swift` — implements `SyntaxRefactoringProvider` on `SequenceExprSyntax`
- Wired into `SyntaxRefactoringCodeActionProvider` adapter
- Tests follow the same pattern as `RemoveRedundantParenthesesTests`

**Limitation:** Only handles simple three-element sequences (one binary operator). Compound expressions like `a + b + c` aren't offered since they require operator folding to know which sub-expression to flip.

Closes #2520

### DIFF
--- a/Sources/SwiftSyntaxCodeActions/CMakeLists.txt
+++ b/Sources/SwiftSyntaxCodeActions/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(SwiftSyntaxCodeActions STATIC
   ConvertStringConcatenationToStringInterpolation.swift
   ConvertZeroParameterFunctionToComputedProperty.swift
   DeclModifierRemover.swift
+  FlipOperands.swift
   FormatRawStringLiteral.swift
   IndentationRemover.swift
   IntegerLiteralUtilities.swift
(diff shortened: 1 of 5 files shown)


ahoppen:
Hi @krishnapermi,

Thanks for the PR. We just merged https://github.com/swiftlang/sourcekit-lsp/pull/2694, which implements this refactoring action. I haven’t looked at your code yet. If you believe that you cover something that isn’t covered as part of that PR, feel free to rebase this one and re-open it.

PR #2707, opened 2026-07-03
Add 'Generate enum associated value accessors' code action

### LINKED ISSUE #2522: Add Generate enum associated value accessors code action

krishnapermi:
Noticed issue #2522 while browsing open code action requests and thought this would be a genuinely useful one to have - generating `as`/`is` accessors by hand is the kind of boilerplate that adds up quickly.

This PR adds a `GenerateEnumAssociatedValueAccessors` syntactic code action. When the cursor is on an enum case that has associated values, it inserts computed properties for each element:

- `var as<Case>: Type?` - pattern-matches and extracts the associated value (returns `nil` for other cases)
- - `var is<Case>: Bool` - checks whether the enum is in that case
For a single associated value the property returns the value directly. For multiple values it returns a tuple. The properties are inserted just before the enum's closing brace, indented to match the case declaration.

Closes #2522

### DIFF
--- a/Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
+++ b/Sources/SwiftSyntaxCodeActions/GenerateEnumAssociatedValueAccessors.swift
@@ -1,0 +1,125 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(SourceKitLSP) package import LanguageServerProtocol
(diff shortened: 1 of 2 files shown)


ahoppen:
Hi @krishnapermi,

https://github.com/swiftlang/sourcekit-lsp/pull/2680 is already implementing it and very close to the finish line, so we’re taking that PR. If you believe that your PR covers something that https://github.com/swiftlang/sourcekit-lsp/pull/2680 does not, feels free to rebase it once https://github.com/swiftlang/sourcekit-lsp/pull/2680 is merged and re-open this PR.

PR #2710, opened 2026-07-04
Avoid opening generated interfaces for root workspace modules

### LINKED ISSUE #2693: `textDocument/definition` on a reference to a project package opens `.swiftinterface` file instead
### Swift version

Swift version 6.5-dev (LLVM 18d2bfb70c14d89, Swift c13d82e5987aecb)

### Platform

Linux

### Editor

Visual Studio Code, latest Swift plugin version

### Description

Navigating via `textDocument/definition` on a reference to a package of the local workspace opens a `.swiftinterface` representation of that package.
IMHO, instead, it should either reference the source folder of the package or reject the request.

I have not researched if a folder is a valid response to `textDocument/definition`.

In an ideal world, the folder would be returned and a rename request would rename references to the package, the directory and update the setup in the Package.swift file. But that's probably out-of-scope for SourceKit LSP.

I originally noticed this with a cloned https://github.com/swiftlang/swift-subprocess, but created a small example to reproduce the problem.

Thank you for your hard work!

### Steps to Reproduce

1. Clone https://github.com/jansorg/sourcekit-lsp-bug-demo
2. Open `Tests/swift_sourcekit_demoTests/swift_sourcekit_demoTests.swift`
3. Ctrl+Click on package reference (`swift_sourcekit_demo`, line 5)

Now VSCode (via SourceKit LSP) opens a .swiftinterface. Depending on previous work, it's either empty or not.
In my tests it was usually empty

### Logging

_No response_

somiljain2006:
Fixes #2693 

Update definition lookup for module symbols to prefer local declarations and indexed definitions before falling back to generated interfaces.

- Uses workspace build metadata to detect root-project modules and returns no definition instead of opening a generated .swiftinterface for local package targets.
- Preserves existing behavior for SDK and external modules so they continue to generate interfaces.
- Threads build server context through the shared definition lookup.
- Adds regression tests covering both root package modules and system modules to protect this boundary.

### DIFF
--- a/Sources/SourceKitLSP/DefinitionLocations.swift
+++ b/Sources/SourceKitLSP/DefinitionLocations.swift
@@ -64,8 +64,7 @@ package func definitionLocations(
   languageService: any LanguageService,
   buildServerManager: BuildServerManager?
 ) async throws -> DefinitionLocationsResult {
-  // If this symbol is a module then generate a textual interface, unless the build graph says that the module is part
-  // of the root project. SourceKit-LSP has no file/range location for a module target itself in that case.
+  // If this symbol is a module then generate a textual interface
   if symbol.kind == .module {
     if let bestLocalDeclaration = symbol.bestLocalDeclaration {
       return DefinitionLocationsResult(locations: [bestLocalDeclaration])
--- a/Sources/SourceKitLSP/DefinitionLocations.swift
+++ b/Sources/SourceKitLSP/DefinitionLocations.swift
@@ -33,8 +33,8 @@ package struct DefinitionLocationsResult {
 
 /// Returns a source-backed definition location for `moduleName` if it is a module in the root SwiftPM package.
 ///
-/// Swift does not have a source declaration for a module. For a SwiftPM root target, the closest existing source-backed
-/// representation is the package manifest, which is the build server configuration file for SwiftPM workspaces.
+/// Swift does not have a source declaration for a module. For SwiftPM root targets,
+/// use the package manifest as the source-backed definition location.
 private func definitionLocationInWorkspacePackage(
   moduleName: String,
   buildServerManager: BuildServerManager?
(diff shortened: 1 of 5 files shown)


somiljain2006:
@ahoppen @rintaro Can you review this pr?

Sources/SourceKitLSP/DefinitionLocations.swift
```diff
@@ -51,6 +94,10 @@ package func definitionLocations(
       return DefinitionLocationsResult(locations: [])
     }
 
+    if await moduleIsPartOfRootProject(moduleName: moduleName, buildServerManager: buildServerManager) == true {
+      return DefinitionLocationsResult(locations: [])
```
ahoppen: Not returning any location here is, in my opinion, a regression to what we have right now. If we can’t jump to the directory on disk, I’d prefer to keep the current behavior.

somiljain2006:
@ahoppen Instead of returning no result, I updated the PR to use the workspace metadata to map root-package modules to their Package.swift manifest and return that as the definition location (falling back to the generated interface if no manifest is available). Would that be an acceptable target, or were you thinking of something else?

ahoppen:
I personally don’t think that makes much sense. Imagine yourself cmd-clicking on a module. What would you expect to happen? Does jumping to the Package manifest provide any value? I would argue that you’d like to see the source code that implements this module.

somiljain2006:
That makes sense, and I agree that jumping to Package.swift isn't a useful destination from a user's perspective. I should have given more thought to this earlier.

While looking into this, I found that SourceKit/IndexStoreDB don't appear to expose a source declaration for a module symbol, and the build graph exposes the set of source files that belong to a target rather than a canonical source location representing the module itself. Because of that, I don't see an obvious source-backed location that naturally represents the module.

Returning an arbitrary source file doesn't seem ideal because the choice isn't semantically meaningful, while returning every source file could become cumbersome for larger targets. Given that, it seems like we may need to identify an existing abstraction that represents the implementation of the module rather than introducing a heuristic.

Until such a source-backed location exists, I think it probably makes sense to keep the current generated-interface behavior rather than replace it with something equally arbitrary.

PR #2711, opened 2026-07-07
Filter enclosing syntactic tokens during semantic token merging

### LINKED ISSUE #2708: @TaskLocal and other attached macro attributes fall through the parser and produce no token
### Swift version

main

### Platform

any

### Editor

_No response_

### Description

Attached macro usages at attribute position (e.g. `@TaskLocal`, `@Observable`) are not emitted as LSP semantic tokens.

Macro declarations (macro Foo() = #externalMacro(...)) and freestanding uses (#stringify(...)) correctly arrive as macro tokens.

### Steps to Reproduce

```
  class C {
      @TaskLocal << want this to be identified as macro
      static var taskLocalLogger: Logger = Logger(label: "", factory: { _ in SwiftLogNoOpLogHandler(label: "") })
  }
```

textDocument/semanticTokens/full returns no token covering TaskLocal. The declaration is in Swift.swiftinterface:

```
  @attached(accessor) @attached(peer, names: prefixed(`$`))
  public macro TaskLocal() = #externalMacro(module: "SwiftMacros", type: "TaskLocalMacro")
```

### Logging

_No response_

somiljain2006:
Fixes #2708 

Update SyntaxHighlightingTokens.mergingTokens to treat containing ranges as overlapping during token merging, allowing more specific semantic tokens to replace enclosing syntactic tokens instead of only removing tokens with identical ranges. This avoids retaining redundant syntactic tokens, such as the attribute modifier covering `@TaskLocal` when a semantic macro token already exists for TaskLocal, while preserving the existing behavior for unrelated tokens. Add a regression test covering attached macro highlighting.

### DIFF
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -89,7 +89,8 @@ extension SwiftLanguageService {
       }
 
     return SyntaxHighlightingTokens(
-      sortedTokens: tree
+      sortedTokens:
+        tree
         .classifications(in: range)
         .flatMap { $0.highlightingTokens(in: snapshot).tokens }
     )
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -88,11 +88,11 @@ extension SwiftLanguageService {
         tree.range
       }
 
-    let tokens = tree
-      .classifications(in: range)
-      .flatMap { $0.highlightingTokens(in: snapshot).tokens }
-
-    return SyntaxHighlightingTokens(sortedTokens: tokens)
+    return SyntaxHighlightingTokens(
+      sortedTokens: tree
+        .classifications(in: range)
+        .flatMap { $0.highlightingTokens(in: snapshot).tokens }
+    )
   }
 
   package func documentSemanticTokens(
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -61,8 +61,8 @@ extension SwiftLanguageService {
     try Task.checkCancellation()
 
     async let semanticTokens =
-          await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
-          ?? SyntaxHighlightingTokens(sortedTokens: [])
+        await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+        ?? SyntaxHighlightingTokens(sortedTokens: [])
 
     let tokens =
       if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -61,8 +61,8 @@ extension SwiftLanguageService {
     try Task.checkCancellation()
 
     async let semanticTokens =
-        await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
-        ?? SyntaxHighlightingTokens(sortedTokens: [])
+      await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+      ?? SyntaxHighlightingTokens(sortedTokens: [])
 
     let tokens =
       if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -70,7 +70,7 @@ extension SwiftLanguageService {
       } else {
         await semanticTokens
       }
-    return tokens.sorted({ $0.start < $1.start })
+    return tokens
   }
 
   private func syntacticTokens(
--- a/Sources/SwiftLanguageService/SemanticTokens.swift
+++ b/Sources/SwiftLanguageService/SemanticTokens.swift
@@ -61,8 +61,8 @@ extension SwiftLanguageService {
     try Task.checkCancellation()
 
     async let semanticTokens =
-      await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
-      ?? SyntaxHighlightingTokens(tokens: [])
+          await orLog("Loading semantic tokens") { try await semanticHighlightingTokens(for: snapshot) }
+          ?? SyntaxHighlightingTokens(sortedTokens: [])
 
     let tokens =
       if self.options.reportSyntacticHighlightInSemanticTokensOrDefault {
@@ -88,11 +88,11 @@ extension SwiftLanguageService {
         tree.range
       }
 
-    return
-      tree
+    let tokens = tree
       .classifications(in: range)
-      .map { $0.highlightingTokens(in: snapshot) }
-      .reduce(into: SyntaxHighlightingTokens(tokens: [])) { $0.tokens += $1.tokens }
+      .flatMap { $0.highlightingTokens(in: snapshot).tokens }
+
+    return SyntaxHighlightingTokens(sortedTokens: tokens)
   }
 
   package func documentSemanticTokens(
@@ -126,7 +126,7 @@ extension SwiftLanguageService {
 extension SyntaxClassifiedRange {
   fileprivate func highlightingTokens(in snapshot: DocumentSnapshot) -> SyntaxHighlightingTokens {
     guard let (kind, modifiers) = self.kind.highlightingKindAndModifiers else {
-      return SyntaxHighlightingTokens(tokens: [])
+      return SyntaxHighlightingTokens(sortedTokens: [])
     }
 
     let multiLineRange = snapshot.positionRange(of: self.range)
@@ -140,7 +140,7 @@ extension SyntaxClassifiedRange {
       )
     }
 
-    return SyntaxHighlightingTokens(tokens: tokens)
+    return SyntaxHighlightingTokens(sortedTokens: tokens)
   }
 }
 
--- a/Sources/SwiftLanguageService/SyntaxHighlightingTokenParser.swift
+++ b/Sources/SwiftLanguageService/SyntaxHighlightingTokenParser.swift
@@ -27,7 +27,7 @@ struct SyntaxHighlightingTokenParser {
   private func parseTokens(
     _ response: SKDResponseDictionary,
     in snapshot: DocumentSnapshot,
-    into tokens: inout SyntaxHighlightingTokens
+    into tokens: inout [SyntaxHighlightingToken]
   ) {
     let keys = sourcekitd.keys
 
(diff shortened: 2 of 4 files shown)


somiljain2006:
@rintaro @ahoppen Can you review this pr?

ahoppen:
@somiljain2006 You don’t need to ping us for a review a few hours after you opened the PR. We get notifications for PRs when they are opened and will review them when we get to them.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,35 @@ package struct SyntaxHighlightingTokens: Sendable {
...
+        let otherRange = otherRanges[mid]
+
+        if token.range.clamped(to: otherRange) == otherRange || otherRange.clamped(to: token.range) == token.range {
+          return false
+        } else if otherRange.lowerBound >= token.range.upperBound {
+          high = mid - 1
+        } else {
+          low = mid + 1
+        }
+      }
```
ahoppen: Does this mean we do a binary search in `other` for every token with a complexity of `O(n * log(n))`. That seems really unfortunate. Since both token ranges are sorted, can’t we iterate through them simultaneously and thus stay in `O(n)` complexity?

somiljain2006:
> @somiljain2006 You don’t need to ping us for a review a few hours after you opened the PR. We get notifications for PRs when they are opened and will review them when we get to them.

Sorry I wasn't sure whether mentioning reviewers was the preferred workflow. I'll avoid pinging for reviews from now on and will wait for the normal review process. Thanks for the clarification.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,53 @@ package struct SyntaxHighlightingTokens: Sendable {
...
+        otherIndex += 1
+      } else if token.range.upperBound <= otherToken.range.lowerBound {
+        filteredTokens.append(token)
+        selfIndex += 1
+      } else {
+        let tokenRange = token.range
+        let otherRange = otherToken.range
+
+        let syntacticEnclosesSemantic = tokenRange.lowerBound <= otherRange.lowerBound && tokenRange.upperBound >= otherRange.upperBound
+        let semanticEnclosesSyntactic = otherRange.lowerBound <= tokenRange.lowerBound && otherRange.upperBound >= tokenRange.upperBound
```
ahoppen: Can’t you use `tokenRange.contains(otherRange)` here?

ahoppen: I don’t think we should call these variables syntactic and semantic because this is merging to arbitrary token sets. Let’s call them `self` and `other` to stick with the terminology in the rest of the file.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,53 @@ package struct SyntaxHighlightingTokens: Sendable {
...
+
+        if syntacticEnclosesSemantic || semanticEnclosesSyntactic {
+          selfIndex += 1
+        } else {
+          if tokenRange.upperBound <= otherRange.upperBound {
+            filteredTokens.append(token)
+            selfIndex += 1
+          } else {
+            otherIndex += 1
+          }
```
ahoppen: This can be flattened to an `else if` chain

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,53 @@ package struct SyntaxHighlightingTokens: Sendable {
...
-  /// preferring the given array's tokens if duplicate ranges are
+  /// preferring the given array's tokens if overlapping ranges are
   /// found.
   package func mergingTokens(with other: SyntaxHighlightingTokens) -> SyntaxHighlightingTokens {
-    let otherRanges = Set(other.tokens.map(\.range))
-    return SyntaxHighlightingTokens(tokens: tokens.filter { !otherRanges.contains($0.range) } + other.tokens)
+    return self.mergingTokens(with: other.tokens)
   }
 
   package func mergingTokens(with other: [SyntaxHighlightingToken]) -> SyntaxHighlightingTokens {
```
ahoppen: Could we merge the two `mergingTokens` functions into one? Seems like we never directly call the function that takes an array.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
ahoppen: In the doc comment you say that you ignore tokens that overlap (which, I think, is the correct behavior) but here you’re checking for tokens enclosed in another. You can use `.overlaps` to check for overlap.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,53 @@ package struct SyntaxHighlightingTokens: Sendable {
...
-    let otherRanges = Set(other.map(\.range))
-    return SyntaxHighlightingTokens(tokens: tokens.filter { !otherRanges.contains($0.range) } + other)
+    var filteredTokens: [SyntaxHighlightingToken] = []
+    filteredTokens.reserveCapacity(tokens.count)
+
+    var selfIndex = 0
+    var otherIndex = 0
```
ahoppen: Instead of keeping track of the indices, I think you can have an iterator for the two arrays (`tokens.makeIterator()`) and then `selfIndex += 1` becomes `token = tokensIterator.next()`.

The advantage of this is that if we at some point want to generalize this to work on any sequence instead of just arrays, it should just be a signature change.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -56,16 +56,35 @@ package struct SyntaxHighlightingTokens: Sendable {
   }
 
   /// Merges the tokens in this array into a new token array,
-  /// preferring the given array's tokens if duplicate ranges are
+  /// preferring the given array's tokens if overlapping ranges are
   /// found.
   package func mergingTokens(with other: SyntaxHighlightingTokens) -> SyntaxHighlightingTokens {
```
ahoppen, line 77: I just realized that `SyntaxHighlightingTokens` aren’t actually guaranteed to be sorted and that we only sort them in `SwiftLanguageService.mergedAndSortedTokens`.

I think what we should do is:
- Add an invariant to `SytnaxHighlightingTokens` declaring that `tokens` are always sorted
- `mergingTokens` could now produce a single merged dictionary that is already sorted in-place (instead of adding `other.tokens` at the end)
- `mergedAndSortedTokens` no longer needs to call `.sorted`

Would you be able to implement that? And sorry for not noticing this earlier.

ahoppen (commented):
This looks a lot nicer. Just noticed one more thing though.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
somiljain2006: No worries at all, I implemented the changes:

- SyntaxHighlightingTokens now maintains the sorted-order invariant.
- mergingTokens now produces a single merged, sorted result directly instead of appending other.tokens and sorting afterward.
- mergedAndSortedTokens no longer calls .sorted().

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -20,6 +20,10 @@ package struct SyntaxHighlightingTokens: Sendable {
   package var tokens: [SyntaxHighlightingToken]
 
   package init(tokens: [SyntaxHighlightingToken]) {
```
ahoppen: We should add the requirement that tokens are sorted as a doc comment to both the initializer and the `tokens` property.

Also, the argument name should be `sortedTokens` to further indicate that these tokens must be sorted as another precaution. 

And if we can’t be 100% sure that the tokens we pass in here are sorted, I think it would be better to be conservative and sort them again, because non-sorted tokens will totally mess up the syntax highlighting.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -20,6 +20,10 @@ package struct SyntaxHighlightingTokens: Sendable {
   package var tokens: [SyntaxHighlightingToken]
```
ahoppen: Can we make this a `let` to ensure the tokens don’t get changed after the assertion was used in the initializer, thus creating `SyntaxHighlightingTokens` that aren’t sorted?

ahoppen (commented):
Nice. Just a few small comments inline.

Sources/SwiftLanguageService/SemanticTokens.swift
```diff
@@ -88,11 +88,11 @@ extension SwiftLanguageService {
...
 
-    return
-      tree
+    let tokens = tree
       .classifications(in: range)
-      .map { $0.highlightingTokens(in: snapshot) }
-      .reduce(into: SyntaxHighlightingTokens(tokens: [])) { $0.tokens += $1.tokens }
+      .flatMap { $0.highlightingTokens(in: snapshot).tokens }
+
+    return SyntaxHighlightingTokens(sortedTokens: tokens)
```
ahoppen:
```suggestion
    return tree
 /// ...
      .flatMap { $0.highlightingTokens(in: snapshot) }
```

Sources/SwiftLanguageService/SyntaxHighlightingTokenParser.swift
```diff
@@ -80,9 +80,9 @@ struct SyntaxHighlightingTokenParser {
   }
 
   func parseTokens(_ response: SKDResponseArray, in snapshot: DocumentSnapshot) -> SyntaxHighlightingTokens {
-    var tokens: SyntaxHighlightingTokens = SyntaxHighlightingTokens(tokens: [])
+    var tokens: [SyntaxHighlightingToken] = []
     parseTokens(response, in: snapshot, into: &tokens)
-    return tokens
+    return SyntaxHighlightingTokens(sortedTokens: tokens)
```
ahoppen: This is a position where I’d like to play it safe and sort the tokens again.

Can we have an initializer in `SyntaxHighlightingTokens` that takes a potentially unsorted list and sorts them? And then call it from here.

Sources/SwiftLanguageService/SyntaxHighlightingTokens.swift
```diff
@@ -17,14 +17,15 @@ import SourceKitLSP
...
-    self.tokens = tokens
+
+  /// Syntax highlighting tokens sorted by their start position.
+  package let tokens: [SyntaxHighlightingToken]
+
+  /// Creates a syntax highlighting token collection.
+  ///
+  /// - Parameter sortedTokens: Syntax highlighting tokens sorted by their start position.
+  package init(sortedTokens: [SyntaxHighlightingToken]) {
+    self.tokens = sortedTokens.sorted { $0.start < $1.start }
```
ahoppen: I meant that if the parameter is called `sortedTokens`, we should just have an assertion on them being sorted and avoid the redundant sort.

ahoppen (commented):
Just a few more nitpicks

ahoppen (approved):

somiljain2006:
@ahoppen Can you run the ci?

ahoppen:
Sorry, forgot to trigger it.

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

somiljain2006:
@ahoppen CI failed due to formatting issues. I have fixed it.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2712, opened 2026-07-07
Always advertise the SourceKit-LSP semantic tokens legend

### LINKED ISSUE #2643: SourceKit-LSP overrides syntax highlighting provided by clangd
Summary:
I noticed that while editing some Objective-C related code, that the colors of all of the variables in my source files changed.

Steps to Reproduce:
1. Install the vscode-clangd extension. 
2. Install the vscode-swift extension.
3. Edit some Swift files, notice that syntax highlighting works properly.
4. Go to an Objective-C or C file, notice that the colors used are completely wrong. (Is this SourceKit-LSP related?)
5. As a bonus, restarting clangd restores the original colors, but then restarting the Swift LSP after that causes them to get overwritten again.

Expected Behavior:
Swift's LSP should only be used for Swift files, and clangd's LSP should be used for C/Objective-C files.

Environment:
 - macOS 26.4
 - Apple Swift version 6.3 (swiftlang-6.3.0.123.5 clang-2100.0.123.102)
 - Visual Studio Code version: 1.118.1 (Universal)
 - vscode-swift version: 2.16.4

Steffeeen:
Fixes #2643.

When the client supports dynamic semantic token registration we previously advertised the clangd token legend while still translating the tokens into the SourceKit-LSP legend. This mismatch lead to incorrect highlighting as seen in the linked issue.

### DIFF
--- a/Sources/ClangLanguageService/ClangLanguageService.swift
+++ b/Sources/ClangLanguageService/ClangLanguageService.swift
@@ -378,13 +378,18 @@ extension ClangLanguageService {
         workspaceFolders: nil
       )
     )
-    self.capabilities = result.capabilities
+    var capabilities = result.capabilities
     if let legend = result.capabilities.semanticTokensProvider?.legend {
+      // Semantic token responses from clangd are translated to SourceKit-LSP's legend below before being sent to the
+      // client, so dynamic registrations must advertise SourceKit-LSP's legend instead of the clangd one.
+      capabilities.semanticTokensProvider?.legend = .sourceKitLSPLegend
       self.semanticTokensTranslator = SemanticTokensLegendTranslator(
         clangdLegend: legend,
         sourceKitLSPLegend: SemanticTokensLegend.sourceKitLSPLegend
       )
     }
+    self.capabilities = capabilities
+
     if let sourceKitLSPServer {
       // Since `ClangLanguageService` is created per toolchain, different clangd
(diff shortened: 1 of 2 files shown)


Sources/ClangLanguageService/ClangLanguageService.swift
```diff
@@ -378,21 +378,25 @@ extension ClangLanguageService {
...
     if let legend = result.capabilities.semanticTokensProvider?.legend {
       self.semanticTokensTranslator = SemanticTokensLegendTranslator(
         clangdLegend: legend,
         sourceKitLSPLegend: SemanticTokensLegend.sourceKitLSPLegend
       )
     }
+    // Semantic token responses from clangd are translated to SourceKit-LSP's legend below before being sent to the
+    // client, so dynamic registrations must advertise SourceKit-LSP's legend instead of the clangd one.
+    var capabilities = result.capabilities
+    capabilities.semanticTokensProvider?.legend = .sourceKitLSPLegend
```
ahoppen: I think this only needs to happen if we set up a `semanticTokensTranslator`, right? Otherwise, we don’t perform any legend translation.

ahoppen (commented):
Thanks for finding the issue here, Steffen! One small comment inline.

ahoppen (approved):
Thank you!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2713, opened 2026-07-08
Resolve reference-like requests from the index for unopened files

rintaro:
`textDocument/references`, `textDocument/prepareCallHierarchy`, `textDocument/implementation`, and `textDocument/prepareTypeHierarchy` previously resolved the symbol at the requested position via the document's language service (`symbolInfo`), which requires the document to be open. Invoked on a position in a file the editor hadn't opened, they returned nothing.

These requests now resolve the USR(s) at the requested position directly from the index when the document is not open. When the document is open we still prefer cursor info from the live buffer, which is more precise and also resolves local, non-indexed symbols. This matches the LSP model (on-disk contents are authoritative for a closed document). The shared resolution lives in a new `usrsOfSymbol` helper, and the resolved USRs are sorted and de-duplicated for deterministic results.

`prepareTypeHierarchy` no longer pre-filters the cursor-info symbols by kind; the existing filter on the index symbol kind is authoritative and applies to both resolution paths.

To support this, `CheckedIndex` gains `symbolOccurrences(inFilePath:)`, returning up-to-date occurrences with their locations.

### DIFF
--- a/Sources/SemanticIndex/CheckedIndex.swift
+++ b/Sources/SemanticIndex/CheckedIndex.swift
@@ -161,6 +161,17 @@ package final class CheckedIndex {
     return try index.symbols(inFilePath: path)
   }
 
+  /// All symbol occurrences recorded in the index for the given file, filtered to those that are up-to-date.
+  ///
+  /// Unlike `symbols(inFilePath:)`, each occurrence carries its `SymbolLocation` (line/column), so callers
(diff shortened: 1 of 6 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #2714, opened 2026-07-08
Fix InlayHintRequest race against shiftedPosition

sean7218:
## Description

The unstructured task `InlayHintRequest` sending to **sourcekitd** will return an empty array of hints with multi-code-point + newline edits like this:

**From**

```swift
let x1️⃣ = 1
```

**To**
```swift
let xabc
    👨‍💻1️⃣ = 1
```

So when `InlayHintManager.processEdits` is triggered, the following two race against each other
1. `InlayHintRequest` to **sourcekitd** that returns nothing. 
2. synchronous position shifting logic that returns `: Int`

The test `testInlayHintShiftingWithMultiCodePointAndNewlineInsertionDirectlyBeforeAHint` is asserting against the second case - position shifting logic, thus it is flaky when the first case wins the race.

The fix is to add a gate inside InlayHintManager to prevent InlayHintRequest's empty response writing to the cache `cache[uri] = []` and let the `shiftedPosition` writing to the cache `cache[uri] = shiftedHints`

### DIFF
--- a/Tests/SourceKitLSPTests/InlayHintTests.swift
+++ b/Tests/SourceKitLSPTests/InlayHintTests.swift
@@ -118,12 +118,13 @@ final class InlayHintTests: SourceKitLSPTestCase {
   private func runInlayHintTestCase(
     initialText: String,
     range: (start: String, end: String)? = nil,
+    hooks: Hooks = Hooks(),
     testBody: (InlayHintTestCaseContext) async throws -> Void
   ) async throws {
     let capabilities = ClientCapabilities(
       workspace: WorkspaceClientCapabilities(inlayHint: RefreshRegistrationCapability(refreshSupport: true))
     )
-    let testClient = try await TestSourceKitLSPClient(capabilities: capabilities)
+    let testClient = try await TestSourceKitLSPClient(hooks: hooks, capabilities: capabilities)
     let uri = DocumentURI(for: .swift)
 
     let positions = testClient.openDocument(initialText, uri: uri)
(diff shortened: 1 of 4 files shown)


Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -575,18 +576,31 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
+    let shouldGateBackgroundRefresh = ThreadSafeBox<Bool>(initialValue: false)
+    let backgroundRefreshMayProceed = MultiEntrySemaphore(name: "Inlay hint background refresh may proceed")
+    var hooks = Hooks()
+    hooks.inlayHintRefreshWillUpdateCache = {
+      // Gate the background refresh so it can't race ahead of the read below, which asserts on the synchronously
+      // shift-computed hints rather than a background recompute.
+      if shouldGateBackgroundRefresh.value {
+        await backgroundRefreshMayProceed.waitOrXCTFail()
+      }
+    }
```
rintaro: We have six instances of this. Can you factor out this somehow?

Tests/SourceKitLSPTests/InlayHintTests.swift
```diff
@@ -575,18 +576,31 @@ final class InlayHintTests: SourceKitLSPTestCase {
...
 
   func testInlayHintShiftingWorks() async throws {
+    let shouldGateBackgroundRefresh = ThreadSafeBox<Bool>(initialValue: false)
+    let backgroundRefreshMayProceed = MultiEntrySemaphore(name: "Inlay hint background refresh may proceed")
+    var hooks = Hooks()
+    hooks.inlayHintRefreshWillUpdateCache = {
+      // Gate the background refresh so it can't race ahead of the read below, which asserts on the synchronously
+      // shift-computed hints rather than a background recompute.
+      if shouldGateBackgroundRefresh.value {
+        await backgroundRefreshMayProceed.waitOrXCTFail()
```
rintaro: Do we actually need to restart the background refresh? What if we just block it indefinitely? e.g.
```swift
while !Task.isCancelled { try? await Task.sleep(for: .seconds(1)) }
```
The request should be cancelled when the server is destructed.

rintaro: On second thought I don't think this is a good idea for hygiene. So please disregard.

rintaro:
@swift-ci please test

rintaro (approved):
LGTM, thanks!

nkcsgexi:
Awesome!

PR #2715, opened 2026-07-08
Support qualified name queries in workspace/symbol

rintaro:
`workspace/symbol` previously matched the query as a flat fuzzy search against bare symbol names, so you couldn't search by qualified name (e.g. `Foo.bar` or `Foo::bar`).

Queries are now split on `.`/`::` into a parent chain + member. A qualified query resolves the container(s) named by the parent chain and returns their members whose name contains the member as a case-insensitive subsequence. An empty member (`Foo.`) lists all members. Both separators are accepted and can be mixed (`Foo::Bar.baz`), and a partial chain works (`Inner.member` resolves a nested type without naming its outer scope). Unqualified queries take the flat-search path unchanged.

Unlike the unqualified path, which filters out system symbols, a qualified query deliberately surfaces system SDK results — `String.description` finds the standard-library member. Those results point into the module's generated interface via a `sourcekit-lsp://generated-swift-interface` location (reusing the existing reference-document mechanism); only system namespaces (e.g. `std`) are skipped, to avoid enumerating an entire system scope.

Qualified results carry the fully-qualified name in `name` (e.g. `Foo.bar()`) with no `containerName`, so clients that filter against the label (e.g. VS Code) don't drop them.

No changes to the protocol, `workspace/symbolNames`/`symbolInfo`, or IndexStoreDB.

(`SourceKitLSPServer.swift` getting pretty long for `workspace/symbol` related code. I'm going to make a follow-up PR just for moving them to a dedicated file)

### DIFF
--- a/Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
+++ b/Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
@@ -1,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A `workspace/symbol` query split into its qualified parts.
+///
+/// For example:
+///   `"String.description"`   → `containerChain: ["String"], member: "description"`
+///   `"Foo::bar"`             → `containerChain: ["Foo"], member: "bar"`
+///   `"Outer.Inner.method"`   → `containerChain: ["Outer", "Inner"], member: "method"`
+///   `"Foo."`                 → `containerChain: ["Foo"], member: ""`
+///   `"Foo.bar(baz:)"`        → `containerChain: ["Foo"], member: "bar(baz:)"` (a lone `:` is literal)
+///
+/// An unqualified query (no separator, e.g. `"description"`) fails to parse and returns `nil`.
+package struct QualifiedWorkspaceSymbolQuery: Equatable {
+  /// Container chain in outer-to-inner order. Always non-empty for a successfully-parsed query.
+  package let containerChain: [String]
+
+  /// The trailing component the user is searching for. May be empty (e.g. for the query `Foo.`).
+  package let member: String
+
+  /// Parse a `workspace/symbol` query, splitting on the trailing `.` or `::` qualifier separators.
+  ///
+  /// Returns `nil` if the query has no qualifier or the container chain is empty (so callers can fall
+  /// back to the unqualified search path). A trailing separator with an empty member (e.g. `Foo.`)
+  /// is a valid qualified query that lists all members of the container.
+  package init?(_ query: String) {
+    // Split the query into components on the `.` and `::` separators. The last component is the member
+    // being searched for; the preceding components are the container chain. A lone `:` is not a separator
+    // — it is a literal character, e.g. an argument label in `Collection.append(contentsOf:)`.
+    var components =
+      query
+      .replacing("::", with: ".")
+      .split(separator: ".", omittingEmptySubsequences: false)
+      .map(String.init)
+
+    // Without a separator the query isn't qualified; callers fall back to the unqualified search path.
+    guard components.count >= 2 else { return nil }
+    let member = components.removeLast()
+    let containerChain = components
+    // Container chain segments must be non-empty (rejects leading and doubled separators). An empty
+    // member is allowed (e.g. `Foo.` lists all members of the container).
+    guard !containerChain.contains(where: \.isEmpty) else { return nil }
+
+    self.containerChain = containerChain
+    self.member = member
+  }
+}
+
+extension String {
+  /// Returns `true` if the characters of `subsequence` appear in order within `self`, compared
+  /// case-insensitively. An empty `subsequence` always matches.
+  package func fuzzilyContains(subsequence: String) -> Bool {
+    var remaining = Substring(subsequence.lowercased())
+    for character in self.lowercased() where character == remaining.first {
+      remaining = remaining.dropFirst()
+    }
+    return remaining.isEmpty
+  }
+}
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1584,6 +1584,11 @@ extension SourceKitLSPServer {
   /// Map a `SymbolOccurrence` from the index to a `WorkspaceSymbolItem`, or `nil` if it has no
   /// representable location.
   ///
+  /// If `useQualifiedName` is `true` and the symbol has a container, the item's `name` is the fully-qualified
+  /// name (e.g. `Foo.bar`) and `containerName` is dropped. This is used for qualified queries so that clients
+  /// which filter workspace symbols by matching the query against the item's `name` (e.g. VS Code) keep the
+  /// result — the qualified query wouldn't match the bare member name otherwise.
+  ///
   /// - Parameter referenceDocumentMainFile: The project file whose build settings are used to open the
   ///   symbol's generated interface. **Passing a non-`nil` value changes the shape of the result**: the
   ///   symbol is returned as a `WorkspaceSymbol` with a `sourcekit-lsp://generated-swift-interface`
@@ -1595,18 +1600,28 @@ extension SourceKitLSPServer {
     for symbolOccurrence: SymbolOccurrence,
     in index: CheckedIndex,
     copiedFileMap: CopiedFileMap,
-    referenceDocumentMainFile: DocumentURI?
+    referenceDocumentMainFile: DocumentURI?,
+    useQualifiedName: Bool = false
   ) throws -> WorkspaceSymbolItem? {
     let containerNames = try index.containerNames(of: symbolOccurrence)
-    let containerName: String? =
-      if !containerNames.isEmpty {
-        switch symbolOccurrence.symbol.language {
-        case .cxx, .c, .objc: containerNames.joined(separator: "::")
-        case .swift: containerNames.joined(separator: ".")
-        }
-      } else {
-        nil
+    let separator =
+      switch symbolOccurrence.symbol.language {
+      case .cxx, .c, .objc: "::"
+      case .swift: "."
       }
+    let containerName: String? = containerNames.isEmpty ? nil : containerNames.joined(separator: separator)
+
+    // For qualified queries, put the qualified name in the label (which clients filter against) and drop the
+    // now-redundant container name.
+    let name: String
+    let displayContainerName: String?
+    if useQualifiedName, let containerName {
+      name = "\(containerName)\(separator)\(symbolOccurrence.symbol.name)"
+      displayContainerName = nil
+    } else {
+      name = symbolOccurrence.symbol.name
+      displayContainerName = containerName
+    }
 
     if let referenceDocumentMainFile {
       let (interfaceModuleName, groupName) = Self.splitModuleNameAndGroup(symbolOccurrence.location.moduleName)
@@ -1625,9 +1640,9 @@ extension SourceKitLSPServer {
       )
       return WorkspaceSymbolItem.workspaceSymbol(
         WorkspaceSymbol(
-          name: symbolOccurrence.symbol.name,
+          name: name,
           kind: symbolOccurrence.symbol.kind.asLspSymbolKind(),
-          containerName: containerName,
+          containerName: displayContainerName,
           location: .uri(.init(uri: try urlData.uri)),
           data: data.encodeToLSPAny()
         )
@@ -1638,11 +1653,11 @@ extension SourceKitLSPServer {
     let location = symbolLocation.adjusted(for: copiedFileMap)
     return WorkspaceSymbolItem.symbolInformation(
       SymbolInformation(
-        name: symbolOccurrence.symbol.name,
+        name: name,
         kind: symbolOccurrence.symbol.kind.asLspSymbolKind(),
         deprecated: nil,
         location: location,
-        containerName: containerName
+        containerName: displayContainerName
       )
     )
   }
@@ -1818,6 +1833,13 @@ extension SourceKitLSPServer {
     guard req.query.count >= minWorkspaceSymbolPatternLength else {
       return []
     }
+    if let qualified = QualifiedWorkspaceSymbolQuery(req.query) {
+      return try await qualifiedWorkspaceSymbols(qualified)
+    }
+    return try await unqualifiedWorkspaceSymbols(query: req.query)
+  }
+
+  private func unqualifiedWorkspaceSymbols(query: String) async throws -> [WorkspaceSymbolItem] {
     var items: [WorkspaceSymbolItem] = []
     for workspace in workspaces {
       guard let index = await workspace.index(checkedFor: .deletedFiles) else {
@@ -1826,7 +1848,7 @@ extension SourceKitLSPServer {
       let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
       var symbols: [SymbolOccurrence] = []
       try index.forEachCanonicalSymbolOccurrence(
-        containing: req.query,
+        containing: query,
         anchorStart: false,
         anchorEnd: false,
         subsequence: true,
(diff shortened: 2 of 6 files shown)


Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,115 @@
...
+  /// The trailing component the user is searching for.
+  package let member: String
+
+  /// Whether the query has any parent qualifier.
+  package var isQualified: Bool { !parentParts.isEmpty }
+
+  /// Parse a `workspace/symbol` query, splitting on the trailing `.` or `::` qualifier separators.
+  ///
+  /// Returns `nil` if the query has no qualifier or the parent half is empty (so callers can fall back
+  /// to the existing unqualified search path). A trailing separator with an empty member (e.g. `Foo.`)
```
ahoppen: I don’t think *existing* makes sense in the doc comment going forward.


```suggestion
  /// to unqualified search). A trailing separator with an empty member (e.g. `Foo.`)
```

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,115 @@
...
+    let member = String(scalars[memberStart...])
+    // An empty member is allowed (e.g. `Foo.` lists all members); an empty parent is not.
+    guard !parentRaw.isEmpty else { return nil }
+
+    // Split the parent half on `.` or `::` to support multi-level qualifiers like
+    // `Outer.Inner.method` or `Outer::Inner::method`.
+    var parts: [String] = []
+    var current: [Character] = []
+    var iter = parentRaw.makeIterator()
+    while let c = iter.next() {
```
ahoppen: Instead of scanning here again, can’t we record the separator above and then call `parentRaw.split(separator:)` here?

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    self.parentParts = parts
+    self.member = member
+  }
+}
+
+extension String {
+  /// Returns `true` if the characters of `subsequence` appear in order within `self`, compared
+  /// case-insensitively. An empty `subsequence` always matches.
+  package func fuzzilyContains(subsequence: String) -> Bool {
```
ahoppen, line 61: Is this used anywhere outside of tests?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
     return items
   }
 
+  /// Handle a `workspace/symbol` request whose query contains a qualifier separator (`.` or `::`).
+  ///
+  /// Drives the search from the member side using the existing fuzzy index lookup, then filters the
```
ahoppen: Similar here, I think *existing* doesn’t make sense in a doc comment in the long term because two years from now nobody will remember how this evolved.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+  /// members. Members declared in extensions are included. Members from all matching containers are
+  /// unioned and de-duplicated by USR.
+  private nonisolated func members(
+    ofContainerChain chain: [String],
+    matching member: String,
+    in index: CheckedIndex
+  ) throws -> [SymbolOccurrence] {
+    guard let containerName = chain.last else {
+      return []
+    }
```
ahoppen, line 1948: Should we throw (or at least log) an error here? This should never happen, right?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+    for (offset, part) in ancestors.reversed().enumerated() {
+      guard let parentRel = relations.first(where: { $0.roles.contains(.childOf) }),
+        parentRel.symbol.name.caseInsensitiveCompare(part) == .orderedSame
+      else {
+        return false
+      }
+      if offset == ancestors.count - 1 {
+        break
+      }
+      let parentOccs = (try? index.occurrences(ofUSR: parentRel.symbol.usr, roles: .definition)) ?? []
```
ahoppen: Should we use `primaryDefinitionOrDeclarationOccurrence` here?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+    matching member: String,
+    in index: CheckedIndex
+  ) throws -> [SymbolOccurrence] {
+    guard let containerName = chain.last else {
+      return []
+    }
+    let ancestors = Array(chain.dropLast())
+
+    // Resolve the innermost container(s) by exact name, verifying the outer scope chain.
+    var containerUSRs: [String] = []
```
ahoppen: Do we need `containerUSRs` at all? Wouldn’t `seenContainerUSRs` be sufficient?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+        containerUSRs.append(symbol.symbol.usr)
+      }
+      return true
+    }
+    try Task.checkCancellation()
+
+    // Members declared in an extension are `childOf` the extension symbol, not the extended type. A
+    // type's occurrences carry `extendedBy` relations pointing at its extensions, so add those USRs to
+    // the set of containers whose children we enumerate.
+    for typeUSR in containerUSRs {
```
ahoppen, line 2001: Do we show members of `Foo` when searching for `Namespace.Foo` in the following?

```swift
enum Namespace {}

extension Namespace {
    struct Foo {
        var bar: Int
    }
}
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+      try Task.checkCancellation()
+      for occurrence in try index.occurrences(ofUSR: typeUSR, roles: .all) {
```
ahoppen: We only need to fetch `.extendedBy` here, right?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,136 @@ extension SourceKitLSPServer {
...
+    }
+
+    var members: [SymbolOccurrence] = []
+    var seenUSRs: Set<String> = []
+    for containerUSR in seenContainerUSRs {
+      try Task.checkCancellation()
+      let children = try index.occurrences(relatedToUSR: containerUSR, roles: .childOf)
+      for child in children
+      where !child.location.isSystem && !child.roles.contains(.accessorOf)
+        && child.symbol.name.gi(subsequence: member)
```
ahoppen: Typo with `gi` here?

Tests/SourceKitLSPTests/WorkspaceSymbolsTests.swift
```diff
@@ -193,4 +193,186 @@ class WorkspaceSymbolsTests: SourceKitLSPTestCase {
...
+
+  func testQualifiedWorkspaceSymbolFuzzyMemberMatch() async throws {
+    let project = try await IndexedSingleSwiftFileTestProject(
+      """
+      struct Foo {
+        func 1️⃣longMethodName() {}
+      }
+      """
+    )
+    let response = try await project.testClient.send(WorkspaceSymbolsRequest(query: "Foo.longMethod"))
```
ahoppen: This isn’t actually a fuzz match, is it? It’s a substring match

Tests/SourceKitLSPTests/WorkspaceSymbolsTests.swift
```diff
@@ -193,4 +193,186 @@ class WorkspaceSymbolsTests: SourceKitLSPTestCase {
...
+      """
+    )
+    let response = try await project.testClient.send(WorkspaceSymbolsRequest(query: "Foo."))
+    let names = (response ?? []).compactMap { item -> String? in
+      guard case .symbolInformation(let info) = item else { return nil }
+      return info.name
+    }
+    // All of `Foo`'s members are listed with their qualified name, and members of other types are not.
+    XCTAssertTrue(names.contains("Foo.bar()"), "\(names)")
+    XCTAssertTrue(names.contains("Foo.baz"), "\(names)")
```
ahoppen: We have `assertContains` that already prints the collection being searched.

ahoppen (commented):
Very nice 🤩

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: I simplified the query parsing - just split `.` or `::` and use the last component as the member name.
At this point I feel we should not differentiate `::` and `.`. E.g. we should accept `swift::Info.range` for
```cpp
namespace ns {
struct Info {
  Range range;
};
}
```

This is a deliberate choice. It might seem strange to accept `::` as a separator in Swift (especially after module selector), but there's no strong reason to exclude matched Swift results even if the query contains `::`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I accidentally rewrote `fuzzilyContains` to `gi` right before commiting 😅

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: You're right!

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Yes. Added a test case for this.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Simplified

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: `fuzzilyContains` is used for member name matching. I accidentally rewrote it `gi`...

Tests/SourceKitLSPTests/WorkspaceSymbolsTests.swift
rintaro: Changed the query to `Foo.lmn` to exercise fuzzy matching.

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
ahoppen: I agree that we can be lenient here and allow `::` as a separator in Swift (and `.` in C++).

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,88 @@
...
+    var current = ""
+    var iterator = query.makeIterator()
+    while let character = iterator.next() {
+      switch character {
+      case ".":
+        components.append(current)
+        current = ""
+      case ":":
+        // Must be a `::` separator; a lone `:` is invalid.
+        guard iterator.next() == ":" else { return nil }
```
ahoppen: How about if you want to search for `String.components(separatedBy:)`. That contains a lone but shouldn’t be invalid.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1816,6 +1838,142 @@ extension SourceKitLSPServer {
...
+  /// of the chain. `member` is matched as a case-insensitive subsequence; an empty `member` matches all
+  /// members. Members declared in extensions are included. Members from all matching containers are
+  /// unioned and de-duplicated by USR.
+  private nonisolated func members(
+    ofContainerChain chain: [String],
+    matching member: String,
+    in index: CheckedIndex
+  ) throws -> [SymbolOccurrence] {
+    guard let containerName = chain.last else {
+      throw ResponseError.internalError("members(ofContainerChain:matching:in:) requires a non-empty chain")
```
```suggestion
      throw ResponseError.internalError("\(#function) requires a non-empty chain")
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: One more thing I’d like to test. In the following, the parent of `Foo` will be the extension of `Sub` and I think we don’t manage to jump to `Namespace` as its parent. Or I’m missing something.

```swift
enum Namespace {
  enum Sub {
  }
}

extension Namespace.Sub {
  struct Foo {}
}
```

ahoppen (commented):
One more test I would

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: You were right. Added extension handling in `occurrence(_:matchesAncestors:in:)` and test cases.

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: Good catch. Fixed.

rintaro:
Enabled system symbols for qualified searches when the client supports both `workspace/getReferenceDocument` and `workspaceSymbol/resolve`.
It may seem a bit odd that system symbols are available for qualified searches but not for unqualified searches. But I don't think that's a blocker. System symbols in qualified searches are far less noisy.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,92 @@
...
+///
+/// For example:
+///   `"String.description"`   → `parentParts: ["String"], member: "description"`
+///   `"Foo::bar"`             → `parentParts: ["Foo"], member: "bar"`
+///   `"Outer.Inner.method"`   → `parentParts: ["Outer", "Inner"], member: "method"`
+///   `"Foo."`                 → `parentParts: ["Foo"], member: ""`
+///   `"Foo.bar(baz:)"`        → `parentParts: ["Foo"], member: "bar(baz:)"` (a lone `:` is literal)
+///
+/// An unqualified query (no separator, e.g. `"description"`) fails to parse and returns `nil`.
+@_spi(Testing)
```
ahoppen: Do we need `@_spi(Testing) public` here? Wouldn’t `package` be sufficient?

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,92 @@
...
+public struct QualifiedWorkspaceSymbolQuery: Equatable {
+  /// Parent chain in outer-to-inner order. Always non-empty for a successfully-parsed query.
+  package let parentParts: [String]
```
ahoppen: We end up with three names for the same list of outer scopes: `parentParts` here, `chain` in `members(ofContainerChain:matching:includeSystemSymbols:in:)` (whose doc comment then calls it a "parent chain"), and `ancestors` in `occurrence(_:matchesAncestors:in:)` — and the value is passed straight through untransformed. Could we settle on one term?

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,92 @@
...
+  /// is a valid qualified query that lists all members of the container.
+  package init?(_ query: String) {
+    // Split the query into components on the `.` and `::` separators. The last component is the member
+    // being searched for; the preceding components are the parent chain. A lone `:` is not a separator —
+    // it is a literal character, e.g. an argument label in `Collection.append(contentsOf:)`.
+    var components: [String] = []
+    var current = ""
+    let characters = Array(query)
+    var index = 0
+    while index < characters.count {
```
ahoppen: Just an idea: I think the following would be the same. It does an additional string copy but since this is just query parsing, I don’t think it matters too much.

```swift
var components = query
  .replacing("::", with: ".")
  .split(separator: ".", omittingEmptySubsequences: false)
  .map(String.init)
```

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
```diff
@@ -0,0 +1,92 @@
...
+
+extension String {
+  /// Returns `true` if the characters of `subsequence` appear in order within `self`, compared
+  /// case-insensitively. An empty `subsequence` always matches.
+  package func fuzzilyContains(subsequence: String) -> Bool {
+    var remaining = Substring(subsequence.lowercased())
+    guard let firstNeedle = remaining.first else {
+      return true
+    }
+    var needle = firstNeedle
```
ahoppen: Do we need `needle` at all? It's always `remaining.first`, so this should collapse to the following, which also drops both early returns.
```suggestion
    var remaining = Substring(subsequence.lowercased())
    for character in self.lowercased() where character == remaining.first {
      remaining.removeFirst()
    }
    return remaining.isEmpty
```

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -14,7 +14,7 @@ import BuildServerIntegration
 @_spi(SourceKitLSP) import BuildServerProtocol
 import Dispatch
 import Foundation
-import IndexStoreDB
+@preconcurrency import IndexStoreDB
```
ahoppen: Why is `@preconcurrency` needed now? Can’t we mark the types that need to be sendable as such in  `IndexStoreDB`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1595,18 +1600,28 @@ extension SourceKitLSPServer {
     for symbolOccurrence: SymbolOccurrence,
     in index: CheckedIndex,
     copiedFileMap: CopiedFileMap,
-    referenceDocumentMainFile: DocumentURI?
+    referenceDocumentMainFile: DocumentURI? = nil,
```
ahoppen: Should we omit these default values? The doc comment right above says that a non-`nil` `referenceDocumentMainFile` changes the shape of the result and that enforcing the capability check "is the caller's responsibility" — it was a required parameter for that reason. All three call sites pass both explicitly, so the defaults don't save anyone anything; they just let a future call site silently get the wrong shape.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+          useQualifiedName: true
+        )
+      }
+    }
+    return items
+  }
+
+  /// Whether `occurrence`'s enclosing scopes match `ancestors` (outer-to-inner), walking the `childOf`
+  /// relation chain. An empty `ancestors` imposes no constraint and always matches.
+  private nonisolated func occurrence(
```
ahoppen: `occurrence` returns a `Bool`, so `self.occurrence(symbol, matchesAncestors: ancestors, in: index)` reads like it hands back an occurrence rather than asserting something about one — and the base name shadows its own first parameter. How about making it an extension on `SymbolOccurrence` so it reads `symbol.matchesAncestors(ancestors, in: index)`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+
+    var members: [SymbolOccurrence] = []
+    var seenUSRs: Set<String> = []
+    for containerUSR in containerUSRs {
+      try Task.checkCancellation()
+      let children = try index.occurrences(relatedToUSR: containerUSR, roles: .childOf)
+      for child in children {
+        guard
+          !child.roles.contains(.accessorOf),
+          includeSystemSymbols || !child.location.isSystem,
```
ahoppen, line 2021: `includeSystemSymbols` only filters the individual children at the very end, so for a client without `workspace/getReferenceDocument` support the query `String.` still resolves stdlib `String`, walks every `.extendedBy` extension of it across the SDK, and enumerates all of their `childOf` members and then only reports results that the user may have declared as extensions on `String` in their own module, which would also be surprising behavior.

Since the `isSystemNamespace` check above already avoids this for namespaces, should we bail out on a system container up front too when `includeSystemSymbols` is `false`?

Tests/SourceKitLSPTests/WorkspaceSymbolsTests.swift
```diff
@@ -193,4 +193,468 @@ class WorkspaceSymbolsTests: SourceKitLSPTestCase {
...
+            name: "Foo.bar()",
+            kind: .method,
+            location: Location(uri: project.fileURI, range: Range(project.positions["1️⃣"]))
+          )
+        )
+      ]
+    )
+  }
+
+  func testQualifiedWorkspaceSymbolWithColonSeparator() async throws {
```
ahoppen: This builds and indexes a whole second `IndexedSingleSwiftFileTestProject` with byte-identical source, only to swap `.` for `::` in the query. Can't we send both queries against the one project, the way `testQualifiedWorkspaceSymbolMatchesCXXMemberInNamespace` already does with three?

Tests/SourceKitLSPTests/WorkspaceSymbolsTests.swift
```diff
@@ -193,4 +193,468 @@ class WorkspaceSymbolsTests: SourceKitLSPTestCase {
...
+            name: "Outer.Inner.method()",
+            kind: .method,
+            location: Location(uri: project.fileURI, range: Range(project.positions["1️⃣"]))
+          )
+        )
+      ]
+    )
+  }
+
+  func testQualifiedWorkspaceSymbolMatchesMemberViaInnermostContainerOnly() async throws {
```
ahoppen: Same source as `testQualifiedWorkspaceSymbolMultiLevelChain` right above, just a different query. Could this be a second `send` on that project instead of paying for another index build?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+  /// relation chain. An empty `ancestors` imposes no constraint and always matches.
+  private nonisolated func occurrence(
+    _ occurrence: SymbolOccurrence,
+    matchesAncestors ancestors: [String],
+    in index: CheckedIndex
+  ) -> Bool {
+    var relations = occurrence.relations
+    for (offset, part) in ancestors.reversed().enumerated() {
+      guard let parentRel = relations.first(where: { $0.roles.contains(.childOf) }),
+        parentRel.symbol.name.caseInsensitiveCompare(part) == .orderedSame
```
ahoppen: Instead of using the Foundation `caseInsensitiveCompare`, should we compare the lowercased value of the symbol and part? Seems more Swift-like to me.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+      else {
+        return false
+      }
+      if offset == ancestors.count - 1 {
```
ahoppen: Just thinking if this reads easier as `offset == 0` if you use `ancestors.enumerated().reversed()` above.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+  /// The innermost name is matched exactly (case-insensitive) and its outer scopes must match the rest
+  /// of the chain. `member` is matched as a case-insensitive subsequence; an empty `member` matches all
+  /// members. Members declared in extensions are included. Members from all matching containers are
+  /// unioned and de-duplicated by USR.
+  ///
+  /// System (SDK/stdlib) members are only included if `includeSystemSymbols` is `true`. They can only be
+  /// navigated to through a generated interface, so callers pass `false` when the client can't open one.
+  private nonisolated func members(
+    ofContainerChain chain: [String],
+    matching member: String,
```
ahoppen: Should this parameter be named `fuzzilyMatching` to convey the fuzzy matching behavior?

Tests/SourceKitLSPTests/QualifiedWorkspaceSymbolQueryTests.swift
```diff
@@ -0,0 +1,120 @@
...
+    XCTAssertEqual(colonQuery?.member, "subscript(at:)")
+  }
+
+  func testEmptyParentSegmentRejected() {
+    XCTAssertNil(QualifiedWorkspaceSymbolQuery("Foo..bar"))
+    XCTAssertNil(QualifiedWorkspaceSymbolQuery("Foo.::bar"))
+  }
+}
+
+final class StringFuzzilyContainsTests: XCTestCase {
```
ahoppen: Should this be in a separate file? I think we only have one `XCTestCase` per file everywhere else.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1856,6 +1878,185 @@ extension SourceKitLSPServer {
...
+        break
+      }
+      // A symbol declared in an extension is `childOf` the extension, and the extension's own occurrence
+      // has no `childOf` link to the extended type's enclosing scope. The extended type records the
+      // inverse `extendedBy` relation, so resolve it and continue the walk from the extended type's
+      // definition (whose `childOf` chain reaches the outer scopes).
+      let parentUSR: String
+      if parentRel.symbol.kind == .extension {
+        guard
+          let extendedOccrs = try? index.occurrences(relatedToUSR: parentRel.symbol.usr, roles: .extendedBy),
```
nkcsgexi: Without too much background, I suspect this extension special casing isn't the first occurrence in the LSP codebase. Can we extract a common utility instead of implementing a new one?

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: > only reports results that the user may have declared as extensions on String in their own module, which would also be surprising behavior.

But we do suggest members in `extensions String` in user modules for unqualified queries.
I'd argue we should suggest user declarations even if they're on system containers.
I updated the logic to skip looking into containers declared in system modules.

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: I updated this and other places to `container`/`containerChain`,  but I kept `ancestors` because that is for ancestors of the container.

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: Nice. I updated to it. Performance is not a concern here.

Sources/SourceKitLSP/QualifiedWorkspaceSymbolQuery.swift
rintaro: Not needed. Simplified as you suggested.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: This wasn't needed. Removed.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: This was a residue of the early development. Removed.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Moved to matching `index.containerNames(of:)`

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Nice suggestion, actually we already have it. `CheckedIndex.containerNames(of:)`.
Changed to use it.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
nkcsgexi: Thank you!

nkcsgexi (approved):
Excited to see this happening, Rintaro!

ahoppen (approved):

PR #2716, opened 2026-07-09
Update BackgroundIndexingTests to work with both build backends

owenv:
Change how these tests compare target identifiers so they're compatible with both the builtin SwiftPM indexing functionality and the new BSP implementation. In combination with https://github.com/swiftlang/swift-build/pull/1526 this fixes many of the remaining test failures if I hardcode the BSP implementation as the default.

### DIFF
--- a/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
+++ b/Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
@@ -529,11 +529,11 @@ final class BackgroundIndexingTests: SourceKitLSPTestCase {
     var testHooks = Hooks()
     let expectedPreparationTracker = ExpectedIndexTaskTracker(expectedPreparations: [
       [
-        try ExpectedPreparation(target: "LibA", destination: .target),
-        try ExpectedPreparation(target: "LibB", destination: .target),
+        try ExpectedPreparation(targetName: "LibA"),
+        try ExpectedPreparation(targetName: "LibB"),
       ],
       [
-        try ExpectedPreparation(target: "LibB", destination: .target)
+        try ExpectedPreparation(targetName: "LibB")
       ],
     ])
     testHooks.indexHooks = expectedPreparationTracker.testHooks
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test windows

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -2659,13 +2657,10 @@ final class BackgroundIndexingTests: SourceKitLSPTestCase {
...
-      preparationRequests.value.flatMap(\.targets),
-      [
-        try BuildTargetIdentifier(target: "LibA", destination: .target),
-        try BuildTargetIdentifier(target: "LibB", destination: .target),
-      ]
-    )
+    let preparedTargets = preparationRequests.value.flatMap(\.targets)
+    XCTAssertEqual(preparedTargets.count, 2)
+    XCTAssert(preparedTargets[0].matchesTargetName("LibA"))
+    XCTAssert(preparedTargets[1].matchesTargetName("LibB"))
```
ahoppen: Can you refactor this so the test doesn’t crash if `preparedTargets` has less than 2 elements?

Tests/SourceKitLSPTests/ExpectedIndexTaskTracker.swift
```diff
@@ -37,35 +37,24 @@ enum BuildDestination {
...
-    ]
-
-    struct FailedToConvertSwiftBuildTargetToUrlError: Swift.Error, CustomStringConvertible {
-      var target: String
-      var destination: String
-
-      var description: String {
-        return "Failed to generate URL for target: \(target), destination: \(destination)"
-      }
+  // For testing only. Determine if the BuildTargetIdentifier corresponds to a target with the given name. Matches
```
ahoppen: I don’t think we need `For testing only` if this is in a test target. Not sure why we had it before. Probably an artifact of things moving around.

ahoppen (approved):
Two minor comments, otherwise looks good to me.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
owenv: yep, that's my bad, fixed

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
LLDB flakiness

@swift-ci test windows

PR #2719, opened 2026-07-14
Add 'Extract tuple into struct' code action

### LINKED ISSUE #2511: Add Extract tuple into struct code action
### Description

Extract a named tuple into a struct and update all usage sites.

**Before:**

```swift
func getUser() -> (name: String, age: Int, email: String) {
    return ("Alice", 30, "alice@example.com")
}

let user = getUser()
print(user.name)
```

**After:**

```swift
struct User {
    let name: String
    let age: Int
    let email: String
}

func getUser() -> User {
    return User(name: "Alice", age: 30, email: "alice@example.com")
}

let user = getUser()
print(user.name)
```

krishnapermi:
Implements the "Extract tuple into struct" code action requested in #2511.

When the cursor is on a named tuple type — one where every element has a label — a new code action appears in the lightbulb menu. It generates a `struct` with matching stored properties immediately before the enclosing declaration and replaces the tuple type annotation with the new struct name.

**Before:**
```swift
func getUser() -> (name: String, age: Int, email: String) {
    return ("Alice", 30, "alice@example.com")
}
```

**After:**
```swift
struct GetUserResult {
    let name: String
    let age: Int
    let email: String
}

func getUser() -> GetUserResult {
    return ("Alice", 30, "alice@example.com")
}
```

The struct name is inferred from context:
- Function return type → `<FunctionName>Result` (e.g. `getUser` → `GetUserResult`)
- Variable annotation → capitalised variable name
- Type alias → the alias name itself
- Fallback → concatenated capitalised element labels

Indentation is matched to the surrounding source using `BasicFormat.inferIndentation`.

**Scope:** This is the syntactic-only step discussed in #2511 — only the local type annotation is rewritten. Renaming call sites that construct or destructure the tuple is left to a follow-up rename refactoring (the pattern @ahoppen described as an acceptable first step).

**What doesn't trigger it:** unlabelled tuples, single-element tuples, and tuples with `_` wildcard labels, since none of those produce clean property names.

Resolves #2511.

### DIFF
--- a/Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
+++ b/Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
@@ -1,0 +1,199 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftBasicFormat
+import SwiftExtensions
+@_spi(SourceKitLSP) import LanguageServerProtocol
+import SourceKitLSP
+import SwiftRefactor
+package import SwiftSyntax
+
+/// Syntactic code action that extracts a named tuple type into a named struct.
+///
+/// When the cursor is on a named tuple type — one where every element carries a
+/// label — this action generates a `struct` definition with matching stored
+/// properties immediately before the enclosing declaration, then replaces the
+/// tuple type annotation with the new struct name.
+///
+/// Only the local type annotation is rewritten. Renaming call sites that
+/// construct or destructure the tuple is left to a follow-up rename refactoring.
+///
+/// ## Before
+///
+/// ```swift
+/// func getUser() -> (name: String, age: Int, email: String) {
+///     return ("Alice", 30, "alice@example.com")
+/// }
+/// ```
+///
+/// ## After
+///
+/// ```swift
+/// struct GetUserResult {
+///     let name: String
+///     let age: Int
+///     let email: String
+/// }
+///
+/// func getUser() -> GetUserResult {
+///     return ("Alice", 30, "alice@example.com")
+/// }
+/// ```
+struct ExtractTupleIntoStruct: SyntaxCodeActionProvider {
+  package static func codeActions(in scope: SyntaxCodeActionScope) -> [CodeAction] {
+    guard let tupleType = findNamedTupleType(in: scope) else {
+      return []
+    }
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
```diff
@@ -0,0 +1,199 @@
...
+///     let name: String
+///     let age: Int
+///     let email: String
+/// }
+///
+/// func getUser() -> GetUserResult {
+///     return ("Alice", 30, "alice@example.com")
+/// }
+/// ```
+struct ExtractTupleIntoStruct: SyntaxCodeActionProvider {
```
ahoppen, line 51: Could you implement this in terms of  `SyntaxRefactoringCodeActionProvider`?

Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
```diff
@@ -0,0 +1,199 @@
...
+    guard let tupleType = findNamedTupleType(in: scope) else {
+      return []
+    }
+    guard let enclosingDecl = findEnclosingInsertionPoint(of: Syntax(tupleType)) else {
+      return []
+    }
+
+    let structName = deriveStructName(for: tupleType)
+
+    // Match the indentation style of the surrounding source.
```
ahoppen, line 62: I don’t think this comment provides significant value as it just describes what the code below already clearly states. Please remove such comments.

Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
```diff
@@ -0,0 +1,199 @@
...
+    }.joined(separator: "\n")
+
+    // Insert right before the first real token of the enclosing declaration
+    // (after its leading trivia). The trivia — typically a leading newline and
+    // the declaration's indentation — becomes the struct's own prefix, so the
+    // struct text itself omits an explicit leading indent but includes a
+    // trailing blank line and indentation to restore spacing before the
+    // original declaration.
+    let insertPos = enclosingDecl.positionAfterSkippingLeadingTrivia
+    let structText = "struct \(structName) {\n\(fields)\n\(baseIndent)}\n\n\(baseIndent)"
```
ahoppen, line 81: This would  read nicer as a multi-line string literal

Sources/SwiftSyntaxCodeActions/ExtractTupleIntoStruct.swift
```diff
@@ -0,0 +1,199 @@
...
+// MARK: - Locating the insertion point
+
+/// Walks up the syntax tree to find the nearest `DeclSyntax` that is a direct
+/// child of a `CodeBlockItemSyntax` or `MemberBlockItemSyntax`.
+///
+/// That structural property means a new sibling declaration can be inserted
+/// immediately before it, either at the top level of a function body or as a
+/// nested member of a type.
+private func findEnclosingInsertionPoint(of node: Syntax) -> DeclSyntax? {
+  var current: Syntax? = node
```
ahoppen, line 135: Can’t you use `findParentOfSelf` here instead of walking the tree manually?

ahoppen (commented):
Some initial comments. Could you also add test cases for this?

PR #2720, opened 2026-07-14
Fix issues causing `Remove Unused Imports` to not show up in VS Code

ahoppen:
We had two issues here:
1. VS Code does not show code actions with kind `sourceOrganizeImports` in the lightbulb menu. Change it to `empty` so that it does show up.
2. We didn’t declare the command in `SwiftLanguageService`, so VS Code didn’t know to invoke SourceKit-LSP to perform the `remove.unused.imports.command`

### DIFF
--- a/Sources/SwiftLanguageService/RemoveUnusedImports.swift
+++ b/Sources/SwiftLanguageService/RemoveUnusedImports.swift
@@ -96,7 +96,7 @@ extension SwiftLanguageService {
     return [
       CodeAction(
         title: command.title,
-        kind: .sourceOrganizeImports,
+        kind: .refactor,
         diagnostics: nil,
         edit: nil,
         command: command.asCommand()
--- a/Sources/SwiftLanguageService/SwiftCommand.swift
+++ b/Sources/SwiftLanguageService/SwiftCommand.swift
@@ -51,6 +51,7 @@ extension SwiftLanguageService {
     [
       SemanticRefactorCommand.self,
       ExpandMacroCommand.self,
+      RemoveUnusedImportsCommand.self,
     ].map { (command: any SwiftCommand.Type) in
       command.identifier
     }


ahoppen:
@swift-ci Please test

rintaro:
How about providing `RemoveUnusedImportsCommand` regardless of the `scope` and leaving the kind as `sourceOrganizeImports`. Because that's what this kind is for, right?

ahoppen:
I haven’t been able to figure out when VS Code shows `sourceOrganizeImports` code actions. Do you know?

rintaro:
I looked into this and tested locally:

* **Command Palette `>Organize Imports`** - for this to show up, the server needs to advertise `source.organizeImports` explicitly in the server capability `codeActionProvider.codeActionKinds`. When executing it, VS Code requests the code actions with: `"only": ["source.organizeImports"]"`. Interestingly, this didn't surface "Remove Unused Imports", it was just executed. I haven't tried what happens if we return multiple `source.organizeImports` actions.
* **Context Menu > Source Actions...** - for this, the server needs to advertise `source` explicitly in the capability, Also the code action is requested with `"only": ["source"]"`. This shows the sub menu that lists "Remove Unused Imports"

So to support this, we need to advertise  `source.organizeImports` (this covers `source` as well) and handle `only` for `source.organizeImports` AND `source`.

ahoppen:
Oh, interesting. Thanks for investigating this, Rintaro. 

I just thought about this a bit more and I think that `refactor` is actually the correct code action kind we want here:

- Organize imports: I would imagine that this is mostly about sorting imports (organizing doesn’t imply removal to me) and since `Remove Unused Imports` doesn’t sort the imports, I don’t think this matches
- General `source` actions: The LSP spec describes them as *Source code actions apply to the entire file.* Since this code action only applies to imports (which in >99.9% of all cases for one coherent block at the source of the start of the file), I think doesn’t make sense to classify this code action as the `source` kind. 


So, I would continue with the implementation as it is in this PR. What do you think?

rintaro:
> Organize imports: I would imagine that this is mostly about sorting imports (organizing doesn’t imply removal to me) and since Remove Unused Imports doesn’t sort the imports, I don’t think this matches

THE standard implementation [does remove unused imports](https://code.visualstudio.com/docs/typescript/typescript-refactoring#_organize-imports), fwiw. But it's true it's mainly for sorting them.

> General source actions: The LSP spec describes them as Source code actions apply to the entire file. Since this code action only applies to imports (which in >99.9% of all cases for one coherent block at the source of the start of the file), I think doesn’t make sense to classify this code action as the source kind.

But `source.organizeImport` IS a `source` action.

---

That said, I'm totally fine with going with `refactor`.

> So, I would continue with the implementation as it is in this PR. What do you think?

Why `.empty` not `.refactor`?

ahoppen:
> Why `.empty` not `.refactor`?

I felt like this was a bit of its own kind of code action, more like simplification than refactoring. But thinking about this a bit more, I don’t feel like my argument holds up. Changed it to `refactor` now.

ahoppen:
@swift-ci Please test

rintaro (approved):

rintaro:
@swift-ci Please test Windows

PR #2722, opened 2026-07-15
Fix completion replacement range extending into the next token

### LINKED ISSUE #2721: No code completion results displayed in VS Code if code completion is not at the end of a function
For example, when completing in the following after `ba`, VS Code does not show any code completion results. As far as I can tell, this is due to the end line for being off by one after the introduction of `InsertReplaceEdit`

```swift
struct Foo {
    let bar: Int
}

func test() {
    let x = Foo(bar: 1)
    x.ba|
    print("abc")
}
```

Steffeeen:
Fixes #2721

`findEndOfIdentifier` could incorrectly return the end position of the next token, if that token was also an identifier or keyword.

### DIFF
--- a/Sources/SwiftLanguageService/CodeCompletionContext.swift
+++ b/Sources/SwiftLanguageService/CodeCompletionContext.swift
@@ -40,9 +40,10 @@ private func findEndOfIdentifier(
   snapshot: DocumentSnapshot,
   tree: SourceFileSyntax
 ) -> Position? {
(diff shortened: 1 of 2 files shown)


Steffeeen:
@swift-ci Please test

Steffeeen:
@swift-ci Please test linux

ahoppen (approved):
Perfect 👍🏽 Thanks for the quick fix

PR #2726, opened 2026-07-20
Fix textDocument/implementation for protocol requirements

### LINKED ISSUE #1600: `textDocument/implementation` on a protocol requirement returns locations at which conformance to that protocol is stated instead of their implementations
Consider doing an implementation request at 1️⃣ in the following.

```swift
protocol MyProto {
  func 1️⃣myFunc()
}

class ClassA: MyProto {
  func 2️⃣myFunc() {}
}
class ClassB: MyProto {
  func 3️⃣myFunc() {}
}
class 4️⃣ClassBSubA: ClassB {}
class 5️⃣ClassBSubB: ClassB {}

class RetroactiveConformanceClassWithMyFuncInClassDecl {
  func 6️⃣myFunc() { }
}
extension 7️⃣RetroactiveConformanceClassWithMyFuncInClassDecl: MyProto {}

class RetroactiveConformanceClassWithMyFuncInExtension {}
extension RetroactiveConformanceClassWithMyFuncInExtension: MyProto {
  func 8️⃣myFunc() { }
}
```

Currently, we report 2, 3, 4, 5, 7, 8

We should not be reporting locations 4, 5 and 7 because they don't actually contain myFunc. We should, however, be reporting location 6.

AnkanMisra:
Fixes #1600

- Return actual method implementations instead of subclass and conformance locations
- Preserve C++ declarations and out-of-line definitions that share a USR
- Also includes Swift and C++ regression tests

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
     guard let index = await workspaceForDocument(uri: req.textDocument.uri)?.index(checkedFor: .deletedFiles) else {
       return nil
     }
-    let locations = try symbols.flatMap { (symbol) -> [Location] in
-      guard let usr = symbol.usr else { return [] }
-      var occurrences = try index.occurrences(ofUSR: usr, roles: .baseOf)
-      if occurrences.isEmpty {
-        occurrences = try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
-      }
 
-      return occurrences.compactMap { $0.location.lspLocation }
+    // Preserve baseOf-first dispatch used for type hierarchies (subclasses / protocol conformers).
+    // When baseOf is empty (typical for methods), keep explicit overrideOf definition/declaration
+    // locations and remap only pure-implicit sites to their primary definition (e.g. protocol
+    // requirement satisfied in a type body with a separate conformance extension).
+    // See https://github.com/swiftlang/sourcekit-lsp/issues/1600
+    var candidateLocations: [Location] = []
+    var fallbackOccurrences: [SymbolOccurrence] = []
+
+    for symbol in symbols {
+      guard let usr = symbol.usr else { continue }
+      let baseOccurrences = try index.occurrences(ofUSR: usr, roles: .baseOf)
+      if !baseOccurrences.isEmpty {
+        candidateLocations += baseOccurrences.compactMap { $0.location.lspLocation }
+      } else {
+        fallbackOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
+      }
     }
+
+    // Deduplicate USRs only for pure-implicit occurrences that need primary-definition lookup.
+    var implicitUSRsNeedingLookup = Set<String>()
+    var uniqueImplicitUSRsInOrder: [String] = []
+
+    for occurrence in fallbackOccurrences {
+      let isExplicitDefinitionOrDeclaration =
+        occurrence.roles.contains(.definition) || occurrence.roles.contains(.declaration)
+      if isExplicitDefinitionOrDeclaration {
+        // Preserve every explicit declaration/definition location (important for C++ where
+        // in-class declaration and out-of-line definition share a USR).
+        if let location = occurrence.location.lspLocation {
+          candidateLocations.append(location)
+        }
+      } else if occurrence.roles.contains(.implicit) {
+        let implUSR = occurrence.symbol.usr
+        if implicitUSRsNeedingLookup.insert(implUSR).inserted {
+          uniqueImplicitUSRsInOrder.append(implUSR)
+        }
+      }
+      // Other non-explicit, non-implicit roles: ignore.
+    }
+
+    for implUSR in uniqueImplicitUSRsInOrder {
+      if let primary = try index.primaryDefinitionOrDeclarationOccurrence(ofUSR: implUSR),
+        let location = primary.location.lspLocation
+      {
+        candidateLocations.append(location)
+      }
+      // Pure implicit with no resolvable primary: drop (avoids subclass / conformance sites).
+    }
+
     let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
-    let remappedLocations = locations.adjusted(for: copiedFileMap)
-    return .locations(remappedLocations.sorted())
+    let remappedLocations = candidateLocations.adjusted(for: copiedFileMap)
+    return .locations(remappedLocations.unique.sorted())
   }
 
   func references(
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2228,36 +2228,32 @@ extension SourceKitLSPServer {
       return nil
     }
 
-    // Preserve baseOf-first dispatch used for type hierarchies (subclasses / protocol conformers).
-    // When baseOf is empty (typical for methods), keep explicit overrideOf definition/declaration
-    // locations and remap only pure-implicit sites to their primary definition (e.g. protocol
-    // requirement satisfied in a type body with a separate conformance extension).
+    // Report baseOf locations used for type hierarchies (subclasses / protocol conformers). For overrideOf
+    // occurrences, keep explicit definition/declaration locations and remap only pure-implicit sites to their primary
+    // definition (e.g. a protocol requirement satisfied in a type body with a separate conformance extension).
     // See https://github.com/swiftlang/sourcekit-lsp/issues/1600
-    var candidateLocations: [Location] = []
-    var fallbackOccurrences: [SymbolOccurrence] = []
+    var resultLocations: [Location] = []
+    var overrideOccurrences: [SymbolOccurrence] = []
 
     for symbol in symbols {
       guard let usr = symbol.usr else { continue }
       let baseOccurrences = try index.occurrences(ofUSR: usr, roles: .baseOf)
-      if !baseOccurrences.isEmpty {
-        candidateLocations += baseOccurrences.compactMap { $0.location.lspLocation }
-      } else {
-        fallbackOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
-      }
+      resultLocations += baseOccurrences.compactMap { $0.location.lspLocation }
+      overrideOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
     }
 
     // Deduplicate USRs only for pure-implicit occurrences that need primary-definition lookup.
     var implicitUSRsNeedingLookup = Set<String>()
     var uniqueImplicitUSRsInOrder: [String] = []
 
-    for occurrence in fallbackOccurrences {
+    for occurrence in overrideOccurrences {
       let isExplicitDefinitionOrDeclaration =
         occurrence.roles.contains(.definition) || occurrence.roles.contains(.declaration)
       if isExplicitDefinitionOrDeclaration {
         // Preserve every explicit declaration/definition location (important for C++ where
         // in-class declaration and out-of-line definition share a USR).
         if let location = occurrence.location.lspLocation {
-          candidateLocations.append(location)
+          resultLocations.append(location)
         }
       } else if occurrence.roles.contains(.implicit) {
         let implUSR = occurrence.symbol.usr
@@ -2272,13 +2268,13 @@ extension SourceKitLSPServer {
       if let primary = try index.primaryDefinitionOrDeclarationOccurrence(ofUSR: implUSR),
         let location = primary.location.lspLocation
       {
-        candidateLocations.append(location)
+        resultLocations.append(location)
       }
       // Pure implicit with no resolvable primary: drop (avoids subclass / conformance sites).
     }
 
     let copiedFileMap = await workspace.buildServerManager.cachedCopiedFileMap
-    let remappedLocations = candidateLocations.adjusted(for: copiedFileMap)
+    let remappedLocations = resultLocations.adjusted(for: copiedFileMap)
     return .locations(remappedLocations.unique.sorted())
   }
 
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -2243,6 +2243,19 @@ extension SourceKitLSPServer {
       overrideOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
     }
 
+    func preferredImplicitOccurrence(_ lhs: SymbolOccurrence, _ rhs: SymbolOccurrence) -> SymbolOccurrence {
+      let lhsIsContainedByExtension = lhs.relations.contains {
+        $0.roles.contains(.containedBy) && $0.symbol.kind == .extension
+      }
+      let rhsIsContainedByExtension = rhs.relations.contains {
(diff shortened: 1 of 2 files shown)


Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
+    var candidateLocations: [Location] = []
+    var fallbackOccurrences: [SymbolOccurrence] = []
+
+    for symbol in symbols {
+      guard let usr = symbol.usr else { continue }
+      let baseOccurrences = try index.occurrences(ofUSR: usr, roles: .baseOf)
+      if !baseOccurrences.isEmpty {
+        candidateLocations += baseOccurrences.compactMap { $0.location.lspLocation }
+      } else {
+        fallbackOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
```
ahoppen: Should we name this variable `overrideOccurrences` because that’s what it is?

Actually, I’m also just thinking why we only want to report these if the `baseOf` locations are empty. Maybe we should always be iterating through both.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
-      var occurrences = try index.occurrences(ofUSR: usr, roles: .baseOf)
-      if occurrences.isEmpty {
-        occurrences = try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
+
+    // Preserve baseOf-first dispatch used for type hierarchies (subclasses / protocol conformers).
+    // When baseOf is empty (typical for methods), keep explicit overrideOf definition/declaration
+    // locations and remap only pure-implicit sites to their primary definition (e.g. protocol
+    // requirement satisfied in a type body with a separate conformance extension).
+    // See https://github.com/swiftlang/sourcekit-lsp/issues/1600
+    var candidateLocations: [Location] = []
```
ahoppen: I don’t think that these are just candidates. They are the locations we report, so `resultLocations` or something like that would fit better, right?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
+      if !baseOccurrences.isEmpty {
+        candidateLocations += baseOccurrences.compactMap { $0.location.lspLocation }
+      } else {
+        fallbackOccurrences += try index.occurrences(relatedToUSR: usr, roles: .overrideOf)
       }
+    }
+
+    // Deduplicate USRs only for pure-implicit occurrences that need primary-definition lookup.
+    var implicitUSRsNeedingLookup = Set<String>()
+    var uniqueImplicitUSRsInOrder: [String] = []
```
ahoppen: Why is in-order important here? Don’t we sort them in the end anyway?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
 
-      return occurrences.compactMap { $0.location.lspLocation }
+    for occurrence in fallbackOccurrences {
+      let isExplicitDefinitionOrDeclaration =
+        occurrence.roles.contains(.definition) || occurrence.roles.contains(.declaration)
+      if isExplicitDefinitionOrDeclaration {
```
ahoppen: You don‘t actually check if it’s explicit here. Not an issue, just a misleading variable name. I’d just inline the check anyway.


```suggestion
      if occurrence.roles.contains(.definition) || occurrence.roles.contains(.declaration) {
```

Tests/SourceKitLSPTests/ImplementationTests.swift
```diff
@@ -249,9 +249,10 @@ final class ImplementationTests: SourceKitLSPTestCase {
   }
 
   func testOverrideProtocolFunc() async throws {
-    // TODO: We should not be reporting locations 4, 5 and 7 because they don't actually contain myFunc.
-    // We should, however, be reporting location 6. (https://github.com/swiftlang/sourcekit-lsp/issues/1600)
-
+    // https://github.com/swiftlang/sourcekit-lsp/issues/1600
+    // Report actual method implementations only: not subclass names (4, 5) or
+    // conformance-only extensions (7). Retroactive methods declared in the type
+    // body (6) must still be found via the conformance extension's index relation.
```
ahoppen: I think we can drop this comment now. The test should be self-explanatory on what it does.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
+        // Preserve every explicit declaration/definition location (important for C++ where
+        // in-class declaration and out-of-line definition share a USR).
+        if let location = occurrence.location.lspLocation {
+          candidateLocations.append(location)
+        }
+      } else if occurrence.roles.contains(.implicit) {
+        let implUSR = occurrence.symbol.usr
```
ahoppen: `impl` reads as short for “implementation” to me, which is confusing in a function that’s all about implementations. Could we spell it `implicitUSR`, matching `implicitUSRsNeedingLookup` right above?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,59 @@ extension SourceKitLSPServer {
...
+        if implicitUSRsNeedingLookup.insert(implUSR).inserted {
+          uniqueImplicitUSRsInOrder.append(implUSR)
+        }
+      }
+      // Other non-explicit, non-implicit roles: ignore.
```
ahoppen: Do you know what falls into the neither branch here? If you don’t think this can be hit, should we emit a log message to ease the ability to debug this, should we hit it?

ahoppen (commented):
Thanks for fixing this @AnkanMisra. Overall the implementation looks great, I just have a small comments inline.

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: `overrideOccurrences` seems more clear, ill also collect both `baseOf` and `overrideOf` occurences here.

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: make sense, ill rename it to `resultLocations`

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: the final results are sorted anyway, ill just simplify this to just use the set

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: ill inline the check

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: sure, `implicitUSR` is much clear here

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: i am not aware of a case that hits this branch, but logging it would help if one comes up, ill add that 👍🏼

AnkanMisra:
done, the implementation now collects both `baseOf` and `overrideOf` occurrences for every symbol, keeps explicit declaration/definition locations, and resolves implicit occurrences through their primary definition

since the final locations are sorted, i dropped the extra ordered USR array. also cleaned up the naming, added a fault log for unexpected roles, and removed the redundant test comment

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,52 @@ extension SourceKitLSPServer {
...
+    }
 
-      return occurrences.compactMap { $0.location.lspLocation }
+    for implicitUSR in implicitUSRsNeedingLookup {
+      if let primary = try index.primaryDefinitionOrDeclarationOccurrence(ofUSR: implicitUSR),
+        let location = primary.location.lspLocation
+      {
+        resultLocations.append(location)
+      }
+      // Pure implicit with no resolvable primary: drop (avoids subclass / conformance sites).
```
ahoppen: What do you think to report the implicit location instead of implicitly dropping it? At least that gives the user a sense that there is something somewhere and usually the implicit location is at least somewhat related to what they might be interested in.

ahoppen (commented):
Nice. Reads ver nice and straightforward now 🙂 

One more small comment left, otherwise good to mere.

Sources/SourceKitLSP/SourceKitLSPServer.swift
AnkanMisra: yeah, that makes sense, i dropped the implicit location to avoid bringing back the misleading results from the issue, but i was treating “no exact location” the same as “bad location”

using `primaryDefinitionOrDeclarationOccurrence` when available and falling back to the `.implicit` site otherwise seems to do better 👀

one thing im curious about: if the same `USR` has multiple implicit occurrences, would you expect all of those fallback locations to be reported and let `.unique.sorted()` handle them, or should we keep a single representative?

ahoppen: Reporting it once should be sufficient. The user won’t be able to see any of the supplementary information like roles, so reporting it multiple times doesn’t give any additional value.

AnkanMisra:
So i kept one deterministic implicit occurrence per USR and only use it when `primaryDefinitionOrDeclarationOccurrence` has no indexed result, basically the real definition still wins while the fallback is reported once

and i added a binary only module regression with three implicit sites sharing the same USR
it catches both dropping the fallback and returning every site

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -2227,18 +2227,52 @@ extension SourceKitLSPServer {
...
+    for occurrence in overrideOccurrences {
+      if occurrence.roles.contains(.definition) || occurrence.roles.contains(.declaration) {
+        // Preserve every explicit declaration/definition location (important for C++ where
+        // in-class declaration and out-of-line definition share a USR).
+        if let location = occurrence.location.lspLocation {
+          resultLocations.append(location)
+        }
+      } else if occurrence.roles.contains(.implicit) {
+        let usr = occurrence.symbol.usr
+        implicitOccurrencesNeedingLookup[usr] = min(implicitOccurrencesNeedingLookup[usr] ?? occurrence, occurrence)
```
AnkanMisra, line 2258: i noticed this depends on source order,like if i move the conformance extension below the subclasses,`min` picks the first subclass instead

so would it make more sense to prefer the extension here,or is either location fine?

ahoppen: Do you have an example where this manifests? In any case, it would be good to have a test case for this and I agree with you that we should have consistent behavior independent of source order. 

Also would like to hear your opinion on which location would be best to report.

AnkanMisra: yeah, i can reproduce it with the [`testImplicitImplementationWithoutIndexedDefinitionIsReportedOnce`](https://github.com/swiftlang/sourcekit-lsp/blob/7f9999e4de71132154be5ff4f1a7c587148ad5f3/Tests/SourceKitLSPTests/ImplementationTests.swift#L284-L347) case

if i move the extension below the two subclasses then the implementation jumps to `SubclassA` while with the extension first it points to the extension

locally, the reordered test returns line 7 (`SubclassA`) instead of line 9 (the extension),so moving three lines around currently changes where we send the user, which feels a little too magical i assume?😅

i think the extension is more useful here since thats where the conformance is added and the subclasses just inherit it

i can keep the extension last in the test, prefer it when its available and use the containing symbol `USR` to keep the remaining choice stable

does that sound reasonable?

ahoppen: I agree that reporting the extension makes more sense. What I would do, is more the extension between the two subclass definitions and still check that it gets reported. That should capture all cases going forward. Maybe also add a short comment summarizing what you said above so that we don’t wonder why the `extension` is between the subclasses in the future.

Also: Thanks for noticing, reducing and presenting this issue this well. 🙏🏽

AnkanMisra: got it about the extension between the subclasses 👍

AnkanMisra, line 2258: one thing i wanted to check before changing is `.containedBy` with `symbol.kind == .extension` the right signal to identify the conformance extension here?

AnkanMisra: i found [`.containedBy` used for containers](https://github.com/swiftlang/sourcekit-lsp/blob/7f9999e4de71132154be5ff4f1a7c587148ad5f3/Sources/SourceKitLSP/SourceKitLSPServer.swift#L2404-L2421) and [`.extension` checked on related symbols](https://github.com/swiftlang/sourcekit-lsp/blob/7f9999e4de71132154be5ff4f1a7c587148ad5f3/Sources/SourceKitLSP/TestDiscovery.swift#L111-L120) elsewhere, so this seems consistentm with the existing codebase,just wanted to confirm the relation is reliable for implicit `overrideOf` occurrences

ahoppen: I’m not entirely sure what you mean here. Currently, there’s no `containedBy` check here. Are you proposing to change this lookup to something else? If so, to what are you proposing changing it?

AnkanMisra: yeah sorry i explained that badly 😭, i meant changing [the fallback on line 2258](https://github.com/swiftlang/sourcekit-lsp/blob/7f9999e4de71132154be5ff4f1a7c587148ad5f3/Sources/SourceKitLSP/SourceKitLSPServer.swift#L2258) to prefer the implicit occurrence contained by the extension and only use `min` if there isnt one

the primary definition lookup stays the same and ill move the extension between the subclasses in the test like you suggested

AnkanMisra:
updated the implicit implementation fallback to prefer an occurrence contained by an extension

and i also moved the extension between both subclasses so the test can’t pass by picking the first or last result

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I think that makes sense. Could you just add a short doc comment explaining why we want to prefer implicit declarations in extensions, maybe with a short example?

AnkanMisra: i have added the doc comment above `preferredImplicitOccurrence` in [694a4439](https://github.com/swiftlang/sourcekit-lsp/commit/694a4439fab0fee62057c396b3435ba06db0afc8)

it explains that subclasses can inherit a conformance introduced in an extension and produce implicit occurrences for the same requirement,so the extension is the better fallback when the index has no definition or declaration

does that example cover what you had in mind?

ahoppen (approved):
Looks good. Thanks for working through all the subtleties here, @AnkanMisra 🙏🏽

AnkanMisra:
thanks for sticking with me through it, i learned a lot about how the index represents these implicit occurrences 🙌

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2727, opened 2026-07-21
Honor the build server's canonical-target preference when selecting a target

rintaro:
Depends on https://github.com/swiftlang/swift-tools-protocols/pull/74

For a file that belongs to multiple targets, the build server can now rank the candidates via `SourceKitSourceItemData.canonicalTargetPreference` (lower is preferred). `BuildServerManager.canonicalTarget(for:)` picks the target with the lowest preference; targets without a preference rank after those with one, and ties fall back to the existing deterministic ordering by URI.


rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/74
@swift-ci Please test

PR #2729, opened 2026-07-21
Add diagnostics for missing/broken Symbol Links in DocC comments, Tutorial and Markdown files 

Padmashree06:
This PR adds diagnostics for unresolved symbol links in DocC comments, Markdown, and Tutorial files, using both push and pull diagnostics. For `Swift` files, where both `SwiftLanguageService` and `DocumentationLanguageService` can report diagnostics, the responses from both are merged into a single combined report.

### Changes 
- Adds both push and pull diagnostics in `DocumentationLanguageService` for unresolved symbol references. 
- For `.swift` files, merges diagnostics from both `SwiftLanguageService` and `DocumentationLanguageService` into a single response (generalizes `handleRequest` to support merging responses across language services via an optional `combine` parameter so that the existing callers are unaffected since none of them pass `combine`). 
- Falls back  to push diagnostics only when the pull diagnostics is not supported by the client.
- For  `.swift` files the push notifications of both `SwiftLanguageService` and `DoucumenattionLanguageService` are merged and sent to the client.  
- Symbol graph is considered as the source of truth to resolve the Symbol links. If the symbol graph is not available (module hasn't built or failed to build) no diagnostics are reported. 

### User Impact
Users now see diagnostics (red squiggles) for unresolved symbol links in DocC comments, Markdown and Tutorial files, reported via pull diagnostics where supported and push diagnostics otherwise.

### DIFF
--- a/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
+++ b/Sources/DocumentationLanguageService/DocCSymbolInformation.swift
@@ -38,10 +38,11 @@ struct DocCSymbolInformation {
   }
 
   func matches(_ link: DocCSymbolLink) -> Bool {
-    guard link.components.count == components.count else {
+    guard link.components.count <= components.count else {
       return false
     }
-    return zip(link.components, components).allSatisfy { linkComponent, symbolComponent in
+    let matchingSuffix = components.suffix(link.components.count)
+    return zip(link.components, matchingSuffix).allSatisfy { linkComponent, symbolComponent in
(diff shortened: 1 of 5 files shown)


ahoppen (commented):
This is pretty much blocked on the same underlying issues of generating the symbol graph in the background as https://github.com/swiftlang/sourcekit-lsp/pull/2698, right?

snprajwal:
Yes, this will also get refactored, so it might be good to hold off on review until that happens (at least for the bits that use the existing symbol graph logic).

PR #2730, opened 2026-07-21
Emit a stable generated-interface URI for SDK workspace symbols

rintaro:
`workspace/symbolInfo` returned SDK/stdlib symbols with a deferred `file://<interface>?module=<name>` location that `workspaceSymbol/resolve` *replaced* with a `sourcekit-lsp://` reference-document URI. Replacing the URI on resolve requires the client to advertise `location.uri` resolution, so clients that resolve only `location.range` (e.g. VS Code) couldn't use it.

This emits the `sourcekit-lsp://generated-swift-interface` URI up front instead. It's self-describing, shared by every symbol of the module, and `workspaceSymbol/resolve` now leaves it untouched and only fills in the range, extending the feature to clients with `location.range`-only resolution and avoiding the percent-encoding issue via the custom scheme.

The interface's main file (for build settings) is resolved via `mainFiles(containing:)` and memoized per interface, so the number of `mainFiles(containing:)` calls is bounded by the distinct system modules imported in the workspace, not the number of symbols in the response.

The `WorkspaceSymbol`'s `data` now also carries the `.swiftinterface`/`.swiftmodule` path and the fully-qualified module name, so clients can render the candidate without parsing the opaque location URI.

The path is gated on the client supporting both generated-interface reference documents and `workspaceSymbol/resolve`; otherwise it falls back to a plain `file://` interface location. The server now advertises `workspaceSymbolProvider` with `resolveProvider: true`, and `clientSupportsWorkspaceSymbolResolve` is relaxed to accept `location` or `location.range`. Only `workspace/symbolInfo` is affected — `workspace/symbol` filters out system symbols.

### DIFF
--- a/Contributor Documentation/Open Quickly.md
+++ b/Contributor Documentation/Open Quickly.md
@@ -44,13 +44,14 @@ The shape of each result item depends on the symbol's origin:
   }
 ```
 
-**SDK/stdlib symbols** — returned as `WorkspaceSymbol` with `location: .uri(file:// URL)` (no range) pointing to the
-`.swiftinterface` or `.swiftmodule` file from the index record, when the client advertises
-`workspace.symbol.resolveSupport.properties` containing `"location"` or both `"location.uri"` and `"location.range"`.
-The fully-qualified module name (e.g. `Swift.String`) is appended as a `?module=` query parameter on the location URL
-so clients can derive a display path without inspecting the `data` dictionary. The symbol's USR is stored in the
-`data` dictionary. The client must call `workspaceSymbol/resolve` to obtain the exact location within the generated
-interface.
+**SDK/stdlib symbols** — returned as `WorkspaceSymbol` with `location: .uri(sourcekit-lsp:// URL)` (no range) and a
+`data` payload. The `data` is a `SourceKitWorkspaceSymbolData` that clients decode via `WorkspaceSymbol.sourceKitData`,
+carrying the symbol's USR, the `.swiftinterface`/`.swiftmodule` path (`interfaceURI`), and the fully-qualified module
+name (`moduleName`, e.g. `Swift.String`) that clients can use to render the candidate. The URI is a
+fully-parameterized `sourcekit-lsp://generated-swift-interface/` reference-document URL: `workspaceSymbol/resolve`
+fills in its range and `workspace/getReferenceDocument` returns its content. This form is emitted when the client
+advertises **both** generated-interface reference documents (`GetReferenceDocumentRequest`) and
+`workspaceSymbol/resolve` support (see [Client Capabilities](#client-capabilities)).
 
 ```
 → WorkspaceSymbolInfoRequest { names: ["String"] }
@@ -59,37 +60,39 @@ interface.
       WorkspaceSymbol {
         name: "String",
         kind: .struct,
-        location: { uri: "file:///path/to/Swift.swiftmodule/arm64-apple-macosx.swiftinterface?module=Swift.String" },
-        data: { "usr": "s:SS" }
+        location: { uri: "sourcekit-lsp://generated-swift-interface/Swift.String.swiftinterface?moduleName=Swift&groupName=String&sourcekitdDocument=Swift.String.12345678&buildSettingsFrom=file:///path/to/Sources/main.swift" },
+        data: {
+          "usr": "s:SS",
+          "interfaceURI": "file:///path/to/Swift.swiftmodule/arm64-apple-macosx.swiftinterface",
+          "moduleName": "Swift.String"
+        }
       }
     ]
   }
 ```
 
-Without that capability, the raw `file://` URI of the `.swiftinterface` or `.swiftmodule` file from the index record
-is returned as `SymbolInformation` instead.
+Without those capabilities, the raw `file://` URI of the `.swiftinterface` or `.swiftmodule` file from the index
+record is returned as `SymbolInformation` instead.
+
+> [!NOTE]
+> The `data` payload is a typed SourceKit-LSP API (`SourceKitWorkspaceSymbolData`): clients should decode it via
+> `WorkspaceSymbol.sourceKitData` rather than reading raw JSON keys.
 
 The response is a flat array of `WorkspaceSymbolItem` values. Each item carries the symbol name in its `name` field.
 
-### `workspaceSymbol/resolve` — Location Resolution
+### `workspaceSymbol/resolve` — Range Resolution
 
-Resolves the lazy location of a `WorkspaceSymbol` returned by `sourcekit/workspace/symbolInfo`. The server parses
-`moduleName` and `groupName` from the `?module=` query parameter of the location URL, reads `usr` from the `data`
-dictionary, opens the generated Swift interface for the symbol's module, finds the symbol's position using the USR,
-and returns the same symbol with `location` replaced by a full `Location` (URI + range).
-
-> [!NOTE]
-> The LSP 3.17 spec defines `workspaceSymbol/resolve` as filling in a missing `range` for a URI-only location — it
-> does not anticipate replacing `location.uri` itself. SourceKit-LSP goes beyond the spec here: it substitutes the
-> `file://` module-file URL with a `sourcekit-lsp://` URI that the client can pass to `workspace/getReferenceDocument`
-> to retrieve the generated interface content.
+Fills in the range of a `WorkspaceSymbol` returned by `sourcekit/workspace/symbolInfo`. The server parses
+`moduleName`, `groupName`, and `buildSettingsFrom` from the location URI, reads `usr` from `sourceKitData`,
+opens the generated Swift interface for the symbol's module, finds the symbol's position using the USR, and returns
+the same symbol with `location` upgraded from a URI-only location to a full `Location` (URI + range).
 
 ```
 → WorkspaceSymbolResolveRequest {
     workspaceSymbol: WorkspaceSymbol {
       name: "String",
       kind: .struct,
-      location: { uri: "file:///path/to/Swift.swiftmodule/arm64-apple-macosx.swiftinterface?module=Swift.String" },
+      location: { uri: "sourcekit-lsp://generated-swift-interface/Swift.String.swiftinterface?moduleName=Swift&groupName=String&sourcekitdDocument=Swift.String.12345678&buildSettingsFrom=file:///path/to/Sources/main.swift" },
       data: { "usr": "s:SS" }
     }
   }
@@ -103,14 +106,10 @@ and returns the same symbol with `location` replaced by a full `Location` (URI +
   }
 ```
 
-The resolved URI is a fully-parameterized `sourcekit-lsp://generated-swift-interface/` URL containing
-`sourcekitdDocument` and `buildSettingsFrom` derived from a real source file in the workspace via
-`mainFiles(containing:)`.
-
-The client must treat the resolved `sourcekit-lsp://` URI as **opaque** — it should not parse or extract information
-from the query parameters. The path component (e.g. `Swift.String.swiftinterface`) may be used as the editor tab
-title. The URI is otherwise only valid as an input to `workspace/getReferenceDocument`; its query parameter structure
-is an implementation detail subject to change.
+The client must treat the `sourcekit-lsp://` URI as **opaque** — it should not parse or extract information from the
+query parameters. The path component (e.g. `Swift.String.swiftinterface`) may be used as the editor tab title. The URI
+is otherwise only valid as an input to `workspace/getReferenceDocument`; its query parameter structure is an
+implementation detail subject to change.
 
 ### `workspace/getReferenceDocument` — Content Retrieval
 
@@ -124,8 +123,8 @@ provider — it returns the document text and nothing else.
   }
 ```
 
-The URI passed here must be a fully resolved URI (with `sourcekitdDocument` set), as returned by
-`workspaceSymbol/resolve`.
+The URI passed here is the same one carried by the `WorkspaceSymbol` from `sourcekit/workspace/symbolInfo`; the client
+does not need to have resolved the range first.
 
 ## Workflow
 
@@ -140,13 +139,13 @@ Client                                            Server
   │── sourcekit/workspace/symbolInfo                 │
   │     {["String", "Stride", ...]} ────────────────▶│
   │◀─ [WorkspaceSymbol] ─────────────────────────────│
-  │     (location: "file://...?module=Swift.String") │
+  │     (location: "sourcekit-lsp://...")            │
   │                                                  │
   │  [user selects "String"]                         │
   │                                                  │
   │── workspaceSymbol/resolve ──────────────────────▶│
   │◀─ WorkspaceSymbol                                │
-  │     (url: "sourcekit-lsp://...", range: ...) ────│
+  │     (location: uri + range) ─────────────────────│
   │                                                  │
   │── workspace/getReferenceDocument ───────────────▶│
   │◀─ { content: "..." } ────────────────────────────│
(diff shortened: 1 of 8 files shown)


rintaro:
@swift-ci Please test

Sources/SourceKitLSP/MessageMetadata.swift
```diff
@@ -36,9 +36,21 @@ struct ResolveItemData: Codable, Hashable, LSPAnyCodable {
...
   var usr: String
 
-  init(usr: String) {
+  /// The `.swiftinterface`/`.swiftmodule` file the symbol is declared in. Some clients use this to render
+  /// the candidate in their UI without parsing the opaque location URI.
+  var interfaceURI: DocumentURI?
+
+  /// The fully-qualified module name recorded in the index (e.g. `Swift.String`). Some clients use this to
+  /// render the candidate in their UI without parsing the opaque location URI.
+  var moduleName: String?
```
rintaro: I didn't want to make `data` a parsable structure, but I think this is the least bad place to embed these informations.
Adding these to `WorkspaceSymbol` top-level properties feels risky at this point. If we really want that, we can do it later when we stabilize the feature more.

Sources/SourceKitLSP/CapabilityRegistry.swift
```diff
@@ -128,15 +128,16 @@ package final actor CapabilityRegistry {
...
   }
 
-  /// Whether the client supports `workspaceSymbol/resolve` for lazy location resolution.
-  /// Requires `workspace.symbol.resolveSupport.properties` to contain `"location"` or
-  /// both `"location.uri"` and `"location.range"`.
+  /// Whether the client supports `workspaceSymbol/resolve` for lazily filling in a symbol's range.
+  ///
+  /// The location's URI is always sent up front and is left unchanged by `workspaceSymbol/resolve`, which
+  /// only fills in the range. So it suffices for the client to resolve `location` or just `location.range`;
+  /// resolving `location.uri` is not required.
```
ahoppen: I’m not sure if explicitly spelling out what's not required makes too much sense here. If we start spelling out what’s not required in every doc comment, the documentation would explode.

Sources/SourceKitLSP/MessageMetadata.swift
```diff
@@ -36,9 +36,21 @@ struct ResolveItemData: Codable, Hashable, LSPAnyCodable {
...
+
+  init(usr: String, interfaceURI: DocumentURI? = nil, moduleName: String? = nil) {
```
ahoppen: Should we omit the default values here so callers can’t just forget to pass `interfaceURI` and `moduleName`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1644,6 +1640,49 @@ extension SourceKitLSPServer {
     )
   }
 
+  /// Split an index module name of the form `Module` or `Module.Group` into its module and optional group
+  /// components, on the first `.`.
+  private nonisolated static func splitModuleNameAndGroup(_ fullModuleName: String) -> (module: String, group: String?)
+  {
```
```suggestion
  private nonisolated static func splitModuleNameAndGroup(
    _ fullModuleName: String
  ) -> (module: String, group: String?) {
```

Contributor Documentation/Open Quickly.md
```diff
@@ -59,37 +60,53 @@ interface.
...
-> `file://` module-file URL with a `sourcekit-lsp://` URI that the client can pass to `workspace/getReferenceDocument`
-> to retrieve the generated interface content.
+Fills in the range of a `WorkspaceSymbol` returned by `sourcekit/workspace/symbolInfo`. The server parses
+`moduleName`, `groupName`, and `buildSettingsFrom` from the location URI, reads `usr` from the `data` dictionary,
+opens the generated Swift interface for the symbol's module, finds the symbol's position using the USR, and returns
+the same symbol with `location` upgraded from a URI-only location to a full `Location` (URI + range).
+
+The location URI is **unchanged** — resolve only adds the range. This matches the LSP 3.17 contract for
+`workspaceSymbol/resolve` (fill in a missing `range` for a URI-only location) and means the client only needs to
+advertise resolution of `location` or `location.range`; it does not need to advertise `location.uri`.
```
ahoppen: I don’t think this is worth explicitly spelling out. Not changing the URI is standard behavior in LSP.

Contributor Documentation/Open Quickly.md
```diff
@@ -158,90 +171,19 @@ Client                                            Server
...
-   `.first`
-4. Call `openGeneratedInterface(document: primaryFile, moduleName:, groupName:, symbolUSR:)`
-5. Return the symbol with `location` replaced by a full `Location` (resolved `sourcekit-lsp://` URI + range)
+## `sourcekit-lsp://` URI for SDK/stdlib Symbols
 
-## `sourcekit-lsp://` URI for Resolved Locations
-
-After `workspaceSymbol/resolve`, the location URI is a fully-parameterized `sourcekit-lsp://generated-swift-interface/`
-URL.
+The location URI for an SDK/stdlib `WorkspaceSymbol` is a fully-parameterized
```
ahoppen, line 170: Similar here. I don’t think we need to keep spelling out that URI is unchanged.

Contributor Documentation/Open Quickly.md
```diff
@@ -280,6 +225,20 @@ sourcekit-lsp://generated-swift-interface/Swift.String.swiftinterface
...
+
+The `WorkspaceSymbol`/`.uri` path in `sourcekit/workspace/symbolInfo` is gated on the client advertising **both** of
+the following. When either is missing, `sourcekit/workspace/symbolInfo` falls back to `SymbolInformation` with the raw
+`file://` URI from the index record.
+
+- `GetReferenceDocumentRequest` support (an experimental client capability) — signals that the client can open the
+  `sourcekit-lsp://` reference document via `workspace/getReferenceDocument`.
+- `ClientCapabilities.workspace.symbol.resolveSupport.properties` containing `"location"` or `"location.range"` (LSP
+  3.17) — signals that the client can call `workspaceSymbol/resolve` to obtain a range-bearing location. Resolving
+  `location.uri` is *not* required, because the URI is stable and only the range is filled in.
```
ahoppen: Same here, not worth selling out that URI is unchanged.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1576,13 +1576,19 @@ extension SourceKitLSPServer {
...
+  ///   interface location built from this main file's build settings, with the USR in `data`. Callers pass
+  ///   it only for SDK/stdlib symbols when the client supports reference documents and
+  ///   `workspaceSymbol/resolve`, so its presence alone gates that path. Otherwise a `SymbolInformation`
+  ///   with the plain `file://` location from the index is returned.
   private nonisolated func workspaceSymbolItem(
     for symbolOccurrence: SymbolOccurrence,
     in index: CheckedIndex,
     copiedFileMap: CopiedFileMap,
-    canUseWorkspaceSymbolResolve: Bool
+    generatedInterfaceMainFile: DocumentURI? = nil
```
ahoppen: Should we remove the default value so that callers can’t just forget to pass `generatedInterfaceMainFile`?

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1576,13 +1576,19 @@ extension SourceKitLSPServer {
...
-  /// Map a `SymbolOccurrence` from the index to a `WorkspaceSymbolItem` suitable for returning in a
-  /// `workspace/symbolInfo` response.
+  /// Map a `SymbolOccurrence` from the index to a `WorkspaceSymbolItem`, or `nil` if it has no
+  /// representable location.
+  ///
+  /// - Parameter generatedInterfaceMainFile: If non-`nil`, emit a stable `sourcekit-lsp://` generated-
+  ///   interface location built from this main file's build settings, with the USR in `data`. Callers pass
+  ///   it only for SDK/stdlib symbols when the client supports reference documents and
+  ///   `workspaceSymbol/resolve`, so its presence alone gates that path. Otherwise a `SymbolInformation`
+  ///   with the plain `file://` location from the index is returned.
```
ahoppen: It’s surprising to me that `generatedInterfaceMainFile` communicates client capabilities. I would prefer if we either checked whether the client supports reference documents and `workspaceSymbol/resolve` in here or explicitly pass it in as a parameter.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1644,6 +1640,49 @@ extension SourceKitLSPServer {
...
+    var mainFiles: [DocumentURI: DocumentURI] = [:]
+    for symbol in symbols {
+      let path = symbol.location.path
+      guard path.hasSuffix(".swiftinterface") || path.hasSuffix(".swiftmodule"),
+        let interfaceURI = symbol.location.uri,
+        mainFiles[interfaceURI] == nil
+      else {
+        continue
+      }
+      if Task.isCancelled { return [:] }
```
ahoppen: Should we `try Task.checkCancellation()` here to avoid callers unintentionally using an empty dictionary as a valid result if they don’t end up checking cancellation themselves? I don’t see a reason why this function wouldn’t be able to throw.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1595,38 +1601,28 @@ extension SourceKitLSPServer {
...
+      )
       return WorkspaceSymbolItem.workspaceSymbol(
         WorkspaceSymbol(
           name: symbolOccurrence.symbol.name,
           kind: symbolOccurrence.symbol.kind.asLspSymbolKind(),
           containerName: containerName,
-          location: .uri(.init(uri: DocumentURI(locationURL))),
-          data: data
+          location: .uri(.init(uri: try urlData.uri)),
+          data: data.encodeToLSPAny()
```
ahoppen, line 1627: Can we continue to pass `nil` here if the encoded data is empty? Just thinking that omitting `data: ` altogether will result in a smaller JSON response than having `data: {}`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1706,67 +1759,41 @@ extension SourceKitLSPServer {
 
   /// Handle a `workspaceSymbol/resolve` request.
   ///
-  /// If the symbol has a `location: .uri(moduleFileURL?module=...)` (as emitted by
-  /// `workspace/symbolInfo` for SDK/stdlib symbols), opens the generated Swift interface, resolves
-  /// the symbol position using `data["usr"]`, and returns the symbol with `location: .location(...)`.
+  /// If the symbol has a `location: .uri(sourcekit-lsp://generated-swift-interface?...)` (as emitted by
+  /// `workspace/symbol` and `workspace/symbolInfo` for SDK/stdlib symbols), opens the generated Swift
+  /// interface, resolves the symbol position using `data["usr"]`, and returns the symbol with the exact
+  /// range. The URI is kept unchanged so it stays stable for the client; only the range is filled in.
```
ahoppen: Same here, not sure if it’s worth spelling out that URI stays unchanged.

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1706,67 +1759,41 @@ extension SourceKitLSPServer {
...
-          groupName: groupName,
-          symbolUSR: usr
-        )
-      }
-      if let details {
-        symbol.location = .location(
-          Location(uri: details.uri, range: Range(details.position ?? Position(line: 0, utf16index: 0)))
-        )
-      }
+    // Treat an empty USR as absent so we don't run a position lookup that can't match.
```
ahoppen: Did you see this happen somewhere in practice?

Sources/SourceKitLSP/MessageMetadata.swift
ahoppen: I would prefer us to put the data where we want it long-term (in the `WorkspaceSymbol`) straight away because it will be annoying to move it later (or we may just forget). Also, this way we will have these fields documented in `LSP Extensions.md` and we pass less data back to LSP during a `workspaceSymbol/resolve` request. 

If we want to keep these properties flexible, we should just be hiding them behind an experimental feature.

ahoppen (commented):
I like the stable URI approach a lot better 🙂

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: We are checking the client capability whether to derive the `generatedInterfaceMainFile`. So checking it here again or passing-in is technically redundant. Since this is a private function, I don't think we want it just for the clarity.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: The only reason was that this is called inside `concurrentMap`.  But I'm going to wrap it with `orLog` in the caller.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I don't think so. This is just a defensive guard.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Previously, we passed `nil` iff `usr.isEmpty` but that shouldn't reallly happen. So I don't think it's worth to add explicit checking.

Sources/SourceKitLSP/MessageMetadata.swift
rintaro: Let's follow the `SourceKitSourceItemData` pattern: use `data` as the underlying storage, but expose it through a public typed property.
https://github.com/swiftlang/swift-tools-protocols/pull/75

> we pass less data back to LSP during a workspaceSymbol/resolve request

Clients request `workspaceSymbol/resolve` with the entire `WorkspaceSymbol` structure anyway, so moving things out of `data` doesn't change the message size.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/75
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/75
@swift-ci Please test Windows

Sources/SourceKitLSP/MessageMetadata.swift
ahoppen: Good with me 👍🏽

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: I would at least like to include some hint that the value of this significantly changes the structure of the returned workspace symbol and to indicate that a non-nil value must only be passed if the client supports `workspace/getReferenceDocument` and `workspaceSymbol/resolve`. 
 
What do you think about changing the `generatedInterfaceMainFile` parameter name to something like `resolvableDocumentURIUsingMainFile`? I’m not a huge fan of that name myself right now but something along those lines maybe?

Sources/SourceKitLSP/SourceKitLSPServer.swift
ahoppen: OK, maybe worth spelling this out.

ahoppen (approved):
LGTM, just have the following comments open. Feel free to merge after addressing them:
- [x] https://github.com/swiftlang/sourcekit-lsp/pull/2730#discussion_r3645038020
- [x] https://github.com/swiftlang/sourcekit-lsp/pull/2730#discussion_r3645071839

Sources/SourceKitLSP/SourceKitLSPServer.swift
```diff
@@ -1649,6 +1640,50 @@ extension SourceKitLSPServer {
     )
   }
 
+  /// Split an index module name of the form `Module` or `Module.Group` into its module and optional group
+  /// components, on the first `.`.
+  private nonisolated static func splitModuleNameAndGroup(
```
hamishknight: I know this isn't new with this PR, but this doesn't seem like it will handle submodules correctly? I guess we need to be properly tracking that in the index data?

hamishknight (approved):

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/75
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/75
@swift-ci Please test Windows

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: I revised the doc-comment, and changed the parameter name to `referenceDocumentMainFile`.

Sources/SourceKitLSP/SourceKitLSPServer.swift
rintaro: Updated to only handle `Swift.*` because in practice `Swift` is the only module using groups.

rintaro:
https://github.com/swiftlang/swift-tools-protocols/pull/75
@swift-ci Please test Windows

PR #2731, opened 2026-07-23
Add toggle test enabled/disabled code action

### LINKED ISSUE #2524: Toggle disabled test
### Description

Toggle a test between enabled and disabled states.

**Swift Testing:**

**Before:**

```swift
@Test
func testExample() {
    #expect(2 + 2 == 4)
}
```

**After:**

```swift
@Test(.disabled())
func testExample() {
    #expect(2 + 2 == 4)
}
```

**XCTest:**

**Before:**

```swift
func testExample() {
    XCTAssertEqual(2 + 2, 4)
}
```

**After:**

```swift
func testExample() throws {
    throw XCTSkip("Disabled")
    XCTAssertEqual(2 + 2, 4)
}
```

somiljain2006:
Fixes #2524 

Introduce a new refactoring that toggles Swift Testing `@Test` functions by adding or removing the `.disabled()` trait, and XCTest test methods by inserting or removing `throw XCTSkip("Disabled")` while preserving `throws` when still required. Add comprehensive tests covering both enable/disable flows, edge cases such as existing traits and remaining `try` expressions, and document the new refactoring in the available code actions.


Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+
+    return [
+      SourceEdit(
+        range: editRange,
+        replacement: refactoredFunction.description
+      )
+    ]
+  }
+
+  private static func findTestAttribute(in function: FunctionDeclSyntax) -> (Int, AttributeSyntax)? {
```
ahoppen: Could you add labels to the tuple elements to clarify especially what the integer represents?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+    }
+    return nil
+  }
+
+  private static func toggleSwiftTestingTest(
+    function: FunctionDeclSyntax,
+    attributeIndex: Int,
+    testAttribute: AttributeSyntax
+  ) -> FunctionDeclSyntax {
+    var newFunction = function
```
ahoppen: I would declare this variable only down below before it’s used because it doesn’t get modified in the first part of the body.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+struct ToggleTestCodeAction: SyntaxRefactoringCodeActionProvider {
+
+  package static let title: String = "Toggle Test Enabled/Disabled"
+
+  package static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> FunctionDeclSyntax? {
+
```
ahoppen: Superfluous newline

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      if argumentsArray.isEmpty {
+        newAttribute.leftParen = nil
+        newAttribute.arguments = nil
+        newAttribute.rightParen = nil
+      } else {
+        newAttribute.arguments = .argumentList(LabeledExprListSyntax(argumentsArray))
+      }
+    } else {
+
```
ahoppen: Superfluous newline.

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,253 @@
...
+    return newFunction.formatted().as(FunctionDeclSyntax.self) ?? newFunction
+  }
+
+  private static func isDisabledTrait(_ expr: ExprSyntax) -> Bool {
+    guard let functionCall = expr.as(FunctionCallExprSyntax.self),
+      let memberAccess = functionCall.calledExpression.as(MemberAccessExprSyntax.self)
+    else {
+      return false
+    }
+    return memberAccess.declName.baseName.text == "disabled"
```
ahoppen: Should this also check for `disabled(if:)`, in which case the refactoring action shouldn’t be available because we can’t unconditionally toggle the test between enabled and disabled.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+  private static func toggleSwiftTestingTest(
+    function: FunctionDeclSyntax,
+    attributeIndex: Int,
+    testAttribute: AttributeSyntax
+  ) -> FunctionDeclSyntax {
+    var newFunction = function
+    var newAttribute = testAttribute
+    var argumentsList = LabeledExprListSyntax([])
+
+    if case .argumentList(let exprList) = testAttribute.arguments {
```
ahoppen: If the arguments are present but not `.arguemntList`, we are in some kind of unexpected scenario (I can’t imagine how it would come up but better play it safe). In that case, I think we should fail the refactoring instead of assuming that the test is not disabled.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      argumentsList = exprList
+    }
+
+    var argumentsArray = Array(argumentsList)
+
+    // Check if the `.disabled(...)` trait is already present in the arguments.
+    if let disabledArgIndex = argumentsArray.firstIndex(where: { isDisabledTrait($0.expression) }) {
+
```
ahoppen: Superfluous newline

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      )
+      let functionCall = FunctionCallExprSyntax(
+        calledExpression: disabledExpr,
+        leftParen: .leftParenToken(),
+        arguments: [],
+        rightParen: .rightParenToken()
+      )
+
+      if !argumentsArray.isEmpty {
+        argumentsArray[argumentsArray.count - 1].trailingComma = .commaToken(trailingTrivia: .space)
```
ahoppen: We should only add a comma if none exists yet. Otherwise we might override comments that may be associated to an existing comma’s trivia.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      } else {
+        newAttribute.arguments = .argumentList(LabeledExprListSyntax(argumentsArray))
+      }
+    } else {
+
+      // The test is currently enabled, so disable it by appending the `.disabled()` trait.
+      let disabledExpr = ExprSyntax(
+        MemberAccessExprSyntax(declName: DeclReferenceExprSyntax(baseName: .identifier("disabled")))
+      )
+      let functionCall = FunctionCallExprSyntax(
```
ahoppen: Let’s name this `disabledCall` because `functionCall` sounds more like the test function itself.

```suggestion
      let disabledCall = FunctionCallExprSyntax(
```

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+        leftParen: .leftParenToken(),
+        arguments: [],
+        rightParen: .rightParenToken()
+      )
+
+      if !argumentsArray.isEmpty {
+        argumentsArray[argumentsArray.count - 1].trailingComma = .commaToken(trailingTrivia: .space)
+      }
+
+      argumentsArray.append(LabeledExprSyntax(expression: ExprSyntax(functionCall)))
```
ahoppen: Do you have any opinions on whether `.disabled` should be inserted at the start or end? I feel like I would naturally put it at the start when editing this manually.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      newAttribute.leftParen = testAttribute.leftParen ?? .leftParenToken()
+      newAttribute.arguments = .argumentList(LabeledExprListSyntax(argumentsArray))
+      newAttribute.rightParen = testAttribute.rightParen ?? .rightParenToken()
+    }
+
+    var attributesArray = Array(function.attributes)
+    attributesArray[attributeIndex] = .attribute(newAttribute)
+    newFunction.attributes = AttributeListSyntax(attributesArray)
+
+    return newFunction.formatted().as(FunctionDeclSyntax.self) ?? newFunction
```
ahoppen: We should not call `.formatted()` because it may change the existing test function’s formatting. We should just ensure that all edits are already well-formatted without `.formatted()`.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+  private static func isXCTestFunction(_ function: FunctionDeclSyntax) -> Bool {
+    let name = function.name.text
+
+    guard name.hasPrefix("test"), name.count > 4 else {
+      return false
+    }
+
+    let indexAfterTest = name.index(name.startIndex, offsetBy: 4)
+    let nextChar = name[indexAfterTest]
+    guard nextChar.isUppercase || nextChar.isNumber || nextChar == "_" else {
```
ahoppen: Where did you get this test detection rule from? Is this defined somewhere in XCTest? As far as I can tell `testabc` would also be considered as a test by XCTest but not by your implementation.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      return false
+    }
+
+    let hasNoParams = function.signature.parameterClause.parameters.isEmpty
```
ahoppen: I think all of these could be early-exit `guard` checks that immediately communicate what happens if the condition is not satisfied. 


```suggestion
    guard function.signature.parameterClause.parameters.isEmpty else {
      return false
    }
```

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+    let hasNoReturnType = function.signature.returnClause == nil
+    let isNotStaticOrClass = !function.modifiers.contains { modifier in
+      let kind = modifier.name.tokenKind
+      return kind == .keyword(.static) || kind == .keyword(.class)
+    }
+
+    return hasNoParams && hasNoReturnType && isNotStaticOrClass
+  }
```
ahoppen: We should share this with the detection logic in `SyntacticSwiftXCTestScanner` instead of duplicating it.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+
+  private static func toggleXCTest(function: FunctionDeclSyntax) -> FunctionDeclSyntax {
+    guard let body = function.body else { return function }
+
+    var newFunction = function
+    var statementsArray = Array(body.statements)
+
+    if let skipIndex = statementsArray.firstIndex(where: isXCTSkipStatement) {
```
ahoppen: I think we should only consider an `XCTest` as disabled if the first statement is a `throw XCTSkip` because otherwise it does some testing before being skipped (which is an odd behavior but who knows what people like to do).

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+      statementsArray.remove(at: skipIndex)
+
+      // Remove `throws` only if the remaining body no longer requires it.
+      let remainingBody = CodeBlockSyntax(statements: CodeBlockItemListSyntax(statementsArray.map { $0.trimmed }))
+      if !bodyRequiresThrows(remainingBody) {
```
ahoppen: `bodyRequiresThrows` does not correctly identify eg.

```swift
if something {
  throw SomethingElse()
}
```

But instead of trying to discover whether a function may throw (which may be its own interesting non-trivial analysis), I think we can just leave the `throws` in there. There’s no harm in marking a test function as `throws`.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+        )
+        newSignature.effectSpecifiers = effects
+        newFunction.signature = newSignature
+      }
+    }
+
+    newFunction.signature.parameterClause.rightParen.trailingTrivia = []
+    newFunction.body?.statements = CodeBlockItemListSyntax(statementsArray.map { $0.trimmed })
+
+    let formattedFunction = newFunction.formatted().as(FunctionDeclSyntax.self)
```
ahoppen: Similar to my comment above, we shouldn’t call `.formatted()` here.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+        effects.throwsClause = ThrowsClauseSyntax(
+          throwsSpecifier: .keyword(.throws)
+        )
+        newSignature.effectSpecifiers = effects
+        newFunction.signature = newSignature
+      }
+    }
+
+    newFunction.signature.parameterClause.rightParen.trailingTrivia = []
+    newFunction.body?.statements = CodeBlockItemListSyntax(statementsArray.map { $0.trimmed })
```
ahoppen: Why are you trimming all the statements here?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,253 @@
...
+        attributeIndex: index,
+        testAttribute: testAttribute
+      )
+    } else if isXCTestFunction(function) {
+      refactoredFunction = toggleXCTest(function: function)
+    } else {
+      throw RefactoringNotApplicableError("Function is not a recognized test")
+    }
+
+    let editRange = function.position..<function.endPosition
```
ahoppen: If there is a way to only report the actual edit of adding/removing the code necessary to enable/disable the test, I would prefer that compared to reporting the entire function body again. I am not aware of any editors that do it, but you could imagine that an editor wants to visualize exactly what the action would do before it’s applied and such a narrower edit would be nicer to visualize.

ahoppen (commented):
Thanks for providing this refactoring action, @somiljain2006. I have left some comments inline.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
somiljain2006: I opted to append it to avoid breaking tests that use a custom display name. Since @Test accepts a string as its first positional argument (for example, @Test("Custom Name", .trait)), unconditionally inserting .disabled() at the beginning would produce @Test(.disabled(), "Custom Name"), which is no longer a valid argument order.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
somiljain2006: I looked into sharing the logic with SyntacticSwiftXCTestScanner, but ToggleTestCodeAction lives in the SwiftSyntaxCodeActions module, whereas SyntacticSwiftXCTestScanner lives in SwiftLanguageService. Because of that module boundary, SwiftSyntaxCodeActions can't see the extension.

ahoppen: We could implement it in `SwiftSyntaxCodeActions`, make it `package` and then use it from `SwiftLanguageService`.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
ahoppen: Oh, fair. I’m fine with adding it at the end then.

Sources/SwiftLanguageService/SyntacticSwiftXCTestScanner.swift
```diff
@@ -51,19 +52,8 @@ final class SyntacticSwiftXCTestScanner: SyntaxVisitor {
...
-      }
-      guard function.modifiers.map(\.name.tokenKind).allSatisfy({ $0 != .keyword(.static) && $0 != .keyword(.class) })
-      else {
-        // Test methods can't be static.
-        return nil
-      }
-      guard function.signature.returnClause == nil, function.signature.parameterClause.parameters.isEmpty else {
-        // Test methods can't have a return type or have parameters.
-        // Technically we are also filtering out functions that have an explicit `Void` return type here but such
-        // declarations are probably less common than helper functions that start with `test` and have a return type.
```
ahoppen, line 65: Let’s keep this comment in the pulled out implementation

Sources/SwiftSyntaxCodeActions/FunctionDeclSyntax+IsXCTestFunction.swift
```diff
@@ -0,0 +1,40 @@
...
+    guard signature.returnClause == nil else {
+      return false
+    }
+
+    guard !modifiers.contains(where: {
+      let kind = $0.name.tokenKind
+      return kind == .keyword(.static) || kind == .keyword(.class)
+    }) else {
+      return false
+    }
```
ahoppen: I think the previous implementation is a little more concise. I’d prefer to keep that.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+    } else if function.isXCTestFunction {
+      return try toggleXCTestEdits(function: function)
+    } else {
+      throw RefactoringNotApplicableError("Function is not a recognized test")
+    }
+  }
+
+  private static func findTestAttribute(
+    in function: FunctionDeclSyntax
+  ) -> (index: Int, attribute: AttributeSyntax)? {
```
ahoppen: I don’t think we ever use the index.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+      return disabledCall.arguments.contains { $0.label?.text == "if" }
+    }
+  }
+
+  private static func testAttributeRange(_ testAttribute: AttributeSyntax) -> Range<AbsolutePosition> {
+    let range = testAttribute.trimmedRange
+    if !range.isEmpty {
+      return range
+    }
+    return testAttribute.positionAfterSkippingLeadingTrivia..<testAttribute.endPositionBeforeTrailingTrivia
```
ahoppen: Isn’t this exactly the definition of `trimmedRange`?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  private static func toggleSwiftTestingTest(testAttribute: AttributeSyntax) throws -> AttributeSyntax {
```
ahoppen: Should this return the `[SourceEdit]` directly? That way the caller doesn’t need to correlate it the the range of the attribute.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+
+      if argumentsArray.isEmpty {
+        newAttribute.leftParen = nil
+        newAttribute.arguments = nil
+        newAttribute.rightParen = nil
+      } else {
+        newAttribute.arguments = .argumentList(LabeledExprListSyntax(argumentsArray))
+      }
+    } else {
+
```
ahoppen: Superfluous newline

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+internal import LanguageServerProtocol
+internal import SourceKitLSP
+import SwiftBasicFormat
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+struct ToggleTestCodeAction: SyntaxRefactoringCodeActionProvider {
+
```
ahoppen: Superfluous newline

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+  private static func isDisabledTrait(_ expr: ExprSyntax) -> Bool {
+    guard let disabledCall = expr.as(FunctionCallExprSyntax.self),
+      let memberAccess = disabledCall.calledExpression.as(MemberAccessExprSyntax.self)
+    else {
+      return false
+    }
+    return memberAccess.declName.baseName.text == "disabled"
+  }
+
+  private static func isXCTestFunction(_ function: FunctionDeclSyntax) -> Bool {
```
ahoppen: Isn’t this duplicate with the code you just pulled out?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+
+  private static func toggleXCTestEdits(function: FunctionDeclSyntax) throws -> [SourceEdit] {
+    guard let body = function.body else {
+      throw RefactoringNotApplicableError("XCTest function has no body")
+    }
+
+    if let firstStatement = body.statements.first, isXCTSkipStatement(firstStatement) {
+      return [
+        SourceEdit(
+          range: firstStatement.positionAfterSkippingLeadingTrivia..<firstStatement.endPositionBeforeTrailingTrivia,
```
ahoppen: We should probably also remove the trivia so we aren’t left with an empty line

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+            range: insertionPosition..<insertionPosition,
+            replacement: "throw XCTSkip(\"Disabled\")\n\(indentation)"
+          )
+        )
+      } else {
+        insertionPosition = body.rightBrace.positionAfterSkippingLeadingTrivia
+
+        let hasNewline = body.rightBrace.leadingTrivia.description.contains("\n")
+
+        let baseIndentation = indentationBefore(function)
```
ahoppen: Can’t you infer the indentation using `.indentationOfLine`?

Also, I would infer it from the function body instead of the function declaration so that you don’t need to make any assumption about the user’s indentation settings.

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+        leftParen: .leftParenToken(),
+        arguments: [],
+        rightParen: .rightParenToken()
+      )
+
+      if !argumentsArray.isEmpty, argumentsArray[argumentsArray.count - 1].trailingComma == nil {
+        argumentsArray[argumentsArray.count - 1].trailingComma = .commaToken(trailingTrivia: .space)
+      }
+
+      argumentsArray.append(LabeledExprSyntax(expression: ExprSyntax(disabledCall)))
```
ahoppen: For a parameterized test this appends after the `arguments:` argument, so `@Test(arguments: [1, 2])` becomes `@Test(arguments: [1, 2], .disabled())`, which does not compile — the variadic traits have to come before `arguments:`. `traits(ofTestAttribute:)` in `SwiftTestingScanner.swift` slices at the first labeled argument for exactly this reason. Could we insert the trait before the first labeled argument instead of always appending?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,281 @@
...
+
+      let name = attr.attributeName.trimmedDescription
+      if name == "Test" || name == "Testing.Test" {
+        return (index: index, attribute: attr)
+      }
+    }
+    return nil
+  }
+
+  private static func hasConditionalDisabledTrait(_ testAttribute: AttributeSyntax) -> Bool {
```
ahoppen, line 77: `.disabled { someCondition() }` is conditional as well but has no `if:` label, so we offer the toggle for it and then silently delete the condition along with the trait. `TestingAttributeData.isDisabled` in `SwiftTestingScanner.swift` handles this by also checking `functionCall.trailingClosure != nil` — can we pull out that implementation and share it, similar to how we detect XCTest cases?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    if let (_, testAttribute) = findTestAttribute(in: function) {
+      if hasConditionalDisabledTrait(testAttribute) {
+        return nil
+      }
+      return function
+    }
+
+    guard function.isXCTestFunction else {
```
ahoppen: `isXCTestFunction` only looks at the function’s own name and signature, so we offer this for any function starting with `test`, including a plain global function in non-test code. For `func testConnection()` in some networking file we would offer to insert `throw XCTSkip("Disabled")`, which does not even compile there. `SyntacticSwiftXCTestScanner` only considers functions that are members of a class or extension — should we require the same here?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,281 @@
...
+      let function = scope.innermostNodeContainingRange?.findParentOfSelf(
+        ofType: FunctionDeclSyntax.self,
+        stoppingIf: { _ in false }
+      )
+    else {
+      return nil
+    }
+
+    if let (_, testAttribute) = findTestAttribute(in: function) {
+      if hasConditionalDisabledTrait(testAttribute) {
```
ahoppen, line 40: Could you add a comment on why we do not offer the action at all when a `.disabled(if:)` trait is present? The action silently disappearing is surprising and I do not think a reader can reconstruct the reasoning from the code. This branch also has no test — could you add one next to `testToggleTestNotOfferedForNonTestFunction`?

Sources/SwiftSyntaxCodeActions/ToggleTestCodeAction.swift
```diff
@@ -0,0 +1,281 @@
...
+//===----------------------------------------------------------------------===//
+
+internal import LanguageServerProtocol
+internal import SourceKitLSP
+import SwiftBasicFormat
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+struct ToggleTestCodeAction: SyntaxRefactoringCodeActionProvider {
```
ahoppen: None of the other providers in this directory carry a `CodeAction` suffix — `SwapBinaryOperands`, `ApplyDeMorganLaw`, `GenerateEnumCaseAsAccessors`, `ConvertCommentToDocComment`. How about `ToggleTestDisabled` (and renaming the file to match)? Also, could you add a doc comment on the type?

Sources/SwiftSyntaxCodeActions/FunctionDeclSyntax+IsXCTestFunction.swift
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension FunctionDeclSyntax {
```
ahoppen: Files that contain nothing but an extension are named `<Type>+<Feature>.swift` in this repo — `Toolchain+PluginPaths.swift`, `DocumentSnapshot+FromFileContents.swift`, … Could you rename this to `FunctionDeclSyntax+IsXCTestFunction.swift`, and the `CMakeLists.txt` entry with it? As it stands the name reads like it declares a type called `XCTestFunction`.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -3353,6 +3353,338 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+      in: """
+        1️⃣func helper() {
+          print("hi")
+        }2️⃣
+        """,
+      atMarker: "1️⃣"
+    )
+  }
+
+  func testToggleXCTestDisabledToEnabled_keepsThrowsIfTryPresent() async throws {
```
ahoppen: How does this differ from `testToggleXCTestDisabledToEnabled` above? Both assert the exact same single edit, and since `throws` is never removed the `try doSomething()` does not change anything about what is verified. If the intent is “we keep `throws` because a `try` remains”, the interesting half — dropping `throws` when nothing else in the body throws — is not implemented, so the name promises more than the test checks.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -3353,6 +3353,338 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testToggleXCTestEnabledToDisabled() async throws {
```
ahoppen, line 3421: Could you add a case for an `async` test method? `toggleXCTestEdits` has a dedicated branch anchoring the `throws` insertion after the `async` specifier and it is untested — `func testExample() async {` needs to become `async throws`, not `throws async`.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -3353,6 +3353,338 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                )
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testToggleSwiftTestingTestDisabledToEnabled_withDisabledReason() async throws {
```
ahoppen: None of the existing tests in this file use `_` in their names (repo-wide there are two, and those separate a build-system suffix). Could you camel-case these, e.g. `testToggleSwiftTestingTestDisabledToEnabledWithDisabledReason`?

Documentation/Refactoring Actions.md
```diff
@@ -60,6 +60,7 @@ The specific refactorings available depend on what code is selected or where the
 | **Convert to zero parameter function** | Cursor on a read-only computed property |
 | **Add documentation** | Cursor on a function, type, property, or macro declaration |
 | **Convert comment to doc comment** | Cursor on a regular comment preceding a declaration |
+| **Toggle Test Enabled/Disabled** | Cursor on a Swift Testing @Test function or an XCTest test method |
```
ahoppen: Nit: the other rows put code in backticks — “Cursor on the `switch` keyword”, “Cursor on `if let x = x`”. So how about ``Cursor on a Swift Testing `@Test` function or an XCTest test method``?

Sources/SwiftSyntaxCodeActions/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_library(SwiftSyntaxCodeActions STATIC
   ConvertZeroParameterFunctionToComputedProperty.swift
   DeclModifierRemover.swift
   FormatRawStringLiteral.swift
+  FunctionDeclSyntax+IsXCTestFunction.swift
```
ahoppen: This PR adds three new files but only two of them are listed here — `SwiftTestingDisabledTrait.swift` is missing, which will break the CMake (Windows) build. Could you add it?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+
+    return [
+      SourceEdit(
+        range: testAttribute.trimmedRange,
+        replacement: newAttribute.trimmedDescription
+      )
+    ]
+  }
+
+  private static func isDisabledTrait(_ expr: ExprSyntax) -> Bool {
```
ahoppen: `isDisabledTrait` only checks that the member's base name is `disabled`, while `isSwiftTestingDisabledTrait` also validates the base is absent, `ConditionTrait`, or `Testing.ConditionTrait`. So `@Test(MyTraits.disabled())` is a disabled trait to one and not the other, and the toggle would silently strip an unrelated trait. Can `toggleSwiftTestingEdits` just use `isSwiftTestingDisabledTrait`?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    return function
+  }
+
+  package static func textRefactor(syntax function: FunctionDeclSyntax, in context: Void) throws -> [SourceEdit] {
+    // Route to the appropriate toggling logic based on which test framework is detected.
+    if let testAttribute = findTestAttribute(in: function) {
+      return try toggleSwiftTestingEdits(testAttribute: testAttribute)
+    } else if function.isSyntacticXCTestMethod { // <-- Consistent with nodeToRefactor
```
ahoppen: Instead of having `<---` here, I would add a comment at the start of the `if` to say that this needs to be kept in sync with `nodeToRefactor`.

Sources/SwiftSyntaxCodeActions/FunctionDeclSyntax+IsXCTestFunction.swift
```diff
@@ -0,0 +1,53 @@
...
+  package var isXCTestFunction: Bool {
```
ahoppen: Could you add doc comments to these? `isXCTestFunction` encodes a heuristic (name prefix, no parameters, no return type, not static/class) whose rationale used to be spelled out inline in `SyntacticSwiftXCTestScanner`. The surviving comment at the call site now only explains the return type/parameter part and sits above a guard that doesn't show any of it — I'd move the whole explanation here.

Sources/SwiftSyntaxCodeActions/SwiftTestingDisabledTrait.swift
```diff
@@ -0,0 +1,58 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+package extension FunctionCallExprSyntax {
```
ahoppen: Only `isSwiftTestingUnconditionalDisabledTrait` is used outside this module. Could `isSwiftTestingDisabledTrait` and `isSwiftTestingConditionalDisabledTrait` (and `isSyntacticXCTestMethod` in the sibling file) be `fileprivate`/`internal`?

ahoppen:
@somiljain2006 As the issue you are fixing is marked as a good first issue, could you please indicate if or how much you used AI to generate this PR? I also want to make you aware of the AI guidelines we have for the swiftlang org here: https://github.com/swiftlang/project-operations/blob/main/contributing/ai-tools.md

somiljain2006:
@ahoppen When I initially picked up any issue and submitted my first implementation, I didn't rely much on AI. I spent time understanding the relevant files, the existing implementation, and how everything fit together, so I was comfortable making the initial changes myself.

As the review progressed, though, I found some of the review comments quite difficult to understand, especially since I'm still getting familiar with this part of the codebase. That's where I used AI the most, not to blindly generate code, but to help me understand the feedback, discuss possible approaches, and explain why certain changes were being requested. I still implemented the changes myself, tested them, and made sure I understood and agreed with what I was submitting before updating the PR.

If you feel this PR has ended up going in a different direction than what would be appropriate, I'm open to closing it if that's the best course of action.

ahoppen:
If you personally learn something during this interaction, that’s OK. I just wanted to make sure that I’m not directly talking to an AI agent. These PR reviews take time and for these kinds of issues, the biggest benefit is growing contributors. If that’s happening here, then everything is good.

Sources/SwiftSyntaxCodeActions/SwiftTestingDisabledTrait.swift
```diff
@@ -0,0 +1,58 @@
...
+      return true
+    }
+
+    return arguments.lazy
+      .compactMap(\.label?.text)
+      .contains("if")
+  }
+
+  var isSwiftTestingUnconditionalDisabledTrait: Bool {
+    isSwiftTestingDisabledTrait && !isSwiftTestingConditionalDisabledTrait
```
ahoppen: We are effectively calling `isSwiftTestingDisabledTrait` twice here - once through `isSwiftTestingConditionalDisabledTrait`. Would it make sense to have a `swiftTestingDisabledTrait` that returns an enum with either `none`, `disabled` or `conditionallyDisabled`. That way callers are also forced to always think about the conditionally disabled case.

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  private static func findTestAttribute(
+    in function: FunctionDeclSyntax
+  ) -> AttributeSyntax? {
+    for element in function.attributes {
+      guard case .attribute(let attr) = element else { continue }
+
+      let name = attr.attributeName.trimmedDescription
+      if name == "Test" || name == "Testing.Test" {
```
ahoppen: Could we re-use `AttributeSyntax.isNamed(_:inModuleNamed:)` here instead, which also accounts for whitespace between `Testing` and `Test`?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+    return memberAccess.declName.baseName.text == "disabled"
+  }
+
+  private static func toggleXCTestEdits(function: FunctionDeclSyntax) throws -> [SourceEdit] {
+    guard let body = function.body else {
+      throw RefactoringNotApplicableError("XCTest function has no body")
+    }
+
+    if let firstStatement = body.statements.first, isXCTSkipStatement(firstStatement) {
+      return [
```
ahoppen, line 167: I think this still leaves an empty line at the start of the function because the newline will be leading trivia to the next token. Could you check if the leading trivia of the next token is purely whitespace (ie. newlines and tabs/spaces) and remove those as well in that case?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+      }
+
+      if let firstStatement = body.statements.first {
+        let insertionPosition = firstStatement.positionAfterSkippingLeadingTrivia
+        let statementIndentation = firstStatement.firstToken(viewMode: .sourceAccurate)?.indentationOfLine ?? []
+
+        edits.append(
+          SourceEdit(
+            range: insertionPosition..<insertionPosition,
+            replacement: "throw XCTSkip(\"Disabled\")\n\(statementIndentation.description)"
```
ahoppen: Nitpick: I usually feel like multi-line string literals represent these multi-line strings better than having `\n` in them.

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
```diff
@@ -0,0 +1,236 @@
...
+          )
+        )
+      }
+
+      return edits
+    }
+  }
+
+  private static func isXCTSkipStatement(_ item: CodeBlockItemSyntax) -> Bool {
+    guard let throwStmt = item.item.as(StmtSyntax.self)?.as(ThrowStmtSyntax.self),
```
ahoppen: I don’t think the intermediate case to `StmtSyntax` is needed.

```suggestion
    guard let throwStmt = item.item.as(ThrowStmtSyntax.self),
```

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
somiljain2006: Should we just remove it?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
somiljain2006: The newline after throw XCTSkip(...) is part of the next token’s leading trivia. In this implementation, the edit starting at firstStatement.position still results in a single newline before the following statement rather than leaving an empty line. I also added a test covering the case where a comment follows XCTSkip, and it preserves the expected formatting.

somiljain2006:
Sorry, I accidentally referenced the wrong PR in the previous mention. Please disregard that reference

Sources/SwiftSyntaxExtensions/AttributeSyntax+isNamed.swift
ahoppen, line 1: Could we place this in `SKUtilities`? I think that would be a good place for these general utilities that purely depend on swift-syntax. Also, on second thought, that would be a good location for the `isXCTestFunction` and `swiftTestingDisabledTrait` implementations.

Sources/SwiftSyntaxCodeActions/SwiftTestingDisabledTrait.swift
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+      if let baseMemberAccess = base.as(MemberAccessExprSyntax.self),
+        baseMemberAccess.declName.baseName.text == "ConditionTrait",
+        let moduleRef = baseMemberAccess.base?.as(DeclReferenceExprSyntax.self)
+      {
+        return moduleRef.baseName.text == "Testing"
+      }
+
+      return false
```
ahoppen: Why can’t we use the `fullyQualifiedName` implementation that we had previously here?

Sources/SwiftSyntaxCodeActions/ToggleTestDisabled.swift
ahoppen: Ah, I forgot that you’re removing the newline that’s part of the `XCTSkip()` statement.

ahoppen (commented):
Just two more nitpicks, otherwise looks good now.

ahoppen (approved):
LGTM. Thank you

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

somiljain2006:
@ahoppen The recent Windows CI failure was due to SwiftSyntax missing from SKUtilities's target_link_libraries in CMakeLists.txt. Could you rerun the CI?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

somiljain2006:
@ahoppen Sorry to bother you, but I'm having trouble identifying the exact cause of the Windows CI failure. I can see that the build fails with the missing required module SwiftSyntax while building SemanticIndex, but I haven't been able to figure out where the missing dependency is being introduced. I checked the relevant CMake files (SKUtilities and SemanticIndex), but nothing obvious stood out to me. Can you tell me what I'm missing?

ahoppen:
@rintaro Could you take a look at the Windows CI failure here, you know a lot more about that CMake build? My best bet would be that `SKUtilities` now needs to be imported publicly by all modules so that transitive dependencies can find SwiftSyntax but I’m not 100% sure if that’s indeed the problem.

rintaro:
`SKUtilities` now exposes SwiftSyntax types in its API, so loading `SKUtilities.swiftmodule` requires `SwiftSyntax.swiftmodule` to be findable. `SemanticIndex` doesn't import `SKUtilities` itself, but it links `BuildServerIntegration` and `ToolchainRegistry`, both of which import `SKUtilities` and both link it `PRIVATE`. For a static library, a `PRIVATE` link still propagates the link itself, but not the usage requirements, which for Swift is the module search path.

I believe the fix is to move `SKUtilities` from the `PRIVATE` to the `PUBLIC` `target_link_libraries` list in the targets that import it:

- `Sources/BuildServerIntegration/CMakeLists.txt`
- `Sources/ToolchainRegistry/CMakeLists.txt`
- `Sources/SwiftLanguageService/CMakeLists.txt`
- `Sources/Diagnose/CMakeLists.txt`
- ~`Sources/SourceKitLSP/CMakeLists.txt`~ - already PUBLIC
- ~`Sources/SwiftSourceKitPlugin/CMakeLists.txt`~ - nothing imports this

A short comment next to `target_link_libraries(SKUtilities PUBLIC SwiftSyntax::SwiftSyntax)` explaining that consumers have to link `SKUtilities` publicly for this reason would be helpful, since the `PUBLIC` links otherwise look like they contradict the internal `import SKUtilities` in those modules.

Why this didn't came up until now? All the first-party modules (i.e. `sourcekit-lsp/Sources/*`) land in a single `CMAKE_Swift_MODULE_DIRECTORY` that every target already has on its search path, which is why `PRIVATE` has been fine for them so far, however `SwiftSyntax` comes from a separate build directory and is only reachable through the `SwiftSyntax::SwiftSyntax` imported target, so it needs a propagation path.

--

BUT...

`SKUtilitiesForPlugin` compiles the same source list, so the sourcekitd plugin now builds these four files and links SwiftSyntax into its dylib, even though none of the plugin sources use them. Hopefully dead-code elimination strips them all, but I'm skeptical it does. So I'd rather move the SwiftSyntax helpers into their own small module (e.g. `SwiftSyntaxExtensions`), shared by `SwiftLanguageService`, `SwiftSyntaxCodeActions`, and `SourceKitLSP`, all of which already depend on SwiftSyntax. @ahoppen wdyt?

Sources/SKUtilities/AttributeSyntax+isNamed.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
```
rintaro: This should be `package` under `InternalImportsByDefault` enabled

```suggestion
package import SwiftSyntax
```

rintaro:
Thinking about this more. It's even possible that the SwiftSyntax symbols in the plugin could conflict with those in SourceKitService (they don't at this point because we're using different ABI names, but still...).
So yeah, I'd rather not link SwiftSyntax into anything that's included by the plugins.

ahoppen:
Great point with the duplicate symbols for the plugin, @rintaro. I agree with you that a new `SwiftSyntaxExtensions` module would be great here.

@somiljain2006 Could you add that?

ahoppen (approved):
Thank you

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

somiljain2006:
@ahoppen Can you run the CI again? This failure was due to import ordering.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftLanguageService/SwiftTestingScanner.swift
```diff
@@ -15,6 +15,8 @@
 import SourceKitLSP
 import SwiftParser
 import SwiftSyntax
+import SwiftSyntaxCodeActions
```
rintaro: This doesn't look right. What in this file is using `SwiftSyntaxCodeActions`?

Sources/SwiftLanguageService/SyntacticSwiftXCTestScanner.swift
```diff
@@ -13,6 +13,8 @@
 @_spi(SourceKitLSP) import LanguageServerProtocol
 import SourceKitLSP
 import SwiftSyntax
+import SwiftSyntaxCodeActions
```
rintaro: Ditto

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

ahoppen:
Thanks for the contribution, @somiljain2006.

PR #2733, opened 2026-07-25
[WIP] Update test to account for updated SwiftPM handling of build tool dependencies

owenv:
Update a test which looks like it was depending on slightly incorrect behavior. Add an explicit product to make it clear the executable is intended to build for the target and host


owenv:
Testing alongside https://github.com/swiftlang/swift-package-manager/pull/10328

ahoppen:
Ideally, we wouldn't need a product declaration to provide semantic functionality in the package. Why is it needed?

PR #2734, opened 2026-07-26
Add support for hooking ExternalBuildServerAdapter requests in tests

owenv:
This threads the existing build server hooks through to the external build server adapter to allow a few more tests to pass when using the external SwiftPM BSP implementation.


owenv:
@swift-ci test

ahoppen (approved):

PR #2735, opened 2026-07-26
Allow extra SwiftPM build server arguments

valwalt:
SwiftPM's experimental build-server flags can evolve faster than SourceKit-LSP's typed configuration. There is currently no generic way to pass a newly introduced flag to `swift package experimental-build-server`.

This PR adds a `swiftPM.extraArguments` configuration option and appends its values verbatim to the builtin and external SwiftPM build-server invocation.

### DIFF
--- a/Documentation/Configuration File.md
+++ b/Documentation/Configuration File.md
@@ -33,6 +33,7 @@ The structure of the file is currently not guaranteed to be stable. Options may 
   - `buildToolsSwiftCompilerFlags: string[]`: Extra arguments passed to the compiler for Swift files or plugins. Equivalent to SwiftPM's `-Xbuild-tools-swiftc` option.
   - `forceResolvedVersions: boolean`: Equivalent to SwiftPM's `--force-resolved-versions` option. Makes all processes (including background indexing) treat Package.resolved as a lock file.
   - `disableSandbox: boolean`: Disables running subprocesses from SwiftPM in a sandbox. Equivalent to SwiftPM's `--disable-sandbox` option. Useful when running `sourcekit-lsp` in a sandbox because nested sandboxes are not supported.
+  - `extraArguments: string[]`: Additional arguments appended to SwiftPM when starting or preparing the build server.
   - `buildSystem: "native"|"swiftbuild"`: Which SwiftPM build system should be used when opening a package.
 - `compilationDatabase`: Dictionary with the following keys, defining options for workspaces with a compilation database.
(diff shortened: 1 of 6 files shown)


ahoppen (approved):
LGTM but I’d like to hear @rintaro’s opinion on this as well before merging.

Documentation/Configuration File.md
```diff
@@ -33,6 +33,7 @@ The structure of the file is currently not guaranteed to be stable. Options may
   - `buildToolsSwiftCompilerFlags: string[]`: Extra arguments passed to the compiler for Swift files or plugins. Equivalent to SwiftPM's `-Xbuild-tools-swiftc` option.
   - `forceResolvedVersions: boolean`: Equivalent to SwiftPM's `--force-resolved-versions` option. Makes all processes (including background indexing) treat Package.resolved as a lock file.
   - `disableSandbox: boolean`: Disables running subprocesses from SwiftPM in a sandbox. Equivalent to SwiftPM's `--disable-sandbox` option. Useful when running `sourcekit-lsp` in a sandbox because nested sandboxes are not supported.
+  - `buildServerArguments: string[]`: Additional arguments appended to `swift package experimental-build-server`. Only used when the SwiftPM build system is `swiftbuild`.
```
rintaro: I prefer naming this `extraArguments`.
Also, could you explain why we should limit this to `swiftbuild` build system? It doesn't seem hard to apply this to the builtin `SwiftPMBuildServer.prepare(singleTarget:)`.

Sources/BuildServerIntegration/ExternalBuildServerAdapter.swift
```diff
@@ -192,6 +192,9 @@ struct BuildServerConfig: Codable {
     if options.swiftPMOrDefault.disableSandbox == true {
       args.append("--disable-sandbox")
     }
+    if let buildServerArguments = options.swiftPMOrDefault.buildServerArguments {
```
rintaro: Could you add a comment stating that this block should come after everything else?

Documentation/Configuration File.md
valwalt: `extraArguments ` sounds good to me. I was under the impression that the native system was deprecated starting from 6.4 but you're right it's not hard to support it in any case, will update

rintaro:
@swift-ci Please test

rintaro (approved):
Thanks!

PR #2737, opened 2026-07-28
Exclude opaque build-server data from the index batching key

rintaro:
Background indexing batches a target's Swift files into a single compiler invocation when they share the same build settings, keyed on `FileBuildSettings`. That key included the opaque `data` field passed through from the build server's `textDocument/sourceKitOptions` responses. Since `data` doesn't affect the compiler invocation, a build server returning a per-file `data` value gave every file a unique key and defeated batching, running one compilation per file instead of a batch.

This clears `data` before keying, so files differing only in `data` are still batched togethe. `data` is still forwarded to clients unchanged, and the built-in SwiftPM build server sends no `data`, so this is a no-op for it.

Also adds an `updateIndexStoreTaskDidComputePartitions` index hook so tests can observe how files are grouped into invocations, with a test pinning the batching key: files differing only in `-index-unit-output-path` or `data` batch together, while files with genuinely different arguments are indexed separately.

rdar://183491001

### DIFF
--- a/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
+++ b/Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
@@ -90,22 +90,52 @@ package struct FileAndOutputPath: Sendable, Hashable {
   fileprivate var sourceFile: DocumentURI { file.sourceFile }
 }
 
-/// A single file or a list of files that should be indexed in a single compiler invocation.
-private enum UpdateIndexStorePartition {
-  case multipleFiles(filesAndOutputPaths: [(file: FileToIndex, outputPath: String)], buildSettings: FileBuildSettings)
-  case singleFile(file: FileAndOutputPath, buildSettings: FileBuildSettings)
-
-  var buildSettings: FileBuildSettings {
-    switch self {
-    case .multipleFiles(_, let buildSettings): return buildSettings
-    case .singleFile(_, let buildSettings): return buildSettings
-    }
+/// A set of files that should be indexed by a single compiler invocation.
+package struct UpdateIndexStorePartition {
+  /// The parts of a file's build settings that determine how the compiler is invoked, and thus whether two files can
+  /// be indexed by the same invocation.
+  package struct Invocation: Hashable {
+    package var language: Language
+    package var workingDirectory: String?
+    package var compilerArguments: [String]
   }
 
-  var files: [FileToIndex] {
-    switch self {
-    case .multipleFiles(let filesAndOutputPaths, _): return filesAndOutputPaths.map(\.file)
-    case .singleFile(let file, _): return [file.file]
+  /// The files to index, together with the index unit output path that should be recorded for it.
+  package enum Files {
+    /// Files indexed together, each with the index unit output path recorded via an `-output-file-map`.
+    case multipleFiles([(file: FileToIndex, indexUnitOutputPath: String)])
+
+    /// A single file for which no output path is known; indexed by its path without an `-output-file-map`.
+    case singleFile(FileToIndex)
+  }
+
+  package var invocation: Invocation
+  package var files: Files
+
+  /// The files this invocation indexes, regardless of how their output paths are handled.
+  package var indexedFiles: [FileToIndex] {
+    switch files {
+    case .multipleFiles(let filesAndOutputPaths): return filesAndOutputPaths.map(\.file)
+    case .singleFile(let file): return [file]
+    }
+  }
+}
+
+extension UpdateIndexStorePartition.Invocation {
+  fileprivate init(_ buildSettings: FileBuildSettings) {
+    self.init(
+      language: buildSettings.language,
+      workingDirectory: buildSettings.workingDirectory,
+      compilerArguments: buildSettings.compilerArguments
+    )
+  }
+
+  fileprivate var workingDirectoryPath: AbsolutePath? {
+    get throws {
+      guard let workingDirectory else {
+        return nil
+      }
+      return try AbsolutePath(validating: workingDirectory)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


rintaro:
@swift-ci Please test

ahoppen (commented):
I think we only use `compilerArguments` and `workingDirectory` from `FileBuildSettings`. Technically `language` and `isFallback` also don’t influence the batching directly (both are handled earlier on). Would it make sense to define a different struct to use as the batching key?

Otherwise looks good to me.

rintaro:
I ended up with a major modification of `UpdateIndexStoreTaskDescription`.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -90,22 +90,52 @@ package struct FileAndOutputPath: Sendable, Hashable {
...
+    package var language: Language
+    package var workingDirectory: String?
+    package var compilerArguments: [String]
+  }
 
-  var buildSettings: FileBuildSettings {
-    switch self {
-    case .multipleFiles(_, let buildSettings): return buildSettings
-    case .singleFile(_, let buildSettings): return buildSettings
+  /// A file to index, together with the index unit output path that should be recorded for it.
```
ahoppen:
```suggestion
  /// The files to index, together with the index unit output path that should be recorded for it.
```

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -90,22 +90,52 @@ package struct FileAndOutputPath: Sendable, Hashable {
...
+extension UpdateIndexStorePartition.Invocation {
+  fileprivate init(_ buildSettings: FileBuildSettings) {
+    self.init(
+      language: buildSettings.language,
+      workingDirectory: buildSettings.workingDirectory,
+      compilerArguments: buildSettings.compilerArguments
+    )
+  }
+
+  fileprivate var workingDirectoryPath: AbsolutePath? {
```
ahoppen, line 133: Now that this exists on `Invocation`, the `fileprivate extension FileBuildSettings { var workingDirectoryPath }` at the bottom of the file is dead — its only caller moved here. Can we delete it?

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -289,7 +319,7 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
     // the batch can be indexed by a single compiler invocation. However, we might discover that two Swift files within
     // the same target have different build settings in the build server. In that case, the best thing we can do is
     // trigger two compiler invocations.
-    var swiftFileInfosByBuildSettings: [FileBuildSettings: [(file: FileToIndex, outputPath: String)]] = [:]
+    var swiftBatches: [UpdateIndexStorePartition.Invocation: [(file: FileToIndex, indexUnitOutputPath: String)]] = [:]
```
ahoppen: `batches` already means something specific in this file — `UpdateIndexStoreTaskDescription.batches(toIndex:)` splits files into scheduling batches, which is exactly what the comment right above this line is talking about. Using it for the grouping-by-normalized-invocation makes the two read like the same concept. How about `swiftFilesByInvocation`?

That comment above also still refers to `fileInfosByBuildSettings` — could you update it to whatever name we settle on?

Also affects `sortedBatches` below.

rintaro:
@swift-ci test

rintaro:
@swift-ci test Windows

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -372,35 +412,43 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
             )
           } catch {
-            logger.error("Updating index store for \(fileInfo.mainFile.forLogging) failed: \(error.forLogging)")
-            BuildSettingsLogger.log(settings: partition.buildSettings, for: fileInfo.mainFile)
+            logger.error("Updating index store for \(indexFile.mainFile.forLogging) failed: \(error.forLogging)")
+            BuildSettingsLogger.log(
+              compilerArguments: partition.invocation.compilerArguments,
+              workingDirectory: partition.invocation.workingDirectory,
+              for: [indexFile.mainFile]
+            )
```
ahoppen: Just a thought: Would it make sense to add an extension to `BuildSettingsLogger` in this file that takes an `Invocation` to shorten these calls. Don’t too strong of an opinion though.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
```diff
@@ -456,21 +511,34 @@ package struct UpdateIndexStoreTaskDescription: IndexTaskDescription {
...
         "-index-file",
         // batch mode is not compatible with -index-file
         "-disable-batch-mode",
       ]
-    args = try await addOrReplaceIndexStorePath(in: args, for: partition.files.map(\.mainFile))
+    let indexFiles = partition.indexedFiles.map(\.mainFile)
+    args = try await addOrReplaceIndexStorePath(in: args, for: indexFiles)
+
+    // The output file map is only needed while the compiler runs. Deferring its removal to the end of this function
+    // (instead of the end of the `switch` case) keeps it alive until the indexing process has finished reading it.
```
ahoppen: I’m not sure if this comment is really needed. It makes sense in terms of this PR but it’s pretty self-explanatory going forward, I’d say.

ahoppen (approved):
Just a few stylistic nitpicks. Feel free to merge after addressing them.

Sources/SemanticIndex/UpdateIndexStoreTaskDescription.swift
rintaro: `fileInfosByBuildSettings` it is

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

PR #2738, opened 2026-07-28
Add missing semantic token modifier `parameterLabel` to documentation

jansorg:
Sync the semantic token modifiers with https://github.com/swiftlang/swift-tools-protocols/blob/main/Sources/LanguageServerProtocol/SupportTypes/SemanticTokenModifiers.swift

A commit in swift-tools-protocols added `parameterLabel`, which is not yet documented here.

As far as I understand, both should be in sync.
I was unable to actually see `parameterLabel` being returned in the list of available modifiers with Swift 6.3.3. But that may be caused by just testing with Objective-C files?

### DIFF
--- a/Contributor Documentation/LSP Extensions.md
+++ b/Contributor Documentation/LSP Extensions.md
@@ -59,6 +59,8 @@ functionScope = 'functionScope'
 classScope = 'classScope'
 fileScope = 'fileScope'
 globalScope = 'globalScope'
+//  Marks a token as a parameter label in a function signature or call.
+parameterLabel = 'parameterLabel'
 ```
 
 ## Semantic token types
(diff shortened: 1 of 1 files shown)


Contributor Documentation/LSP Extensions.md
```diff
@@ -59,6 +59,8 @@ functionScope = 'functionScope'
 classScope = 'classScope'
 fileScope = 'fileScope'
 globalScope = 'globalScope'
+//  Marks a token as a parameter label in a function signature or call.
+parameterLabel = 'parameterLabel'
```
rintaro: This is not from clangd. could you add a separate paragraph for this? Following "Semantic token types"

````suggestion
```

Added case

```ts
/**
 * Marks a token as a parameter label in a function signature or call.
 */
parameterLabel = 'parameterLabel'
````

jansorg:
Thanks for the suggestion. It's applied now.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Windows failure is unrelated. Let’s try again

@swift-ci Please test Windows

PR #2740, opened 2026-08-04
Add support for pkgconfig option when using Swift Build BSP backend

owenv:
I forgot to add this when the option was initially added, we just need to forward --pkg-config-path along when we launch the build server


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

ahoppen (approved):

PR #2741, opened 2026-08-04
Test compatibility updates for Swift Build BSP backend

owenv:
- Update an output path assertion which differs between backends
- Update two tests to build before making cross-module definition requests in an Objective-C package. Previously, the native build system would construct module maps at build planning time. Swift Build generates these during the build (or preparation build). This is a behavioral change, but the new behavior is in general much better for performance and the correctness of incremental builds, so I don't think we want to try applying some hack to avoid the build/preparation here.

### DIFF
--- a/Tests/SourceKitLSPTests/DefinitionTests.swift
+++ b/Tests/SourceKitLSPTests/DefinitionTests.swift
@@ -232,7 +232,7 @@ class DefinitionTests: SourceKitLSPTestCase {
     let project = try await SwiftPMTestProject(
       files: [
         "LibA/include/LibA.h": """
-        @interface 1️⃣LibAClass2️⃣
+        @interface 1️⃣LibAClass
         - (void)doSomething;
         @end
         """,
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

owenv:
There's some inconsistency in whether definition requests return a range or position depending on if the index is populated I need to sort out here

owenv:
Should be fixed now. The main diff between the old and new build systems is that the old one wrote module maps to disk as a side effect of requesting compiler args, while the new one properly integrates them into the build/prepare operation. I did have to update a couple tests to run a build to lay down those module maps, but I think that should be ok

owenv:
@swift-ci test

Tests/SourceKitLSPTests/DefinitionTests.swift
```diff
@@ -259,13 +259,18 @@ class DefinitionTests: SourceKitLSPTestCase {
         )
         """
     )
+    try await SwiftPMTestProject.build(at: project.scratchDirectory)
```
ahoppen: Instead of building the project, could you enable background indexing instead? That should be more performant than a full build.

owenv: sgtm

owenv:
@swift-ci test

owenv:
@swift-ci test windows

ahoppen (approved):

PR #2743, opened 2026-08-06
Use '/' to separate group name levels in sourcekitd requests

rintaro:
`workspace/symbol` and `workspace/symbolInfo` return SDK symbols as generated-interface reference documents whose module and group are derived from the module name that the index records. The index spells the levels of a group with `.`, recording the stdlib's `Math/Integers` group as the module name `Swift.Math.Integers`, but sourcekitd matches `key.groupname` against the group name as the module stores it, which separates levels with `/`. The dotted spelling matched no group, so the interface was generated without the group's declarations and `workspaceSymbol/resolve` could not locate the symbol in it, leaving the resolved position at the start of an interface that doesn't contain the symbol.

Translate `.` back to `/` when splitting the index module name into its module and group. Rendering a group for display keeps using `.`.

### DIFF
--- a/Sources/SourceKitLSP/SourceKitLSPServer.swift
+++ b/Sources/SourceKitLSP/SourceKitLSPServer.swift
@@ -1674,7 +1674,10 @@ extension SourceKitLSPServer {
     guard fullModuleName.hasPrefix(swiftModulePrefix) else {
       return (fullModuleName, nil)
     }
-    return ("Swift", String(fullModuleName.dropFirst(swiftModulePrefix.count)))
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

hamishknight (approved):

rintaro:
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift-format/pull/1261
@swift-ci test macOS

PR #2745, opened 2026-08-07
[Test] Limit several external build server tests to 6.5+ toolchains

rintaro:
`SwiftPMBuildServerTests` runs most of its tests against both the native and the out-of-process SwiftPM build server. Three of them depend on behavior of the out-of-process build server that was added during Swift 6.5 development and thus fail with a Swift 6.4 or earlier toolchain.

- `testBuildSetup` expects `-O` for a package configured for release, which requires the build server to prepare a target with the package's optimization level instead of the index build's. (swiftlang/swift-build#1470)
- `testSymlinkInWorkspaceSwift` and `testSymlinkInWorkspaceCXX` expect the paths reported for a package opened through a symlink to preserve that symlink, which SourceKit-LSP requests using `--experimental-skip-resolving-package-paths`. (swiftlang/swift-package-manager#10199)

Turn `buildServerOptionsToTest` into a function that takes the Swift version a test requires and only includes the out-of-process build server if the toolchain used for testing is at least that version.

### DIFF
--- a/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
+++ b/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
@@ -59,35 +59,16 @@ fileprivate extension SourceKitLSPOptions {
   }
 }
 
-let swiftPMHasExperimentalBuildServer: Bool = {
-  @Sendable func impl() async throws -> Bool {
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

owenv (approved):

PR #2746, opened 2026-08-08
Move workspace symbol implementation out of SourceKitLSPServer.swift

rintaro:
NFC

The `workspace/symbol`, `workspace/symbolNames`, `workspace/symbolInfo` and `workspaceSymbol/resolve` implementations made up roughly 470 lines of SourceKitLSPServer.swift. Move them to WorkspaceSymbols.swift and leave each request handler in SourceKitLSPServer.swift as a single statement that forwards to its implementation.

Fold QualifiedWorkspaceSymbolQuery.swift into the same file, since the query type and its fuzzy matching helper are only used by the qualified `workspace/symbol` search path.

Also remove `languageClass(for:)`, whose last caller went away when language service instances became scoped per workspace.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SourceKitLSP/WorkspaceSymbols.swift
ahoppen, line 1: In other files (eg. `Rename.swift` and `CodeCompletion.swift`) we extend `SourceKitLSPServer`/`SwiftLanguageService` to add the request handler method that you left in `SourceKitLSPServer`. I don’t have a strong preference to either way of approaching request handles implemented in other files (I kind of like what you did) but I think we should be consistent with the approach taken.

rintaro: I agree this is inconsistent with others.
When I was reading through the code, I found it a bit of a pain that some handlers are in the main file while others are in separate extension files. I think keeping the handler methods in the main file makes the code easier to follow.

So let me make the others consistent with this in follow-up PRs.

ahoppen: Great. Thanks for cleaning it up 👍🏽

ahoppen (approved):

PR #2750, opened 2026-08-11
Update the compiler diagnostic in `PullDiagnosticsTests.testNoteInSecondaryFile()`

tshortli:
The location and highlights for Swift compiler diagnostics about unavailable declarations are changing in https://github.com/swiftlang/swift/pull/91369. The `PullDiagnosticsTests.testNoteInSecondaryFile()` test would start failing as a result, so update the test to verify the same fundamental behavior as it did before but using a different compiler diagnostic that isn't changing.


tshortli:
@swift-ci please test

ahoppen (approved):

tshortli:
@swift-ci please test Windows

PR #2751, opened 2026-08-12
Replace SwiftPMTestHooks with more general build server test hooks

owenv:
The SwiftPM test hooks were relatively tightly coupled to the builtin build server. This PR replaces them with the more general build server test hooks that are also supported by the SwiftPM BSP backend, and updates tests accordingly. A few of the test updates might need some scrutiny, I'll leave some comments below.

### DIFF
--- a/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
+++ b/Sources/BuildServerIntegration/SwiftPMBuildServer.swift
@@ -93,12 +93,12 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     case cannotDetermineHostToolchain
   }
 
+  package static let packageReloadingTaskID = TaskId(id: "package-reloading")
+
   // MARK: Integration with SourceKit-LSP
 
   /// Options that allow the user to pass extra compiler flags.
   private let options: SourceKitLSPOptions
-
-  private let testHooks: SwiftPMTestHooks
 
   /// The queue on which we reload the package to ensure we don't reload it multiple times concurrently, which can cause
   /// issues in SwiftPM.
@@ -230,8 +230,7 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     projectRoot: URL,
     toolchainRegistry: ToolchainRegistry,
     options: SourceKitLSPOptions,
-    connectionToSourceKitLSP: any Connection,
-    testHooks: SwiftPMTestHooks
+    connectionToSourceKitLSP: any Connection
   ) async throws {
     self.projectRoot = projectRoot
     self.options = options
@@ -247,7 +246,6 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
     }
 
     self.toolchain = toolchain
-    self.testHooks = testHooks
     self.connectionToSourceKitLSP = connectionToSourceKitLSP
 
     // Start an open-ended log for messages that we receive during package loading. We never end this log.
@@ -494,18 +492,16 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
 
     self.connectionToSourceKitLSP.send(
       TaskStartNotification(
-        taskId: TaskId(id: "package-reloading"),
+        taskId: Self.packageReloadingTaskID,
         data: WorkDoneProgressTask(title: "SourceKit-LSP: Reloading Package").encodeToLSPAny()
       )
     )
-    await testHooks.reloadPackageDidStart?()
     defer {
       signposter.endInterval("Reloading package", state)
       Task {
         self.connectionToSourceKitLSP.send(
-          TaskFinishNotification(taskId: TaskId(id: "package-reloading"), status: .ok)
+          TaskFinishNotification(taskId: Self.packageReloadingTaskID, status: .ok)
         )
-        await testHooks.reloadPackageDidFinish?()
       }
     }
 
--- a/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
+++ b/Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
@@ -162,13 +162,15 @@ struct SwiftPMBuildServerTests {
         ]
       )
       let packageRoot = tempDir.appending(component: "pkg")
+      let connection = LocalConnection(receiverName: "dummy")
+      let handler = IgnoringMessageHandler()
+      connection.start(handler: handler)
       await expectThrowsError(
         try await SwiftPMBuildServer(
           projectRoot: packageRoot,
           toolchainRegistry: ToolchainRegistry(toolchains: []),
           options: SourceKitLSPOptions(),
-          connectionToSourceKitLSP: LocalConnection(receiverName: "dummy"),
-          testHooks: SwiftPMTestHooks()
+          connectionToSourceKitLSP: connection
         )
       )
     }
@@ -198,12 +200,14 @@ struct SwiftPMBuildServerTests {
         ),
         backgroundIndexing: false
       )
+      let connection = LocalConnection(receiverName: "dummy")
+      let handler = IgnoringMessageHandler()
+      connection.start(handler: handler)
       let swiftpmBuildServer = try await SwiftPMBuildServer(
         projectRoot: packageRoot,
         toolchainRegistry: .forTesting,
         options: options,
-        connectionToSourceKitLSP: LocalConnection(receiverName: "dummy"),
-        testHooks: SwiftPMTestHooks()
+        connectionToSourceKitLSP: connection
       )
 
       let dataPath = await swiftpmBuildServer.destinationBuildParameters.dataPath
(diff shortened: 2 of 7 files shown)


owenv:
@swift-ci test

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -494,19 +492,15 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
       )
     )
-    await testHooks.reloadPackageDidStart?()
     defer {
       signposter.endInterval("Reloading package", state)
-      Task {
-        self.connectionToSourceKitLSP.send(
-          TaskFinishNotification(taskId: TaskId(id: "package-reloading"), status: .ok)
-        )
-        await testHooks.reloadPackageDidFinish?()
```
owenv, line 508: A few tests were waiting on semaphores in this reloadPackageDidFinish hook, which I think was effectively a no-op since this is only ever called at the end of this task, so it shouldn't have been blocking anything

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -1144,19 +1142,11 @@ struct SwiftPMBuildServerTests {
...
-    let didReceiveWorkDoneProgressNotification = WrappedSemaphore(name: "work done progress received")
     let project = try await SwiftPMTestProject(
       files: [
         "MyLibrary/Test.swift": ""
       ],
       capabilities: ClientCapabilities(window: WindowClientCapabilities(workDoneProgress: true)),
-      hooks: Hooks(
-        buildServerHooks: BuildServerHooks(
-          swiftPMTestHooks: SwiftPMTestHooks(reloadPackageDidStart: {
-            didReceiveWorkDoneProgressNotification.waitOrXCTFail()
```
owenv, line 1156: This wait didn't seem necessary, but I could be missing something

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
```diff
@@ -1762,31 +1765,19 @@ final class BackgroundIndexingTests: SourceKitLSPTestCase {
 
   func testPauseBackgroundIndexing() async throws {
     try SkipUnless.longTestsEnabled()
-    let backgroundIndexingPaused = WrappedSemaphore(name: "Background indexing was paused")
-    let hooks = Hooks(
-      buildServerHooks: BuildServerHooks(
-        swiftPMTestHooks: SwiftPMTestHooks(
-          reloadPackageDidFinish: {
-            backgroundIndexingPaused.waitOrXCTFail()
```
owenv, line 1770: These are the tests where I'm pretty sure 
```
reloadPackageDidFinish: {
            backgroundIndexingPaused.waitOrXCTFail()
}
```

was a no-op

Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
```diff
@@ -287,16 +288,18 @@ final class SwiftPMIntegrationTests: SourceKitLSPTestCase {
...
       files: [
         "Test.swift": ""
       ],
       hooks: Hooks(
         buildServerHooks: BuildServerHooks(
-          swiftPMTestHooks: SwiftPMTestHooks(reloadPackageDidStart: {
-            receivedDocumentSymbolsReply.waitOrXCTFail()
-          })
+          preHandleRequest: { request in
+            if request is TextDocumentSourceKitOptionsRequest {
```
owenv, line 299: This is a slightly different way of forcing use of fallback build settings, but I think it should achieve the same result

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
```diff
@@ -494,19 +492,15 @@ package actor SwiftPMBuildServer: BuiltInBuildServer {
...
-      }
+      self.connectionToSourceKitLSP.send(
+        TaskFinishNotification(taskId: Self.packageReloadingTaskID, status: .ok)
+      )
```
ahoppen: We need to support building SourceKit-LSP using the latest released Xcode version, which currently has Swift 6.3, so we can’t use async defers yet.

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
ahoppen: If my memory is correct this is needed because otherwise the SwiftPM load could finish before the `WorkDoneProgress` manager has sent the work done progress start before the debounce. It would then not send any work done progress notifications at all.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
ahoppen: This is needed because otherwise the package reload may finish before the `SetOptionsRequest(backgroundIndexingPaused: true)` is handled by SourceKit-LSP and we would thus start background indexing. It’s unlikely that package loading is faster than that request handling but we’ve seen very weird timing races in CI, so it’s better to be safe.

Same below

Tests/SourceKitLSPTests/SwiftPMIntegrationTests.swift
ahoppen: Yes, looks good to me.

ahoppen (commented):
Thanks for loosening that coupling. I think a few of these semaphore are needed though.

Tests/SourceKitLSPTests/BackgroundIndexingTests.swift
owenv: Does blocking in reloadPackageDidFinish actually prevent the reload from completing though? The TaskFinishNotification for package-reloading is sent before the hook is called and then the hook blocks at the end of an unstructured task.

ahoppen: Hmm, I suppose it used to block but it doesn’t anymore with the unstructured task. Good to remove it then

Sources/BuildServerIntegration/SwiftPMBuildServer.swift
owenv: Ah, good catch, I put the wrapping Task {} back here

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
owenv: Makes sense. I used the new notification hook to delay delivery of the TaskFinishedNotification here, which should keep this behavior

owenv:
@swift-ci test

owenv:
@swift-ci test windows

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -1386,7 +1387,7 @@ struct SwiftPMBuildServerTests {
   private func makeServerWithBinaryTargetAndWaitForInitialLoad(
     in tempDir: URL,
     options: SourceKitLSPOptions = SourceKitLSPOptions(),
-    reloadPackageDidStart: (@Sendable () async -> Void)? = nil
+    connectionToSourceKitLSP: (any Connection)? = nil
```
ahoppen: Instead of passing in `nil` here and then defaulting it to `LocalConnection` can be more explicit about what not passing a value for `connectionToSourceKitLSP` does by changing this to


```suggestion
    connectionToSourceKitLSP: (any Connection) = LocalConnection(receiverName: "dummy")
```

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
```diff
@@ -1453,18 +1453,13 @@ struct SwiftPMBuildServerTests {
...
+      let packageReloads = PackageReloadCountingMessageHandler()
 
       let (server, projectRoot) = try await makeServerWithBinaryTargetAndWaitForInitialLoad(
         in: tempDir,
-        reloadPackageDidStart: {
-          if packageInitialized.value {
-            unexpectedReloadStarted.withLock { $0 = true }
-          }
-        }
+        connectionToSourceKitLSP: LocalConnection(receiverName: "dummy", handler: packageReloads)
```
ahoppen, line 1468: I’m a little surprised this works. This will create a `LocalConnection` that has been started (`start` is called by this initializer) while the connection isn’t started by calling `LocalConnection(receiverName: "dummy")`. This doesn’t cause any issues at the moment because:
- `send(_ notification:)` doesn’t actually check `state` (other than `send(_ request:)`)
- We only send notifications at the moment, so we don’t hit the `precondition in `send(_ request:)`.

Would you be able to ensure the connection is in the same state for both branches? For bonus points, check the state in `LocalConnection.send(_ notification:)` as well 😉

owenv: After some refactoring I did earlier, the default value wasn't actually ever used, so I just removed it. Will open a swift-tools-protocols PR to add the missing assert though

owenv:
@swift-ci test

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
owenv: https://github.com/swiftlang/swift-tools-protocols/pull/78

owenv:
@swift-ci test windows

Tests/BuildServerIntegrationTests/SwiftPMBuildServerTests.swift
ahoppen: I suppose this will fail once https://github.com/swiftlang/swift-tools-protocols/pull/78 is merged because you don’t start the `LocalConnection` eg. in line 170. But that can be addressed with https://github.com/swiftlang/swift-tools-protocols/pull/78 if you want.

ahoppen (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test windows

owenv:
@swift-ci test windows

PR #2753, opened 2026-08-19
Use file names instead of full paths in the index log title

rintaro:
The index log title listed the full path of every file in an indexing build. Since a single build can contain up to 25 files, the title grew extremely long. List only the file names; the full paths are still visible in the logged compiler invocation.


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2754, opened 2026-08-20
Add inline variable refactoring

### LINKED ISSUE #2507: Add Inline temp variable code action

somiljain2006:
Fixes #2507 

Implements Inline Variable for local let declarations with lexical lookup

### DIFF
--- a/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
+++ b/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
@@ -26,23 +26,28 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
 
   package static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> PatternBindingSyntax? {
     guard let node = scope.innermostNodeContainingRange else { return nil }
-
-    let variableDecl =
-      node.ancestorOrSelf(mapping: { $0.as(VariableDeclSyntax.self) })
-      ?? node.as(CodeBlockItemSyntax.self)?.item.as(VariableDeclSyntax.self)
-
-    guard let variableDecl = variableDecl,
-      variableDecl.bindingSpecifier.tokenKind == .keyword(.let)
-    else {
-      return nil
+    var current: Syntax? = Syntax(node)
+    while let syntax = current {
+      if let binding = syntax.as(PatternBindingSyntax.self) {
+        guard let variableDecl = binding.parent?.parent?.as(VariableDeclSyntax.self),
+          variableDecl.bindingSpecifier.tokenKind == .keyword(.let)
+        else {
+          return nil
+        }
+        return binding
+      }
+      if let variableDecl = syntax.as(VariableDeclSyntax.self) {
+        guard variableDecl.bindingSpecifier.tokenKind == .keyword(.let) else {
+          return nil
+        }
+        return variableDecl.bindings.first
+      }
+      if syntax.is(CodeBlockSyntax.self) || syntax.is(MemberBlockSyntax.self) {
+        return nil
+      }
+      current = syntax.parent
     }
-
-    // If the cursor is on the identifier or value, return the specific pattern binding.
-    if let specificBinding = node.ancestorOrSelf(mapping: { $0.as(PatternBindingSyntax.self) }) {
-      return specificBinding
-    }
-
-    return variableDecl.bindings.first
+    return nil
   }
 
   package static func textRefactor(syntax binding: PatternBindingSyntax, in context: Void) throws -> [SourceEdit] {
@@ -57,7 +62,7 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
       throw RefactoringNotApplicableError("Could not parse the target identifier")
     }
 
-    guard let blockScope = variableDecl.ancestorOrSelf(mapping: { $0.as(CodeBlockItemListSyntax.self) }) else {
+    guard let blockScope = variableDecl.parent?.parent?.as(CodeBlockItemListSyntax.self) else {
       throw RefactoringNotApplicableError("Could not find the lexical scope for this variable")
     }
 
@@ -73,7 +78,6 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     }
 
     var edits: [SourceEdit] = []
-
     for usageNode in usageSites {
       let replacementText = formatReplacement(
         initializer: initializer,
@@ -97,15 +101,12 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
       // remove only the target binding and clean up the trailing commas.
       var newBindingsArray = Array(variableDecl.bindings)
       newBindingsArray.removeAll { $0.id == binding.id }
-
       if !newBindingsArray.isEmpty {
         newBindingsArray[0].leadingTrivia = []
         newBindingsArray[newBindingsArray.count - 1].trailingComma = nil
       }
-
       var newDecl = variableDecl.with(\.bindings, PatternBindingListSyntax(newBindingsArray))
       newDecl.leadingTrivia = []
-
       let range = variableDecl.positionAfterSkippingLeadingTrivia..<variableDecl.endPosition
       edits.append(SourceEdit(range: range, replacement: newDecl.description))
     }
@@ -131,7 +132,19 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     usageNode: DeclReferenceExprSyntax
   ) -> String {
     if needsParentheses(initializer: initializer, usageNode: usageNode) {
-      return "(\(initializer.description.trimmingCharacters(in: .whitespacesAndNewlines)))"
+      let parenthesized = TupleExprSyntax(
+        leftParen: .leftParenToken(leadingTrivia: initializer.leadingTrivia),
+        elements: [
+          LabeledExprSyntax(
+            expression:
+              initializer
+              .with(\.leadingTrivia, [])
+              .with(\.trailingTrivia, [])
+          )
+        ],
+        rightParen: .rightParenToken(trailingTrivia: initializer.trailingTrivia)
+      )
+      return parenthesized.description
     }
     return initializer.description
   }
@@ -140,27 +153,49 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     initializer: ExprSyntax,
     usageNode: DeclReferenceExprSyntax
   ) -> Bool {
-    let root = usageNode.root
-    let fullText = root.description
-    let initializerText = initializer.description.trimmingCharacters(in: .whitespacesAndNewlines)
-
-    let startPos = usageNode.positionAfterSkippingLeadingTrivia.utf8Offset
-    let endPos = usageNode.endPositionBeforeTrailingTrivia.utf8Offset
-
-    let startIdx = fullText.utf8.index(fullText.utf8.startIndex, offsetBy: startPos)
-    let endIdx = fullText.utf8.index(fullText.utf8.startIndex, offsetBy: endPos)
-
-    var rewrittenText = fullText
-    rewrittenText.replaceSubrange(startIdx..<endIdx, with: initializerText)
-
-    let parsedFile = Parser.parse(source: rewrittenText)
-
-    if parsedFile.hasError {
+    guard let codeBlockItem = containingCodeBlockItem(for: Syntax(usageNode)) else {
       return true
     }
-
+    let scopeText = codeBlockItem.description
+    let initializerText = initializer.trimmedDescription
+    let scopeStartOffset = codeBlockItem.position.utf8Offset
+    let startPos =
+      usageNode.positionAfterSkippingLeadingTrivia.utf8Offset - scopeStartOffset
+    let endPos =
+      usageNode.endPositionBeforeTrailingTrivia.utf8Offset - scopeStartOffset
+    let startIdx = scopeText.utf8.index(
+      scopeText.utf8.startIndex,
+      offsetBy: startPos
+    )
+    let endIdx = scopeText.utf8.index(
+      scopeText.utf8.startIndex,
+      offsetBy: endPos
+    )
+    var rewrittenText = scopeText
+    rewrittenText.replaceSubrange(startIdx..<endIdx, with: initializerText)
+    let parsedScope = Parser.parse(source: rewrittenText)
+    if parsedScope.hasError {
+      return true
+    }
     let targetSpan = startPos..<(startPos + initializerText.utf8.count)
-    return !containsNode(in: Syntax(parsedFile), withSpan: targetSpan, matching: initializerText)
+    return !containsNode(
+      in: Syntax(parsedScope),
+      withSpan: targetSpan,
+      matching: initializerText
+    )
+  }
+
+  private static func containingCodeBlockItem(
+    for node: Syntax
+  ) -> CodeBlockItemSyntax? {
+    var current: Syntax? = node
+    while let syntax = current {
+      if let codeBlockItem = syntax.as(CodeBlockItemSyntax.self) {
+        return codeBlockItem
+      }
+      current = syntax.parent
+    }
+    return nil
   }
 
   static func containsNode(
@@ -168,27 +203,46 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     withSpan span: Range<Int>,
     matching text: String
   ) -> Bool {
+    let finder = TargetNodeFinder(targetSpan: span, targetText: text)
+    finder.walk(node)
+    return finder.found
+  }
+}
+
+/// A visitor that checks if a syntax node with a matching span and description exists.
+private final class TargetNodeFinder: SyntaxAnyVisitor {
+  let targetSpan: Range<Int>
+  let targetText: String
+  var found = false
+
+  init(targetSpan: Range<Int>, targetText: String) {
+    self.targetSpan = targetSpan
+    self.targetText = targetText
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
+    if found {
+      return .skipChildren
+    }
+
+    // Prune subtree traversal if full span does not overlap target range
+    let fullStart = node.position.utf8Offset
+    let fullEnd = node.endPosition.utf8Offset
+    if fullStart > targetSpan.upperBound || fullEnd < targetSpan.lowerBound {
+      return .skipChildren
+    }
+
     let nodeStart = node.positionAfterSkippingLeadingTrivia.utf8Offset
     let nodeEnd = node.endPositionBeforeTrailingTrivia.utf8Offset
-
-    if nodeStart == span.lowerBound && nodeEnd == span.upperBound {
-      if node.description.trimmingCharacters(in: .whitespacesAndNewlines) == text {
-        return true
+    if nodeStart == targetSpan.lowerBound && nodeEnd == targetSpan.upperBound {
+      if node.trimmedDescription == targetText {
+        found = true
+        return .skipChildren
       }
     }
 
-    for child in node.children(viewMode: .sourceAccurate) {
-      let childFullStart = child.position.utf8Offset
-      let childFullEnd = child.endPosition.utf8Offset
-
-      if childFullStart <= span.upperBound && childFullEnd >= span.lowerBound {
-        if containsNode(in: child, withSpan: span, matching: text) {
-          return true
-        }
-      }
-    }
-
-    return false
+    return .visitChildren
   }
 }
 
@@ -207,8 +261,7 @@ final class ReferenceCollector: SyntaxVisitor {
 
   override func visit(_ node: DeclReferenceExprSyntax) -> SyntaxVisitorContinueKind {
     guard let nodeIdentifier = Identifier(node.baseName),
-      nodeIdentifier == targetIdentifier,
-      node.position > targetBinding.endPosition
+      nodeIdentifier == targetIdentifier
     else {
       return .visitChildren
     }
@@ -218,19 +271,16 @@ final class ReferenceCollector: SyntaxVisitor {
       switch result {
       case .fromScope(_, let names):
         let matchesTarget = names.contains { name in
-          name.syntax.id == targetBinding.id || name.syntax.id == targetBinding.pattern.id
+          name.syntax.id == targetBinding.pattern.id
         }
-
         if matchesTarget {
           candidates.append(node)
-          return .visitChildren
         }
         return .visitChildren
       default:
         continue
       }
     }
-
     return .visitChildren
   }
 }
--- a/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
+++ b/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
@@ -214,9 +214,7 @@ private final class TargetNodeFinder: SyntaxAnyVisitor {
       return .skipChildren
     }
 
-    // Prune subtree traversal if full span does not overlap target range
-    let fullStart = node.position
-    let fullEnd = node.endPosition
+    // Prune subtree traversal if the node does not contain the target span.
     guard node.range.contains(targetSpan) else {
       return .skipChildren
     }
--- a/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
+++ b/Sources/SwiftSyntaxCodeActions/InlineVariable.swift
@@ -16,7 +16,7 @@ internal import SourceKitLSP
 import SwiftBasicFormat
 import SwiftExtensions
 @_spi(Experimental) import SwiftLexicalLookup
-import SwiftOperators
+import SwiftParser
 import SwiftRefactor
 import SwiftSyntax
 
@@ -28,7 +28,7 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     guard let node = scope.innermostNodeContainingRange else { return nil }
 
     let variableDecl =
-      node.firstMatch(ofType: VariableDeclSyntax.self)
+      node.ancestorOrSelf(mapping: { $0.as(VariableDeclSyntax.self) })
       ?? node.as(CodeBlockItemSyntax.self)?.item.as(VariableDeclSyntax.self)
 
     guard let variableDecl = variableDecl,
@@ -38,7 +38,7 @@ struct InlineVariable: SyntaxRefactoringCodeActionProvider {
     }
 
     // If the cursor is on the identifier or value, return the specific pattern binding.
-    if let specificBinding = node.firstMatch(ofType: PatternBindingSyntax.self) {
+    if let specificBinding = node.ancestorOrSelf(mapping: { $0.as(PatternBindingSyntax.self) }) {
       return specificBinding
     }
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+extension SyntaxProtocol {
+  /// Traverses up the syntax tree to find the first ancestor of the specified type.
+  func firstMatch<T: SyntaxProtocol>(ofType type: T.Type) -> T? {
+    var current: Syntax? = Syntax(self)
+    while let node = current {
+      if let match = node.as(T.self) { return match }
+      current = node.parent
+    }
+    return nil
+  }
```
ahoppen: We should have `ancestorOrSelf` for this already.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+
+      let range = variableDecl.positionAfterSkippingLeadingTrivia..<variableDecl.endPosition
+      edits.append(SourceEdit(range: range, replacement: newDecl.description))
+    }
+
+    return edits
+  }
+
+  private static func findValidUsages(
+    targetIdentifier: String,
```
ahoppen: Instead of using `String` here, we should base all of this on top of `Identifier`, which should also make sure that we match `let foo = 2` with a usage that spells the variable name in backticks like ``print(`foo`)``.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+    targetBinding: PatternBindingSyntax,
+    in scope: CodeBlockItemListSyntax
+  ) -> [DeclReferenceExprSyntax] {
+    let finder = ReferenceCollector(targetIdentifier: targetIdentifier)
```
ahoppen: Instead of a two-part process, could `ReferenceCollector` directly use SwiftLexicalLookup to check if the reference refers to the variable in question?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+    return finder.candidates.filter { usageNode in
+      guard usageNode.position > targetBinding.endPosition else { return false }
+
+      let lookupResults = usageNode.lookup(Identifier(usageNode.baseName))
+
+      for result in lookupResults {
+        switch result {
+        case .fromScope(_, let names):
+          let matchesTarget = names.contains { name in
+            if let resolvedBinding = name.syntax.firstMatch(ofType: PatternBindingSyntax.self) {
```
ahoppen: Do we need to walk up the tree to a `PatternBindingSyntax` here or could we compare `name.syntax` directly? I’m wondering because walking up the tree could theoretically find a completely unrelated `PatternBindingSyntax`.

Same for the check below of `VariableDeclSyntax`.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+        }
+      }
+      return false
+    }
+  }
+
+  /// Determines whether it is safe to inline the initializer at multiple usage sites.
+  /// If the initializer has side effects (like a function call), duplicating it could
+  /// change the program's behavior.
+  private static func canDuplicateInitializer(
```
ahoppen: Moving the location at which the initializer is called may already change program behavior. If the user invokes the refactoring action, they are opting into moving the initializer around (and potentially duplicating them), so I don’t think we should try to be too clever here. There may be good reasons to inline variables initialized by a function call because the developer knows that it’s save to duplicate.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+    return initializer.description
+  }
+
+  static func isPrimaryExpression(_ expr: ExprSyntax) -> Bool {
+    expr.is(IntegerLiteralExprSyntax.self) || expr.is(FloatLiteralExprSyntax.self)
+      || expr.is(BooleanLiteralExprSyntax.self) || expr.is(StringLiteralExprSyntax.self)
+      || expr.is(NilLiteralExprSyntax.self) || expr.is(DeclReferenceExprSyntax.self)
+      || expr.is(FunctionCallExprSyntax.self) || expr.is(MemberAccessExprSyntax.self)
+      || expr.is(SubscriptCallExprSyntax.self) || expr.is(ArrayExprSyntax.self) || expr.is(DictionaryExprSyntax.self)
+      || expr.is(TupleExprSyntax.self) || expr.is(ClosureExprSyntax.self)
```
ahoppen: I think this would read a little nicer as a switch over `expr.as(ExprSyntaxEnum.self)`.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+    guard let parent = usageNode.parent else { return false }
+
+    if parent.is(MemberAccessExprSyntax.self) || parent.is(SubscriptCallExprSyntax.self)
+      || parent.is(FunctionCallExprSyntax.self) || parent.is(PrefixOperatorExprSyntax.self)
+      || parent.is(PostfixOperatorExprSyntax.self)
+    {
+      return !isPrimaryExpression(initializer)
+    }
+
+    let parentSeq = parent.as(SequenceExprSyntax.self) ?? parent.parent?.as(SequenceExprSyntax.self)
```
ahoppen: Why do we need to check the parent and grandparent but not great-grandparent?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+
+    guard let parentSeq = parentSeq else {
```
ahoppen:
```suggestion
    guard let parentSeq else {
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+      return !isPrimaryExpression(initializer)
+    }
+
+    var newElements: [ExprSyntax] = []
+    var replaced = false
+
+    for element in parentSeq.elements {
+      if element.id == usageNode.id {
```
ahoppen: This doesn’t work if `parentSeq = parent.parent?.as(SequenceExprSyntax.self)`, right?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,341 @@
...
+      || expr.is(TupleExprSyntax.self) || expr.is(ClosureExprSyntax.self)
+  }
+
+  // Determine whether replacing the reference with the initializer would
+  // change the expression's operator grouping.
+  static func needsParentheses(
+    initializer: ExprSyntax,
+    usageNode: DeclReferenceExprSyntax,
+    operatorTable: OperatorTable
+  ) -> Bool {
```
ahoppen, line 155: There are also other places in which parentheses need to be added outside of operator precedence. For example

```swift
let x = try three()
let result = 1 + x
```

but the following is invalid


```swift
let result = 1 + try three()
```

What do you think about the following approach instead: You parse the rewritten source code without parentheses and locate the node at the replaced offset. You then check if that node has the exact same text representation as the initializer that was inserted. If the rewritten text doesn’t parse or you don’t find the node, try again with parentheses.

For example if you replace `2 * x` by `2 * 1 + 1`, then you won’t be able to find any node in the syntax tree that represents `1 + 1`, so you know that parentheses need to be added.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,625 @@ final class CodeActionTests: SourceKitLSPTestCase {
     )
   }
 
+  func testInlineVariableBasic() async throws {
+    try await assertCodeActions(
+      """
+      func doSomething() {1️⃣
+        let 0️⃣x = 422️⃣
```
ahoppen: We generally use `1️⃣` as the first location marker instead of `0️⃣`. Could you change these tests to be consistent with that pattern?

ahoppen (commented):
Thanks for tackling this, @somiljain2006. I left a few suggestions inline.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+/// A code action that replaces a local variable with its assigned value at all usage sites.
+struct InlineVariable: SyntaxRefactoringCodeActionProvider {
+  package static let title: String = "Inline variable"
+
+  package static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> PatternBindingSyntax? {
+    guard let node = scope.innermostNodeContainingRange else { return nil }
+
+    let variableDecl =
+      node.ancestorOrSelf(mapping: { $0.as(VariableDeclSyntax.self) })
+      ?? node.as(CodeBlockItemSyntax.self)?.item.as(VariableDeclSyntax.self)
```
ahoppen: Wouldn’t `node.as(CodeBlockItemSyntax.self)?.item.as(VariableDeclSyntax.self)` already imply that `node.ancestorOrSelf(mapping: { $0.as(VariableDeclSyntax.self) })` returned non-nil?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+
+/// A code action that replaces a local variable with its assigned value at all usage sites.
+struct InlineVariable: SyntaxRefactoringCodeActionProvider {
+  package static let title: String = "Inline variable"
+
+  package static func nodeToRefactor(in scope: SyntaxCodeActionScope) -> PatternBindingSyntax? {
+    guard let node = scope.innermostNodeContainingRange else { return nil }
+
+    let variableDecl =
+      node.ancestorOrSelf(mapping: { $0.as(VariableDeclSyntax.self) })
```
ahoppen: I think we should add `stoppingAt` for `CodeBlockSyntax` and `MemberBlockSyntax` so that we don’t offer this eg. in

```swift
let foo = {
  compute^HERE^Something()
}
```


```suggestion
      node.findParentOfSelf(
      ofType: VariableDeclSyntax.self,
      stoppingIf: { $0.is(CodeBlockSyntax.self) || $0.is(MemberBlockSyntax.self) }
    )
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    guard let variableDecl = variableDecl,
+      variableDecl.bindingSpecifier.tokenKind == .keyword(.let)
+    else {
+      return nil
+    }
+
+    // If the cursor is on the identifier or value, return the specific pattern binding.
+    if let specificBinding = node.ancestorOrSelf(mapping: { $0.as(PatternBindingSyntax.self) }) {
+      return specificBinding
+    }
```
ahoppen: Since we are walking the tree two separate times, I am a little cautious in case `specificBinding` would resolve to a different binding than the variable decl. I think it would happen in this case

```swift
var x = {
  ^HERE^let value = 2
  return value
}()
```

Could we do something like this instead:
1. First find a `PatternBindingDecl`
2. If that doesn’t exist find a `VariableDeclSyntax` and then pick the first pattern (this is so we can offer the refactoring action on `let`)
3. Now walk up from the `PatternBindingDecl` using `.parent.parent.parent` or something like that to fetch the corresponding `VariableDeclSyntax` and check that it is `let`.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+      let initializer = binding.initializer?.value
+    else {
+      throw RefactoringNotApplicableError("Could not extract variable information")
+    }
+
+    guard let targetIdentifier = Identifier(identifierPattern.identifier) else {
+      throw RefactoringNotApplicableError("Could not parse the target identifier")
+    }
+
+    guard let blockScope = variableDecl.ancestorOrSelf(mapping: { $0.as(CodeBlockItemListSyntax.self) }) else {
```
ahoppen: Instead of walking up the entire tree, could you just check if the grandparent  is a `CodeBlockItemListSyntax`. I just want to make sure that we don’t extract the incorrect item eg. in this case:


```swift
func foo() {
  struct Foo {
    let x^HERE^ = "hi"
  }
  print("wohoo")
}
```

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,389 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                TextEdit(
+                  range: positions["2️⃣"]..<positions["3️⃣"],
+                  newText: ""
+                ),
+              ]
+            ]
+          )
+        )
+      ]
+    }
```
ahoppen, line 4371: Could you also add a test case like the following that checks we match identifiers with backticks:

```swift
func foo() {
  let `x` = "hi"
  print(x)
  print(`x`)
}
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+  init(targetIdentifier: Identifier, targetBinding: PatternBindingSyntax) {
+    self.targetIdentifier = targetIdentifier
+    self.targetBinding = targetBinding
+    super.init(viewMode: .sourceAccurate)
+  }
+
+  override func visit(_ node: DeclReferenceExprSyntax) -> SyntaxVisitorContinueKind {
+    guard let nodeIdentifier = Identifier(node.baseName),
+      nodeIdentifier == targetIdentifier,
+      node.position > targetBinding.endPosition
```
ahoppen: Technically, this lexical check isn’t necessary or correct. For example, this compiles in top-level code (albeit, it mis-compiles – known issue – but still):

```swift
print(x)
let x = 1
```

If you want to have some fun, try to guess what it prints before executing it.

I would just trust `SwiftLexicalLookup` for these checks.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    else {
+      return .visitChildren
+    }
+
+    let lookupResults = node.lookup(nodeIdentifier)
+    for result in lookupResults {
+      switch result {
+      case .fromScope(_, let names):
+        let matchesTarget = names.contains { name in
+          name.syntax.id == targetBinding.id || name.syntax.id == targetBinding.pattern.id
```
ahoppen: Why do we need to check both `targetBinding.id` and `targetBinding.pattern.id`?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        if matchesTarget {
+          candidates.append(node)
+          return .visitChildren
```
ahoppen: I think you can avoid this and just use the `return` two lines below.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,389 @@ final class CodeActionTests: SourceKitLSPTestCase {
     )
   }
 
+  func testInlineVariableBasic() async throws {
+    try await assertCodeActions(
+      """
+      func doSomething() {2️⃣
+        let 1️⃣x = 423️⃣
+        print(4️⃣x5️⃣)
+      }
```
ahoppen: Nitpick: Could you order the position markers in-order inside the source code so that you invoke the refactoring action on `2`?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    finder.walk(scope)
+    return finder.candidates
+  }
+
+  static func formatReplacement(
+    initializer: ExprSyntax,
+    usageNode: DeclReferenceExprSyntax
+  ) -> String {
+    if needsParentheses(initializer: initializer, usageNode: usageNode) {
+      return "(\(initializer.description.trimmingCharacters(in: .whitespacesAndNewlines)))"
```
ahoppen: Instead of trimming the newlines, could you create a `TupleExprSyntax` and just transfer the leading and trailing trivia from `initializer` over to it?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    return initializer.description
+  }
+
+  static func needsParentheses(
+    initializer: ExprSyntax,
+    usageNode: DeclReferenceExprSyntax
+  ) -> Bool {
+    let root = usageNode.root
+    let fullText = root.description
+    let initializerText = initializer.description.trimmingCharacters(in: .whitespacesAndNewlines)
```
ahoppen: I think this won’t work if you have eg. 

```swift
let foo = /* comment */
  1 * 2
```

Instead of trimming characters from the string, could you inspect the trivia pieces instead? Or maybe you don’t need to trim the trivia at all.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    let startPos = usageNode.positionAfterSkippingLeadingTrivia.utf8Offset
+    let endPos = usageNode.endPositionBeforeTrailingTrivia.utf8Offset
+
+    let startIdx = fullText.utf8.index(fullText.utf8.startIndex, offsetBy: startPos)
+    let endIdx = fullText.utf8.index(fullText.utf8.startIndex, offsetBy: endPos)
+
+    var rewrittenText = fullText
+    rewrittenText.replaceSubrange(startIdx..<endIdx, with: initializerText)
+
+    let parsedFile = Parser.parse(source: rewrittenText)
```
ahoppen: I don’t think we need to parse the full source file. You could try and find the surrounding `CodeBlockItemSyntax` and then just re-parse that. This should be more efficient for large source files.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,236 @@
...
+    let nodeStart = node.positionAfterSkippingLeadingTrivia.utf8Offset
+    let nodeEnd = node.endPositionBeforeTrailingTrivia.utf8Offset
+
+    if nodeStart == span.lowerBound && nodeEnd == span.upperBound {
+      if node.description.trimmingCharacters(in: .whitespacesAndNewlines) == text {
+        return true
+      }
+    }
+
+    for child in node.children(viewMode: .sourceAccurate) {
```
ahoppen: You’ve effectively re-implemented a `SyntaxVisitor` here. I think it would be simpler to implement this as a `SyntaxAnyVisitor` instead.

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,389 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                TextEdit(range: positions["2️⃣"]..<positions["3️⃣"], newText: ""),
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testInlineVariableIgnoresShadowedReferences() async throws {
```
ahoppen: Isn’t this effectively already covered by `testInlineVariableRespectsShadowing`?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,389 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+            let x = 100
+            print(x)
+          }
+        }
+        """,
+      atMarker: "1️⃣"
+    )
+  }
+
+  func testInlineVariableRespectsShadowingBetweenValidReferences() async throws {
```
ahoppen: Isn’t this effectively already covered by `testInlineVariableRespectsShadowing`?

Tests/SourceKitLSPTests/CodeActionTests.swift
```diff
@@ -4027,6 +4027,389 @@ final class CodeActionTests: SourceKitLSPTestCase {
...
+                ),
+              ]
+            ]
+          )
+        )
+      ]
+    }
+  }
+
+  func testInlineVariableShadowingStartsAfterOuterReference() async throws {
```
ahoppen: I don’t think this adds too much value. You’re just testing one edge case of `SwiftLexicalLookup` here, which should be covered by the `SwiftLexicalLookup` tests.

ahoppen (commented):
Thanks! Looks a lot simpler now already. I have left a few more suggestions inline.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+      return parenthesized.description
+    }
+    return initializer.description
+  }
+
+  static func needsParentheses(
+    initializer: ExprSyntax,
+    usageNode: DeclReferenceExprSyntax
+  ) -> Bool {
+    guard let codeBlockItem = containingCodeBlockItem(for: Syntax(usageNode)) else {
```
ahoppen: Can’t this just be 
```suggestion
    guard let codeBlockItem = usageNode.ancestorOrSelf(mapping: { $0.as(CodeBlockItemSyntax.self) }) else {
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+      return true
+    }
+    let scopeText = codeBlockItem.description
+    let initializerText = initializer.trimmedDescription
```
ahoppen: In `formatReplacement`, we return `initializer.description`, so we should also be using `initializer.description` here instead of `trimmedDescription`.

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+      scopeText.utf8.startIndex,
+      offsetBy: startPos
+    )
+    let endIdx = scopeText.utf8.index(
+      scopeText.utf8.startIndex,
+      offsetBy: endPos
+    )
+    var rewrittenText = scopeText
+    rewrittenText.replaceSubrange(startIdx..<endIdx, with: initializerText)
+    let parsedScope = Parser.parse(source: rewrittenText)
```
ahoppen: Let’s explicitly parse this as a `CodeBlockItemSyntax` here instead of building a full `SourceFileSyntax`.

```suggestion
    var parser = Parser(rewrittenText)
    let parsedScope = CodeBlockItemSyntax.parse(from: &parser)
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+
+  static func containsNode(
+    in node: Syntax,
+    withSpan span: Range<Int>,
+    matching text: String
+  ) -> Bool {
+    let finder = TargetNodeFinder(targetSpan: span, targetText: text)
+    finder.walk(node)
+    return finder.found
+  }
```
ahoppen: Just a thought: Would this read nicer as a static method on `TargetNodeFinder` (also direct you sooner to the type that actually implements it)

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// A visitor that checks if a syntax node with a matching span and description exists.
+private final class TargetNodeFinder: SyntaxAnyVisitor {
+  let targetSpan: Range<Int>
+  let targetText: String
+  var found = false
+
+  init(targetSpan: Range<Int>, targetText: String) {
```
ahoppen: Instead of `Range<Int>`, can you use `Range<AbsolutePosition>` to indicate that these are UTF-8 positions?

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+
+  override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
+    if found {
+      return .skipChildren
+    }
+
+    // Prune subtree traversal if full span does not overlap target range
+    let fullStart = node.position.utf8Offset
+    let fullEnd = node.endPosition.utf8Offset
+    if fullStart > targetSpan.upperBound || fullEnd < targetSpan.lowerBound {
```
ahoppen: Could this be simplified to the following (also, I think `contains` is slightly more restrictive and also sufficient)

```suggestion
    guard node.range.contains(targetSpan) else {
```

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,286 @@
...
+      return .skipChildren
+    }
+
+    let nodeStart = node.positionAfterSkippingLeadingTrivia.utf8Offset
+    let nodeEnd = node.endPositionBeforeTrailingTrivia.utf8Offset
+    if nodeStart == targetSpan.lowerBound && nodeEnd == targetSpan.upperBound {
```
ahoppen: This should also be simpler if you use `Range<AbsolutePosition>`

ahoppen (commented):
I have a couple more nitpicky comments comments, otherwise looks great to me 👍🏽

Sources/SwiftSyntaxCodeActions/InlineVariable.swift
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    // Prune subtree traversal if the node does not contain the target span.
+    guard node.range.contains(targetSpan) else {
+      return .skipChildren
+    }
+
+    let nodeStart = node.positionAfterSkippingLeadingTrivia
+    let nodeEnd = node.endPositionBeforeTrailingTrivia
+    if nodeStart == targetSpan.lowerBound && nodeEnd == targetSpan.upperBound {
```
ahoppen: Isn’t this equivalent to

```suggestion
    if node.trimmedRange == targetSpan {
```

ahoppen (commented):
One last nitpick, then we’re ready to ship.

ahoppen (approved):
Let’s 🚢 it

ahoppen:
@swift-ci Please test

somiljain2006:
@ahoppen I'm having trouble identifying the exact cause of the Windows CI failure. The build fails during CMake generation because the SwiftSyntax::SwiftLexicalLookup target cannot be found. I checked the top-level CMakeLists.txt, where SwiftSyntax is already brought in with find_package(SwiftSyntax CONFIG REQUIRED), but I haven't been able to figure out whether the issue is with how the dependency is declared in SourceKit-LSP or with the SwiftSyntax CMake package used by the Windows Stage2 toolchain. Can you tell me what I'm missing?

ahoppen:
Seems like we need to expose `SwiftLexicalLookup` before we can use it because this is the first change that’s using it. Opened https://github.com/swiftlang/swift/pull/91985 to do so.

ahoppen:
The Windows issue should be fixed now. Let's try again. 

@swift-ci Please test Windows

somiljain2006:
@ahoppen How about extending the inline variable code action to support var declarations?

ahoppen:
I don’t think extending it to `var` is all that important and reasoning about whether the variable is never modified is non-trivial (eg. need to also consider if it’s ever passed as inout to a function). But if you have a good proposal of how to implement it, I won’t object.

somiljain2006:
@ahoppen I was thinking of this approach: only offer the action when the variable has an initializer and all resolved references are read-only. I would reject direct mutations (=, +=, etc.), inout uses such as &x, and any cases where a known mutation pattern is detected. More complex cases like closure captures and mutable pointer APIs could remain unsupported initially.

ahoppen:
I think `+=` actually raises a good point where you can’t fully decide if a variable is read-only. You could define your own operator `++=` that takes a first argument as `inout`. Then, if you write, `x ++= [2]`, `x` would be mutated but you had no way of knowing that without running the type checker. 

I personally think that this would be a fairly big investment to do correctly with marginal pay-offs, so I wouldn’t do it.

PR #2760, opened 2026-09-05
Unblock new dependency of XCTest on Swift Testing for Windows

jerryjrchen:
Update compiler args for a test to include the Testing module alongside the XCTest module.

XCTest requires a Swift Testing dependency when built in the toolchain: https://github.com/swiftlang/swift-corelibs-xctest/pull/549

On Windows, the test "IndexedSingleSwiftFileTestProject" manually constructs the compilation flags in a way that previously only included the XCTest module, causing a test failure in the PR that adds the Swift Testing dependency.


ahoppen (approved):
Good with me, thanks.

PR #2763, opened 2026-09-13
Fix textDocument/codeLens method name in LSP Extensions doc and a few typos

toyeshhm:
Documentation-only change, no code touched.

The playgrounds section of `Contributor Documentation/LSP Extensions.md` tells clients to filter for `swift.play` CodeLens returned by the `textDocument/codelens` request. The LSP method is `textDocument/codeLens` (capital L), which is also the spelling `SourceKitLSPServer` reads from the initialization options, so a client copying the method name from the doc would send a request the server doesn't recognise. Fixed the casing.

While there, removed a doubled "the" in the `SourceKitOptionsRequest` comment in the same file, and fixed a few small slips in the VS Code note in `CONTRIBUTING.md` ("that that", "the settings is", and a capital "If" mid-sentence).


ahoppen (approved):
Thanks. LGTM.

Issue #2766, opened 2026-09-15
The BackgroundIndexingTests are very flaky on Windows CI


finagolfin:
### Swift version

6.4 and trunk

### Platform

Windows Server 2019

### Editor

None

### Description

We've been seeing a bunch of [flakes that break Windows CI runs](https://ci-external.swift.org/job/swift-PR-windows/61110/console) for pulls to all toolchain repos, causing us to have to rerun the CI till it doesn't flake:
```
Test Case 'BackgroundIndexingTests.testLibraryUsedByExecutableTargetAndPackagePlugin' started at 2026-09-15 06:09:40.438
<EXPR>:0: error: BackgroundIndexingTests.testLibraryUsedByExecutableTargetAndPackagePlugin : threw error "ResponseError(code: LanguageServerProtocol.ErrorCode(rawValue: -32001), message: "workspace/synchronize request timed out after 180.0 seconds", data: nil)"
Test Case 'BackgroundIndexingTests.testLibraryUsedByExecutableTargetAndPackagePlugin' failed (180.778 seconds)
Test Suite 'BackgroundIndexingTests' failed at 2026-09-15 06:12:41.216
```
May be related to #2346?

### Steps to Reproduce

Run the Windows CI

rintaro:
Some data from ~200 Windows CI builds. 

### The signature to grep for

`XCTUnwrap failed: expected non-nil value of type "CallHierarchyItem"` and `XCTAssertNotEqual failed: ("Optional([])") is equal to ("Optional([])")` are downstream effects. The actual cause only shows up as:

```
Terminating process using SIGINT because task was cancelled: [swiftc.exe -parse-as-library -module-name MyLibrary …]
🟨🟥⬜️ Finished with error in 120.00164099999999 seconds: Timed out
Updating index store failed: Timed out.
```

Counting `Updating index store failed: Timed out` matches the number of failed tests one-for-one in every affected build I looked at.

### It looks like a hang, not a slow machine

[...]

PR #2770, opened 2026-09-19
[DO NOT MERGE] Bound and report implicit module lock waits in tests

rintaro:
Investigating #2766 

All tests share one module cache, so a test can wait on a module lock
that another test holds. Cancelling an index or preparation task kills
the compiler, and `TaskScheduler.shutDown` cancels every in-flight task
when a server shuts down, so tests routinely kill compilers that may
hold a lock. `LockFileManager.processStillExecuting` only detects a
dead lock owner on Unix, so on other platforms the waiter blocks for
the full `-fimplicit-modules-lock-timeout`, which defaults to 90
seconds, before clearing the lock and building the module itself. That
exceeds what an index task can absorb within its own timeout, and the
resulting failure looks like slow indexing rather than lock contention.

Lower the wait to 30 seconds so that it cannot consume the index store
timeout, and pass `-Rmodule-build` so that `remark_module_lock_timeout`
identifies the module whose lock was contended.

Whether lock contention actually causes the observed stalls is not
established; the remark is what would confirm or rule it out.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

